PR #2, opened 2015-01-25
Update index.rst

piskvorky:
Fix two broken links.

Couldn't find link to repo in the [docs](http://honnibal.github.io/spaCy/) => went to check the source, found the sciPy link at the top didn't render.

Similarly, the link to commercial license lead to http://honnibal.github.io/spaCy/license.rst which is a 404.


honnibal:
Thanks!

PR #85, opened 2015-08-22
Add a script to generate the specials.json file

NSchrading:
Takes care of handling uppercase and missing apostrophe contractions.
Allows for adding exceptions to the generation of particular token + contraction cases.
When run, overwrites the specials.json file.
This has added new specials to the json file (only new uppercase / no apostrophe versions, I didn't add any entirely new words, although it is certainly possible). Please look over the specials it has generated and edit the generator as necessary to add / remove ones you want.

I also added some comments in for various tokens wondering why they don't have POS tags or lemmas in certain instances.


honnibal:
Thanks for this. Looks like it fixes the immediate problems.

Going forward it'd be nice to figure out a better way to both separate the code and data, and allow the generalisations.

Could follow the instructions here? https://github.com/honnibal/spaCy/blob/master/contributors/cla_template.md

Specifically, after reading the agreement, if you agree, please add a copy of the file named after your username, and filled in, to this commit.

Thanks!

PR #86, opened 2015-08-24
Correctly pass link_args in c_ext() in setup.py

vsolovyov:
`c_ext()` incorrectly passed `compile_args` twice, both to `extra_compile_args` and to `extra_link_args`. This pull request fixes that.

### DIFF
--- a/contributors/vsolovyov.md
+++ b/contributors/vsolovyov.md
@@ -1,0 +1,95 @@
+Syllogism Contributor Agreement
+===============================
+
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks, nice catch.

Unfortunately I have to ask for the contributor's agreement with the commit --- see here https://github.com/honnibal/spaCy/blob/master/contributors/cla.md

Also, did this fix a bug for you, or did you just read the code and see it was wrong?

vsolovyov:
I've just read the code and seen it was wrong. Give me 5 min for that contributor thing

PR #99, opened 2015-09-22
Force SSL for downloading English language data.

rw:
It would also be nice to have a checksum for this.


honnibal:
Thanks.

Actually the whole script is pretty inadequate. There's no caching mechanism, and no way to download part of the data but not all of it. If you'd like to suggest a rethink of it, I'd definitely be open to that.

For now, could you fill out the CLA? https://github.com/honnibal/spaCy/blob/master/contributors/cla.md Sorry about this, especially for such a small change.

PR #111, opened 2015-09-28
add doctests for website 'api'-section

henningpeters:


henningpeters:
- I marked failing tests with "xfail"
- please check whether we should remove the preambles (currently commented out) in the samples, e.g.
```
    # from spacy.en import English
    # nlp = English()
    doc = nlp('Some text.') # Applies tagger, parser, entity
```

PR #117, opened 2015-09-30
Changing deprecated codecs.open to io.open

alvations:
@honnibal I am looking for `goodfirstbug` to fix in spaCy as I am picking cython and reading Kurt Smith's book. I caught the deprecated `codecs.open` from the .py code. But if there are any other `goodfirstbug` to fix, esp. from the .pyx code, I will be glad to help out.

P/S: Also, it's great to see that @nltk is replacing the outdated maxent tagger with your perceptron tagger.

### DIFF
--- a/bin/init_model.py
+++ b/bin/init_model.py
@@ -20,7 +20,7 @@ from pathlib import Path
 
 from shutil import copyfile
 from shutil import copytree
-import codecs
+import io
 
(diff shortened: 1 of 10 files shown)


honnibal:
Hey,

Sorry I missed this before! Will merge this today. Thanks.

PR #126, opened 2015-10-06
Improve slicing support for both Doc and Span

tomtung:
Due to data being missing from the repository, I could not run some of the tests.

These changes are tested with: `py.test tests/tokens/test_tokens_api.py::test_getitem --models`


honnibal:
Thanks for this --- seems clear this fixes several bugs.

You probably noticed the code duplication between `Span` and `Doc` while doing this. In general I don't love inheritance, but it might be the best way forward. If you have a clever idea please let me know.

In the meantime, would you mind filling out the contributor's agreement? https://github.com/honnibal/spaCy/blob/master/contributors/cla.md
Even though we've changed to the MIT license, we're still trying to keep uniform copyright for the project, so that we can dual license if there's a need to in the future.

tomtung:
Filling out contributor's agreement soon...

honnibal:
Locating the logic in a `utils` function looks better than the duplication. I'll try to think of a nicer way to organize things, so that it's obvious where the implementations will live. But for now this is good.

Is the CLA proving a problem? We're trying to figure out the best compromises, so we appreciate feedback about what's resulting in unnecessary, annoying process.

tomtung:
Currently waiting for the legal okay from our company attorney, but should be fine :)

honnibal:
I'm going to go ahead and merge this. In the super unlikely scenario that there's a problem we can unwind and deal with it then. Cheers!

PR #129, opened 2015-10-07
Fix size of allocation when creating a pattern

chrisdubois:
Each pattern object currently contains two AttrValues rather than just one.

Found this while using spacy in a setting that uses tcmalloc, then building/running with sanitize flags.

No unit tests are included, unfortunately.

### DIFF
--- a/contributors/chrisdubois.md
+++ b/contributors/chrisdubois.md
@@ -1,0 +1,95 @@
+Syllogism Contributor Agreement
+===============================
+
+This Syllogism Contributor Agreement (“SCA”) is based on the Oracle Contributor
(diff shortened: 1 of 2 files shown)


honnibal:
Nice!!

Thanks very much for the process tip using tcmalloc. I've wanted to run spaCy under Valgrind for a long time, but it seems complicated because of Python. The process you recommend might be a nice easier way.

Can you fill out the CLA? https://github.com/chrisdubois/spaCy/blob/master/contributors/cla.md

If it's too much hassle I can apply the patch myself.

chrisdubois:
I've added a contributor file.

I'm glad the change might be useful. I'm looking forward to exploring the package more in the coming weeks!

PR #133, opened 2015-10-08
Fix typo

pquentin:
I know that first pull request need a  CLA, but that was the easiest way to say "here's a typo". Please apply this yourself if this is possible?


honnibal:
Actually for the site source, there's no problem accepting a patch --- we only need the CLA for the code.

Thanks!

PR #138, opened 2015-10-13
Python 2.7 Win64 automatic build with appveyor

maxirmx:
* Build scripts 
    * .appveyor.yml
    * appveyor subdirectory

* setup.py
    * Compiler options management through build_ext subclassing. This may be a kind of heavy-weighted but it is the only MinGW\cygwin 100% friendly method that I know

* Readme to support Windows manual build
* Changes to .gitignore to make Windows life easier


honnibal:
This looks nice. Thanks for the compiler management stuff --- this should resolve another ticket, on using GCC in OSX.

maxirmx:
I believe in OSX     `if sys.prefix == 'darwin'`     shall be   `if sys.platform.startswith('darwin')`  
Cannot check right now

henningpeters:
Here are the values from OSX 10.10.5

```
>>> sys.prefix
'/Users/hp/...'
>>> sys.platform
'darwin'
```

According to python docs:
```
sys.prefix
A string giving the site-specific directory prefix where the platform independent Python files are installed.
sys.platform
This string contains a platform identifier
```

Hence, ```sys.prefix == 'darwin'``` is probably always False.

maxirmx:
Yes, `if sys.prefix == 'darwin'` shall be replaced with `if sys.platform.startswith('darwin')`

I do not want to make this fix without test environment

honnibal:
K, made the change.

PR #140, opened 2015-10-14
Fix docs to match code (like_number)

bittlingmayer:
s/like_num/like_number

It should match language.py.

Also, `is_oov()` is not actually accessible yet, is it?

If it is implemented, then I can add it to language.py.  Otherwise I can rm it from the docs in this PR.

### DIFF
--- a/website/src/jade/docs/_api.html
+++ b/website/src/jade/docs/_api.html
@@ -246,7 +246,7 @@ array([[11880],
       </li>
       <li><span class="declaration"><code>like_url</code></span> Does the word resembles a URL?
       </li>
-      <li><span class="declaration"><code>like_num</code></span> Does the word represent a number? e.g. “10.9”, “10”, “ten”, etc
+      <li><span class="declaration"><code>like_number</code></span> Does the word represent a number? e.g. “10.9”, “10”, “ten”, etc
       </li>
       <li><span class="declaration"><code>like_email</code></span> Does the word resemble an email?
       </li>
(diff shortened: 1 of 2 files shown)


bittlingmayer:
Also, the example `en.like_number(u'ten')` actually returns `False`.  Should I remove it from the docs?  Or mention it as coming soon?

bittlingmayer:
Also, besides the forms like 'one', ... 'ten' etc, this feature seems like a good candidate for either a universal utility (language agnostic eg `from spacy.i18n import i18n`) or partial language support (because it already works for, say, Spanish).

honnibal:
Thanks for the suggestions on this.

The feature is actually called `LIKE_NUM` in the `spacy.attrs` enum, which is why that name gets used in the docs. You can see the current implementation here:

https://github.com/honnibal/spaCy/blob/master/spacy/orth.pyx#L106

The way these features work is a little bit round-about. What happens is that the `Vocab` class receives a dictionary mapping attribute IDs to functions from the `orth` string to the attribute value. When the language is loaded, it loads pre-computed values for the features from disk, with some additional values computed off-line. For instance, the `prob` and `cluster` values are read from external sources. The `bin/init_models.py` script does this work.

For about 98% of tokens, their string is hashed and a pointer to a struct holding these pre-computed features is fetched from the `Vocab`. So, we can have these features doing lots of work, and even be sort of expensive, because we so rarely actually have to re-compute them. For the 2% of out-of-vocabulary tokens, we do have to compute these things, though. That's when the functions stored in `Vocab.get_lex_attr` are accessed.

The actual implementations for these things are mostly in `spacy/orth.pyx`. But there's an additional indirection: the `get_lex_attr` dict references a bunch of static methods on `Language`. I did this so that you'd be able to just replace some of those defaults. Maybe that was a bad idea; I haven't needed it yet.

PR #145, opened 2015-10-20
better error reporting, cleanup

henningpeters:


henningpeters:
Latest commit should fix all pending immediate problems.

PR #159, opened 2015-11-02
fixed error when printing unicode

andreasgrv:
Corrected __str__ and __repr__ to return bytes - error was raised when printing strings that actually contained unicode characters - example: café . Added doctests to check printing for doc, span and token.

### DIFF
--- a/spacy/tests/print/test_print.py
+++ b/spacy/tests/print/test_print.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+import pytest
+
+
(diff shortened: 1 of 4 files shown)


honnibal:
Is this correct for both Python 2 and Python 3?

I think for Python 3 `__str__` should return a unicode object.

andreasgrv:
Good catch, one sec, i'll check

andreasgrv:
Yup, It's not compatible with python3 -_- , I'll review it

PR #161, opened 2015-11-02
fixed error in python2 when printing unicode

andreasgrv:
Should now be py[23] compatible - tested it locally. Used six to check python versions.


honnibal:
Nice, thanks.

PR #167, opened 2015-11-04
Update spans automatically after each merge

andreasgrv:
Used some of the existing tests + added a few simple ones. Not sure what more to check - would be wise to add some more tests about subtree - lefts and rights to make sure merging doesn't do something bad. However, I haven't really changed much in doc.pyx merge function, apart from assigning to token.lex later on and moving the beginning into another function. Not sure what else could go wrong now.

PS.
I think you used something like s/spans/span/g and meddled with setup.py tests/spans in setup.py ^_^


andreasgrv:
It passes all tests locally - I don't understand how it gets this wrong - this is before any merging happens:
``` python
        tokens = EN(u'Los Angeles start.') 
        assert len(tokens) == 4 
        assert tokens[0].head.orth_ == 'Angeles' 
```
Local test:
``` python
In [1]: from spacy.en import English

In [2]: nlp = English()

In [3]: tokens = nlp(u'Los Angeles start.')

In [4]: tokens[0].head.orth_
Out[4]: u'Angeles'
```

andreasgrv:
Found a bug, will add better checks too.

honnibal:
Can you rebase onto the current master? I did rename spans.pyx to span.pyx, for consistency. There were a couple of other refactors too.

andreasgrv:
Sure. Currently checking out bug, will pull when done.

PR #168, opened 2015-11-05
Update spans automatically after each merge

andreasgrv:
Couldn't get the rebase to work nicely - So worked on a new fork - will try and keep it tidy this time. Need to dive deeper into git. Fixed bug, should be ok now.

PS.

Shouldn't line 166 in setup.py be:
 'spacy.tests.spans',
instead of 
'spacy.tests.span'
?

### DIFF
--- a/spacy/tests/spans/test_merge.py
+++ b/spacy/tests/spans/test_merge.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 import pytest
 
-@pytest.mark.models
 def test_merge_tokens(EN):
     tokens = EN(u'Los Angeles start.')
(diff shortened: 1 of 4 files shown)


honnibal:
Hmm.

I've gone through this now and I think I prefer a different implementation, that is less total code, and doesn't introduce the potential performance problems of making `Span.start` and `Span.end` properties.

I've pushed to master, but there are still bugs that I'm working out. Thanks for your work on this.

andreasgrv:
A property was indeed overkill performance-wise, but by using a property I had one thing less to worry about while digging into the code. Thank you for fixing this :+1:

PR #202, opened 2015-12-07
access model via sputnik

henningpeters:


spacy/language.py
```diff
@@ -134,73 +135,44 @@ def default_ner_labels(cls):
...
-    def default_data_dir(cls):
-        return path.join(path.dirname(__file__), 'data')
-
-    @classmethod
-    def default_vocab(cls, data_dir=None, get_lex_attr=None):
-        if data_dir is None:
-            data_dir = cls.default_data_dir()
+    def default_vocab(cls, package=None, get_lex_attr=None):
+        if package is None:
+            package = default_package()
```
honnibal: The `Language.default_data_dir()` thing was fiddly, but there's an advantage to it: when you subclass, you have a way to specify a new default. If we read the default from this `default_package()` function, the subclass can't change that.

It could be that we don't need this capability. But that's why I did it like that.

honnibal:
I'm a little bit concerned about the API difference we're getting here.

Before there was:

`Language.from_dir`

We never really made this convenient, because of the awkwardness of installing the data in different places.

But now we have:

`Language.from_package`

So the user has to instantiate this package object.

Given that this is the library entry-point, we need to have something a bit easier. Is it possible to have something like this:

    nlp = spacy.en.English('<some resource descriptor>')
    nlp = spacy.Language('en', '<some resource descriptor>')

Inside `English.__init__`, we would try to create a Package object from the string, if a string is what the user gave us. If the user gives us a `Package` object we continue on our way. Fallback to default if nothing specified.

henningpeters:
a model can now be specified:

1) by a path to the model directory (DEPRECATED)
  - ```Language(data_dir='path/to/data')```

2) by a language identifier (and optionally a package root dir)
  - ```Language(lang='en')```
  - ```Language(lang='en', data_dir='spacy/data')```

3) by a model name/version (and optionally a package root dir)
  - ```Language(model='en_default')```
  - ```Language(model='en_default', version='1.0.0')```
  - ```Language(model='en_default', version='1.0.0', data_dir='spacy/data')```

```English()``` works the same, but with lang='en' preset.

henningpeters:
Btw: is ```templates.json``` still un use?

honnibal:
Let me check again that I've understood how this works properly.

Let's say our back-end was a DB. Would this data model work?

* Root resource is the table
* Each model is a row
* Model version is the primary key
* Model language is a column
* Boolean column package_default
* Boolean columns for en_default, it_default, fr_default etc

There's a global default root resource: a directory `data/` that's a sibling to `spacy.__init__.__file__`. But you can point to a different root resource instead. In future functionality, the root resource need not be a file-system.

Within the root resource, models are uniquely identified by version string. You can also ask the resource to supply its default model, or the model that's default for a given language.

You can't ask for version X of language Y, because there's only one model with version X, and it has exactly one language.

Question: Does the same root resource know about models for different *spaCy* versions? If so, does spaCy (silently) send its version number, as part of the "SELECT" statement it's making for the model?

henningpeters:
Yes, your database design would almost work, small correction: you forgot the model name and we assume that all models you can view are ones that are compatible with your current spaCy installation (does this answer already your question?).

Hence, within the root resource, models are uniquely identified by _name_ and version string. This means there could be also models that are not default model in any language, but still addressable by name/version.

In future functionality root resource could be also outside the filesystem (e.g., REST API), but we currently lack a mechanism to flexibly specify such resources. I overloaded the semantics of ```data_dir``` specifically not to invent such a scheme, but to reuse the path-spec that already existed.

honnibal:
Okay so it's more like:

* The root resource is the DB connection
* The spaCy version currently running specifies the table
* The model name is a string that serves as the primary key

Is that more like it?

henningpeters:
Yep, that works.

honnibal:
Okay, so does this make sense?

    Language(model='en_default', version='1.0.0')

The version is specified by the spaCy code currently executing, isn't it?

henningpeters:
Yes, it does. There could be still multiple compatible versions of the same model. Example: you created a model that contained a "bug". The bugfix doesn't require to release a new spaCy version, but only a new model version that ships as new default. At the same time you might want to pin your spaCy instance to a particular model version to avoid changes in your application even if a new model version is released. Hence, there is still the need for selecting a model by version. The check for the spaCy version compatibility is there to avoid mistakes and offering functioning upgrade paths.

honnibal:
Hmm. It'd be pretty nice to drop the extra parameter, though!

Why not just concatenate the version onto the model string, if a user wants to specify a version constraint?

Structured strings aren't a great solution, but this is how it works for semver, and there might be a use-case for specifying >= , <, etc.

Scenario: Let's say you're building something, and you have two teams. One team develops new models for spaCy, e.g. with new word vectors, new named entity features for your domain, etc. The other team consumes those models. The first team wants to just crank out models, but the second team might not be ready to consume them right away, e.g. a new model might introduce new entity types, or it might cause you to retrain your models.

henningpeters:
Agree, this is much cleaner (see new commit). Sputnik already works this way internally, hence it was easy to change.

PR #222, opened 2016-01-13
integrate with sputnik

henningpeters:
Sputnik now includes a new DirPackage class that offers a Package interface around a model directory. You can now instantiate spaCy as follows:

```
1) default
nlp = spacy.en.English()

2) model directory without dealing with packages (e.g., for development)
nlp = spacy.en.English('my/directory')
nlp = spacy.en.English(data_dir='my/directory')

3) with a different model
nlp = spacy.en.English('fr_default')

4) with a relocated package base
nlp = spacy.en.English(model='en_default', data_dir='new/root')

5) with an existing package
package = sputnik.package(<app_name>, <app_version>, 'my_model >=1.0.0', data_path='packages')
nlp = spacy.en.English(package)
```

Tokenizer/Tagger/Vocab/etc load() accepts string, directory or package
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,17 +1,14 @@
 {
-    "name": "en_default",
-    "version": "0.100.0",
-    "description": "english default model",
+    "name": "en_test",
+    "version": "1.0.0",
+    "description": "english test model",
     "license": "public domain",
     "include": [
-        "deps/*",
-        "ner/*",
-        "pos/*",
-        "tokenizer/*",
-        "vocab/*",
-        "wordnet/*"
-    ],
-    "compatibility": {
-        "spacy": "==0.100.0"
-    }
+        ["deps", "*"],
+        ["ner", "*"],
+        ["pos", "*"],
+        ["tokenizer", "*"],
+        ["vocab", "*"],
+        ["wordnet", "*"]
+    ]
 }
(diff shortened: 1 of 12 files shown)


henningpeters:
Please have a look at the error messages at https://github.com/henningpeters/spaCy/blob/9b75d872b08fe66a5a5c0a8534e36c0e252d3089/spacy/util.py#L32 whether they need tuning.

henningpeters:
Please have a look at the error messages at https://github.com/honnibal/spaCy/blob/d9471f684fbfb4776bfaf2a4e2265c19dd758db2/spacy/util.py#L32 whether they need tuning.

honnibal:
Error messages look good, better than the current average standard.

My eventual goal is for error messages to almost always expose all relevant local state, that led to the error. If you can think of a neat way to write that into these messages, that might make them even better.

henningpeters:
Agree, will think of something. Are you fine just writing to stderr or did I miss some code that can be reused for error output?

honnibal:
I was thinking to integrate it into the text.

I haven't thought this through properly, but I have this idea of using the
errors-are-objects thing more, so that the state is attached to the object.
Then the class knows how to write its message, by formatting its arguments.

Note that this isn't how the rest of Python really does things. So, it
might be bad, to the extent that it's weird. But it's always bugged me that
there's all this state, and the error message could expose it but just
doesn't.

PR #223, opened 2016-01-15
refactored data_dir->via, add zip_safe, add spacy.load()

henningpeters:
```
a model can be specified:

1) by calling a Language subclass
  - spacy.en.English()

2) by calling a Language subclass with via (previously: data_dir)
  - spacy.en.English('my/model/root')
  - spacy.en.English(via='my/model/root')

3) by package name
  - spacy.load('en_default')
  - spacy.load('en_default==1.0.0')

4) by package name with a relocated package base
  - spacy.load('en_default', via='/my/package/root')
  - spacy.load('en_default==1.0.0', via='/my/package/root')
```


henningpeters:
Now we can choose between load(via) and from_package(pkg) calls. Also, English() now has a ```package``` param. Result: ```via``` and ```package``` don't collude anymore.

PR #265, opened 2016-02-18
add tokenizer files for German, add/change code to train German tagger

wbwseeker:
- add files to specify rules for German tokenization
- change generate_specials.py to generate from an external file (abbrev.de.tab)
- copy gazetteer.json from lang_data/en/

- init_model.py
	- change doc freq threshold to 0
- add train_german_tagger.py
	- expects conll09-formatted input


honnibal:
Looking good :). We just have to merge this `german_tagger.py` script and we'll have the tokenizer and part of speech tagger done!

PR #278, opened 2016-03-01
replace codecs.open with io.open

elyase:


honnibal:
Thanks! We need to get these scripts tested across platforms properly.

PR #280, opened 2016-03-03
German parser

wbwseeker:
This adds functionality to deal with non-projective dependency structures.
It implements the pseudo-projective parsing idea from Nivre & Nilsson 2005.
This enables the parser to be trained on non-projective treebanks.


honnibal:
Geil :)

PR #284, opened 2016-03-10
Added reloadable English() example for inventory counting

olegzd:
Added an example of reusing the English() object through using module reloading. On top of that, added an example of determining inventory counts for a few test sentences, and the way to interact with Doc objects, finding wether they're a noun with descriptive compound objects ('fresh ground beef' VS 'beef')

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,5 +96,8 @@ setup.py
 # Windows local helper files
 *.bat
 
+# Mac OS X
+*.DS_Store
+
 # Komodo project files
 *.komodoproject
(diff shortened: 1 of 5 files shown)


honnibal:
Hey,

As I said over email, thanks for this :)

Some binary files seem to have been caught in the commit though. You'll need to get rid of them, and make sure they don't appear in the history. Unfortunately I can't remember how to do this off the top of my head.

olegzd:
> Some binary files seem to have been caught in the commit though.

Whoops! When I forced added to the examples/ folder it included the binary .pyc files in the commit. Updated :)

honnibal:
You've still got a .DS_STORE file in there.

olegzd:
Removed the .DS_Store + added an entry to .gitignore so no one else makes the same mistake.

PR #287, opened 2016-03-11
add function to Token for setting head and dep (and dep_) 

wbwseeker:
change PseudoProjectivity.deprojectivize to use these functions

previously, the deprojectivization was using Doc.from_array to change the head of tokens.
Doc.from_array calls set_children_from_head() which introduces sentence boundaries.
however, this function assumes projective trees which is not true anymore, so it was introducing
incorrect sentence boundaries.
Token.head and Token.dep (and Token.dep_) can now be set from the outside.
deprojectivize doesn't call from_array anymore. from_array is still unsafe if the tree is non-projective.

### DIFF
--- a/spacy/syntax/nonproj.pyx
+++ b/spacy/syntax/nonproj.pyx
@@ -118,15 +118,18 @@ class PseudoProjectivity:
         # reattach arcs with decorated labels (following HEAD scheme)
         # for each decorated arc X||Y, search top-down, left-to-right,
         # breadth-first until hitting a Y then make this the new head
-        parse = tokens.to_array([HEAD, DEP])
-        labels = [ tokens.vocab.strings[int(p[1])] for p in parse ]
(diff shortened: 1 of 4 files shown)


honnibal:
It's a shame to have the left/right branches duplicated, but I don't see a better way. I think the abstraction over the two would produce code that's more difficult to read.

The care with comments and tests is appreciated. I've introduced a lot of subtle bugs with this stuff.

Should we rethink the way left and right children are stored, now that we have non-projective trees? The left and right edges are convenient when the tree is projective because all subtrees are contiguous. But with non-projective trees it's a bit more complicated.

wbwseeker:
I think I thought of a slightly simpler way of doing this that should also be less costly. Let me change it and update the pull request.

about storing the left/right children: we could for example store explicit arrays of pointers to all children. This would increase the memory footprint but we wouldn't have to search through the entire subtree every time.

PR #305, opened 2016-03-24
multiple langs in download script

henningpeters:


syllog1sm:
Maybe a bit nitpicky, but I'm reluctant to have meaningful defaults hidden away in the about.py. I also think it makes the code in `load()` quite indirect.

Can we make the `English` and `German` classes own their defaults? I'd also rather have something like a `languages` dict, which maps the ID strings to the classes. Each language could be responsible for adding itself to the dict. It'd probably be nice to have an explicit function like `register_language`, which just adds to the global. This way users can register their subclass, so that `load()` works correctly with it.

syllog1sm:
Changes to `download.py` look good.

henningpeters:
The only defaults I added are `module` and `class`, but those can't live in English/German, because they are needed to find English/German. `package` was living in about.py already before, hence no change.

syllog1sm:
I don't think it's so much extra code, and we'll need the mapping from letter codes to classes, both for our use and as part of the API. So we may as well put it in the right place.

I think the natural place for it is either in `spacy.__init__` or in `spacy.language`. I would prefer it to be in the `__init__`.

henningpeters:
Looks like you'll get into circular dependency problems as language and util depend on each other. You can't put it into __init__ either as you cannot import from __init__, can you? Thought a bit about a language class self-register feature, but all solutions that I came up with had too much import magic.

syllog1sm:
I'm not seeing where the circular dependencies arise yet. I'm also not sure why we need any import magic. I see what you mean about the `__init__.py` problem, and yes that's nasty. So how about this:

In language.py we have:

```python

LANGUAGES = {}

def register_lang(name, clas, overwrite=False):
    global LANGUAGES
    if name in LANGUAGES and not overwrite:
        raise KeyError("TODO msg")
    LANGUAGES[name] = clas

def get_lang(name):
    # TODO: Behaviour on missing?
    return LANGUAGES[name]
```

The in spacy/en/__init__.py:

```python
from ..language import Language, register_lang
class English(Language):
    name = 'en'
    ...

register_lang(English.name, English)
```

Then in spacy/__init__.py:

```python

from .language import get_lang
from . import util

def load(name, vectors=None, via=None):
    clas = get_lang(name)
    package = util.get_package_by_name(name, via=via)
    vectors_package=util.get_package_by_name(vectors, via=via))
    return clas(package=package, vectors_package=vectors_package)
    return English(		 +    if name not in __models__:
```

henningpeters:
Looks good, but how do you ensure that spacy{en,de}/__init.py gets executed before spacy/__init__.py?

syllog1sm:
True. You'd have to import them into the `__init__.py` script. Which means that we should just call `register_lang('en', English)` from within the `__init__.py`.

henningpeters:
I thought its adding unnecessary latency to loading time, but it seems thats not significant. Due to circular dependency between `language.py` and `util.py` you cannot place the language registration in `language.py`, I've put it now in `util.py` until we find a better place.

Keep in mind that you might still want to express versions in `spacy.load()`, hence I added some logic to handle that.

syllog1sm:
Okay thanks, looks good.

Final quibble: it's okay for the `register_lang` and `get_lang` functions to be implemented in `util`, but having user-facing API in `spacy.util` is unideal. The easy fix is to just do a relative import into `spacy/__init__.py`. This way users get:

```python

import spacy

spacy.register_lang
spacy.get_lang
spacy.get_package
```

This looks quite good now, imo. Users can easily subclass `Language` themselves, and register their subclass. Then they can use `spacy.load()` and it'll work just the same as our subclasses. They can even patch in their own replacements to the languages we implement, replacing 'en', 'de' etc.

henningpeters:
Language registration and package retrieval functions are now relatively imported into `__init__.py`:

```
import spacy

spacy.set_lang_class
spacy.get_lang_class
spacy.get_package
spacy.get_package_by_name
```

works. More use cases:

```
import spacy
spacy.load('en>=1.0.0')
spacy.load('en')
spacy.load('de')
spacy.load('abc')  # raises exception
```

syllog1sm:
Nice! Merging this. While you're looking at this bit of code, it might be a good time to look at the `parser`, `tagger`, and `entity` arguments we wanted for `load()`.

PR #306, opened 2016-03-24
add German noun chunk functionality

wbwseeker:
This adds an iterator to provide noun chunks also for German.
To make this work with multiple languages, the vocab object now has a lang property that comes from the respective language class that was used to create the vocab object.
The parser uses the lang property to set the appropriate noun chunk iterator when a document is being parsed.

Because of this change, this version is not compatible to the current English model anymore, which will need to be retrained. We probably want to wait with a merge until this is done. (also all the test fail that presuppose the English model)

While I was at it, I also implemented is_bracket, is_quote, is_left_punct, and is_right_punct in orth.pyx and added the respective constants to attrs.pxd. I think this doesn't actually have an effect just yet because these features aren't being used. I think they will help us predict better sentence boundaries around punctuations symbols like quotes and brackets.

oh, and I undid the changes I introduced to init_model.py in a previous PR. It should now work as it used to.


syllog1sm:
Looks good!

I'm trying to think of a better name for `DocIterator`. Its name sounds to me like it should be yielding `Doc` objects, rather than iterating over the `Doc`.

Also, should the `sents` iterator be done the same way? It might be a nice way to allow a lightweight SBD, that doesn't require the parse.

PR #307, opened 2016-03-24
remove internal redundancy and overhead from StringStore

scoder:
This should generally speed up the `StringStore` class, both for loading and for general usage. It does not change anything about the strings file format, nor does it remove the load/store via JSON.


syllog1sm:
Looks good. What's `cython.final` do?

scoder:
`@cython.final` declares a method as non-overridable, thus allowing Cython to call its underlying C function directly rather than going through a method dispatch. The C compiler can then see which function is called and optimise the call. It's usually a good thing for internal/private methods.

PR #312, opened 2016-03-28
add restrictions to L-arc and R-arc to prevent space heads

wbwseeker:
Proposal for additional restrictions for LeftArc and RightArc that should prevent the parser from predicting space tokens as heads of other tokens.

I think there shouldn't be any weird interactions between the transitions. 
First, I thought several space tokens at the end might be a problem. If there are several space tokens at the end of the buffer, this could get the parser into a dead end because it may get into a situation where it is not allowed to shift but also cannot introduce arcs. But I think this won't happen because there can't be consecutive space tokens, they would be merged into one. And even if this was so, unshift could fix it.
Can you think of any boundary case where additional restrictions may lead to desaster?

Can there be any interaction with the cost scheme or arc reachability? I think not because the heads of space tokens aren't specified in the training data and as such have no influence on the costs.


syllog1sm:
Haven't had time to go through this in detail, so just quickly:

I've had some trouble with this before --- a similar problem was arising with the way spaces were being attached around sentence boundaries.

I think your logic seems sound, but I'd really like to see some tests for these things. Could you add a test that fails for the current code, that this fixes?

I follow your logic about how we can't have multiple space tokens given the current tokenization rules. But this seems a bad thing to rely on in the parser. So, if you could find a way around this, that'd be quite a bit better.

honnibal:
I've lost track of this one a little bit --- I think we talked some more, and you reiterated why you preferred to have this as a transition constraint, rather than model guidance. I'm on board with that, but I'd still really like to have tests with the patch.

wbwseeker:
As I said on slack, I changed the system so that space tokens are treated entirely inside stateC.fast_forward(). Now, space tokens are always attached to their last preceding non-space token except in the beginning of a document, when they are attached to the first following one. In input that consists exclusively of space tokens, the last one serves as the head of all others.

The advantage of this is that the attachment of space is rule-based and thus never decided by the statistical model. It is not possible to use space tokens as heads for other tokens anymore. The statistical model never sees space tokens on the stack or the buffer. It may see them in the feature set as left-edge/right-edge. We might want to check whether this behavior is desirable (I think so) or should be augmented by additional features that look for the left-most non-space token as well (I think so too).
I removed the additional restrictions from LeftArc and RightArc that I proposed originally because they are not necessary anymore.

This change made some small changes in Break.is_valid() necessary. Before, the parser could only introduce sentence breaks between adjacent tokens. Now, it can also do it if there are only space tokens between two adjacent real tokens. Space tokens can never be sentence starts except if they are the first token in a document.

I also added some tests to test the functionality. Let me know whether they meet your expectations. I can change them or add more. I also ran it on some English test and it looked fine. We probably need to retrain the models for English (and German) though.

Issue #322, opened 2016-04-06
inconsistent sentence boundaries before and after serialization


thricedotted:
I've been running into a problem where a parse's sentence boundaries change after converting it to a bytestring:

```python
> text = u"I bought a couch from IKEA. It wasn't very comfortable."

> parse = nlp(text)

> parse_from_bytes = Doc(nlp.vocab).from_bytes(parse.to_bytes())

> [s.text for s in parse.sents]
[u"I bought a couch from IKEA. It wasn't very comfortable."]

> [s.text for s in parse_from_bytes.sents]
[u'I bought a couch from IKEA.', u"It wasn't very comfortable."]

> parse.to_bytes() == parse_from_bytes.to_bytes()
True
```

This happened to be one where the sentence boundaries were more correct *after* the conversion, but I have other examples where ~~it actually breaks the parse~~ EDIT: the parse is already broken; in the original, two ROOTs appear in the same sentence, whereas in the from_bytes version, the ROOTs are forced to be in different sentences.

Not sure if this means there is a bug in the serialization or initial sentence boundary detection!

honnibal:
Thanks, there's definitely something wrong here.

honnibal:
(atting @wbwseeker because we were talking about this bug on Slack)

I've just gone back over the code and realised that I'd forgotten how my transition system works, with respect to the Break transition. It's really not written down anywhere, and it's in fact rather different from the paper that the code cites as inspiration. So, I'll give some background here.

The intention is that all sentences are connected trees, so there's one word per sentence that is its own head, and that has the label ROOT. Mostly, sentence boundaries are inserted by the Break action. The Break action flags the first word of the buffer as the start of the next sentence. The parser then acts as though the buffer is exhausted until the stack has only one word. That is, it continues parsing using the "Unshift" action to connect the stack, until only one word is left. That word then becomes the root of the sentence, it's popped, and parsing continues.

There is however another way that we can get a sentence boundary. If the buffer is fully exhausted (i.e. we're really at the end of the sentence), the parser might end up with two root words on the stack. It's then allowed to join them with a left or right arc, using the label ROOT. This should be interpreted as saying "These are both root words, of different sentences. Insert a sentence boundary between them." In the code, there's a flag `USE_ROOT_ARC_SEGMENT` that toggles this behaviour.  It was used as a baseline strategy when I was experimenting with the definition of the Break transition. 

[...]

PR #346, opened 2016-04-21
introduce sentence boundaries for additional root tokens

wbwseeker:
added the one line that @honnibal suggested to fix issue #322

### DIFF
--- a/spacy/syntax/arc_eager.pyx
+++ b/spacy/syntax/arc_eager.pyx
@@ -399,31 +399,34 @@ cdef class ArcEager(TransitionSystem):
         cdef TokenC* orig_head
         cdef int new_edge
         cdef int child_i
-        cdef TokenC* head_i
+        cdef int head_i
         for i in range(st.length):
             if st._sent[i].head == 0 and st._sent[i].dep == 0:
                 st._sent[i].dep = self.root_label
             # If we're not using the Break transition, we segment via root-labelled
             # arcs between the root words.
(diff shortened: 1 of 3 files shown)


wbwseeker:
the parser should not get into a loop again.

There is one issue with this way of introducing additional root nodes, and this is serious:
if we have a parse like this: heads [0,-1,-2,-3], deps [ROOT, ROOT, nsubj, nsubj]
arc_eager.finalize_state() will make the second token it's own root and therefore sentence
by changing it to [0,1,-2,-3] with [ROOT, ROOT, nsubj, nsubj]

This is a non-projective structure and the sentence boundary at the second token will
disconnect tokens 2 and 3 from their head 0. This will not be a valid state, but there is currently no mechanism to prevent this. The only reason why this is probably not showing currently is because the statistical model will rarely predict such a structure for proper input.
However, this is something we shouldn't rely on because the model could be used to parse unfamiliar input and will then break on it.

honnibal:
That structure is familiar --- I thought about that when doing the joint segmentation and parsing for speech. I thought I had constraints which prevented this, but obviously not. I'll have a look at this.

honnibal:
Definitely no constraints.

I'll start digging through the history to get a sense of what I was thinking and what options I tried.

wbwseeker:
cool. we should probably merge the pull request already anyway, because the loop bug is currently on the master branch

PR #384, opened 2016-05-16
Fix get_lang_class parsing

daylen:
We want the get_lang_class to return "en" for both "en" and "en_glove_cc_300_1m_vectors". Changed the split rule to "_" so that this happens. See https://github.com/spacy-io/spaCy/issues/338 for details

### DIFF
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -23,7 +23,7 @@ def set_lang_class(name, cls):
 
 
 def get_lang_class(name):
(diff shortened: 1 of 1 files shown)


honnibal:
Thanks!

honnibal:
Ah damn. Merged too soon. This caused some tests to fail, when the `lang_class` comes with a semver string.

I don't have time to look at this now, but if you figure it out, please send another PR.

PR #422, opened 2016-06-10
plac version in requirements

RahulKulhari:
current new version of plac(0.9.3) is creating problem but it is working <0.9.3


pquentin:
It should work with 0.9.4 too.

PR #526, opened 2016-10-16
Added Gitter badge, alt text for all badges to README.rst

### LINKED ISSUE #487: Gitter chat for spaCy
As the community grows, it might be useful to have a gitter chat room for spaCy (http://gitter.im). 

Any repo admin should be able to get it going with a couple of clicks, and anyone with a github account can sign in and lurk/chat/ask questions. I've found it quite useful for other projects I'm part of, and highly recommend we do it - especially when the community can answer many questions in real time.

cc @honnibal

crawfordcomeaux:
Out of curiosity, I wonder if there's any sort of preferred grammar for alt text when it comes to accessibility.

Like, would someone using a screen reader prefer "Our Gitter chat room" or "Click here for our Gitter" over "spaCy on Gitter"?

Just a random thought.

Possibly fixes #487 (though may make sense to open a separate issue for connecting Gitter to Slack/IRC, if so)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,15 +9,23 @@ the MIT license.
 
 .. image:: http://i.imgur.com/wFvLZyJ.png
     :target: https://travis-ci.org/explosion/spaCy
-
+    :alt: spaCy on Travis CI
+    
 .. image:: https://travis-ci.org/explosion/spaCy.svg?branch=master
     :target: https://travis-ci.org/explosion/spaCy
+    :alt: Build Status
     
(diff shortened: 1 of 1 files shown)


ines:
Good catch, thanks! Also sorted out the Gitter mess and set up a proper channel for the Explosion org and a room for spaCy – this way, all discussion stays in one place and we can open up more rooms in case people want to discuss sense2vec, displaCy etc.

PR #545, opened 2016-10-20
added ujson dependency to setup.py

adamhadani:
Noticed after updating to latest spaCy I got breakage due to missing ujson - added this dependency in setup.py here


adamhadani:
FYI @syllog1sm

honnibal:
Oops. Thanks.

PR #551, opened 2016-10-22
Add Windows Compiler installation instructions

sixhobbits:
The ReadMe currently says that Windows Spacy Users should install a full Visual Studio version in order to compile Spacy. Some might prefer to only install the compiler provided by Microsoft (or MinGW if this is confirmed to work?). I added a link to the Wiki that points to the various Microsoft downloads, as these are quite difficult to find otherwise.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -179,6 +179,11 @@ Install a version of Visual Studio Express or higher that matches the version
 that was used to compile your Python interpreter. For official distributions 
 these are VS 2008 (Python 2.7), VS 2010 (Python 3.4) and VS 2015 (Python 3.5).
 
+If you don't want to install the entire Visual Studio, you can install a
+stand-alone compiler. Make sure that you install the correct version for 
+your version of Python. See https://wiki.python.org/moin/WindowsCompilers for 
+links to download these.
+
 Run tests
 =========
 


honnibal:
Thanks! I'd like to get back to building binaries for spaCy, but in the meantime these instructions are good.

On another note: I happened to come across [your book on Flask](https://realpython.com/blog/python/flask-by-example-part-1-project-setup/), which looks very nice. How interested are you in spaCy?

sixhobbits:
Thanks, Matthew. That very popular Flask blog post is not actually mine -- an unfortunate name clash, but my publishers were doing a "By Example" series so they decided to go ahead with a clashing title anyway. Mine is https://www.packtpub.com/web-development/flask-example 

I'm pretty interested in spaCy -- definitely solves many of the NLTK pain-points. I'll probably be using it for my upcoming master's thesis. Why do you ask?

honnibal:
Oops! Fixed the link. I was looking at your book and didn't realise the post wasn't by you.

I ask because we really need this sort of material about spaCy. I know a book is a big commitment, but every part helps. I'm sort of the worst person to write this, because I'm the furthest away from the people who need it. The chicken-and-egg problem is that until the material gets written, not enough people know the content at all.

What would you need from me to help you write "spaCy by example"?

sixhobbits:
The chicken and egg problem is real. The main obstacle would probably be my lack of expertise -- I've been using Flask for ~5 years and even though the book only really covered basic usage, I still felt it pushing at the boundaries of what I was accurately able to teach (someone once said something about only teaching 10% of what you know). In terms of spaCy, I haven't done much more than use its POS, NER, and sentiment analysis functionality -- so I would feel horribly under-qualified to author a book on it.

That said, I'm hoping to publish some more screencast video tutorials soon. I don't have a syllabus yet, but NLP in Python is the most likely candidate at the moment, so we can definitely talk about that. I'm not convinced that books are the best way to learn libraries these days -- smaller project walkthroughs (like the posts you already have on the spaCy homepage) seem way better to me. I'd definitely be keen to author material in that style once I have more experience with spaCy. Feel free to email me at gareth@dwyer.co.za.

PR #561, opened 2016-10-22
Working version of rule matcher tutorial

chssch:
Hi everybody,

I added the merge_phrases function to the tutorial and fixed a minor bug. Let me know if something should be restructured.


chssch:
No Idea why the build is failing here, just changed docs and checked the results with harp.

honnibal:
The failure's not your fault — the Travis log is a little hard to read because it's asking my server to build a distribution, then installing that. Thanks for the fix!

PR #567, opened 2016-10-23
Fix full example links

spalger:
While browsing the docs I found and fixed several broken links. I'm on mobile right now, so I can't verify the fix by rebuilding the docs, but I hope this helps get the fixes started  😀

### DIFF
--- a/website/_harp.json
+++ b/website/_harp.json
@@ -24,6 +24,10 @@
 
         "SPACY_VERSION": "1.0",
         "SPACY_STARS": "2500",
-        "GITHUB": { "user": "explosion", "repo": "spacy" }
+        "GITHUB": {
+            "user": "explosion",
+            "repo": "spacy",
+            "file_url_base": "https://github.com/explosion/spaCy/blob/master"
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


ines:
Thanks for pointing this out! Just fixed it manually – can't merge the PR at this point because now that I see how easily those link mistakes happen, I wanna rewrite some more of the Jade logic to make it easier to handle GitHub links and avoid those messy, concatenated URLs in general.

The GitHub org name should stay in only one global variable (`SOCIAL.github`), so maybe a function could be nice. Just sketching this out here so I don't forget...

```jade
gh("spaCy", "examples/training/train_tagger.py")
//- https://github.com/explosion/spaCy/blob/master/examples/training/train_tagger.py

gh("displacy")
//- https://github.com/explosion/displacy
```

```javascript
function gh(repo, filepath, branch) {
    return 'https://github.com/' + SOCIAL.github + '/' + repo + (filepath ? '/blob/' + (branch || 'master') + '/' + filepath : '' );
}
```

spalger:
👍

PR #569, opened 2016-10-24
updated training examples to v1.1.2

kendricktan:
On spaCy v1.1.2 the `vector` parameter is changed to `add_vector`. I've updated the examples to reflect spaCy v1.1.2.

[x] - `vectors` -> `add_vectors`


honnibal:
Thanks!

PR #570, opened 2016-10-24
Fixed training examples

kendricktan:
#### 1. train_ner won't crash if no data directory is not found
Since v1.1.2, `nlp.tagger` will return `NoneType` if the data directory is not found. This breaks the example file and I've made a change so that `nlp.tagger` will contain the feature_templates from `Tagger` if data directory is not found, should be helpful for people who are just starting out.

#### 2. Fixed train_tagger expected `spacy.gold.GoldParse, got list`
I finally for to play around with `train_tagger.py` but was greeted with the following message:

```python
  File "train_tagger.py", line 74, in <module>
    plac.call(main)
  File "/home/kendricktan/.virtualenvs/laice/lib/python3.5/site-packages/plac_core.py", line 328, in call
    cmd, result = parser.consume(arglist)
  File "/home/kendricktan/.virtualenvs/laice/lib/python3.5/site-packages/plac_core.py", line 207, in consume
    return cmd, self.func(*(args + varargs + extraopts), **kwargs)
  File "train_tagger.py", line 60, in main
    tagger.update(doc, tags)
TypeError: Argument 'gold' has incorrect type (expected spacy.gold.GoldParse, got list)
```

Fixed it by adding a `GoldParse` to `tag`. File now runs.

### DIFF
--- a/examples/training/train_ner.py
+++ b/examples/training/train_ner.py
@@ -6,6 +6,7 @@ import random
 import spacy
 from spacy.pipeline import EntityRecognizer
 from spacy.gold import GoldParse
+from spacy.tagger import Tagger
 
 
 def train_ner(nlp, train_data, entity_types):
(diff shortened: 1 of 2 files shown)


honnibal:
Awesome! Really appreciate the help.

PR #582, opened 2016-10-26
Fix small bug in code of mark-adverbs tutorial

pokey:


honnibal:
Thanks!

Btw, the values of various things will probably be wrong in that tutorial if you're using >1.0, because the new default is to use the GloVe vectors.

pokey:
Ok.  One other thing I noticed is that if you run the tutorial as is, the similar words you get include different capitalizations.  I fixed this by changing line 60 to 

```
words = [w for w in nlp.vocab if w.has_vector and w.is_lower]
```

honnibal:
Nice, thanks.

pokey:
And if it's helpful, here's the result that I get:

```
1-20 bragged, remonstrated, enquired, demurred, sighed, mused, intimated, retorted, entreated, motioned, ranted, confided, countersued, gestured, texted, implored, interceded, muttered, marvelled, bickered
```

PR #620, opened 2016-11-11
Missing import statement for spacy.matcher.Matcher

savkov:
## Description
I've added an import statement at the top of the `Matcher` example.

## Motivation and Context
First-time users and lazy people like me don't know how to import it. The `spacy.matcher` package is not listed in the autocomplete of PyCharm (not sure about other IDEs), so one either has to remember or look it up in the code or other examples.


honnibal:
Thanks!

Issue #364, opened 2016-05-05
German stopwords


bobbyrenwick:
Hey, it would be great if you could add `spacy.de.STOPWORDS` to match `spacy.en.STOPWORDS`.

honnibal:
Good point! Forgot about these.

In the meantime you can set your own as follows:

    import spacy
    
    de_nlp = spacy.load('de')
    for stopword in my_stop_words:
        lexeme = nlp.vocab[stopword]
        lexeme.is_stop = True

The attribute will be inherited by all tokens, because all tokens of a given type receive a reference to that type. So when you write to the lexical type via the vocab, all tokens will see the updated attribute.

bobbyrenwick:
Thanks for the workaround!

ines:
[Fixed!](https://github.com/explosion/spaCy/blob/master/spacy/de/language_data.py)

PR #638, opened 2016-11-19
Add German Stopwords

souravsingh:
Add a list of stopwords for  German Language.

Fixes Issue #364

### DIFF
--- a/spacy/de/language_data.py
+++ b/spacy/de/language_data.py
@@ -4,11 +4,11 @@ import re
 
 
 STOP_WORDS = set("""
-a ab aber ach acht achte achten achter achtes
+ab aber ach acht achte achten achter achtes
 ag alle allein allem allen aller allerdings alles
 allgemeinen als also am an andere anderen
 andern anders au auch auf aus ausser außer
-ausserdem außerdem b bald bei beide beiden
+ausserdem außerdem bald bei beide beiden
 beim beispiel bekannt bereits besonders
 besser besten bin bis bisher bist da
 dabei dadurch dafür dagegen daher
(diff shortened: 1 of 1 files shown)


souravsingh:
@ines Thanks for the review. The ``a`` and ``b`` in the texts are the result of my mistake. I apologize for that

ExplodingCabbage:
@souravsingh, how did you compile this list? Is it taken from a single source, or multiple? To what extent have you personally tweaked that data by adding words you noticed were missing or removing words you thought were inappropriate? This list looks similar, but not identical, to https://github.com/wgpsutherland/stopwords/blob/master/dist/de.json; I couldn't find any other plausible-looking sources on Google.

(Note that I'm not a project maintainer, just a random guy from the internet, but if I were in @ines's place I'd personally want to know what the source of the data was before merging.)

souravsingh:
@ExplodingCabbage The list was taken from multiple sources. The list was compiled from the website here-http://codingwiththomas.blogspot.in/2012/01/german-stop-words.html and from the stopwords list from Apache Lucene. I had some knowledge of German, so identifying which ones are plausible stopwords wasn't really difficult.

ines:
Thanks for the info! Will merge this now and make a few edits and additions.

Btw, @honnibal and I went over the current state of the language data earlier and it could definitely need some better organisation. Starting with basic formatting, but also more complex stuff – for example, having a global module for emoticons that can be imported across languages (instead of having the same data live in each language).

So I'll be taking this on over the next week or so 😃  English and German will be easy (native German speaker here), but we might post a call for native speakers for the other languages soon, just to have another pair of eyes making sure it's all good.

We have such a great community from all over the world, so we should be able to do this! 💪

PR #642, opened 2016-11-20
Let --data-path be specified when running download.py scripts

ExplodingCabbage:
## Description
Adds a `--data-path` (or `-d`) argument to the download.py scripts that lets the path to download to be specified.

## Motivation and Context
Makes downloading to a custom location easier; resolves https://github.com/explosion/spaCy/issues/637

## How Has This Been Tested?
Manually. After `pip install -e .`, I did a `git clean -xdf` to clean my local repo, then, in sequence:

- Checked that 

  ```
  import spacy
  spacy.load('en')('Hello, world')[0].pos_
  ```

  gives the empty string
- Ran `python3 -m spacy.en.download all -d ~/some_dir`
- Checked that `spacy.load('en')('Hello, world')[0].pos_` still gave the empty string but that `spacy.util.set_data_path('/home/mark/some_dir'); spacy.load('en')('Hello, world')[0].pos_` didn't.
- Ran `python3 -m spacy.en.download all`
- Checked that `spacy.load('en')('Hello, world')[0].pos_` no longer gave the empty string.
- Tried running `python3 -m spacy.en.download all` and `python3 -m spacy.en.download all -d ~/some_dir` again and confirmed that both exited early saying that the model was already installed.

## Types of changes

- [ ] Bug fix (non-breaking change fixing an issue)
- [x] New feature (non-breaking change adding functionality to spaCy)
- [ ] Breaking change (fix or feature causing change to spaCy's existing functionality)
- [ ] Documentation (Addition to documentation of spaCy)

## Checklist:

- [x] My code follows spaCy's code style. (*as far as I can tell - there's no actual style guide in CONTRIBUTING.md yet, so I just tried to imitate what was nearby)
- [x] My change requires a change to spaCy's documentation. (*ideally, anyway*)
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes. (*no tests for the downloader currently exist; any that we'd want to add would probably want to be CI-only, since they'll be extremely slow for anyone on a slow internet connection*)
- [x] All new and existing tests passed.

PR #644, opened 2016-11-20
Fix a bunch of missing spaces of the website

ExplodingCabbage:
Having never touched Jade before, I ran into this annoying syntax trap while writing https://github.com/explosion/spaCy/pull/642/commits/a0c4b29dcbacfa784e7522653e18ca6c78084419 and immediately wondered if there was anywhere in the existing codebase where it had already been screwed up. Sure enough, a regex search in Sublime for `\| [^ ]` turned up these 4 locations, all of which are currently rendered with a missing space on the website (although it's not too noticeable in the cases where there's a code block on either side of a comma).

### DIFF
--- a/website/docs/api/doc.jade
+++ b/website/docs/api/doc.jade
@@ -164,8 +164,8 @@ p
         +cell #[code other]
         +cell -
         +cell
-            | The object to compare with. By default, accepts #[code Doc],
-            | #[code Span], #[code Token] and #[code Lexeme] objects.
+            |  The object to compare with. By default, accepts #[code Doc],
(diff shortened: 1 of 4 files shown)


ines:
Thanks!

And yeah, as much as I like Jade, hard-wrapping can be pretty annoying. Sorry about that. It just makes the most sense for the docs, which is why we opted for the pipe + 2 spaces convention.

I've been trying to cover the most important syntax and gotchas [here](https://github.com/explosion/spaCy/tree/master/website#making-changes-to-the-site) – so if there's anything that should be added or documented, let me know!

PR #646, opened 2016-11-20
Ignore entire data folder

ExplodingCabbage:
Previously only some of its content was ignored, so running `python -m spacy.en.download all` after installing from a local repo would create unstaged changes.


honnibal:
Thanks!

PR #676, opened 2016-12-11
Add acl to symbols.pyx

pokey:
Add acl to symbols.pyx

## Description
Add acl to symbols.pyx

## Types of changes

- [ ] Bug fix (non-breaking change fixing an issue)
- [ ] New feature (non-breaking change adding functionality to spaCy)
- [ ] Breaking change (fix or feature causing change to spaCy's existing functionality)
- [ ] Documentation (Addition to documentation of spaCy)

## Checklist:

- [ ] My code follows spaCy's code style.
- [ ] My change requires a change to spaCy's documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #677, opened 2016-12-11
Revert "Add acl to symbols.pyx"

honnibal:
Reverts explosion/spaCy#676

This patch needs to add the `acl` symbol to the actual `enum`, in the `symbols.pxd` file. Unfortunately, there's an unfortunate problem at the moment.

I made the questionable decision of adding all of these symbols to the `StringStore`, so that when you do `nlp.vocab["NOUN"]` you get back the same integer symbol as `spacy.symbols.NOUN`. That's nice, but the problem is I added these at the *start* of the string table. This means that if we add anything to the symbols list, all of the other strings get pushed back! This means all of the models have to be retrained.

Version 2 will fix this by adding these special symbols towards the *end* of the strings table. Until then we have to just log the changes, because I don't want to make master incompatible with the current model.

### DIFF
--- a/spacy/symbols.pyx
+++ b/spacy/symbols.pyx
@@ -371,7 +371,6 @@ IDS = {
     "ORDINAL": ORDINAL,
     "CARDINAL": CARDINAL,
 
-    "acl": acl,
     "acomp": acomp,
     "advcl": advcl,
     "advmod": advmod,


pokey:
@honnibal Ok; I have a workaround anyways

What's the rational for having all these in `strings` / `vocab`?  I'd argue that all symbols should just be in symbols.pyx.  Then we can avoid code like https://github.com/explosion/spaCy/blob/master/spacy/syntax/iterators.pyx#L7-L12 and just import all those symbols from symbols.pyx

honnibal:
The logic is that it's pretty nasty if `vocab.strings[NOUN] != 'NOUN'`. The integer ID we assign to `NOUN` will resolve to some word or other. We should make sure it resolves to `"NOUN"`. Anything else is super confusing.

The answer will be to have these map to higher numbers. This works better across languages, too. Every language can get a vocabulary counting from 0, while the non-language symbols will count down from the maxint.

pokey:
I guess what confuses me is why `'NOUN'` should be in `vocab.strings` at all.  My understanding is that `vocab.strings` contains the vocabulary of the language, whereas `'NOUN'` is not part of the vocabulary of the language, but rather is a tag applied to tokens of the language.  It seems like `vocab.strings` is being overloaded.  Even if we did happen to see `'NOUN'` in a sentence, I don't think it should have the same value as the symbol `NOUN`, because it just leads to conflation of the concepts of `'NOUN'` the word and `NOUN` the pos tag.

pokey:
I reckognize that `NOUN` is an integer, so that `vocab.strings[NOUN]` is technically a well-typed operation.  However, I think this argues for `NOUN` not to be an integer, but rather a richer semantic type that wraps an integer.  Then the user won't accidentally try to index into `vocab.strings` with a pos tag, which is not a semantically meaningful operation imo

honnibal:
Internally `NOUN` has to be an integer, because you need to be able to put it in a numpy array or a tensor. We also need to make sure the internal storage requirements are low. We could change the Python API so that when the user sees it, it's wrapped in an object. Python objects are relatively heavyweight, though, and I think it's better to not have the Python layer diverge too much from the underlying representation.

pokey:
The richer semantic type thing makes a lot of sense in a language like C++, where you can have an object which is completely different from an integer from the perspective of the type checker, but compiles to assembly that is identical to using an unwrapped integer.  Ie:

```
class PosTag {
   private int value;
   public int tensor() {
      // For use in numpy array
      return value;
   }
};
```

I must confess I know very little about Cython, but was hoping something similar might be possible.  If not, that's ok, I was mainly pointing out that it's confusing to explicitly support indexing into `vocab.strings` with a pos tag.  I think it's a good thing to have a clear distinction between elements of the vocabulary and meta symbols like pos tags.  The documentation could support this distinction.  If the user gets confused and tries to index into `vocab.strings` with a pos tag, ideally this would be caught by a type checker, but given python doesn't have a type checker, it's probably better to have this be a run-time error.  

I guess maybe you'd argue it's more convenient to have `vocab.strings` be the universal place where the user goes to get a string when they have a symbol, be it for vocab element or pos tag?  I guess there's something to this, especially if you're planning to support different strings for `NOUN` in the `vocab.strings` of English vs German.  Not sure that's a great idea, though, given the way that these strings are used interchangeably with their corresponding symbols in the code.

honnibal:
Cython supports C stuff very well, and C++ stuff a little bit. You can't easily get C++ zero-cost abstractions in Cython. Because the POS tags are enums, you do get a little bit of semantic typing in the C representation — the compiler won't convert them into integers explicitly. You have to ask.

No matter the representation, though, at some point the type information is likely to get lost from the object. For instance, as soon as we do put the tags into the numpy array, we're going to lose the type. Users are also likely to ask for tags by (string) name, rather than by symbol.

Remember that there are also lots of other pieces of information we're trying to map to and from strings and integers, including:

* Extended POS tags (NN, VBZ, etc)
* Dependency relations (nsubj, dobj, etc)
* Morphological features
* Named entity types

It's much less error prone to have a single, consistent string-to-integer mapping, that avoids hard-coding semantic distinctions between different symbol types.

pokey:
Ok your last argument seems reasonable.  I'm not sure I agree that we should take such a cavalier attitude towards type information getting lost; would think we can recover based on context?  Just something to consider; not sure where it should be for priorities

Also, regardless of whether symbols appear in `vocab.strings`, I think it would be better if https://github.com/explosion/spaCy/blob/master/spacy/syntax/iterators.pyx#L4-L12 didn't utilize that.  Seems better to import from symbols.pyx.  Ie:

```
from spacy.symbols cimport nsubj, dobj, nsubjpass, pcomp, pobj, attr, ROOT, root, conj, NP


def english_noun_chunks(obj):
    '''Detect base noun phrases from a dependency parse.
    Works on both Doc and Span.'''
    np_deps = [nsubj, dobj, nsubjpass, pcomp, pobj, 
              attr, ROOT, root]
    for i, word in enumerate(obj):
```

PR #695, opened 2016-12-20
Added Swedish morph rules

magnusburton:
## Description
I added morph rules based on the ones over at the [Swedish Wiktionary (See the table)](https://sv.wiktionary.org/wiki/deras). Any questions about how I thought when creating these rules are appreciated. Thanks guys!

## Types of changes

## Checklist:

### DIFF
--- a/spacy/sv/morph_rules.py
+++ b/spacy/sv/morph_rules.py
@@ -1,0 +1,68 @@
+# encoding: utf8
+from __future__ import unicode_literals
+
+from ..symbols import *
+from ..language_data import PRON_LEMMA
+
+# Used the table of pronouns at https://sv.wiktionary.org/wiki/deras
+
+MORPH_RULES = {
+    "PRP": {
+        "jag":          {LEMMA: PRON_LEMMA, "PronType": "Prs", "Person": "One", "Number": "Sing", "Case": "Nom"},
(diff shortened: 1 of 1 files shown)


honnibal:
Thanks!

Just so you're aware, I think the morphology rules aren't being wired up correctly in the code at the moment. So, if it doesn't work, it's probably us, not you!

I'm particularly interested to find out whether the morphology system is sufficiently powerful for Swedish. In English, this is pretty much all we need, but for many languages (e.g. Finnish), we can't expect a flat tag scheme to express enough of the morphology.

magnusburton:
@honnibal The Swedish language is quite complex in it's ways to define _mine_, _yours_, _his_, _hers_ etc. First if all we have _min_ and _mitt_. They are reflexive first person singular possessions. The difference between them is comparable to _a_ and _an_ in English (except it's not linked to vowels).

_My table_ in Swedish would be _mi**tt** bord_, whereas _my chair_ would become _mi**n** stol_. I gave them similar options in the rules because I didn't think you had options for these kind of settings.

Thank you for your comment. Will keep committing with updates as I see fit so if you're not ready for a merge yet, I totally understand.

ines:
Yes, I think this is totally fine to merge! 

Also, thanks for the insights into the language. I think a great next step would be adding some tokenizer exceptions. I assume there won't be as many in Swedish as there are [in English](spacy/en/tokenizer_exceptions.py) (mostly contractions), so just having the most common abbreviations should do for now (like in [German](spacy/de/tokenizer_exceptions.py)).

PR #697, opened 2016-12-20
Fixed missing vec_path declaration that was failing if 'add_vectors' …

jmizgajski:
Added vec_path variable declaration to avoid accessing it before assignment in case 'add_vectors' is in overrides.

Added vec_path variable declaration to avoid accessing it before assignment in case 'add_vectors' is in overrides.

## Description
Added vec_path variable declaration to avoid accessing it before assignment in case 'add_vectors' is in overrides.

## Motivation and Context
Without this fix one cannot make loading the vectors optional (7s overhead)

## How Has This Been Tested?
1 line fix, tested manually

## Types of changes

- [x] Bug fix (non-breaking change fixing an issue)
- [ ] New feature (non-breaking change adding functionality to spaCy)

## Checklist:

### DIFF
--- a/spacy/en/__init__.py
+++ b/spacy/en/__init__.py
@@ -44,6 +44,7 @@ def _fix_deprecated_glove_vectors_loading(overrides):
     else:
         path = overrides['path']
         data_path = path.parent
+    vec_path = None
     if 'add_vectors' not in overrides:
         if 'vectors' in overrides:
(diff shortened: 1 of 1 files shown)


jmizgajski:
I don't think this is the best PR I could have made, but I wanted to let you know that the bug exists and where.

honnibal:
Great, thanks.

Don't worry so much about the bureaucracy for this sort of thing :)

PR #702, opened 2016-12-22
fixed minor typo

fnorf:
Peformance -> Performance


ines:
Wow, so this must have been there for a very long time 😮  It's always the most obvious ones that we miss...

Thanks a lot for spotting and fixing this!

PR #703, opened 2016-12-22
Added Swedish abbreviations

magnusburton:
## Description
Added Swedish abbreviations with tokenizer exceptions coming soon.

## Types of changes

## Checklist:


ines:
Thanks, this looks good!

FYI, I made a few small edits:
- removed exceptions containing whitespace (`"i sht"`, `"i st"`)
- removed exceptions not containing special characters (like `"kcal"`) – in case some of the Swedish-specific ones also exist with a `.`, you might want to re-add those

Since the tokenizer deals with whitespace-delimited chunks, an exception containing whitespace, like `"i sht"`, will never match. For the same reason, tokenizer exceptions are only relevant for abbreviations containing special characters that would otherwise be split off. `"kcal"` is already a whitespace-delimited chunk, so the exception is not necessary. 

Unlike, for example, `"förf."` – by default, it'd become `"förf"`, `"."`. Instead, it'll now be considered one token `"förf."`, which is correct. 

For unambiguous abbreviations that always translate to the same normalised form, you could even consider adding them to `TOKENIZER_EXCEPTIONS` instead. For example:

```python
"Jan.": [
    {ORTH: "Jan.", LEMMA: "January"}
],
```

In case Swedish has abbreviations that "translate" to an inflected form, you can specify that as `NORM` and use the base form for `LEMMA`:

```python
# I made this up because English doesn't really have those 
# (at least I can't think of a good example)

"pzzs.": [
    {ORTH: "pzzs.", LEMMA: "pizza", NORM: "pizzas"}
],
```

PR #705, opened 2016-12-23
Initial support for Hungarian

oroszgy:
## Description
Added initial support for Hungarian containing a tokenizer and stop-words.

## Motivation and Context
Hungarian is one of the morphologically most complex languages, why should not we support it in spaCy? :)

## How Has This Been Tested?
Several test cases were implemented to test the behaviour of the Hungarian tokenizer. `tests/hu/tokenizer/test_tokenizer.py`

## Types of changes

- [X] New feature (non-breaking change adding functionality to spaCy)

## Checklist:


ines:
Wow, thanks – this is amazing! 💯  And thanks for adding the tests!

I was just looking at the prefixes/suffixes/infixes in [puncuations.py](https://github.com/oroszgy/spaCy.hu/blob/45e045a87b0170395efc6fc25ed34c5603f0ec8f/spacy/hu/punctuations.py) – how about we merge these entirely with the base punctuation in [language_data/punctuation.py](https://github.com/explosion/spaCy/blob/master/spacy/language_data/punctuation.py)? I don't think there's any ambiguity here and it'd be nice to have the additional characters (`üóőúéáűíÜÓŐÚÉÁŰÍ` etc.) available in other languages as well, just in case. This will also make it easier to maintain.

honnibal:
This is super good!

I think Hungarian will be an interesting test case once we have the morphology logic hooked up properly, and once we have neural network models (including character-based networks). But the first step is always tokenisation :)

If you want a fun project at some later date: I've always wanted to try seq2seq NMT models for lemmatization. I think this should work quite well, and it will remove one of the most difficult requirements for language-specific resources.

ines:
Btw, when you have a second, would you mind filling in the [Contributor Agreement](https://github.com/explosion/spaCy/blob/master/.github/CONTRIBUTOR_AGREEMENT.md)? Just to make sure we can use and integrate this across the project etc. 😃

oroszgy:
@ines Done. 

@honnibal The idea sounds good, but did you consider using morfessor in spaCy?

honnibal:
I haven't really looked into Morfessor much — I should probably give it more thought.

@oroszgy : What's the deal with the -ben, -e etc affixes in Hungarian? These are case markers, right?

Would it make sense to split these into their own tokens? Especially for `-ben`, which is attaching to numeric entities, splitting these looks appealing to me. I notice that's not how the Hungarian treebank is doing this, so I was wondering what your take would be.

ines:
Yes, I'm also curious about this!

To explain the recent changes to the tokenizer: I've [reorganised the global prefixes, suffixes and infixes](https://github.com/explosion/spaCy/commit/347c4a2d06cd08c0edd65c87cd558dae4f5a63c7) to generalise some of the basic shared rules, and grouped up alpha unicode characters to be used in the character sets. The idea behind this is that all languages should know the most common characters, even if they're not present in the language itself (for example, `ő` might not be used in English, but if we do come across it in a token, we'd still prefer to be able to match the punctuation rules). In the long term, we should probably consider using unicode ranges instead of writing out each character, but for now, it might be best to keep it explicit and more readable.

It all works fine for Hungarian – except a few edge-cases where the expected tokenization specified in the tests is quite different from the standard behaviour of the tokenizer. (It's mostly combinations numbers/hyphens and `-ben` affixes, which is why I've temporarily removed testing for numbers and hyphens.) 

If this is behaviour is specific to Hungarian, we should obviously add more rules again or overwrite some of the defaults.

oroszgy:
@honnibal "ben" is a case marker, however "-e" is not, it is the question tag particle. (E.g. "ül" means *sit*, "ül-e" means *does he hit*) In general, case markers and other suffixes are simply appended to words, but "-e" is always given with the hyphen. This means that "-e" should be treated as a single token, however the others are part of the wordform. Looking at the testcases you can find a lot of "<NUMBER>-ben" examples that are actually correct: inflexions are appended to numbers or quoted words/phrases with hyphens. 

Actually, I am in the progress of [building vocabulary models for Hungarian](https://github.com/oroszgy/spaCy.hu/tree/hu_vocab) and found a few bugs especially around tokenizing numbers. Hopefully you can expect a bugfix in a few days. In general, I would like to follow the principles the Szeged Treebank (as much as possible), since this is the way how I can incorporate a tagger and a parser as well.

@ines I really like the reorganization part. It makes sense to allow other tokenizers to handle foreign words well. :) As I said, I am in the progress of improving the tokenization of numeric expressions, so hopefully can address those problems as well.

honnibal:
Sounds good! I think we'll need to make another adjustment to the tokenizer rules for Hungarian then. If `-ben` definitely needs to attach to the numbers, I'm not sure that's still being handled.

Btw, spaCy's `GoldParse` class can perform alignment of tokens to a raw string during training, so you can actually train on the "real" tokens. This makes the system a little less sensitive to divergences between the tokeniser and the treebank scheme. Obviously it's best to stay close to the treebank, though, especially on common words. By default misalignments are handled as missing values, although you could come up with heuristics to enact a different policy.

oroszgy:
Thanks! I'll look into the `GoldParse` class when I get there.

PR #726, opened 2017-01-09
Added Swedish abbreviations as token exceptions

magnusburton:
## Description
I added abbreviations for januari (January) and the abbreviations for the two biggest cities in Sweden (Stockholm & Göteborg)

## Types of changes

## Checklist:

### DIFF
--- a/spacy/sv/tokenizer_exceptions.py
+++ b/spacy/sv/tokenizer_exceptions.py
@@ -62,6 +62,12 @@ TOKENIZER_EXCEPTIONS = {
     ],
     "sön.": [
         {ORTH: "sön.", LEMMA: "söndag"}
+    ],
+    "sthlm": [
+        {ORTH: "sthlm", LEMMA: "Stockholm"}
+    ],
+    "gbg": [
+        {ORTH: "gbg", LEMMA: "Göteborg"}
     ]
 }
 
--- a/spacy/sv/tokenizer_exceptions.py
+++ b/spacy/sv/tokenizer_exceptions.py
@@ -6,7 +6,63 @@ from ..language_data import PRON_LEMMA
 
 
 TOKENIZER_EXCEPTIONS = {
-    
+    "jan.": [
+        {ORTH: "jan.", LEMMA: "januari"}
+    ],
+    "febr.": [
+        {ORTH: "febr.", LEMMA: "februari"}
+    ],
+    "feb.": [
+        {ORTH: "feb.", LEMMA: "februari"}
+    ],
+    "apr.": [
+        {ORTH: "apr.", LEMMA: "april"}
+    ],
+    "jun.": [
+        {ORTH: "jun.", LEMMA: "juni"}
+    ],
+    "jul.": [
+        {ORTH: "jul.", LEMMA: "juli"}
+    ],
+    "aug.": [
+        {ORTH: "aug.", LEMMA: "augusti"}
+    ],
+    "sept.": [
+        {ORTH: "sept.", LEMMA: "september"}
+    ],
+    "sep.": [
+        {ORTH: "sep.", LEMMA: "september"}
+    ],
+    "okt.": [
+        {ORTH: "okt.", LEMMA: "oktober"}
+    ],
+    "nov.": [
+        {ORTH: "nov.", LEMMA: "november"}
+    ],
+    "dec.": [
+        {ORTH: "dec.", LEMMA: "december"}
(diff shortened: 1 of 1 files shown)


ines:
Looks good, thanks! 👍

What are the capitalisation rules like in Swedish btw? It's mostly lowercase, right? Just in case any of these abbreviations also commonly occur capitalised, it might make sense to add two versions.

I also came across [this list of Swedish contractions](https://en.wiktionary.org/wiki/Category:Swedish_contractions) – are any of these relevant? A lot of it seems pretty random, but if any of these are actually common, it could be nice to have them included as well.

magnusburton:
@ines Unlike English we don't capitalize months and days, therefore I should probably add versions of the abbreviations capitalized as well, correct? (Take the sentence _Monday comes before Tuesday_ -> _Måndag kommer före tisdag_).

Also, many thanks for that list of Swedish contractions, totally forgot about these words although _tjarrå_ (meaning bye) is one of my personal favorites. I'll work on adding those in!

ines:
Yeah, as a rule of thumb, it's always better to have them, just in case.

Also, I'm not sure if this is relevant in Swedish, but: If there are contractions/abbreviations that translate to more than one token but are only ever used as this exact phrase, it's fine to set them up as one token with a multi-word lemma, instead of splitting them up.

I can't think of an example in English, but German has a lot of these: `z.Zt.` for example stands for `zur Zeit` ("at the moment"). Instead of splitting this up into an inflected pronoun (`zur`) and a noun (`Zeit`) straight away, we currently keep it as one token in the tokenizer exceptions and provide the written-out form as the lemma.

magnusburton:
@ines Also, I was wondering; I realized that the list of Swedish contractions you sent can be made much much bigger. After thinking I found over 20 different "new" contractions all based on verbs;

`hajaru` -> `hajar du`, which means `do you get it`
`hörru` -> `hör du`, which means `do you hear me`
`fixaru [det]` -> `fixar du [det]`, which means `are you fixing it`
`tränaru` -> `tränar du`, which means `are you (training/practicing/working out)`

I figured the list can be made very long if I just sit down and think for a moment. Is there an easy way to turn these contractions into exceptions without having to go through every single one of them?

ines:
Does this apply to a set of specific verbs only, or is this more of a general pattern that applies to most/a lot of verbs? If it's more of a general pattern (like the [reflexive verbs](https://en.wikipedia.org/wiki/Spanish_pronouns#Reflexive_pronouns_and_intensifiers) in Spanish that take suffixes like `-se`, `-lo` etc.), it's a bit trickier and should be handled later in the lemmatizer, because we don't want to list every possible option manually.

If the set of verbs is managable, they should definitely be handled in the tokenizer exceptions. Since it's all done in Python, it's fine to get a little creative here and use loops to avoid doubling markup, wherever it seems reasonable. I did [something like this](https://github.com/explosion/spaCy/blob/master/spacy/en/tokenizer_exceptions.py#L206) for English contractions of verbs that follow a general pattern (like `can't`, and `won't`). You can also loop over the lowercase and title case versions to save you some work and keep the code neat.

For the examples you mentioned, here's how this could look (not sure if this is 100% correct for Swedish, but you get the idea):

```python
for verb_data in [
    {ORTH: "harjar", LEMMA: "..."},
    {ORTH: "hörr", LEMMA: "..."}
]:
    verb_data_tc = dict(verb_data)
    verb_data_tc[ORTH] = verb_data_tc[ORTH].title()

    for data in [verb_data, verb_data_tc]:
        EXC[data[ORTH] + "u"] = [
            dict(data),
            {ORTH: "u", LEMMA: PRON_LEMMA, NORM: "du"}
        ]
```

To check that the exceptions are all created correctly, just sort and print them:

```python
print "\n".join(sorted(TOKENIZER_EXCEPTIONS))
```

Then run the basic quick tokenizer tests with `-s`:

```bash
py.test -s spacy/tests/tokenizer
```

magnusburton:
@ines Great informative reply, much appreciated, thanks! It could be applied for almost every verb in Swedish but it is rarely used except for a select number of verbs.

I mean, it's much more common to say (or write) `driveru` meaning `are you kidding` than for example the contraction `vattnaru blommorna?` meaning `are you watering the plants` (where `vattnar` and `watering `is the same verb).

Single it's a manageable amount och contracted verbs I might just go ahead and use the code you supplied and add in a loop for capitalization. Thanks a lot!

PR #732, opened 2017-01-11
Avoid "True if ... else False"

danielhers:
Minor change, maintains code semantics.


honnibal:
I wonder why I wrote it like that. Weird.

Thanks! More like this please, if you have time and inclination :)

PR #738, opened 2017-01-12
Fix #737: support loading word vectors with " " as a word

### LINKED ISSUE #737: load_vectors should allow whitespace tokens for Gensim compatibility
I am calling `load_vectors` on the text file produced from the following:
* Download [GoogleNews-vectors-negative300.bin.gz](https://drive.google.com/file/d/0B7XkCwpI5KDYNlNUTTlSS21pQmM/edit?usp=sharing)
* Use Gensim to convert to text (need about 8G RAM and disk space): http://stackoverflow.com/a/33183634/223267

I am getting:
```
spacy.vocab.VectorReadError: Error reading word vectors from <_io.TextIOWrapper name='word_vectors/GoogleNews-vectors-negative300.txt' mode='r' encoding='UTF-8'> on line 431879.
All vectors must be the same size.
Prev size: 300
Curr size: 299
```

I looked at line 431880 (the error message above is zero-based), and indeed there are just 300 elements in it (instead of 301 for the token + numbers) because the "token" there is just a space...
You could say the file is faulty for that reason, but since it's widely used I think spaCy should be able to handle it.

* Python Version Used: 3.5
* spaCy Version Used: 1.5.0

danielhers:


honnibal:
Thanks!

PR #742, opened 2017-01-14
Improved Hungarian tokenizer

oroszgy:
## Description

Fixes bugs that were introduced during refactoring tests, improves numeric expression handling and fixes parsing for hyphenated tokens. The tokenizer now mostly follows Hungarian grammar rules.

## Motivation and Context

* During the refactoring of the test module, the Hungarian tokenizer stopped passing some of the test cases. 
* There were a few minor bugs in`spacy/language_data/punctuation.py` which are also fixed. 
* Did an extensive test on a wiki dump, made the tokenizer more robust.
* Improved compatibility with the Hungarian Szeged Treebank.
* New abbreviations are added.

## How Has This Been Tested?

* New tests added, old ones passing.
* Manual comparison with the sentences of the Szeged Corpus.

## Types of changes

- [X] Bug fix (non-breaking change fixing an issue)
- [ ] New feature (non-breaking change adding functionality to spaCy)

## Checklist:


ines:
This looks great – pretty excited about the progress on Hungarian btw. And thanks for fixing the punctuation as well!

Now that I look at the abbreviations again, maybe it'd be a good idea to just add all unicode alpha characters + `.` to the exceptions by default.  Basically, something like this:

```python

SINGLE_LETTER_ABBR = [strings_to_exc(a + '.') for a in ALPHA.split()]
```

I initially thought this might be overkill, but it does make sense, since we're now including all of these characters in the regexes as well. And if we're still asking every language to include their own special characters as single-letter abbreviations, this sort of defeats the point of that change (and introduces more potential errors and missing abbreviations as we're adding more languages).

oroszgy:
Thanks for merging/
@ines might be a good idea, but I think that it should happen in a way that one should be able to overwrite this as there might valid one letter words: http://jalu.ch/languages/one_letter_words.php

ines:
@oroszgy Good point – maybe we should then just let every language specify their one-letter words (not too many after all, according to the list) and exclude them + `.` from the exceptions? Still hacky, but it seems more reasonable than asking each language to specify all one-letter abbreviations that are *not* words.

oroszgy:
@ines I'd not try to collect all one letter word abbreviation in one language independent list, since it might be hard to predict in advance what is an abbreviation and what is not. For examples `"a"` is a one letter word (a determiner) in Hungarian but it never goes alone, meaning that it must be followed by a noun. In practice, it implies that if there is a `"a."` string there is a high chance that it is an abbreviation.

ines:
@oroszgy Fair enough – maybe it's best not to generalise too much then. The more I think about this, it also seems a little risky to assume too much data by default and ask languages to override and exclude...

PR #806, opened 2017-02-04
Fix issue #805

wallinm1:
## Description
Aims to fix issue #805 by importing the Swedish-specific `ORTH_ONLY` and `TOKENIZER_EXCEPTIONS`-exceptions in the `language_data.py`-file. This requires for the `s.`-exception to be removed from the Swedish-specific exceptions to not conflict with global abbreviations.

## Types of changes

- [x] **Bug fix** (non-breaking change fixing an issue)
- [ ] **New feature** (non-breaking change adding functionality to spaCy)
- [ ] **Breaking change** (fix or feature causing change to spaCy's existing functionality)
- [ ] **Documentation** (addition to documentation of spaCy)

## Checklist:

- [ ] My change requires a change to spaCy's documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #807, opened 2017-02-04
Added swedish lemma rules and more verb contractions

magnusburton:
## Description
I added some swedish lemma rules and some more verb contractions including a loop for `fattaru` etc.

I have a question about how I've go ahead with implementing the last two adjective categories [here](http://www.worddive.com/grammar/sv/svensk-grammatik/5-komparation-av-adjektiv/). Since the base word is for example `god` (nice or kind) and it's an irregular adjective. What's the best approach to words like this?

## Types of changes

## Checklist:

### DIFF
--- a/spacy/sv/lemma_rules.py
+++ b/spacy/sv/lemma_rules.py
@@ -1,0 +1,45 @@
+# encoding: utf8
+from __future__ import unicode_literals
+
+
+LEMMA_RULES = {
+    "noun": [
+        ["t", ""],
+        ["n", ""],
+        ["na", ""],
+        ["na", "e"],
+        ["or", "a"],
+        ["orna", "a"],
+        ["et", ""],
+        ["en", ""],
+        ["en", "e"],
+        ["er", ""],
+        ["erna", ""],
+        ["ar", "e"],
+        ["ar", ""],
+        ["lar", "el"],
+        ["arna", "e"],
+        ["arna", ""],
+        ["larna", "el"]
+    ],
+
+    "adj": [
+        ["are", ""],
+        ["ast", ""],
+        ["re", ""],
+        ["st", ""],
+        ["ägre", "åg"],
+        ["ägst", "åg"],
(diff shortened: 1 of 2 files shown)


honnibal:
The `Morphology` class is a bit broken at the moment, but in theory it allows you to define a morphological analysis for the token based on a part-of-speech key. So you can say "best + JJ = "good + comparative", and "best + VBP = best". I think there are example morphology rules in the language data. If you define the data in that format, it should work once the morphology class is fixed.

ines:
Looks good – thanks for your great work on the Swedish language data!

Btw, in case you haven't seen it: @wallinm1 (who contributed the Finnish tokenizer) made some small fixes to the existing Swedish data in #806, including a tokenizer test.

I'll create a directory `sv` in [`tests`](https://github.com/explosion/spaCy/tree/master/spacy/tests) and copy the existing test over, and maybe add one for the abbreviations. So you'll be able to add more test cases there if you like.

ines:
Okay, just added the Swedish tests directory: [`tests/sv`](https://github.com/explosion/spaCy/tree/master/spacy/tests/sv).

Also fixed a small bug in the tokenizer exceptions (see 6715615d553f57ecbf9db674611f4d56bebf1a5e). `EXC = {}` wasn't declared at the top, so it wasn't able to add the verb exceptions and spaCy failed running. So I added that, plus a test for the verb contractions, and it passes.

You should be able to run the tests via `fab test` from the spaCy directory, or alternatively, `py.test spacy` from within your virtualenv.  (You can also do `py.test spacy/tests/sv` to just run the Swedish tests.)

magnusburton:
@honnibal Great tips. I'll start working on some rules based on the format you provided. Couldn't however find the example morphology rules you mentioned.

@ines Thank you for the merge and for the `fattaru` test. Will start working on some tests in the near future to make sure all my previous contributions are totally covered.

PR #814, opened 2017-02-08
Make the website nav header's hysteresis a bit more robust

wehlutyk:
## Description

When browsing the website in Firefox, the header navbar (which disappears when scrolling down) keeps reappearing every time you stop scrolling down. The `scroll` event seems to fire a bit more than on Chrome, leading `scrollUp` to go back to `true` when scrolling ends, when in fact we're not moving any more.

This change just keeps the latest `scrollUp` value if we haven't actually moved.

No tests needed, no documentation change, so the checklist is empty.

## Types of changes

## Checklist:

### DIFF
--- a/website/assets/js/main.js
+++ b/website/assets/js/main.js
@@ -14,7 +14,7 @@
     const updateNav = () => {
         const vh = updateVh()
         const newScrollY = (window.pageYOffset || document.scrollTop) - (document.clientTop || 0)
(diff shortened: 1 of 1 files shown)


ines:
Oh, this is great – thanks a lot for getting to the bottom of this and fixing it! 👍

wehlutyk:
Pleasure! Thanks for merging :)

Issue #783, opened 2017-01-28
Support for French tokenization exceptions


raphael0202:
I recently submitted a PR so that spaCy supports tokenization in French. I'm currently working to add the numerous tokenization exceptions existing in French.
With the help of wiktionary, I ended up with ~100k tokenization exceptions. 
You can find [here](https://github.com/explosion/spaCy/files/737066/tokenization_exceptions.gz) the gzipped exception file.

I first tried to directly load the gzipped file in `fr/language_data.py`, but it had a major impact on spaCy loading time: it took 2.0s just to `import spacy` (1.8s spent in `fr/language_data.py`).
By saving the processed `TOKENIZER_EXCEPTIONS` in a file with `numpy.save`, and loading it with `numpy.load`, it only took 0.8s to load `fr/language_data.py`. It's still too much, as this file is imported everytime spacy is loaded.

I wanted to have some insights from spaCy developers, in order to find a good solution concerning this tokenizer exceptions problem. One possibility would be to do lazy loading, by transforming `French.tokenizer_exceptions` into a lazy object which only imports the tokenizer exceptions if some of its methods (such as `.items()`) are called.

honnibal:
Hmm!

Maybe we should have a frequency-pruned version of this in the `TOKENIZER_EXCEPTIONS`, and load the rest when you call `nlp = spacy.load('fr')`?

I'm sure the top 5k or top 10k entries here will give very good coverage. I would run this over the common crawl data here: http://statmt.org/ngrams/pages/raw-data.html

raphael0202:
Do we actually need `TOKENIZER_EXCEPTIONS` to be loaded with spaCy or can we simply load it during `spacy.load('fr')`?

honnibal:
We can load it during `spacy.load('fr')`. I'd like to ship decent tokenizer data with the library, though — I want to make sure people don't need to download anything just to use the library.

oroszgy:
Guys, looked into the data, and it looks like that a decent amount of them could covered with good old regexes. (Maybe here: https://github.com/explosion/spaCy/pull/700) What do you think?

honnibal:
If regular expressions work, that's a fine solution. I think to a great extent these things should be to the taste of the developer.

[...]

PR #818, opened 2017-02-09
Add tokenizer exceptions for French

raphael0202:
Add tokenizer exceptions in French

## Description
Add tokenizer exceptions in French, using a list of tokens (saved in a file) and a regex-based matcher (using the BaseDefault.token_match attribute). See issue #783 for the rational behind the approach chosen.

Some tokenizer exceptions could not be spotted however - essentially specific tokenization in hyphen-delimited expression such as `a-t-elle` that should be tokenized as `['a', '-t', '-elle']`- so just the most common of these exceptions were manually added.

## Types of changes

## Checklist:


ines:
Thanks, this looks great – and seems like this was really a good use case for the new `token_match`.

Before I merge, could you have a quick look at that conflict? I'm not entirely sure what it's complaining about, it might just be related to some of the reordering / reformatting.

raphael0202:
The conflict was due to recent additions to the list of tokenizer exceptions, it's now resolved.

ines:
Ok perfect, merging now! (Ignore the Travis fails btw, this is still an issue with our build server.)

raphael0202:
Great, thanks!

ines:
@raphael0202 Sorry, I had to temporarily revert this commit because it kept hitting the `sorry, but this version only supports 100 named groups` Python error. But making all groups non-capturing (e.g. `di(?:sney)?`) fixed the problem 👍

The new changes are now on master (see f08e180a47e71d2af60df1f53de1c7877a974b51). All tests pass, so fingers crossed for the build.

raphael0202:
@ines Oh ok, it's weird because all the tests passed. How did it occur? Thanks for making the fix

ines:
Okay, so after digging around for a bit, I think I found the explanation: Are you using Python 3 by any chance? Apparently, the 100 groups limit in Python 2 has been removed (or at least, increased) in Python 3. So that might be why I hit this error and you didn't.

Here's what I got in Python 2 when running the tests (or doing anything with spaCy, since the groups limit prevented the tokenizer exceptions from being compiled):

```bash
  File "spacy/fr/tokenizer_exceptions.py", line 215, in <module>
    TOKEN_MATCH = re.compile('|'.join('({})'.format(m) for m in REGULAR_EXP), re.IGNORECASE).match
  File "spacy/.~env/lib/python2.7/re.py", line 194, in compile
    return _compile(pattern, flags)
  File "spacy/.~env/lib/python2.7/re.py", line 249, in _compile
    p = sre_compile.compile(pattern, flags)
  File "spacy/.~env/lib/python2.7/sre_compile.py", line 583, in compile
    "sorry, but this version only supports 100 named groups"
AssertionError: sorry, but this version only supports 100 named groups
``` 

It's good that this came up, though. I'll try to remember to add a note about the group limit in Python 2 when we add the `token_match` docs.

The [Travis builds](https://travis-ci.org/explosion/spaCy/builds/200020992) for Python 3 still failed, but for a different reason: `test_issue768.py` is importing the `TOKENIZER_EXCEPTIONS` from `fr/language_data.py` – but they don't actually exist there anymore. So I fixed that as well in the above commit.

It's maybe not ideal to add this one extra step (`tokenizer_exceptions` -> `language_data` -> `__init__`),  but for now, it keeps it consistent with the other languages. The original idea here was that all language-specific data is bundled in`[language]/language_data.py`, so it can always be imported predictably from there.

raphael0202:
Sorry for my late answer.
Yes I'm using Python3, and I haven't launched the tests on Python2. And I only launched the default 'fast' tests.

PR #865, opened 2017-02-28
Add basic Bengali language support

aniruddha-adhikary:
## Description
This is adding initial support for the Bengali language.

## Types of changes

## Checklist:


honnibal:
Thanks!!

Are the basic tokenizer rules for punctuation (, . ; ? etc) relevant for Bengali, and if so, do they work correctly?

I think we set up the regex on these to only be sensitive to Latin alphabetic characters. This is already a problem for Cyrrilic support, so I was meaning to look into the `regex` library, which seems to support unicode better.

aniruddha-adhikary:
@honnibal Yes they do! In Bengali, "।" indicates the end of a sentence; often modern speakers also use a period. The rest are same as Latin.

Issue #781, opened 2017-01-28
💫 Lemmatizer should apply rules on OOV words


honnibal:
@juanmirocks points out in #327 that the lemmatizer fails on OOV words:

```python

>>> nlp.vocab.morphology.lemmatizer(u'endosomes', 'noun', morphology={'number': 'plur'})set([u'endosomes'])
>>> nlp.vocab.morphology.lemmatizer(u'chromosomes', 'noun', morphology={'number': 'plur'})
set([u'chromosome'])
```

Suggested patch to lemmatizer.py

```python

    oov_forms = []
    for old, new in rules:
        if string.endswith(old):
            form = string[:len(string) - len(old)] + new
            if form in index or not form.isalpha():
                forms.append(form)
            else:
                oov_forms.append(form)
    if not forms:
        forms.extend(oov_forms)
```

## Your Environment

* Operating System: 
* Python Version Used: 
* spaCy Version Used: 
* Environment Information:

juanmirocks:
@honnibal thanks for looking into this -- Know I understand the problem better.

Looking forward to a fix as I think it will greatly improve the performance of my methods.

Kindly let me know if I can help with the patch.

ines:
@juanmirocks If you have time to add the patch to `lemmatizer.py` and make a pull request, that would be great 👍  

The only thing that's important is to make sure it's properly tested. In this case, I think it'd be fine to just add a [regression test](https://github.com/explosion/spaCy/blob/master/CONTRIBUTING.md#fixing-bugs) for this issue, using the above examples.

juanmirocks:
Thanks Ines.

I will gladly help and try. Is that patch code all what is necessary as per
decided for Hannibal?

I will dive in

juanmirocks:
Other examples that I will use for debugging and testing:

* "colocalizes" is lemmatized the same, "colocalizes"
  * > PP2A colocalizes with shugoshin at centromeres and is required for centromeric protection .

honnibal:
Merged!

Issue #435, opened 2016-06-23
Lemma for 'dive' as noun


raghavgupta93:
Shows up as 'diva' in my experiments - can you please check? Thanks!

honnibal:
I had assumed there was a simple data error here, but actually this points to something deeper.

The WordNet data gives the mapping of the *plural* forms of nouns to their base form. The lemmatizer was applying this mapping to all noun forms --- both plural and non-plural. This means that "dive" as a singular noun was matching a weird plural form of "diva", "dive". I doubt any instance of the token "dive" will *ever* mean "divas" --- it's much more likely to be a POS tagging error. But the problem in the lemmatization logic may have had other effects.

Thanks for the report!

PR #866, opened 2017-03-01
#781 Apply suggested patch -- passing regression test added

juanmirocks:
## Description

I applied the suggested patch of issue #781 

It's unclear whether the line `forms.extend(exceptions.get(string, []))` is anymore needed (also related to #435). I suggest to remove it. The regression test passes with and without that line. It seems redundant, from my current understanding of the code.

## Types of changes

## Checklist:

- [ ] All new and existing tests passed. (**note: general spaCy tests failed previous change, at least in my installation, so I cannot test this. New test passes**)

### DIFF
--- a/spacy/lemmatizer.py
+++ b/spacy/lemmatizer.py
@@ -86,13 +86,16 @@ def lemmatize(string, index, exceptions, rules):
     #if string in index:
     #    forms.append(string)
     forms.extend(exceptions.get(string, []))
+    oov_forms = []
     for old, new in rules:
         if string.endswith(old):
             form = string[:len(string) - len(old)] + new
             if form in index or not form.isalpha():
                 forms.append(form)
+            else:
+                oov_forms.append(form)
     if not forms:
-        forms.append(string)
+        forms.extend(oov_forms)
     return set(forms)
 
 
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks for this, looks good!

About that line: we do need to look up the exceptions, for arbitrary cases like `people -> person`. That exceptions logic isn't triggered anywhere else, right?

juanmirocks:
Thanks.

You're right, the exceptions logic is not triggered anywhere else (the lookup `exceptions.get` is literally only called in this line out of the entire code base). I was first wrongly assuming it was handling oov exceptions too. 

Therefore, we must keep that line.

spacy/lemmatizer.py
```diff
@@ -86,13 +86,16 @@ def lemmatize(string, index, exceptions, rules):
     #if string in index:
     #    forms.append(string)
     forms.extend(exceptions.get(string, []))
+    oov_forms = []
```
honnibal: Maybe this needs to be:

    oov_forms = [string]

The tests are currently failing when we don't match any rules.

honnibal:
Added a comment on the first commit. I think I see why the tests fail atm.

PR #869, opened 2017-03-03
Add support for Universal Dependencies v2.0

rominf:
## Description
I'm working on Russian language support. This commit is a prerequisite (Russian UD corpus uses UD v2.0).

## Types of changes
I looked at UD v2.0 documentation and added some symbols. To keep old code working, I didn't remove old symbols, even if it was required by new standard.

## Checklist:

### DIFF
--- a/spacy/attrs.pyx
+++ b/spacy/attrs.pyx
@@ -93,7 +93,7 @@ NAMES = [key for key, value in sorted(IDS.items(), key=lambda item: item[1])]
 
 def intify_attrs(stringy_attrs, strings_map=None, _do_deprecated=False):
     '''Normalize a dictionary of attributes, converting them to ints.
-    
+
     Arguments:
         stringy_attrs (dict):
             Dictionary keyed by attribute string names. Values can be ints or strings.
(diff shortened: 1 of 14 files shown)


honnibal:
Thanks!

There's currently a bad design mistake with these attributes: The `vocab` object adds these symbol names at the start of the `StringStore`. This means that adding these attributes changes the offset of the string-to-integer table, breaking compatibility with the pre-trained models.

I therefore have to hold off on merging this until I get the models retrained. The plan is to change the vocab code so that there's a fixed-size region of indices reserved for these special symbols --- say, 1000 or so. This way, we won't have to offset the vocabulary words when we map these symbols, and the same problem won't reoccur.

rominf:
@honnibal 
OK, understand. Should I use this commit in my `russian`-support branch (I think, yes)?

honnibal:
Yes.

honnibal:
I've got this merged on the develop branch, and have the models training. I've been doing some experiments with L1 regularization, which should finally produce small models for testing. I hope to have something uploaded on this today.

PR #872, opened 2017-03-04
[Bengali] basic tag map, morph, lemma rules and exceptions

aniruddha-adhikary:
## Description
Added a tag map with basic pos tags. A few morph rules for pronouns were added. A few common tokenization exceptions were covered. The lemmatizer got a few basic rules.

## Types of changes

## Checklist:


ines:
Thanks so much, this is really coming together well! 👍
I'm very curious how the Bengali tokenizer currently performs out of the box – are there any benchmarks available that we could run?

Also, do you think you have time to add one or two tests for Bengali?  If you add the tokenizer to `LANGUAGES` in [conftest.py](https://github.com/explosion/spaCy/blob/master/spacy/tests/conftest.py), it'll automatically be included in the basic sanity tokenizer tests. For the language-specific tests, something simple [like this](https://github.com/explosion/spaCy/tree/master/spacy/tests/fr) would be totally fine (we mostly want to prevent regressions and make sure non-Bengali-speakers have a simple check available to make sure it all works.)

aniruddha-adhikary:
@ines I am afraid not much work is available online for NLP on Bengali. However, we can start building these now.

I just added some test in #873 !

aniruddha-adhikary:
Bengali Tokenization is pretty easy. It is completely like Latin, except for an extra character '।' for marking the end of a sentence.

PR #876, opened 2017-03-08
[Bangla] Update "tokenizer_exceptions.py"

shuvanon:
## Description

I'm working on Bangla language support. This commit update ABBREVIATIONS in tokenizer_exceptions.py

## Types of changes

## Checklist:


ines:
Thanks a lot – and great to see more people working on Bengali! 👍

I have a question about two of the abbreviations: 
(cc: @aniruddha-adhikary)

> "কি.মি": [
>     {ORTH: "কি.মি", LEMMA: "কিলোমিটার"},
>     {ORTH: "কি.মি.", LEMMA: "কিলোমিটার"},
> ],
> "সে.মি": [
>     {ORTH: "সে.মি", LEMMA: "সেন্টিমিটার"},
>     {ORTH: "সে.মি.", LEMMA: "সেন্টিমিটার"},
> ]

If a tokenizer exception is split up into more than one token, the concatenated `ORTH` values have to match the original string. This doesn't seem to be the case here. (For example, কি.মি would become কি.মিকি.মি. using `Doc.text`.)

Is this a mistake, and is this actually supposed to express that both সে.মি and সে.মি. exist as an abbreviation that should be assigned the same lemma? If so, they should be listed separately and with only one token each. (If there are many abbreviations like this and you want to avoid duplication, you could also use a loop, [like this](https://github.com/explosion/spaCy/blob/master/spacy/en/tokenizer_exceptions.py#L15)).

If কি.মি should be split into two tokens, you'd have to split it into two `ORTH`s, for example কি and .মি or কি. and মি. How this is done can be a little arbitrary sometimes, like "gonna" in English:

```python
"gonna": [
    {ORTH: "gon", LEMMA: "go", NORM: "going"},
    {ORTH: "na", LEMMA: "to"}
]
```

(Btw, I added the above example as a quote instead of code, because for some reason, the characters won't render properly for me in the monospace font: `কি.মি`. Are you seeing this too / is this an issue with GitHub's font stack choice? Or is this just a problem with my local font settings or browser?)

shuvanon:
Thank you @ines for pointing out my mistake and suggest some possible solution.
I will solve this soon.

ines:
Looks good – fixed a small typo and merging this now!

Issue #669, opened 2016-12-03
Add `lower_` property to `Span` class?


bdewilde:
This is a feature request, not a bug. (Is there a better place to post re: enhancements?)

For convenience, it would be great to have a `lower_` property on spans analogous to the [`lemma_`](https://github.com/explosion/spaCy/blob/master/spacy/tokens/span.pyx#L364) property. Looks like it would be a super straightforward change, maybe even just

```
cdef class Span:
    ...
    property lower_:
        def __get__(self):
            return ' '.join([t.lower_ for t in self]).strip()
```

Thanks in advance!

honnibal:
Maybe! If so, we should support the other string methods too, right?

I think we should think more deeply about an ABC over `Token`, `Span` and `Doc`.

bdewilde:
Agreed! I just didn't want to ask for more than I had immediate use for. :)

bhargavvader:
I'd like to take a shot at this; any pointers?

honnibal:
Sorry for letting this slip --- adding the other string methods sounds good. Adding the help wanted tag here.

The file to change is https://github.com/explosion/spaCy/blob/master/spacy/tokens/span.pyx

Tests would go here: https://github.com/explosion/spaCy/blob/master/spacy/tests/spans/test_span.py

ines:
Fixed in #883!

PR #883, opened 2017-03-11
Add `lower_` and `upper_` properties and `map` fn to `Span` class

ericzhao28:
Added lower_ and upper_ properties as described in https://github.com/explosion/spaCy/issues/669.
I also added a generic map function to execute on the string representations of each token in a span.

## Types of changes
I added some simple features via very minor additions to `spacy/tokens/span.pyx`, and added corresponding tests to `spacy/tests/spans/test_span.py`.

## Checklist:
- [ ] My change requires a change to spaCy's documentation. (Not sure about this, sorry)
- [ ] I have updated the documentation accordingly. (Should I do so in this case?)

This is my first commit to a major open-source project - please let me know if I should fix/add anything, and apologies ahead of time in case I somehow break something.

- Eric

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,7 +106,6 @@ website/announcement.jade
 website/www/
 website/.gitignore
 
-# Personal (Eric)
+# Python virtualenv
 venv
 venv/*
-.gitignore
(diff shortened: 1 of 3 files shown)


ericzhao28:
Hi,
I removed mapStr and pushed that commit with only `upper_` and `lower_`. I guess mapStr is a bit extraneous.
- Eric

honnibal:
Thanks!

PR #901, opened 2017-03-21
Split CONLLX file using tabs and not default split separators

raphael0202:
I encountered various errors while trying to train the parser for French using train_ud.py script.

- the different columns of the CONLLX file were splitted using split function default separators, instead of tabs. It could lead to an error if one field contains space (for instance the token '100 000'), as the number of field is greater than expected.
- use of the LangGlass defined above instead of the base Language class to instanciate the language.
Without this, I got the following error, due to recent change in Spacy Language class (commit 9605cf39cc1c8119535616ebd1df0d011cb4e9f8):

```
  File "train_ud.py", line 139, in <module>
    plac.call(main)
  File "/home/raphael/.virtualenvs/spacy/lib/python3.5/site-packages/plac_core.py", line 328, in call
    cmd, result = parser.consume(arglist)
  File "/home/raphael/.virtualenvs/spacy/lib/python3.5/site-packages/plac_core.py", line 207, in consume
    return cmd, self.func(*(args + varargs + extraopts), **kwargs)
  File "train_ud.py", line 132, in main
    nlp = Language(vocab=vocab, tagger=tagger, parser=parser)
  File "/home/raphael/git_core/spaCy/spacy/language.py", line 282, in __init__
    path = util.get_data_path() / self.lang
  File "/usr/lib64/python3.5/pathlib.py", line 889, in __truediv__
    return self._make_child((key,))
  File "/usr/lib64/python3.5/pathlib.py", line 681, in _make_child
    drv, root, parts = self._parse_args(args)
  File "/usr/lib64/python3.5/pathlib.py", line 643, in _parse_args
    % type(a))
TypeError: argument should be a path or str object, not <class 'NoneType'>
```
## Types of changes

## Checklist:

### DIFF
--- a/bin/parser/train_ud.py
+++ b/bin/parser/train_ud.py
@@ -1,18 +1,13 @@
 from __future__ import unicode_literals
 import plac
 import json
-from os import path
-import shutil
-import os
 import random
-import io
 import pathlib
 
 from spacy.tokens import Doc
 from spacy.syntax.nonproj import PseudoProjectivity
 from spacy.language import Language
 from spacy.gold import GoldParse
-from spacy.vocab import Vocab
 from spacy.tagger import Tagger
 from spacy.pipeline import DependencyParser, BeamDependencyParser
 from spacy.syntax.parser import get_templates
@@ -21,7 +16,6 @@ from spacy.scorer import Scorer
 from spacy.language_data.tag_map import TAG_MAP as DEFAULT_TAG_MAP
 import spacy.attrs
 import io
-
 
 
 def read_conllx(loc, n=0):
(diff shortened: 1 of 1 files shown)


honnibal:
Sorry I've got some local changes that need to be pushed on this. Give me a few minutes.

raphael0202:
Well, as a start I think some functions may be properly unit-tested (for instance the one reading CONNLLX files).

ines:
@raphael0202 That's a good idea – we were actually thinking about adding a `train` command to the new `cli` module. Not sure yet how exactly this should be set up to make it as practical as possible (suggestions welcome!).

Also, in case you wanna test it for models you're training: The `develop` branch has an (experimental) implementation of a `package` command that creates a Python package from a model data directory and lets you set up the meta data all from the command line. So you can do:

```bash
python -m spacy package /your/model/data /output/parent/directory
```

PR #902, opened 2017-03-22
Fix infinite recursion in spacy.info

pavlin99th:
## Description

Remove definition of `info` from `__init.py__` as it is imported from `cli` module. Removed function caused `RecursionError: maximum recursion depth exceeded`.

## Types of changes

## Checklist:


ines:
Good catch, thanks – this shouldn't have been there!

Issue #909, opened 2017-03-23
Model symlinks not working on Windows (Python 2.7)


solresol:
It's not exactly spaCy's fault. It looks like it's a pathlib problem creating symlinks on Windows.

Anyway, on a fairly fresh Win2k8 (an Amazon workspaces desktop, in fact) pip install spacy works as long as the compiler is in place.

 `python -m spacy download en`  fails with the following error message:

> Traceback (most recent call last):
  File "c:\python27\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "c:\python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "c:\python27\lib\site-packages\spacy\__main__.py", line 71, in <module>
    plac.Interpreter.call(CLI)
  File "c:\python27\lib\site-packages\plac_ext.py", line 1142, in call
    print(out)
  File "c:\python27\lib\site-packages\plac_ext.py", line 914, in __exit__
    self.close(exctype, exc, tb)
  File "c:\python27\lib\site-packages\plac_ext.py", line 952, in close
    self._interpreter.throw(exctype, exc, tb)
  File "c:\python27\lib\site-packages\plac_ext.py", line 964, in _make_interpreter
    arglist = yield task
  File "c:\python27\lib\site-packages\plac_ext.py", line 1139, in call
    raise_(task.etype, task.exc, task.tb)
  File "c:\python27\lib\site-packages\plac_ext.py", line 380, in _wrap
    for value in genobj:
  File "c:\python27\lib\site-packages\plac_ext.py", line 95, in gen_exc
    raise_(etype, exc, tb)
  File "c:\python27\lib\site-packages\plac_ext.py", line 966, in _make_interpreter
    cmd, result = self.parser.consume(arglist)
  File "c:\python27\lib\site-packages\plac_core.py", line 207, in consume
    return cmd, self.func(*(args + varargs + extraopts), **kwargs)
  File "c:\python27\lib\site-packages\spacy\__main__.py", line 45, in link
    cli_link(origin, link_name, force)
  File "c:\python27\lib\site-packages\spacy\cli\link.py", line 14, in link
    symlink(origin, link_name, force)
  File "c:\python27\lib\site-packages\spacy\cli\link.py", line 50, in symlink
    link_path.symlink_to(model_path)
  File "c:\python27\lib\site-packages\pathlib.py", line 1167, in symlink_to
[...]

PR #921, opened 2017-03-25
Possible solution to #909

solresol:
## Description

For Python 2 on Windows, pathlib's symlink capability doesn't work. Instead, we shell out to mklink

## Types of changes

## Checklist:


ines:
Thanks a lot for taking this on – looks good!

I think I'll also move the platform and version check out to helper functions, so we can do stuff like `is_windows()` and `is_python2()` (because I'm sure there might be other stuff like this that needs to be fixed in the future).

Issue #934, opened 2017-03-28
Linking to model success but not visible under spacy info.


jreeter:
I have downloaded the md model and running the link with:
```
PS C:\dev\rasa_nlu> python -m spacy link c:\Python27\Lib\site-packages\en_core_web_sm en --force
Invalid switch - "Python27".

    Linking successful

    c:/Python27/Lib/site-packages/en_core_web_sm -->
    C:/Python27/lib/site-packages/spacy/data/en

    You can now load the model via spacy.load('en').
```

However it doesn't seem to be linked under spacy :

```PS C:\dev\rasa_nlu> python -m spacy info

    Info about spaCy

    Python version     2.7.13
    Platform           Windows-10-10.0.14393
    spaCy version      1.7.3
    Installed models
    Location           C:\Python27\lib\site-packages\spacy
```

## Info about spaCy

* **Python version:** 2.7.13
* **Platform:** Windows-10-10.0.14393
* **spaCy version:** 1.7.3
* **Installed models:**

ines:
Thanks for the report! This is interesting... I need to test this, might be related to the `pathlib` issues with symlinks on Windows / Python 2 (see #909). Maybe the symlinks are not recognised correctly as directories [here](https://github.com/explosion/spaCy/blob/master/spacy/cli/info.py#L52) (which, according to `pathlib`, should work fine).

When you have a second, could you run the following code and tell me if it's getting the paths correctly and if not, which one fails / returns an empty list?

```python
from spacy.util import get_data_path

data_path = get_data_path()
print([p for p in data_path.iterdir()])
print([p for p in data_path.iterdir() if p.isdir()])
print([p.parts[-1] for p in data_path.iterdir()]
print([p.parts[-1] for p in data_path.iterdir() if p.isdir()]
```

jreeter:
@ines per your request:

```
from spacy.util import get_data_path data_path = get_data_path()print([p for p in data_path.iterdir()])print([p for p in data_path.iterdir() if p.is_dir()])print([p.parts[-1] for p in data_path.iterdir()])print([p.parts[-1] for p in data_path.iterdir() if p.is_dir()])

Find
No Results

from spacy.util import get_data_path

data_path = get_data_path()
print([p for p in data_path.iterdir()])
print([p for p in data_path.iterdir() if p.is_dir()])
print([p.parts[-1] for p in data_path.iterdir()])
print([p.parts[-1] for p in data_path.iterdir() if p.is_dir()])
[...]

PR #942, opened 2017-03-31
Issue #934 symlink should not convert paths as_posix under windows.

jreeter:
## Description
symlink function in link.py was incorrectly converting WindowPaths to the posix version. Per Python docs Path should correctly instantiate as a WindowsPath and generate paths with backslashes. Windows process mklink expects paths to be valid windows paths(with backslashes).

## Types of changes
- [ X] **Bug fix** (non-breaking change fixing an issue)

## Checklist:

### DIFF
--- a/spacy/cli/link.py
+++ b/spacy/cli/link.py
@@ -46,7 +46,7 @@ def symlink(model_path, link_name, force):
     # Add workaround for Python 2 on Windows (see issue #909)
     if util.is_python2() and util.is_windows():
         import subprocess
-        command = ['mklink', '/d', link_path.as_posix(), model_path.as_posix()]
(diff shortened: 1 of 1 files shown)


ines:
Thank you! Copy-pasting over my reply from #934:

> Thanks so much for your analysis and submitting the PR 👍
>
> The `mlink` fix originates from the symlink issues described in #909 and it seems to have solved the issue for this particular case. So I can only guess that in some setups, `is_posix()` is resolved correctly to the backslash Windows path, and in others, like yours, it isn't?
> 
> I hope there's no other `pathlib` bug or weirdness going on here and the `Path` is really interpreted correctly across Windows versions etc. Otherwise we might have to introduce some additional checks.

PR #943, opened 2017-03-31
Portuguese improvements

mamoit:
## Description
Improve the Portuguese language word lists.
Let me know if I can help with something PT related.

## Types of changes

## Checklist:

### DIFF
--- a/spacy/pt/stop_words.py
+++ b/spacy/pt/stop_words.py
@@ -3,18 +3,19 @@ from __future__ import unicode_literals
 
 
 STOP_WORDS = set("""
-à às acerca adeus agora ainda algo algumas alguns ali além ambos ano
-anos antes ao aos apenas apoio apontar após aquela aquelas aquele aqueles aqui
-aquilo area área as assim através atrás até aí
+à às acerca adeus agora ainda algo algumas alguns ali além ambas ambos ano
+anos antes ao aos apenas apoio apoia apontar após aquela aquelas aquele aqueles
+aqui aquilo área as assim através atrás até aí
 
-baixo bastante bem bom breve
+baixo bastante bem boa bom breve
 
 cada caminho catorze cedo cento certamente certeza cima cinco coisa com como
-comprido conhecido conselho contra corrente custa cá
+comprido comprida conhecida conhecido conselho contra corrente custa cá
 
-da daquela daquele dar das de debaixo demais dentro depois desde desligado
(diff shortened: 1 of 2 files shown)


mamoit:
Do not merge yet, I still need to review the whole stop word list.

ines:
Thanks – just post an update as soon as it's ready to merge!

A next, easy step would be to [add tokenizer exceptions](https://spacy.io/docs/usage/adding-languages#tokenizer-exceptions), especially abbreviations and possibly contractions (for inspiration, see [`en/tokenizer_exceptions.py`](https://github.com/explosion/spaCy/blob/master/spacy/en/tokenizer_exceptions.py)).

For contractions, only include closed-class contractions here. If the language has open-ended contractions (like verbs in Spanish), those need to be dealt with at a later point during lemmatization. A contraction exception could look something like this:

```python
TOKENIZER_EXCEPTIONS = {
    "do": [
        {ORTH: "d", NORM: "de"},
        {ORTH: "o"}
    ]
}
```

I don't speak Portuguese, so I just googled for Portuguese contractions, sorry 😜 The keys need to always match the two `ORTH` values combined. The way the original token is split up can be a bit arbitrary sometimes... for example, "num" would probably become "n" (norm: "em") and "um". If possible, you can also add the `TAG` or `LEMMA` here. For more info on this, see the [docs](https://spacy.io/docs/usage/adding-languages#tokenizer-exceptions).

mamoit:
@ines quick question... In cases such as:
```
a + a = à
a + as = às
```
how would you construct that exception?
I'm quite new to NLP, so I'm trying to keep up with the definitions.

ines:
@mamoit The first one doesn't really work, because "à" can't be split up, and we can't have a zero-width token. The second one could be something like:

 ```
"às": [
    {ORTH: "à", NORM: "a"},
    {ORTH: "s", NORM: "as"}
]
```

If you need more inspiration, there are also some good tokenizer exceptions examples on the other languages – I think French, Hungarian and Swedish are quite far along as well now.

ines:
Just merged and made one small edit (ad8bf18) to make sure the exceptions are actually loaded and merged with the defaults, like emoticons.

When you have time, it'd also be great to have a few simple tests for Portuguese, to make sure it all works as expected. Something simple like the [French tests](https://github.com/explosion/spaCy/tree/master/spacy/tests/fr) would be totally fine for a start!

mamoit:
:+1:  I had something urgent to attend to, so I could not go any further yesterday.

Issue #906, opened 2017-03-22
Unexpected lemmatization for "it" / "its"


davechallis:
Basic example:

    import spacy
    nlp = spacy.load('en')
    for token in nlp("IT it ITS its"):
        print(token, token.lemma_)

Output:

    IT it
    it -PRON-
    ITS it
    its -PRON-

Expected:

    IT it
    it it
    ITS it
    its it

## Info about spaCy

* **spaCy version:** 1.7.2
* **Platform:** Darwin-16.4.0-x86_64-i386-64bit
* **Python version:** 3.6.0
* **Installed models:** en

honnibal:
This is the expected behaviour --- the 1.6 behaviour was a regression.

If you find the current behaviour particularly inconvenient, could you describe why? There seems to me no really satisfying solution for the lemma for pronouns, but I understand that many don't like the current policy.

ddofer:
Well, it's not what some users might expect naively from lemmatization ("let's find the root meaning/word"). It can also have a fair amount of info loss, e.g. "they" vs "I". 
(I came acorss this thread while trying to understand where the "-PRON-" in my lemmatized text came from, I thought i'd messed it up).

davechallis:
I wouldn't say that I found it inconvenient, it was more that the output was unexpected, so thought it was a bug rather than intended (i.e. I thought the lemmatizer would always return a word, not a part of speech).

I'm not a linguist, so don't know what the correct lemma is, this is the only really unexpected thing I encountered when moving from NLTK's lemmatization to Spacy's, so perhaps it's just a documentation issue.

ines:
Closing this to keep the discussion in #962.

Issue #898, opened 2017-03-21
Dependencies not deprojectivized in spaCy 1.7


adam-ra:
I've noticed suspiciously large amount of evident parser errors after migrating from Spacy 1.6.0 and generic ‘en’ model to 1.7.2 + ‘en_depent_web_md’.

Environment: Python 3.4.3 / 3.5.1 on 64-bit Linux (Ubuntu).

Some example below (please note the abundance of proper noun tags).

1.6.0: pain in lower back
```
 0) ROOT    pain                    NOUN  NN    ROOT
 1) prep   ---- in                  ADP   IN    prep
 3) pobj   -------- back            NOUN  NN    pobj
 2) amod   ------------ low         ADJ   JJR   amod
```

1.7.2: pain in lower back
```
 3) ROOT    back                    PROPN NNP   ROOT
 0) nsubj  ---- pain                NOUN  NN    nsubj
 1) nmod   ---- in                  X     XX    nmod
 2) compound ---- lower             PROPN NNP   compound
```

1.6.0: I feel pain in lower back
```
 1) ROOT    feel                    VERB  VBP   ROOT
 0) nsubj  ---- i                   PRON  PRP   nsubj
 2) dobj   ---- pain                NOUN  NN    dobj
 3) prep   -------- in              ADP   IN    prep
 5) pobj   ------------ back        NOUN  NN    pobj
 4) amod   ---------------- low     ADJ   JJR   amod
```

1.7.2: I feel pain in lower back
```
 1) ROOT    feel                    VERB  VBP   ROOT
 0) nsubj  ---- i                   NOUN  NN    nsubj
 2) dobj   ---- pain                NOUN  NN    dobj
 3) prep   -------- in              ADP   IN    prep
 5) pobj   ------------ back        PROPN NNP   pobj
 4) compound ---------------- lower PROPN NNP   compound
```

1.6.0: sores on my dick
```
 0) ROOT    sore                    NOUN  NNS   ROOT
 1) prep   ---- on                  ADP   IN    prep
 3) pobj   -------- dick            NOUN  NN    pobj
 2) poss   ------------ my          ADJ   PRP$  poss
```

1.7.2: sores on my dick
```
 0) ROOT    sore                    NOUN  NN    ROOT
 1) prep   ---- on                  ADP   IN    prep
 3) pobj   -------- dick            PROPN NNP   pobj
 2) compound ------------ my        PROPN NNP   compound
```

honnibal:
[...]

PR #952, opened 2017-04-03
Add contractions for won't

kinow:
Hi, I tried implementing the fix for this issue, but alas it did not work in my local environment.

The issue is that Won't & won't return lemma wo and nt. Whereas I would expect to get will not.

Saw a similar issue for Let's -> Let us, but it also didn't work for me. I wonder if there is any documentation on how to test the tokenizer exceptions before submitting pull requests? Anyway, feel free to close it and implement in a better way if necessary.

Cheers
Bruno

## Types of changes

- [ X ] **Bug fix** (non-breaking change fixing an issue)

## Checklist:

- [ X ] All new and existing tests passed.

### DIFF
--- a/spacy/en/tokenizer_exceptions.py
+++ b/spacy/en/tokenizer_exceptions.py
@@ -491,6 +491,16 @@ OTHER = {
         {ORTH: "'s", LEMMA: PRON_LEMMA, NORM: "us"}
     ],
 
+    "won't": [
+        {ORTH: "wo", LEMMA: "will"},
+        {ORTH: "n't", LEMMA: "not"}
+    ],
+
+    "Won't": [
+        {ORTH: "Wo", LEMMA: "will"},
+        {ORTH: "n't", LEMMA: "not"}
+    ],
+
     "\u2014": [
         {ORTH: "\u2014", TAG: ":", LEMMA: "--"}
     ],


ines:
Thanks for giving this a go!

I just had a look at this problem and it turned out that it was caused by a missing POS tag in the tokenizer exceptions. The exceptions for the verbs are already handled [here](https://github.com/explosion/spaCy/blob/master/spacy/en/tokenizer_exceptions.py#L209), and they did have the correct lemma. But because they were missing a `TAG`, it was later overwritten by the lemmatizer. 

I added the missing tags and it should work properly now!

kinow:
Thanks @ines.

I couldn't understand why it was not working, and started looking at the Tokenizer class. Thanks for pointing where I was supposed to be looking at :-) 

Will give it a try today.
Bruno

kinow:
Arrived home, checked out the latest version

```
git log -n 1
commit 808cd6cf7f184e20d9b8e42364f7e10f045028dc
Author: ines <ines@ines.io>
Date:   Mon Apr 3 18:12:52 2017 +0200

    Add missing tags to verbs (resolves #948)
```

Then `python setup.py build && python setup.py install`. Executed a test code.

```
from spacy.en import English
from spacy.symbols import ORTH, LEMMA, POS

from pprint import pprint as pp

nlp = English()

text = "Don't you use NLP? Won't you need it? Let's use it !!!"

tokens = nlp(text)
for token in tokens:
    print("%s - %s - %s" % (token.text, token.lemma_, token.pos_))
```

Got:

```
Do - do - 
n't - not - ADV
you -  - 
use -  - 
NLP -  - 
? -  - 
Wo - will - VERB
n't - not - ADV
you -  - 
need -  - 
it -  - 
? -  - 
Let - let - 
's - -PRON- - 
use -  - 
it -  - 
! -  - 
! -  - 
! -  -
```

Then downloaded the models `python3 -m spacy.en.download --force all`, and re-executed the code:

```
Do - do - VERB
n't - not - ADV
you - -PRON- - PRON <---- ??? why the lemma here is not you?
use - use - VERB
NLP - nlp - PROPN
? - ? - PUNCT
Wo - will - VERB
n't - not - ADV
you - -PRON- - PRON <--- ditto
need - need - VERB
it - -PRON- - PRON <--- ditto
? - ? - PUNCT
Let - let - VERB
's - 's - PRON <----- why not us here?
use - use - VERB
it - -PRON- - PRON
! - ! - PUNCT
! - ! - PUNCT
! - ! - PUNCT
```

Am I missing anything? Maybe I should have executed something else, or there's something wrong with my code?

f11r:
Regarding lemmatisation to `-PRON-` see: https://github.com/explosion/spaCy/issues/906 and https://github.com/explosion/spaCy/issues/898#issuecomment-288164755.

kinow:
Thanks @f11r 

I am working on an application that would need the values to be matched against a dictionary. When I have have contractions, like "Let's", I need the values let and us to match against a list of words.

For "Let's", the lemmas will have what I am looking for (i.e. [let, us]). But for "It", if I use the lemma, then I will get -PRON- - if I understand it correctly.

In this case I would either have to think about another strategy, or maybe always use lemma and, if I find any word surrounded by dashes, fallback to using the text.

PR #982, opened 2017-04-13
Fixed python2 issue with the package command

oroszgy:
## Description

When running `python -m spacy package` the package creation failed with an error in Python 2 environments. It was due to `json.dumps` returns `str` not `unicode`. The fix aims to make the CLI work equally well on both python versions.

## Types of changes

- [X] **Bug fix** (non-breaking change fixing an issue)

## Checklist:

### DIFF
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -6,9 +6,15 @@ import shutil
 import requests
 from pathlib import Path
 
+import six
+
 from .. import about
 from .. import util
 
+if six.PY2:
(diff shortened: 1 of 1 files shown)


ines:
Thanks – I really like that solution! 👍 

For spaCy v2.0, we definitely want to come up with a better way of standardising the compatibility checks across the code base. This is all pretty messy at the moment.  Maybe a `spacy.compat` helper would be a good solution for this. All functions that require different versions depending on the platform or Python version could then be imported directly from there. (I'll open a ticket on the issue tracker about this as well.)

PR #996, opened 2017-04-19
Fix for issue 995

beneyal:
## Description

Where the `re` module was used, I changed it to `import regex as re` to utilize the new features without breaking the API.
The important thing to notice regarding this change is that `regex.DEFAULT_VERSION` should be assigned the value `regex.VERSION1` before spaCy does anything `regex`-related. Right now, this assignment is in the `punctuation.py` file, and the tests pass, but I'm not 100% that this is the right place for it.

Alphabetic characters are no longer defined manually, but are defined using `regex`'s `\p{...}` syntax together with the support for set operations.
To retain the previous behavior, `ALPHA_LOWER` and `ALPHA_UPPER` were changed to `LATIN_LOWER` and `LATIN_UPPER`, respectively, using intersection of the Unicode Latin script and the Letter categories.
I think the only non-latin languages that needed attention were Hebrew and Bengali, so I added the patterns for them. Note that they are intersected with the entire Letter category (I think they are both in "Letter, other").

The definitions for ALPHA_LOWER, ALPHA_UPPER, and ALPHA are just the relevant set unions of the above definitions.

pytest reports 1175 passes, 2285 skipped, and 10 xfailed.

Thank you, @ines, for the tests and kind guidance :)

## Types of changes

## Checklist:


ines:
Thanks so much for taking this on and the detailed PR – looks great! 👍  
I'll also add a little note on this in the docs on adding languages.

PR #1014, opened 2017-04-24
Make object of the deep learning tutorial clearer

julien-c:
This is a great tutorial, but I think it is weirdly explained in the current form. The largest part of the code is about implementing the actual sentiment analysis model, not about counting entities. (which is not even present in the `deep_learning_keras.py` script in `examples`)


honnibal:
Thanks!

Issue #846, opened 2017-02-18
💫 Generic lemmatization & morphology


honnibal:
Simple and generic lemmatization functionality can be implemented as a look-up table. The lemmatization lists here look good: http://www.lexiconista.com/datasets/lemmatization/

To make this work, we need a simple class that behaves similarly to the existing `spacy.lemmatizer.Lemmatizer` class. A language subclass can then create this lookup lemmatizer in its `Language.Defaults.create_lemmatizer()` method.

This will give us a low-effort start on lemmatization in lots of languages, that can be replaced by more sophisticated strategies on a case-by-case basis.

## Wishlist extension

It would be very interesting to try a sequence-to-sequence model to generate the lemmas, using the existing lists as training. The sequence-to-sequence model would then be used when the lookup fails. I think this might perform quite well, especially if a POS tag can be supplied as a feature.

To be specific, the sequences are the characters, and the problem is analogous to neural machine translation. So, example NMT architectures would be the best place to start.

## Details

* **Difficulty:** easy / suitable for new contributors. The extension is more complicated and requires some ML experience.
* **Files to change:** [`lemmatizer.py`](https://github.com/explosion/spaCy/blob/master/spacy/lemmatizer.py), [`language.py`](https://github.com/explosion/spaCy/blob/master/spacy/language.py), possibly [`morphology.pyx`](https://github.com/explosion/spaCy/blob/master/spacy/morphology.pyx)
* **Tests:** `tests/lemmatizer` (not created yet)
* **Related issues:** #390

hiteshpaul:
Already started to implement!

ines:
Copying over the current state from #390 and #974 (and closing those) to make this one the master issue 🎉

**Comment by @Liebeck on #390:**
> Hi, I just wanted to point you to one of my older projects called IWNLP which produces a list of form -> lemma (e.g., Schwimmbäder -> Schwimmbad) for German words based on Wiktionary. Check out https://github.com/Liebeck/IWNLP and www.iwnlp.com or query me for more information.
The produced mappings are in the form of (form, POS) -> lemma, are under a CC-lincense, and I'd love to see them implemented ;) I might be able to dump another format (or a generic format, if specified) if that's desired? The latest evaluation results are listed here: http://www.iwnlp.com/iwnlp_results.html

**Response by @honnibal on #390:**
> This is really nice! We'd definitely like to see this implemented. Sorry I missed this comment before!
>
[...]

PR #1024, opened 2017-04-27
Lemmatization (lookup) and some corrections in the french language

Gregory-Howard:
## Description
Initial lemmatization feature following : #846
I would like to change the name of spacy/lemmatizerlookup.py,it's awfull and I'm not inspired.
Bug fix : adding title insensitive tokenization in french (there was no issue)

_Added 2 xfail tests and reasons, because when a word is tagged as PROPN or AUX it's not lemmatized_

## Types of changes

- [X] **Bug fix** (non-breaking change fixing an issue)
- [X] **New feature** (non-breaking change adding functionality to spaCy)

## Checklist:


honnibal:
Thanks, this is really nice! This has been outstanding for way too long.

A thought: currently are the lemmatizer files imported by default? I still think it's better to have this data in `.py` files instead of `json`, but I think we should be careful to avoid importing the module until it's used.

Gregory-Howard:
Yeah lemmatizer are already imported for every language. Should I write files with the ```python -m spacy download en``` command ?

honnibal:
It would be nice to find a solution where the languages are lazy-loaded. I'll think about how to do this --- probably a bit of crafty magic in `spacy/__init__.py`.

We wouldn't want to use the `download` command --- that's really just for the large binary data packs.

Gregory-Howard:
Something like this ? : 
``` 
def __init__(self):
    super()
    if not LOOK_UP:
        # wget somerepo/spacy/lang/lemmatize.py
        # or
        # wget http://www.lexiconista.com/Datasets/lemmatization-lang.zip
        # parsing and save in repo
```

honnibal:
Listing out solutions:

1. Lemmatizer data is in repository as .py files
2. Lemmatizer data is in repository as data files
3. Lemmatizer data is packaged in model
4. Lemmatizer data is downloaded in separate mechanism, outside of model

I think 1 is okay --- the only drawback is that we have to make sure that when you do `import spacy`, you don't accidentally load 500,000 lines of lemmatizer data for languages you don't care about. That's why I want some sort of lazy-loading.

Currently the problem is that we import the different languages within the `spacy/__init__.py` file --- and the different languages tend to import the data for the tokenizer etc within their package `__init__.py`.

So currently:

* User does `import spacy`, executing `spacy/__init__.py`
* `spacy/__init__.py` does `import fr, hu, de` etc
* These ultimately execute `import spacy.fr.tokenizer_exceptions, spacy.fr.lemmatizer_data`, etc

Adding up over all the languages, the current tokenizer exceptions are already leading to a lot of data being loaded. I think the lemmatizer data in this PR tips this over into being too much.

What we need is for the data not to be loaded until the user does `nlp = spacy.load('fr')`. At that point, it's fine for the lemmatizer data to be imported and loaded.

Gregory-Howard:
Lemmatization files are already .py not json files.
Created with : https://pastebin.com/Q34YfaWt (need to improve it a bit)

Okay I fully understand the problem. 
http://stackoverflow.com/questions/128478/should-python-import-statements-always-be-at-the-top-of-a-module
>The best reasons I've seen to perform lazy imports are:

> -Optional library support. If your code has multiple paths that use different libraries, don't break if an optional library is not installed.
> -In the __init__.py of a plugin, which might be imported but not actually used. Examples are Bazaar plugins, >which use bzrlib's lazy-loading framework.

Gregory-Howard:
I'm not sure if what I've done works in python 2.7 normally yes but this time it's a **Breaking change**

honnibal:
Really thanks a lot for doing this --- you've really had to wade into lots of not-very-fun details of Python on this. Getting the lemmatization in and getting this all right will be really great.

ines:
Thank you – and I agree, this will definitely make a big difference!

It'll be even better in v2.0 once we've made some more changes to the language data. Now that there are so many languages, things have been getting a little messy again... 

Here are some of the ideas we've had:

* Move all languages to a module `languages` (just need to find a good solution to still allow `from spacy.en import English` etc.)
* Improve the way a `Language` subclass and its `Defaults` are created and make it more obvious which components are needed and where they are coming from (also get rid of random wildcard imports)
* Improve the way tokenizer exceptions are put together – maybe create one, more powerful helper function or class, instead of relying on random helpers like `update_exc()`, `strings_to_exc()`, `expand_exc()` etc.
* Include word lists like `NUM_WORDS` and all the hard-coded stuff that's currently in [`orth.pyx`](https://github.com/explosion/spacy/blob/master/spacy/orth.pyx) with the language data. Each language could define its own functions for things like `like_number`, which are then used to set the respective flags. This would also make it much easier to add more flags in the future.

Gregory-Howard:
>Move all languages to a module languages (just need to find a good solution to still allow from spacy.en import English etc.)
In which file do you want to do `spacy.en import English`?

ines:
@Gregory-Howard Ah, I meant, we shouldn't break backwards compatibility for things like:

```python
from spacy import en
from spacy.en import English
```

At least, I think it'd be a bad idea to just move `spacy.en` to `spacy.languages.en` in v2.0 and break people's code.

Gregory-Howard:
Oh okay, I have no idea how we can allow backwards compatibility for this :/

Gregory-Howard:
Actually we can have into spacy/\_\_init\_\_.py
``` from spacy.languages.en import *
en = spacy.languages.en.English
```
Or 
``` from spacy.languages.en import English as en ```

But with this metho we can't have lazy loading AND backwards compatibility

Liebeck:
@honnibal You opted for "Lemmatizer data is in repository as .py files"

Won't that design be too restrictive for other resources that might clash due to license issues? I'm copying my unaswered comment from #846: 

`Do you want me to include the mapping from IWNLP directly into the code of spaCy? Won't there be a clash between your MIT License and the CC license (which I'm forced to use since the Wiktionary data is CC licensed)? I guess an easy way to circumvent this would be to create a file format/model which can then be downloaded by spaCy?`

honnibal:
@Liebeck I don't think there's a clash. It follows the same logic as porting in code originally released under another license. Two scenarios:

1. The other license is more restrictive, e.g. GPL. In this scenario spaCy can't depend on the data.
2. The other license is less restrictive, e.g. CC-by-SA or Apache. In this situation, I believe we're allowed to distribute the data, including the original license or attribution as necessary.

I don't think exposing the data as text vs as .py files changes anything.

ines:
See #1413 🎉  (closing this PR since it's all covered on `develop` now)

PR #1025, opened 2017-04-27
fix typo, CONLL format

Ferdous-Al-Imran:
tried to google about connlu format. Saw there is conll format, not connlu.

## Types of changes

## Checklist:

### DIFF
--- a/website/docs/usage/adding-languages.jade
+++ b/website/docs/usage/adding-languages.jade
@@ -544,7 +544,7 @@ p
 p
     |  You can now train the model using a corpus for your language annotated
     |  with #[+a("http://universaldependencies.org/") Universal Dependencies].
-    |  If your corpus uses the connlu format, you can use the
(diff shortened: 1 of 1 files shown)


ines:
Sorry this was confusing!

In this case, we do have to make it clear that `spacy convert` only supports *one* of the CoNLL formats, i.e. [CoNLL-U](http://universaldependencies.org/docs/format.html) (files with the `.conllu` extension). See [here](https://github.com/UniversalDependencies/UD_Spanish-AnCora) for an example corpus. We definitely want to add more converters in the future, though.

To make the docs less confusing, they should probably use the proper spelling and capitalisation ("CoNLL-U" instead of "conllu"), and link to the Universal Dependencies site for explanation. I'll also include a mention of the file extension.

Ferdous-Al-Imran:
Great.

PR #1039, opened 2017-05-03
Corretions for model test example

akYoung:
The sentences of test data in sentence entailment example should be generated with integers limited to vocab_size.


honnibal:
Thanks!

Issue #1069, opened 2017-05-18
`Doc.sents` returns a generator even when a non-co-routine user-hook is defined


raphael0202:
Here is a snippet to reproduce the bug:
```
import spacy

class SentenceBoundaryDetector:
    def __init__(self, model):
        self._model = model

    def __call__(self, doc):
        doc.user_hooks['sents'] = self.sents

    @staticmethod
    def sents(doc):
        print("In SentenceBoundaryDetector.sents")
        return list(range(5))


def load_nlp(lang_id='en'):
    def create_pipeline(nlp):
        return [nlp.tagger, nlp.entity, SentenceBoundaryDetector('test')]
    return spacy.load(lang_id, create_pipeline=create_pipeline)

nlp = load_nlp('fr')
doc = nlp("Here is a sentence. Here is another one.")

print(doc.sents)
print(list(doc.sents))
```
Stdout:
```
<generator object at 0x7fe916f028c8>
In SentenceBoundaryDetector.sents
[]
```
The problem comes from `Doc.sents` being a generator, so returning `self.user_hooks['sents'](self)` (line 434) break the generator, which is therefore empty.

We need to iterate over `self.user_hooks['sents'](self)`, and yield each sentence from this.

## Your Environment

* Operating System: linux
* Python Version Used: 3.5
* spaCy Version Used: HEAD
* Environment Information:

PR #1070, opened 2017-05-18
Fix issue #1069 with custom hook `Doc.sents` definition

raphael0202:
See bug #1069 

## Types of changes

## Checklist:


honnibal:
Thanks! Interesting case.

Could we use `yield from` here? If it gives the same result I think it's clearer. In Cython this works in both Python 2 and Python 3.

raphael0202:
Yes, I didn't know yield from was compatible with python2 in Cython

PR #1076, opened 2017-05-19
Deleting (legacy?) whitespace attribute in doc

raphael0202:
token.whitespace now raises an AttributeError

## Types of changes

- [x] **Documentation** (addition to documentation of spaCy)

## Checklist:

### DIFF
--- a/website/docs/api/token.jade
+++ b/website/docs/api/token.jade
@@ -238,11 +238,6 @@ p An individual token — i.e. a word, punctuation symbol, whitespace, etc.
         +cell #[code text_with_ws]
         +cell unicode
         +cell Text content, with trailing space character if present.
-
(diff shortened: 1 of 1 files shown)


honnibal:
Hm, I don't remember intentionally deciding to deprecate this.

Maybe we should have something like this though? In the attrs the boolean flag is named `SPACY`, which is cute but a little misleading.

The complication is, how should this behave when we have a whitespace *token* like `\n`? This doesn't own a trailing whitespace necessarily, but it would potentially be irritating to have its `whitespace_` attribute return nothing. So we have to be careful about the naming.

Anyway. Merging this docs change, because no matter what we decide, the docs should reflect the library. Thanks again.

PR #1096, opened 2017-06-02
Spanish model features

frascuchon:
## Description

This PR :
- Fix the spanish sentence segmentation by reconfiguring tokenizer exceptions
- Add a spanish nouns chunker

## Types of changes

## Checklist:

### DIFF
--- a/spacy/es/tokenizer_exceptions.py
+++ b/spacy/es/tokenizer_exceptions.py
@@ -6,44 +6,15 @@ from ..language_data import PRON_LEMMA, DET_LEMMA
 
 
 TOKENIZER_EXCEPTIONS = {
-    "al": [
-        {ORTH: "a", LEMMA: "a", TAG: ADP},
-        {ORTH: "el", LEMMA: "el", TAG: DET}
-    ],
-
-    "consigo": [
-        {ORTH: "con", LEMMA: "con"},
-        {ORTH: "sigo", LEMMA: PRON_LEMMA, NORM: "sí"}
-    ],
-
-    "conmigo": [
-        {ORTH: "con", LEMMA: "con"},
-        {ORTH: "migo", LEMMA: PRON_LEMMA, NORM: "mí"}
-    ],
-
-    "contigo": [
-        {ORTH: "con", LEMMA: "con"},
-        {ORTH: "tigo", LEMMA: PRON_LEMMA, NORM: "ti"}
-    ],
-
-    "del": [
-        {ORTH: "de", LEMMA: "de", TAG: ADP},
-        {ORTH: "l", LEMMA: "el", TAG: DET}
-    ],
-
-    "pel": [
-        {ORTH: "pe", LEMMA: "per", TAG: ADP},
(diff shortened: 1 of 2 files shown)


ines:
Thanks, this look great! Will merge it here and then port it over to the develop branch as well.

Btw, in spaCy v2.0, we're also going to implement some improvements to the syntax iterators. Instead of having them live randomly in `/syntax`, each language can have its own `syntax_iterators.py`, which will be used by spaCy if it exists. The ultimate goal is to finally have all language-specific data grouped together within `spacy/lang` 😄

ines:
Just FYI, had to revert one change:

```python
"pal": [
    {ORTH: "pa", LEMMA: "para"},
    {ORTH: "l", LEMMA: DET_LEMMA, NORM: "el"}
],
```

The `ORTH` values always need to match the original string, so the second `ORTH` here has to be "l" and can't be "el" – otherwise, the exception won't match.

frascuchon:
oh, i didn't know it. Thanks ines!!

ines:
No worries! 

Btw, we just released v2.0.0 alpha, which finally makes better use of the `NORM` attributes. A language can now set [norm exception rules](https://alpha.spacy.io/docs/usage/adding-languages#norm-exceptions), similar to tokenizer exceptions, that map a string to its norm. For example, British spelling vs. American spelling in English, currency symbols or different styles of punctuation marks like `”` and `"`. The norms are used as features during training, which can improve accuracy, as it ensures that similar words receive similar representations (even if they weren't present in the training data.)

So this might be relevant for Spanish as well! 🙂

adam-ra:
I see you refer to lemmas in the above snippets… does this mean that lemmatisation is implemented for Spanish in 2.0? I've checked on 1.8.2 an it was not there. BTW I think lemmatisation is a key feature for any language (I'd say even more important than entities or semantic vectors) and it would be nice if it was included in the table at https://alpha.spacy.io/docs/usage/models#available

ines:
@adam-ra spaCy v2.0.0 alpha now supports simple, lookup-based lemmatization for English, German, French, Spanish, Italian, Hungarian, Portuguese and Swedish. For examples, see the `lemmatizer.py` files in [`spacy/lang`](https://github.com/explosion/spacy/tree/develop/spacy/lang). We're hoping to extend this in the future. 

Currently, info about all features, including lemmatization, is available in the API docs in the [Languages section](https://alpha.spacy.io/docs/api/language-models). But I agree that this is a bit messy at the moment... we've been struggling with finding a good format to present _all_ language and model info in one place, while still keeping it compact.

adam-ra:
@ines thanks for this info! Speaking of naive lemmatization, I guess for many languages it will fail miserably, especially for morphologically-rich languages such as Polish. I don't know how complex is Spanish inflection, probably pretty simple (at least for nominal phrases). Perhaps a still a bit crude but significantly better solution would be to base lemmatization on orth + POS tag (general wordclass), possibly having a fallback on orth only (as it is now). I'll look into the language models code and think it over anyway.

PR #1120, opened 2017-06-10
Implementation of dependency pattern-matching algorithm

raphael0202:
Currently, there is no easy way in spaCy to match a complex pattern with respect to a given dependency tree.  Stanford CoreNLP offers this kind of feature with its Semgrex patterns (https://nlp.stanford.edu/nlp/javadoc/javanlp/edu/stanford/nlp/semgraph/semgrex/SemgrexPattern.html).

This PR contains an implementation of a simple pattern matching algorithm, along with some high level wrapping classes.

A quick demonstration snippet:
```
import spacy
from spacy.pattern import PatternParser, DependencyTree

nlp = spacy.load('en')
doc = nlp("The quick brown fox jumped over the lazy dog.")
tree = DependencyTree(doc)

query = """fox [word:fox]=f
           [lemma:quick]=q >/am.*/ fox
           [word:/brown|yellow/] > fox"""

pattern = PatternParser.parse(query)
matches = tree.match(pattern)

assert len(matches) == 1
match = matches[0]

assert match['f'] == doc[3]
```
We start by creating a `DependencyTree` for the `Doc`. This class models the document dependency tree.
Then we compile the query into a `Pattern` using the `PatternParser`. The syntax is quite simple:
- we define a node named 'fox', that must match in the dep tree a token whose `orth_` is 'fox'.
- an anonymous token whose lemma is 'quick' must have fox as parent, with a `dep_` matching the regex `am.*`
- another anonymous token whose `orth_` matches the regex `brown|yellow` has fox as parent, with whathever `dep_`

`DependencyTree.match` returns a list of `PatternMatch`. Notice that we can assign names to anonymous or defined nodes (`[word:fox]=f`). We can get the `Token` mapped to the fox node using `match['f']`.

~~Unit tests are missing~~ (edit: unit tests added), and the docstrings should be improved a bit. I'm open to any suggestion or remarks :)

Disclaimer: To write the query parser, I took inspiration from jinja2, and used 3 classes from the project as baseline (Token, TokenStreamIterator and TokenStream). I don't know if there could be licensing issues because of this.

## Types of changes

## Checklist:

### DIFF
--- a/spacy/compat.py
+++ b/spacy/compat.py
@@ -16,6 +16,10 @@ try:
 except ImportError:
     import copyreg as copy_reg
 
+try:
+    import Queue as queue
+except ImportError:
+    import queue
 
 is_python2 = six.PY2
 is_python3 = six.PY3
@@ -32,6 +36,7 @@ if is_python2:
     basestring_ = basestring
     input_ = raw_input
     json_dumps = lambda data: ujson.dumps(data, indent=2).decode('utf8')
+    intern = intern
 
 elif is_python3:
     bytes_ = bytes
@@ -39,6 +44,7 @@ elif is_python3:
     basestring_ = str
     input_ = input
     json_dumps = lambda data: ujson.dumps(data, indent=2)
+    intern = sys.intern
 
 
 def symlink_to(orig, dest):
--- a/spacy/pattern/__init__.py
+++ b/spacy/pattern/__init__.py
@@ -1,0 +1,4 @@
+# coding: utf-8
+
+from .pattern import DependencyTree
+from .parser import PatternParser
--- a/spacy/pattern/parser.py
+++ b/spacy/pattern/parser.py
@@ -38,7 +38,17 @@ class PatternParser(object):
         if not pattern.nodes:
             return
 
+        cls.check_pattern(pattern)
         return pattern
+
+    @staticmethod
+    def check_pattern(pattern):
+        if not pattern.is_connected():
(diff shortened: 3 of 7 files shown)


honnibal:
Thanks for this! Trying to leave comments with GitHub's review functionality. Hopefully I don't mess things up.

raphael0202:
Hey! Sorry for my late reply. I can take care of 1. and 3. by myself. I think it's better than you deal with 2., as I'm not sure what changes are required to comply with the library style.
About 4., tests are already using pytest, they are only organized in classes for convenience. As these classes are stateless, it can however be splitted in functions too if you need it.
About 5., I can write the content, but I guess it will be easier if you take care of the formatting and the integration in spaCy documentation.

wadetb:
Is there a version of this patch for the spaCy 2.0 alpha?

raphael0202:
@honnibal Can you give me some indications about how you would like the documentation to be structured (new page distinct from 'Rule-based matching' or the same,...) ? And if you want the rename a few classes/functions, I can update them if you give me the new names.

honnibal:
@raphael0202 Apolgies for taking so long on this.

I retract the suggestion I made about class method vs global functions. It's actually better as you've implemented it, in a class --- this will make it easier to add a new parser if people want.

I'm going to go ahead and merge this, and port to v2. This leaves the documentation outstanding, but I'll add the example you give to the `spacy/exampes/` directory, and add a docs ticket for the docs.

PR #1126, opened 2017-06-12
Added html5lib==1.0b8 to requirements.txt

ianmobbs:
There's a frequent issue with `six` that occasionally causes spaCy to fail. The solution was found [here](https://stackoverflow.com/questions/39658404/python-cannot-import-name-viewkeys).


ines:
Thanks a lot for the fix and explanation – this is super valuable!  👍

Josleee:
After upgrading html5lib to 1.0b8, it works.
Thanks for the requirements file.

PR #1128, opened 2017-06-12
Changed the capital of Lithuania to Vilnius

thinline72:
Hi,
There is a typo about the capital of Lithuania.

Vilnius is the capital of Lithuania https://en.wikipedia.org/wiki/Vilnius
Ljubljana is the capital of Slovenia https://en.wikipedia.org/wiki/Ljubljana

### DIFF
--- a/website/docs/usage/word-vectors-similarities.jade
+++ b/website/docs/usage/word-vectors-similarities.jade
@@ -77,7 +77,7 @@ p
 
 +code.
     doc1 = nlp(u"Paris is the largest city in France.")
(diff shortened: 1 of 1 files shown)


honnibal:
Fuuuck I knew I should've double checked that :(

Thanks for pointing this out!

ines:
Ahh this is awkward! Can't believe we spent so much time playing with this example but missed that... I swear I actually knew this, haha.

Thanks for fixing the example **and** the similarities btw, this is great! 👍

PR #1157, opened 2017-06-27
Add basic Japanese Tokenizer Test

polm:
## Description

I added a basic test for the Japanese tokenizer. I tried to make it work similarly to other tokenizer tests, which required some changes to the way the base Japanese tokenizer was set up. I think it's an improvement since the tokenizer is re-used instead of being recreated every time now, but since no other language seems to work this way, let me know if something should be different.

## Types of changes

(sorry, not clear to me which of these adding tests counts as)

## Checklist:

### DIFF
--- a/spacy/ja/__init__.py
+++ b/spacy/ja/__init__.py
@@ -3,21 +3,39 @@ from __future__ import unicode_literals, print_function
 
 from os import path
 
-from ..language import Language
+from ..language import Language, BaseDefaults
+from ..tokenizer import Tokenizer
 from ..attrs import LANG
 from ..tokens import Doc
 
 from .language_data import *
 
-
-class Japanese(Language):
-    lang = 'ja'
-
-    def make_doc(self, text):
+class JapaneseTokenizer(object):
+    def __init__(self, cls, nlp=None):
+        self.vocab = nlp.vocab if nlp is not None else cls.create_vocab(nlp)
         try:
             from janome.tokenizer import Tokenizer
         except ImportError:
             raise ImportError("The Japanese tokenizer requires the Janome library: "
                               "https://github.com/mocobeta/janome")
-        words = [x.surface for x in Tokenizer().tokenize(text)]
+        self.tokenizer = Tokenizer()
+
+    def __call__(self, text):
+        words = [x.surface for x in self.tokenizer.tokenize(text)]
         return Doc(self.vocab, words=words, spaces=[False]*len(words))
+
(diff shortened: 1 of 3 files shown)


ines:
Thanks for adding the test and making that fix. Will check out the changes to the code (cc: @honnibal) and if it's all good, port it over to v2 as well.

Travis currently fails, but this is because the Japanese tokenizer requires the Janome library, which is not actually a dependency of spaCy itself (and we obviously want to keep it this way).

So in order to make Travis pass, here's my suggestion: pytest comes with an [`importorskip` method](https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency) to skip a test if a module can't be imported. So if we simply add that to the fixture, the test should only run if Janome is installed:

```python
@pytest.fixture
def ja_tokenizer():
    janome = pytest.importorskip("janome") # if this fails, test will be skipped
    return Japanese.Defaults.create_tokenizer()
```

polm:
Thanks! Added `importorskip` and parametrized the tests - let me know if there's anything else I should do.

ines:
Thanks a lot! 

And hmm, for some weird reason, Travis still fails... maybe my suggestion wasn't quite right (which is weird, because I remember using the same logic in other tests and it worked fine... and the fact that it's within a fixture shouldn't matter either, [see here](https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization-of-optional-implementations-imports)). 

Anyway, I'm sure we'll find a solution for this!

polm:
Sorry for the late reply on this - I found the issue.

If I run just the Japanese-specific tests in an environment without `janome`, they're skipped, as they should be. 

The issue was with the generic Tokenizer tests. I added Japanese to the `LANGUAGES` list in `conftest.py`, and creating the tokenizer there raises the exceptions that cause Travis to fail. If I remove Japanese from the `LANGUAGES` list all tests pass. I think this makes sense for externally implemented tokenizers, and I note Chinese also isn't in the list of `LANGUAGES`.

If Japanese _should_ be run through the generic Tokenizer tests, let me know and I'll figure out how to do that.

ines:
Ahh, you're right – no idea how I missed this. Thanks for fixing! 👍  Looks like Travis is green now, so I'll go ahead and merge.

I think for now, it doesn't make much sense to run the basic tokenizer tests on Japanese, as they're specifically written for spaCy's tokenizer. It includes a lot of special cases and rules (like, emoji etc.) and I don't expect Janome to handle things the same way.

PR #1161, opened 2017-06-28
French NUM_WORDS and ORDINAL_WORDS

AlexisEidelman:
## Description
Minor change maybe not even operative

Pretty easy to understand.

I'd be pleased to have @Gregory-Howard opinion ?

### DIFF
--- a/spacy/fr/stop_words.py
+++ b/spacy/fr/stop_words.py
@@ -86,3 +86,28 @@ votre vous vous-mêmes vu vé vôtre vôtres
 
 zut
 """.split())
+
+
+
+# Number words
+
+NUM_WORDS = set("""
+zero un deux trois quatre cinq six sept huit neuf dix
+onze douze treize quatorze quinze seize dix-sept dix-huit dix-neuf 
+vingt trente quanrante cinquante soixante septante quatre-vingt huitante nonante
+cent mille mil million milliard billion quadrillion quintillion
+sextillion septillion octillion nonillion decillion
+""".split())
+
+# Ordinal words
+
+ORDINAL_WORDS = set("""
+premier deuxième second troisième quatrième cinquième sixième septième huitième neuvième dixième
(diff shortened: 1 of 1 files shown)


ines:
Thanks! 👍

And yes, in v1.x, this data is not yet used – but we've started collecting it anyways to make it easier to add in v2 and allow people to use it to create their own custom flags.

In v2.0, individual languages can overwrite the lexical getters – for example, `LIKE_NUM`. [Here's how this is implemented in English](https://github.com/explosion/spaCy/blob/develop/spacy/lang/en/lex_attrs.py). For more info, see the [alpha docs](https://alpha.spacy.io/docs/usage/adding-languages#lex-attrs).

If I remember correctly, French numbers are mostly composed of the same number words and hyphenated, right? So the tokenizer should split them, and the `like_num` function could simply check the string against a list of number word components, like it's done [in English](https://github.com/explosion/spaCy/blob/develop/spacy/lang/en/lex_attrs.py). Or are there many special cases?

Gregory-Howard:
Oups didn't see that.
This is an example how number are in French : http://orthonet.sdv.fr/pages/lex_nombres.html
Absolutely horrible, I know. 
For the moment I don't have the time sorry. (I need to write my Internship report)

honnibal:
This seems safe to merge, right?

I get that this isn't the last word on French number processing, but the patch does seem better than before. Thanks for your help!

AlexisEidelman:
Found [this](http://www.miakinen.net/vrac/nombres)
I'm working in pythonizing it.

I'll be glad to have your opinion for a more general question. I'd like not only to know if a word is a number, I'd appreciate if in case it's look like a number to know to which number it is close to. Is spaCy the appropriate place to implement this ?

PolyglotOpenstreetmap:
If you can explain to me how to go about getting words next to each other but separated by spaces or dashes, I would like to attempt parsing such numbers (and determining the represented values).
I did it for Dutch, at least for the parts that are written together without spaces. That's not the complete number though.

PR #1171, opened 2017-07-03
Improve basic support for Danish

mollerhoj:
## Description
I've made some minor improvements to the support for Danish. I also noticed two strange stop words in the English list, so I removed them. I've trained a model for POS tagging and dependency parsing + some morphological information. Performance shown below.

```
# Performance of model
{
  "token_acc":100.0,
  "ents_p":0.0,
  "ents_r":0.0,
  "uas":81.4341736695,
  "tags_acc":92.3017107309,
  "ents_f":0.0,
  "las":76.7394957983
}
```
I have an issue with the pos tags - they are only available using `token.tag_`. `token.pos_` always returns `X`. I think this is because they include morphological features (Eg. `PRON__Case=Nom|Gender=Com|Number=Sing|Person=3|PronType=Prs`), thus the tag_map does not work. I guess this is because my JSON training data in the correct format. However, I can't figure out what the format should look like. @ines, could you please show an example of what the values of the JSON format should be?

## TODO
- [ ] Fix POS tags + Morphological features
- [ ] Supply lists of lemmas

## Types of changes

## Checklist:


lucaskjaero:
Native Danish speaker here. This looks like a good list.

I'd say that etc in the English list isn't wrong though. etc is short for etcetera, which means the same thing as osv.

mollerhoj:
@lucaskjaero - Thanks!

I'm questioning if `etc.` and `osv.` are considered stop words? (I guess that's a matter of taste). Anyways, I guess I won't remove `etc.` then - and add `osv.` to the list of Danish stop words ;-)

@ines - Would it be possible for you to show an example of what the values in the JSON training data should look like?

honnibal:
@mollerhoj Apologies for missing this question. I've attached an example file, which is taken from the English Wall Street Journal portion of the Penn Treebank. It shows dependencies and part-of-speech.

I would suggest using the `spacy convert` tool to create these json files. I have to say I don't find my own format very convenient :(. It's like this because I wanted to support all the annotations, including raw text and paragraph/sentence structure.

[spacy-json-example.txt](https://github.com/explosion/spaCy/files/1410204/spacy-json-example.txt)

ines:
I also just resolved the conflict that resulted from a small change I've made to the language data `__init__.py` files in the meantime. So once the CI checks go green, this PR would technically be ready to merge!

mollerhoj:
Thanks guys. I wanna include a list of lemmas as well (I have access to some very good resources for this, that have previously been hidden in the strange maze they call academia.). But I'll just make a new PR. I've removed the WIP tag from the title. Yay for Denmark ;-)

ines:
Yay, merging! 🎉

Btw, also just updated the JSON training format docs with more details and another example: https://alpha.spacy.io/api/annotation#json-input

PR #1177, opened 2017-07-05
Dutch NUM_WORDS and ORDINAL_WORDS

swierh:
## Description
I've added the Dutch numbers and ordinals.
To do this I've also renamed the `stop_words.py` to `word_sets.py`, and fixed the import in `language_data.py`, these changes reflect the structure for the English word sets.

## Types of changes

## Checklist:


ines:
Thanks! 👍

Also looking forward to porting this over to the develop branch and finally using this data properly in the new version! Here are some details on how the `NUM_WORDS` etc. will be used in spaCy v2.0: https://github.com/explosion/spaCy/pull/1161#issuecomment-312423088 In short, there will be a `like_num` getter that takes a string and returns a boolean.

Dutch number words are similar to German number words (as in, "five hundred and thirty nine thousand" is one token), right?

swierh:
@ines Yes, from what I remember from high school it's identical to German.
There is one minor detail, in Dutch words that contain two different vowel sounds after eachother, sometimes the second vowel gets a diaeresis (looks identical to umlaut). This is an issue with numbers that have as single digits 2 or 3:
tweeëntwintig (two and twenty), drieëntwintig (three and twenty), honderdtweeënveertig (hundred two and forty).

ines:
Thanks! This is good to know – so we'll have to find a nice solution for that when we port this over to v2. 

In English, this is easier, because we can simply take the token string and check whether it's part of a list of number words.  In cases like this Dutch, German etc., we'll have to take a different approach (because we can't include a list of _all_ number words, obviously). So instead, we could include a list of number word "fragments" and check whether the string contains the number fragment – and a list of strings for the numbers from 1 to 13 or something, which we also check against (otherwise, it'd return `True` for all words containing "twee" or whatever, which would likely be bad). This won't always be _perfect_ – but it's also why the attribute is called **like**_num.

PR #1186, opened 2017-07-12
.../cli/#foo is 404

lgenerknol:
https://spacy.io/docs/usage/cli/#package is a 404.  
Changed to https://spacy.io/docs/usage/cli#package 

Definitely a larger fix possible to deal with trailing slashes

## Types of changes

## Checklist:


ines:
Thanks for the fix! 👍

I'll do a quick search and replace to make sure there are no other links with `/#`.  Maybe I'll add a rewriting rule on the server as well to handle those cases – but I think it should be find without, as long as our site doesn't include any broken links. (If users click on an anchor and copy-paste the link, it'll be correct anyways, and I always prefer avoiding URL rewriting if possible.)

PR #1220, opened 2017-07-25
Install models to users homedirectory instead of globally

leezu:
Use pip install --user. It is not advisable to install the packages globally as this not only requires root access on the machine, [but also messes with files managed by the system package manager](https://wiki.gentoo.org/wiki/Pip#Package_installation) (apt-get, emerge, ...)

### DIFF
--- a/spacy/cli/download.py
+++ b/spacy/cli/download.py
@@ -64,7 +64,7 @@ def download_model(filename):
     util.print_msg("Downloading {f}".format(f=filename))
     download_url = about.__download_url__ + '/' + filename
     subprocess.call([sys.executable, '-m',
-        'pip', 'install', '--no-cache-dir', download_url],
+        'pip', 'install', '--user', '--no-cache-dir', download_url],
         env=os.environ.copy())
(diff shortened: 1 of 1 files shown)


honnibal:
The most common workflow is to install to virtualenv -- specifying `--user` here breaks that, right?

The right thing would be to figure out whether spaCy was installed with `--user`, and if so, install the model that way too. I haven't been able to figure out how to do that. Any ideas?

leezu:
Hey Matthew, I was not aware that specifying `--user` breaks the workflow im virtualenv. Your suggestion seems good to me, though until you figure it out, what about introducing a `--user` command line switch for the spaCy module? I will adapt the pull request for that if it's in order with you.

follnoob:
Maby you can add a flag '--user' to the download command like

    python -m spacy download en --user

so you can use your appropriated way of installation.

honnibal:
I think @follnoob 's suggestion is good. I'm going to close this PR and open an enhancement ticket instead.

Issue #1214, opened 2017-07-23
Japanese POS Tag Support


polm:
Would it be OK to add support for Japanese POS tags to spaCy? 

Janome, the library used for tokenization, already provides POS tags since they're generated as part of tokenization. That means this wouldn't require any extra dependencies or processing time over what's already being done when tokenizing Japanese.

If I wanted to implement this, is there anything in particular I should read first or keep in mind while working on it?

honnibal:
This sounds very nice! A pull request would be very welcome.

The main thing to keep in mind is that you'll need to make sure that there are entries in the `Japanese.Defaults.tag_map` dictionary, to allow the `token.pos_` attribute to be computed. an example entry:
    
    from spacy.attrs import POS
    from spacy.symbols import NOUN

    `"NN": {POS: NOUN}

polm:
Thanks, that sounds good.

I started working on a tag map but ran into an issue - turns out Japanese Universal Dependency tags aren't all direct mappings from raw tags, some require looking at the way a particular token is used in the sentence. This is documented in the notes at the bottom [here](http://universaldependencies.org/ja/overview/morphology.html#morphology).

These rules are also kind of unclear, and I haven't able to find any code that actually handles tag conversion, so I'm going to see if I can get some clarification from the Universal Dependencies folks.

polm:
I'm still waiting on some information from the UD team on the mapping, but I've run into an issue - a simple dictionary like `tag_map` can't handle conversion from raw tags to Japanese UD tags. 

A paper introducing the Japanese UD project in English [here](http://www.ar.media.kyoto-u.ac.jp/publications/tanaka-LREC16.pdf) touches on the issue; to give an example:

![Same Unidic tag, different UD tag](https://user-images.githubusercontent.com/286278/28833025-20618cc8-7719-11e7-825a-21d341a3a057.png)

The "report" in the two sentences would have the same raw tag, but a different UD tag. (Another issue is resolving the UD tag would require a full dependency parse...)

Is there a way to deal with non-direct mappings like this in spaCy? Looking at existing tag maps I notice Spanish seems to append a lot of raw features into the source tag, but that won't necessarily resolve the ambiguity in this case.

polm:
I'm closing this since the related changes were merged in #2036 and #1246. Thanks!

PR #1246, opened 2017-08-07
[wip] Sample implementation of Japanese Tagger (ref #1214)

polm:
This is far from complete but it should be enough to check some things related to #1214.

1. Mecab transition. Janome doesn't support Unidic, only IPAdic, but UD
tag mappings are based on Unidic. This switches out Mecab for Janome to
get around that.

2. Raw tag extension. A simple tag map can't meet the specifications for
UD tag mappings, so this adds an extra field to ambiguous cases. For
this demo it just deals with the simplest case, which only needs to look
at the literal token. (In reality it may be necessary to look at the
whole sentence, but that's another issue.) Is this the right way to deal
with ambiguous mappings?

3. General code structure. Seems nobody else has implemented a custom
Tagger yet, so still not sure this is the correct way to pass the
vocabulary around, for example.

Any feedback would be greatly appreciated. -POLM

## Types of changes

- [x] **Breaking change** (fix or feature causing change to spaCy's existing functionality) (changing to Unidic will change tokenizer output)

## Checklist:

- [x] I have added tests to cover my changes. (Just a minimal demo)
- [ ] All new and existing tests passed. (Current JA Tokenizer tests fail due to transition from IPAdic to Unidic)


honnibal:
This will be a great addition. Thanks!

> 1. Mecab transition. Janome doesn't support Unidic, only IPAdic, but UD
tag mappings are based on Unidic. This switches out Mecab for Janome to
get around that.

I need to do some reading before I can comment on this. Will get back to you.

> 2. Raw tag extension. A simple tag map can't meet the specifications for
UD tag mappings, so this adds an extra field to ambiguous cases. For
this demo it just deals with the simplest case, which only needs to look
at the literal token. (In reality it may be necessary to look at the
whole sentence, but that's another issue.) Is this the right way to deal
with ambiguous mappings?

I'll explain the best-case solution, and then we can think about the best trade-off given the constraints.

Normally we'd want to force the UD mapping to be deterministic given the `TAG`. The `TAG` can be any arbitrary string, so we can always make this work, even if we have to decorate the `TAG` with the UD POS value. Even if we ended up with a large tagset, the tagger can still learn the task fairly easily.

The problem comes when we only have the tool's output, and we don't have the training data. In this case i think it's fine that the mapping is a bit lossy. We would just pick the most common UD tag, and accept that it'll be wrong in some cases. For common words we can also use the `morphs.py` mapping table, which is keyed by both word and TAG.

> General code structure. Seems nobody else has implemented a custom
Tagger yet, so still not sure this is the correct way to pass the
vocabulary around, for example.

Looks good :). A tiny detail is that it'd be better to move the try/except around the import into a utility function. I still need to spend a little more time with the code though, so there may be something I've missed.

Thanks again!
Matt

polm:
Thanks for the feedback! Good to know I'm moving in the right direction.

> The problem comes when we only have the tool's output, and we don't have the training data. In this case i think it's fine that the mapping is a bit lossy. We would just pick the most common UD tag, and accept that it'll be wrong in some cases.

This sounds reasonable to me. Unfortunately some of the ambiguous tags are extremely common - for example, basically all "to be" verbs are always potentially either `VERB` or `AUX` - so just picking one will diverge substantially from a proper UD style conversion. That said, what we can do is make a simple mapping that is technically not in line with UD but is at least unsurprising. Unidic POS tags are structured like `(main category),(secondary category),(etc),(etc)`, and we can use the "main category" as a default. Especially given the UD conversion rules aren't finalized this should be a good start. This is actually what the `TAG_MAP` I included already does for tags if an annotation like `NOUN` or `VERB` is not appended.

> For common words we can also use the morphs.py mapping table, which is keyed by both word and TAG.

This sounds like it'd be useful for 連体詞 - the whole category has ~100 terms, and a subset of those need a different UD tag from the rest. I couldn't find a `morphs.py` file though, is the map you mention something in the `Morphology` class somewhere?

About the Mecab transition: if there's any information I can provide let me know. While I understand adding even an optional dependency on an external C++ program is undesirable, Mecab is basically the standard tool for tokenizing Japanese - Janome uses a Mecab dictionary, when the national agency [NINJAL](http://www.ninjal.ac.jp/) [releases Unidic](https://ja.osdn.net/projects/unidic/) it's prepared specifically for use with Mecab, and basically all other tools compare themselves to Mecab. One nice bonus of switching from Janome is that using Mecab is roughly 10 times faster (mentioned in the [Janome JP FAQ](http://mocobeta.github.io/janome/#id13) but not the English one for some reason...).

polm:
Hey, could I get some feedback on this?

honnibal:
@polm Looks great! Certainly there's nothing here that'll cause problems outside the `spacy.jp` package, so I'm ready to merge when you are. Sorry for the delay -- I thought it was still WIP.

> About the Mecab transition: if there's any information I can provide let me know. While I understand adding even an optional dependency on an external C++ program is undesirable, Mecab is basically the standard tool for tokenizing Japanese

I think at least having the option would be good. Would it be possible to make a container Python package, that shipped Mecab inside it? This could be released as both a source package and a wheel. That way the dependency can be specified via pip, and it's just like another Python dependency. If Mecab itself has system dependencies, then there's not much point. But if it's otherwise self-contained, this can work pretty well.

ines:
Thanks for the great work on this!  👍 
From looking at the code, this should be pretty easy to port over to `develop` for spaCy v2.0, right? If you want to have a go at it, let me know – otherwise, we're happy to take care of it as well.

polm:
Thanks! Sorry I wasn't more clear about the status - I think it's good to merge at this point.

The UD mapping is not perfect, but since the Japanese UD rules aren't finalized I think it's as good as possible for now. I'll update it as the rules are clarified.

Could you clarify what the "`morphs.py` mapping table" is? I never did find that... 

> Would it be possible to make a container Python package, that shipped Mecab inside it?

That sounds like a good idea. Mecab only depends on libiconv, so that could be included if necessary. I haven't made a container package like this before so I'll have to figure out how that's done - if you have any examples you can point me at that'd be a big help.

Regarding spaCy v2, I'll try porting it over - it'll be a good chance for me to get up to speed on v2.

Thanks for all your help!

ines:
> Could you clarify what the "`morphs.py` mapping table" is? I never did find that...

I think @honnibal was referring to the `morph_rules.py`, which can be an additional component of the language data directory. Previously, this hasn't been documented very well – but there's now [a section on it in the v2.0 alpha docs](https://alpha.spacy.io/usage/adding-languages#morph-rules).

honnibal:
@polm Two examples for container packages, with different strategies:

1) https://github.com/explosion/murmurhash

This ships the .cpp files, and specifies them in the setup.py as source files. setuptools then triggers the compilation.

2) https://github.com/honnibal/cython-blis

I'm not sure this example works 100% correctly. In this example, the source files are included as package data, that setuptools doesn't know anything about. The setup.py then calls out to the shell to use make the library, using the `./configure` and `Makefile` provided.

Solution 2 is obviously a worse end-state than solution 1. On the other hand, build systems can be pretty complicated. I have a tonne of trouble even reading people's configure scripts and Makefiles --- let alone recreating the logic in setuptools, which doesn't exactly have great support for this stuff. So if the C/C++ project is non-trivial, calling out to the project's build-system might be a good compromise.

I totally understand if you don't want to spend time doing this. Having a system dependency isn't even as bad as it used to be, since everyone's using Docker etc anyway.

polm:
Thanks for the merge and the package advice! 

It's not working yet, but I've got code [here](https://github.com/polm/mecab-packed) and there was some interest in a bundled mecab+unidic, so I'll see if I can figure it out. If it gets working I'll create another PR here. 

Thanks again!

Issue #1235, opened 2017-07-29
Unexpected processing of some text ending with 4g, 2g, ...


slremy:
From my understanding, the code below should produce a Doc object of length 15, however instead it produces an object of length 16 since one term "e2g" is split into two.

This doesn't appear to be the appropriate behavior.
For reference, the length of the testword list doesn't appear to matter. It seems to be the case the many strings that ends in G and have a number preceding that G (2g, V2G) may also produce this behavior.

> regex =>     (.*)\d+g

```
import spacy
nlp = spacy.load('en_core_web_md') 
testwords=u'utilizing utm utmost utms utterly uwb ux v1 v2 e2g v2i v2v v3 vacant vacating'
doc=nlp(testwords);print len(doc), len(testwords.split(" "))
```
16 15
```
zip(doc, testwords.split(" "))
```
[(utilizing, u'utilizing'),
 (utm, u'utm'),
 (utmost, u'utmost'),
 (utms, u'utms'),
 (utterly, u'utterly'),
 (uwb, u'uwb'),
 (ux, u'ux'),
 (v1, u'v1'),
 (v2, u'v2'),
 (e2, u'e2g'),
 (g, u'v2i'),
 (v2i, u'v2v'),
 (v2v, u'v3'),
 (v3, u'vacant'),
 (vacant, u'vacating')]

* **Python version:** 2.7.13
* **Platform:** Darwin-16.7.0-x86_64-i386-64bit
* **spaCy version:** 1.8.2
* **Installed models:** en, en_core_web_md, en_core_web_sm

slremy:
Digging a little further.. It's probably the case that these terms are being treated as if they are a number that is in units of grams.

[spacy/language_data/punctuation.py](https://github.com/explosion/spaCy/blob/df64e8dbb32c2e5ea9abb98b162f07b5a3d0f521/spacy/language_data/punctuation.py#L9)

Accordingly, they are being parsed as if they were like the term 100GB, and split into 100, and GB.

This is just a hypothesis.. but I hope that I'm wrong!

ines:
Thanks for the report and analysis! this is interesting – the "g" should definitely only be split off if it's preceded by _only numbers_ (in which case it's fair to assume that it most commonly stands for grams or gigabytes). 

So this might indeed be an issue with the suffix rules and should be easy to fix with some fiddling and a simple regression test. Will look into this! (Also tagging this issue with `help wanted (easy)` in case someone else wants to investigate this).

[...]

PR #1265, opened 2017-08-16
Better token splitting regexes for numbers preceding unit suffixes (Issue #1235)

pmccarter:
Thought I'd help out with an easy issue, so worked on addressing issue #1235 to better split numeric quantities with joined unit suffix.

## Description
Replaced simple tokenizer regex with 2 regexes to ensure splitting only happens when prefixed by numeric quantities, with possible decimal and thousands place comma separators. Added a unit test to show new splitting functionality.

## Types of changes

## Checklist:

### DIFF
--- a/spacy/language_data/punctuation.py
+++ b/spacy/language_data/punctuation.py
@@ -84,7 +84,8 @@ TOKENIZER_SUFFIXES = (
         r'(?<=[0-9])\+',
         r'(?<=°[FfCcKk])\.',
         r'(?<=[0-9])(?:{c})'.format(c=CURRENCY),
-        r'(?<=[0-9])(?:{u})'.format(u=UNITS),
+        r'(?<=^[0-9]+(\.[0-9]*)?)(?:{u})'.format(u=UNITS),
+        r'(?<=^[0-9]{{1,3}}(,[0-9]{{3}})*(\.[0-9]*)?)(?:{u})'.format(u=UNITS),
         r'(?<=[0-9{al}{p}(?:{q})])\.'.format(al=ALPHA_LOWER, p=r'%²\-\)\]\+', q=QUOTES),
         r'(?<=[{au}][{au}])\.'.format(au=ALPHA_UPPER),
         "'s", "'S", "’s", "’S"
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks for helping out!

I have to say, a quick look at the code left me a little puzzled as to how this works? If I'm not mistaken, you have a variable-length pattern within a look-behind. I thought that wasn't allowed?

pmccarter:
It's apparently supported by the nonstandard regex module used by spacy.

[https://stackoverflow.com/questions/24987403/variable-width-lookbehind-issue-in-python](url)

honnibal:
Hm. Interesting!

I'm pretty suspicious of the performance implications of this though. We're already seeing some problems from the URL regex in weird edge cases.

Is there another way to phrase the pattern?

pmccarter:
I'd assume noncapturing lookbehind would be cheap for non matching tokens, but I can do some analysis. Do you have any rule of thumb for acceptable regex performance?

honnibal:
Well...The problem is really that look-behinds and other regex complexities can lead to non-linear worst-case complexity. This is hard to benchmark, as the behaviour will be implementation dependent, and the benchmark may not trigger the patthological cases.

This already bit me recently: spaCy died after processing ~4b tokens of text on a large job, with the `regex` module asserting "too many look behinds". I don't know what token caused this --- it could've been a token that was hundreds of characters long, for instance. If so we can add some length constraints. But obviously, it's unacceptable that some string input crashes the library.

I think I'd like to have the tokenizer expressions simplified further, probably removing the look-arounds in the prefix and suffix expressions. So we'll need to find another way of phrasing the regex in this patch, so that it's easier to reason about the regex's performance.

pmccarter:
If you tested the reverse of the tokens, couldn't you get reliable linear time match regexes?  I could also see trimming the prefixes/suffixes that are tested to a constant upper limit to avoid full scans on bad cases.

honnibal:
Not 100% sure I'm reading your suggestion correctly -- but, if you mean to add some post-processing logic after the matches are identified, we'd like to avoid hard-coding things. The match logic has to happen inside the expresion, to ensure the tokenizer can be customized.

pmccarter:
No I meant preprocessing the unsplit whitespace separated strings so the prefix and suffix matchers only access a slice of the full string, so massive tokens don't kill it. But that would mean the single joined match regexes would have to be split up.

honnibal:
@pmccarter That's a good idea! If we have a maximum prefix or suffix length, then we can feed only a slice of the token into the regex. This would make sure the regex performance was predictable.

Issue #1271, opened 2017-08-17
spacy.load fails on python < 3.6 (v2)


danohuiginn:
spacy.load fails with python 3.5:

```
In [2]: en = spacy.load('en')
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-a732401aeb32> in <module>()
----> 1 en = spacy.load('en')

/src/spaCy/spacy/__init__.py in load(name, **overrides)
     11 def load(name, **overrides):
     12     name = resolve_load_name(name, **overrides)
---> 13     return util.load_model(name, **overrides)
     14
     15

/src/spaCy/spacy/util.py in load_model(name, **overrides)
    101     if isinstance(name, basestring_):
    102         if name in set([d.name for d in data_path.iterdir()]): # in data dir / shortcut
--> 103             return load_model_from_link(name, **overrides)
    104         if is_package(name): # installed as package
    105             return load_model_from_package(name, **overrides)

/src/spaCy/spacy/util.py in load_model_from_link(name, **overrides)
    114     """Load a model from a shortcut link, or directory in spaCy data path."""
    115     init_file = get_data_path() / name / '__init__.py'
--> 116     spec = importlib.util.spec_from_file_location(name, init_file)
    117     try:
    118         cls = importlib.util.module_from_spec(spec)

/usr/lib/python3.5/importlib/_bootstrap_external.py in spec_from_file_location(name, location, loader, submodule_search_locations)

AttributeError: 'PosixPath' object has no attribute 'endswith'

```
I believe this is because the loader is relying on importlib.util.spec_from_file_location accepting a pathlib.Path arugument, which is [only supported from python 3.6](https://docs.python.org/3/library/importlib.html#importlib.util.spec_from_file_location). 

This change fixes the problem; I've not looked to see if similar issues occur elsewhere:

```
diff --git a/spacy/util.py b/spacy/util.py
index ccb81fbe..d83fe341 100644
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -113,7 +113,7 @@ def load_model(name, **overrides):
 def load_model_from_link(name, **overrides):
     """Load a model from a shortcut link, or directory in spaCy data path."""
     init_file = get_data_path() / name / '__init__.py'
-    spec = importlib.util.spec_from_file_location(name, init_file)
+    spec = importlib.util.spec_from_file_location(name, str(init_file))
     try:
         cls = importlib.util.module_from_spec(spec)
     except AttributeError:

[...]

PR #1272, opened 2017-08-17
Allow loading with python < 3.6

danohuiginn:
Don't rely on recent python features to load models

Fixes Issue #1271

## Types of changes

- [x ] **Bug fix** (non-breaking change fixing an issue)

## Checklist:


ines:
Thanks for spotting and fixing this so quickly! 👍  

(I was surprised this hadn't come up before – but I think the current solution does work on Python 2 and Python 3.6, and since not so many people are using 3.5, so we missed this. Sorry about that. I think we might not have 3.5 covered in our test either, so I'll check Travis and add it if necessary.)

PR #1276, opened 2017-08-19
Ported model cli from v1

oroszgy:
Ported Model CLI for v2

## Description
Took the model cli from the master branch, and adapted it for the v2 workflow. (Probably needs to be reviewed)

## Types of changes

- [X] **New feature** (non-breaking change adding functionality to spaCy)

## Checklist:

### DIFF
--- a/spacy/__main__.py
+++ b/spacy/__main__.py
@@ -3,15 +3,21 @@ from __future__ import print_function
 # NB! This breaks in plac on Python 2!!
 #from __future__ import unicode_literals
 
-
 if __name__ == '__main__':
     import plac
     import sys
-    from spacy.cli import download, link, info, package, train, convert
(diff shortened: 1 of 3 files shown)


ines:
Thanks!  👍 If I remember correctly, we originally didn't include it because we thought it'd require too many changes and that it'd be redundant now that the serialisation is fixed. But of course, the frequencies etc. are still relevant. So thanks for taking this on!

Now that I look at it again, maybe we should consider renaming the command? `spacy model` is a bit vague and possibly confusing to users. What do you think? I don't immediately have an idea, so we're open for suggestions!

oroszgy:
As this command assembles the vocabulary from multiple sources, I would find a name related to "vocabulary"

BTW, just realized that vectors were not added, fixing this and will do another PR. 
Another question: where should I store the `oov_prob` value in the vocabulary?

Issue #1292, opened 2017-08-30
Customized tokenizer reports 'AssertionError' with specific infixes


Vimos:
When I trying to add '—\-' to the infix_re, whole re is `'[\[\]!&:,()\*—\-]'`, for a custom tokenizer, I got the following error. 
```
Traceback (most recent call last):
  File "/home/vimos/Public/github/NLP/keras-nli/examples/squad/sequence_pointer/preprocessing.py", line 15, in <module>
    o.preprocessing()
  File "/home/vimos/Public/github/NLP/keras-nli/kerasnli/preprocessing/squad.py", line 223, in preprocessing
    self.to_sequence()
  File "/home/vimos/Public/github/NLP/keras-nli/kerasnli/preprocessing/squad.py", line 119, in to_sequence
    context = [word.text for word in nlp(p['context'])]
  File "/data/home/vimos/Public/git/github/NLP/spaCy/spacy/language.py", line 320, in __call__
    doc = self.make_doc(text)
  File "/home/vimos/Public/github/NLP/keras-nli/kerasnli/preprocessing/squad.py", line 36, in <lambda>
    return lambda text: tokenizer(text)
  File "tokenizer.pyx", line 152, in spacy.tokenizer.Tokenizer.__call__
  File "tokenizer.pyx", line 217, in spacy.tokenizer.Tokenizer._tokenize
  File "tokenizer.pyx", line 315, in spacy.tokenizer.Tokenizer._attach_tokens
  File "doc.pyx", line 469, in spacy.tokens.doc.Doc.push_back
AssertionError
```
I am using the head of spacy, but same error on an earlier version. Only '—\-' is causing this problem. 

After some hacking, I find a suspect for the error, trailing '-' in the string
```
# coding: utf-8
from __future__ import unicode_literals

from ... import load
from ...tokenizer import Tokenizer
from ... import util

import pytest


def test_customized_tokenizer_handles_infixes():
    def custom_tokenizer(nlp_model):
        prefix_re = util.compile_prefix_regex(nlp_model.Defaults.prefixes)
        suffix_re = util.compile_suffix_regex(nlp_model.Defaults.suffixes)
        custom_infixes = ['\.\.\.+',
                          '(?<=[0-9])-(?=[0-9])',
                          # '(?<=[0-9]+),(?=[0-9]+)',
                          '[0-9]+(,[0-9]+)+',
                          u'[\[\]!&:,()\*—–\/-]']

[...]

PR #1294, opened 2017-09-01
Fix issue #1292 and add test case for the Assertion Error

Vimos:
## Description

Added a check to the span sub string before push back to the tokens, to avoid a null sub string which may cause Assertion Error.
Tested on python 2.7 with a Ubuntu16.04.

## Types of changes

## Checklist:


honnibal:
Looks good! I tweaked the test slightly to remove a layer of indirection. Hopefully I didn't mess it up.

PR #1389, opened 2017-10-05
Lemmatizer obey exceptions

### LINKED ISSUE #1387: Lemmatizer exceptions: `lemma_strings` get sorted anyway
Lemmatization exceptions have been working inconsistently, even when they are added directly in `corpora/en/wordnet/dict/verb.exc`

The minimal test case of `coping` at https://github.com/explosion/spaCy/issues/389 reveals that, at 
https://github.com/explosion/spaCy/blob/master/spacy/lemmatizer.py#L94, the lemmatizer's list of  potential forms (where the thing in lemmatizer.exceptions is item [0]) is cast to a `set` (and so loses ordering); then in https://github.com/explosion/spaCy/blob/master/spacy/morphology.pyx#L149, that `set` gets sorted. So lemmatizer exceptions only work if they also come first alphabetically!

I've implemented the fix for this, but I need this issue in order to submit the PR!

One question, though, for @honnibal: this can be fixed either as I did it locally -- return the whole list, with exceptions and then anything that comes back from the lemmatizer's `rules` -- or skip the rules altogether if we used an exception. I think it's more useful downstream if we keep all lemma candidates, even if we're not using them in the default pipeline. But it also seems only destructive to do `sorted(set())` on them!

mdcclv:
## Description
Fix #1387 in the simplest way: don't apply `rules` if we got a lemma from `exc`eptions.

First commit, e81a608, is the regression test; next, ffb50d2, fixes it; in the third commit I copied the regression test to the tests in `tagger/test_lemmatizer.py` but I'm not sure if that's desirable. I thought it might be nice to have a stub there for other tests of exceptions, but at the moment it's purely redundant.

## Types of changes

## Checklist:

### DIFF
--- a/spacy/lemmatizer.py
+++ b/spacy/lemmatizer.py
@@ -78,15 +78,16 @@ def lemmatize(string, index, exceptions, rules):
     #    forms.append(string)
     forms.extend(exceptions.get(string, []))
     oov_forms = []
-    for old, new in rules:
-        if string.endswith(old):
-            form = string[:len(string) - len(old)] + new
-            if not form:
-                pass
-            elif form in index or not form.isalpha():
-                forms.append(form)
-            else:
-                oov_forms.append(form)
+    if not forms:
+        for old, new in rules:
+            if string.endswith(old):
+                form = string[:len(string) - len(old)] + new
+                if not form:
+                    pass
+                elif form in index or not form.isalpha():
+                    forms.append(form)
+                else:
+                    oov_forms.append(form)
     if not forms:
         forms.extend(oov_forms)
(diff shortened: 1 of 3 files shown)


honnibal:
Thanks for this! I think the same patch should apply cleanly to v2 as well?

tarskiandhutch:
Hello! Am I correct in interpreting that the fix giving lemmatizer exceptions precedence over the built-in rules has since been undone?

If so, what was the reason for this, and is there any way to ensure now that the exceptions one has loaded will not be ignored?

Thanks!

Issue #235, opened 2016-01-26
Forcing sentence segmentation for newlines


scottyli:
Is there a way to force sentence segmentation when a newline `\n` character is found?

For example,

``` md
Hey Honnibal,

This is a great library for 2 reasons:
 - It's fast
 - It's accurate

```

This is parsed as 1 sentence using `nlp(text)`. However, it'd be great if it was parsed as 4 sentences, because of the newline. E-mail data tends to be in this format.

I've searched the docs but couldn't find anything. Is there a workaround?

Thanks!

Nogreedy:
How can I contact you about TipperCoin/TwitterCoin (it doesn't work anymore and my Bitcoins are stuck) ?

wbwseeker:
If you are sure that newlines will always coincide with sentence boundaries, you can just split your documents at newlines and feed them to spaCy piece by piece.

honnibal:
I think it's weird at the moment that there's really no way to set the sentence boundaries yourself. If we had some way to do that, then users could set the attribute before parsing. Then we just have to make sure the transition system respects the existing data in the token.

The tricky thing is that writing to `token.sent_start` might leave the parse in an invalid state, and I'm not sure I like the idea of quietly fixing up a bunch of other words' `.dep`, `.head` etc attributes because you wrote to one token's `sent_start`. A `doc.break_sent(i)` method is probably more sensible. On the other hand, if `doc.is_parsed == False`, then writing the sentence boundary should be unproblematic. 

We should also have a `SENT_START` attribute. It's pretty weird that it's missing --- it means there's no way to export the sentence boundaries into the array at the moment. We would have to decide how to handle it in `doc.from_array` though.

So, here's my proposal. It's designed for now not to break anything, and to be better than the current status quo:

* Add a `SENT_START` attribute.
* Raise error if both `SENT_START` and `HEAD` are being set at the same time in `doc.from_array`
* Allow writing to `token.sent_start`. Raise if `self.doc.is_parsed == True`.
* Debate how to relax the constraint that you can only break *after* parsing.
* Consider changing the features in the tagger, so that the `N1` and `N2` token features take `EOL` values when they hit a sentence boundary. This will allow users to make sure that the tagger behaves just as if the string ends there.

[...]

PR #1400, opened 2017-10-08
💫 Force parser to respect preset sentence boundaries

honnibal:
This patch allows use of separate sentence segmentation and parsing components. Previously this didn't work correctly, because you could add sentence boundaries, but they didn't properly constrain the parser.

The patch fixes the problem by introducing ternary value semantics for the `token.sent_start` attribute. Previously this attribute took two values in `{True, False}`. It now takes values in `{True, False, None}`. In the C data, the mapping is `{1: True, 0: None, -1: False}`.

When `token.sent_start` is set to `None` (default), the parser can predict a sentence starting with that token. If you set it to `True`, the parser *must* predict a sentence on that token. `False` prevents a sentence.

So let's say you get sentence boundaries from some other function you have, that takes text and returns `(start_char, end_char)` tuples, that give the sentence spans. With the new pipeline management API, you'll be able to do the following:

```python

from somewhere_else import sentence_segmenter

def set_sent_starts(doc):
    for sent_start_char, sent_end_char in sentence_segmenter(doc.text):
        sent = doc.char_span(sent_start_char, sent_end_char)
        sent[0].sent_start = True
        for token in sent[1:]:
            token.sent_start = False
    return doc

nlp = spacy.load('en')
nlp.add_pipe(set_sent_starts, name='sentence_segmenter', before='parser')
```

As a small extra, I've also included an addition `Span` method, `span.as_doc()`. This has been suggested elsewhere, and it provides a useful alternative for some situations. The `span.as_doc()` method creates a `Doc` object with the same tokens as the span, **without copying the token data**. This means the new `Doc` object acts as a view, so when you write to it, your changes are reflected in the original `Doc`. This way, if you have a component that works over `Doc` objects, you can easily convert it to process sentences independently. We should probably have a wrapper to handle the most general transformation. Something like this:

```python

def doc_to_sent_adaptor(component):
    def doc2sent_inner(doc):
        for sent in doc.sents:
            component(sent.as_doc())
        return doc
    return doc2sent_inner
```

You would be able to wrap any component that has in-place semantics this way.

Re #235  #363 #1105 

CC @christian-storm        

## Types of changes

## Checklist:

### DIFF
--- a/spacy/structs.pxd
+++ b/spacy/structs.pxd
@@ -61,13 +61,13 @@ cdef struct TokenC:
     attr_t sense
     int head
     attr_t dep
-    bint sent_start
 
     uint32_t l_kids
     uint32_t r_kids
     uint32_t l_edge
     uint32_t r_edge
 
+    int sent_start
     int ent_iob
     attr_t ent_type # TODO: Is there a better way to do this? Multiple sources of truth..
     hash_t ent_id
--- a/spacy/syntax/_state.pxd
+++ b/spacy/syntax/_state.pxd
@@ -307,6 +307,8 @@ cdef cppclass StateC:
             this._stack[this._s_i] = this.B(0)
         this._s_i += 1
         this._b_i += 1
+        if this.B_(0).sent_start == 1:
+            this.set_break(this.B(0))
         if this._b_i > this._break:
             this._break = -1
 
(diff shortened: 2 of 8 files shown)


huajunzeng:
Hi, thanks a lot since this is a very desired feature. Is this only possible in v2, or also can be done in 1.x?

PR #1414, opened 2017-10-12
Adding Russian language support

yuukos:
## Types of changes

## Checklist:


yuukos:
Thank you for fast feedback and suggestions.
I'll fix it and make new pull request. And I can fill contributor agreement if it's necessary. 
I'm going to start porting russian language support to v2.x mb next month, because I do not have enough time now.

ines:
@yuukos Thanks and no worries! You don't have to make a new pull request just to add updates and changes by the way – you can just keep pushing to your branch, and it'll be updated over here.

yuukos:
oh... I got it. Thanks again ^^

yuukos:
Hi. All tests passed finally. And what should I do anything else now? (except for filling contributor agreement)

ines:
Yay! Very excited about this contribution – Russian has been requested a lot, so I hope this will make it easier for others to contribute and work towards full, built-in support!

If you want, you can add one or two simple tests  for the Russian tokenizer ([see here](https://github.com/explosion/spaCy/tree/master/spacy/tests) for examples and testing docs), so we can always ensure Russian works as expected. But those can also be added later 😊

yuukos:
I'm glad to see that my work can be useful for someone. I'm going to tuned it later and adapt for v2.x. 
And I have one question. I found two things: file with contributors and directory with contributor agreements. I should add myself into the file and add agreement into the directory. Is it correct?

ines:
@yuukos Adding the agreement file [to this directory](https://github.com/explosion/spaCy/tree/master/.github/contributors) is the most important thing. Once this is done, I'm happy to add you to the `CONTRIBUTORS.md` as well.

The agreement is just a formality to make sure that we can use your contribution in spaCy (we're now trying to remind everyone when they open a PR).

yuukos:
Hello. 
I have a question. Can I use Russian language from spacy just after `pip install -U spacy`? Or I have to do something else to use Russian language?

ines:
@yuukos No, you'll have to wait until the next release – we're currently busy with getting v2.0.0 ready (so there are currently more frequent releases on `spacy-nightly`), but there'll definitely be another update for v1.10.0. As soon as this is up on pip, you can install it.

In the meantime, you can always [compile spaCy from source](https://spacy.io/docs/usage/#source). This will give you the latest version of spaCy that's currently on `master`, including your Russian code.

Btw, one more question: I had a look at `pymorphy2` and saw that it also supports Ukrainian – does that mean we could adapt the `Russian` class and add `Ukrainian` as well using the same approach (assuming we can crowdsource some language data like stop words and tokenizer exceptions)?

yuukos:
@ines well. I got it. I suppose that `pip install -e .` will work for me.

About Ukrainian... Honestly, I do not have any idea about that. I suppose that adapting the `Russian` language class as `Ukrainian` will work (just because they are very similar as languages). I'm so sorry. I can just suppose.

spacy/ru/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        RussianTokenizer._morph = RussianTokenizer._create_morph(MorphAnalyzer)
+
+        self.vocab = nlp.vocab if nlp else cls.create_vocab(nlp)
+        self._spacy_tokenizer = spacy_tokenizer
+
+    def __call__(self, text):
+        words = [self._normalize(RussianTokenizer._get_word(token))
+                 for token in self._spacy_tokenizer(text)]
+
+        return Doc(self.vocab, words, [False] * len(words))
```
honnibal: Thanks for all this! Definitely an exciting patch. I think the engine in `pymorphy2` might be useful for other languages as well.

Now that I'm rereading the patch, I think this line might have a bug. `[False] * len(words)` is setting the `spacy` attribute on the tokens, which sets whether the token 'owns' a trailing space. This is used to build the `doc.text` and `span.text` strings.

If you can't determine the spacing from the text, it's best to set this to `True`, not `False`. But to be fully correct, we should match into the string, and check whether there's a space. I haven't tested this, but the loop would look something like this:

```python

def get_spaces(token_strings, text):
    i = 0
    spaces = []
    for token in token_strings:
        i += len(token)
        if text[i] == ' ':
            spaces.append(True)
            i += 1
        else:
            spaces.append(False)
    return spaces
```

yuukos: Oh.. Thank you for review and suggestion! I'll think it over and fix, but not right now. In few hours. Again thank you.

PR #1420, opened 2017-10-15
[ja] Stash tokenizer output for speed

polm:
## Description

Before this commit, the Mecab tokenizer had to be called twice when
creating a Doc- once during tokenization and once during tagging. This
creates a JapaneseDoc wrapper class for Doc that stashes the parsed
tokenizer output to remove redundant processing. 

I originally tried doing this without a Doc wrapper class by handling the shared info in the `Japanese.make_doc` function. However, it seems `make_doc` isn't used the way I thought... I expected it to be called in the below code:

    import spacy
    ja = spacy.load('ja')
    doc = ja("日本語だよ")

But it seems it wasn't, and as a result it couldn't handle sharing information between the Tokenizer and Tagger.

Except for adding the one attribute to store Mecab output `JapaneseDoc` is a pass-through class, so I think this is OK, but if there's a better way to do this let me know. While Mecab is relatively fast, it's slow compared to everything else in Doc creation, so avoiding redundant calls would be helpful for high-volume processing (model building etc.).

## Types of changes

- [x] none of the above

## Checklist:

Since this is an implementation detail I don't believe any extra tests are required.

### DIFF
--- a/spacy/ja/__init__.py
+++ b/spacy/ja/__init__.py
@@ -16,12 +16,7 @@ from collections import namedtuple
 
 ShortUnitWord = namedtuple('ShortUnitWord', ['surface', 'base_form', 'part_of_speech'])
 
-class JapaneseDoc(Doc):
-    def __init__(self, detailed_tokens, vocab, words=None, spaces=None, orths_and_spaces=None):
-        super(JapaneseDoc, self).__init__(vocab, words, spaces, orths_and_spaces)
-        # This saves tokenizer output so mecab doesn't have to be called again
-        # when determining POS tags.
-        self.detailed_tokens = detailed_tokens
+DETAILS_KEY = 'mecab_details'
 
 def try_mecab_import():
     """Mecab is required for Japanese support, so check for it.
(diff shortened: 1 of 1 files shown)


honnibal:
Could you put the data in `doc.user_data` instead of creating the extra class?

In spaCy 2 the system for storing extra data on the Doc is much better. See here: https://alpha.spacy.io/usage/processing-pipelines#custom-components-attributes

polm:
OK, it uses `user_data` now; I had assumed that was reserved for use outside spaCy proper.

The spaCy 2 model for custom attributes looks very nice.

honnibal:
You're right that user data is generally reserved for outside-of-spaCy use. I think this usage is okay; it'll definitely get in the way less than having a subclass.

I thought about suggesting a small cache instead, which would randomly drop an item when it becomes full. This might be useful for other components as well. Providing general-purpose caching is very error-prone, though: often the cache function will be "almost correct" (the worst kind of incorrect...). For example, it's normal to want to modify a document (e.g. split/merge tokens) and reapply an earlier analysis. This would be frustrated by a cache if the key isn't sensitive to the right information.

Issue #285, opened 2016-03-10
Streaming Data Memory Growth


ELind77:
I have been using spacy for streaming data (twitter and news stories mostly) and I believe that the fundamental design of the vocab/StringStore in spacy is problematic for streaming processing.  When used for batch jobs the additional memory overhead of storing a new lexeme struct for each new word form encountered in parsing is negligible compared to the speed gains, and because most text conforms to the assumption that vocabulary size grows logarithmically as the total number of tokens grows linearly this is usually a safe bet.  But for streaming text, especially for social media where new terms are invented by the minute (hashtags and URLs in particular) this assumption no longer holds and the spacy vocabulary storage represents a dynamic element in what should be a completely static production deployment.   

In order to test this assumption, I took one million tweets and performed a rudimentary analysis using the resources module in python to get the maximum memory used by the program at regular intervals during processing.  I first performed some minor preprocessing to remove newlines from the data so that it could be read line by line so that it wouldn't all be kept in memory, then I ran spacy with all models set to false, only the tokenizer loaded.  I then did the same thing again after removing all URLs, hashtags, and twitter mentions from the data , and then filtering all empty strings (this resulted in a 1.4% data loss in terms of total tweets processed but that's fairly minor).  

The final result was that spacy used an additional 278.6 MB after tokenizing the raw tweets and 60.99 MB of additional memory when tokenizing the pre-processed tweets.  This result confirms my hypothesis but also shows that the memory increase really isn't all that significant (especially at the relatively low volume that I am currently processing).  But it still points to a potential flaw in the design of the library.   

My suggestion/request in the near term would be to have an option to make the vocabulary read only so that users who want to be able to leave spacy alone to do streaming data processing don't need to worry about changing memory requirements.  In the long term, I think that an optimal solution would be to add some functionality for a timeout on vocabulary entries that aren't loaded at initialization.  E.g. if this lexeme hasn't been accessed for the last _n_ seconds, delete it from the StringStore.  And _n_ would be user configurable. 

My code and results are available here: https://github.com/ELind77/spacy_memory_growth 

Thanks again for continuing to develop such a great library! 

-- Eric

honnibal:
I really need to fix this issue. Thanks.

syllog1sm:
I think this patch from a few weeks ago might have fixed this. Unfortunately the patch wasn't pushed to master --- I've only just merged it now.

[...]

PR #1424, opened 2017-10-16
💫 Fix streaming data memory growth (!!)

honnibal:
This patch follows through on the changes to the StringStore introduced in v2, and fixes by far the longest open bug: #285 Streaming Data Memory Growth

The solution implemented allows a weakref to be created on `Doc` objects, so that we can figure out when it's safe to flush out old strings. A rolling buffer is created using two StringStore objects. One StringStore contains the set of original strings and strings from recent documents. The other string store is the one current on the Vocab object. The only difference between the two `StringStore` objects are in strings created by increasingly old documents. Once we figure out that all of those old documents have passed out of scope and been reference counted, we figure it's safe to flush the obsolete strings.

This logic is only applied in the `Language.pipe()` method, and requires no API changes. However, it does require breaking the former 2-cycle between the `Doc` and `Token` objects, that had been created by caching the `Token` instances within the `Doc`. I think breaking this cycle is a Good Thing, independent of the `StringStore` changes. We don't need to cache those objects --- almost nothing is done in `Token.__init__`, so there should be no problem with recreating the object. If we like, we could use a weakref on `Token`, but it's simpler to just not cache.

The resolution was checked with this script, to verify the string store does not grow as data streams through:

```python
import spacy
import numpy.random

def generate_strings():
    while True:
        yield str(numpy.random.random())

def main():
    nlp = spacy.blank('xx')
    nlp.tokenizer.infix_finditer = None
    for i, doc in enumerate(nlp.pipe(generate_strings())):
        if not i % 1000:
            print(i, len(nlp.vocab.strings))

if __name__ == '__main__':
    main()
```

## Checklist:

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -526,7 +526,7 @@ class Language(object):
             for word in doc:
                 recent_strings.add(word.text)
             recent_refs.add(doc)
-            if nr_seen < 1000:
+            if nr_seen < 10000:
                 old_refs.add(doc)
                 nr_seen += 1
             elif len(old_refs) == 0:
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -8,6 +8,7 @@ import random
 import ujson
 from collections import OrderedDict
 import itertools
+import weakref
 
 from .tokenizer import Tokenizer
 from .vocab import Vocab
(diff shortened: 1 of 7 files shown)


ashaffer:
Hey, thanks for fixing this. Is this going to be backported into v1 though?

Issue #641, opened 2016-11-20
💫 Indic language tokenizers (Hindi, etc)


honnibal:
I think there are now quite a lot of Indian users of spaCy. Let's get started on the tokenizers :).

Hindi is whitespace-delimited, right? The docs for adding tokenizers can be found here: https://spacy.io/docs/usage/customizing-tokenizer

souravsingh:
@honnibal I would like to work on adding the Hindi tokenizers

kaustubhn:
Hi I have started working on a Indian language model for Marathi.
Things I have:
1. A corpus of news articles gathered from news sites size. 3 GB
2. Basic understanding of Natural language processing and Machine Learning

Things required to implement a language in Spacy.
1. Word Frequencies
2. Brown Clusters
3. Word Vectors
4. Stop Words List

Things I have done below things so far.
1. Brown Clusters

I am working towards getting the remaining things.

I wanted to know a few things.
1. Amy going in the right direction as far as implementing a language in spacy?
2. Is the data enough to implement a language model?
3. How to reduce time to generate brown clusters ? 
    i. I performed 500 clusters on 3 GB data it took around 50 Hours.

Other data that I have is for languages Hindi and Tamil.

SanketSKasar:
@kaustubhn Can you please update with the progress?
I am interested, especially for Marathi and Hindi but need some guidance to get started with it.

kaustubhn:
@SanketSKasar anyway we can communicate over email? would like to share the process and files. let me know how to get in touch.

SanketSKasar:
@kaustubhn contact me on kasar.sanket.s@gmail.com.

gauravgr:
@kaustubhn Interested in contributing in this effort. can we get it touch. 
gauravgr@gmail.com

kaustubhn:
@SanketSKasar @gauravgr  you can remove you comments (email) .

ines:
I've just added a simple `Language` class for Hindi over on a feature branch of develop (v2.0.0): https://github.com/explosion/spaCy/tree/feature/hindi-tokenizer/spacy/lang/hi

It currently uses spaCy's basic tokenizer, adds stop words and a simple function setting a token's `NORM` attribute to the word stem, if available (adapted from [here](http://research.variancia.com/hindi_stemmer/) / [here](http://computing.open.ac.uk/Sites/EACLSouthAsia/Papers/p6-Ramanathan.pdf)).

[...]

PR #1425, opened 2017-10-16
💫 Basic Hindi tokenization support

ines:
Related issues: #641
Also CC'ing everyone who participated in the above issue:
@kaustubhn, @gauravgr, @SanketSKasar, @souravsingh 

This pull request adds a basic Hindi `Language` class to support tokenization with spaCy. It also includes a getter for the `NORM` attribute that adds the stem word if available (adapted from [here](http://computing.open.ac.uk/Sites/EACLSouthAsia/Papers/p6-Ramanathan.pdf)).

Since Hindi support has been requested a lot in the past, I hope this will make it easier for others to contribute and improve the language data. (I don't know Hindi, so I was only able to add the very basics – feedback and contributions are appreciated!)

## Todos and further ideas

*  Add `LIKE_NUM` and `IS_DIGIT` getters and include the Hindi digit representations (see [here](https://github.com/explosion/spaCy/issues/641#issuecomment-336636719)). They should probably also be included in the `NORM`. One challenge here is to find the best and most performant way to check whether a string consists of Hindi digits.
* Add tokenizer exceptions and [other language data](https://alpha.spacy.io/usage/adding-languages)
* **Add tests!** Even a few simple examples covering text and expected tokens, or text and expected norms (stem words etc.) would be incredibly helpful.


spacy/lang/hi/stop_words.py
```diff
@@ -0,0 +1,177 @@
...
+कई
+कर
+करता
+करते
+करना
+करने
+करें
+कहते
+कहा
+का
```
smarthi: Also add 'अर्थात', 'किन्तु', 'परन्तु', 'अपितु', 'कदाचित'  ???  - need a Hindi Pandit to ratify ?

kaustubhn: Isn't 'कदाचित' Marathi word, I have almost never seen it used in Hindi.

smarthi: Neh, its in Hindi, Marathi and Sanskrit (check ??).  U definitely see its wide usage in periodic movies and epic hindi serials.

smarthi: .. but i could be wrong too - i would say let's keep that out of stopwords - if some Hindi literary guy could comment on its frequency of usage and if it deserves to be a StopWord or not - let's go with that

kaustubhn: Actually the base script used is 'Devanagri' (https://en.wikipedia.org/wiki/Devanagari) so its a valid word in both languages (Hindi, Marathi), Also not sure what qualifies as a valid token in NLP is it just the occurrence or the usage in the actual language? I wish someone with Hindi Literature understanding would take a look that would be helpful.

smarthi: +!, hence my earlier comment about a Hindi Pandit having to ratify this.

ines: Stop words don't have to be 100% perfect, but in general (also for tokenizer exceptions and other language data), we always try to make sure spaCy also covers the *actual* usage in the real world. For example, in the English stop words, including the word "thru" is almost more important and relevant than the word "whence" (which is usually pretty uncommon in the type of text people are using with spaCy). So keeping a good balance here is always good 👍

smarthi: That's a great point @ines, i think we are good then with what we have here. +1 to merge.

ines: Great – merging this now so it's on `develop` and can be tested! If anyone wants to contribute more data or tests, you can submit a PR to the `develop` branch 😊  [See here](https://alpha.spacy.io/usage/adding-languages) for the usage guide on adding language data to spaCy in v2.0.

PR #1435, opened 2017-10-18
Support single value for attribute list in doc.to_array

ramananbalakrishnan:
## Description
- Updates `Doc.to_array(attr_ids)` to accept a single value in `attr_ids`
- Partial fix for discussion in issue https://github.com/explosion/spaCy/issues/1410.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly.

### DIFF
--- a/.github/contributors/ramananbalakrishnan.md
+++ b/.github/contributors/ramananbalakrishnan.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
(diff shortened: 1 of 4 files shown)


ramananbalakrishnan:
I am not a Cython expert, so could not get the `reshape` command working as-is. If there is a better to way to achieve the same result, do let me know.

Also, can I use directly use [`intify`](https://github.com/explosion/spaCy/blob/cd9378c8/spacy/attrs.pyx#L98) to map a string name ("LEMMA") in the input list to the attribute id (`spacy.attrs.lemma`) ?

ramananbalakrishnan:
I have ported the changes over to the `develop` branch as well, for merging into v2.  ( PR https://github.com/explosion/spaCy/pull/1440 )
I should be able to keep track of suggestions from either PR and fix across both branches.

honnibal:
Thanks for this! It will be merged in time for the new v2 release.

honnibal:
> Also, can I directly use intify_attrs to map a string name ("LEMMA") in the input list to the attribute id (spacy.attrs.lemma) ?

I *think* so. It's possible there's some quirks, but yes that's what we would want to be using.

ramananbalakrishnan:
Done. `to_array` now supports string names in `attr_ids`.
Updated tests and the v2 PR as well.

honnibal:
:tada:

Thanks for this! Made some small changes, especially changing use of `type` in the conditionals. This is a little bit like the weird equality checking stuff in Javascript --- it's best to avoid type-based checks in Python if possible, because the language makes it hard to make the logic fully correct. `hasattr` checks are ugly, but it's a bit easier to cover all the cases this way.

ramananbalakrishnan:
got it! looks much cleaner now.

Also, since you are directly returning `output.reshape()`, looks like the `output_1D` variable is redundant now.

PR #1461, opened 2017-10-25
💫 Add Language.disable_pipes(), to temporarily edit pipeline and update code examples

honnibal:
Often we want to work on some text with only part of the pipeline. This is especially common during training. To make this easier, this patch adds a method to the `Language` class that lets you remove pipes by name. The main use is as a context manager, but you can also use it as a method. See patch for examples.

### Types of changes

New feature

## Checklist

- [x] My changes don't require a change to the documentation, or if they do, I've added all required details.


ines:
Also started cleaning up the examples and somehow ended up pushing everything to this branch, since the training examples now all use `Language.disable_pipes()`.

So this branch now also includes new and updated examples for v2.0:
* updating the named entity recognizer
* training a new entity type
* training the parser
* training the tagger
* entity relation extraction
* navigating the parse tree and subtrees
* phrase matcher

Issue #1202, opened 2017-07-19
Version pin prevents numpy 1.13 usage


zhiruiwang:
I try to install spacy from conda-forge after upgrade my numpy to 1.13.1, while the conda force to downgrade my numpy to 1.12. I wonder whether the spacy support numpy 1.13, or will there be a plan for that?

honnibal:
Hi,

This is purely a problem with the version pin --- there's no *actual* incompatibility with numpy 1.13. All we need is a pull request to the conda forge repo with the version pin and SHA updated. If you have a minute, would you mind doing it? Otherwise I can do it, if it's unfamiliar.

Thanks for pointing this out!

zhiruiwang:
Hi,
I'm not familiar with how to pull request to conda forge. If you have time, please do so. 
Thank you!

zhiruiwang:
Still no one is doing this?

Erlemar:
The problem is still actual.It would be great if it is fixed.

marikgoldstein:
Not quite used to conda forge but happy to submit a PR.

This is the file in question, right? 
spacy-feedstock/recipe/meta.yaml

dataanalyst4lyfe:
I'm still seeing this issue as well.

ines:
Should be fixed in the new version, v2.0. We're currently waiting to get it built for conda (it's just taking a bit longer because of Travis errors and the OSX backlogs 😓). 

Once the version is up, could you try again? [See here](https://spacy.io/usage/v2) for details on what's new in v2.0. If there are still problems with conda install and the new version, feel free to reopen.

marikgoldstein:
Will do!

PR #1479, opened 2017-11-02
Require numpy>=1.12

IamJeffG:
`spacy.util.from_disk` fails below numpy 1.12, so require `numpy >= 1.12`.

## Description

Earlier versions of numpy are not set up to read a `Path`, as is passed from Spacy.

```
/usr/local/lib/python2.7/dist-packages/en_custom/__init__.pyc in load(**overrides)
---> 39     language = load_model_from_init_py(__file__, **overrides)
/usr/local/lib/python2.7/dist-packages/spacy/util.pyc in load_model_from_init_py(init_file, **overrides)
--> 167     return load_model_from_path(data_path, meta, **overrides)
/usr/local/lib/python2.7/dist-packages/spacy/util.pyc in load_model_from_path(model_path, meta, **overrides)
--> 150     return nlp.from_disk(model_path)
/usr/local/lib/python2.7/dist-packages/spacy/language.pyc in from_disk(self, path, disable)
--> 573         util.from_disk(path, deserializers, exclude)
/usr/local/lib/python2.7/dist-packages/spacy/util.pyc in from_disk(path, readers, exclude)
--> 497             reader(path / key)
/usr/local/lib/python2.7/dist-packages/spacy/language.pyc in <lambda>(p)
--> 560             ('vocab', lambda p: self.vocab.from_disk(p)),
vocab.pyx in spacy.vocab.Vocab.from_disk()
vectors.pyx in spacy.vectors.Vectors.from_disk()
/usr/local/lib/python2.7/dist-packages/spacy/util.pyc in from_disk(path, readers, exclude)
--> 497             reader(path / key)
vectors.pyx in spacy.vectors.Vectors.from_disk.load_keys()
/usr/local/lib/python2.7/dist-packages/numpy/lib/npyio.pyc in load(file, mmap_mode, allow_pickle, fix_imports, encoding)
--> 391         magic = fid.read(N)
AttributeError: 'PosixPath' object has no attribute 'read'
```

I can reproduce this (and the fix) on spacy 2alpha. I did not try it on spacy 1+.

One thing for reviewer to be aware of is: https://github.com/explosion/spaCy/issues/1202 — I don't want the now-tighter range of allowed versions to make for impossible dependency resolution in conda.  However that issue does make it sound like `numpy==1.12` is a totally valid solution for both my bug and for conda. 

### Types of changes

- `requirements.txt`
- `setup.py`

## Checklist

- [ ] My changes don't require a change to the documentation, or if they do, I've added all required details.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 cython>=0.24,<0.27.0
 pathlib
-numpy>=1.7
+numpy>=1.12
 cymem>=1.30,<1.32
 preshed>=1.0.0,<2.0.0
 thinc>=6.9.0,<6.10.0
--- a/setup.py
+++ b/setup.py
@@ -191,7 +191,7 @@ def setup_package():
             ext_modules=ext_modules,
             scripts=['bin/spacy'],
             install_requires=[
-                'numpy>=1.7',
+                'numpy>=1.12',
                 'murmurhash>=0.28,<0.29',
                 'cymem>=1.30,<1.32',
                 'preshed>=1.0.0,<2.0.0',


IamJeffG:
PR to add my contributing agreement to `master` is  https://github.com/explosion/spaCy/pull/1481

honnibal:
Thanks! However looking at this I think it's better to say the problem is in our code: we'd rather avoid tightening the pin, and instead just make sure we pass string objects to numpy, instead of `pathlib.Path` objects.

honnibal:
Amongst other things I doubt `cupy` can be relied upon to take `Path` objects, so this would be broken on GPU. I think we should expect to pass string paths into other libraries.

PR #1511, opened 2017-11-08
Fix typo in message

danielhers:


ines:
Ah damn – well spotted! Thanks! 👍

PR #1524, opened 2017-11-09
Fix Doc mistake, showing visual of similarity when dissimilar

MathiasDesch:
## Description
* I just changed the condition on similarity to display similar icon when similarity is > 0.5
* I also fixed a broken link to SCA template that I saw when I filled it

### Types of change
Small bug fix


ines:
Thanks a lot – also for fixing that link in the contributing guide 👍

PR #1542, opened 2017-11-10
Fix typo. Add missing '='.

Wahib:


ines:
Ahh, well spotted – thanks! 👍

PR #1552, opened 2017-11-11
Try to add ability to clean up StringStore in pipe

### LINKED ISSUE #1506: KeyError during stream processing
@ligser writes on Gitter:

```
Hello all! I have some troubles with spacy-nightly==2.0.0rc1 (a18 has same behavior) with en_core_web_lg model. When I run nlp.pipe with a generator of a texts I get the exception: KeyError: 4405041669077156115..
Exception raised after amount of texts (average 10000).
Stacktrace looks like that:
    nlp.pipe((c.content_text for c in texts), batch_size=24, n_threads=8)
  File "doc.pyx", line 375, in spacy.tokens.doc.Doc.text.__get__
  File "doc.pyx", line 232, in __iter__ 
  File "token.pyx", line 178, in spacy.tokens.token.Token.text_with_ws.__get__
  File "strings.pyx", line 116, in spacy.strings.StringStore.__getitem__
KeyError: 4405041669077156115
That looks like a bug with a StringStore cleanup or something related (maybe shared string-store that clean-up by one of threads?).
My code just get a texts from mysql split it to texts and ids and do: for id, doc in zip(ids_gen,nlp.pipe(docs_gen, ...)).
```

I think this is likely due to the solution added in spaCy 2 to address the streaming data memory growth.

ligser:
## Description
I try to fix #1506 by allowing StringStore to clean up old strings when using it to iterate through many documents in a pipe. I try to do it with little effort on memory or speed, but I might not see all effect of my changes.
Because that method is written as the feature only for `pipe` method of the Language class — the method marked as internal by the underscore.

### Types of change
It's bug fix.

## Checklist

### DIFF
--- a/.github/contributors/ligser.md
+++ b/.github/contributors/ligser.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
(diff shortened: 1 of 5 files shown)


honnibal:
This looks very promising. Thank you for taking the time to do this efficiently.

honnibal:
Very good, thanks!

ligser:
@honnibal I realize that [here](https://github.com/explosion/spaCy/pull/1552/files#diff-b6b3e43f85a70fd78f0b9e8395857783L567) I broke swapping of recent and old refs. I afraid that change can cause the error: if `nr_seen` raise upper than 10000 and `old_refs` is not empty at some iterations we can lose some of the refs that not recorded to old_refs. When the old_refs become empty we initiate clean-up and remove strings from N-1th iteration (because of hits — N iteration still here). If one of the docs that we lose from old_refs still not cleaned-up after N+1th iteration and we try to get string data from it — we should get the crash.
I try to create the test case for that — but cannot reproduce that behavior. Maybe I miss something?

ligser:
I understand. My fix not cleans up strings — just keys. 😐

ligser:
I made a new pull request to fix it. Sorry.

PR #1564, opened 2017-11-13
cleaned encoding problems

DuyguA:
Some Turkish only letters had some kind of encoding problems. For instance line13 "altmýþ", is indeed line14 "altmış". I cleaned the duplicates that was led by this problem, also went over the word list once.

## Description

Deleted duplicate lines that look different but indeed the same word
Corrected some words
Go over all the words and see how inclusive is the list

### Types of change

An enhancement

## Checklist

- [x ] My changes don't require a change to the documentation, or if they do, I've added all required information.


ines:
Thanks a lot - this is super valuable! 👍 

I mostly copy-pasted the list from that open-source stop words collection when I set up the `Turkish`  template class and didn't check this properly. Sorry about that. 

Btw, if you have any other suggestions for Turkish (for example, abbreviations or contractions for the `tokenizer_exceptions.py`), let us know. It'd be so cool to get spaCy to support Turkish properly, and maybe even train a model in the future!

DuyguA:
I guessed so, copy pastes usually have encoding problems :)

Sure, I will work on tokenizer exceptions. I'll think of possible additions meanwhile.

Training models requires much more data than English, due to rich word forms. We can collobrate when the time comes :100:

Issue #1248, opened 2017-08-09
Tokenizer: -ing contraction parsed incorrectly


Jeiwan:
Spacy doesn't properly tokenize words with contracted '-ing' ending:
```python
import en_core_web_sm
nlp = en_core_web_sm.load()
doc = nlp("I'm lovin' it")
print(doc[1])
# 'm – CORRECT!
print(doc[1].lemma_)
# be – CORRECT!
print(doc[2])
# lovin – INCORRECT!
print(doc[2].lemma_)
# lovin – INCORRECT!
print(doc[2].pos_)
# ADJ – INCORRECT!
```

## Info about spaCy

* **spaCy version:** 1.9.0
* **Platform:** Darwin-16.4.0-x86_64-i386-64bit
* **Python version:** 3.6.2
* **Installed models:** en_core_web_sm

Is there a workaround?

honnibal:
Thanks!

ines:
Sorry, looks like we forgot to fix this in the tokenizer exceptions for v2.0 – just added it and it will be included in the next model release.

MathiasDesch:
I think there is actually not much of words having this contraction commonly. I believe "lovin'" was introduced by Mac Donald a few years ago and stayed until now. This contraction is actually not registered as official ones.

Therefore, I think it might be a solution to assume that it's the only case where it is correct or that this contraction can be made with all 'ing' ending verbs but not to have a list of exceptions. It means making change in the tokenizer. This seems to be doable to have a general rule for it.

ines:
@MathiasDesch Yeah, I agree – the only other common ones I can think of are maybe goin', doin' and havin'... although, if you're working with texts that use these colloquially (and not as part of slogans etc.), the apostrophe is probably also less common. So maybe those forms could be handled in the [new norm exceptions](https://spacy.io/usage/adding-languages#norm-exceptions), to make sure they receive the same representations as the full forms.

MathiasDesch:
My PR does not solve the POS related problem though. `Lovin'` is still predicted as an `ADJ`

ines:
@MathiasDesch Thanks so much for taking this on! And yeah, I assume the model likely hasn't seen any of those words during training, so it makes sense that spaCy still gets it wrong. 

[...]

PR #1566, opened 2017-11-13
Add exceptions to tokenizer and norm

MathiasDesch:
Add exceptions both in tokenizer and norm as discussed in #1248. This improve the parsing of some -ing verbs using not common contraction in `in`or `in'`

## Description
Should solve #1248
I am not sure if I should add an unittest for this specific case. If yes, I can totally do it.

### Types of change
Bug fix and enhancement. Changes are only on exceptions cases

## Checklist

I have tried on a small example and it seems to work fine.


ines:
Thanks, looks good! 

I actually ended up checking all additions without an apostrophe in a dictionary, just to make sure they're definitely not regular English words (to be honest, I was quite surprised that "goin" wasn't some quaint old term for... I dunno, a farming tool or something like that, haha)

PR #1576, opened 2017-11-14
Actually reset caches in pipe [wip]

ligser:
## Description
After previous PR #1552, StirngStore not actually does any cleanup. It just erases keys to stale data, not the data itself. Now that behavior was fixed and StringStore cleans up correctly. 
Sorry for the inconvenience.

### Types of change
It is a bug fix.

## Checklist


ligser:
And again that does not work. Because of Tokenizer cache.

ligser:
Looks like there is no any simple solution.
String store is not only one method to access that map, also there is cache in Vocab (`_by_hash` and `_by_orth`) and Tokenizer(`_cache`) that store mapping between hash and string.

ligser:
I try to clean all caches when cleanup StringStore but think that solution looks ugly and hacky. If you decide to not accept that code I totally appreciate it.

ligser:
I need some suggestions here. My solution interrupts internal processes in spaCy. That might be the not best way to fix the problem. 

I see few ways:
* My solution that doing a hard cleanup of internal caches.
* Discard cleanup at all. Get `pipe` method back as it looks before fixed memory growth.
* Refactor internal caches to allow them cleanups correctly.

The last solution looks best. But it also requires more time to implement and deep dive into spaCy internals. I can try to implement it but before I done — cleanup did not work.
Maybe you already have some ideas how to do it, and my work concurrent with your)

honnibal:
@ligser Thanks for your work on this. I'm thinking about what to do.

This problem is one of the oldest in spaCy, e.g. see here: https://github.com/explosion/spaCy/issues/285 . So, don't feel bad that your fix had a problem :). Actually you got much further on this than anyone else, and as you saw my own fix to this was broken.

ligser:
I try to make the first option working (no more selective cleanup of tokenizer cache).
That code should be measured and if you would like — you can review and merge it.
I'll try to measure that by myself but if you already have a script to do that — you can just write result here.
I don't do performance measurements in python before. It can take some time for me.

ligser:
I still thinking about a more convenient fix but that one can be used as the hotfix to let the current solution just works.

honnibal:
I've been working on this recently. Try

    python -m spacy profile en_core_web_sm

There are some patches on master which may conflict with your changes. If you don't want to bother with that yet, you could just do `git cherry-pick 8d692771f68` to get the improvements to `spacy profile` into your branch.

After running the profiler, you may also be interested to do `pip install gprof2dot`. You can then run `ggprof2dot -f pstats Profile.prof -o callingGraph.dot` and view the `callingGraph.dot` with GraphViz. For the tokenizer this might not be necessary.

ligser:
I merge master from origin repo, ok!

ligser:
Oh! Looks like cache fix in tokenizer can allow me to get one more try to selective cleanup in the tokenizer.
Is there an available case to reproduce the bug with cache? Looks like I have the same problem in 
3e21680 but I cannot reproduce it.

ligser:
Profile shows promising results — 125 (without the fix) and 123(with it) seconds.

honnibal:
Merging this and looking more closely at it. Seems good so far -- many thanks!

honnibal:
I've added an option `cleanup` to `Language.pipe()`, which is set to `False` by default. This way we can avoid problems while we test this more thoroughly.

Issue #1582, opened 2017-11-15
fast text example  ValueError: could not convert string to float:


yogendrasoni:
vector = numpy.asarray([float(v) for v in pieces[1:]], dtype='f')
ValueError: could not convert string to float:

rtrim line before rsplit is solving this issue
pieces = line.**rstrip()**.rsplit(' ',nr_dim)
## Info about spaCy

* **Python version:** 2.7.13
* **Platform:** Linux-4.4.75-loc-x86_64-Intel-R-_Xeon-R-_CPU_E5-2630_v3_@_2.40GHz-with-slackware-14.2
* **spaCy version:** 2.0.2
* **Models:** en-1.1.0, en_core_web_lg, en_vectors_web_lg, en_glove_cc_300_1m_vectors-1.0.0, en, __cache__, en_core_web_md

ines:
Thanks for the report and solution! Do you want to submit a pull request for this?
(If not, we're also happy to take care of fixing this later.)

PR #1583, opened 2017-11-15
Add rstrip after reading line from  vec file #1582

yogendrasoni:
loading english fast text giving error because line contains new line at the end and rsplit is splitting it incorrectly

## Checklist


ines:
Yay, thanks a lot! 👍

PR #1594, opened 2017-11-16
Update _spacy.jade

pavillet:
Doc example gives 'object is not subscriptable' error.
Correcting as an attribute

## Description

Doc example gives 'object is not subscriptable' error.
Correcting as an attribute

### Types of change

Documentation fix

## Checklist


ines:
Thanks, well spotted!

PR #1604, opened 2017-11-17
Fix typo in documentation

raphael0202:
### Types of change
Update in documentation.
## Checklist


ines:
Thanks! 👍 

I feel like I need to finally find a good solution for spell/typo checking in my text editor... I never really got this to work properly, because it's difficult to make it ignore the markup. Then again, I hope that with all the incoming pull requests, we *must* eventually run out of typos to correct 😜

Issue #1494, opened 2017-11-05
Custom tokenizer with infix_finditer is not working correctly.


damianoporta:
## Your Environment
* Operating System: 16.04
* Python Version Used: 3.5.2 
* spaCy Version Used: develop

Ok, I am brute-forcing the tokenizer :) pardon @ines in advance haha.
Maybe I have found a bug with custom infix pattern. The code:

```
import spacy
import re
from spacy.tokenizer import Tokenizer

infix_re = re.compile(r'''[^a-z]''')

def my_tokenizer(nlp):
    return Tokenizer(nlp.vocab,
                     {},
                     infix_finditer=infix_re.finditer
    )

nlp = spacy.load('en')
nlp.tokenizer = my_tokenizer(nlp)

doc = nlp("This is a ...test")

for token in doc:
    print(token.text)
```
The output is:

```
This
is
a
.
.
.test

```

As you can see it holds the last DOT with the word "test". 
It occurs when there are three consecutive characters outside the [^a-z]. So, for example:

`doc = nlp("This is a 123test")` returns:

```
This
is
a
1
2
3test

```
`doc = nlp("This is a .-/test")` returns:

```
This
is
a
.
-
/test

```
and so on...
In this case the 3 is not a perfect number :d

The following are working correctly:

`doc = nlp("This is a 12test")`
```
This
is
a
1
2
test

```
`doc = nlp("This is a -/test")`

```
This
is
a
-
/
test

```

Hope this helps.

fsonntag:
According to the [documentation](https://spacy.io/usage/linguistic-features#how-tokenizer-works), this behaviour is correct.
Check this part:
```
            elif find_infixes(substring):
                infixes = find_infixes(substring)
                offset = 0
                for match in infixes:
                    tokens.append(substring[offset : match.start()])
                    tokens.append(substring[match.start() : match.end()])
                    offset = match.end()
                substring = substring[offset:]
```

In the case of `"A 123test"`, it finds `"2"` and `"3"` as infixes, with a `match(start=3, end=4)` and `match(start=4, end=5)`. The tokenizer starts to split at the first match and appends `"1"` (`tokens.append(substring[offset : match.start()])`) and `"2"` (`tokens.append(substring[match.start() : match.end()]`) to `tokens`.
`offset` is set to to 4 and in the next iteration `match.start()` is also 4, so `tokens.append(substring[offset : match.start()])` has no effect anymore.

[...]

PR #1611, opened 2017-11-19
Solving #1494

fsonntag:
## Description
See #1494.

### Types of change
Changed behaviour of tokenizer infix search. When infix_start equals the current offset, previously the whole infix tokenization was skipped. Now we add the infix itself, but still ignore the previous tokens (which are basically not there.

## Checklist

### DIFF
--- a/.github/contributors/fsonntag.md
+++ b/.github/contributors/fsonntag.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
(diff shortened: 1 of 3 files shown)


honnibal:
Looks good! I'll hold off on merging this until I've played with it more thoroughly, but it certainly seems elegant. Test looks good too.

honnibal:
What happens if we have your pattern and just `hello 3test`? This will segment into `hello, 3, test` -- so the infix pattern will match at the start of the token, like a prefix. This will be against the intentions of the pattern, since if you want to split that, you should put the logic in the prefix pattern.

It seems like we can cover this case by setting a variable `start_before_infixes = start` before looping over the infix matches. We can then check `infix_start == start_before_infixes` to `continue` the loop as before. This should split `hello...test` into `hello, ., ., ., test` if you have `.` as an infix, without splitting `hello .test`.

honnibal:
Awesome, thanks!

fsonntag:
I have to say thank you for this great tool :)

PR #1614, opened 2017-11-20
Use flake8 to look for syntax errors, undefined names

cclauss:
## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ install:
 script:
   - "pip install pytest pytest-timeout"
   - if [[ "${VIA}" == "compile" ]]; then python -m pytest --tb=native spacy; fi
-  - if [[ "${VIA}" == "flake8" ]]; then flake8 . --count --exclude=spacy/lang --select=E901,E999,F821,F822,F823 --show-source --statistics; fi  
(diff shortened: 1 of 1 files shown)


ines:
Thanks! I actually wouldn't mind integrating this into the build process, and I like `flake8` (although I know that @honnibal usually finds its defaults too strict 😜 – but this shouldn't be a problem here considering the explicitly specified codes).

The only problem here is that this will only catch errors in the Python files and not the Cython source (which I guess is still better than nothing, and making linting work for Cython is pretty tricky anyways). It's also really nice to at least have a sanity check for the examples.

Just summarising the included error/warning codes again to make this easier to review:

| Code | Description |
| --- | --- |
| `E901` | SyntaxError or IndentationError |
| `E999` | SyntaxError: invalid syntax |
| `F821` | undefined name | 
| `F822` | undefined name name in `__all__` |
| `F823` | local variable name ... referenced before assignment |

The error code range should also be narrow enough to avoid having to add too many random `# noqa` exceptions to the code base (which is something we definitely want to avoid.)

honnibal:
Thanks! I'll merge this, as the balance will shift more and more in its favour as the project grows and the balance of work on the repo shifts towards managing pull requests.

PR #1619, opened 2017-11-20
Use six.builtins and six.moves to simplify compat

cclauss:
* http://python-future.org/compatible_idioms.html
* https://pythonhosted.org/six/#binary-and-text-data

I believe that [six.b() and six.u()](https://pythonhosted.org/six/#binary-and-text-data) could be used to simplify compat.py

### DIFF
--- a/spacy/compat.py
+++ b/spacy/compat.py
@@ -7,6 +7,9 @@ import sys
 import ujson
 import itertools
 import locale
+from six.builtins import bytes as bytes_
+from six.builtins import str as unicode_
+from six.moves import input as input_
 
 from thinc.neural.util import copy_array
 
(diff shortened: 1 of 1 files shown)


bdewilde:
@ines fwiw i've run into unexpected subtleties in behavior when using `six` or `future` instead of a more "manual" `compat.py` setup as `spacy` currently uses. since the compat module is relatively narrow and straightforward as-is, this may be a solution in search of a problem. none of this is to say that your changes are bad or wrong, @cclauss , just wanted to share my experience with this when developing `textacy`.

ines:
@bdewilde Thanks, this is good to know! I remember adopting the compat module from `textacy`... Do you remember which problems you came across specifically?

In any case, maybe we can put this on hold for now, at least until we've been able to test this reliably – we definitely want to avoid accidentally breaking things. And with the current solution, we at least know that it works without problems.

bdewilde:
I don't recall specifically, tbh — it was ~1.5 years ago — but had something to do with subclassing the built-in types. That was with `python-future`, actually, and not `six`. I just took a peek [under the hood](https://github.com/benjaminp/six/blob/master/six.py) at `six`, and it looks like they effectively do what `spacy` and `textacy` both do, albeit for _all_ Py2/3 incompatibilities, rather than just the ones that affect us.

cclauss:
@bdewilde What is your recommendation on this PR?

| six module | Python 2 | Python 3 |
| --- | --- | --- |
| six.class_types | (<type 'type'>, <type 'classobj'>) | (<class 'type'>,) |
| six.integer_types | (<type 'int'>, <type 'long'>) |  (<class 'int'>,) |
| six.string_types | (<type 'basestring'>,) | (<class 'str'>,) |
| six.text_type | <type 'unicode'> | <class 'str'> |
| six.binary_type | <type 'str'> | <class 'bytes'> |

bdewilde:
@cclauss I think it probably wouldn't break anything, but that nothing is currently broken. :)

I defer to @ines!

honnibal:
I think I prefer to close this, for the reasons @bdewilde notes. Using a dependency instead of our own code isn't an intrinsic benefit, especially for small arbitrary stuff like the compatibility functions. Thanks!

PR #1624, opened 2017-11-21
Add support for Russian

GreenRiverRUS:
## Description
- returned some code of @yuukos from v1.10 branch.
- added some more exceptions, also `like_num` lex attribute
- added cyrillic char classes
- added custom lemmatizer based on `pymorphy2`
- wrote some tests
- added custom tag map based on UD Russian SynTagRus corpus for tagger to support Russian morphology (which is used by lemmatizer for correct lemma choosing)

P.S. I also have a trained model with vectors and brown clusters trained on wiki, and tagger trained on UD. How can I contribute it?

### Types of change

## Checklist


honnibal:
@GreenRiverRUS Awesome! Thanks

Did you need to do anything special to train the Russian model? Also, what text did you use for the vectors?

I think what we've found from previous experience is that we have to be careful to have the supported languages in our build pipeline, because otherwise we release new versions of spaCy and support for the different languages comes and goes.

The model building process is still a bit flakey (dev ops isn't my strong suit...), and doesn't include the vector training. But we might be able to add support for Russian already. Alternatively, you could make a Github repo and attach the trained model as a "release file". This would let people download the model easily, without you having to host it for them.

GreenRiverRUS:
@honnibal 
I have trained model on wiki corpus, obtained as described [here](https://github.com/explosion/spacy-dev-resources/tree/master/vocab) with minor fixies of code as it was written for spacy v1.x. I didn't need smth special except writing a [script to create model itself from raw data](https://github.com/explosion/spaCy/pull/1646).

PR #1631, opened 2017-11-23
Use local parameter in example MyComponent

markulrich:
## Description

Documentation typo fix

### Types of change

Typo fix and added contribute agreement

## Checklist


honnibal:
Thanks. Added another small fix while I was looking at the page.

Issue #1618, opened 2017-11-20
Drop support for Python versions that are EOL?


cclauss:
[Python 2.6 and 3.3 have already reached End of Life.](https://docs.python.org/devguide/index.html#branchstatus)  Is it useful to continue supporting these legacy versions of Python?

https://github.com/explosion/spaCy/blob/master/setup.py#L214-L216

honnibal:
Thank, I think the classifier in the setup is actually wrong. I suspect py2.6 has been broken for some time. py3.3 was horrible and nobody used it.

PR #1644, opened 2017-11-26
Drop support for EOL Python 2.6 and 3.3

hugovk:
## Description

Fixes https://github.com/explosion/spaCy/issues/1618.

Drop support for EOL Python 2.6 and 3.3, which also means set literals may be used. 

### Types of change

I think a bug fix.

## Checklist


honnibal:
Thanks for this! Agree with updating the docs and dropping declared support for 2.6 and 3.3. However I don't think switching to set literals is of immediate benefit.

I mean, the code's been working on 2.6 right? So why go out of our way to break it? We don't want to take trouble to support these old versions, but if the code happens to work on them, that's nice.

hugovk:
Well, we all as a community benefit if people move off 2.6, see the links below, but, sure, I can revert that commit when I'm back at a keyboard.

https://snarky.ca/stop-using-python-2-6/
http://www.curiousefficiency.org/posts/2015/04/stop-supporting-python26.html

honnibal:
@hugovk Thanks, revert when you can and then I'll merge.

I quite disagree with the philosophy in that post. It's not a kindness to break code that would otherwise work in some environment, to "helpfully" discourage people from using that environment.

hugovk:
@honnibal Done!

honnibal:
Thanks!

PR #1645, opened 2017-11-26
Fixed spaCy version string in default meta

GreenRiverRUS:
Default spaCy version string in language meta causes error if one tries to install the model package.

## Description
I've noticed that by default blank language [creates meta with current spaCy version](https://github.com/explosion/spaCy/blob/master/spacy/language.py#L150). If one packages this model with default meta, a setup file will be created from [this](https://github.com/explosion/spaCy/blob/master/spacy/cli/package.py#L149) template, which builds requirements from meta and expects `==`, `>=`, or `<=` before version number. So after running `setup.py` a wrong requirement `spacy2.x.x` (instead of `spacy>=2.x.x` or smth like that) causes the error.
I'm not sure, I fixed it in the right place, but it works at least.

### Types of change

- [x] Bug fix (non-breaking change fixing an issue)

## Checklist

- [x] I have submitted the spaCy Contributor Agreement (in previous PR)

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -147,7 +147,7 @@ class Language(object):
         self._meta.setdefault('lang', self.vocab.lang)
         self._meta.setdefault('name', 'model')
         self._meta.setdefault('version', '0.0.0')
-        self._meta.setdefault('spacy_version', about.__version__)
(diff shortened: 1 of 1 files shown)


ines:
Ah damn, no idea how this happened! The `package` command does this correctly. Thanks a lot for the fix.

PR #1646, opened 2017-11-26
Added model command to create models from raw data

GreenRiverRUS:
I'm not sure that this command should be added, but since I have already written it for myself, why not share it :smile:

## Description
I've merged v1.x `model` cli command with v2.x `vocab` command into a new one, which helps to create models from raw data: words counts, brown clusters and vectors.
The command requires the following parameters:
- `lang`: model language
- `output_dir`: model output directory to store to
- `freqs_loc`: path to word counts file, should be in format, which described [here](https://spacy.io/usage/adding-languages#training)
- `clusters_loc`: [optional] location of Brown clusters data in format produced by [this](https://github.com/percyliang/brown-cluster) script
- `vectors_loc`: [optional] location of [word vectors](https://spacy.io/usage/adding-languages#word-vectors) file in GenSim text format (created by default)
- `prune_vectors`: [optional] number of vectors to prune to

### Types of change

## Checklist

- [x] I have submitted the spaCy Contributor Agreement (in first PR)

### DIFF
--- a/spacy/cli/model.py
+++ b/spacy/cli/model.py
@@ -1,0 +1,150 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+import plac
+import math
+
+from tqdm import tqdm
+
+import spacy
+import numpy
+from ast import literal_eval
+from pathlib import Path
+from preshed.counter import PreshCounter
+
+from spacy.compat import fix_text
+from spacy.vectors import Vectors
+from spacy.util import prints, ensure_path
+
+
+@plac.annotations(
+    lang=("model language", "positional", None, str),
+    output_dir=("model output directory", "positional", None, Path),
+    freqs_loc=("location of words frequencies file", "positional",
+               None, Path),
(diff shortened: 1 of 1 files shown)


ines:
Thanks, this is cool! 👍 

One thing I've been thinking of: In spaCy v1.x, we never really liked the name `model` that we came up with for this command, because it's so vague... so maybe you have an idea for a better name? It'd definitely be great to bring this functionality back in spaCy v2.x.

GreenRiverRUS:
@ines
Maybe `init` as `git init`, but it isn't speaking for itself :thinking: . I'm not good at naming, unfortunately.

ines:
Hmm, maybe `init-model`, `model-data` or `lang-model`? Just `spacy init` might be a bit weird, because it sounds like you're... initialising spaCy or something. I'll think about this and see what @honnibal thinks!

ines:
Merging this and will make a few more edits afterwards, including the name change 👍

PR #1647, opened 2017-11-26
Corrected char index instead of token index

atomobianco:
Changed the index used to add the label because `displacy.render` apparently uses char index

## Checklist


ines:
Thanks, well spotted – displaCy definitely needs the char index, because it operates on the raw text. Also thanks for fixing the formatting, looks much better now 👍

PR #1650, opened 2017-11-27
Add several Danish alternative spellings

sorenlind:
## Description
I have added several alternative spellings to the Danish normalization exceptions. I have tried to consistently normalize to the most "modern" spelling. For example, "bedstemoder" is normalized to "bedstemor" and "akvarium" is normalized to "akvarie". I've added tests for a couple of examples.

### Types of change
Enhancement.

## Checklist


ines:
Thanks, this looks good! 👍

There was one test failure, but only on Python 2 – I suspect this might have been caused by the missing `unicode_literals` import in the norm exceptions file, which means the strings were passed in as byte strings instead of unicode. So I just updated that and I'm pretty sure it should pass now.

sorenlind:
Thanks! I had only run the tests on Python 3. Will not make that mistake again :) I've just run the tests locally with the added `unicode_literals` using Python and they pass so hopefully they will on the CI too. Thanks for the fix!

ines:
No problem – that exactly why we have the CI integration set up to run on each PR. Running the tests locally is good, but we don't expect you to run extensive cross-platform and cross-version tests before every PR (unless you already have a complex setup in place and this is super easy for you 😜)

Tests just passed, so I'm merging now! 🎉

PR #1667, opened 2017-11-30
Bugfixies in russian support

GreenRiverRUS:
## Description
Fixed some bugs in russian language support code, that I haven't noticed previously. As for tests, I ran then so many times locally that didn't noticed when return statement was accidentally deleted. Sorry :disappointed: 

### Types of change
- [x] Bugfix

## Checklist


ines:
Thanks and no worries! 

The nice thing about the language data in v2.0 and the lazy loading is that at least, languages are now pretty isolated, so bugs in one language won't affect the rest of spaCy's functionality. (However, sometimes this also means that bugs in a language can go unnoticed for longer – we're still thinking about how to improve the test framework for this,  especially the languages using external dependencies that are not part of spaCy's dependencies.)

PR #1668, opened 2017-11-30
Add more Danish morph rules and clean up existing ones

sorenlind:
## Description
I have made some changes to the Danish morph rules. Specifically I've done the following:

* I systematically went through all forms of Danish personal pronouns, adding missing ones and fixing incorrect ones.
* I used the Danish Universal Dependencies for reference in all cases, except 'den'/'det' (Danish for the English word 'it'). I believe  Danish UD contains systematic errors for those words: Case is always specified as 'acc' even when the word refers to the subject (and consequently, Case should be set to 'nom'). I have left Case unspecified for 'den' and 'det'.
* I removed use of 'Masc' and 'Fem' for Gender as Danish only has the genders  'Neut' and 'Com' (common).

### Types of change
Bug fix / enhancement

## Checklist

### DIFF
--- a/spacy/lang/da/__init__.py
+++ b/spacy/lang/da/__init__.py
@@ -22,7 +22,7 @@ class DanishDefaults(Language.Defaults):
     lex_attr_getters[NORM] = add_lookups(Language.Defaults.lex_attr_getters[NORM],
                                          BASE_NORMS, NORM_EXCEPTIONS)
     tokenizer_exceptions = update_exc(BASE_EXCEPTIONS, TOKENIZER_EXCEPTIONS)
-    # morph_rules = MORPH_RULES
+    morph_rules = MORPH_RULES
     tag_map = TAG_MAP
     stop_words = STOP_WORDS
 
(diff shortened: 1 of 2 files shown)


sorenlind:
I was a little too quick, I hadn't realized the Danish morph rules are currently disabled in `__init__.py`. I will look further into this.

sorenlind:
@ines : My [last commit](https://github.com/explosion/spaCy/pull/1668/commits/d86b537a3820b23d66b5a8d52d15ae5d11c2b34b) enables the morph rules for Danish. I believe they work now but any hints on how to test this would be greatly appreciated :). 

When using a trained model, I can do  this (the text is Danish for "I see you"):
```python
for token in nlp("jeg ser dig"):
    print(token, token.tag_, token.lemma_, token.norm_)
```
which results in the following:
```
jeg PRON -PRON- jeg
ser VERB ser ser
dig PRON -PRON- dig
```

honnibal:
Thanks!

PR #1672, opened 2017-12-01
Adds Turkish Lemmatization

cbilgili:
- Ported Turkish lemmatization lookup dictionary from Turkish dictionary site http://st2.zargan.com/duyuru/Zargan_Linguistic_Resources_for_Turkish.html
- Adds simple Turkish test suite and lemmatization test

### DIFF
--- a/.github/contributors/cbilgili.md
+++ b/.github/contributors/cbilgili.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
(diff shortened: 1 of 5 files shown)


ines:
Hi @cbilgili – thanks for your work on this! 👍 I just had a look at the data source and the license there is listed as CC BY-NC-SA 4.0 – this also applies to the lemmatizer data, right?

If that's the case, this is a problem, because it's not compatible with spaCy's MIT license, and we won't be able to merge it into the core library. So we'd have the following options:

* Find a different data source that's compatible with the MIT license.
* Since you've already put in the work, you could take your code, wrap it as an extension package or a [custom pipeline component](https://spacy.io/usage/processing-pipelines#custom-components) and publish it on your account, licensed as CC BY-NC-SA 4.0. This means that people could use it with spaCy under those license terms – but of course, it wouldn't be as nice and as convenient.

cbilgili:
@ines I have written to the creator for permission. I hope we could get permission. Because actually we don't use his work commercially.

ines:
@cbilgili Thanks – keep us updated on the progress! If the creator has any questions about spaCy or the licensing terms, they can also always get in touch with us via email.

The main problem here is that the [MIT license](https://github.com/explosion/spaCy/blob/master/LICENSE) allows commercial use. Many spaCy users are also using the library for commercial purposes. This is why we can't ship any code or data with spaCy that doesn't allow commercial use, or is otherwise incompatible with the MIT license.

cbilgili:
@ines The author (Orhan Bilgin) approved to use their work in spaCy as I used. Now how could we proceed? Getting permission by a letterhead would be ok? Whom would he officially permit?

ines:
@cbilgili Yay, good news – thanks for taking care of this!  I think the easiest solution would be if he just fills in the [contributor agreement](https://github.com/explosion/spaCy/blob/master/.github/CONTRIBUTOR_AGREEMENT.md).

All he needs to do is:

1. Copy the `CONTRIBUTOR_AGREEMENT.md` to [`.github/contributors`](https://github.com/explosion/spaCy/tree/master/.github/contributors), fill in the details and change the file name to `[his_name_or_github_username].md`.
2. Submit a pull request with the contributor agreement.

cbilgili:
@ines I see, but for him I think it is a bit difficult. How about I send him text and get it back as scanned document? :)

ines:
@cbilgili Sure, if that's easier, that's no problem! 👍

cbilgili:
@ines Scanned document is here;
[spaCy_zargan.pdf](https://github.com/explosion/spaCy/files/1575771/spaCy_zargan.pdf)

ines:
Perfect! 👍 I've added the agreement to the `contributors` directory, and will merge the PR. Thanks again for your work on this!

PR #1674, opened 2017-12-03
fix setup.py spacy req string for packaging

twerkmeister:
## Description
spacy requirement string for model packages was broken.
It should be `spacy>=2.0.2` instead of `spacy2.0.2`

### Types of change
bug fix

## Checklist
- [x] I ran the tests, and all new and existing tests passed. (There are no tests for the cli packaging)


honnibal:
Thank!

PR #1683, opened 2017-12-05
Fix link to CLEAR Style dependency labels PDF

mkdynamic:


ines:
Thanks, merging! (Ignore the build failure btw – I think @honnibal accidentally broke master 😜)

PR #1687, opened 2017-12-06
doc: Add missing *-operator to nlp.disable_pipes() in Lightning Tour

mpuels:
## Description

I'm using SpaCy version 2.0.3. If I don't use the *-operator in the example, Python throws an error message. With the operator it works fine. Also according to the documentation of the function `nlp.disable_pipes()`, it expects one or more strings as arguments and not one argument being a list of strings.

### Types of change
Change in documentation in spaCy 101.


ines:
Ah yes, you're right – thanks for the fix! 👍

PR #1688, opened 2017-12-06
doc: Fix assert statement in Lightning Tour

mpuels:
## Description
Python 3 throws an error message on the original assert statement. Also, according to the Python documentation regarding the assert statement (https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement), `assert` takes at least one argument and at most two. In the two-argument form the second argument is meant as an error message to be displayed when the assertion fails. I don't think this is intended in this case.

### Types of change
Change in documentation in Lighting Tour.

### DIFF
--- a/website/usage/_spacy-101/_lightning-tour.jade
+++ b/website/usage/_spacy-101/_lightning-tour.jade
@@ -188,7 +188,10 @@ p
     pasta = doc[6]
     hippo = doc[8]
     assert apple.similarity(banana) > pasta.similarity(hippo)
-    assert apple.has_vector, banana.has_vector, pasta.has_vector, hippo.has_vector
+    assert apple.has_vector
+    assert banana.has_vector
+    assert pasta.has_vector
+    assert hippo.has_vector
 
 p
     |  For the best results, you should run this example using the


ines:
Thanks a lot! I think this must have happened when we were trying to condense the example without testing it properly. I guess we could also do:

```python
assert apple.has_vector and banana.has_vector and pasta.has_vector and hippo.has_vector
```

But this is definitely getting quite long... From a user's perspective, what do you think looks cleaner and easier to understand? All in one line to save space, or more explicit as one assert statement per line?

mpuels:
Personally, I like to keep lines shorter than 80 characters in production code, so that I can have two files open side by side. But your proposed line

    assert apple.has_vector and banana.has_vector and pasta.has_vector and hippo.has_vector

wouldn't be the longest one in the tutorial, which is (in section "Calculate inline markup on original string")

                output.append(html.format(classes=classes, word=token.text, space=token.whitespace_))

So I think your suggested solution matches the style of the existing documentation and should be fine.

And thanks for the great documentation by the way!

ines:
Yeah, you're right – in my own code, I definitely like to stick to the 80/79 characters as well. I'll just merge your fix – it's fine, PEP8 compatible and I guess it's never wrong to be more explicit in the introduction examples 👍

PR #1708, opened 2017-12-10
Fix Issue 1622

IsaacHaze:
## Description
Fix exception while trying to save the final trained model when training a new model with cli/train.py

### Types of change
Bugfix

## Checklist


ines:
Thanks a lot! 👍

There's currently one test failure on Python 2, and I think it's because `TemporaryDirectory` is only Python 3+... It shouldn't be hard to fix, though – we might even be able to replace it with `pytest`'s `tmpdir` ([see here](https://stackoverflow.com/a/36071167/6400719))?

Issue #1698, opened 2017-12-07
Is it expected behaviour that LIKE_URL also matches on e-mail addresses?


Bri-Will:
For the code below, an e-mail address triggers both matching rules: 

```
import en_core_web_sm
from spacy.matcher import Matcher

nlp = en_core_web_sm.load()
matcher = Matcher(nlp.vocab)

matcher.add('EMAIL', None, [{'LIKE_EMAIL': True}])
matcher.add('URL', None, [{'LIKE_URL': True}])

doc = nlp(u'What happens with an email@address.com and how about a URL like http://www.google.com?')
matches = matcher(doc)

print (matches)
```
Results:
`[(17587345535198158200, 4, 5), (2582013287274679728, 4, 5), (2582013287274679728, 11, 12)]`

Seems like they should be distinct...I didn't think an "@" symbol could be present in a URL...

## Your Environment

* Operating System: Windows
* Python Version Used: 3
* spaCy Version Used: 2.0
* Environment Information:

ines:
Thanks for the report 👍  This is definitely not intentional.

For performance reasons, the `like_url` getter uses simple conditionals instead of regular expressions. This means it should be easy to adjust – maybe even as simple as adding something along the lines of `'@' not in text`. The relevant source is here:

https://github.com/explosion/spaCy/blob/51d3ab21378a298c0c5abcc8fdba1ddf22223cc9/spacy/lang/lex_attrs.py#L76-L95

I'll add a `help wanted` label in case anyone wants to give it a go, adjust the getter and add a few tests for it. Otherwise, we're happy to take care of this, too.

Bri-Will:
FYI, I added an `if` statement like so:
```
if '@' in text:
    return False
```
and it seems to works fine.

I'd love to add the fix, but I'm too much of a newbie to be trusted to do that!

ines:
@Bri-Will Thanks for updating – nice to hear that it's working!

> I'd love to add the fix, but I'm too much of a newbie to be trusted to do that!

Don't worry – of course, you don't *have to*, but if you'd like to submit your first PR, I'm happy to assist! Your fix looks fine and you don't have to worry about adding tests for now (I'll take care of this later).

[...]

PR #1715, opened 2017-12-11
Update lex_attrs.py. Fix like_url from matching on e-mail

Bri-Will:
## Description

Fixes issue #1698

Checks token text for presence of "@" symbol in order to prevent LIKE_URL from matching on e-mail addresses.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/.github/contributors/Bri-Will.md
+++ b/.github/contributors/Bri-Will.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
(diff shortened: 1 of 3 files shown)


ines:
Thank you, this looks good! 👍 I made one small edit and moved the `'@' in text` check up – if the text looks like an email address, we might as well return `False` immediately and not perform other checks, split the text etc.

f11r:
@ines I think the check was actually in the correct place so that the function still returns `True` for URLs like https://en.wikipedia.org/wiki/%22@%22_%28album%29

ines:
@f11r Ah, thanks – I didn't consider cases like this. This makes sense, will revert my edit then 👍 (Sorry for messing with your code, @Bri-Will!)

I'm sure there might still be cases where it doesn't produce perfect results, but I guess we can get away with that – I mean, it's called **like**_url for a reason.

PR #1719, opened 2017-12-13
fix: Add missing period in train data

mpuels:
## Description
The annotations `heads` and `deps` each contain 6 elements and the last element of `deps` is `'punct'`, so I guess there is a period missing at the end of the example sentence.

### Types of change
Fix in example code.

### DIFF
--- a/examples/training/train_parser.py
+++ b/examples/training/train_parser.py
@@ -21,7 +21,7 @@ TRAIN_DATA = [
         'heads': [1, 1, 4, 4, 5, 1, 1],
         'deps': ['nsubj', 'ROOT', 'compound', 'punct', 'nmod', 'dobj', 'punct']
     }),
-    ("I like London and Berlin", {
+    ("I like London and Berlin.", {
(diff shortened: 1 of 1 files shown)


ines:
Thanks a lot – this is interesting, now I'm kind of surprised this example worked before. Especially since we're not actually providing the gold-standard tokenization anywhere 🤔

mpuels:
Just out of interest, what exactly surprised you:

1. That no exception was thrown during training because of the mismatch in number of tokens between training sentence and annotations?
2. That the model "learned" something, despite the mismatch?

ines:
@mpuels Mostly point 1, because I remember issues being reported that were caused by mismatched tokenization (i.e. where the text's tokenization didn't match the tags passed in). I need to go back and check, but a possible explanation would be that in those cases, there were actually **not enough** tags, while in this case, there were **too many**, so spaCy ignored the additional ones. Not sure if this is acceptable behaviour, or a bug. Probably a bug, though – I might open an issue about this, so we can at least investigate it properly.

PR #1724, opened 2017-12-14
doc: Fix minor mistakes

mpuels:
## Description
In the first change I added a missing 'that'.

In the documentation words derived from 'to optimize' are written with a 'z' instead of an 's'. I corrected the two usages that were inconsistent with the rest of the text.


ines:
Thanks a lot! 👍

> In the documentation words derived from 'to optimize' are written with a 'z' instead of an 's'. I corrected the two usages that were inconsistent with the rest of the text.

This is an ongoing issue we still haven't found a perfect solution for... both @honnibal and I normally use British spelling, so in order to keep things consistent, we originally decided to use British spelling for the docs and tutorials as well. However, keeping it consistent turned out to be difficult, because in some cases, British spelling just seems off – especially for words that also in the API (tokenizer, optimizer etc).

PR #1735, opened 2017-12-18
Documentation example fix : token.head needs '==' rather than 'is'

mdda:
(similar change to #1689, it seems).  

## Description

Also the assertion needs to include the case subject==descendant, since it appears possible to be in its own subtree, without being its own ancestor. 

### Types of change

Documentation fix

## Checklist


ines:
Thank you! 👍

PR #1742, opened 2017-12-19
Two corrections in the da lan.

kimfalk:
Corrected what seems to be a automatic translation error in the examples file and added a tokenizer exception, which is used often in Danish. 

### Types of change

change to the documentation.
## Checklist


ines:
Thank you! 👍

PR #1749, opened 2017-12-20
Tune Danish tokenizer to more closely match Universal Dependencies

sorenlind:
## Description

I have made several changes to the Danish tokenizer exceptions. I've added a Danish `punctuation.py`.  The motive for these changes is to more closely match the tokenizer used in the Danish Universal Dependencies. I have written tests.

## Changes to Exceptions
I removed the fellowing because, in addition to being abbreviations, they are also words in Danish:

* bio.
* bot.
* da.
* eg.
* el.
* fem.
* form.
* kat.
* min.
* nat.
* par.
* Pas.
* sen.
* sms.
* sort.
* sø.
* tab.
* ti.
* to.
* øv.

I removed the following because they are specified in the base exceptions:

* d.
* f.
* l.
* p.
* s.
* t.
* v.

To support abbreviations written in lowercase (in-sentence) as well ones written with first letter uppercased (beginning of sentence), I now add both the orgiginal, lowercase `orth` as well as `orth.capitalize()` to the dictionary for most exceptions.

Some exceptions, such as ones that form proper nouns when capitalized are only added using lowercase. One example is "mia." which is short for "milliard" (*billion*) while "Mia." is most likely the proper noun "Mia" followed by a full stop.

Finally I removed they following:
* "" (the empty string, I see no reason to have this as an exception?)
* "K." (I believe this is short for Kelvin but the correct abbreviation is simply *K*)

## Punctuation
The main reason for adding a `punctuation.py` was to make sure words containing dashes (such as *FN-operationen*) are kept as one token. Secondly, the *'s* and *'* indicating possessive are now no longer split from the (proper) noun. This is in line with the tokenization in the Danish Universal Dependencies.

### Types of change

Enhancement.

## Checklist

Edit: Minor cleanup of description.


ines:
Thanks so much for the PR and the detailed explanation! Adding the punctuation rules seems reasonable, too – it's always good to have more language-specific rules, and I like that they're reusing the global ones.

I just merged #1742 which included a change to the Danish exceptions, so this caused a conflict with your PR (sorry – but I figured merging the small change first was a better solution, since it'd otherwise be more difficult to reconcile the changes). 

I just resolved the conflict, but when you have a second, could you look over it one more time to make sure I merged correctly?

sorenlind:
Thanks, @ines! No problem, I will take a look at it and let you know :)

sorenlind:
@ines The change made to the exceptions in #1742 was lost when merging master into da_ud_tokenization. I've added the change back in and I believe we are ready to merge :)

ines:
Thanks and sorry about that – good thing I asked you to look over it again 😉 All checks are green, so merging the PR now!

Issue #1774, opened 2017-12-28
missing lang value in meta.json when pickling zh lang


zqhZY:
when I try to run follow script:
```
import spacy
nlp = spacy.blank("zh", vector=False)
nlp.to_disk("./zh_model")
spacy.load("./zh_model")

```
I got the follow issue:

```
Traceback (most recent call last):
  File "/home/zqh/mygit/Rasa_NLU_Chi-master/spacy_test/load_wv.py", line 4, in <module>
    spacy.load("./zh_model")
  File "/usr/local/lib/python3.6/site-packages/spacy-2.0.5-py3.6-linux-x86_64.egg/spacy/__init__.py", line 19, in load
    return util.load_model(name, **overrides)
  File "/usr/local/lib/python3.6/site-packages/spacy-2.0.5-py3.6-linux-x86_64.egg/spacy/util.py", line 116, in load_model
    return load_model_from_path(Path(name), **overrides)
  File "/usr/local/lib/python3.6/site-packages/spacy-2.0.5-py3.6-linux-x86_64.egg/spacy/util.py", line 144, in load_model_from_path
    meta = get_model_meta(model_path)
  File "/usr/local/lib/python3.6/site-packages/spacy-2.0.5-py3.6-linux-x86_64.egg/spacy/util.py", line 196, in get_model_meta
    raise ValueError(msg % setting)
ValueError: No valid 'lang' setting found in model meta.json
```

when reading the source code, I find  this may because code in file  [lang/zh/__init__.py](https://github.com/explosion/spaCy/blob/master/spacy/lang/zh/__init__.py)  have not created and realized the subclass of  Language.Defaults . 

I also find pull request [#1210](https://github.com/explosion/spaCy/pull/1210) may solve this problem but can`t be merged since other things. So may I pull a request to solve this problem for this issue. 

## Your Environment

* Operating System: ubuntu 14.0.1
* Python Version Used: python 3.6
* spaCy Version Used: 2.0.5
* Environment Information:

zqhZY:
#1775

PR #1775, opened 2017-12-28
fix issue #1774

zqhZY:
## Description
fix No valid 'lang' setting found in model meta.json issue when pickling zh lang model 

### Types of change
bugfix

## Checklist


ines:
Thanks so much 👍 I somehow thought we had fixed this, but looks like we didn't!

PR #1710, opened 2017-12-11
Remove abbreviation for positional plac argument

sorenlind:
## Description

When running `spacy init-model` I get an exception from plac:

```
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/sorenlind/Documents/Developer/spacy-init/spacy/__main__.py", line 31, in <module>
    plac.call(commands[command])
  File "/Users/sorenlind/.virtualenvs/spacy-init/lib/python3.6/site-packages/plac_core.py", line 324, in call
    parser = parser_from(obj)
  File "/Users/sorenlind/.virtualenvs/spacy-init/lib/python3.6/site-packages/plac_core.py", line 133, in parser_from
    parser.populate_from(obj)
  File "/Users/sorenlind/.virtualenvs/spacy-init/lib/python3.6/site-packages/plac_core.py", line 257, in populate_from
    a = Annotation.from_(ann)
  File "/Users/sorenlind/.virtualenvs/spacy-init/lib/python3.6/site-packages/plac_core.py", line 91, in from_
    return cls(*obj)
  File "/Users/sorenlind/.virtualenvs/spacy-init/lib/python3.6/site-packages/plac_core.py", line 78, in __init__
    assert abbrev is None, abbrev
AssertionError: f
```

This appears to be caused by plac not supporting *abbreviations* for positional arguments. So I've removed the abbreviation `f` from the `freqs_loc`argument.

### Types of change

Bug fix.

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #1789, opened 2018-01-03
Fix CLI arguments for init-model

sorenlind:
## Description

I have made the following changes to `init-model`:
* When I removed the abbreviation for positional plac args in #1710, I overlooked the fact that the first argument passed is the name of the CLI argument (`init-model`). This caused all arguments to be shifted by one. I've added a dummy-argument called `_cmd`.
* `read_vectors` has two return values, but only one default value was specified when no word vectors were specified.
* `create_model` checked for the existence of word vectors be calling `len` but this caused an exception when no word vectors were specified.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/cli/init_model.py
+++ b/spacy/cli/init_model.py
@@ -25,7 +25,7 @@ from ..util import prints, ensure_path, get_lang_class
     prune_vectors=("optional: number of vectors to prune to",
                    "option", "V", int)
 )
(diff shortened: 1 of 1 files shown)


ines:
Thanks, this is great! I also like the underscore before the dummy argument – I think we've had it as `cmd` in the other commands, and it's definitely more clear with the underscore. So we should probably adjust this across the other commands as well.

sorenlind:
@ines Cool, I can run through the code and change `cmd` in the other commands tonight.

PR #1793, opened 2018-01-03
Prefix dummy argument names with underscore

sorenlind:
## Description

As suggested in https://github.com/explosion/spaCy/pull/1789#issuecomment-354997281, I have prefixed dummy argument names in the CLI commands with an underscore. This makes it more clear that they are dummy variables.

### Types of change

Enhancement.

## Checklist


ines:
This is amazing, thanks for taking care of that suggestion 👍

(Btw, if the checks fail, it might be because Travis is flakey today... already had two random failures earlier which seem to turn out fine after restarting the build.)

PR #1801, opened 2018-01-04
Don't pass CLI command name as dummy argument

sorenlind:
## Description

Running spaCy from the CLI, specifying a command but omitting one or more required arguments, will print a help text. This help text includes names of required and optional arguments. However, it also print both the command name *and* the name of the dummy variable containing the command name.

For example, calling `spacy package` would result in the following output:
```
usage: spacy package [-h] [-m None] [-c] [-f] _cmd input_dir output_dir
spacy package: error: the following arguments are required: input_dir, output_dir
```

Note that the first line contains `spacy package` but it also lists the variable name `_cmd` as one of the arguments. I think that's confusing as it can lead the user to think he or she needs to specify a command *in addition* to the already specified command. 

This PR completely avoids the dummy variables (`_cmd`) by skipping the first command line argument when passing arguments to the plac-annotated functions. This is done by changing `plac.call(commands[command])` to `plac.call(commands[command], sys.argv[1:])`.

This makes the above command simply print the following:
```
usage: spacy package [-h] [-m None] [-c] [-f] input_dir output_dir
spacy package: error: the following arguments are required: input_dir, output_dir
```

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/__init__.py
+++ b/spacy/__init__.py
@@ -25,4 +25,4 @@ def blank(name, **kwargs):
 
 
 def info(model=None, markdown=False):
-    return cli_info(None, model, markdown)
+    return cli_info(model, markdown)
--- a/spacy/__main__.py
+++ b/spacy/__main__.py
@@ -28,7 +28,7 @@ if __name__ == '__main__':
     command = sys.argv.pop(1)
     sys.argv[0] = 'spacy %s' % command
     if command in commands:
-        plac.call(commands[command])
+        plac.call(commands[command], sys.argv[1:])
     else:
         prints(
             "Available: %s" % ', '.join(commands),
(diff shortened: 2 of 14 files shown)


ines:
Thanks so much for your great work on this – this makes a lot of sense! I remember hacking around this issue and not really getting it to work, so we settled for the `cmd` solution. (I think I only really got good at doing advanced stuff with Plac after writing the Prodigy and its CLI/recipe solution. So it's totally possible that we just didn't figure it out properly back when we re-wrote the spaCy CLI.)

Before I merge, there's one little thing I want to check: I think there might be a few cases where we import and call the CLI functions directly, e.g. in other commands. To work around the dummy argument, we ended up passing in `None` as the first argument. So those calls will need to be adjusted as well.

sorenlind:
Thanks a lot! I ran the tests but didn't check for other calls to those functions. I should have - apologies for that. (Maybe something like pylint could be handy for checking for this sort of problems automatically?) I have now found three calls to the functions and updated them by removing the dummy variable.

Issue #1769, opened 2017-12-27
like_num returns false if a numeric word is not lowercased


fucking-signup:
I am using rule-based matching and noticed that some patterns containing `"LIKE_NUM": True` are not matching phrases, where numeric words are not in lowercase (in my case some sentences start with numeric words that are capitalized).

It could be easily checked:

```python
import spacy
nlp = spacy.load('en')

def like_num(text):
    doc = nlp(text)
    print(doc[0].like_num)

like_num("twenty") # True
like_num("20")     # True
like_num("Twenty") # False
like_num("TWENTY") # False
like_num("tWenTY") # False
```

Maybe `like_num` attribute should check lowercased form of a token?

## Environment
* Operating System: macOS High Sierra
* Python Version Used: 3.6.3
* spaCy Version Used: 2.0.3

ines:
Thanks, good point! The implementation is here:

https://github.com/explosion/spaCy/blob/master/spacy/lang/en/lex_attrs.py#L15-L25

It looks like just changing it to `if text.lower() in _num_words` should work, right? If you want to test this and submit a PR, this would be great 👍

fucking-signup:
@ines sure! Will do within a week or two.

PR #1808, opened 2018-01-08
Fix issue #1769

fucking-signup:
## Description
This PR:
- Fixes issue where `like_num` returned `False` when numeric word was not lowercased (#1769).
- Adds checks for `like_num` in ordinal words lists, where ordinal words were specified but left unused (French, Dutch and Portuguese). Previously, only Danish checked for `like_num` in ordinals words list.
- Updates code example in [Adding Languages](https://spacy.io/usage/adding-languages) in Usage section on the website according to proposed changes.

### Types of change
Bugfix and enhancement

## Checklist

### DIFF
--- a/.github/contributors/fucking-signup.md
+++ b/.github/contributors/fucking-signup.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
+documentation, or any other material posted or submitted by you to the project.
+
+2. With respect to any worldwide copyrights, or copyright applications and
+registrations, in your contribution:
+
+    * you hereby assign to us joint ownership, and to the extent that such
+    assignment is or becomes invalid, ineffective or unenforceable, you hereby
+    grant to us a perpetual, irrevocable, non-exclusive, worldwide, no-charge,
+    royalty-free, unrestricted license to exercise all rights under those
+    copyrights. This includes, at our option, the right to sublicense these same
+    rights to third parties through multiple levels of sublicensees or other
+    licensing arrangements;
+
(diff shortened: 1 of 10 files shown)


fucking-signup:
Hmm. The build has failed on AppVeyor but it is not related to proposed changes. Is there a way to re-run the build process?

ines:
@fucking-signup Thanks a lot for the detailed PR! 🙏  Just restarted the Appveyor build... not sure what went wrong there – sometimes the builds can be randomly flakey.

fucking-signup:
@ines great, thank you! 😊

fucking-signup:
Interesting. Still fails on Windows under Python 2.7, yet no errors related to changes.

ines:
Yeah, this is strange... I also don't think it's related to some weird recursion or other issue caused by the language data imports, because that should fail more consistently? 

I might just try it again tomorrow, because who knows... if it still fails, we'll try and rebuild master, and see if we get the same error there...

fucking-signup:
Could be an issue with AppVeyor's environment with Python 2.7?
There are no issues with Python 3.6 on any platform and no issues with Python 2.7 on Linux.

honnibal:
One thing that can go wrong on Windows Python2.7 but not other environments is that `Path.open` won't necessarily default to a utf8 encoding on that platform, while on the other environments it does. If we have any calls to `Path.open()` without the `encoding` arg explicit, this can cause an error.

However, I can't match that problem up to this error --- I would've expected an encoding bug to trigger an error right away.

I wonder whether we're just on the edge of the Appveyor VM memory limit, and indexing a few extra tests with pytest is just enough to tip things over? This could be exacerbated by the double parametrize, which maybe pytest handles inefficiently?

fucking-signup:
Sure, will try today!

fucking-signup:
@honnibal yup, we are on edge on Appveyor VM. Build on Windows under Python 2.7 just succeeded after I removed all tests but one.

ines:
Thanks for the fix, merging now! 👍 

Sorry this was kinda frustrating – but it's still a good thing the memory limit came up. I guess we'll have to tidy up the tests soon!

fucking-signup:
@ines wait, I didn't fix it haha
Just verified that this was an issue. Fixing now.

ines:
Ahhh, sorry, should have checked this more carefully – I thought you had just simplified the tests. I don't think we need to revert everything in #1808 – I think it'd be fine to just update the tests. I'm pretty confident it'll work as expected, considering the fix is fairly minor.

I think one of the main issues might indeed be the double parametrize – if I remember correctly, this will run the test for all combinations of the available parameters, which is pretty overkill in our case. So we could simplify it to something like this:

```python
@pytest.mark.parametrize('word', ['onze', 'quadragésimo')])
def test_pt_lex_attrs(word):
    assert pt_like_num(word) == pt_like_num(word.upper())
```

What do you think?

fucking-signup:
@ines no problem, I should have expressed myself more clearly :)
I have rewritten tests, removed parametrize altogether and used fixtures as suggested by Matthew. Still didn't work.

fucking-signup:
Just checked your suggestion, fails as well 😞

ines:
Thanks for trying! I think I might know what's going on then: when importing a sub-mobule from the language data, the top-level module is imported as well. The language data can get pretty large (especially with all the lemmatization data etc.), which is why the respective class is usually lazy-loaded only when it's needed.

Here, the tests are importing 7 languages in total, including some pretty large ones. So what we could do instead is lazy-load them within the tests instead (i.e. via `util.get_lang_class()`), mark the tests with `@pytest.mark.slow` and only leave one (e.g. the English one) as a sanity check that runs on each build.

fucking-signup:
@ines great idea!
Will do that.

fucking-signup:
It works! Thank you! 😊 
Created new PR.

PR #1835, opened 2018-01-12
Revert "Fix issue #1769"

fucking-signup:
Reverts explosion/spaCy#1808

At the time of merging, the build issue on Windows (Python 2.7) was not solved. I just removed most of the tests to confirm that there is something with AppVeyor's memory limit and the build has succeeded. 

Although it verifies that proposed changes work for at least one language, I think we need to make tests work for other languages as well.


ines:
See my comment in #1808!

PR #1841, opened 2018-01-15
Update _pos-tags.jade

matatusko:
Labels: docs

really small changes to English tags description, but might help some people while working on projects
1) -PRB- should be -RRB- instead 
2) space gets tagged as _SP, and not SP

## Checklist


honnibal:
Thanks! These small changes are really helpful --- this could've gone unfixed for a long time.

PR #1842, opened 2018-01-15
💫 Update numpy headers

honnibal:
Users on unusual platforms such as PowerPC are having trouble installing due to our outdated numpy headers. This patch tries the simplest thing: updating to the latest. This may be incorrect.

## Checklist


sdmonov:
Regarding the numpy headers I was thinking maybe creating a symlink in setup.py to the currently installed numpy headers under the site-packages should work fine?

Issue #1503, opened 2017-11-07
Pattern matcher regression


emste:
Hey!
I'm working with the Matcher and experienced changed behaviour between v2.0.0a17 and v2.0.0a19. The pattern operator '*' seems to be less greedy than it was in a17:

```
import spacy

from spacy.matcher import Matcher

nlp = spacy.load('en_core_web_sm')

def on_match(matcher, doc, i, matches):
    match_id, start, end = matches[i]
    match = doc[start : end].text
    print(f"on_match: {match}")

matcher = Matcher(nlp.vocab)
matcher.add('MyMatcher', on_match,  [ {'ORTH': 'Hello'}, {'ORTH': 'World', 'OP': '*'} ])

doc1 = nlp('Hello World')
doc2 = nlp('Hello World World World')
matcher(doc1)
matcher(doc2)
# Output for a17
# on_match: Hello World
# on_match: Hello World World World
# Output for a19
# on_match: Hello World
# on_match: Hello World
```

Is this an intended change?

## Env a17
* **spaCy version:** 2.0.0a17
* **Platform:** Linux-4.9.49-moby-x86_64-with-debian-8.9
* **Python version:** 3.6.3
* **Models:** en_core_web_sm, xx_ent_wiki_sm

## Env a19
* **spaCy version:** 2.0.0a19
* **Platform:** Linux-4.9.49-moby-x86_64-with-debian-8.9
* **Python version:** 3.6.3
* **Models:** en_core_web_sm, xx_ent_wiki_sm, de_core_news_sm

wpm:
Looks like [the same thing](https://support.prodi.gy/t/op-does-not-appear-to-be-greedy/181) is happening with the `+` qualifier.

ines:
Should be fixed in #1876!

PR #1876, opened 2018-01-22
Pattern matcher fixes

GregDubbin:
#1503 
## Description
This PR includes fixes and improvements to the pattern matching behavior of spacy.matcher.Matcher.  Specifically, the '*' and '+' operators are not greedy and are inconsistent with the output of standard regular expression libraries.

The PR adds a ADVANCE_PLUS action that behaves as if both ADVANCE_ZERO and REPEAT were the action.  This allows greedy matching to be implemented without backtracking.  Due to the restrictions of the token patterns (i.e. one operator per token), the partial match list can be pruned to never be larger than twice the number of specs in a pattern.

### Types of change

1. Fixed bug in matching state logic preventing '*' from ever being greedy
2. Expanded functionality to allow multiple variable length quantifiers without backtracking 

## Checklist

### DIFF
--- a/spacy/matcher.pyx
+++ b/spacy/matcher.pyx
@@ -8,12 +8,14 @@ from cymem.cymem cimport Pool
 from preshed.maps cimport PreshMap
 from libcpp.vector cimport vector
 from libcpp.pair cimport pair
-from libcpp.unordered_map cimport unordered_map as umap
 from cython.operator cimport dereference as deref
 from murmurhash.mrmr cimport hash64
 from libc.stdint cimport int32_t
 
-from libc.stdio cimport printf
+try:
+    from libcpp.unordered_map cimport unordered_map as umap
+except:
+    from libcpp.map cimport map as umap
 
 from .typedefs cimport attr_t
 from .typedefs cimport hash_t
--- a/spacy/matcher.pyx
+++ b/spacy/matcher.pyx
@@ -361,7 +361,8 @@ cdef class Matcher:
                 if action == PANIC:
                     raise Exception("Error selecting action in matcher")
                 
-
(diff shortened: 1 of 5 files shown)


honnibal:
Hmm, it compiles fine for me but the CI servers are giving:

```
/usr/include/c++/4.8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
```

honnibal:
I think the compile error is coming from the unordered map. We could replace this with the `PreshMap` hash table I use elsewhere in spaCy, but I think it's a bit inconvenient, since it's nice to map directly to a struct, as you do here. `PreshMap` only supports 64-bit keys and values. I'll try adding the compiler flag they suggest.

I wish I could make the overall matching code simpler and better. There are a few block that are fairly repetitive, but it's hard to refactor it into functions without actually making things worse.

GregDubbin:
I believe you are correct.  An alternative is to include a pointer to a MatchEntryC in TokenPatternC data structure, initializing it so that every TokenPatternC in a pattern points to the same MatchEntryC.  This should save some memory and time as we wouldn't have to move to the end of the pattern to check the last match for the pattern.

GregDubbin:
I believe that the test examples in test_issue1450 are incorrect.  Specifically, `('a b b c', 0, 2)` should be `('a b b c', 0, 3)` and `('a b b', 0, 2)` should be `('a b b', 0, 3)`.  I have changed the code accordingly, let me know if you disagree.

honnibal:
Amazing work! And yes I think you're right about that test: I think it matched the incorrect performance.

Okay I think this is good to merge --- a major improvement!

PR #1879, opened 2018-01-23
Add Persian character and symbols

azarezade:
## Description
Add Persian characters and the following:
- `٪` used instead of `%`
- `؟` used instead of `?`
- `﷼` used instead of `$`
- `،` used instead of `,`
- `؛` used instead of `;`

### Types of change
Enhancement

## Checklist


honnibal:
Awesome, thanks!

Issue #1889, opened 2018-01-25
`is_stop` depends on capitalisation


peterroelants:
Spacy's `token.is_stop` seems to be outputting different results depending on the capitalization of the word. For example "the" at the start of the sentence is not recognized as a stop word if it's capitalized.

How to replicate:
```
print(spacy.about.__version__)
nlp = spacy.load('en')
for s in ["The store", "the store"]:
    doc = nlp(s)
    print('\n{}'.format(s))
    for t in doc:
        print('{}\t{}'.format(t.text, t.is_stop))
```

Output is:
```
2.0.5

The store
The	False
store	False

the store
the	True
store	False
```

I suspect that this is not intended behavior? Since words at the start of sentences are usually capitalized, but this doesn't mean that their meaning changes.

## My Environment

* **spaCy version:** 2.0.5
* **Platform:** Darwin-17.4.0-x86_64-i386-64bit
* **Python version:** 3.6.4
* **Models:** en_core_web_sm, en
* Spacy installed via `conda install spacy` with conda version 4.4.7

fucking-signup:
You are right, this is not intended behavior. I will fix it.
Thanks for pointing out!

ines:
See #1891!

PR #1891, opened 2018-01-25
Fix issue #1889

fucking-signup:
## Description
This PR fixes an issue where `is_stop` returned `False` when a stop word was not lowercased (#1889).

**Before this PR:**

```python
import spacy
nlp = spacy.load('en')

def is_stop(text):
  doc = nlp(text)
  print(doc[0].is_stop)

is_stop('the') # True
is_stop('The') # False
```

**After this PR:**

```python
import spacy
nlp = spacy.load('en')

def is_stop(text):
  doc = nlp(text)
  print(doc[0].is_stop)

is_stop('the') # True
is_stop('The') # True
```

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/lang/lex_attrs.py
+++ b/spacy/lang/lex_attrs.py
@@ -136,7 +136,7 @@ def is_lower(string): return string.islower()
 def is_space(string): return string.isspace()
 def is_title(string): return string.istitle()
 def is_upper(string): return string.isupper()
-def is_stop(string, stops=set()): return string in stops
+def is_stop(string, stops=set()): return string.lower() in stops
 def is_oov(string): return True
 def get_prob(string): return -20.
 
--- a/spacy/tests/regression/test_issue1889.py
+++ b/spacy/tests/regression/test_issue1889.py
@@ -1,0 +1,11 @@
+# coding: utf-8
+from __future__ import unicode_literals
+from ...lang.lex_attrs import is_stop
+from ...lang.en.stop_words import STOP_WORDS
+
+import pytest
+
+
+@pytest.mark.parametrize('word', ['the'])
+def test_lex_attrs_stop_words_case_sensitivity(word):
+    assert is_stop(word, STOP_WORDS) == is_stop(word.upper(), STOP_WORDS)


honnibal:
Thanks!

This seems like a good change, but I'll leave it until 2.1.0. We try to avoid making changes that can affect people's pre-trained models on patch releases. If someone has a model that pre-processes with spaCy's `is_stop` attribute, changing the set of words filtered could lead to surprising results.

fucking-signup:
Sure!

bellabie:
Apologies if I'm doing this wrong, but I tried to apply the change `def is_stop(string, stops=set()): return string.lower() in stops` to my own lex_attrs file, but it doesn't seem to fix the issue of stop words only being detected if they are lowercase.

I ended up sticking a `text = text.lower()` as a pre-processing step since that seemed to solve the problem

fucking-signup:
@bellabie hard to tell what's wrong, need more information :)
Can you provide some details such as spaCy version, what were you trying to do and why?

davidchall:
It seems that this PR hasn't yet made it into a release. https://github.com/explosion/spaCy/blob/master/spacy/lang/lex_attrs.py#L147

bellabie:
Apologies I didn't get back sooner.

I meant that since this change was not incorporated, I changed the lex_attrs file as suggested here:https://github.com/explosion/spaCy/pull/1891/files

Although I made the change to the file, the case-sensitivity persisted. That is to say:
`is_stop('the') # True
is_stop('The') # False`

Essentially, I don't think that this change fixes the problem, at least assuming I did it right.

fucking-signup:
@bellabie it does fix the problem.

Firstly, validate that you have applied changes correctly:

```python
from spacy.lang.lex_attrs import is_stop
is_stop('The', set(['the'])) # should be True, if applied correctly
```

Secondly, you have to use a **blank model** for changes to work. Pre-trained models such as `en_core_web_sm` store lexeme attribute values in a vocabulary. They don't call functions in `lex_attrs` file if a lexeme is already stored in the vocabulary. That's why you might assume that proposed changes aren't working.

You could easily validate it as such:

```python
import spacy
nlp_blank = spacy.blank('en')

nlp.vocab['The'].is_stop # False because information about this lexeme is already stored in the vocab
nlp_blank.vocab['The'].is_stop # True because this is not stored in the vocab and it calls is_stop function in the lex_attrs file
```

bellabie:
Ah! I was not using a blank model, you're right.

jdukatz:
@fucking-signup Say my language processing application uses `en_core_web_sm` rather than training its own model. When spacy 2.1.0 is released with this PR included and I upgrade the version in my application, I'm still not going to see the change then, correct? It sounds like spacy needs to release a new version of the pretrained models with updated case sensitive lexeme attributes for this to truly be fixed.

For example:
`>>> nlp.vocab[u'The'].is_stop`
`False`
because this result comes from the lexeme attribute and not the `is_stop` function, this fix will only apply to custom trained models, not pretrained. Am I understanding that correctly?

fucking-signup:
@jdukatz when there is a minor spaCy release, models need to be retrained, you are correct!
If you upgrade to spaCy 2.1.0, you will also need to upgrade your models, as previous models won't be compatible with it. So you will see a change.

As of 2.1.0 alpha, spaCy has already released a new version of pretrained models.
If you are willing to try it, you might install `spacy-nightly` and download new models.

jdukatz:
Got it, thanks. I didn't realize that every minor version came with new pretrained models. One more reason to look forward to v2.1!

Sireus:
> 

@fucking-signup hey, can u send me an e-mail thesireus@proton.me, please :)

Issue #1872, opened 2018-01-22
Tokenizer Exceptions with regex


azarezade:
I'm trying to add Persian language. 
For tokenization, there are some two-words verbs in Persian like "می روم" that are actually one token and should not be tokenize as `["می", "روم"]`.
But I can't use `TOKENIZER_EXCEPTIONS.py`, I need a way to except regular expression in tokenizer.

I tried to add the following lines in `lang/fa/tokenizer_exceptions.py`:
```
_verb = "می\s[آ-ی]*"
TOKEN_MATCH = re.compile("({v})|({u})".format(u=URL_PATTERN, v=_verb)).match
```
and edit Persian class in `lang/fa/__init__.py` as follows:
```
class PersianDefaults(Language.Defaults):
    lex_attr_getters = dict(Language.Defaults.lex_attr_getters)
    lex_attr_getters[LANG] = lambda text: 'fa'
    tokenizer_exceptions = update_exc(BASE_EXCEPTIONS, TOKENIZER_EXCEPTIONS)
    token_match = TOKEN_MATCH
```
But I fails the following assertion in my test:
```
nlp = Persian()
tokenizer = Persian().Defaults.create_tokenizer(nlp)
tokens = tokenizer(u'می روم')
assert len(tokens) == 1
```
There are two tokens, which should be one!

##  Environment
* Operating System: Mac OS X
* Python Version Used: 3.6
* spaCy Version Used: 2

azarezade:
Regarding the second part of my question, "token normalization for Persian", I wonder if **Norm Exceptions** can handle regular expressions instead of look table? 
Then in Persian for example, I can replace any character "ك" with "ک" or character "ي" with "ی", or many other complicated normalizations.

abedkhooli:
Tried using the fa tokenizer for a sample Arabic string. I can see some simple rules implemented, but does Spacy have built-in support for languages like Arabic ( Persian, Hebrew, ...) that can be easily coded as explained [here](https://spacy.io/usage/adding-languages)? 
I am thinking of diacritics, prefixes, suffixes, hamza, al (alif-lam), .. etc.

azarezade:
Can you say some example (in Arabic) of what you need.

abedkhooli:
[...]

PR #1893, opened 2018-01-27
Add Persian language

azarezade:
Add Persian language, including `__init__.py`, `example.py` and `stop_words.py`.

Regarding the issue #1872, although the base tokenizations works either in Persian, but some improvements are needed, as I mentioned there.

## Checklist


honnibal:
Thanks! Sorry I missed your notes in #1872 --- will reply one I get a chance.

For now, I've run a quick benchmark that suggests the current changes don't affect the general tokenization speed. This is nice, because otherwise we'd have to be more careful about having the Persian characters only in the Persian language regex, instead of in the general one.

azarezade:
No problem, that is good!

PR #1905, opened 2018-01-29
Fixed auto linking after download and added simple test to check

### LINKED ISSUE #1904: Dummy command argument wasn't removed from auto-link after download
Auto-linking of downloaded languages was broken in https://github.com/explosion/spaCy/commit/7f0ab145e95036a55af4802184a4b1c496557d0a. The dummy argument wasn't removed from the `link` call at https://github.com/explosion/spaCy/blob/master/spacy/cli/download.py#L44. I can make a PR for the fix unless it's easier for a maintainer to just fix this quickly.

## Your Environment
* Operating System: Docker python:3.6
* Python Version Used: 3.6.2
* spaCy Version Used: Changes made since 2.0.5
* Environment Information: Docker python:3.6

Kimahriman:
Fixes #1904, removing dummy command argument from the link call after a download.

## Description

I added a quick and dirty test that just looks at the console output. Not sure if this is necessary/too fragile if the messages change in the future.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/.github/contributors/Kimahriman.md
+++ b/.github/contributors/Kimahriman.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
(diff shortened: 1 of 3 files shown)


ines:
Thanks a lot! 👍 

> I added a quick and dirty test that just looks at the console output. Not sure if this is necessary/too fragile if the messages change in the future.

Yeah, we might want to take the test out for now – tests that make network requests are always a little problematic, because it means the test suite can't be run without an internet connection and depends on the download links (which are hosted on GitHub and can occasionally time out). We're still thinking about a more satisfying way to test the CLI commands, but it's kinda difficult with the current setup.

Kimahriman:
Yeah I thought that might be the case, wasn't sure how to test it without doing the network request. I'll remove it

ines:
Thanks a lot!

Btw, I just remembered that our internal spaCy build process (the one that we run before each release) actually uses the `download` command to download and run a selection of models to test that everything works correctly. So at least we'll always be able to ensure that it's not fundamentally broken.

PR #1882, opened 2018-01-24
Add norwegian bokmål ('nb') lemmatizer and tag_map

ohenrik:
## Description
Adding support for lemmatization and simple tagging using tag_map

### Types of change
New features to language nb (Norwegian bokmål)

## Checklist
**Will check if boxes bellow as i get closer to completing. Need to start the PR to get help with tag_map. Do not merge yet!** 
- [x] I have submitted the spaCy Contributor Agreement.
- [ ] I ran the tests, and all new and existing tests passed.
- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #1913, opened 2018-01-30
Norwegian Language (nb) - Added french syntax iterator with explanation

ohenrik:
Added syntax iterator to Norwegian Bokmål (nb)

## Description
As suggested [here](https://github.com/explosion/spaCy/pull/1882#issuecomment-361409573) by @ines I added the french syntax iterator as both languages are tagged with universal dependencies.

### Types of change
new feature

## Checklist

### DIFF
--- a/spacy/lang/nb/__init__.py
+++ b/spacy/lang/nb/__init__.py
@@ -13,6 +13,12 @@ from ...language import Language
 from ...attrs import LANG, NORM
 from ...util import update_exc, add_lookups
 
+# Borrowing french syntax parser because both languages use
+# universal dependencies for tagging/parsing.
+# Read here for more:
+# https://github.com/explosion/spaCy/pull/1882#issuecomment-361409573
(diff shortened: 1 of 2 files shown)


ohenrik:
I tried the syntax iterator and it seems to be working :) However I'm far from an expert so some one should probably take a look.

ines:
Thanks a lot! It definitely makes sense that it works – and if it passes the basic sanity checks, we're happy to add it 😊 

One small thing: I normally wouldn't advocate to duplicate code, but in this case, I think it'd make sense to copy-paste the `syntax_iterators.py` and add it to `nb`. Having Norwegian "depend" on French is a little inconvenient and possible too confusing. And if Norwegian does require one or more special-case hacks, it'll make it easier for others to contribute.

ohenrik:
@ines  i now copied the code to the nb language directory.

I was uncertain myself about this, knowing people are really careful about copying code. But i find your reasoning better and more future proof :)

PR #1914, opened 2018-01-30
fix broken link to test suite models

hassanshamim:
Simple fix for broken github link.  It sent you to `https://github.com/explosion/spacy/blob/master/tests/conftest.py` when it should be `https://github.com/explosion/spacy/blob/master/spacy/tests/conftest.py`


ines:
Ah, thanks a lot! 👍

Issue #1928, opened 2018-02-03
Problem with init_model 


azarezade:
I am trying to train a model for Persian (fa) language. I prepared the freqs, brown cluster and word2vec data using the instructions [here](https://github.com/explosion/spacy-dev-resources/tree/master/vocab). Then following the doc, I used:
```
python -m spacy init-model fa data/model/fa/ data/model/fa/fa_wiki.freqs -c data/model/fa/brown/paths -v data/model/fa/fa_wiki.word2vec
```
But it returns error:
```
usage: spacy init-model [-h] [-c None] [-v None] [-V -1]
                        lang output_dir freqs_loc
spacy init-model: error: unrecognized arguments: data/model/fa/fa_wiki.freqs
```
###  Environment
* Operating System: Mac OS X
* Python Version Used: 3.6
* spaCy Version Used: 2.0.5

azarezade:
Sorry, I just found that it is resolved in #1789

azarezade:
No `init-model` works but it returns another error :(
```
Counting frequencies...
923130it [00:03, 302459.31it/s]
Reading vectors...
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/user/Documents/spacy-dev-resources/.venv/lib/python2.7/site-packages/spacy/__main__.py", line 31, in <module>
    plac.call(commands[command])
  File "/home/user/Documents/spacy-dev-resources/.venv/local/lib/python2.7/site-packages/plac_core.py", line 328, in call
    cmd, result = parser.consume(arglist)
  File "/home/user/Documents/spacy-dev-resources/.venv/local/lib/python2.7/site-packages/plac_core.py", line 207, in consume
    return cmd, self.func(*(args + varargs + extraopts), **kwargs)
  File "/home/user/Documents/spacy-dev-resources/.venv/local/lib/python2.7/site-packages/spacy/cli/init_model.py", line 40, in init_model
    vectors_data, vector_keys = read_vectors(vectors_loc) if vectors_loc else None, None
  File "/home/user/Documents/spacy-dev-resources/.venv/local/lib/python2.7/site-packages/spacy/cli/init_model.py", line 87, in read_vectors
    shape = tuple(int(size) for size in f.readline().split())
[...]

Issue #1832, opened 2018-01-12
Error in init_model cli command


jarozajac:
the init_model command fails during the Creating model step with error

> Traceback (most recent call last):
>   File "/home/ec2-user/anaconda3/lib/python3.6/runpy.py", line 193, in _run_module_as_main
>     "__main__", mod_spec)
>   File "/home/ec2-user/anaconda3/lib/python3.6/runpy.py", line 85, in _run_code
>     exec(code, run_globals)
>   File "/home/ec2-user/spaCy/spacy/__main__.py", line 31, in <module>
>     plac.call(commands[command], sys.argv[1:])
>   File "/home/ec2-user/anaconda3/lib/python3.6/site-packages/plac_core.py", line 328, in call
>     cmd, result = parser.consume(arglist)
>   File "/home/ec2-user/anaconda3/lib/python3.6/site-packages/plac_core.py", line 207, in consume
>     return cmd, self.func(*(args + varargs + extraopts), **kwargs)
>   File "/home/ec2-user/spaCy/spacy/cli/init_model.py", line 42, in init_model
>     nlp = create_model(lang, probs, oov_prob, clusters, vectors_data, vector_keys, prune_vectors)
>   File "/home/ec2-user/spaCy/spacy/cli/init_model.py", line 73, in create_model
>     nlp.vocab.vectors = Vectors(data=vectors_data, keys=vector_keys)
>   File "vectors.pyx", line 56, in spacy.vectors.Vectors.__init__
> AttributeError: 'tuple' object has no attribute 'shape'
> 

The problem is, that in init_model.py
> vectors_data, vector_keys = read_vectors(vectors_loc) if vectors_loc else None, None

returns vectors_data as a tuple instead of numpy array, it's caused by 1 of the latest merges adding the second "None" part in the if

The second error in init_model.py is caused by this line:

> if vectors_data:

It produces another exception:

> Traceback (most recent call last):
>   File "/home/ec2-user/anaconda3/lib/python3.6/runpy.py", line 193, in _run_module_as_main
>     "__main__", mod_spec)
>   File "/home/ec2-user/anaconda3/lib/python3.6/runpy.py", line 85, in _run_code
>     exec(code, run_globals)
>   File "/home/ec2-user/spaCy/spacy/__main__.py", line 31, in <module>
>     plac.call(commands[command], sys.argv[1:])
>   File "/home/ec2-user/anaconda3/lib/python3.6/site-packages/plac_core.py", line 328, in call
[...]

PR #1930, opened 2018-02-03
Fix init_model issue

azarezade:
Fixing issue described in #1928 

### Types of change
bug fix

## Checklist


ines:
Thanks so much – looks like this is also the fix for #1832! 👍

Issue #1439, opened 2017-10-18
💫 Which lexical attributes would you like spaCy to support?


ines:
Symbols live in [`spacy/symbols.pyx`](https://github.com/explosion/spacy/tree/develop/spacy/symbols.pyx) and are used to reference attributes internally and externally without using strings. They include strings of the annotation scheme like `POS`, entity labels like `PERSON` or lexical attributes and flags like `IS_ALPHA`.

In spaCy v1.x, adding new symbols was difficult and inconvenient, because it required re-training all models to prevent mismatched integer IDs. This is [not the case anymore](https://alpha.spacy.io/usage/v2#features-hash-ids) in v2.0, since spaCy now uses hash values to generate vocabulary-independent IDs for strings. So we're free to add more symbols and lexical attributes!

💬 **COMMUNITY QUESTION: Is there anything you think is currently missing from the lexical attributes that spaCy should support and allow languages to add functions for?**

Lexical attributes always refer to the [`Lexeme`](https://alpha.spacy.io/api/lexeme), i.e. an **entry in the vocabulary** – not an individual token in context. Token attributes can now be added using [custom attribute extensions](https://alpha.spacy.io/usage/processing-pipelines#custom-components-attributes). Built-in lexical attributes should be **general-purpose flags** that are useful across all languages. (If your application needs very specific flags, you can always add them yourself via [`Vocab.set_flag`](https://alpha.spacy.io/api/vocab#add_flag)!) 

[See here](https://alpha.spacy.io/usage/adding-languages#lex-attrs) for how lexical attributes can be used and overwritten by a `Language` subclass. See [`spacy/lang/lex_attrs.py`](https://github.com/explosion/spaCy/blob/develop/spacy/lang/lex_attrs.py) for the language-independent functions used by spaCy.

The following symbols are currently available:

| Symbol | Description | Example |
| --- | --- | --- |
| `IS_ALPHA` | consists of alphanumeric characters | `token.is_alpha` |
| `IS_ASCII` | consists of ASCII character | `token.is_ascii` |
| `IS_DIGIT` | is a digit | `token.is_digit` |
| `IS_LOWER` | is in lowercase | `token.is_lower` |
| `IS_PUNCT` | is punctuation | `token.is_punct` |
| `IS_SPACE` | is whitespace | `token.is_space` |
| `IS_TITLE` | is titlecase | `token.is_title` |
| `IS_UPPER` | is uppercase | `token.is_upper` |
[...]

PR #1932, opened 2018-02-04
New lexical attribute: is_currency

DuyguA:
As already discussed in  #1439 , I'd like to add is_currency lexical attribute under lang/. 

## Description
I made use of unicodedata.categories, currency symbols have category 'Sc'. [This category](https://www.fileformat.info/info/unicode/category/Sc/list.htm) has 54 currency symbols.

I added the test under tests/lang/test_attrs.py

Note that **is_currency** returns True only for currency symbols. Currency words e.g. dollar, euro, cent; I'll override in lang subhooks.  

I can add an update to [Lexeme API](url) as well. 

### Types of change
New feature.

## Checklist

- [x ] I ran the tests, and all new and existing tests passed.


spacy/attrs.pxd
```diff
@@ -18,6 +18,7 @@ cdef enum attr_id_t:
     IS_QUOTE
     IS_LEFT_PUNCT
     IS_RIGHT_PUNCT
+    IS_CURRENCY
 
     FLAG18 = 18
```
ines: In order to not confuse the order of flags, we should remove this line and replace the next one with:

```python
FLAG19 = 19
```

spacy/attrs.pyx
```diff
@@ -21,6 +21,7 @@ IDS = {
     "IS_QUOTE": IS_QUOTE,
     "IS_LEFT_PUNCT": IS_LEFT_PUNCT,
     "IS_RIGHT_PUNCT": IS_RIGHT_PUNCT,
+    "IS_CURRENCY": IS_CURRENCY,
     "FLAG18": FLAG18,
```
ines: Same here - essentially, the new `IS_CURRENCY` flag replaces `FLAG18`.

spacy/attrs.pxd
ines: In order to not confuse the order of flags, we should remove this line and replace the next one with:

```python
FLAG19 = 19
```

DuyguA:
Ooooopsi, flags messed up 😱
Kapish, I'll make the changes soon and notify from here, I hope that works.

PR #1956, opened 2018-02-08
Website documentation changes Linguistic features, dependency parse

pktippa:
## Description

- [x] Updated table under Navigating Parse tree with new values, see the [ipynb](https://github.com/pktippa/ai-training/blob/master/natural-language-processing/spacy/Linguistic_features.ipynb)
- [x] Pleases update the current codepen from [here](https://codepen.io/pktippa/pen/jZyWom)

Ques: rather than having a static svg content, can we generate this using displacy.js ?

### Types of change
change to the documentation

## Checklist

### DIFF
--- a/.github/contributors/pktippa.md
+++ b/.github/contributors/pktippa.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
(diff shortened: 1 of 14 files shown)


ines:
Thanks a lot! 👍

> Ques: rather than having a static svg content, can we generate this using displacy.js ?

In theory, yes, but we didn't really want to introduce another dependency, especially one that needs to make requests to a server running the models etc. In general, we're trying to keep the website as static as possible. There's [this Jade mixin](https://gist.github.com/ines/b53926f0ab50c135338c31ff2b7f6a00) I made for displaCy which we could use instead of the CodePens. I think the main reason I used the copy-pasted markup is to show examples that look exactly like the ones created by the built-in `displacy` module. But maybe this was a bad idea, I'm not sure!

pktippa:
> In general, we're trying to keep the website as static as possible.
I agree with this.

> There's this Jade mixin I made for displaCy which we could use instead of the CodePens.
Jade mixin is good 👍 
I think we can still continue to use CodePens, in the JS section, we can add the JS file which we are using in [demos displacy](https://demos.explosion.ai/displacy/) CodePen (another idea making the displacy.js node module browserify and give the parsed JSON data to generate SVG?) and give the sentence as input and get the rendered svg as output, so that it will be consistent.

pktippa:
Please verify and merge, get the svg from my codepen and update the same on explosion codepen.

ines:
Thanks, updated and will push the updates to the website in a bit! 👍

PR #1957, opened 2018-02-08
update link to Honnibal and Johnson 2015

mdcclv:
aclweb.org is throwing a gateway timeout on the URL as `https`+`aclweb.org`, but is fine with `https`+`www.aclweb.org` (also with `http`+`aclweb.org`, but let's keep it in `https`, shall we?

Arguably this is an aclweb server configuration issue, but it's quicker to make a PR to spaCy than to fix aclweb!

Checklist:
- [x] I have submitted the spaCy contributor agreement
- [x] Tests pass
- [x] This is a minor documentation change


honnibal:
Thanks!

PR #1968, opened 2018-02-11
New lexical feature is_currency

DuyguA:
Lexical feature for currency symbols €, $...

## Description
New lexical feature
I corrected the mess with attribute IDs. Tests look normal and I checked entity tags once more against any inconvenience. 

```
>>> doc = nlp(u'Apple is looking at buying U.K. startup for €1 billion')
>>> [(t, t.is_currency) for t in doc]
[(Apple, False), (is, False), (looking, False), (at, False), (buying, False), (U.K., False), (startup, False), (for, False), (€, True), (1, False), (billion, False)]
```

### Types of change
New feature

## Checklist


honnibal:
Thanks, looks good! I think this could definitely be useful to people.

DuyguA:
Thanxi:) I also think so. Many people are working in finance area, including myself. I hope everyone benefits.

Issue #1974, opened 2018-02-12
Importing spacy modifies random.seed


emulbreh:
`spacy.__init__` imports `spacy.cli.train` and `spacy.cli.evaluate`, both of which [set a global random seed](https://github.com/explosion/spaCy/blob/master/spacy/cli/evaluate.py#L15). Spacy shouldn't modify the global rng outside of cli commands.

## Info about spaCy
* **spaCy version:** 2.0.7
* **Platform:** Darwin-16.0.0-x86_64-i386-64bit
* **Python version:** 3.6.0
* **Models:** de

PR #1979, opened 2018-02-13
Don't fix random seeds on import

emulbreh:
Fixes https://github.com/explosion/spaCy/issues/1974


honnibal:
This is a good point! I'm sorry that this probably caused you some grief. Thanks for the fix.

PR #1987, opened 2018-02-14
Make span.sent work when only manual / custom sbd

thomasopsomer:
## Description

Hi,

I noticed that the function allowing to retrieve the sentence from a span => `span.sent` works only when the document is parsed. I added the ability to get the sentence using the `sent_start` attribute, by looking on the left and right of the span for the next sentence boundaries.

Don't know if you want that, or if it should belong to a personal `user_span_hooks`, but in case it's here :)

### Types of change

bug fix ?

## Checklist

### DIFF
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -19,6 +19,15 @@ def doc(en_tokenizer):
     return get_doc(tokens.vocab, [t.text for t in tokens], heads=heads, deps=deps)
 
 
+@pytest.fixture
+def doc_not_parsed(en_tokenizer):
+    text = "This is a sentence. This is another sentence. And a third."
+    tokens = en_tokenizer(text)
+    d = get_doc(tokens.vocab, [t.text for t in tokens])
+    d.is_parsed = False
+    return d
+
+
 def test_spans_sent_spans(doc):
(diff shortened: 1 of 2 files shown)


honnibal:
I think there's likely a bug in your implementation: in the C data, the `sent_start` attribute takes the ternary values `(0, -1, 1)` -- which I guess should be changed to an enum...

You're checking for truthiness, which will hit on -1 (not a boundary) and 1 (is a boundary). You want to be checking `== 1`.

Additionally, should we be complaining if the sentence boundaries are unset and `is_parsed==False`? We could do that by checking whether all the values are 0.

thomasopsomer:
Yes I mixed up with the `is_sent_start`, it's fixed.

Maybe a `has_sentence` property on the `Doc` would make sense and could be reused...

honnibal:
Probably a good idea! I'm not sure what to call it though. Maybe `is_sentenced`? `has_sentence` sounds like it should be `False` if the document doesn't have a sentence boundary within it.

ines:
Yeah, as weird as `is_sentenced` sounds, it's probably the best option to keep it consistent with `is_parsed`.

honnibal:
Great --- merging this, will then add the `is_sentenced` property

Issue #1959, opened 2018-02-09
Custom component not executing when calling nlp.pipe


enerrio:
I created a custom component to filter out stop words and punctuation and add it to my pipeline like so:

```python
nlp = spacy.load('en')

punctuations = string.punctuation
stopwords = spacy.lang.en.STOP_WORDS

def clean_component(doc):
    """ Clean up text. Tokenize, lowercase, and remove punctuation and stopwords """
    print("Running cleaner")
    # Remove punctuation, symbols (#) and stopwords
    doc = [tok.text for tok in doc if (tok.text not in stopwords and tok.pos_ != "PUNCT" and tok.pos_ != "SYM")]
    # Make all tokens lowercase
    doc = [tok.lower() for tok in doc]
    doc = ' '.join(doc)
    return nlp.make_doc(doc)

nlp.add_pipe(clean_component, name='cleaner', after='tagger')
print(nlp.pipe_names) # ['tagger', 'cleaner','parser', 'ner']
```

But when I run nlp.pipe on some text "Running cleaner" is printed but the text isn't filtered.

``` python
for doc in nlp.pipe(data['text'][:2]):
    print(doc)
```
The output is the same as the input. Am I using pipe wrong? Thanks.

## Your Environment

* Operating System: MacOS 10.13.3
* Python Version Used: 3.6.4
* spaCy Version Used: 2.0.7
* Environment Information:

honnibal:
Thanks, this is a bug. When component functions don't have a `.pipe()` method, we call a helper function to pipe them, here: https://github.com/explosion/spaCy/blob/master/spacy/language.py#L721

This function should be yielding the result, but is instead yielding the original doc. Here's a minimal hack that should make your code work for now:

```python

def clean_component(doc):
    """ Clean up text. Tokenize, lowercase, and remove punctuation and stopwords """
    print("Running cleaner")
    # Remove punctuation, symbols (#) and stopwords
    doc = [tok.text for tok in doc if (tok.text not in stopwords and tok.pos_ != "PUNCT" and tok.pos_ != "SYM")]
    # Make all tokens lowercase
    doc = [tok.lower() for tok in doc]
    doc = ' '.join(doc)
    return nlp.make_doc(doc)

def pipe_clean(docs, **kwargs):
    for doc in docs:
        yield clean_component(doc)

# Yes, adding attributes to functions works...It's just a bit dirty-looking. Arguably less confusing to
# make it a class. Shrug.
[...]

PR #1988, opened 2018-02-15
Fix Issue #1959

enerrio:
## Description
This PR fixes issue #1959 where calling nlp.pipe won't modify text using a custom component. Here is some example code:

```python 
nlp = spacy.load('en')
texts = ['Apple is looking at buying U.K. startup for $1 billion.']

def clean_component(doc):
    """ Clean up text. Make lowercase and remove punctuation and stopwords """
    # Remove punctuation, symbols (#) and stopwords
    doc = [tok.text.lower() for tok in doc if (not tok.is_stop
                                               and tok.pos_ != 'PUNCT' and
                                               tok.pos_ != 'SYM')]
    doc = ' '.join(doc)
    return doc

nlp.add_pipe(clean_component, name='cleaner', after='ner')
for doc in nlp.pipe(texts):
    print(doc)
```
This is the output before this PR:
``` 'Apple is looking at buying U.K. startup for $1 billion.'```

After this PR:
```'apple looking buying u.k. startup 1 billion'```

### Types of change

Big Fix

## Checklist

### DIFF
--- a/.github/contributors/enerrio.md
+++ b/.github/contributors/enerrio.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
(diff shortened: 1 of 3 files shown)


ines:
Sorry, there was an issue on `master`, so this might be why your PR is currently failing. Will restart the build!

honnibal:
Thanks!

PR #2000, opened 2018-02-18
Polish tag map

jimregan:
## Checklist


jimregan:
I haven't looked at this in a couple of months; I held off the pull request until the animacy fixes went in, but I don't remember what state it was in before that

PR #2036, opened 2018-02-26
Port Japanese mecab tokenizer from v1

polm:
## Description

This brings the Mecab-based Japanese tokenization introduced in #1246 to
spaCy v2. There isn't a JapaneseTagger implementation yet, but POS tag
information from Mecab is stored in a token extension. A tag map is also
included.

As a reminder, Mecab is required because Universal Dependencies are
based on Unidic tags, and Janome doesn't support Unidic.

Things to check:

1. Is this the right way to use a token extension? I couldn't find any other uses of extensions in language-related code.

2. What's the right way to implement a JapaneseTagger? The approach in
 #1246 relied on `tag_from_strings` which is gone now. I guess the
best thing is to just try training spaCy's default Tagger?

### Types of changes

This is a port of the enhancements in #1246.

## Checklist

**Note:** All Japanese-related tests pass. I tried running the full test suite and got a lot of unrelated errors; since I also got the errors running the commit before this I assume it's not related to my changes, but if you find something let me know.


honnibal:
Thanks! It's really nice to have this Japanese support.

Btw, I've been working on a "native" strategy as well, as part of the UD parsing work for CoNLL 2018. spaCy now supports learning to merge tokens. If you create a `GoldParse` object with a `Doc` that has over-segmented, the parser will learn to assign a special label to the tokens to merge.

For Chinese and Japanese, we just have to make the initial tokens the characters, and then the parser learns to merge them, jointly with the syntactic analysis. This has been working pretty well in my CoNLL 2017 tests so far.

> Is this the right way to use a token extension? I couldn't find any other uses of extensions in language-related code.

It looks good to me. I haven't reviewed it *that* closely, but I don't see any obvious problems.

> What's the right way to implement a JapaneseTagger? The approach in #1246 relied on tag_from_strings which is gone now. I guess the best thing is to just try training spaCy's default Tagger?

That method disappeared because you can now just write to `token.tag_`. So, setting the tags is still the best thing to do -- it's just a simple loop now.

polm:
Thanks for the feedback!

> That method disappeared because you can now just write to token.tag_. So, setting the tags is still the best thing to do -- it's just a simple loop now.

This is great! It's so simple I just put it in the tokenizer - since all the work is happening there anyway it was just one extra line. If I should make a `JapaneseTagger` to be consistent with other languages let me know.

> Btw, I've been working on a "native" strategy as well, as part of the UD parsing work for CoNLL 2018.

I've been meaning to take a look at this - haven't had a chance yet but it sounds great. I will try to run the test script you posted in the other issue to give it a look.

honnibal:
Sorry for the hold-up on merging this. Looking forward to trying it out!

PR #2040, opened 2018-02-27
Make token.lemma property return hash instead of unicode

thomasopsomer:
## Description

`token.lemma` wasn't always returning an integer ! When the lemma wasn't set on the c attribute, the `lemma` property calls the `lemmatizer` lookup table but it returned the unicode instead of the hash.

Small fix, but quite annoying bug ^^.

### Types of change

bug fix

## Checklist


honnibal:
Thanks!

Issue #1490, opened 2017-11-03
💫 Language data for Turkish, Croatian and Romanian


ines:
To make it easier for the spaCy community to contribute to new languages, I've started adding language data skeletons, including the language class setup and the minimum amount of data required to make them available to spaCy. In particular, I've been focussing on adding languages that are available via [Universal Dependencies](https://github.com/UniversalDependencies) and licensed under CC BY-SA (permitting commercial use). This will allow us – and other spaCy users – to train language models later on.

**If you speak any of the languages listed below, feel free to contribute!** 🎉 Unfortunately, I don't speak any of them, so I've only added stop words I found in an open-source collection (which can vary in quality – so it's always useful to proofread them) and a few abbreviations or contractions I found online. But I hope that those are useful to give you an idea of what could be included in the tokenizer exception patterns.

| Language |  Source | UD data license |
| --- | --- | --- |
| Turkish | [`spacy/lang/tr`](https://github.com/explosion/spacy/tree/master/spacy/lang/tr) | CC BY-NC-SA, CC BY-SA |
| Croatian | [`spacy/lang/hr`](https://github.com/explosion/spacy/tree/master/spacy/lang/hr) | CC BY-SA |
| Romanian | [`spacy/lang/ro`](https://github.com/explosion/spacy/tree/master/spacy/lang/ro) | CC BY-SA |

## Documentation

* [Adding languages](https://spacy.io/usage/adding-languages)
* [spaCy 101 guide](https://spacy.io/usage/spacy-101) for beginners

DuyguA:
Hello @ines  ,

Native Turkish speaker here :+1:  I went over the stopwords list and made a bit  bad encoding cleaning. Here you can see the pull request: https://github.com/explosion/spaCy/pull/1564

 Tokenizer exception, I'll have a closer look tonight.
Do you need anything else? In case you need assistance, you can always ask me directly.

I'll make an open source morphological analyzer soon anyway, if you're interested  it can be integrated into Spacy.

ines:
@DuyguA Ah, only saw your comment now (after merging and commenting on your PR). And omg, yes, this would be amazing 🎉 If you want to test your morphological analyser with spaCy, a nice way to implement this would probably be to start off with a [custom pipeline component](https://spacy.io/usage/processing-pipelines#custom-components). This lets you test it in an isolated environment – and once it works well, it'll be super easy to integrate into the core library.

Some other ideas:

[...]

Issue #1107, opened 2017-06-06
💫 Idea: Example sentences for testing languages


ines:
When testing models and spaCy's performance in general, it's always nice to have a few text examples handy – especially for the languages you don't speak. We noticed this while testing the new Spanish and multi-language NER models. It's not always easy to find good, representative sentences, and copy-pasting whatever is on the Wikipedia front page that day isn't ideal either.

So here's the idea: For each language, we want to add **a simple list of text examples** that can be imported and processed – for example, to test if a model is working as expected. Ideally, those examples should include different types of sentences, grammatical constructions and named entities. The examples could also be used in spaCy's [test suite](https://github.com/explosion/spacy/tree/master/spacy/tests).

Examples could live with the language's data in `spacy.lang`.

```python
examples = [
    'Apple is looking at buying U.K. startup for $1 billion',
    'Autonomous cars shift insurance liability toward manufacturers',
    'San Francisco considers banning sidewalk delivery robots',
    'London is a big city in the United Kingdom.'
]
```

```python
from spacy.lang.en import examples

doc = nlp(examples[0]) # use one example
print([(t.text, t.pos_, t.tag_) for t in doc])

from spacy import displacy
docs = [nlp(doc) for doc in examples]
displacy.serve(*docs) # visualise all example sentences on one page
```

## ✅ Languages with models (8/8)

- [x] English `en` (can be covered by us)
- [x] German `de` (can be covered by us)
- [x] French `fr` (thanks @Gregory-Howard & @Tpt)
- [x] Spanish `es` (thanks @lucianosilvi)
- [x] Portuguese `pt` (thanks @oximer)
- [x] Italian `it` (thanks @MartinoMensio)
- [x] Dutch `nl` (thanks @redevries)
- [x] Greek `el` (thanks @Eleni170 & @giannisdaras)

## Languages with only tokenizers (29/40)

- [x] Albanian `sq` (thanks @AlbionaHoti)
- [x] Arabic `ar` (thanks @tzano)
- [x] Bengali `bn` (thanks @roshni-b)
- [x] Bulgarian `bg` (thanks @KristiyanVachev)
- [x] Catalan `ca` (thanks @mpuig) 
- [x] Chinese `zh` (thanks @renewang)
- [x] Croatian `hr` (thanks @isaric)
- [x] Danish `da` (thanks @mollerhoj)
- [x] Finnish `fi` (thanks @tjkemp)
- [x] Hebrew `he` (thanks @beneyal)
[...]

Issue #2011, opened 2018-02-21
💫 Participating in CoNLL 2018 Universal Dependencies evaluation (Team spaCy?)


honnibal:
**Update 06/06/2018. Best way to run the CoNLL experiments is:**

```
git clone https://github.com/explosion/spaCy -b develop
cd spaCy
make
./dist/spacy.pex ud-train --help
```

The Conference for Natural Language Learning (CoNLL) 2017 shared task is a great standard for evaluating parsing algorithms. Unlike previous parsing evaluations, CoNLL 2017 is end-to-end: from raw text to dependencies, across many languages. While we missed the 2017 evaluation, I'd like to participate in 2018.

To participate in [CoNLL 2018](http://universaldependencies.org/conll18/), we would need to:

* Adapt tokenizers for to match UD tokenization more closely.

* Add pipeline component for statistical lemmatization, to improve lemmatizer coverage across languages.

* Add pipeline component to predict morphological tags.

* Support joint segmentation and tagging or parsing, for languages like Chinese.

All of these are great goals, regardless of the competition! However, it's a lot of work, especially the tokenization, which really needs speakers of the various languages.

Even if we don't get everything done in time to participate in the official evaluation, it will be a great step for spaCy to publish accuracy figures using the official evaluation software and methodology. This will allow direct comparison against other systems, and make quality control across languages much easier.

What would be really awesome is if we got a few people working on this together, so we could participate as **"Team spaCy"**. Ideally we'd have people taking ownership of some of the [main languages](https://github.com/explosion/spaCy/tree/master/spacy/lang), e.g. French, Spanish, German, Chinese, Japanese etc. It's much easier to work on a specific language that you're well familiar with. The official evaluation will consider all language equally, but I'm okay with having low accuracy on like, Ancient Greek or Dothraki.

The official testing period will run **April 30 to June 26**. However, we can get started right away by working with the CoNLL 2017 data.

To get started, I've made a quick script to run an experiment, which I've been testing on the English data. You can run it by building the `feature/better-gold` branch, and running the `examples/training/conllu.py` script like so:

```
python examples/training/conllu.py en ~/data/ud-treebanks-conll2017/UD_English/en-ud-train.conllu ~/data/ud-treebanks-conll2017/UD_English/en-ud-train.txt  ~/data/ud-treebanks-conll2017/UD_English/en-ud-dev.conllu ~/data/ud-treebanks-conll2017/UD_English/en-ud-dev.txt /tmp/dev.conllu
```

[...]

PR #2080, opened 2018-03-09
Turkish data

DuyguA:
## Description
Referencing issue #1490 , basic Turkish data is here.

### Types of change
I cleaned some bad characters from the stopwords

I made some additions to the stopwords. I added frequently used cases. For instance: `lemma: kimi(some people), cases: kimini(to some people), kiminin(some people's), kiminden(from some people)`. Note that not all cases of the same lemma should go to stopwords. 

I deleted some negated words from the stopwords that might mislead sentiment classifers. For instance `oldugunu(it is that way)` is a stopword, but negated form `olmadigini(not that way)` should not go into stopwords(in my opinion).

I make an `is_num` addition to the lexical attributes.

I added a list of frequently used abbreviations to the tokenizer exceptions.

I also included example sentences as in #1107 . I'll write the translations to this issue.

## Checklist

- [ ]x I have submitted the spaCy Contributor Agreement.


DuyguA:
@Ines requested tokenizer tests for Turkish again in  #1490 . I want to prepare the tests in a separate PR and give special attention. Would that be OK?

ines:
Thank you so much – looks great and a big step towards training a Turkish model! 🎉

Re tests: Yes, that's totally fine! For now, the tests can be fairly simple – for example [something like this](https://github.com/explosion/spaCy/blob/master/spacy/tests/lang/de/test_prefix_suffix_infix.py) for the tokenizer exceptions. They're mostly nice to have as a santiy check, to make sure that we (or others) don't accidentally introduce changes that break Turkish tokenization (also see [this thread](https://github.com/explosion/spaCy/issues/2011#issuecomment-368474523) for changes we're planning for the tokenizer – here, it'd be very important to have decent test coverage).

DuyguA:
Sure. I'll carefully go over the specs and prepare the tests soon as possible.

Issue #2063, opened 2018-03-06
Conjugation tag is being lemmatized: also becomes conjurer


masscognition:
The word "also" becomes "conjurer" when lemmatized. Obviously the conjugation tag is being lemmatized not the word. Try this:

>>> import spacy
>>> nlp = spacy.load('en')
>>> pdoc = nlp("this is also three times also also")
>>>pdoc = nlp(u"this is also three times also also")
>>>for ii in pdoc: 
>>>      print ii.lemma_

You get:

this
be
conjurer
three
time
conjurer
conjurer

## Your Environment

* Operating System:  Ubunutu 16.04 LTS
* Python Version Used: 2.7.12
* spaCy Version Used: 2.0.3
* Environment Information:

DuyguA:
I'm on Ubuntu 17.04 Zesty and Python 2.7.13. I was unable to reproduce the bug I also tried on my CentOS developer instance, both with Python2 and Python3 and unable to produce this bug.

masscognition:
Hmm. It reproduces every time for me. You have slightly different OS and Python versions, I know it is VERY unlikely this is the cause but....

I will try it with 2.7.13 and see if it reproduce. Thx.

DuyguA:
I think I found the cause.

@ines  , would it be possible to check  lang/en/lookup.py ? 

`also` is in adverbs list, but somehow  in lookup.py its lemma is `conjurer' ? :)

ines:
Thanks a lot for the report and fix! 👍 

Looks like the lemmatization table we're using contained this typo, and it was ported over. Very strange. Since `v2.0.3`, we've also improved the way lemmas are set in English (rules based on POS tags first if a tagger is available, then lookup table). This also explains why the weird behaviour wasn't reproducable in the latest version (where spaCy correctly uses the POS tag to determine the lemma).

DuyguA:
I viewed the file history, I don't have a clue how that line came :)
Yesterday I made a  awk + grep to check other lemmas. Rest of the file looks normal, only this line contains a typo somehow.

PR #2095, opened 2018-03-14
Quick typo fix

DuyguA:
Very minor typo fix

## Description
Referencing #2063 , I noticed that  `also` has lemma `conjurer`, I thought it might be a copy paste typo. I mapped it to itself, adverbs has lemma themselves (with some exceptions:) ).

The reason I'm making a one liner pull request is that, "also" is a common adverb and bad lemmatization cause huge shifts at the end.

### Types of change
Very minor change, I only removed one line.
I also made a quick awk to go over adverb and adjective lists and check if any of them exists somewhere else. 

## Checklist


ines:
Thanks so much! ✨

Issue #2096, opened 2018-03-14
issue in lemmatizer


azarezade:
I'm want to implement Persian lemmatizer, but I first tried to understand how English lemmatize works. I wonder why the output of
```
from spacy.lemmatizer import Lemmatizer
from spacy.lang.en import LEMMA_INDEX, LEMMA_EXC, LEMMA_RULES
lemmatizer = Lemmatizer(LEMMA_INDEX, LEMMA_EXC, LEMMA_RULES)
lemmas = lemmatizer(u'corpus', u'noun')
print(lemmas)
```
is `corpu`, which is false! But the output of
```
from spacy.lang.en import English
nlp = English()
doc = nlp(u"corpus")
print([token.lemma_ for token in doc])
```
is `corpus`.

I think **there is an issue** in in `lemmatize` function in `spacy/lemmatizer.py`:
```
def lemmatize(string, index, exceptions, rules):
    orig = string
    string = string.lower()
    forms = []
    forms.extend(exceptions.get(string, []))
    oov_forms = []
    if not forms:
        for old, new in rules:
            if string.endswith(old):
                form = string[:len(string) - len(old)] + new
                if not form:
                    pass
                elif form in index or not form.isalpha():
                    forms.append(form)
                else:
                    oov_forms.append(form)
    if not forms:
        forms.extend(oov_forms)
    if not forms:
        forms.append(orig)
    return list(set(forms))
```
the last lines should be 
```
    if not forms:
        forms.append(orig)
    if not forms:
        forms.extend(oov_forms)
```
to resolve the mentioned problem! I can create a pull request for that if it is true.

Moreover, It seems that the second code snippet, uses `lookup.py`, but the first one uses `lemmatizer` function. Why it doesn't use `_nouns.py` and etc in `spacy/lang/en/lemmatizer` folder?

##  My Environment
* Operating System: Mac OS X 
* Python Version Used: 3.6
* spaCy Version Used: 2.0.9

honnibal:
> lemmas = lemmatizer(u'corpus', u'noun')
> is corpu

Notice that the `lemmatizer.__call__` function also takes morphology keyword arguments. That's why `corpus` gets correctly lemmatized when you pass it through spaCy: the tagger is predicting not only that it's a noun, but that it's singular. This lets us know we can avoid lemmatizing it entirely.

[...]

PR #2109, opened 2018-03-17
Fix lemmatize issue

azarezade:
Referring to issue #2096, for some words that are not plural, like "corpus" the old lemmatize function returns "corpu" as lemma which is false! Changing the last two `if` commands resolves the issue!

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/lemmatizer.py
+++ b/spacy/lemmatizer.py
@@ -106,7 +106,7 @@ def lemmatize(string, index, exceptions, rules):
                 else:
                     oov_forms.append(form)
     if not forms:
-        forms.extend(oov_forms)
-    if not forms:
         forms.append(orig)
+    if not forms:
+        forms.extend(oov_forms)
     return list(set(forms))


honnibal:
Hmm, this will never append the OOV forms though. If the tagger tags corpus as plural, I think 'corpu' is a good guess from the lemmatizer. Now if the tagger *didn't* assign plural, then yes we definitely have a bug here1

azarezade:
@honnibal That is right, but you mean the lemmatize result is different than `lemma_`, which I think use lookup table not the `_nouns.py` and etc in `spacy/lang/en/lemmatizer` folder?

Actually I wanted to the add a lemmatizer for Persian. So, I decide to create a list of singular nouns, and etc for Persian like the files in `spacy/lang/en/lemmatizer`, and a set of rules for making singular out of plural tokens (which in Persian,  similar to English, we just remove some suffix from tokens),  then for example, if a word be in the list of nouns, then we don't apply the rules and it is already a singular noun, otherwise we apply the rules to make it singular.

To have a above mentioned thing I decided to use lemmatize, but it doesn't seems to work like I wanted. Do you have any suggestion?

honnibal:
I updated your thread, with a bit of explanation of how the lemmatizer uses morphology features. I'll close this in the meantime, thanks!

PR #2123, opened 2018-03-21
drop should be a float

iann0036:
## Description
drop should be a float

### Types of change
Doco change

## Checklist


ines:
Ah yes, that's correct – thanks for the attention to detail and submitting the fix! 👍

PR #2128, opened 2018-03-22
WIP: linear regexes and tokeniser optimisations

### LINKED ISSUE #1642: 💫 Improve tokenization, refactor regular expressions and get rid of "regex" dependency
Related issues: #1504, #1623, #1371, #1265, #1235, #1184, #3002

### Problems

* `re.DEFAULT_VERSION` is currently set globally, which is bad and can break other modules. There might also be other unintended side-effects of this (see #1623).
* The non-standard versioning of the module, e.g. `0.1.20110315` vs. `2017.06.23` seems to cause problems in combination with other packages that specify older versions of `regex` (see #1184).
* The tokenizer is getting slower (see #1371), parts of which can likely be tracked down to bad regular expressions and lookbehinds defined in the punctuation rules, and potentially the growing character classes.

We've already done some profiling and the suffix rules definitely seem to be a bigger problem than the prefix and infix rules. The biggest issue seems to be variable-width lookbehinds, caused by character classes which actually consist of multiple-character tokens (like `’’`). This means that they need to be grouped into disjunctive expressions using `|`, rather than character classes using `[]` which only require a set lookup. However, variable-width lookbehinds mean that the tokenizer has to keep backtracking. This is also the reason most regex implementations don't support variable-width lookarounds. Russ Cox (who wrote `re2`) has a [very comprehensive overview of these issues](https://swtch.com/~rsc/regexp/regexp1.html).

In spaCy's case, the `regex` module seem to be less efficient that `re`. If we simplify our regular expressions, we could also have an optional dependency on `re2`, which is even stricter than Python's `re`, but much faster. It can be used via a [wrapper like this one](https://pypi.python.org/pypi/re2/) and will require the system-level dependency to be installed – but if it is, spaCy will be much faster.

### Target

We want the tokenizer to run at around 500,000 words per second.

### Considerations

* Simplifying the tokenization rules may mean that we have to accept *some* regressions. This is okay, as long as it only affects very specific cases. We might also want to adopt a flag system that lets the user enable additional, advanced rules that may result in slower tokenization speed, but higher accuracy (e.g. enabling full emoji and symbols support).

* We initially thought it'd be a good idea to include all alphanumeric characters in the [character classes](https://github.com/explosion/spaCy/blob/master/spacy/lang/char_classes.py) used by the global punctuation rules. The idea here was that even if you're processing English text and come across a non-English alphanumeric character, you'd still want the punctuation rules to work as expected. While this still makes sense for characters like `ä` or `ø`, we should probably re-think this approach for character sets of Hebrew, Russian etc. Instead, those languages should probably define their own rules using and character classes in the respective language data – not globally.

* On the plus side, our [language-specific](https://github.com/explosion/spaCy/tree/master/spacy/tests/lang) and [language-independent](https://github.com/explosion/spaCy/tree/master/spacy/tests/tokenizer) tokenizer tests are pretty good and comprehensive. This should hopefully help a lot with debugging and testing possible solutions.

### Solutions

* Refactor and simplify regular expressions defined in the global and language-specific rules.
* Move classes for alphanumeric characters in Hebrew, Russian etc. to the language-specific data so they're only used when needed.
* Drop `regex` dependency.
* Add support for `re2`.

### Relevant source and helpers

* [`spacy/lang/punctuation.py`](https://github.com/explosion/spaCy/blob/master/spacy/lang/punctuation.py)
* [`spacy/lang/char_classes.py`](https://github.com/explosion/spaCy/blob/master/spacy/lang/char_classes.py)

There's also the (currently experimental and thus undocumented) `profile` CLI command, which profiles the spaCy pipeline to find out which functions and operations take the most time. The command takes a model name or language ID, and an optional input file (JSONL with one `"text"` key per line). If no input is specified, the IMDB dataset provided by Thinc is used.

```bash
python -m spacy profile en
python -m spacy profile en_core_web_sm /tmp/my_input.jsonl
```

RXminuS:
**WIP**: fix #1642 . I'm going to start work fully on this tomorrow but wanted to open a PR straight away to help capture some of the ideas we had around optimising tokeniser performance and provide some guidance towards the final cleanup...since I'm not great at Python.

Target: 500.000 WPS tokenisation...ideally scaling parallel

Changes I'm planning to make:
- [ ] transform postive lookaheads and lookbehinds to make use of capture groups (which are O(n))
- [ ] transform negative lookaheads and lookbhinds to make use of a combination of linearisation techniques I came up with for my go-tokenizer.
- [ ] settle on a schema for defining regexes where we explicitly define the capture group we're looking for. I remember there were 1 or 2 regexes that were hard to transform because they relied on capturing "multiple parts" in between lookaheads/behinds. Also wrapping the existing API/functions so they make use of a capturing groups transparently. https://github.com/explosion/spaCy/blob/15eb54feccfb8dac50f7635ee9802280d8a56354/spacy/tokenizer.pyx#L408
- [ ] use the re2 regex library
- [ ] add some performance tests? --- i might need some help with this?
- [ ] see if I can port my parallelisation and other optimisations to the tokenisers. Might have to rethink some of it since it's written in Go but should be doable.

## Checklist

### DIFF
--- a/spacy/lang/punctuation.py
+++ b/spacy/lang/punctuation.py
@@ -12,12 +12,12 @@ _prefixes = (['§', '%', '=', r'\+'] + LIST_PUNCT + LIST_ELLIPSES + LIST_QUOTES 
 
 _suffixes = (LIST_PUNCT + LIST_ELLIPSES + LIST_QUOTES + LIST_ICONS +
              ["'s", "'S", "’s", "’S"] +
-             [r'(?<=[0-9])\+',
-              r'(?<=°[FfCcKk])\.',
-              r'(?<=[0-9])(?:{})'.format(CURRENCY),
-              r'(?<=[0-9])(?:{})'.format(UNITS),
-              r'(?<=[0-9{}{}(?:{})])\.'.format(ALPHA_LOWER, r'%²\-\)\]\+', QUOTES),
-              r'(?<=[{a}][{a}])\.'.format(a=ALPHA_UPPER)])
+             [r'(?:[0-9])(\+)',
+              r'(?:°[FfCcKk])(\.)',
+              r'(?:[0-9])({})'.format(CURRENCY),
+              r'(?:[0-9])({})'.format(UNITS),
+              r'(?:[0-9{}{}(?:{})])(\.)'.format(ALPHA_LOWER, r'%²\-\)\]\+', QUOTES),
+              r'(?:[{a}][{a}])(\.)'.format(a=ALPHA_UPPER)])
 
 
 _infixes = (LIST_ELLIPSES + LIST_ICONS +


RXminuS:
@ines Is there a benchmark test to get a WordsPerSecond for the Tokeniser? In Go tests & benchmarks are the same thing but not sure how it works in Python? Could only find more unit-like tests.

honnibal:
I've been using this to benchmark multi-processing over the pipeline. If you use `for text in texts: doc = nlp.make_doc(text)` instead of `nlp.pipe()`, it should just profile the tokenizer. You'll probably want to undo the joblib stuff and just call the function directly.

One thing to be careful of when profiling the tokenizer: the vocab matters, because we cache the result for every chunk where all the words are in the vocabulary. For instance, if "but," is seen as a chunk, and we have both "but" and "," in the vocab, then the tokenizer will cache the result. This is nice because we can cache straight to the `TokenC*` array we'll need to contruct --- but we can't cache like this if we have OOV items, or else we grow in memory too much.

So, be careful that these are *not* equivalent: `tokenizer = spacy.blank('en').tokenizer` and `nlp = spacy.load('en_core_web_sm').tokenizer`. The latter should be significantly faster.

```
from thinc.extra.datasets import imdb
import plac
from timeit import default_timer as timer
import tqdm
import spacy
import joblib
from toolz import partition_all

def count_words(model_loc, texts):
    nlp = spacy.load(model_loc)
    n_words = 0
    for doc in nlp.pipe(texts):
        n_words += len(doc)
    return n_words


def main(model_loc):
    train, test = imdb()
    texts = [text for (text, label) in train[:3000]]
    #nlp.disable_pipes(*nlp.pipe_names)
    begin = timer()
    partitions = partition_all(1000, texts)
    executor = joblib.Parallel(n_jobs=3)
    do = joblib.delayed(count_words)
    tasks = (do(model_loc, batch) for batch in partitions)
    word_counts = executor(tasks)
    n_words = sum(word_counts)
    end = timer()
    print(n_words, end-begin, n_words/(end-begin))

if __name__ == '__main__':
    plac.call(main)
```

RXminuS:
Ok, but that's not part of the tokeniser though right? So if I use the same "reset" model between benchmarks I should be fine right?

honnibal:
> Ok, but that's not part of the tokeniser though right? So if I use the same "reset" model between benchmarks I should be fine right?

Yes, using the same reset model between benchmarks is fine.

honnibal:
Hmm. So the plan is to have the bit we want to split off as a group, right? But we're making one long alternating expression, so we'll have a lot of these groups.

@RXminuS how did your Go implementation work? Did you make these multiple expressions, and loop over them?

honnibal:
Closing this to keep things tidy, as things have been moving on.

Still super interested in getting the regexes cleaned up, but I suspect a different trick will apply. I think the best solution will be to simplify the token match regex, and rely on matcher post-processing for URLs.

PR #2135, opened 2018-03-23
bugfix for global-variable-change-in-runtime related issue

### LINKED ISSUE #2133: Few variables was changed globally by mistake in runtime
`spacy.lang.zh.ChineseDefaults.use_jieba` should be `True` by default (default value). But actually it is `False` which is changed by `spacy.cli.ud_train` in runtime (when loading spacy). `spacy.cli.ud_train` also make few other changes by accident, for example: `lang.ja.Japanese.Defaults.use_janome = False`, `random.seed(0)` and `numpy.random.seed(0)`. This will cause bugs which is very difficult to find.

**I will make a PR to fix this bug later.**

## Your Environment

* Operating System: Any
* Python Version Used: Any
* spaCy Version Used: 2.1.0.dev4
* Environment Information: Any

howl-anderson:
This PR try to fix #2133 

## Description

Put global variable change operations to main() method to keep it from auto execute during importing

### Types of change

bug fix

## Checklist

### DIFF
--- a/.github/contributors/howl-anderson.md
+++ b/.github/contributors/howl-anderson.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks, true!

PR #2139, opened 2018-03-23
Finnish numbers

ottosulin:
Added Finnish numbers. Please advise if I didn't get the use of lex_attrs.py correctly :)

### DIFF
--- a/spacy/lang/fi/lex_attrs.py
+++ b/spacy/lang/fi/lex_attrs.py
@@ -1,0 +1,26 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+# import the symbols for the attrs you want to overwrite
+from ...attrs import LIKE_NUM
+
+# check if token resembles a number
+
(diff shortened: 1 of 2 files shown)


ines:
Thanks a lot! The `lex_attrs` look good to me – but I obviously can't really comment on the language side of things, as I can't read Finnish 😛 If you want to test that it all works, you could have a look at how they're integrated into the language class in [`lang/en/__init__.py`](https://github.com/explosion/spaCy/blob/master/spacy/lang/en/__init__.py). You should then be able to test it by checking the token attribute:

```python
from spacy.lang.fi import Finnish

nlp = Finnish()
doc = nlp(u"seitsemän")
assert doc[0].like_num
```

If you like, you could even [add a test](https://github.com/explosion/spaCy/tree/master/spacy/tests) to [`tests/lang/fi`](https://github.com/explosion/spaCy/tree/master/spacy/tests/lang/fi), e.g. something like this:

```python
@pytest.mark.parametrize('text', ['seitsemän', 'kahdeksan'])  # maybe some more complex ones
def test_lex_attrs_like_num(fi_tokenizer, text):
    tokens = fi_tokenizer(text)
    assert tokens[0].like_num
```

If you can think of any potential false positives, a test that checks that those return `False` for `like_num` could also be nice.

Issue #1660, opened 2017-11-28
ValueError: shapes (4,0) and (300,128) not aligned from numpy at Thinc 


zhaow-de:
I need to handle German and English languages with a single application. It worked fine with spaCy 1.8.2, 1.9.0 and 1.10.0 but gets broken with spaCy 2.0.3.

To reproduce the issue:
```
 >>> import spacy
 >>> nlpEN = spacy.load('en')
 >>> nlpDE = spacy.load('de')
 >>> doc = nlpEN('Hello world!')
```
The error messages:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Projects/foobar/.env/lib/python3.6/site-packages/spacy/language.py", line 333, in __call__
    doc = proc(doc)
  File "pipeline.pyx", line 390, in spacy.pipeline.Tagger.__call__
  File "pipeline.pyx", line 402, in spacy.pipeline.Tagger.predict
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/neural/_classes/model.py", line 161, in __call__
    return self.predict(x)
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/api.py", line 55, in predict
    X = layer(X)
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/neural/_classes/model.py", line 161, in __call__
    return self.predict(x)
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/api.py", line 293, in predict
    X = layer(layer.ops.flatten(seqs_in, pad=pad))
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/neural/_classes/model.py", line 161, in __call__
    return self.predict(x)
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/api.py", line 55, in predict
    X = layer(X)
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/neural/_classes/model.py", line 161, in __call__
    return self.predict(x)
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/neural/_classes/model.py", line 125, in predict
    y, _ = self.begin_update(X)
  File "/Projects/foobar/.env/lib/python3.6/site-packages/thinc/api.py", line 372, in uniqued_fwd
    Y_uniq, bp_Y_uniq = layer.begin_update(X[ind], drop=drop)
[...]

PR #2158, opened 2018-03-28
💫 Fix loading of multiple vector models

honnibal:
This patch addresses #1660, which was caused by keying all pre-trained
vectors with the same ID when telling Thinc how to refer to them. This
meant that if multiple models were loaded that had pre-trained vectors,
errors or incorrect behaviour resulted.

The vectors class now includes a .name attribute, which defaults to:
{nlp.meta['lang']_nlp.meta['name']}.vectors
The vectors name is set in the cfg of the pipeline components under the
key pretrained_vectors. This replaces the previous cfg key
pretrained_dims.

In order to make existing models compatible with this change, we check
for the pretrained_dims key when loading models in from_disk and
from_bytes, and add the cfg key pretrained_vectors if we find it.

## Checklist

### DIFF
--- a/spacy/_ml.py
+++ b/spacy/_ml.py
@@ -242,6 +242,10 @@ class PrecomputableAffine(Model):
 
 def link_vectors_to_models(vocab):
     vectors = vocab.vectors
+    if vectors.name is None:
+        raise ValueError(
+            "Unnamed vectors -- this won't allow multiple vectors "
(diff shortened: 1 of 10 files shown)


CYUlysses:
Thanks a lot for the fix!

It might be a rookie's question, but may I ask what the best way to apply this patch to the local module is? Or is it better that I wait for the next general version update?

PR #2184, opened 2018-04-04
Add Danish lemmatizer

mollerhoj:
Add's lemmas for Danish

## Description
This adds lemmas for Danish to spaCy. The dataset (Den store danske SprogTeknologiske Ordbase, STO) has been made available for public with courtesy of The University of Copenhagen. 

### Types of change
Added lemmas for Danish

## Checklist

### DIFF
--- a/.github/contributors/mollerhoj.md
+++ b/.github/contributors/mollerhoj.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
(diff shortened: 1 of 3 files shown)


ines:
Cool, thanks a lot! 👍 

One small thing: I think you forgot to copy the `CONTRIBUTOR_AGREEMENT.md` to `.github/contributors/mollerhoj.md`, so you accidentally filled in the template instead. If you could fix that, that'd be great – aside from that, it all looks good. Also thanks for adding a test!

(Now that we have so much language and lemmatization data for various different languages, maybe we also need to re-think the way we're storing them. Moving it out to a submodule seems weird and inconvenient – but for ~600k lines of lemmatizer data, maybe it is more efficient to store it in a text-based format and then convert it to a binary msgpack file when we compile spaCy. Anyway, mostly thinking out loud – but also open for suggestions if you happen to have opinions on this!)

mollerhoj:
Hey @ines. Sorry about the `CONTRIBUTOR_AGREEMENT.md`, just a silly mistake of course, I just fixed it. Notice that this lemma data -JUST- left academia. No other libraries have it, so that's pretty cool. I agree that it's weird to store this data in python code. But on the other hand, it's really easy to understand what's going on, so my suggestion is to just keep things as-is. :-)

I really love what you guys a doing in general - keep up the good work!

Issue #1480, opened 2017-11-02
Implement Fast-Text vectors with subword features


honnibal:
FastText is a nice tool, and their algorithm makes a lot of sense for word vector learning. For many languages, it's really not sufficient to have one ID per whitespace-delimited word. In languages such as French, Russian, Arabic, Hindi etc, word formation is compositional: stem + affixes. FastText includes character ngram features to handle this.

spaCy should read in not only FastText's word vector table, but also its character ngram weights. This will allow the vectors to be derived for novel words on the fly.

https://github.com/facebookresearch/fastText/issues/328

r-wheeler:
Came here looking for this, great idea.

alexsgit:
Looking for updates

0ruben:
This would be a very cool feature

samitthange:
Came here looking for FastText support. Currently getting killed by Out of Vocabulary words. Would be nice to see this feature.

honnibal:
@samitthange Do you have gradients by which you could train a word representation?

fmontenbruck:
Leaving a comment to second this request

skrcode:
@honnibal How do we model the character n grams in Vectors class? How do we leverage the existing data and key2row objects to store ngram vectors for each word as well?

PR #2247, opened 2018-04-20
1480 - Implement Fast-Text vectors with subword features

skrcode:
Implement Fast-Text vectors with subword features #1480 

## Description

This commit incorporates subword information to get embeddings for words using fasttext.
For words found in the input vocab model, the vector is returned as it is. For out of vocab words, average of all the subword vectors are computed and returned.
ngram range can be given as optional params. If no range is specified, no subwords are used and the word is used as it is. 
This enables simple and easy integration of subword information and is agnostic of the vocab model supplied.

Usage:
```vocab.get_vector(text,minn,maxn)```
## Checklist

### DIFF
--- a/spacy/tests/vectors/test_vectors.py
+++ b/spacy/tests/vectors/test_vectors.py
@@ -23,6 +23,18 @@ def vectors():
         ('juice', [5, 5, 10]),
         ('pie', [7, 6.3, 8.9])]
 
+@pytest.fixture
+def ngrams_vectors():
+    return [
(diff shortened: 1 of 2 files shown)


honnibal:
This is very nice, thanks! Need to think about this a bit more before merging, and will need to update docs. But we definitely want this functionality in the library.

Does FastText store its subword vectors in the main table? I thought they were output separately, but I haven't checked in detail.

honnibal:
Merging this onto develop so we can test it out in `spacy-nightly`!

Issue #2257, opened 2018-04-25
Unexpected lemma for "coding" -> "cod"


aleksandertym:
I think it should be "code" instead of "cod". It happens in more cases e.g. "riding" -> "rid" (should be "ride").

```
import spacy

nlp = spacy.load("en")

print([x.lemma_ for x in nlp("I like coding")])
```
output:

`['-PRON-', 'like', 'cod']`

* **spaCy version:** 2.0.11
* **Models:** en
* **Platform:** Linux-4.13.0-38-generic-x86_64-with-Ubuntu-16.04-xenial
* **Python version:** 3.5.2

bellabie:
I had this same issue for coded --> cod.
Changing _verbs_irreg.py seemed to fix:
`"coded": ("code",),
"coding": ("code",),`

Should I list as a separate issue?

ines:
Fixed on the `develop` branch (see 2ec2192000e99414a7bbbf11c069c92606da4dc5) and will be included in the upcoming `v2.1.x`.

PR #2258, opened 2018-04-25
Improves word lemmatization

aleksandertym:
## Description
Fixes issue #2257 

### Types of change
Bug fix

### Checklist


spacy/lemmatizer.py
```diff
@@ -92,18 +92,32 @@ def lemmatize(string, index, exceptions, rules):
...
     oov_forms = []
+    exp_forms = []
     if not forms:
         for old, new in rules:
             if string.endswith(old):
                 form = string[:len(string) - len(old)] + new
                 if not form:
                     pass
                 elif form in index or not form.isalpha():
+                    exp_forms.append((string[:len(string)-len(old)] , old, form))
```
therealronnie: This won't probably pass the linter. You should use something like Flake8 to make sure the code complies with the style guide. That should be rewritten to:
```
exp_forms.append((string[:len(string) - len(old)], old, form))
```

spacy/lemmatizer.py
```diff
@@ -92,18 +92,32 @@ def lemmatize(string, index, exceptions, rules):
...
+        new_forms = []
+        for exp_form in exp_forms:
+            key = [key for key, value in exceptions.items() 
+                   if key.startswith(exp_form[0]) and key.endswith(exp_form[1]) 
+                   and value[0] == exp_form[2]]
+            if not key:
+                new_forms.append(exp_form[2])
+        if new_forms:
+            forms = new_forms
+
```
honnibal: Could you explain the logic here? I'm not sure this is correct.

ines:
See #2257! 🙂

Issue #2219, opened 2018-04-13
Indexing issue during Token-Token similarity


hunan-rostomyan:
Given these sentences:

1. cat on **the** mat
2. cat on **a** mat

I want to look at the SpaCy [similarity](https://spacy.io/api/token#similarity) scores between every pair of tokens in each of these sentences. The first one works, but I'm getting an error on the second one.

### Code

```python
nlp = spacy.load('en')

tokens = nlp('cat on a mat')

for tok1 in tokens:
    for tok2 in tokens:
        print(tok1, tok2, tok1.similarity(tok2))
```

### Error

```
cat cat 1.0
cat on -0.0020153373
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-133-e2870c8253a9> in <module>()
      3 for tok1 in tokens:
      4     for tok2 in tokens:
----> 5         print(tok1, tok2, tok1.similarity(tok2))
      6 
      7 heat = np.zeros(shape=(len(tokens), len(tokens)))

token.pyx in spacy.tokens.token.Token.similarity()

TypeError: 'spacy.tokens.token.Token' object does not support indexing
```

I've had the same error on several other sentences, so I don't think it's specific to "a". Not sure how to make sense of this inconsistent behavior.

## Your Environment

* Operating System: OS X 10.11.4 (15E65)
* Python Version Used: python 3.5+
* spaCy Version Used: 2.0.9
* Environment Information: Jupyter Notebook running a python3 kernel

knoxdw:
I just encountered the same error, and agree that it looks like a bug. The problem is in lines 128–129 of `token.pyx`, where if  `other` is length 1 there's an attempt to get the `orth` attribute of `other[0]`. I don't understand the full context, but it looks like this expects to unwrap a token from a single-item list, and instead it generates an error on a single-character string.

dsorato:
Got the same error.

alba83:
Got the same error too.

malokroman:
Also got the same error.

knoxdw:
This deserves a bug label, rather than feature labels. The test included with my pull request illustrates the problem.

Briefly, `token1.similarity(token2)` will generate an error if token2 is a single-character word.

PR #2272, opened 2018-04-28
Test and fix for Issue #2219: Token.similarity() failed if single letter

knoxdw:
Fixes bug described in issue #2219; regression test demonstrates the problem.

## Description
Token similarity was failing when the `other` token happened to be a single letter, apparently because case analysis for span didn't take into account that a token could satisfy `len(other) == 1`. 

### Types of change
Since the problem is that `other[0]` generated an error because the "Token object does not support indexing," adding a check for `hasattr("__getitem__)"` is all that's needed.

## Checklist


honnibal:
Thanks!

PR #2282, opened 2018-05-01
Fix code sample for Doc.set_extension

alexvy86:
## Description
The assertion at the end of the previous sample code for `set_extension` fails because `city_getter`  checked if the whole document text matches any of the city names. Now it checks if any of the city names is contained somewhere in the document text.

Ran the old and new code snippets in a Python 3.6.5 environment, with the expected behaviors (assertion fails for old sample code, passes after fix).

I built and ran the tests for the project and got a bunch of errors (most of them of the form `AttributeError: 'Namespace' object has no attribute 'slow'`), but since I didn't touch the actual project code I imagine these are either issues with my environment...

### Types of change
This is a bug fix to sample code **in the documentation**.

## Checklist

### DIFF
--- a/.github/contributors/alexvy86.md
+++ b/.github/contributors/alexvy86.md
@@ -1,0 +1,87 @@
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks! For small changes like this it's okay to rely on the CI stuff. I know that building the library can be a hassle.

PR #2286, opened 2018-05-02
Fix code sample for span.set_extension

alexvy86:
## Description
Same fix as [PR 2282](https://github.com/explosion/spaCy/pull/2282) but for `Span._set_extension` this time, didn't notice that this one had the same bug. And I checked that `Token.set_extension` doesn't need a fix :).

### Types of change
This is a bug fix to sample code **in the documentation**.

## Checklist


ines:
Thank you! 👍

PR #2297, opened 2018-05-04
Fix bug in custom_component_countries_api.py script

ivyleavedtoadflax:
## Description
Fix bug in the custom_component_countries_api.py script.
Set a `default` when setting extension to prevent the following `ValueError`:

```
ValueError: [E083] Error setting extension: only one of `default`, `method`, or `getter` (plus optional `setter`) is allowed. Got: 0
```

### Types of change
Fixes bug in example script.

## Checklist


ines:
Thanks a lot! (spaCy used to be not strict enough about the keyword arguments, so we fixed that in the latest release. Looks like we forgot to adjust the example!)

PR #2305, opened 2018-05-08
Simplify is_config() and normalize_string_keys()

cclauss:
## Description

* Use __in__ to avoid the nested _ands_ and _ors_.
* Dict comprehension directly tracks with the doc string

### Types of change

Simplifications

## Checklist

### DIFF
--- a/spacy/compat.py
+++ b/spacy/compat.py
@@ -109,22 +109,17 @@ def symlink_to(orig, dest):
 
 
 def is_config(python2=None, python3=None, windows=None, linux=None, osx=None):
-    return ((python2 is None or python2 == is_python2) and
-            (python3 is None or python3 == is_python3) and
-            (windows is None or windows == is_windows) and
-            (linux is None or linux == is_linux) and
-            (osx is None or osx == is_osx))
+    return (python2 in (None, is_python2) and
+            python3 in (None, is_python3) and
+            windows in (None, is_windows) and
+            linux in (None, is_linux) and
+            osx in (None, is_osx))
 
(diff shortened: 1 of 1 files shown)


spacy/compat.py
```diff
@@ -109,22 +109,17 @@ def symlink_to(orig, dest):
...
-    return ((python2 is None or python2 == is_python2) and
-            (python3 is None or python3 == is_python3) and
-            (windows is None or windows == is_windows) and
-            (linux is None or linux == is_linux) and
-            (osx is None or osx == is_osx))
+    return (python2 in (None, is_python2) and
+            python3 in (None, is_python3) and
+            windows in (None, is_windows) and
+            linux in (None, is_linux) and
+            osx in (None, is_osx))
```
honnibal: Thanks! This is neat.

spacy/compat.py
```diff
@@ -109,22 +109,17 @@ def symlink_to(orig, dest):
...
     """Given a dictionary, make sure keys are unicode strings, not bytes."""
-    new = {}
-    for key, value in old.items():
-        if isinstance(key, bytes_):
-            new[key.decode('utf8')] = value
-        else:
-            new[key] = value
-    return new
+    return {key.decode('utf8') if isinstance(key, bytes_) else key: value
+            for key, value in old.items()}
```
honnibal: I think this might be a bit dense. The combination of the dictionary comprehension and the ternary conditional makes it fairly tricky, especially if someone's even a little less familiar with Python.

PR #2372, opened 2018-05-25
Update _custom.jade

idealley:
It seems based on the doc and trying out that the `en` or `[lang]` is missing from the `spacy model-init`

## Checklist


ines:
Ah, good point – thanks for the fix! 👍

Issue #2385, opened 2018-05-28
iob_to_biluo converter does not handle multi-word entities


mpszumowski:
The `spacy.gold.iob_to_biluo` function mismatches NER tags for entities longer than one word. The problem seems to lie in `_consume_ent` function which increments `length` only for each I- tag and having encountered L- tag does not treat it as a part of entity. Thus it sets `length` = 1 where len(entity) == 1 token and `length` = len(entity) - 1 where len(entity) >= 2 tokens (in pseudocode, of course).

Another problem I see is that the line: `target = tags.pop(0).replace('B', 'I')` corrupts all labels with a 'B' character, eg. it converts the following NER labels into the form: 
U-BRAWLERS -> U-IRAWLERS
U-BAWLERS -> U-IAWLERS
U-BASTARDS -> - U-IASTARDS
In fact, this sets all labels with a 'B' to U- labels, because the formatted tag I-IAWLER will be compared to 
yet intact tag from the list I-BAWLER.

## How to reproduce the behaviour

You can reproduce the behavior sending to the read_iob (or any other converter from cli.converters) a list of line or lines in a iob (or appropriate) format.

```
def read_iob(raw_sents):
    sentences = []
    for line in raw_sents:
        if not line.strip():
            continue
        tokens = [t.split('|') for t in line.split()]
        if len(tokens[0]) == 3:
            words, pos, iob = zip(*tokens)
        else:
            words, iob = zip(*tokens)
            pos = ['-'] * len(words)
        biluo = iob_to_biluo(iob)
        sentences.append([
            {'orth': w, 'tag': p, 'ner': ent}
            for (w, p, ent) in zip(words, pos, biluo)
        ])
    sentences = [{'tokens': sent} for sent in sentences]
    paragraphs = [{'sentences': [sent]} for sent in sentences]
    docs = [{'id': 0, 'paragraphs': [para]} for para in paragraphs]
    return docs


def iob_to_biluo(tags):
    out = []
    curr_label = None
    tags = list(tags)
    while tags:
        out.extend(_consume_os(tags))
        out.extend(_consume_ent(tags))
    return out


def _consume_os(tags):
    while tags and tags[0] == 'O':
        yield tags.pop(0)


def _consume_ent(tags):
    if not tags:
        return []
    target = tags.pop(0).replace('B', 'I')
    length = 1
    while tags and tags[0] == target:
        length += 1
        tags.pop(0)
    label = target[2:]
    if length == 1:
[...]

PR #2392, opened 2018-05-29
Fix bug in CLI iob and ner converter

mpszumowski:
Fix a bug in iob_to_biluo function in case of labels with a 'b' character and modify _consume_ent for cli.converter to handle tags already in biluo format.

## Description
https://github.com/explosion/spaCy/issues/2385

If tags in BILUO format are passed to the CLI converter, the L-flag goes unprocessed and n-word tokens are read as (n-1)-word tokens. This is because the CLI iob and ner 2json converters both presuppose that NER tags are passed in IOB1/2 format and convert them to BILUO in gold.iob_to_biluo. 

This may not always be the case - especially the conll_ner2json converter is not self-evident to accept specifically IOB format. What is more, the currently suggested workflow of preparing the dataset for the isolated NER pipeline component is to convert any offset NER tags into BILUO using biluo_tags_from_offsets, write a script converting the docs into .iob or .conll format and then use the CLI converter on the BILUO-tagged docs to obtain the json file accepted by CLI train.

This shows that a general-purpose writer that will also convert offset tags straight into json would be handy. Before it's ready I have modified the iob_to_biluo function to accept both IOB and BILUO formats. In the process I have also fixed a bug that would corrupt labels with a 'B' character due to the str.replace('B', 'I') method.

### Types of change
Bug fix

### DIFF
--- a/.github/contributors/mpszumowski.md
+++ b/.github/contributors/mpszumowski.md
@@ -87,7 +87,7 @@ U.S. Federal law. Any choice of law rules will not apply.
 7. Please place an “x” on one of the applicable statement below. Please do NOT
 mark both statements:
 
-    * [ ] I am signing on behalf of myself as an individual and no other person
+    * [x] I am signing on behalf of myself as an individual and no other person
     or entity, including my employer, has or will have rights with respect to my
     contributions.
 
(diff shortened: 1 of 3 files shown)


mpszumowski:
Yes, it's ready

PR #2415, opened 2018-06-07
replace janome with mecab in the documentation and the test

himkt:
This PR is simple document fix.

## Description

In my glance, spaCy now uses MeCab for Japanese Tokenizer.
But the documentation still refers Janome, which is confusing for me.

## __NOTION__
This PR only changes the documentation.
So I create this PR although unittest fails in my environment.
I will further investigate about unittest failing.

### Types of change
change to the documentation

## Checklist

### DIFF
--- a/.github/contributors/himkt.md
+++ b/.github/contributors/himkt.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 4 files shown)


ines:
Thanks! Merging this onto `develop`, since it'll be included in the upcoming version (`v2.0.11` [still uses janome](https://github.com/explosion/spaCy/blob/v2.0.11/spacy/lang/ja/__init__.py)).

PR #2423, opened 2018-06-08
fix wrong documentations

himkt:
Simple PR about sample codes in documentation.
These line in not functions, so I think `return` should not be put.

## Description
Remove unnecessary `return` statements.

### Types of change
change to the documentation

## Checklist


ines:
Thanks, this makes sense! 👍

PR #2433, opened 2018-06-10
💫 Document Cython API

ines:
## Description

This PR adds the most relevant documentation of spaCy's Cython API.

(Todo for when we publish this: rewrite `/api/#section-cython` and `/api/#cython` to `/api/cython#conventions`.)

### Types of change
docs

## Checklist


website/api/_cython/_tokenc.jade
```diff
@@ -0,0 +1,268 @@
...
+        +cell #[code uint32_t]
+        +cell Offset of the rightmost token of this token's syntactic descendents.
+
+    +row
+        +cell #[code sent_start]
+        +cell #[code int]
+        +cell
+            |  Ternary value indicating whether the token is the first word of
+            |  a sentence. #[code 0] indicates a missing value, #[code -1]
+            |  indicates #[code False] and #[code 1] indicates #[code True].
```
thomwolf: I am not 100% sure about this one.
Usually in my experiments, sent_start is 0 for every token that is not at the beginning of a sentence and 1 for each token at the beginning of a sentence **with the exception of the first token of the doc which as it at 0 also.**

honnibal: I think you're right -- will amend that, thanks.

honnibal:
Introduction section:

```
This section documents spaCy's C-level data structures and interfaces, intended for use from Cython. Some of the attributes are primarily for internal use, and all C-level functions and methods are designed for speed over safety --- if you make a mistake and access an array out-of-bounds, the program may crash abruptly.

With Cython there are four ways of declaring complex data types. Unfortunately we use all four in different places, as they all have different utility:

* `class`: A normal Python class.

* `cdef class`: A Python extension type. Differs from a normal Python class in that its attributes can be defined on the underlying struct. Can have C-level objects as attributes (notably structs and pointers), and can have methods which have C-level objects as arguments or return types.

* `cdef struct`: A struct is just a collection of variables, sort of like a named tuple, except the memory is contiguous. Structs can't have methods, only attributes.

*  `cdef cppclass`: A C++ class. Like a struct, this can be allocated on the stack, but can have methods, a constructor and a destructor. Differs from `cdef class` in that it can be created and destroyed without acquiring the Python global interpreter lock. This style is the most obscure.
```

PR #2500, opened 2018-07-02
return string if found in index, on lemmatizer

mgogoulos:
### Types of change
On a complex language such as the Greek there are many rules that affect lemmatization of words. 

In general, nouns ending with 'α' are coming from nouns that end in 'ο', for example

πρόβατα --> πρόβατο   (sheeps)
ζώα --> ζώο    (animals)
δέντρα --> δέντρο (trees)

so a rule as 
    ["α","ο"],
is valid

However, there are nouns ending with 'α' that this is their singular form, example

στόμα (mouth)
βλέμμα (look)  

The above rule will transform them and the new string that will be returned as their lemma will be wrong. 

A really quick solution here would be to check if the words exists on the index, and if so return them without any transformations, aka prioritize the existence of a word on the index for it's POS over any transformations. In this case, the words 'στόμα', 'βλέμα' will be returned correctly since they are found on the _nouns.py set

I believe that otherwise these words have to be added as exceptions, which adds much more work in order to prepare the datasets. 

What do you think? 

## Checklist

### DIFF
--- a/.github/contributors/mgogoulos.md
+++ b/.github/contributors/mgogoulos.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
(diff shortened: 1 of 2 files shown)


honnibal:
Unfortunately I can't merge this --- it's too difficult to tell whether it disrupts the English lemmatization too.

What we need is a frequency-based approach here, so that we can have a better way to tell which lemmatization to prefer. We also need to key the rules by the morphological attributes, so that the lemmatization makes more sense.

We've been thinking about a number of changes to the lemmatization algorithm to improve it, and we hope to release notes on that over the next couple of weeks. If you have availability to test this against Greek, would you like us to keep you in the loop?

mgogoulos:
Agreed, a better solution is needed here indeed.
Whatever changes are introduced I'd like to test them in Greek for sure

PR #2513, opened 2018-07-05
💫 Rule-based NER component

ines:
## Description

This is probably one of the most common pipeline components built by users, so we want to ship a reusable component and factory with the core library. The component takes match patterns, each with an assigned label, finds them in `Doc` and adds them to the `doc.ents`. This makes it easier to combine rule-based and statistical approaches to NER (e.g. use both the model's predictions and a lexicon.) Pattern files can also be shipped with a model package.

### Features

* Accepts both token patterns (one dict per token describing its attributes) or phrase patterns (exact string matches).
* Can be added before or after an existing entity recognizer in the pipeline. If added before, spaCy's NER model will respect the existing entities and the new constraints, which can potentially lead to better accuracy. If added after, you can optionally set `overwrite_ents=True` to overwrite existing entities.
* If matches overlap (e.g. via token patterns with operators), the component will try to find the best possible combination of entities based on the matches. 
* When you save out the `nlp` object, the patterns will be saved as a `.jsonl` file in the model directory. This lets you ship the patterns **with your model package** 🎉
* Bonus: The patterns have the same format as [Prodigy](https://prodi.gy)'s pattern files. So if you're using Prodigy, you can load in your existing pattern files.

### Usage example

Consider a patterns file like this:

```json
{"label": "ORG", "pattern": "Apple"}
{"label": "GPE", "pattern": [{"lower": "san"}, {"lower": "francisco"}]}
```

```python
import spacy

ruler = EntityRuler(nlp).from_disk('patterns.jsonl')
nlp.add_pipe(ruler, before='ner')
```

### Todo

- [ ] Figure out the naming – see [this Twitter thread](https://twitter.com/_inesmontani/status/1010961474532073472)

### Types of change
enhancement

## Checklist


imaurer:
On the develop branch.

Running into a random segmentation fault and have been able to isolate the issue with a dummy example below that corresponds to my real-world example. Not sure if it is a problem with the pattern logic itself.

```python
import spacy
nlp = spacy.load("en_core_web_lg")
ruler = nlp.create_pipe('entity_ruler')
nlp.add_pipe(ruler, before="ner")
ruler.add_patterns([{"label": "PRODUCT", "pattern": [{"lower": "arial"}, {"lower": "("}, {"lower": "ab987"}, {"lower": ")"}, {"lower": "foo"}, {"lower": "bar"}, {"lower": "baz"}, {"lower": "boo"}]}])
nlp("Arial (ab987) foo bar baz boo")
```

Results in:`Segmentation fault: 11`


The library performs normally except when triggering this example. Even the below example works fine:

```python
import spacy
nlp = spacy.load("en_core_web_lg")
ruler = nlp.create_pipe('entity_ruler')
nlp.add_pipe(ruler, before="ner")
ruler.add_patterns([{"label": "PRODUCT", "pattern": [{"lower": "arial"}, {"lower": "("}, {"lower": "ab987"}, {"lower": ")"}]}])
nlp("Arial (ab987) foo bar baz boo")
```

Let me know if you need me to open up a bug ticket.

mr-bjerre:
First of all - very neat! I am gonna refactor my code for sure.

> * Accepts both token patterns (one dict per token describing its attributes) or phrase patterns (exact string matches).

How do you add one or the other matcher? Is that just giving it either a string or a list as `pattern`? I see that phrase matcher can now be done one lowercase as well. How would I then do that here?

> **NEW**: Allow `PhraseMatcher` to match on token attributes other than `ORTH`, e.g. `LOWER` (for case-insensitive matching) or even `POS` or `TAG`.

mr-bjerre:
@ines I get this error inconsistently when running `nlp(text)` where `EntityRuler` in my pipeline.

    ValueError: [E084] Error assigning label ID 39067104 to span: not in StringStore.

How can I check what is wrong? If I get error with source code `A` then I change it to `B` and run it. It might work. I change it back to `A` and then it still works. So I'm clueless here.

mr-bjerre:
And now I also come across this one.

    Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)

Can't really narrow down whats going on. Maybe you can advice on how to investigate? My example is not minimal.

PR #2521, opened 2018-07-06
Add sent property to token

### LINKED ISSUE #2518: Get sentence from token
## Feature description
I cannot find any simple way of extracting the sentence associated with a token. Would it be possible to create a map on the Doc object to ask for the sentence for any given token. 

## Example

```
doc = nlp("Maybe this will work one day. Maybe it won’t")
token = doc[2]
assert token.text == "will"

sentence = doc.get_sent(token)
assert sentence.text == "Maybe this will work one day."
```

This would be really handy (at least for me) related to putting for example entity tokens in perspective of a sentence.

ohenrik:
Adds token.sent property to fetch the sentence the the token is part of.

## Description
This fixes #2518

### Types of change
New feature

## Checklist

### DIFF
--- a/spacy/tests/doc/test_token_api.py
+++ b/spacy/tests/doc/test_token_api.py
@@ -56,6 +56,14 @@ def test_doc_token_api_str_builtin(en_tokenizer, text):
     assert str(tokens[0]) == text.split(' ')[0]
     assert str(tokens[1]) == text.split(' ')[1]
 
+@pytest.fixture
+def doc(en_tokenizer):
+    text = "This is a sentence. This is another sentence. And a third."
+    heads = [1, 0, 1, -2, -3, 1, 0, 1, -2, -3, 0, 1, -2, -1]
+    deps = ['nsubj', 'ROOT', 'det', 'attr', 'punct', 'nsubj', 'ROOT', 'det',
+            'attr', 'punct', 'ROOT', 'det', 'npadvmod', 'punct']
+    tokens = en_tokenizer(text)
+    return get_doc(tokens.vocab, [t.text for t in tokens], heads=heads, deps=deps)
 
 def test_doc_token_api_is_properties(en_vocab):
     text = ["Hi", ",", "my", "email", "is", "test@me.com"]
@@ -162,3 +170,11 @@ def test_is_sent_start(en_tokenizer):
     assert doc[5].is_sent_start is True
     doc.is_parsed = True
     assert len(list(doc.sents)) == 2
+
+def test_tokens_sent(doc):
+    """Test token.sent property"""
+    assert len(list(doc.sents)) == 3
+    assert doc[1].sent.text == 'This is a sentence .'
+    assert doc[7].sent.text == 'This is another sentence .'
+    assert doc[1].sent.root.left_edge.text == 'This'
+    assert doc[7].sent.root.left_edge.text == 'This'
--- a/spacy/tokens/token.pyx
+++ b/spacy/tokens/token.pyx
@@ -350,6 +350,22 @@ cdef class Token:
         def __get__(self):
             return self.c.r_kids
 
+    property sent:
+        """RETURNS (Span): The sentence span that the span is a part of."""
+        def __get__(self):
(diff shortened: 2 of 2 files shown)


honnibal:
Thanks! Really appreciate the PR on this, as we've got a lot of catching up to do getting the v2.1.0a0 release ready :)

There are a couple of small things to fix here -- copy-paste errors from the `Span` method thi is sourced from.

I think it might be better to use @ines 's suggested trick of delegating to `Span.sent` as the implementation of this. This would avoid the code duplication.

spacy/tests/doc/test_token_api.py
```diff
@@ -162,3 +170,11 @@ def test_is_sent_start(en_tokenizer):
...
     doc.is_parsed = True
     assert len(list(doc.sents)) == 2
+
+def test_tokens_sent(doc):
+    """Test token.sent property"""
+    assert len(list(doc.sents)) == 3
+    assert doc[1].sent.text == 'This is a sentence .'
+    assert doc[7].sent.text == 'This is another sentence .'
+    assert doc[1].sent.root.left_edge.text == 'This'
+    assert doc[7].sent.root.left_edge.text == 'This'
```
honnibal: Test looks good! Thanks.

spacy/tokens/token.pyx
```diff
@@ -350,6 +350,22 @@ cdef class Token:
         def __get__(self):
             return self.c.r_kids
 
+    property sent:
+        """RETURNS (Span): The sentence span that the span is a part of."""
+        def __get__(self):
+            if 'sent' in self.doc.user_span_hooks:
```
honnibal: This should be `doc.user_token_hooks`

spacy/tokens/token.pyx
```diff
@@ -350,6 +350,22 @@ cdef class Token:
         def __get__(self):
             return self.c.r_kids
 
+    property sent:
+        """RETURNS (Span): The sentence span that the span is a part of."""
```
honnibal: Should be "that the token is a part of", not "that the span is a part of".

spacy/tokens/token.pyx
```diff
@@ -350,6 +350,22 @@ cdef class Token:
...
+        def __get__(self):
+            if 'sent' in self.doc.user_span_hooks:
+                return self.doc.user_span_hooks['sent'](self)
```
honnibal: Needs to be `user_token_hooks`

spacy/tokens/token.pyx
```diff
@@ -350,6 +350,22 @@ cdef class Token:
...
+            self.doc.sents
+            cdef int n = 0
+            root = &self.doc.c[self.i]
+            while root.head != 0:
+                root += root.head
+                n += 1
+                if n >= self.doc.length:
+                    raise RuntimeError
+            return self.doc[root.l_edge:root.r_edge + 1]
+
```
honnibal: Overall I think the delegation approach is probably better here. It's maybe a bit slower, but probably not a material difference. If it's problematic we'd rather refactor. Something like this:

```
def __get__(self):
    if 'sent' in doc.user_token_hooks:
        # I don't really see why this would be useful, but for completeness...
        return doc.user_token_hooks['sent'](self)
     return doc[self.i : self.i+1].sent
```

ohenrik:
Glad i can help! :)

I have ifixed the copy paste errors and refactored using the suggested solution from @ines

honnibal:
Great! Thanks.

simonkeng:
Hi @honnibal, @ohenrik,  

I'm excited to see that this is a feature now! Thank you all who were working on this. I just have one question about an attribute error I am experiencing. Maybe I don't quite understand how it should be used. 

```python
raw = unicode(textract.process('test-data.pdf'), 'utf-8')
doc = nlp(raw)

for token in doc:
    if 'fragment' in token.text:
        print token.sent

# AttributeError: 'spacy.tokens.token.Token' object has no attribute 'sent'      
```
Any insight on this would be ultra appreciated. Thanks!

PR #2525, opened 2018-07-06
Conll(u)-bio converter

ohenrik:
Converter for the conllu biluo format for ner training. WIP

## Description
I noticed that there where no converter that included biluo ner tags from the conllu format. The current collu format seems to exclude all ner tags or set them to "O" (nothing).

ref:
https://github.com/explosion/spaCy/blob/master/spacy/cli/converters/conllu2json.py#L61
https://github.com/explosion/spaCy/blob/master/spacy/cli/converters/conllu2json.py#L81

So i have for now duplicated the current connlu2json converter and made it include the ner tag (former _2).

This is work in progress and the first commit contains allot of code duplication. Feel free to suggest ways of including this feature in an elegant way.

### Types of change
New feature

## Checklist


ohenrik:
@ines / @honnibal: This works pretty well now, i have used it to train a NER tagger of the converted data with success. However there is of course a lot of duplicate code and I’m not sure if this should be included as feature in SpaCy. Let me know what you think :)

This implementation is set up to use the original `_2` variable from the conllu2json converter and treat it like a BIO tag. It will then convert the BIO tags into BILUO tags and create the document as normal.

Is the `_2` variable normally reserved for something else? It seems like the conllu format has set the last column to be an extension property so i guess it should be ok to use this if the documentation for this converter clearly states where it expects the BIO tag to be located.

The datasett i have used will soon be publicly available, I’ll let you know once it is released so you can take a look at it.

honnibal:
> This implementation is set up to use the original _2 variable from the conllu2json converter and treat it like a BIO tag. It will then convert the BIO tags into BILUO tags and create the document as normal.

Honestly there are so many variations on the CoNLL format that I can't keep track of them well. If the docs say it's reserved for extensions, I agree this seems like a fine use-case!

Issue #2486, opened 2018-06-28
Problems and errors in German lemmatizer


disimone:
## How to reproduce the behaviour

```
import spacy
nlp = spacy.load('de')
test = nlp.tokenizer('die Versicherungen') # The insuranceS
for t in test:
    print(t,t.lemma_)
[output] die der 
[output] Versicherungen Versicherung

test = nlp.tokenizer('Die Versicherungen') # The insuranceS
for t in test:
    print(t,t.lemma_)
[output] Die Die
[output] Versicherungen Versicherung

test = nlp.tokenizer('die versicherungen') # The insuranceS
for t in test:
    print(t,t.lemma_)
[output] die der
[output] versicherungen versicherungen
```

## Your Environment

* **Python version:** 3.5.2
* **Models:** de
* **Platform:** Linux-4.4.0-112-generic-x86_64-with-Ubuntu-16.04-xenial
* **spaCy version:** 2.0.11

Hi all,

I hope the code snippet exemplifies the problem clearly enough.

Basically, I fail to see how the German lemmatization should be used.

Nouns are only lemmatized if they are Capitalized, all other text elements are only lemmatized if they are lower case. So turning all words to lower() means throwing away all nouns lemmas. Trusting the input to have proper capitalization means losing all cases where a non-noun is at the beginning of a sencence (hence not lower case).

How do people actually use this in a real use-case?

Thanks for your help,

Andrea.

DuyguA:
Hello all,

I checked the German lemmatizer file, https://raw.githubusercontent.com/explosion/spaCy/master/spacy/lang/de/lemmatizer.py

Here, the string "versicherungen"    does not exist as a key, hence go unrecognized. 

The thing with German textt that, almost all NLP applications depends on capitals are correctly written. (as for some words writing capital/small can mess up statistical algorithms).

What about this idea @ines , can we add "safe" nouns' small versions to the lemmatizer file as well?


Cheers,
Duygu.

disimone:
Hi @DuyguA,

indeed, my main problem is not with Versicherungen (which exists in the lemmatizer lookup table with with the correct capitalization), but the fact that "Die" is not recognized. In general, every time a verb/adjective/pronoun/article is at the beginning of a sentence, it will not be recognized, because the lemmatizer only knows it in lower-case.

And of course if I lower() everything, I lose all the nouns, as you pointed out. The same if I lower() only the first word of a sentence, since from time to time nouns will be there too...

[...]

PR #2529, opened 2018-07-08
German lemmatizer additions

DuyguA:
##  Quick fix to German lemmatizer capitals issue

I made a quick fix to the issue at #2486 . In this fix I included capitalized versions of the words belonging to the  classes that are likely to occur at the beginning of a sentence. I added capital versions of the words that are risk free i.e. has only one possible lemma. Here are the classes:

Personal pronouns
Relative pronouns
Demonstratives
Indefinite articles
Definite articles
Particles
Auxiliary verbs
Modal verbs

Members of these classes usually have only one possible lemma anyway.

### Types of change
Additions are word-lemma pairs, for instance:

Die:der
Das:der
Welche:welch

- I was very careful not to override any word-lemma pairs that already exists.
- New behavior is as follows: When one of these words is in capital, lemma is the small version i.e. "Die" -> "der". Please see the tests.
- I made a list of the word -lemma pairs, sorted by word categories: https://github.com/DuyguA/spaCy/tree/german_lemmatizer_additions  . Again in the case if a word existed as a key in existing lemmatizer file, old version is not overriden. I tried to disturb the existing word-lemma pairs as less as possible.
- In the case of interrogative sentences,  verbs can come to the beginning of the sentence as well. I'll make a separate PR for a verbs list.

## Checklist

### DIFF
--- a/spacy/tests/lang/de/test_lemma.py
+++ b/spacy/tests/lang/de/test_lemma.py
@@ -7,7 +7,9 @@ import pytest
 @pytest.mark.parametrize('string,lemma', [('Abgehängten', 'Abgehängte'),
                                           ('engagierte', 'engagieren'),
                                           ('schließt', 'schließen'),
-                                          ('vorgebenden', 'vorgebend')])
+                                          ('vorgebenden', 'vorgebend'),
+                                          ('die', 'der'),
+                                          ('Die', 'der')])
 def test_lemmatizer_lookup_assigns(de_tokenizer, string, lemma):
     tokens = de_tokenizer(string)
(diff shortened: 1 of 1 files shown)


honnibal:
Thanks! This definitely makes things better. Obviously the situation is pretty unsatisfying --- I hope we can get a decent lemmatizer in place for German soon.

DuyguA:
I think situation is not bad anyway, but it can be more consistent and efficient. I suggest we open a new issue, discuss discuss thoroughly and make a complete revision. This way I see that results can become inconsistent due to capitalization or umlauts etc...

PR #2537, opened 2018-07-10
lemmatizer: correct lemma for Rang

stefan-it:
## Description

This PR corrects the German lemma form for the word "Rang". Initially, the lemma form was "ringen", which is not correct, because it refers to the verb ("ringen") and not to the noun ("Rang").

### Types of change

The lemma form for "Rang" is corrected to "Rang", see also the [Duden](https://www.duden.de/rechtschreibung/Rang) entry.

## Checklist


ines:
Thanks, this makes sense!

*In theory*, there could be cases where "ringen" would be the correct lemma (like, "Rang er mit seinem Leben?"), but I think it's fair to say that "Rang" &rarr; "Rang" is far more common, especially in the general-purpose texts we want spaCy to be optimised for.

This is also a good case that shows the limitations of the lookup lemmatizer. We're hoping to transition more languages to a rule-based implementation, which would let us handle edge cases better, take part-of-speech tags into account etc.

PR #2549, opened 2018-07-16
Add Japanese stop words

polm:
List created by taking the 2000 top words from a Wikipedia dump and
removing everything that wasn't hiragana.

### Types of change

This is an enhancement, previously there were no Japanese stop words.

## Checklist

- [x] My changes don't require a change to the documentation

### DIFF
--- a/spacy/lang/ja/stop_words.py
+++ b/spacy/lang/ja/stop_words.py
@@ -1,0 +1,49 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+# This list was created by taking the top 2000 words from a Wikipedia dump and
+# filtering out everything that wasn't hiragana. ー (one) was also added.
+# Considered keeping some non-hiragana words but too many place names were
+# present.
+STOP_WORDS = set("""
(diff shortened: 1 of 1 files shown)


ines:
Thanks a lot, this is great! 👍

One quick question: Since the words are sourced from Wikipedia, does the list include any words that are very frequent but *do not* have low semantic information? For instance, a word like "person" may be super frequent, but spaCy wouldn't define that as a stop word (other lists handle this differently, but we often see the mass exclusion of frequent words as problematic, especially if combined with machine learning approaches).

polm:
Ah, good point, I should have made it clearer - this doesn't include anything I'd consider a content word. While the base list was the top 2000 words, limiting it to hiragana words took it down to 153. Common hiragana words are mostly conjunctions, determiners, pronouns, etc. 

I'm not sure where to draw the line for "low semantic information" but I would consider this a pretty conservative list.

ines:
Ah perfect, that's exactly what I wanted to know!

PR #2558, opened 2018-07-17
Greek language optimizations

giannisdaras:
### Types of change
This pull request contains various improvements for Greek language support that was very recently added.

1. Stop words list improvements. 
 It improves the quality of stop-words by removing tokens that are not in Greek dictionaries or are not commonly used. There are also additions to the list regarding missing commonly used words. The stop words are acquired by counting token frequencies in large corpus and by cross-validation of the results with other sources.
2. Norm exceptions added.
 A list of over 2500 norm exceptions is added to Greek language model. The collection of the words involves parsing Greek dictionaries.
3. Update of tag map. 
The general tag map is updated so it can match the Greek UD treebank annotation that can be found [here](https://github.com/UniversalDependencies/UD_Greek-GDT). There are models trained on these datasets contaning Vocabulary, Syntax (POS and DEP taggers), Vectors (and soon Entities), and they are going to be uploaded soon.
4. New approach to lemmatizer.
 Till now the lemmatizer was based on a lookup table. This pull request changes the approach by establishing rule-based lemmatization, following the English language standards. It contains collections of lemmas for every pos tag available in the language tag map and exceptions for tokens in each of the pos tags categories. The lemmas are acquired by parsing Greek wiktionary. The rules and the exceptions are specific to each category (different for noun, verbs, adverbs, etc) and are written after careful examination of the language morphological features.
5. Extension of lexical attributes script.
 Improves like_num function.
6. Regex expressions for Greek alphabet are added to the char_classes file.
7. Init file is updated so model can support all the functionalities mentioned above.

## Checklist


honnibal:
Thanks so much for this! Greek definitely poses some difficult challenges, and you've taken some very thoughtful approaches. I think the general strategies might improve some other languages as well.

Merging!

PR #2560, opened 2018-07-17
Fix venv command examples

x-ji:
## Description

The documentation refers to `venv`, which is native to Python3. However, the command examples are as if they were still using `virtualenv`, which is a package independent of `venv`:
- `venv` doesn't need to be installed via `pip`. In fact `pip install venv` would
return an error.
- The correct way to invoke `venv` is `python3 -m venv`, not `venv`, which would
return command not found.

See https://docs.python.org/3/library/venv.html

I suspect that at some point the documentation simply replaced all occurrences of `virtualenv` with
`venv`. However they are different modules and are used differently. The documentation should either stick with `virtualenv` or correctly use `venv`, which is what this PR tries to do.

### Types of change
Change to the documentation.

## Checklist

(I'm not sure what "or if they do, I've added all required information." refers to. I have documented why I feel it necessary to change the relevant commands in the documentation.)

### DIFF
--- a/.github/contributors/x-ji.md
+++ b/.github/contributors/x-ji.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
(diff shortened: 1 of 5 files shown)


x-ji:
Eh I think I put it in here already?: https://github.com/explosion/spaCy/pull/2560/files#diff-593d0bd81635de6489438e03e51181ce

ines:
Thanks a lot, really appreciate the attention to detail! I'm pretty sure that's what happened and we just didn't adjust the commands correctly. In general, we definitely want to standardise on Python 3, but provide fallbacks for Python 2-specific stuff for users who are still working on older code bases.

> Eh I think I put it in here already?

Yes, that's all good! 👍 Sorry, our bot isn't very clever. It's currently only checking the existing repo and not the files included in the PR, because that ended up being tricky to get right. (I might look into this again, though!)

PR #2596, opened 2018-07-25
Fixed typo

AndriyMulyar:
Changed 'The index of the first character after the span.' to 'The index of the last character after the span' in description of doc.char_span under field 'end'.

## Checklist


ines:
Ah, damn, that must have been a copy-paste mistake! Thanks so much for spotting and fixing this 👍

PR #2598, opened 2018-07-25
ammended the Case to the standard model

DimaBryuhanov:
In the traditional Russian grammer there are:

1. Nominative        - usually the Subject of a sentence, thus perceived as the initial form of a noun
2. Genitive              - used like:    "There's no...          + noun in Genitive case
3. Dative                 - used like:    "I'm giving to...       + noun in Dative case
4. Accusative          - used like:    "I'm blaming...        + noun in Accusative case
5. Instrumental       -used like:    "I'm creating with... + noun in Instrumental case
6. Prepositional      - used like:   "I am thinking of...   + noun in Prepositional case

cases. There are 2 known obsolete cases, they are: the Vocative case & the Pair case.

The Vocative case was used when addressing to a person or God with a question or a request. For example:
#from Pushkin:
"Что тебе надобно, старче?" = "What do you need, old man?" - a magic Goldfish asked the old man who initially caught and let it go and afterwards came with requests to the magic fish from his greedy wife.
Here, the last word is in the Vocative case. In modern Russian it only remained in prayers to God: ("Господи, Иисусе Христе,..." = "Lord, Jesus Christ,...").
Without this Vocative case it would be Nominative case, like: ("Господь, Иисус Христос,...")

The Pair case was used untill the XVth century for items that usually exist in pairs, such as: eyes, ears, knees, shoulders, horns, sleeves, etc.
In modern Russian the plural form of these words is what was once the Pair Case: "уши, глаза, колени, плечи, рога, рукава и т.д."

## Checklist

### DIFF
--- a/spacy/lang/ru/lemmatizer.py
+++ b/spacy/lang/ru/lemmatizer.py
@@ -139,15 +139,13 @@ def oc2ud(oc_tag):
             'perf': 'Perf',
         },
         'Case': {
-            'ablt': 'Ins',
             'accs': 'Acc',
             'datv': 'Dat',
-            'gen1': 'Gen',
-            'gen2': 'Gen',
-            'gent': 'Gen',
-            'loc2': 'Loc',
+            'gen': 'Gen',
+            'instr': 'Ins',
             'loct': 'Loc',
             'nomn': 'Nom',
+            'pair': 'Pair',
             'voct': 'Voc',
         },
         'Degree': {


honnibal:
@DimaBryuhanov Do these cases match the scheme in the Universal Dependencies data? That's the treebank we're basing off, and they have a single scheme across languages. So we need the features to match that data.

DimaBryuhanov:
@honnibal, the names of the cases are specific to Russian Language and may be similar with western slavic languages, such as Ukranian & Belorussian, however they slightly differ from many other languages, since the other languages do not use so many cases. However, the category 'Case' existed before the branch that I suggested. You cannot possibly have one scheme for all the languages, since there are substantial differences in their grammars. If spaCy describes and learns the language it should adapt to its individual pecularities, since the language will not adapt to the needs of spaCy. Otherwise, spaCy will never be precise trying to learn something imaginary universal, but not Natural Languages. Ask Ines to tell you about the Korean language experience. Please, let me know.

Even though English, German and Russian are all Indo-European languages and all have such category as Case:

English has 3 cases, mainly illustrated by pronouns:
Nominative = I
Dative-Accusative-Genitive = me
Posessive mine

German has 4 proper cases
Nominative
Genitive
Dative
Accusative

Russian has the German scheme with two additional cases:

Nominative
Genitive
Dative
Accusative
Instrumental
Prepositional

And two obsolete cases, that still have rudiments in the modern language in several words:

Vocative case - used in prayers only now
Pair case - typical for very ancient languages.

Since some time ago Germanic & Slavic tribes were one tribe and spoke one language, it is possible that some time ago all the three languages above had the same grammar and vocabulary, but due to living separately for several thousand years they all simplified but to different extent. 

However, the three languages above beling to one language family and are quite close to each other in the roots of the words. But, compared to other language families they can be very different. Compare them to Arabic or Chinese and you will see.

If you try to find 6 cases in English your model will fail looking for them. However, if you try to describe Russian with only 3 cases you will be very confused either, since the remaining 3 cases will have to become exceptions, that are not treated properly.

If you call the cases with unfamiliar words, like 2nd Dative, how will the people who mark the tokens in prodigy know what to mark? This will cause confusion and mistakes, since some people will use it for Instrumental case, some for prepositional and some will skip it, since they've never heard of it.

Since every language library is imported separately, there's nothing bad in slight differences in the grammar models. But this way every language is described according to its natural live pecularities.

Please, let me know.

DimaBryuhanov:
Dear Mathew @honnibal, thank you for your message.
It does not precisely match the scheme. I decided to withdraw my request for a branch.

PR #2599, opened 2018-07-26
Feature/span ents

### LINKED ISSUE #2589: Feature: `Span.ents` property similar to Doc.ents
It would be nice with a `Span.ents` property like the one on Doc
https://github.com/explosion/spaCy/blob/master/spacy/tokens/doc.pyx#L411

Calling `sent.as_doc().ents` seems to run the entity matcher again and removes custom attributes like `my_entity._.is_special`.

So at the moment you have to manually sort through Doc.ents based on the sent.start and sent.end properties to get the relevant ents for a given sentence correctly. This is not optimal, and a Sent.ents does seem natural :)

ohenrik:
Adds ents property to Span objects. This fixes #2589 

### Types of change
New Feature 

## Checklist

### DIFF
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -137,20 +137,27 @@ def test_span_as_doc(doc):
 def test_span_ents_property(doc):
     """Test span.ents for the """
     doc.ents = [
+        (doc.vocab.strings['PRODUCT'], 0, 1),
         (doc.vocab.strings['PRODUCT'], 7, 8),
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks! Will keep this open for now so we don't forget the docs change.

ines:
I'll just go ahead and merge this now – when we do the release notes, I usually also go through the enhancements to make sure all new stuff is documented. So we'll just add the respective docs then.

PR #2613, opened 2018-07-28
Add abbreviations from UD_Swedish-Talbanken

EmilStenstrom:
## Description
This is part of #2608. It reduces the tokenization errors from 216 -> 196. The remaining errors has to do with token splitting, I'll handle that in a separate PR. All added abbreviations are from UD_Swedish-Talbanken.

### Types of change
Bug fix.

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

Issue #2608, opened 2018-07-27
Improve tokenization of Swedish based on UD_Swedish-Talbanken


EmilStenstrom:
This is a followup for #2578 for addressing the tokenization of Swedish. I plan to fix each of this issues in separate PR:s, but thought I'd summarize all of them in one issue.

## How to reproduce the behaviour
I've [downloaded UD_Swedish-Talbanken](https://raw.githubusercontent.com/UniversalDependencies/UD_Swedish-Talbanken/master/sv_talbanken-ud-dev.conllu), and reconstructed the orignal text from those files, run SpaCy on them, and compared the result. [Code available as a gist](https://gist.github.com/EmilStenstrom/dc67b8909ce9b5ea2c81153f57574c22).

It did the analysis on the sentence level, and marked that sentence as "incorrect" when I found the first mismatching token. So there are probably more mismatching tokens, but matching them up when an earlier one was too much work. The results look like this:

**Total sentences**: 6029
**Correct sentences**: 5488 (91.03% where correctly tokenized)
**Incorrect sentences**: 538

### ERROR_TYPES.DASH: 253
These errors occur when "EEC-apparaten" is tokenized as "EEC", "-", "apparaten". There are a lot of errors like this. 

### ERROR_TYPES.ABBR: 219
These errors occur when there are Swedish abbreviations that SpaCy doesn't yet support. Based on this, I have a new list that I'll try adding to SpaCy to see if these kinds of errors go away.

### ERROR_TYPES.GENITIVE: 41
These errors occur for two cases of Swedish genitive form. For uppercase organizations, the form "ECC:s" is used. The other case is the rare "Magnus' saker", where "Magnus" is a name the ends with and s. You can optionally add a ' after the s in that case. This form only occurs twice in this (slightly old-fashioned) corpus, so I'm not sure it's worth supporting.

### ERROR_TYPES.LISTS: 12
All over the corpus there are examples of lists inside sentences. A translated example could be: "Are you 1) happy to see me or 2) not happy at all?". Pieces with letters instead of numbers exist, as well as versions with "(a)" and "(b)" too.

### ERROR_TYPES.SINGLELETTER: 8
This is the error I reported in #2578, where single letters at the end of the sentence get grouped together with the period. Except for the case with "i" (translated "in"), I also found sentences like "The sea level is 3 m." (where "m" is short for "meters").

### ERROR_TYPES.PARENTESISEQUAL: 3
[...]

PR #2615, opened 2018-07-30
Swedish: Exceptions for single letter words ending sentence

EmilStenstrom:
This is part of #2608.

## Description
Sentences ending in "i." (as in "... peka i."), "m." (as in "...än 2000 m."), should be tokenized as two separate tokens, not one.

In the corpus there is a single case of "let's call that variable e."; that gets incorrectly tokenized into "e.". But this variable name could be anything, so to fix it in the general case, I'd have to remove all single letter exceptions. There are two ways to do this:

1) In the language specific code, I could override the single letter versions, to make sure they are two different tokens again. This runs into the problem that "s." is an abbreviation in Swedish, which should be one unit. So one option would be "removing everything except s". Doesn't feel like "clean code" to me...

2) I also tried removing the single letter exceptions from BASE_EXCEPTIONS. Problem is, "s." is still an abbreviation, and if Swedish has one of them, so does probably other languages... So removing the BASE_EXCEPTIONS means going through all supported languages and make sure to add single letter abbreviations where needed. That's too big of a task for me.

Since none of those two options feel good, I've opted for just adding "i." and "m." as exceptions. This is what this PR does. Feel free to suggest I should go with either 1) or 2) instead.

### Types of change
Bug fix

## Checklist

- [x] I have submitted the spaCy Contributor Agreement (As part of #2613)

### DIFF
--- a/spacy/lang/sv/tokenizer_exceptions.py
+++ b/spacy/lang/sv/tokenizer_exceptions.py
@@ -1,7 +1,7 @@
 # coding: utf8
 from __future__ import unicode_literals
 
-from ...symbols import ORTH, LEMMA, TAG, NORM, PRON_LEMMA
+from ...symbols import ORTH, LEMMA, TAG, NORM, PRON_LEMMA, PUNCT
 
 
 _exc = {}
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks!

PR #2680, opened 2018-08-17
Added lemmatizer and tag-map to Hindi

aryaprabhudesai:
## Checklist


aryaprabhudesai:
Hey,
The lemmatizer has code has been written by me. I wrote a simple code to
extract the lemma of the words from my corpus by using the HDTB(Hindi
Dependencies Treebank). Then I formatted the file according to the given
template on the spacy github page to make lemmatizer.py

I infact am trying to develop the model for Hindi. But I am facing problems
while creating the vocabulary(word-vectors and frequency) using the
spacy-dev-resources since the compiler is consistently giving me an error
that Hindi model is not available and hence is not able to load it which is
a little contradictory :P . Can you please look into this error so that I
can train the model.

Thank you.
Regards,
Arya

aniruddha-adhikary:
@aryaprabhudesai You can use `spacy-dev-resources` easily if you just make a few slight adjustments to `vocab/__init__.py`. A few things have gone out of date. spaCy 2.x no longer supports bz2 compressed vectors, and the `model` command has been replaced by the similar `init-model` command. After you take care of these, it should work.

honnibal:
Hi @aryaprabhudesai ,

First, thanks a lot for your work on this. Unfortunately there are a couple of annoying problems I don't think you could have foreseen easily :(

> The lemmatizer has code has been written by me. I wrote a simple code to extract the lemma of the words from my corpus by using the HDTB(Hindi Dependencies Treebank). Then I formatted the file according to the given template on the spacy github page to make lemmatizer.py

It's not very intuitive, but I think we can't accept this list of lemmas. The Hindi Treebank is licensed CC BY-NC, which means it can't be used for commercial projects. I don't know whether a list of lemmas would be considered a derived work, but I think it probably would, so we wouldn't be able to distribute the lemmas under the MIT license.

> I infact am trying to develop the model for Hindi. But I am facing problems while creating the vocabulary(word-vectors and frequency) using the spacy-dev-resources since the compiler is consistently giving me an error that Hindi model is not available and hence is not able to load it which is
a little contradictory :P . Can you please look into this error so that I can train the model.

You should probably try using the `develop` branch (installable over pip as `spacy-nightly`). This has some improvements to the `init-model` command.

Issue #2379, opened 2018-05-27
Some regression tests fail (locally) on current master


ansgar-t:
## How to reproduce the behaviour
Here's what I did: 
```
git clone https://github.com/explosion/spaCy.git explosion.spaCy
mkvirtualenv -p /usr/local/opt/python3/bin/python3 spaCyTmp 
cd explosion.spaCy/
pip install -r requirements.txt
python setup.py build_ext --inplace
python -m spacy download en_core_web_md
python -m spacy download en_core_web_sm
python -m spacy download fr_core_news_sm
python -m spacy download es_core_news_sm
python -m spacy download es_core_news_md
python -m spacy download en
python -m spacy download de
python -m spacy download fr
python -m spacy download es
py.test spacy --models --all --slow
```

This results in some test failures:
```
====================================================================================================================== FAILURES ======================================================================================================================
________________________________________________________________________________________________ test_issue401[en_core_web_sm-Jane's got a new car-1] ________________________________________________________________________________________________

EN = <spacy.lang.en.English object at 0x11da6b518>, text = "Jane's got a new car", i = 1

    @pytest.mark.models('en')
    @pytest.mark.parametrize('text,i', [("Jane's got a new car", 1),
                                        ("Jane thinks that's a nice car", 3)])
    def test_issue401(EN, text, i):
        """Text that 's in contractions is not lemmatized as '."""
        tokens = EN(text)
>       assert tokens[i].lemma_ != "'"
E       AttributeError: 'str' object has no attribute 'lemma_'

spacy/tests/regression/test_issue401.py:13: AttributeError
___________________________________________________________________________________________ test_issue401[en_core_web_sm-Jane thinks that's a nice car-3] ____________________________________________________________________________________________

EN = <spacy.lang.en.English object at 0x11da6b518>, text = "Jane thinks that's a nice car", i = 3

    @pytest.mark.models('en')
    @pytest.mark.parametrize('text,i', [("Jane's got a new car", 1),
                                        ("Jane thinks that's a nice car", 3)])
    def test_issue401(EN, text, i):
        """Text that 's in contractions is not lemmatized as '."""
        tokens = EN(text)
>       assert tokens[i].lemma_ != "'"
E       AttributeError: 'str' object has no attribute 'lemma_'

spacy/tests/regression/test_issue401.py:13: AttributeError
____________________________________________________________________________________________________ test_issue686[en_core_web_sm-He is the man] _____________________________________________________________________________________________________

[...]

Issue #653, opened 2016-11-22
Slow entity merging on large texts


ELind77:
First, I'd like to say that I'm really impressed with what you guys have been doing with spacy and the 1.X releases look great.

My issue is that when I do entity merging on a large document (novel length) it takes an unbelievably long time (as much as 10 minutes).  

This takes about 12 seconds on four novels:
```python
def tokenize(s):
    return [tok.lemma_ for tok in nlp(s, parse=False) if not tok.is_space]
```

But this takes up to 50 minutes on the same inputs:
```python
def tokenize_merge(s):
    doc = nlp(s, parse=False)
    for ent in doc.ents:
        ent.merge(ent.root.tag_, ent.text, ent.label_)
    return [tok.lemma_ for tok in doc if not tok.is_space]
```
(It actually took 24 minutes on a system running 0.101 and 51 minutes on a comperable machine running 1.2.0 but I only ran it a couple of times on each machine so that's not really conclusive.)

The second function is adapted from the old documentation and the sense2vec code.

My theory, based on documentation that says that token indices are changed when merging occurs, is that `merge` is O(n) on the length of the document.  Rather than try and dive into the code I thought I'd just ask, what is going on here?

Is there a better way to do this than what I'm doing?  It looks like some of the new callback functionality may be able to perform the merging for me?  What if I can guarantee that the document will never be sliced into, it will only ever be iterated over from the beginning, so it doesn't matter if the indices are messed up?  

I love the ability to merge tokens but unless there's a faster way to do it (I really hope I'm just doing it wrong), I don't think it's usable on input of the size I'm working with. 

#### Details
- Combined length of the input docs ~829000 tokens
- Python 2.7.6
- Ubuntu 14.04 (yes, I know it's time to upgrade, I'm working on it)

#### Side notes:
It's even worse when I do NPs too (I assume because there are more of them), but I don't have the times because I didn't let it finish.
If I'm right, and this is O(n), you could get a tiny decrease in constant factors by starting from the end.

honnibal:
I bet there's some sort of ["Shlemiel the painter"](http://wiki.c2.com/?ShlemielThePainter) behaviour here. It's an easy mistake to make because the children aren't stored explicitly (as that would require N**2 space).

If you have time, could you have a look at the implementation in doc.merge() and see whether you can figure out what might not be O(n)? It would be something in the way the heads are calculated, for translating the heads.

ELind77:
[...]

PR #2696, opened 2018-08-22
Introduces a bulk merge function, in order to solve issue #653

grivaz:
Introduces a bulk merge function, in order to solve issue #653

## Description
I introduced a bulk merge function in Doc. Most of the changes are in the Retokenizer class, where I replaced the _merge function by a _bulk_merge function. I adapted some unit tests to test the bulk merge (they pass). I tested the implementation on a subset of War and Peace and got the following results:
Before my changes: 165s
With bulk merge: 15s
With bulk merge used sequentially in a for loop: 440s

This is work in progress, depending on feedback, in particular on the following points:
- The bulk merge is slower than the original merge when used sequentially. I could reintroduce the original merge for that case.
- The bulk merge requires the tokens to merge to be given in document order. I could introduce a function that tests whether or they are ordered, and that returns an error if this assumption is not correct (or that orders them)
- The bulk merge assumes the tokens to merge to have no intersection with each other (e.g. no nested merges). I can add a function that removes any conflicting token before merging, and that then merges the remaining tokens, for example.
- I get the same issue as https://github.com/explosion/spaCy/issues/2379 when running the model tests. This behavior is the same with or without my changes, though, so I expect it is not linked to them. --all --slow passes.

This is my first github contribution, and I am somewhat new to pyhon and cython. While I do put a lot of effort into avoiding them, I do apologize in advance for rookie mistakes.

### Types of change
Performance enhancement

## Checklist
Apart from model tests, but they don't pass, even without my changes (on another recent clone of spaCy)
Waiting for feedback on api to change doc.

### DIFF
--- a/spacy/tokens/_retokenize.pyx
+++ b/spacy/tokens/_retokenize.pyx
@@ -61,8 +61,8 @@ def _merge(Doc doc, int start, int end, attributes):
     `start_idx` and `end_idx `do not mark start and end token boundaries,
     the document remains unchanged.
 
-    start_idx (int): Character index of the start of the slice to merge.
-    end_idx (int): Character index after the end of the slice to merge.
+    start (int): Token index of the start of the slice to merge.
+    end (int): Token index after the end of the slice to merge.
     **attributes: Attributes to assign to the merged token. By default,
         attributes are inherited from the syntactic root of the span.
     RETURNS (Token): The newly merged token, or `None` if the start and end
--- a/spacy/tokens/_retokenize.pyx
+++ b/spacy/tokens/_retokenize.pyx
@@ -7,6 +7,8 @@ from __future__ import unicode_literals
 from libc.string cimport memcpy, memset
 from libc.stdlib cimport malloc, free
 
+from cymem.cymem cimport Pool
+
 from .doc cimport Doc, set_children_from_heads, token_by_start, token_by_end
 from .span cimport Span
 from .token cimport Token
@@ -15,22 +17,29 @@ from ..structs cimport LexemeC, TokenC
 from ..attrs cimport TAG
 from ..attrs import intify_attrs
 from ..util import SimpleFrozenDict
-
+from ..errors import Errors
 
 cdef class Retokenizer:
     """Helper class for doc.retokenize() context manager."""
     cdef Doc doc
     cdef list merges
     cdef list splits
+    cdef set tokens_to_merge
     def __init__(self, doc):
         self.doc = doc
         self.merges = []
         self.splits = []
+        self.tokens_to_merge = set()
 
     def merge(self, Span span, attrs=SimpleFrozenDict()):
         """Mark a span for merging. The attrs will be applied to the resulting
         token.
         """
+        for token in span:
+            if token.i in self.tokens_to_merge:
+                raise ValueError(Errors.E097.format(token=repr(token)))
+            self.tokens_to_merge.add(token.i)
+
         attrs = intify_attrs(attrs, strings_map=self.doc.vocab.strings)
         self.merges.append((span, attrs))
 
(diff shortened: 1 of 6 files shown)


spacy/tokens/_retokenize.pyx
```diff
@@ -47,87 +48,137 @@ cdef class Retokenizer:
...
-    for i in range(doc.length - offset, doc.length):
-        memset(&doc.c[i], 0, sizeof(TokenC))
-        doc.c[i].lex = &EMPTY_LEXEME
-    doc.length -= offset
-    for i in range(doc.length):
+    cdef Span span
+    cdef const LexemeC* lex
+    cdef TokenC* token
+    spans = []
+    cdef TokenC** tokens = <TokenC **>malloc(len(merges) * sizeof(TokenC))
```
honnibal: The `try/finally` block here definitely works, but the increase in nesting is a bit unideal. To avoid this, we use this utility: https://github.com/explosion/cymem

So, we'd have:

```
cdef Pool mem = Pool()
tokens = <TokenC**>mem.alloc(len(merges), sizeof(TokenC))
```

When the `Pool` object's reference count drops to 0, it will be destroyed, and its pointers will be freed. So we just have to make sure that memory it allocates doesn't escape its scope, and we'll have no memory leaks or use-after-free errors.

spacy/tokens/_retokenize.pyx
```diff
@@ -47,87 +48,137 @@ cdef class Retokenizer:
...
+
+            # Assign attributes
+            for attr_name, attr_value in attributes.items():
+                if attr_name == TAG:
+                    doc.vocab.morphology.assign_tag(token, attr_value)
+                else:
+                    Token.set_struct_attr(token, attr_name, attr_value)
+
+            # Make sure ent_iob remains consistent
+            if doc.c[end].ent_iob == 1 and token.ent_iob in (0, 2):
```
honnibal: I'm a bit worried about this block. If possible we'd like to be making the `ent_iob` consistent at the end.

spacy/tokens/doc.pyx
```diff
@@ -934,6 +948,20 @@ cdef class Doc:
...
+
+    def fix_attributes(self, attributes):
+        if 'label' in attributes and 'ent_type' not in attributes:
+            if isinstance(attributes['label'], int):
+                attributes[ENT_TYPE] = attributes['label']
+            else:
+                attributes[ENT_TYPE] = self.vocab.strings[attributes['label']]
+        if 'ent_type' in attributes:
+            attributes[ENT_TYPE] = attributes['ent_type']
+
```
honnibal: `remove_label_if_necessary` and `fix_attributes` should probably be functions, not methods. At the very least, they would need to be private methods --- as we don't want to support these indefinitely going forward.

grivaz: Thank you so much for all your feedback! I'm on these changes.

grivaz:
I implemented the suggestions. New stats on the same corpus are 190s for the sequential case and 36s for the bulk case.

I was not sure what you meant by "At the end of the doc.retokenize() block, we could call doc._bulk_merge() if we accumulated more than one match", since doc doesn't know about merges. I implemented that change directly in the retokenizer. Please let me know if that's ok.

There is no API change, so I didn't change the documentation, but I can introduce the retokenizer API if you want me to.

If the pull request is accepted, I'd like to help with something else next. I feel competent implementing the split function, since I know this code well by now, but I couldn't find a corresponding issue. Would that be useful? I'll find something else to do otherwise.

honnibal:
Great, merging this :tada: . We'll make the docs changes when the next release is ready.

I haven't really thought about how to implement a split function. If you have suggestions, that'd be great.

PR #2709, opened 2018-08-27
When calling getoption() in conftest.py, pass a default option

njsmith:
This is necessary to allow testing an installed spacy by running
`pytest --pyargs spacy`.

Bug fix.


ines:
Thanks a lot and sorry for only getting to this now! 👍

I'll check if we also need this adjustment on `develop` – I kinda rewrote the whole test suite and deleted a bunch of old stuff, so it should also run *a lot* faster.

PR #2730, opened 2018-09-04
Less norm computations in token similarity

pzelasko:
I've removed the redundant vector norm computation in token similarity (was previously computed twice).

## Description

Before the change:

```
In [7]: doc = nlp('similar similarities')

In [8]: %timeit doc[0].similarity(doc[1])
38.8 µs ± 576 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

After the change:

```
In [3]: doc = nlp('similar similarities')

In [4]: %timeit doc[0].similarity(doc[1])
23.4 µs ± 131 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

### Types of change

Enhancement/speedup.

## Checklist

### DIFF
--- a/.github/contributors/pzelasko.md
+++ b/.github/contributors/pzelasko.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks! Not sure how I missed this. Surprised at the nice speed-up, too.

pzelasko:
Glad to help :)

PR #2731, opened 2018-09-04
Update Bengali token rules for Hyphen (-এর) and Digits

aniruddha-adhikary:
Bengali '-এ', '-এর' suffix is a special case that needs to be tokenized differently than other instances.

## Description
The tokenizer was not updated for properly tokenizing '-এর', '-এ' and Bengali digits. This patch fixes those outstanding issues. Regular expressions in `TOKENIZER_INFIXES` have been updated.

### Types of change
Enhancement to Bengali implementation.

## Checklist


ines:
Thanks a lot! 👍

PR #2759, opened 2018-09-12
Add words to portuguese language _num_words

filipecaixeta:
Add words that were missing in _num_words for portuguese

### DIFF
--- a/.github/contributors/filipecaixeta.md
+++ b/.github/contributors/filipecaixeta.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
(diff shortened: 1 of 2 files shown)


ines:
Thanks a lot! 👍

Btw, what about the words that were removed? If I remember correctly, there was a previous discussion somewhere about including both `pt-pt` and `pt-br`, and I think for now, we thought it'd be easiest to just have both in there.

filipecaixeta:
@ines ,
Brazilian and Portuguese grammars are the same so in theory, we don't need to have two separate languages. The differences would be in the meaning of some words. The changes that I made are valid for both.

Btw, there is a [Brazilian Portuguese corpus annotated with POS tags](http://nilc.icmc.usp.br/macmorpho/) that we could use to train a better POS tagger. But we would have to change the tokenizer for Portuguese a little bit. If you think its ok, I can do that.

ines:
Okay, perfect! (Ideally, our vision is to be able to provide a model that can handle both `pt-pt` and `pt-br` – but this will obviously take some time and experimentation.)

> Btw, there is a Brazilian Portuguese corpus annotated with POS tags that we could use to train a better POS tagger. But we would have to change the tokenizer for Portuguese a little bit. If you think its ok, I can do that.

Sure, if you want to experiment with this and see if you can beat the current accuracy and results, that'd be very interesting! 👍

Issue #2743, opened 2018-09-06
Errors while running spacy with keras example


romikforest:
spacy 2.0.11, python Python 3.6.6 :: Anaconda custom (64-bit), win 10 64 bit

I was trying to run spacy with keras example from documentation, but could not use it as is:
I could not save the spacy model after training.

with (model_dir / 'config.json').open('wb') as file_:
                file_.write(lstm.to_json())

had error str can't be saved to binary file (and it doesn't manage directory existence).
Than I can't evalaluate the model.
In rintime spacy loads 'en'. And while training - 'en_vectors_web_lg'
I've tried to load all standard models, standard models and vectors like nlp.vocab.vectors.from_disk('en_vectors_web_lg'). I always have errors:
Tagger is not in model
Wrong vectors shape
TypeError: 'Tagger' object is not iterable

and so on

ines:
Merging this with #2758.

Issue #2501, opened 2018-07-02
Deep learning with keras Sentiment analysis


mihirpatel7:
Deep learning with keras: 

   File "D:/Sentiment/datasets/DL_test_small.py", line 157, in evaluate
    for doc in nlp.pipe(texts, batch_size=1000, n_threads=4):
  File "C:\Python36\lib\site-packages\spacy\language.py", line 558, in pipe
    for name, proc in self.pipeline:
TypeError: 'Tagger' object is not iterable

## Your Environment

* Operating System: windows 10
* Python Version Used: 3.6
* spaCy Version Used: 2.0.11
* Environment Information:

danielcanueto:
I have the same issue when evaluating the results.

ines:
Merging this with #2758.

Issue #2209, opened 2018-04-12
Different problems to test "Text classification with Keras" example


lfdharo:
Hello,

I have found three main problems when training and evaluating the Text classifier using Keras example.

1. Saving the 'config.json' file after training in line 222, instead of opening the file with 'wb' it should be just 'w' since the write function in the next line will complain that it cannot handle a 'str' produced by lstm.to_json()

2. For testing (using the is_runtime flag to True) the program crashes when trying to set the weights of the embeddings and lstm in line 33. 

ValueError: Dimension 0 in both shapes must be equal, but are 1070971 and 0. Shapes are [1070971,300] and [0,0]. for 'Assign' (op: 'Assign') with input shapes: [1070971,300], [0,0].

The reason is that during training the model loads the embeddings from 'en_vectors_web_lg' but during test it does not load them because it is using 'en' model instead. Besides, the model is saved using pickle.dump(weights[1:], file_) in line 221 which clearly left the embeddings out. 

My partial solution is to load 'en_vectors_web_lg' inside the load function of the SentimentAnalyser class, get the embeddings and set the weights. But probably saving the embeddings directly with pickle should do the same thing.

3. After that, the system crashes in line 153 when iterating over the data using parallel batches. 
  File "/usr/local/lib/python3.5/dist-packages/spacy/language.py", line 558, in pipe
    for name, proc in self.pipeline:
TypeError: 'Tagger' object is not iterable

So, I solved this issue by not using the create_pipeline function and instead to use:
nlp = spacy.load('en')
nlp.add_pipe(SentimentAnalyser.load(model_dir, nlp, max_length=max_length))

I don't add the tagger and parser since they are already included in 'en'.

however, after all these changes the system prints 0.5 as accuracy and it is taking a long time to present this result (more than 10 minutes) even when I have a GPU.

Please let me know what could be wrong or if it is expected this accuracy for this example.

## Info about spaCy

* **spaCy version:** 2.0.11
* **Platform:** Linux-4.4.0-119-generic-x86_64-with-Ubuntu-16.04-xenial
* **Models:** en, en_vectors_web_lg, en_core_web_lg
* **Python version:** 3.5.3

## Info about models
    Installed models (spaCy v2.0.11)
    /usr/local/lib/python3.5/dist-packages/spacy

    TYPE        NAME                  MODEL                 VERSION                                   
    package     en-core-web-lg        en_core_web_lg        2.0.0    ✔      
[...]

Issue #1917, opened 2018-01-31
More guidance needed re deep_learning_keras.py example


colingoldberg:
More out of curiosity than knowledge - I am a novice here, I ran the deep_learning_keras.py example.

It ran for +/- an hour (on macosx High Sierra), producing the output below. Having done this, I was hoping to run a test or two after the training. But I see no files saved (where??), and it is not obvious how to proceed. Any help is appreciated.

# Output from training:
$ python deep_learning_keras.py 
/Users/colingoldberg/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:34: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
Using TensorFlow backend.
2018-01-31 09:26:24.138046: I tensorflow/core/platform/cpu_feature_guard.cc:137] Your CPU supports instructions that this TensorFlow binary was not compiled to use: SSE4.2 AVX AVX2 FMA
Downloading data from http://ai.stanford.edu/~amaas/data/sentiment/aclImdb_v1.tar.gz
83886080/84125825 [============================>.] - ETA: 0sUntaring file...
Loading spaCy
WARNING:tensorflow:From /Users/colingoldberg/anaconda3/lib/python3.6/site-packages/keras/backend/tensorflow_backend.py:1238: calling reduce_sum (from tensorflow.python.ops.math_ops) with keep_dims is deprecated and will be removed in a future version.
Instructions for updating:
keep_dims is deprecated, use keepdims instead
WARNING:tensorflow:From /Users/colingoldberg/anaconda3/lib/python3.6/site-packages/keras/backend/tensorflow_backend.py:1340: calling reduce_mean (from tensorflow.python.ops.math_ops) with keep_dims is deprecated and will be removed in a future version.
Instructions for updating:
keep_dims is deprecated, use keepdims instead
Parsing texts...
/Users/colingoldberg/anaconda3/lib/python3.6/site-packages/keras/models.py:874: UserWarning: The `nb_epoch` argument in `fit` has been renamed `epochs`.
  warnings.warn('The `nb_epoch` argument in `fit` '
Train on 310254 samples, validate on 304112 samples
Epoch 1/5
310254/310254 [==============================] - 1398s 5ms/step - loss: 0.5942 - acc: 0.6774 - val_loss: 0.5696 - val_acc: 0.6980
[...]

PR #2763, opened 2018-09-14
Fixed spaCy+Keras example

free-variation:
Updated the Keras sentiment analysis example to work with version 2.012 of spaCy.

## Description
The fix addresses #2501, #2743 and #2209, and possibly #1917.

The basic problem was that the language model/pipeline used needed to be fixed at `en-vectors-web-1g`, and the pipeline construction process for the validation stage needed to reflect the pipeline used in training.

Also, `config.json` was being written to a *binary* file, which caused an error.

### Types of change
Bug fixes.

## Checklist

### DIFF
--- a/.github/contributors/free-variation.md
+++ b/.github/contributors/free-variation.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks a lot for this! Super helpful, as several people were trying to use this, but we hadn't had time to get the example set up again to debug it. Hopefully it will stay more stable now, since Keras and Tensorflow are moving a little less quickly...

PR #2786, opened 2018-09-21
Adding French hyphenated first name

mauryaland:
## Description

I have added French hyphenated first name taken from this [web page](https://everybodywiki.com/Liste_de_pr%C3%A9noms_compos%C3%A9s_fran%C3%A7ais). 

### Types of change

Enhancement

## Checklist


mauryaland:
I wonder if it is not a good idea to add a regex like this one _\b[A-ZÉ]{1}\w*-{1}[A-ZÉ]{1}\S+\b_  to the tokenizer in order to match every word that has the following structure _Xxx-Xxx_, especially to integrate hyphenated names from other languages or [double-barrelled name](https://en.wikipedia.org/wiki/Double-barrelled_name).

I have also checked for the english model **en_core_web_md** and hyphenated names are also split into 3 tokens . Is this behaviour desired? For example:
```python
nlp_en = spacy.load('en_core_web_md')
doc = nlp_en('Mary-Jane')
for token in doc:
    print(token)
```
returns:
Mary 
\- 
Jane 

To finish, when I run the NER on the french model **fr_core_news_md**  with the integration of the French hyphenated first name, the result is "LOC" for some. For example, "Jean et Paul" and "Jean-Paul" return both "PER" but "Anne et Lise" and "Anne-Lise" return respectively "PER" and "LOC"

I hope that discussion will be useful and please do not hesitate to tell me if I 'm going in the wrong direction.

honnibal:
Thanks! We might eventually want to move away from having such large files within the repository, and we might want to reorganise the tokenizer exceptions list. But for now it's valuable to have the list, and certainly tokenizing these names better will help the entity recognizer.

PR #2802, opened 2018-09-25
Fix error

darindf:
```
ValueError: cannot resize an array that references or is referenced
by another array in this way.  Use the resize function
```

This error occurs after freezing an application using spacy using pyinstaller. The relevant stack is

```
File "Lib\site-packages\spacy\language.py", line 554, in pipe
  File "Lib\site-packages\spacy\language.py", line 578, in pipe
  File "pipeline.pyx", line 434, in pipe
  File "pipeline.pyx", line 472, in spacy.pipeline.Tagger.set_annotations
  File "doc.pyx", line 870, in spacy.tokens.doc.Doc.extend_tensor
```

It's interesting that this is resizing an array of size zero to the size of the tensor, so not sure where array is being referenced.

### DIFF
--- a/.github/contributors/darindf.md
+++ b/.github/contributors/darindf.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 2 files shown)


honnibal:
That is indeed weird! Thanks for the fix.

PR #2806, opened 2018-09-27
Add jupyter=True to displacy.render in documentation

charlax:
## Description
Fix minor typo in doc

### Types of change
Documentation fix

## Checklist
- [X] I have submitted the spaCy Contributor Agreement. See #2805.
- [X] I ran the tests, and all new and existing tests passed. N/A

### DIFF
--- a/website/usage/_visualizers/_jupyter.jade
+++ b/website/usage/_visualizers/_jupyter.jade
@@ -12,10 +12,10 @@ p
     from spacy import displacy
 
     doc = nlp(u'Rats are various medium-sized, long-tailed rodents.')
-    displacy.render(doc, style='dep')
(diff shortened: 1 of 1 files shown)


charlax:
Mmmh now that I re-read the doc, you clearly state that this arg should be set only if auto-detection fails. Might want to leave it as is then. In my case, auto-detection did fail (Jupyter 5.7.0).

ines:
Thanks a lot! 👍 

(Speaking of Jupyter, I was also going to look into fixing the auto-detection, because in theory, spaCy should detect whether the user is running a notebook or not, and set the flag automatically. But it seems to not work as expected at the moment...)

charlax:
Awesome thanks!

PR #2807, opened 2018-09-27
💫 Fix handling of morphological features

honnibal:
This patch updates the `Morphology` class to handle the universal dependencies features. The features are provided in an enum, so they're hard-coded. This has some pros and cons --- I'm not sure it's what we want in total. Maintaining the list is a hassle, but it does avoid some problems, as the code raises errors if invalid labels are used. The enum also makes some operations easier to implement efficiently, especially without acquiring the GIL.

The implementation is as follows. The `TokenC` data structure receives a 64-bit hash that represents the set of morphological features assigned to it. This 64-bit hash is a key into the `Morphology` table, owned by the vocab. The `Morphology` table is used to resolve the key to a `RichTagC` struct, which is used internally to hold the features.

The point of this implementation is to keep the `TokenC` struct small and fixed-size. Including the `RichTagC` in-place within the `TokenC` would greatly increase how much data the token required to represent, as potentially many morphological attributes may be present. What we exploit is that within a given language, the *combinations* of morphological attributes are quite limited, so we're guaranteed to see far fewer morphological types than we do tokens.

An alternate implementation would be to hold a const pointer to the morphological analysis on the token. This is analogous to how the `LexemeC` struct works. This might indeed be better; I didn't think of it at the time. It should be a simple change to make.

The patch also includes a scrappy but working class to predict the morphological attributes, `Morphologizer`. This is just like the tagger, except its final layer is a new type of layer I've called `MultiSoftmax`. This layer is just like running a bunch of Softmax layers that all have the same input --- it just joins them together so that we only make one bigger matrix multiplication call, instead of lots of little ones.

With the morphological attributes working, we can finally improve the lemmatizer, by making the lemmatization rules specific to morphological features. For instance, instead of having a general rule "rewrite 'ing' to ''", we would key that rule by the morphological feature `Tense=Past`, so that it can't apply to verbs marked present-tense like "ring".

## Things to do before merging

* [x] ~Reconsider whether to have an enum of the feature values.~ The enum was bad. Replaced.
* [x] ~If we do want the enum, make sure all features in Universal Dependencies v2.2 are present.~
* [x] ~If we do want the enum, update the symbols.pyx , to make sure that the symbols values in morphology.pyx and symbols.pyx align~
* [x] ~Consider giving the `TokenC` a pointer to the morphological analysis, rather than a hash ID.~ I think the ID is fine.
* [x] Add API on the Token, to use the new morphological attributes. 
* [ ] Document new API to use the morphological attributes.


honnibal:
Updated the PR to branch to current develop, and removed the ugly enum.

I've also drafted out a new class, `spacy.tokens.morphanalysis.MorphAnalysis`, which can provide the usage API. Moving this under a new class lets us provide a richer API without poluting the `Token` class itself.

honnibal:
This has been merged onto `develop`, so closing the PR.

PR #2818, opened 2018-10-02
Rule-based French Lemmatizer

mauryaland:
## Description

Add a rule-based French Lemmatizer following the english one and the excellent PR for [greek language optimizations](https://github.com/explosion/spaCy/pull/2558) to adapt the Lemmatizer class.

### Types of change

- Lemma dictionary used can be found [here](http://infolingu.univ-mlv.fr/DonneesLinguistiques/Dictionnaires/telechargement.html), I used the XML version.
- Add several files containing exhaustive list of words for each part of speech 
- Add some lemma rules
- Add POS that are not checked in the standard Lemmatizer, i.e PRON, DET, ADV and AUX
- Modify the Lemmatizer class to check in lookup table as a last resort if POS not mentionned
- Modify the lemmatize function to check in lookup table as a last resort
- Init files are updated so the model can support all the functionalities mentioned above
- Add words to tokenizer_exceptions_list.py in respect to regex used in tokenizer_exceptions.py

## Checklist


amperinet:
Hello guys,
I have been working on the French lemmatization as well, and wanted to add my contribution. It's a nice suprise to see that some work has already been done by @mauryaland !! I guess I should wait for this contribution to be approved, check the differences with what I've done and then create a pull request with changes if necessary. Do you know if the validation of @mauryaland could be long? Or how long it could be?
Thanks for your nice work @ines and @honnibal !

PR #2836, opened 2018-10-09
WIP - Dependency Tree Matcher (Validations + Semgrex operators)

skrcode:
Added validation and all major Semgrex operators
## Description

1. Added checks to validate input.
2. Added all major Semgrex operator support. Includes support for'<','>','>>','<<','.','$+','$-','$++','$--'.
Details regarding the operators can be found here : https://nlp.stanford.edu/nlp/javadoc/javanlp/edu/stanford/nlp/semgraph/semgrex/SemgrexPattern.html
Note: The subtree matching code needs to be re-looked at. I guess there are some caveats in the Semgrex pattern matching algorithm that need to be implemented. Unordered subtree matching takes exponential time and the heuristics implemented in Semgrex can be of help.
### Types of change

Enhancement
## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -261,8 +261,13 @@ class Errors(object):
              "Span objects, or dicts if set to manual=True.")
     E097 = ("Invalid pattern: expected token pattern (list of dicts) or "
             "phrase pattern (string) but got:\n{pattern}")
-
-
+    E098 = ("Invalid pattern specified: expected both SPEC and PATTERN.")
+    E099 = ("First node of pattern should be a root node. The root should "
+            "only contain NODE_NAME.")
+    E100 = ("Nodes apart from the root should contain NODE_NAME, NBOR_NAME and "
+            "NBOR_RELOP.")
+    E101 = ("NODE_NAME should be a new node and NBOR_NAME should already have "
+            "have been declared in previous edges.")
 @add_codes
 class TempErrors(object):
     T001 = ("Max length currently 10 for phrase matching")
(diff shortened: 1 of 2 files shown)


skrcode:
@honnibal There's definitely something better than can be done on the pattern matching algorithm. I've hit a roadblock in trying to find a combinatorial approach for calculating all matches correctly and quickly. I've gone through a lot of the literature on this as well as existing code modules and could not find anything suitable that we could extend on to. Any thoughts on this?

skrcode:
@honnibal Yeah, I agree ! I think even corenlp does not handle those exponential cases.
While trying out this example `{} > {} > {} > {} > {} > {} > {} > {} > {} > {} > {} > {} > {} > {} > {} > {}`
for the text `The quick brown fox jumped over the lazy dog.`, though the length of the input is small,
a large number of matches could be possibly found leading to a timeout or too much memory consumption.

skrcode:
@honnibal I've rewritten some of the code here: https://github.com/explosion/spaCy/pull/3465
I've thought about the above performance problem in depth and figured out that for atleast some input specification, there are going to be a large number(exponential) of output subtrees. For example, if the input pattern is something like "give me all subtree matchings with any node having any two children", then the result would obviously have to contain all subtrees(without ordering too) from the dependency parse containing a node with two children. This result in itself would be exponential and so it has to run in atleast exponential time and memory footprint. I feel that there isn't any significant performance improvement that can be theoretically done, although I've added some constant time optimizations that could perform fairly well in practical use-cases, considering that an average sentence length would be some number hovering around 10.

PR #2856, opened 2018-10-16
Redundant ')' in the Stop words' example

digest0r:
## Checklist


ines:
Ah, well-spotted, thanks! 👍

PR #2885, opened 2018-10-29
Modify lemmatizer.py to check if the word is in regular lists of words for French language

mauryaland:
## Description

On my previous PR, I forgot to add the code to check if the word is in the regular lists of words for each part of speech. That leads to lemmas taken from the lookup table while the word is already in its lemma form. Consequently, bad results were occurring.

### Types of change

Bug fix/enhancement

## Checklist


mauryaland:
I modified the wrong lemmatizer.py file. I will change the good one as soon as possible and reopen the PR. 
Sorry about that!

PR #2886, opened 2018-10-29
Check if the word is in one of the regular lists specific to each POS for French Language

mauryaland:
## Description

On the previous PR ([#2818](https://github.com/explosion/spaCy/pull/2818)), I forgot to add the code to check if the word is in the regular lists of words for each part of speech. That leads to lemmas taken from the lookup table while the word is already in its lemma form. Consequently, bad results were occurring.

### Types of change

Bug fix/enhancement

## Checklist


honnibal:
Thanks!

PR #2899, opened 2018-11-04
more pep8 conformity; faster membership operations

ethankoch4:
## Description

I have added more conformity to PEP8 guidelines. Namely, when checking Truth-iness, one should use:

```
if some_variable:
```

while in the repo, I found many occurrences of:

```
if some_variable is True:
```

which is the worst way to check this, [according to PEP8.](https://www.python.org/dev/peps/pep-0008/#id51)

Additionally, I made marginal speed-ups by using membership operations on a set (`O(1)` operation), instead of a list (`O(n)` operation).

The few other changes are for uniformity to the syntax of the rest of the files as well as readability.

Here are my timing results (showing that membership in a `set` is faster than a `list` or `tuple`):
screen shot 2018-11-04 at 4 41 35 pm

### Types of change

Enhancement

## Checklist

### DIFF
--- a/.github/contributors/ethankoch4.md
+++ b/.github/contributors/ethankoch4.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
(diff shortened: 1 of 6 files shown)


honnibal:
I appreciate the thought, but your changes aren't correct. The statements `if value is True` and `if value` aren't equivalent, and when we write `if value is True` we mean what we say: we're trying to check whether the value is literally, and only, the value `True`. If we change this to `if value:`, the condition will be true whenever `value` is any other truthy value, which will be incorrect.

ethankoch4:
Sorry, but if it is the case that the value could be something other than boolean, that is a different question and should be addressed elsewhere, correct? There

Also, if you feel strongly, I can revert those changes and only include the performance improvements in this PR.

ethankoch4:
You could bump performance a bit too, especially in frequently called functions if you avoid the unnecessary comparisons:

screen shot 2018-11-04 at 5 14 02 pm

ethankoch4:
additionally:
 
I removed an unused variable from a function.

I formatted multiline variables more correctly and with adherence to PEP8 & the rest of the file.

I changed variables whose only purpose was as a placeholder to be a generator object instead of a list, decreasing memory and increasing speed.

ethankoch4:
@honnibal

honnibal:
> Sorry, but if it is the case that the value could be something other than boolean, that is a different question and should be addressed elsewhere, correct?

Yes, the value might not be `True`, `False` or `None` --- it could be pretty much any Python object. Given that, by far the best way to check whether it's the value `True` and no other value is to write `if value is True`.

ethankoch4:
I will make that change then and re-submit the PR for the performance improvements.

honnibal:
...Look, your pull request introduced a bunch of bugs, without introducing any functionality. You didn't take the time to understand the code, and just assumed you were right, when you weren't. That alone should make you slow down here.

At the same time you're trying to make micro-optimisations without benchmarking the actual code in question to check whether it's in a hotspot, and I think without understanding that the code you're editing isn't even directly run by Python. It's compiled by Cython and then executed as a C extension.

ethankoch4:
Understood, and you are correct -- I will keep all of that in mind in the future. Hopefully I can contribute in a meaningful way

PR #2905, opened 2018-11-06
Remove u-strings from example on website index

adamchainz:
## Description

u-strings are Python 2 syntax that no-op in Python 3 and Binder runs the example on Python 3.

### Types of change

documentation/bug fix

## Checklist
- [x] I have submitted the spaCy Contributor Agreement. - #2906
- [n/a] I ran the tests, and all new and existing tests passed.
- [n/a] My changes don't require a change to the documentation, or if they do, I've added all required information.

### DIFF
--- a/website/index.jade
+++ b/website/index.jade
@@ -63,13 +63,13 @@ include _includes/_mixins
                 nlp = spacy.load('en_core_web_sm')
 
                 # Process whole documents
-                text = (u"When Sebastian Thrun started working on self-driving cars at "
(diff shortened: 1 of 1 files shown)


honnibal:
spaCy supports both Python 2 and Python 3 though! And if someone's on Python2 (often because they don't know much about this and innocently typed 'python') they should be able to paste the code from the website without getting confusing errors.

So, thanks for the thought, and we really do appreciate small changes like this. But the u-strings are intentional, and need to stay.

PR #2909, opened 2018-11-08
Removed space in docs + added contributor indo

mikelibg:


ines:
Thanks for the attention to detail – can't believe we didn't catch this! 🤦‍♀️

PR #2922, opened 2018-11-13
Allow input text of length up to max_length, inclusive

danielhers:
## Description
The name `max_length` implies that input of exactly that length is still valid, so this should be the case.
I encountered this problem when I saw the error message and tried to do `text = text[:nlp.max_length]` to resolve it, but of course it was still too long.

### Types of change
Minor bug fix.

## Checklist


honnibal:
Fair point! Thanks.

Issue #1971, opened 2018-02-12
💫 Better, faster and more customisable matcher


ines:
Related issues: #1567, #1711, #1819, #1939, #1945, #1951, #2042

We're currently in the process of rewriting the match loop, fixing long-standing issues and making it easier to extend the `Matcher` and `PhraseMatcher`. The community contributions by @GregDubbin and @savkov have already made a big difference – we can't wait to get it all ready and shipped.

This issue discusses some of the planned new features and additions to the match patterns API, including matching by **custom extension attributes** (`Token._.`), **regular expressions**, **set membership** and **rich comparison for numeric values**.

### Custom extension attributes
spaCy v2.0 introduced [custom extension attributes](https://spacy.io/usage/processing-pipelines#custom-components-attributes) on the `Doc`, `Span` and `Token`. Custom attributes make it easier to attach arbitrary data to the built-in objects, and let users take advantage of spaCy's data structures and the `Doc` object as the "single source of truth". However, not being able to match on custom attributes was quite limiting (see #1499, #1825).

The new patterns spec will allow an `_` space on token patterns, which can map to a dictionary keyed by the attribute names:

```python
Token.set_extension('is_fruit', getter=lambda token: token.text in ('apple', 'banana'))

pattern = [{'LEMMA': 'have'}, {'_': {'is_fruit': True}}]
matcher.add('HAVING_FRUIT', None, pattern)
```

Both regular attribute extensions (with a default value) and property extensions (with a getter) will be supported and can be combined for more exact matches.

```python
pattern = [{'_': {'is_fruit': True, 'fruit_color': 'red', 'fruit_rating': 5}}]
```

### Rich comparison for numeric values
Token patterns already allow specifying a `LENGTH` (the token's character length). However, matching tokens of between five and ten characters previously required adding 6 copies of the exact same pattern, introducing unnecessary overhead. Numeric attributes can now also specify a dictionary with the predicate (e.g. `'>'` or `'<='`) mapped to the value. For example:

```python
pattern = [{'ENT_TYPE': 'ORG', 'LENGTH': 5}]          # exact length
pattern = [{'ENT_TYPE': 'ORG', 'LENGTH': {'>=': 5}}]  # length with predicate
```

The second pattern above will match a token with the entity type `ORG` that's 5 or more characters long. Combined with custom attributes, this allows very powerful queries combining both linguistic features and numeric data:

```python
# match a token based on custom numeric attributes
[...]

PR #2925, opened 2018-11-14
💫 Allow matching non-ORTH attributes in PhraseMatcher

ines:
Related issue: #1971

## Description

This PR introduces the ability to match other token attributes in the `PhraseMatcher` – for example, to find sequences of part-of-speech tags that match an example sentence. On initialization, the `attr` keyword argument can specify an attribute string name.

```python
matcher = PhraseMatcher(nlp.vocab, attr='POS')
matcher.add('PATTERN', None, nlp('I love cats'))
doc = nlp('yes, you hate dogs')
assert len(matcher(doc)) == 1  # matched 'you hate dogs' based on token.pos
```

Most useful use case: case-insensitive phrase matches! 🎉

```python
matcher = PhraseMatcher(nlp.vocab, attr='LOWER')
matcher.add('PATTERN', None, nlp('iPhone X'))
doc = nlp('i have an iphone x')
assert len(matcher(doc)) == 1  # matched 'iphone x' based on token.lower
```

It even works for boolean flags:

```python
matcher = PhraseMatcher(nlp.vocab, attr='IS_PUNCT')
matcher.add('PATTERN', None, nlp('hello!'))
doc = nlp('no.')
assert len(matcher(doc)) == 1  # matched 'no.' based on token.is_punct
```

### Implementation details

Under the hood, the `PhraseMatcher` now creates a "phantom doc" that consists of the attribute values instead of the `token.orth`. For example, `Doc(vocab, words=[token.pos_ for token in doc])`.

Since matching requires setting flags on the lexemes, we create new tokens for non-`ORTH` attributes by concatenating the attribute name and value, to prevent polluting the lexemes. For example, if we match on `POS`, the phantom doc's tokens become `'matcher:POS-VERB'` (very unlikely token) instead of just `'VERB'`. This prevents the lexical entry for `'VERB'` from receiving flags that can potentially cause false positives later on.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/matcher.pyx
+++ b/spacy/matcher.pyx
@@ -631,8 +631,9 @@ cdef class PhraseMatcher:
             `doc[start:end]`. The `label_id` and `key` are both integers.
         """
         matches = []
-        match_doc = doc
-        if self.attr != ORTH:
+        if self.attr == ORTH:
+            match_doc = doc
+        else:
             # If we're not matching on the ORTH, match_doc will be a Doc whose
             # token.orth values are the attribute values we're matching on,
             # e.g. Doc(nlp.vocab, words=[token.pos_ for token in doc])
--- a/spacy/matcher.pyx
+++ b/spacy/matcher.pyx
@@ -557,8 +557,11 @@ cdef class PhraseMatcher:
         self.max_length = max_length
         self.vocab = vocab
         self.matcher = Matcher(self.vocab)
-        self.attr = self.vocab.strings[attr]
-        self.phrase_ids = PreshMap()
+        if isinstance(attr, long):
+            self.attr = attr
+        else:
+            self.attr = self.vocab.strings[attr]
+^        self.phrase_ids = PreshMap()
         abstract_patterns = [
             [{U_ENT: True}],
             [{B2_ENT: True}, {L2_ENT: True}],
--- a/spacy/matcher.pyx
+++ b/spacy/matcher.pyx
@@ -561,7 +561,7 @@ cdef class PhraseMatcher:
             self.attr = attr
         else:
             self.attr = self.vocab.strings[attr]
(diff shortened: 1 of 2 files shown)


spacy/matcher.pyx
```diff
@@ -546,16 +546,18 @@ cdef class PhraseMatcher:
...
 
-    def __init__(self, Vocab vocab, max_length=0):
+    def __init__(self, Vocab vocab, max_length=0, attr='ORTH'):
         if max_length != 0:
             deprecation_warning(Warnings.W010)
         self.mem = Pool()
         self.max_length = max_length
         self.vocab = vocab
         self.matcher = Matcher(self.vocab)
+        self.attr = self.vocab.strings[attr]
```
honnibal: If the user passes in e.g. `ORTH` as an ID, this will cause an error. So we probably want something like

```
if isinstance(attr, long):
    self.attr = attr
else:
    self.attr = self.vocab.strings[attr]
```

spacy/matcher.pyx
```diff
@@ -625,8 +628,15 @@ cdef class PhraseMatcher:
...
         matches = []
-        for _, start, end in self.matcher(doc):
-            ent_id = self.accept_match(doc, start, end)
+        match_doc = doc
+        if self.attr != ORTH:
+            # If we're not matching on the ORTH, match_doc will be a Doc whose
+            # token.orth values are the attribute values we're matching on,
+            # e.g. Doc(nlp.vocab, words=[token.pos_ for token in doc])
+            words = [self.get_lex_value(doc, i) for i in range(len(doc))]
+            match_doc = Doc(self.vocab, words=words)
```
honnibal: I'd probably rather have an else clause here, instead of assigning `match_doc = doc` above, and reassigning in the conditional.

PR #2927, opened 2018-11-14
💫 Create random IDs for SVGs to prevent ID clashes

### LINKED ISSUE #2924: displacy arc labels are place incorrectly in Jupyter notebook
| Software | Version |
|----------|---------|
| Python   | 3.6.4   |
| notebook | 5.4.1   |
| spacy    | 2.0.16  |

The SVG contents is the same when I copy the HTML from the inspector both in
jupyter and from the `displacy.serve` webpage.

When viewed through `displacy.serve`:

displacy-serve

And in Jupyter:
displacy jupyter

The SVG generated is the same, so I suspect some CSS issue is overriding the default, I don't know whether that should be solved within Jupyter, or whether it's spaCy's responsibility to make portable SVGs.  

Generated SVG here: https://gist.github.com/willprice/66d9ff961b693abe19e94d541eaacb99

This happens both in Firefox and Chrome.

I had a poke twiddling CSS rules off and on but couldn't seem to find any rules that affected the placement. Let me know if there's anything I can do providing additional info to help resolve the issue.

Thanks for the awesome library!

ines:
Resolves #2924.

## Description
Fixes problem where multiple visualizations in Jupyter notebooks would have clashing arc IDs, resulting in weirdly positioned arc labels. Generating a random ID prefix so even identical parses won't receive the same IDs for consistency (even if effect of ID clash isn't noticable here.)

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/displacy/render.py
+++ b/spacy/displacy/render.py
@@ -1,6 +1,8 @@
 # coding: utf8
 from __future__ import unicode_literals
 
+import random
+
 from .templates import TPL_DEP_SVG, TPL_DEP_WORDS, TPL_DEP_ARCS
 from .templates import TPL_ENT, TPL_ENTS, TPL_FIGURE, TPL_TITLE, TPL_PAGE
 from ..util import minify_html, escape_html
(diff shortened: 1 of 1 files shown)


willprice:
This has resolved the issue, I've now got sensible labelled arcs! Thanks @ines

PR #2931, opened 2018-11-15
💫 Add experimental ULMFit/BERT/Elmo-like pretraining 

honnibal:
Add support for a new command, `spacy pretrain`:

```
usage: spacy pretrain [-h] [-cw 128] [-cd 4] [-er 1000] [-d 0.2] [-i 1] [-s 0]
                      texts_loc vectors_model output_dir

    Pre-train the 'token-to-vector' (tok2vec) layer of pipeline components,
    using an approximate language-modelling objective. Specifically, we load
    pre-trained vectors, and train a component like a CNN, BiLSTM, etc to predict
    vectors which match the pre-trained ones. The weights are saved to a directory
    after each epoch. You can then pass a path to one of these pre-trained weights
    files to the 'spacy train' command.

    This technique may be especially helpful if you have little labelled data.
    However, it's still quite experimental, so your mileage may vary.

    To load the weights back in during 'spacy train', you need to ensure
    all settings are the same between pretraining and training. The API and
    errors around this need some improvement.


positional arguments:
  texts_loc             Path to jsonl file with texts to learn from
  vectors_model         Name or path to vectors model to learn from
  output_dir            Directory to write models each epoch

optional arguments:
  -h, --help            show this help message and exit
  -cw 128, --width 128  Width of CNN layers
  -cd 4, --depth 4      Depth of CNN layers
  -er 1000, --embed-rows 1000
                        Embedding rows
  -d 0.2, --dropout 0.2
                        Dropout
  -i 1, --nr-iter 1     Number of iterations to pretrain
  -s 0, --seed 0        Seed for random number generators
```

The `pretrain` command uses a novel trick, in order to support pre-training with our small models. Previous work has mostly used language modelling objectives over large vocabularies. Meeting that objective requires sufficiently large output layers, which means the hidden layers have to be quite large too. This doesn't work well for our fast and small CNN.

My solution is to instead load in a pre-trained vectors file, and use the vector-space as the objective. This means we only need to predict a 300d vector for each word, instead of trying to softmax over 10,000 IDs or whatever. It also means the vocabulary we can learn is very large, which is quite satisfying.

To make it easier to use the pre-trained weights, `spacy train` now supports a new CLI argument, `-t2v`, which takes a path to a pre-trained weights file. It's the user's responsibility to make sure the settings match up across the two commands, which is a bit fiddly at the moment if you've used non-default depth or width etc.

I thought of this trick a long time ago, and had it implemented in a half-finished state in the `Tensorizer` component. I always assumed it wouldn't work, as it feels too much like compression, and not enough like prediction. The strong performance of the BERT model made me take another look. The BERT model's objective isn't much different from simply using dropout, which we can easily apply.

In preliminary tests, I've already achieved pretty strong improvements for text classification over small training sizes. Training on **1000 documents** from the IMDB corpus, with pre-training I'm able to reach 87% accuracy, which is roughly what Jeremy and Sebastian report in their ULMFit paper ([Figure 3](https://arxiv.org/pdf/1801.06146.pdf)). Without pre-training, the best I could get to was 85%.

Interestingly, the technique seems to work better if the vectors are also used as part of the input. I find this completely surprising -- I expected the opposite. I don't know what's going on with this, but I've currently hard-coded that the vectors should be used as features.

The obvious thing to try is running something like ULMFit or BERT as the target for the CNN to learn from, rather than just using static vectors. I expect that should work better.

## Checklist

### DIFF
--- a/spacy/__main__.py
+++ b/spacy/__main__.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 if __name__ == '__main__':
     import plac
     import sys
-    from spacy.cli import download, link, info, package, train, convert
+    from spacy.cli import download, link, info, package, train, pretrain, convert
     from spacy.cli import vocab, init_model, profile, evaluate, validate
     from spacy.cli import ud_train, ud_evaluate
     from spacy.util import prints
@@ -16,6 +16,7 @@ if __name__ == '__main__':
         'link': link,
         'info': info,
         'train': train,
+        'pretrain': pretrain,
         'ud-train': ud_train,
         'evaluate': evaluate,
         'ud-evaluate': ud_evaluate,
--- a/spacy/cli/__init__.py
+++ b/spacy/cli/__init__.py
@@ -4,6 +4,7 @@ from .link import link
 from .package import package
 from .profile import profile
 from .train import train
+from .pretrain import pretrain
 from .evaluate import evaluate
 from .convert import convert
 from .vocab import make_vocab as vocab
--- a/spacy/cli/pretrain.py
+++ b/spacy/cli/pretrain.py
@@ -1,0 +1,191 @@
+'''This script is experimental.
+
+Try pre-training the CNN component of the text categorizer using a cheap
+language modelling-like objective. Specifically, we load pre-trained vectors
+(from something like word2vec, GloVe, FastText etc), and use the CNN to
+predict the tokens' pre-trained vectors. This isn't as easy as it sounds:
+we're not merely doing compression here, because heavy dropout is applied,
+including over the input words. This means the model must often (50% of the time)
+use the context in order to predict the word.
+
+To evaluate the technique, we're pre-training with the 50k texts from the IMDB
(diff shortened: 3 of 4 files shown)


honnibal:
Quick update in case people revisit this:

* Better hyper-parameter search pushed the baseline for my small textcat experiment above 86%, while the pre-training still only took it to around 87%. So, the gains here weren't so impressive after all. Tests on NER were also largely negative.

* The problem is that the objective of simply predicting the word's vector is indeed too "compressionish". We need to do something that looks more like prediction. The BERT 'masked language model' objective seems to be working very well. I'll push an update to the pretrain script with this objective soon.

honnibal:
@aschmu You should be able to use `spacy init-model`, like this:

```
wget https://dl.fbaipublicfiles.com/fasttext/vectors-crawl/cc.fr.300.vec.gz
spacy init-model fr ./fr_vectors_web_lg --vectors cc.fr.300.vec.gz
```

You should then be able to use the `./fr_vectors_web_lg` directory you created in `spacy pretrain`

PR #2974, opened 2018-11-27
Improved polish tokenizer and stop words.

Gizzio:
## Description

Hi, we (@kowaalczyk @DoomCoder and @MateuszOlko) are working on complementing spaCy with polish language support. This is first PR, which covers tokenizer and stop words improvements and tokenizer tests.

### Types of change
Enhancement

## Checklist


honnibal:
Thanks a lot for this! Looks good.

The LGPL license is a bit of a problem on the exceptions data though :(. We really don't want to complicate the license story on spaCy, which we publish under an MIT license. Having part of the code be LGPL licensed wouldn't work for us.

Is it possible to find an alternative source of the data?

tomekkorbak:
@honnibal  is two-clause BSD license compatible with spaCy's MIT license? We got in touch with the author of tokenizer rules and he's willing to republish under two-clause BSD.

honnibal:
@tomekkorbak Yes, 2-clause BSD would be fine, thanks!

Gizzio:
@honnibal @ines 

It took a while, but finally we have a suitable license. I also added new like-number checking and removed "splitting" exceptions from tokenizer.

As I said, we are working on full support for the polish language. We already have a working rule-based lemmatizer, tag map to UD, the first version of POS tagger (trained on NKJP corpus), and dependency parser (trained on UD). We want to make PR with lemmatization and tag map, but rules in lemmatizer are divided by tags, so our lemmatization and tag map depends on POS tagger. Should we make PR with lemmatizer, or wait for POS tagger to be merged?

honnibal:
Thanks for the changes! Sorry about the delay merging this, I've been traveling.

Issue #2933, opened 2018-11-16
is_ascii unclear documentation


wxv:
Documentation of `is_ascii`: 

> Does the token consist of ASCII characters? Equivalent to `[any(ord(c) >= 128 for c in token.text)]`.

Testing:

```
>>> doc = nlp(u"汉_字")
>>> doc[0]
汉_字
>>> doc[0].is_ascii
False
>>> [any(ord(c) >= 128 for c in doc[0].text)]
[True]
```

Seems to me that the "equivalent" statement returns something like `is_not_ascii`. Also I don't know why it is surrounded by square brackets which would put the bool in a list. I suggest:

> Does the token consist of **only** ASCII characters? Equivalent to `all(ord(c) < 128 for c in token.text)`.

## Which page or section is this issue related to?
https://spacy.io/api/token#attributes

DuyguA:
Hello,
Standart code for is_ascii is all(ord(c) < 128 for c in token.text) i.e. all characters are ASCII. I rechecked the lexeme code and it's as expected.

@ines I think documentation needs a quick fix. I'll do it soon.

PR #2988, opened 2018-11-30
Fix is_ascii documentation

wxv:
Proposed in #2933. 

## Description
Changes the "equivalent code" of `is_ascii` to return a correct Boolean (previously returning a list with wrong truth value).  

### Types of change
Documentation

## Checklist


ines:
Thanks a lot! 👍

Issue #2991, opened 2018-11-30
Tests for French lemmatization are skipped (but the model is installed)


amperinet:
## How to reproduce the problem
I modified the lemmatization files for French (see PR #2992) and went through the tests that are in /spacy/tests/lang/fr, but the tests for lemmatization are skipped and I don't understand why.
From what I read on the tests git page, if a test is skipped it is because the model is not found. So I did as many checks as possible related to the models. 

I ran "pytest -v ." from the French tests directory ('spacy/tests/lang/fr/').

I tested on both Linux and Mac. 

### 1 - Mac 

```bash
MacBook-Pro-2:fr amperinet$ pytest -v .
========================================================================= test session starts =========================================================================
platform darwin -- Python 3.6.5, pytest-4.0.1, py-1.7.0, pluggy-0.8.0 -- /Library/Frameworks/Python.framework/Versions/3.6/bin/python3.6
cachedir: .pytest_cache
rootdir: /Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/spacy/tests/lang/fr, inifile:
collected 18 items                                                                                                                                                    

test_exceptions.py::test_tokenizer_infix_exceptions[aujourd'hui] PASSED                                                                                         [  5%]
test_exceptions.py::test_tokenizer_infix_exceptions[Aujourd'hui] PASSED                                                                                         [ 11%]
test_exceptions.py::test_tokenizer_infix_exceptions[prud'hommes] PASSED                                                                                         [ 16%]
test_exceptions.py::test_tokenizer_infix_exceptions[prud\u2019hommal] PASSED                                                                                    [ 22%]
test_exceptions.py::test_tokenizer_handles_abbr[janv.-janvier] PASSED                                                                                           [ 27%]
test_exceptions.py::test_tokenizer_handles_abbr[juill.-juillet] PASSED                                                                                          [ 33%]
test_exceptions.py::test_tokenizer_handles_abbr[Dr.-docteur] PASSED                                                                                             [ 38%]
test_exceptions.py::test_tokenizer_handles_abbr[av.-avant] PASSED                                                                                               [ 44%]
test_exceptions.py::test_tokenizer_handles_abbr[sept.-septembre] PASSED                                                                                         [ 50%]
test_exceptions.py::test_tokenizer_handles_exc_in_text PASSED                                                                                                   [ 55%]
test_exceptions.py::test_tokenizer_handles_exc_in_text_2 PASSED                                                                                                 [ 61%]
[...]

PR #2992, opened 2018-11-30
Lemmatization of Nouns - French : adding rules and vocabulary

amperinet:
## Description

I added rules for French lemmatization of nouns ; French rules have recently been added (see issue #2818), but for nouns there was only one rule.
It is not sufficient to cover all the morphological cases for nouns.

The last version also misses a lot of general language French vocabulary. Thus, I added more vocabulary in both regular and irregular lists, to cover more general language in French. And I adapted the existing lists (regular/irregular cases) to fit the new rules.

### Types of change

An enhancement:
- Modifying the existing two lists (regular and irregular) and adding words using the LEFFF linguistic resource ([web page](http://www.labri.fr/perso/clement/lefff/telechargement.html)).
- Adding lemma rules for nouns

## Checklist

⚠️ **Tests for lemmatization are skipped and I do not understand why so I opened an issue (#2991)**


ines:
Thanks! 👍

PR #3007, opened 2018-12-04
Updated wordforms for Norwegian lemmatizer

beatesi:
## Description
Updated lists of wordforms for the Norwegian lemmatizer (nouns, verbs, adverbs, adjectives and lookup). Approximately 300K new wordforms added, and a few mistakes in the previous lists have been corrected. The updated lists of wordforms are extracted from the updated version of Norsk Ordbank in Norwegian Bokmål 2005 (updated 27/06/2018). 

### Types of change
Enhancement

## Checklist


ines:
Thanks a lot! 👍

(Btw, if you have some examples of words and their expected expected lemmatization – including more complex ones – it could be really cool to have a few lemmatization tests! This way we can always make sure that new updates don't break any language-specific functionality.)

mrJezy:
I'm having hard time to use this lemmatizer like for example the danish one. The lookup function's vocabulary is not too big. As far as I understand the words are stored here according the wordforms. Do you have to have the POS tag of your word to look up the lemmas of the words here?

beatesi:
The Norwegian lemmatizer works in the same way as the English one, and you are correct in guessing  that it uses the POS-tags of the words. 

I usually use it like this:
```
import spacy
nlp = spacy.load('nb_ud_ner')

doc = nlp(u"Dette er en test.")

for token in doc:
    print(token, token.lemma_)
```

but if you only want to use the Lemmatizer you can do like this:
```
from spacy.lemmatizer import Lemmatizer
from spacy.lang.nb import LEMMA_INDEX, LEMMA_EXC, LEMMA_RULES

lemmatizer = Lemmatizer(LEMMA_INDEX, LEMMA_EXC, LEMMA_RULES)
lemmas = lemmatizer(u'poteter', u'NOUN')
print(lemmas)
```
Hope this helps!

Issue #3012, opened 2018-12-05
Problem using doc array() and bytes() functions together


clippered:
## How to reproduce the behaviour
```
import spacy
from spacy.attrs import ENT_IOB, ENT_TYPE
from spacy.tokens import Doc


nlp = spacy.load('en_core_web_sm')
doc = nlp('This is 10%.')
print(doc[2].orth_, doc[2].pos_, doc[2].tag_, doc[2].ent_type_)  # displays "10 NUM CD PERCENT"

# removing '10%' entity
header = [ENT_IOB, ENT_TYPE]
ent_array = doc.to_array(header)
idx = 2
ent_array[idx, 0] = 0
ent_array[idx, 1] = 0
doc.from_array(header, ent_array)

print(doc[2].orth_, doc[2].pos_, doc[2].tag_, doc[2].ent_type_)  # displays "10 NUM CD "

# serializing then deserializing
bytes = doc.to_bytes()
doc2 = Doc(nlp.vocab).from_bytes(bytes)
print(doc2[2].orth_, doc2[2].pos_, doc2[2].tag_, doc2[2].ent_type_)  # displays "10   "
```

Basically, my use case would be to change some of the tagged entities while keeping other entities. From the small example above, it seems the problem occurs when calling the serialization/deserialization part after entity removal part. When you run it, it shows:
```
10 NUM CD PERCENT
10 NUM CD 
10
```

However, commenting out the entity removal part (the call to `to_array` up to `from_array`) of the code, the serialization works fine and shows:
```
10 NUM CD PERCENT
10 NUM CD PERCENT
10 NUM CD PERCENT
```

## Environment
* Operating System: OSX 10.14
* Python Version Used:  Python 3.6.6
* spaCy Version Used: spaCy 2.0.18
* Environment Information:

clippered:
Apologies if I might be missing some information.
In the example above, I want to keep the POS/TAG attributes after serialization. Something with an output of:
```
10 NUM CD PERCENT
10 NUM CD 
10 NUM CD 
```

honnibal:
It looks to me like the problem might be with the `is_tagged` attribute. If this is set to `False`, the POS tags are not serialized, leading to the mismatch we're seeing here.

honnibal:
Yep, if we do `doc.from_array([ENT_IOB, ENT_TYPE])`, it was seeing that no `TAG` is set, and then clobbering setting `doc.is_tagged` to false. This led us to not serialize the tag data.

[...]

PR #3021, opened 2018-12-06
issue #3012: add test

clippered:
Add regression test for issue#3012

Sorry, I can't get this test to run on my environment. I'm getting an error:
```
E           _pytest.warning_types.RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
E           Please use node.get_closest_marker(name) or node.iter_markers(name).
E           Docs: https://docs.pytest.org/en/latest/mark.html#updating-code
``` 

So please feel free to have a look at it too.

### DIFF
--- a/.github/contributors/clippered.md
+++ b/.github/contributors/clippered.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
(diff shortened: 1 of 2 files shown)


ines:
Thanks for the test! The warning indicates that tests aren't compatible with Pytest 4.x – we did update spaCy's tests to work with the latest Pytest version, but maybe your branch wasn't up to date when you ran the tests?

I ported this over to `develop` and made a few small adjustments (make it work without requiring a model, fix two small typos).

Issue #3011, opened 2018-12-05
French token_match not read correctly from disk ?


svlandeg:
## How to reproduce the behaviour
The string **entr'amis** should match in `TOKEN_MATCH` according to the regular expression defined in [fr/tokenizer_exceptions.py](https://github.com/explosion/spaCy/blob/master/spacy/lang/fr/tokenizer_exceptions.py#L143). This is also checked by a [unit test](https://github.com/explosion/spaCy/blob/master/spacy/tests/lang/fr/test_exceptions.py#L13). However I noticed that when the tokenizer is serialized to and back from bytes, this expression does not match anymore.

To reproduce in [tokenizer.pyx](https://github.com/explosion/spaCy/blob/master/spacy/tokenizer.pyx#L409):

```
if 'token_match' in data:
             print("controle C1 (from bytes)", self.token_match("entr'amis"), self.token_match)
             self.token_match = re.compile(data['token_match']).search
             print("controle C2 (from bytes)", self.token_match("entr'amis"), self.token_match)
```
The print-out shows that the original token_match field was matching our string, but the new, read version doesn't:

> ('controle C1 (from bytes)', <regex.Match object; span=(0, 9), match="entr'amis">, <built-in method match of _regex.Pattern object at 0x000000000B50C180>)
> ('controle C2 (from bytes)', None, <built-in method search of _regex.Pattern object at 0x000000001FA0A2D0>)

This causes downstream problems: a sentence with the word "entr'amis" in it will be tokenized by splitting up "entr" and "amis" into two words:

```
french_sentence = "je suis entr'amis"
nlp = spacy.load('fr_core_news_sm')
doc = nlp(french_sentence)
for token in doc:
        print("TOKEN", token.text, token.lemma_, token.pos_, token.tag_)
```
>TOKEN je il PRON PRON__Number=Sing|Person=1
>TOKEN suis être AUX AUX__Mood=Ind|Number=Sing|Person=1|Tense=Pres|VerbForm=Fin
>TOKEN entr' entr' ADJ ADJ__Gender=Masc|Number=Plur
>TOKEN amis ami NOUN NOUN__Gender=Masc|Number=Plur

 When I disable reading the `self.token_match` in the method `from_bytes` (in the code above), this is fixed:
> TOKEN je il PRON PRON__Number=Sing|Person=1
> TOKEN suis être AUX AUX__Mood=Ind|Number=Sing|Person=1|Tense=Pres|VerbForm=Fin
[...]

Issue #2679, opened 2018-08-16
Non-english (french) models a bit long to load


Kodoyosa:
Hello !

I'm using spacy since few months now. Mainly using downloadable pretrained language models. The problem is that until now I was using the default english model. But then now I want to use the french model (the default one is really not accurate so I currently use the `fr_core_news_md`). But my problem with french models is that it takes me almost half a minute just load those models (`spacy.load('fr')`) while it requires just a second for the english model.

I guess everybody can imagine how painfull it is to wait half a min everytime I run a test.

Does anyone knows why is it so long for these models to load compared to the english ones ?

Does anyone knows any tricks to not reload the models during development ?

Thanks for your help

## My Environment
* Operating System: OSX/10.13.6
* Python Version Used: 3.6.6
* spaCy Version Used: 2.0.9
* Environment Information:
  * **spaCy version:** 2.0.9
  * **Platform:** Darwin-17.7.0-x86_64-i386-64bit
  * **Python version:** 3.6.6
  * **Models:** fr_core_news_md, fr, en

ines:
I think a big problem with French at the moment is actually the language data, especially the [very complex tokenizer exceptions](https://github.com/explosion/spaCy/blob/master/spacy/lang/fr/tokenizer_exceptions.py) that are compiled and loaded (also see `_tokenizer_exceptions_list.py`). We've actually noticed this in our tests as well, and we'd  love to simplify those exceptions in favour of performance. 

One option could be to cut down the exception list to only the most common words and compile all variations (with different unicode apostrophes etc.) once, instead of doing it at runtime ([see here](https://github.com/explosion/spaCy/blob/master/spacy/lang/fr/tokenizer_exceptions.py#L100)). Another option would be to try to express more of this logic in the punctuation rules, and/or at least remove the words from the exceptions list that are already covered in the list.

I'll tag this issue with `help wanted` in case someone from the French spaCy community has some ideas and would like to contribute. We always appreciate help with this kind of stuff!

> Does anyone knows any tricks to not reload the models during development ?

In general, you definitely want to make sure to only load a model once per process and keep it in memory. For example, if you're running tests with pytest, provide the `nlp` object via a fixture thats session-scoped. But of course, this still means that you have to load it *once*.

[...]

PR #3023, opened 2018-12-07
French regular expressions instead of extensive exceptions list

svlandeg:
## Description
This post addresses https://github.com/explosion/spaCy/issues/2679.
* Looked into the French model and analysed where most time was lost
* Rewrote the compilation of _infixes_exc to avoid compilation of redundant patterns (from 270K to 100K)
* Added new regular expressions to catch common tokens
* Removed all redundant cases from the exception list, cutting it down from 34K lines to 16K
* Added a few additional unit tests

### Result
* Loading the fr_core_news_sm model on my system goes down on average from 12-13s to 8-9s 

### Remaining issues

* During testing, I stumbled upon https://github.com/explosion/spaCy/issues/3011 where I noticed that the regular expressions are first properly compiled, but then overwritten by a `from_disk `version of `token_match` in tokenizer.pyx. This means that the new regular expressions may not actually be used until the model from disk is updated (right?)
* 8-9s is probably still too long, compared to 1s for en_core_web_sm !

### Types of change
Speed enhancement

## Checklist


ines:
Wow, this is great news – thanks so much for looking into this! 

Also always love PRs that looks like this 😅
screenshot 2018-12-07 at 17 14 38

Since this PR introduces significant changes to the French language data, I wonder if we should move it over to the `develop` branch and only push it into the upcoming `v2.1.x` release. It shouldn't be too difficult to port over those changes to `develop` (most of what's different on that branch is auto-formatting via `black`, which we've started to adopt).

> This means that the new regular expressions may not actually be used until the model from disk is updated (right?)

Yes, that's correct. The idea here is that models should include their tokenization rules and allow shipping with custom rules. For example, if you've modified your `French` language class to include different rules, they'll be saved out when you save your model, and when you load back in, *your* rules are used instead of spaCy's default ones.

It's a bit unfortunate that we're currently setting up the language class and compiling all the default rules, only to then *replace* them by the model's deserialized rules. This can possibly make loading a model even slower for no reason – but not sure if there's an easy way aroune that at the moment.

svlandeg:
Thanks for the explanation Ines :) The 16K lines worth of exception cases do still slow the model down, indeed. 

I think it makes sense to keep this PR for the next release. Do you want me to rebase against the current develop branch?

ines:
Yeah, I also just realised that it won't make much sense for the `v2.0.x` family anyways, since we're not going to retrain the models!

> Do you want me to rebase against the current develop branch?

Yes, that would be great! I hope you won't be hitting any nast merge conflicts – if so, it's possible that it's easiest to just copy over the parts you've changed.

(Sorry about this btw – we ended up with `develop` being super far ahead of `master` again, because we had to delay `v2.1.x` due to all that infrastructure/[`blis`](https://github.com/explosion/cython-blis) stuff 😭 But we're super close to the release now!)

spacy/lang/fr/tokenizer_exceptions.py
```diff
@@ -76,7 +76,7 @@ def lower_first_letter(text):
                 {LEMMA: pronoun, ORTH: "-" + pronoun}]
 
 for verb, verb_lemma in [
-    ("est","être")]:
+    ("est", "être")]:
     for orth in [verb, verb.title()]:
```
mhham: Dictionary comprehension:
```python
_verb_list = [("est","être")]
_exc.update({
    "{}-ce".format(orth):[
             {LEMMA: verb_lemma, ORTH: orth, TAG: "VERB"},
             {LEMMA: 'ce', ORTH: '-ce'}
     ]
     for verb, verb_lemma in _verb_list
     for orth in [verb, verb.title()]
})
```

spacy/lang/fr/tokenizer_exceptions.py
```diff
@@ -87,64 +87,131 @@ def lower_first_letter(text):
 for pre, pre_lemma in [
     ("qu'", "que"),
     ("n'", "ne")]:
-    for orth in [pre,pre.title()]:
+    for orth in [pre, pre.title()]:
```
mhham: Dictionary comprehension:
```python
_pre_list = [("qu'", "que"), ("n'", "ne")]
_exc.update({
    '%sest-ce' % orth:[
            {LEMMA: pre_lemma, ORTH: orth, TAG: "ADV"},
            {LEMMA: 'être', ORTH: "est", TAG: "VERB"},
            {LEMMA: 'ce', ORTH: '-ce'}
    ]
    for pre, pre_lemma in _pre_list
    for orth in [pre, pre.title()]
})

File: spacy/lang/fr/tokenizer_exceptions.py [hunk 3]
```diff
@@ -87,64 +87,131 @@ def lower_first_letter(text):
...
-    for hyphen_char in ['-', '‐']:
-        _infixes_exc += [infix.replace("'", elision_char).replace('-', hyphen_char)
-                         for infix in FR_BASE_EXCEPTIONS]
-_infixes_exc += [upper_first_letter(word) for word in _infixes_exc]
-_infixes_exc = list(set(_infixes_exc))
+orig_elision = "'"
+orig_hyphen = '-'
+
+# loop through the elison and hyphen characters, and try to substitute the ones that weren't used in the original list
+for infix in FR_BASE_EXCEPTIONS:
```
mhham: As before, it might be worth it to try comprehensions to make things faster : (and also change FR_BASE_EXCEPTIONS from a list to a set)
```python
_infixes_exc = FR_BASE_EXCEPTIONS.copy() #FR_BASE_EXCEPTIONS needs to be a set
_infixes_exc.update([
    word.replace(orig_elision, elision_char) 
    for elision_char in [x for x in ELISION if x != orig_elision]
    for word in _infixes_exc
])
_infixes_exc.update([
    word.replace(orig_hyphen, hyphen_char) 
    for hyphen_char in [x for x in ['-', '‐'] if x != orig_hyphen
    for word in _infixes_exc
])
_infixes_exc.update([upper_first_letter(word) for word in _infixes_exc])
_infixes_exc = list(_infixes_exc)
```

Or even try the double nested loop within the comprehension ?
```python
_infixes_exc = {
   infix.replace("'", elision_char).replace('-', hyphen_char)
   for infix in FR_BASE_EXCEPTIONS 
   for elision_char in [x for x in ELISION if x != orig_elision]
   for hyphen_char in [x for x in ['-', '‐'] if x != orig_hyphen]
}
_infixes_exc.update([upper_first_letter(word) for word in _infixes_exc])
_infixes_exc = list(_infixes_exc)
```

svlandeg:
Good ideas @mhham ! I'll look into it whether that would make it faster :)

mhham:
Also, as a general remark, using sets (or dictionaries) instead of lists for exceptions would be faster

Issue #3018, opened 2018-12-06
New mecab-python3 version breaks Japanese


polm:
The python package used for Mecab, [mecab-python3](https://github.com/SamuraiT/mecab-python3), had a release for the first time since 2014 recently. Unfortunately one of the parts spaCy relies on is broken, see SamuraiT/mecab-python3#19. 

Since this is an optional dependency I don't think we can peg an old version for it, but maybe we should put a note somewhere if it's not fixed quickly. I will see what I can do about fixing it upstream.

## How to reproduce the behaviour

    import spacy
    ja = spacy.blank('ja')
    print(ja("a b c d"))
    # output: a b c db c dc dd

## Your Environment

* **spaCy version:** 2.0.18
* **Platform:** Linux-4.18.9-arch1-1-ARCH-x86_64-with-arch
* **Python version:** 3.7.1

honnibal:
Thanks for letting us know!

Maybe we could set this as an `extensions_require` dependency? Then you can version it. You could install spaCy with something like `spacy[mecab]`.

polm:
That sounds great! I'll figure out how to do that and submit a PR.

PR #3024, opened 2018-12-07
Extras require mecab

polm:
## Description

This just adds an `extras_require` for mecab-python3 at a working version. See #3018.

## Checklist


setup.py
```diff
@@ -217,7 +217,8 @@ def setup_package():
                 'cuda90': ['cupy-cuda90>=4.0', 'thinc_gpu_ops>=0.0.3,<0.1.0'],
                 'cuda91': ['cupy-cuda91>=4.0', 'thinc_gpu_ops>=0.0.3,<0.1.0'],
                 'cuda92': ['cupy-cuda92>=4.0', 'thinc_gpu_ops>=0.0.3,<0.1.0'],
-                'cuda100': ['cupy-cuda100>=4.0', 'thinc_gpu_ops>=0.0.3,<0.1.0']
+                'cuda100': ['cupy-cuda100>=4.0', 'thinc_gpu_ops>=0.0.3,<0.1.0'],
+                'mecab': ['mecab-python3==0.7']
```
ines: Looks good 👍

Maybe we could call the extra `ja`, after the language shortcode? This would also allow us to put more dependencies in there if needed (and add all other languages that depend on external packages in a similar way).

```suggestion
                'ja': ['mecab-python3==0.7']
```

polm: Sounds good to me!

ines:
Perfect, thanks! 👍

PR #3045, opened 2018-12-12
Lemmatization of Adjectives - French : adding rules and vocabulary

amperinet:
## Description

I added rules for French lemmatization of adjectives and removed one rule that was not necessary.  French rules have recently been added (see issue #2818), but for adjectives there were only three rules, with two of them overlapping.
It is not sufficient to cover all the morphological cases for adjectives.

The last version also misses a lot of general language French vocabulary. Thus, I added more vocabulary in both regular and irregular lists, to cover more general language in French. And I adapted the existing lists (regular/irregular cases) to fit both new and existing rules.

### Types of change

An enhancement:
* Modifying the existing two lists (regular and irregular) and adding words using the LEFFF linguistic resource ([web page](http://www.labri.fr/perso/clement/lefff/telechargement.html)).
 * Adding lemma rules for adjectives

## Checklist


amperinet:
Hello!
Thank you Amaury for your comment and for spotting some errors in the vocabulary files. I corrected the files and it should be ok now with only lemmas in _adjectives.py. 

Regarding the rules, I am sorry I think ‘overlap’ was not the right word. First, I only removed one rule, the "es" one. If you look carefully you will see that the "s" rule just moved down in the list of rules. 
I removed the "es" rule, because in all the French adjectives (your vocabulary + the adjectives I added from the LEFFF lexicon), adjectives ending by "es" mainly have a lemma ending by "e" (12060 cases of _"es" > "e"_ vs 4909 of _"es" > ""_). If we keep the _"es" > ""_ rule, we will have more vocabulary in the irregular list, and as you said, it’s better to have smaller files as possible.

I hope this is clearer now. Thank you !

mauryaland:
Hi!

You right, I didn't notice that the rule "s" -> "" was still there. Thank you for your explanation and again great work!

ines:
Thanks for the PR and appreciate the discussion 👍 I'll go ahead and merge this then. 

@mauryaland @amperinet Btw, in case you haven't seen it yet, you might also be interested in #3046, which simplifies how the tokenization rules are compiled and currently reduces loading time of `French` by ~30% on `develop`.

PR #3062, opened 2018-12-18
Added alpha support for Tagalog language

Brixjohn:
Added alpha support for Tagalog language

## Description

I have added alpha support for the Tagalog language from the Philippines. It is the basis for the country's national language Filipino. I have heavily based the format to the EN and ES languages.

I have provided several words in the lemmatizer lookup table, added stop words from a source, translated numeric words to its Tagalog counterpart, added some tokenizer exceptions, and kept the tag map the same as the English language.

While the alpha language passed the preliminary testing that you provided, I think it needs more data to be useful for most cases.

### Types of change
New feature

## Checklist


honnibal:
Thanks! Really appreciate this.

PR #3065, opened 2018-12-18
Fix the first `nlp` call for `ja` (closes #2901)

kbulygin:
## Description

The fix is according to https://github.com/explosion/spaCy/issues/2901#issuecomment-447872494. It wouldn't be possible without investigation made by [Paul O'Leary McCann](https://github.com/polm).

The fix itself is the most trivial (just adding `self.tokenizer.parseToNode('')` after `self.tokenizer = MeCab.Tagger()`). To apply the fix and to check that nothing has been broken (at least for Python 3.7.1), the following Dockerfile was used: https://github.com/kbulygin/spacy-dev/blob/master/Dockerfile.

### Types of change

Bug fix.

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #3078, opened 2018-12-20
Making `lang/th/test_tokenizer.py` pass by creating `ThaiTokenizer`

kbulygin:
## Description

After #3065, I tried to reduce the number of skipped tests by installing other optional dependencies. The status of the only test in `lang/th/test_tokenizer.py` switched from `skipped` to `failed` after installing the latest `pythainlp==1.7.1` (or `pythainlp==1.6.0.7` on Python 2). The reason is that the test essentially uses `ThaiDefaults.create_tokenizer`, which falls back to `Language.Defaults.create_tokenizer`. I implemented `ThaiDefaults.create_tokenizer` using code that was previously in `Thai.make_doc`. I used `JapaneseTokenizer` as a reference. While on that, I also did some linting.

There are a couple of things I'm not quite sure about:
- `class Japanese` contained `Tokenizer = JapaneseTokenizer`, but https://spacy.io/api/language#class-attributes doesn't list `Tokenizer` as a class attribute, so I removed the `Tokenizer = JapaneseTokenizer`.
- It's a bit strange that `create_tokenizer` is used only for `ja` and that I had to introduce `DummyTokenizer`. So I might have missed something. But if not, then at least `vi` and `zh` should be changed in the same way as `th`. (I can do that if that's OK.)
- It's also a bit strange that these things are missed in CI here. Perhaps there should be a `requirements-extra.txt` that lists optional dependencies such that:
  - users can do `pip install spacy[all]` (which behaves like `pip install -r <(cat requirements.txt requirements-extra.txt)`);
  - there's a CI target that tests this `spacy[all]` (and probably checks there are no more `ImportError`s: e.g., `pytest.importorskip` can be replaced by something like `__import__ if is_spacy_all else pytest.importorskip`).

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/lang/ja/__init__.py
+++ b/spacy/lang/ja/__init__.py
@@ -1,19 +1,21 @@
 # encoding: utf8
 from __future__ import unicode_literals, print_function
 
-from ...language import Language
-from ...attrs import LANG
-from ...tokens import Doc, Token
-from ...tokenizer import Tokenizer
-from .tag_map import TAG_MAP
-
 import re
 from collections import namedtuple
+
+from .tag_map import TAG_MAP
+
+from ...attrs import LANG
+from ...language import Language
+from ...tokens import Doc, Token
+from ...util import DummyTokenizer
 
 ShortUnitWord = namedtuple('ShortUnitWord', ['surface', 'lemma', 'pos'])
 
 # XXX Is this the right place for this?
 Token.set_extension('mecab_tag', default=None)
+
 
 def try_mecab_import():
     """Mecab is required for Japanese support, so check for it.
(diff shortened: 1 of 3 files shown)


kbulygin:
Well, if answering the part on the `spacy[all]` idea is overwhelming, I can create a separate issue for that.

honnibal:
@kbulygin 

Hey,

Sorry I let this slip over the holidays. Overall these all look like sensible changes. I think the idea of getting `spacy[all]` working in the CI is definitely worth exploring. The only catch I can see is that we might run into difficult platform problems, if we support extras that only work on some platforms. For instance, I could imagine a useful support for some language not working on Windows, or not working on Python 2.

Other answers:

> I removed the Tokenizer = JapaneseTokenizer.

Looks good.

> at least vi and zh should be changed in the same way as th. (I can do that if that's OK.)

Sure, I think you're right about this, the PR would be welcome.

PR #3079, opened 2018-12-20
[WIP] Enabling `tests/lang/ru/test_lemmatizer.py` and dealing with results

kbulygin:
## Description

After #3078, I continued to reduce the number of skipped tests. I installed `pymorphy2==0.8` (the latest PyPI release for both Python 2 and Python 3, published back in 2014) for `ru`. I removed occurences of `@pytest.mark.models('ru')` in `tests/lang/ru/test_lemmatizer.py`  because https://github.com/explosion/spaCy/blob/071789467e6d12c7a3ca4b154445f88b9de34e5a/CONTRIBUTING.md#adding-tests states: "Tests that require the model to be loaded should be marked with `@pytest.mark.models`" (and there are no loadable Russian models now). So tests in `test_lemmatizer.py` are no longer skipped.

On Python 3, all tests pass. But on Python 2, some "random" ones fail:

```
test_doc_lemmatization PASSED

test_ru_lemmatizer_noun_lemmas[\u0433\u0432\u043e\u0437\u0434\u0438\u043a\u0438-lemmas0] FAILED
test_ru_lemmatizer_noun_lemmas[\u043b\u044e\u0434\u0438-lemmas1] FAILED
test_ru_lemmatizer_noun_lemmas[\u0440\u0435\u043a\u0438-lemmas2] FAILED
test_ru_lemmatizer_noun_lemmas[\u043a\u043e\u043b\u044c\u0446\u043e-lemmas3] PASSED
test_ru_lemmatizer_noun_lemmas[\u043f\u0435\u043f\u043f\u0435\u0440\u043e\u043d\u0438-lemmas4] PASSED

test_ru_lemmatizer_works_with_different_pos_homonyms[\u0440\u043e\u0439-NOUN-None-\u0440\u043e\u0439] PASSED
test_ru_lemmatizer_works_with_different_pos_homonyms[\u0440\u043e\u0439-VERB-None-\u0440\u044b\u0442\u044c] FAILED
test_ru_lemmatizer_works_with_different_pos_homonyms[\u043a\u043b\u0435\u0439-NOUN-None-\u043a\u043b\u0435\u0439] PASSED
test_ru_lemmatizer_works_with_different_pos_homonyms[\u043a\u043b\u0435\u0439-VERB-None-\u043a\u043b\u0435\u0438\u0442\u044c] FAILED
test_ru_lemmatizer_works_with_different_pos_homonyms[\u0442\u0440\u0438-NUM-None-\u0442\u0440\u0438] PASSED
[...]

PR #3084, opened 2018-12-21
Enabling `tests/lang/ru/test_lemmatizer.py`, fixing a `unicode` issue

kbulygin:
## Description

See #3079. Here I'm merging into `develop` instead of `master`.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/lang/ru/lemmatizer.py
+++ b/spacy/lang/ru/lemmatizer.py
@@ -1,6 +1,13 @@
 # coding: utf8
+from __future__ import unicode_literals
+
 from ...symbols import ADJ, DET, NOUN, NUM, PRON, PROPN, PUNCT, VERB, POS
 from ...lemmatizer import Lemmatizer
+
+try:
+    Str = unicode
+except NameError:
+    Str = str
 
 
 class RussianLemmatizer(Lemmatizer):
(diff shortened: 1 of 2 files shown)


spacy/lang/ru/lemmatizer.py
```diff
@@ -1,7 +1,14 @@
...
+from __future__ import unicode_literals
+
 from ...symbols import ADJ, DET, NOUN, NUM, PRON, PROPN, PUNCT, VERB, POS
 from ...lemmatizer import Lemmatizer
 
+try:
+    Str = unicode
+except NameError:
+    Str = str
+
```
ines: To make this more elegant, you can also import our `unicode_` helper from `spacy.compat` – [see here](https://github.com/explosion/spaCy/blob/develop/spacy/compat.py#L51). Because that type of hacky compatibility check comes up a lot, we've added some helpers for it 🙂

For example:

```python
from ...compat import unicode_

if isinstance(univ_pos, unicode_):
   # and so on
```

kbulygin: Thanks for the feedback. At first I wanted to use `six` for that, but since `six` not specified in `requirements.txt`, I used this manual hack. It's fixed now according to your suggestion.

Issue #2396, opened 2018-05-30
get_lca_matrix is incorrect


slavi:
@ramananbalakrishnan 

In spacy version 2.0.11

> s = u"He created a test for spacy"
> doc = nlp(s)
> doc.get_lca_matrix()[3][5]

Expected: 3
Actual: 1

This branch of the recursion is incorrect:

https://github.com/explosion/spaCy/blob/master/spacy/tokens/doc.pyx#L732

It assumes that if n1 and n2 are not adjacent, then LCA(n1, n2) = LCA(n1.head, n2.head).

This is generally not true: there are many counterexamples, the above one with 3 nodes in a chain is the simplest.

honnibal:
Thanks for the report, and yes -- a pull request would be much appreciated!

alvaroabascar:
I reimplemented the function to solve this issue, PR [here](https://github.com/explosion/spaCy/pull/3089).

PR #3089, opened 2018-12-21
Fix issue 2396

alvaroabascar:
## Description
This PR fixes issue #2396: `Doc.get_lca_matrix` had a bug in [this line](https://github.com/explosion/spaCy/blob/master/spacy/tokens/doc.pyx#L739) which caused wrong outputs
for some pairs of tokens.

### Types of change
Bug fix

### Explanation of the bug

I borrow the following example from @slavi (#2396):

He created a test for spacy

```python3
s = "He created a test for spacy"
doc = nlp(s)
doc.get_lca_matrix()[3][5]
```

- Expected output: 3
- Actual output: 1

The above mentioned line is:

```python3
...
            else:
                lca_index = __pairwise_lca(token_j.head, token_k.head, lca_matrix)
...
```

Here, for tokens 3 ("test") and 5 ("spacy"), we are calling `__pairwise_lca` recursively with their heads ("created" and "for" respectively). But in doing so, we are skiping "test", which is the actual lowest common ancestor.

### Solution

The reimplementation consists on [this simple algorihtm](https://thimbleby.gitlab.io/algorithm-wiki-site/wiki/lowest_common_ancestor/). It works as expected and is 3 times faster.

## Checklist

### DIFF
--- a/.github/contributors/alvaroabascar.md
+++ b/.github/contributors/alvaroabascar.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
+documentation, or any other material posted or submitted by you to the project.
+
+2. With respect to any worldwide copyrights, or copyright applications and
+registrations, in your contribution:
+
+    * you hereby assign to us joint ownership, and to the extent that such
+    assignment is or becomes invalid, ineffective or unenforceable, you hereby
+    grant to us a perpetual, irrevocable, non-exclusive, worldwide, no-charge,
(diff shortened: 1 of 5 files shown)


honnibal:
Really nice first PR, thanks!

I noticed some ugliness with the original code as well --- unfortunately the implementation appears to be duplicated in the `spacy/tokens/span.pyx` module. It would probably be good to move the code in `spacy/tokens/doc.pyx` into functions so that `span.pyx` can call into it. If that ends up being tricky, we can also just punt on that and just apply the same fix to the `span.pyx` code, and let the implementations continue in parallel for now.

alvaroabascar:
True! I hadn't seen the duplicated implementation inside `span.pyx`. I can either:

a) Separate the function and call it from both `Span.get_lca_matrix` and `Doc.get_lca_matrix`.
b) Implement the code in `Span.get_lca_matrix`, and from `Doc.get_lca_matrix` just call `self[:].get_lca_matrix()`.

b) seems cleaner to me. But in case you'd prefer a), where would you place this auxiliary function?

honnibal:
Delegating to a span of the whole document is a clever solution, but I think I probably prefer a helper function in `doc.pyx`. I probably wouldn't expect to find the implementation within `span.pyx`, since the `Span` is a view of the doc, so normally the delegation goes the other way.

Moving this out into a helper function also lets us fix that closure, which I always found a bit ugly about the previous implementation. So, we want the function to be something like:

```# within spacy/tokens/doc.pyx

def _get_lca_matrix(Doc doc, int start, int end):
```

We should keep it as a private function so that we don't have to worry about validating the start and end offsets within it. From the span, if we call `_get_lca_matrix(self.doc, self.start, self.end)` we know the start and end are valid.

alvaroabascar:
Done! I first tried to just define a helper `_get_lca_matrix` inside `doc.pyx`, but could not import it from `span.pyx`. It finally worked when I declared it in `doc.pxd`. I'm not experienced in Cython so I hope the approach is right.

I also removed that closure within `_get_lca_matrix`, and I defined it as another helper `_get_tokens_lca`, which made me wonder if that helper should go in `doc.pyx` or `token.pyx`.

Finally,  I guess that both helpers could be further "cythonized" resorting to TokenC, but this would mean renouncing to `token.ancestors`, which makes everything pretty clean.

Let me know what you think :)

honnibal:
Looks great, thanks! I think keeping that helper in the `doc.pyx` is fine. It's nice to have the helper close to where it's used. I think you're right that keeping it as Python so we can use the ancestors property is better.

honnibal:
Hm, the span test is failing, though. Is the test wrong or the code?

alvaroabascar:
It was a mistake in the code, but it should be fine now.

honnibal:
Sweet, merging.

PR #3104, opened 2018-12-28
French lemmatization: correcting wrong lemmas in the lookup dictionnary

amperinet:
## Description

I corrected some lemmas (316 modified, 3 lines removed) in the French lookup dictionnary and corrected accordingly some errors in the adjectives vocabulary file. There were lemmas with a wrong spelling (ex: "éluer" does not exist in French, it is "élu") and some lemmas were word forms, instead of the lemmas (mainly conjugated verbs were conjugated and not in the infinitive mode. Ex; "nageant" instead of "nager").  

I also spotted some errors in the nouns vocabulary files (that I modified in #2992) ; similar errors to adjectives (#3045): there were some word forms (plural/feminine) in the lemma file _nouns.py. Also, some nouns were missing in the irregular list. 

I simplified/grouped some rules for nouns ; in #2992, I added too many rules that can be summed up. 

I corrected wrong line breaks on hyphens (in verbs, adjectives and nouns lemma files). 

### Types of change

An enhancement. 

## Checklist


spacy/lang/fr/lemmatizer/_adjectives_irreg.py
```diff
@@ -51186,5 +51184,5 @@
     "zygotique": ("zygotique",),
     "zymogène": ("zymogène",),
     "zymonucléique": ("zymonucléique",),
-    "zymotique": ("zymotique",)
+    "zymotique": ("zymotiqueg",)
```
mauryaland: I guess there is a small mistake here with "zymotiqueg"

amperinet: yes thanks! :) Corrected.

ines:
Thanks, merging! (Also really like the French collaboration going on here 🥰)

PR #3110, opened 2019-01-01
Bugfix/get lca matrix

alvaroabascar:
This PR adds a test for an untested case of `Span.get_lca_matrix`, and fixes a bug for that scenario, which I introduced in [this PR](https://github.com/explosion/spaCy/pull/3089) (sorry!).

## Description
The previous implementation of get_lca_matrix was failing for the case `doc[j:k].get_lca_matrix()` where `j > 0`. A test has been added for this case and the bug has been fixed.

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/tests/doc/test_doc_api.py
+++ b/spacy/tests/doc/test_doc_api.py
@@ -278,10 +278,30 @@ def test_doc_api_similarity_match():
     assert doc.similarity(doc2) == 0.0
 
 
-def test_lowest_common_ancestor(en_tokenizer):
-    tokens = en_tokenizer('the lazy dog slept')
-    doc = get_doc(tokens.vocab, [t.text for t in tokens], heads=[2, 1, 1, 0])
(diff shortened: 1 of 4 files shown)


alvaroabascar:
Latest commit improves greatly the speed when the `doc` is big and contains several sentences. Tested on a text with 770 tokens and 74 sentences, it works 35 times faster. The assumption used is that two tokens never share an ancestor if they belong to different sentences.

honnibal:
Nice! Definitely agree with that assumption.

honnibal:
I think this is safe to merge, right? It looks like Appveyor failed for no good reason, possibly due to the outages with Github a couple of days ago?

alvaroabascar:
Yes, it seemed to me like a lack of cloud resources to run the tests (`Cannot assign job to clouds/groups 'pro, pro-backup'. All clouds within the group are either failing, or offline, or busy.`).

Tests pass locally, so I think it's safe to merge.

PR #3118, opened 2019-01-04
Joblib site has moved.

alvations:
The joblib link on https://spacy.io/usage/processing-pipelines#multi-processing-example is broken. Changed it to the official site on https://joblib.readthedocs.io/en/latest/parallel.html


alvations:
If it's the pypi link, then it should be https://pypi.org/project/joblib/ but I think linking directly to the maintainers' site would be more helpful =)

ines:
Thanks! 👍

Issue #2894, opened 2018-11-01
cuda10 installation not working


kwhumphreys:
docs state to use `spacy[cuda10]` for installation, but code expects `cuda100` https://github.com/explosion/spaCy/blob/e2ae25d6f55676ecc0f27c2dc1f6fe0b35e729d0/setup.py#L208
however `pip install -U spacy[cuda100]` then fails with
```
Collecting cupy-cuda100>=4.0; extra == "cuda100" (from spacy[cuda100])
  Could not find a version that satisfies the requirement cupy-cuda100>=4.0; extra == "cuda100" (from spacy[cuda100]) (from versions: )
```

## Which page or section is this issue related to?

https://spacy.io/usage/#gpu

honnibal:
I expected `cupy` to support CUDA 10 quickly, so I wrote the extra version qualifier in there so we'd be ready. Unfortunately it seems like supporting the new CUDA is more difficult than I expected, so there's still no cupy support for it. So, currently you have to use CUDA 9.2. Sorry for the inconvenience.

moreymat:
@honnibal CUDA 10 wheels for cupy are now available (released on Dec. 3): https://pypi.org/project/cupy-cuda100/ .

I installed thinc with that wheel (`pip install thinc_gpu_ops thinc[cuda100]`) and compiled the develop branch for spaCy, it works fine on my laptop.
You should be able to restore CUDA 10 support for spaCy :-)

PR #3126, opened 2019-01-07
Support CUDA 10

moreymat:
## Description
This PR adds, or rather restores, support for CUDA 10 now that cupy wheels are available : https://pypi.org/project/cupy-cuda100/ .

### Types of change
Support for CUDA 10 was introduced then withdrawn because wheels were not available at the time #2894 .

## Checklist


honnibal:
Yay, thanks!

PR #3139, opened 2019-01-10
Update call to `mkdir()` to create the parents

retnuh:
- Update the call to `output_dir.mkdir()` to also create the parents if needed

## Description

Very minor enhancement to hopefully save someone else the pain of half an hour training only to have it bomb out when saving :(

### DIFF
--- a/.github/contributors/retnuh.md
+++ b/.github/contributors/retnuh.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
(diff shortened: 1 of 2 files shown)


retnuh:
Surely setting a boolean flag doesn't require a licensing agreement to be signed?

honnibal:
I don't prefer this change actually. If you typo the directory you want to output to, you can end up with things in a weird place. I think it's more reasonable to raise an error if the parent directory is not found.

All pull requests to the library do need the CLA to be signed, even for minor changes. Otherwise we end up in a situation where the rights to the code are unclear, due to an accumulation of small changes from lots of people.

retnuh:
How about checking/making the directory _before_ the training cycle?  It
sucks to spend an hour training a model, expecting the directory to be made
for you, only to have to explode at the very end.

I don't have time to read through the agreement right now, but if you're
amenable to the above idea, I'd be happy to change the PR.

H

retnuh:
the call to `mkdir()` will raise the following:

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/foo/bar/bazz/spoo/asdf'
```

I can wrap that with a value error if you like, which would look something like this:

```
ValueError: Can't create output directory /tmp/foo/bar/bazz/spoo/asdf: [Errno 2] No such file or directory: '/tmp/foo/bar/bazz/spoo/asdf'
```

honnibal:
Thanks! Looks good.

PR #3155, opened 2019-01-13
Allow vectors to be optional in init-model, more robust string counting

DeNeutoy:
## Description
- Vectors are optional (as per the docs) but `init-model` crashes if they are not present
- Added a more graceful fallback to strings which are e.g a part of the Python grammar, making "ast.literal_eval(x)" break due to interpretation as code rather than strings.

### Types of change
Bugfix

## Checklist


DeNeutoy:
@ines This is ready for review/merging now, thanks!

honnibal:
Thanks!

PR #3173, opened 2019-01-20
💫 Update matcher engine for regex and extensions

### LINKED ISSUE #1971: 💫 Better, faster and more customisable matcher

honnibal:
Add support for matching over arbitrary Python predicate functions, and
arbitrary Python attribute getters. This will allow matching over regex
patterns, and allow supporting extension attributes.

The results of the Python predicate functions are cached, so that we don't
call the same predicate function twice for the same token. The extension
attributes are fetched into an array for each token in the doc. This
should minimise the performance impact of the new features.

This patch tips the `matcher.pyx` file over the 1000 line mark :(. We should make
it a subpackage. We also need better pattern validation and errors.

Closes #1971

## Checklist

### DIFF
--- a/spacy/matcher.pyx
+++ b/spacy/matcher.pyx
@@ -2,7 +2,7 @@
 # cython: profile=True
 from __future__ import unicode_literals
 import re
-import json
+import srsly
 from libcpp.vector cimport vector
 from libc.stdint cimport int32_t, uint64_t, uint16_t
 from preshed.maps cimport PreshMap
@@ -579,7 +579,7 @@ def _get_extra_predicates(spec, extra_predicates, seen_predicates):
         if isinstance(value, dict):
             for type_, cls in predicate_types.items():
                 if type_ in value:
-                    key = (attr, type_, json.dumps(value[type_], sort_keys=True))
+                    key = (attr, type_, srsly.json_dumps(value[type_], sort_keys=True))
                     # Don't create a redundant predicates.
                     # This helps with efficiency, as we're caching the results.
                     if key in seen_predicates:
@@ -599,7 +599,7 @@ def _get_extension_extra_predicates(spec, extra_predicates, predicate_types,
         if isinstance(value, dict):
             for type_, cls in predicate_types.items():
                 if type_ in value:
-                    key = (attr, type_, json.dumps(value[type_], sort_keys=True))
+                    key = (attr, type_, srsly.json_dumps(value[type_], sort_keys=True))
                     if key in seen_predicates:
                         output.append(seen_predicates[key])
                     else:


honnibal:
@ines Everything should be working here, but the code's pretty rushed. In particular, I'm not doing anything to verify that the patterns are reasonable, so we'll get things dying in random places. All of the following are invalid for various reasons:

* `{"LENGTH": {"REGEX": "[4-9]"}}` --  Regex requires string valued attribute, length is int valued.
* `{"ORTH": {">=": "apple"}}` -- Comparators require int or float valued attribute, not string.
* `{"_": {"elmo_vector": {"IN": [0.0]}}}` -- `IN` requires either string or integer, extension attributes might return arbitrary types.
* I'm not really verifying that the predicates are valid (currently supported ones are `REGEX`, `IN, `NOT_IN`, comparators). If they pass something arbitrary there, the error might be confusing.
* I'm not checking that we don't get something like `{"ORTH": {"hello": "bye"}}`. I guess this wasn't really checked before.

I think we might consider separating the validation logic from the pattern parser? This might make it easier to write errors.

ines:
> We should make it a subpackage. 

Yeah, then we could have `matcher/matcher.pyx`, `matcher/phrasematcher.pyx` and `matcher/entityruler.pyx`.

> We also need better pattern validation and errors.

Since we're already depending on `jsonschema` in `v2.x`, we could now validate this using a schema? If a pattern is added, it's validated by default (and the behaviour could be turned off by setting `validate=False` on intialization of the matcher, just in case.)

Here's a quick schema suggestion that should roughly describe the constraints (untested):

```json
{
  "$schema": "http://json-schema.org/draft-06/schema",
  "definitions": {
    "string_value": {
      "anyOf": [
        { 
          "type": "string" 
        },
        {
          "type": "object",
          "properties": {
            "REGEX": {
              "type": "string"
            },
            "IN": {
              "type": "array",
              "items": {
                "type": "string"
              }
            },
            "NOT_IN": {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          },
          "additionalProperties": false
        }
      ]
    },
    "integer_value": {
      "anyOf": [
        {
          "type": "integer",
        },
        {
          "type": "object",
          "properties": {
            "REGEX": {
              "type": "string"
            },
            "IN": {
              "type": "array",
              "items": {
                "type": "integer"
              }
            },
            "NOT_IN": {
              "type": "array",
              "items": {
                "type": "integer"
              }
            },
            "==": {
              "type": "integer"
            },
            ">=": {
              "type": "integer"
            },
            "<=": {
              "type": "integer"
            },
            ">": {
              "type": "integer"
            },
            "<": {
              "type": "integer"
            }
          },
          "additionalProperties": false
        }
      ]
    },
    "boolean_value": {
      "type": "boolean"
    },
    "underscore_value": {
      "anyOf": [
        {"type": ["string", "integer", "number", "array", "boolean", "null"]},
        {
          "type": "object",
          "properties": {
            "REGEX": {
              "type": "string"
            },
            "IN": {
              "type": ["string", "integer"]
            },
            "NOT_IN": {
              "type": ["string", "integer"]
            },
            "==": {
              "type": "integer"
            },
            ">=": {
              "type": "integer"
            },
            "<=": {
              "type": "integer"
            },
            ">": {
              "type": "integer"
            },
            "<": {
              "type": "integer"
            }
          },
          "additionalProperties": false
        }
      ]
    }
  },
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "ORTH": {
        "title": "Verbatim token text",
        "$ref": "#/definitions/string_value"
      },
      "TEXT": {
        "title": "Verbatim token text (spaCy v2.1+)",
        "$ref": "#/definitions/string_value"
      },
      "LOWER": {
        "title": "Lowercase form of token text",
        "$ref": "#/definitions/string_value"
      },
      "POS": {
        "title": "Coarse-grained part-of-speech tag",
        "$ref": "#/definitions/string_value"
      },
      "TAG": {
        "title": "Fine-grained part-of-speech tag",
        "$ref": "#/definitions/string_value"
      },
      "DEP": {
        "title": "Dependency label",
        "$ref": "#/definitions/string_value"
      },
      "LEMMA": {
        "title": "Lemma (base form)",
        "$ref": "#/definitions/string_value"
      },
      "SHAPE": {
        "title": "Abstract token shape",
        "$ref": "#/definitions/string_value"
      },
      "ENT_TYPE": {
        "title": "Entity label of single token",
        "$ref": "#/definitions/string_value"
      },
      "LENGTH": {
        "title": "Token character length",
        "$ref": "#/definitions/integer_value"
      },
      "IS_ALPHA": {
        "title": "Token consists of alphanumeric characters",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_ASCII": {
        "title": "Token consists of ASCII characters",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_DIGIT": {
        "title": "Token consists of digits",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_LOWER": {
        "title": "Token is lowercase",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_UPPER": {
        "title": "Token is uppercase",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_TITLE": {
        "title": "Token  is titlecase",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_PUNCT": {
        "title": "Token is punctuation",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_SPACE": {
        "title": "Token is whitespace",
        "$ref": "#/definitions/boolean_value"
      },
      "IS_STOP": {
        "title": "Token is stop word",
        "$ref": "#/definitions/boolean_value"
      },
      "LIKE_NUM": {
        "title": "Token resembles a number",
        "$ref": "#/definitions/boolean_value"
      },
      "LIKE_URL": {
        "title": "Token resembles a URL",
        "$ref": "#/definitions/boolean_value"
      },
      "LIKE_EMAIL": {
        "title": "Token resembles an email address",
        "$ref": "#/definitions/boolean_value"
      },
      "_": {
        "title": "Custom extension token attributes (token._.)",
        "type": "object",
        "properties": {
          "*": {
            "$ref": "#/definitions/underscore_value"
          }
        }
        
      }
    },
    "additionalProperties": false
  }
}
```

honnibal:
@ines +1, looks good!

Do we want to merge this and reserve the reorg and schema validation for a subsequent PR? That'll let us get this out to the nightly.

Since we squash commits, I also want to avoid making the reorg in the same PR as these logic changes, since we won't be able to see the diff for this patch once we do the reorg.

PR #3180, opened 2019-01-21
Fix french lemmatization

foufaster:
This branch fixes an error in the handling of some words in French like "je" and "me".
The error is caused by an expected return type by FrenchLemmartizer. This callable class is supposed to return a list of strings but sometimes returns a list of a tuple containing one string.


honnibal:
Thanks!

PR #3185, opened 2019-01-22
Improving the French lookup dictionnary for ambiguous words

amperinet:
## Description

With the new French lemmatizer, even using the rules the lemmatizer looks into the lookup dictionnary as a last resort. Thus, the quality of the lookup dictionnary is important.
In the current dictionnary, there are many ambiguous words, words that can have different lemmas according to the context/how they are used.  I analysed those cases and made different corrections explained below.

Ambiguous words ; a word form that can have multiple lemmas ; 
- different lemmas and different POS: 
    "plus" => plaire_VERB (_appeal_) and plus_ADV (_more_)
    "mentons" => mentir_VERB (_lie_) and menton_NOUN (_chin_)
- different lemmas for a same POS: 
  "cours_NOUN"  has 2 lemmas => "cour" (_schoolyard_/_courtyard_) or "cours" (_lesson_/_class_)
  "suis_VERB" has 2 lemmas => "suivre" (_follow_) and "être" (_be_)
  "mises_VERB" has 2 lemmas => "miser" (_bet_/_invest_) and "mettre" (_put_/_place_)

In the lookup dictionnary we don't have the POS information, but only a pair "word form/lemma". In case of multiple lemmas for a word form, I checked in all the data available in spacy_fr and chose the "less problematic lemma" when possible, or I completely removed the entry if it is too ambiguous and impossible to choose for only one lemma (in that case, I think it is better to stay with the string, keeping the ambiguity). 

The choices I made to improve the dictionnary in case of multiple lemmas for a word form:
- if the word form has one lemma with a similar spelling (word form = lemma ; ex: "plus / plus_ADV") ; I kept this spelling. That would be a similar option to removing the entry from the dictionnary, and lemmatising keeping the string. 
- if no similar spelling is found within the lemmas; 
     - when there is no big or little meaning shift between the different lemmas (lemmas bear the same general meaning), I kept the most frequent lemma for that word form. Mainly, some words ending with "s" in French can be both nouns and verbs ;  usually, they more frequently correspond to the plural form of the noun (ex: ancres > ancre_NOUN or ancrer_VERB (_anchor_ verb and noun)). 
    - when there is a big meaning shift, too many different lemmas when or the different lemmas are used with similar frequency in French, I removed this entry (ex: marines => mariner_VERB (_marinate_), marin_ADJ (_marine_), marine_NOUN (_navy_)). 

### Types of change

An enhancement made to the lookup dictionnary:
- 276 entries removed
- 5427 entries modified (change of the lemma)

An enhancement to nouns/verbs/adjectives word lists: I added vocabulary that was in the lookup dictionnary but not in the vocabulary files. 

## Checklist


ines:
Thanks for all your work on French, much appreciated! 👍

amperinet:
Hi Ines and Matthew,
Is it possible to do a new relesase with all the last modifications in French so that we can use it in our pipeline ? 

Thanks a lot! 
Amandine

PR #3186, opened 2019-01-22
Fixed tag map for Swedish Talbanken

boena:
## Description
Added tag map to map for UD Swedish Talbanken.

### Types of change
Enhancement to Swedish Language

## Checklist


boena:
I have one question though, we were experimenting a little with the morph abilities but since the tags are quite narrow we wen't with this approach. Is it however possible to add the morph section in the mapping? 

Something like this for example:
`'DT|MAS|SIN|DEF': { "morph": "Definite=Def|Gender=Masc|Number=Sing", POS: DET }` 

@ines @honnibal

PR #3200, opened 2019-01-27
Added functionality to recognise Dutch numerals

PolyglotOpenstreetmap:
## Description

I'm new to this, so I was struggling a bit, but now all the tests pass. I also refactored it a bit, to make it as efficient as possible.

### Types of change
enhancement of recognition of Dutch numerals, which can be written as single compounded words like negenhonderdnegenennegentigduizend(ste) (999000). I'm not sure how to go about grouping them if they need to be separated by a space, like achthonderdduizend achthonderdachtentachtig (800888) or drie zevende (3/7). driekwart is included, as it can be written as a single word or as "drie kwart" as well.

een kwart (1/4), een half, één tweede (1/2), drie achtste (3/8) always need to be written separated.

I'm still trying to figure out if this file is the right place to add this in the pipeline. The code can differ between ordinal and cardinal numbers, but this information is lost now. It also returns the actual value the numeral represents.

## Checklist


spacy/lang/nl/lex_attrs.py
```diff
@@ -3,38 +3,198 @@
...
+DUIZEND = "duizend"
 
-_ordinal_words = set("""
-eerste tweede derde vierde vijfde zesde zevende achtste negende tiende elfde
-twaalfde dertiende veertiende twintigste dertigste veertigste vijftigste
-zestigste zeventigste tachtigste negentigste honderdste duizendste miljoenste
-miljardste biljoenste biljardste triljoenste triljardste
-""".split())
+wrong_ordinals_re = regex.compile(r"""(?x)
+.*(?P<wrong>nulste|eende|tweeste|driede|vierste|vijfste|zeste|zesste|achtde|tienste|honderdde|duizendde)
```
svlandeg: In my dialect, we may actually say "tienste" instead of "tiende". Also found a vague reference here: https://www.encyclo.nl/begrip/tienste 
So maybe excplicitely including it as wrong is not necessary?

spacy/lang/nl/lex_attrs.py
```diff
@@ -3,38 +3,198 @@
...
+$""")
+
+hundreds_units_and_tens_thousand_re = regex.compile(r"""(?x)
+      (
+        (?P<hundreds>twee|drie|vier|vijf|zes|zeven|acht|negen)honderd
+        |
+        (?P<hundred>honderd)
```
svlandeg: I would put `(twee|...|negen)?honderd` and drop the `|(...honderd)` at the end

spacy/lang/nl/lex_attrs.py
```diff
@@ -3,38 +3,198 @@
...
+      )*
+      (
+        (?:
+          (?P<units>eene|tweeë|drieë|viere|vijfe|zese|zevene|achte|negene)
+          n
+        )*
```
svlandeg: Why does the `eene|...|negene` alternation end with a `*` instead of a `?`
You don't want to repeat "eenen", right?
Also, I think it's actually more readable to add the n to each option.

spacy/lang/nl/lex_attrs.py
PolyglotOpenstreetmap: They are verbose regular expressions, spaces are ignored, unless added as explicit r"\s"

What about eenste and drieste? :-) And I removed zevenste and negenste because they are valid in my own dialect, but I'm not convinced that was the right thing to do. Once decided whether to accept them or not, I should definitely add them to the test cases.

I noticed that in another pull request there was a move away from regex to re. Is it better to use re when possible? It shouldn't pose a problem for Dutch.

spacy/lang/nl/lex_attrs.py
PolyglotOpenstreetmap: You are right, I'll implement your suggestions. I know a bit about regular expressions, but there is obviously always more to learn.

spacy/lang/nl/lex_attrs.py
PolyglotOpenstreetmap: I did it the good old fashioned Belgian way: a compromise. I added a parameter. Those who want to adhere strictly to Dutch AN spelling, can set strict_AN_spelling to True.
I removed some of the edge cases from the unit tests.
Thanks for your review, it's really appreciated!

PolyglotOpenstreetmap:
Hi @ines,

I think it's done now. I'd like to figure out how to go about processing several of these like_num tokens when they appear consecutively in a phrase to determine the value they stand for, like for example:

één miljard driehonderdduizend miljoen vijfhonderddrieëntachtigduizend hondervierentwintig en een half.

1 300 583 124.5

PolyglotOpenstreetmap:
Please let me know if there is something I need to do for this pull request to be accepted. In the meantime, I created one for German and I'm working on one for French. I didn't submit the German one, because:

1. I can't figure out how to make the tests pass. (When I create independent tests locally, they do pass)
2. I don't know how to isolate it from this pull request.

ines:
@PolyglotOpenstreetmap Thanks for your work on this and sorry for only getting back to you now! I had to think about this, because we currently don't have a good plan for integrating something like this, and it's quite a bit of specific code that's still experimental in places.

But I have an idea: How about you develop this as a spaCy plugin? This would let you publish it as a separate package people could install alongside spaCy, and you could set various custom attributes on tokens, like `token._.converted_num`. You can put it on PyPi and start publishing alpha and beta releases straight away, we can add it to the [spaCy universe](https://spacy.io/usage/processing-pipelines#section-custom-components-attributes) and people can start playing with it. Plus, it's your package and you'll get the proper credit for it! Once it's stable and working well, we can still talk about integrating it into the core library 😃 Here are more details on extension attributes: https://spacy.io/usage/processing-pipelines#section-custom-components-attributes

I think this would actually be a super cool example of a plugin and pretty straightforward to do. A nice thing about spaCy's `Token` objects is that they expose a `lang` and `lang_` attribute with the language code of the `nlp` object used to process the doc. So you could have one token extension getter that handles all languages supported by the plugin:

```python
def get_converted_num(token):
    if token.lang_ == "nl":
        return compute_dutch_number(token.text)
    elif token.lang_ == "de":
        return compute_german_number(token.text)
    # and so on...

from spacy.tokens import Token
Token.set_extension('converted_num', getter=get_converted_num)
```

This makes tokens expose a `._.converted_num` attribute which could be used like this:

```python
doc = nlp("bla bla 999000")
doc[2]._.converted_num  # negenhonderdnegenennegentigduizend
```

For even more flexibility, you could also add an attribute like this on `Span` objects for multi-token numbers like "één tweede". These are all really nice features that are way beyond the scope of spaCy's token-based `LIKE_NUM` getter. So I really think this would make much more sense as a plugin!

PolyglotOpenstreetmap:
I'm working on this. I'll let you know when/if I get stuck. I'll probably need a hand with it, but right now I have to focus on getting some other stuff done.

ines:
@PolyglotOpenstreetmap Sure, take your time! I'll close this issue in the meantime, just so it's clear that it's not pending in the current state and will be implemented as a plugin instead. If you have any questions about the plugin or want to share it for review, feel free to open a new issue 🙂

PR #3217, opened 2019-01-31
Allow util.minibatch to accept float values for size

theudas:
## Description
I was training a text-classifaction in Spacy this week, when I got an error for passing a float to the minibatch size.
It took me a while to figure out where I made my mistake and want to prevent others to run into the same issue.

Example:

```
batch_sizes = compounding(1, 64, 1.01)

for epoch in range(50):
    batches = minibatch(train_data, size=next(batch_sizes))
    for batch in batches:
        ...
```

### Types of change
enhancement

## Checklist


honnibal:
The change seems reasonable but I'm not sure your code actually does what you want. Are you sure you don't want ` batches = minibatch(train_data, size=batch_sizes)`?

theudas:
I want to increase the batchsize with each epoch, what you are doing is increasing the batch size with each step during each epoch and starting by the start value again in the next epoch. To be honest I don't know what performs better.

honnibal:
@theudas I think it makes most sense to increase the batch size throughout training, so call `compounding` once and then continue drawing from the same generator.

The reason to start at a small batch size is the imitation learning. We follow the analysis predicted by the parser during training, and we don't want to spend too long exploring the bad parses produced when we haven't made many updates. So, we want to update quickly at the start, and then settle into a larger batch size once some learning has been done.

honnibal:
Okay, I think I'll close this PR, because it might cause indicate a deeper problem if the user is passing a float for the batch size.

PR #3218, opened 2019-01-31
Replacing regex library with re to increase tokenization speed

svlandeg:
## Description
Implemented the minimal changes required to replace the `regex` library back with `re`:
- Removed regex from imports and requirements
- Replaced class shortcuts like `\p{Bengali}` (not supported by re) with their actual unicode ranges
- Converted `_quotes`and `ALPHA` variables to single-char concatenated strings, allowing the use of character classes `[]` instead of alternations with `|`
- Rewrote part of the URL expression that was causing Issue #957 (pathological backtracking caused by [greedy operators](https://github.com/explosion/spaCy/blob/master/spacy/lang/tokenizer_exceptions.py#L41))

Some additional small changes:
- Removed a few unnecessary (parts of) regular expressions
- Removed some unnecessary backslashes in character ranges
- Started unifying regular expressions, using the same parameterized format across languages
- Added/enabled/expanded some unit tests, incl. those to flag pathological backtracking (`test_issue957` and `test_issue2835`)
- Fixed few typos in `CONTRIBUTING.md`

### Results
Tested on the UD train corpora:
- Tokenization speed in WPS is 2-3 times faster
- Tokenization accuracy of the blank models remains virtually the same

image

### TODO
- Evaluation of re-trained language models with re instead of regex
- Run for `ja` & `th` (with CLI, as they're not installed properly on my system)

### Open issues (to discuss)
- Temporarily disabled some unit tests with double punctuation, such as "Hello''" and "Тест''", as these now oversegment
- Oversegmentation of double punctuation was found to introduce a few more errors for UD training corpora for `de`, `id` and `ur` specifically. This could be fixed by introducing a new regular expression, unless we want to keep the initial oversegmentation
- The back-tracking issue may creep in again in the future if new expressions with overlapping greedy operators would be introduced

### Future work (next PR)
- Investigate whether the unicode ranges can be further reduced (they are quite ugly now)
- Investigate the necessity & speed impact of e.g. non-latin characters in latin texts
- Use more character classes instead of alternations for the other variables in `char_classes`
- Clean up code further to enhance maintainability of the punctuation rules

### Types of change
Enhancement: tokenization speed

## Checklist


svlandeg:
Oh, looks like French unicode characters É and Î are tripping up the unit tests for Python 2.7 ...

PR #3227, opened 2019-02-04
💫 Add gold.spans_from_biluo_tags helper

ines:
## Description

I was writing examples for a new docs section and noticed that the path from BILUO tags to `doc.ents` was unnecessarily frustrating if you want to set the `doc.ents` as entity spans instead of offsets (unless I'm missing something?).

`gold.spans_from_biluo_tags` takes a `Doc` and a list of tags and returns a list of labeled `Span` objects.`gold.offsets_from_biluo_tags` now delegates to this helper.

On a related note: It should technically be possible to also set `doc.ents` from the offsets returned from `offsets_from_biluo_tags`. However, setting the `doc.ents` from offsets currently requires the labels to be integers, which isn't consistent with the new API. So this should be fixed as well.

(In general, I do prefer instantiating and using `Span` objects to overwrite the `doc.ents`, as opposed to offsets, even if it means creating a new object. The `doc.ents` are a sequence of spans, so it's much more intuitive to also set via a sequence of spans.)

### Types of change

enhancement

## Checklist

### DIFF
--- a/spacy/gold.pyx
+++ b/spacy/gold.pyx
@@ -12,7 +12,7 @@ import srsly
 
 from . import _align
 from .syntax import nonproj
-from .tokens import Doc
+from .tokens import Doc, Span
 from .errors import Errors
 from . import util
 from .util import minibatch, itershuffle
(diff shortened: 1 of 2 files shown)


honnibal:
Looks good!

honnibal:
> On a related note: It should technically be possible to also set `doc.ents` from the offsets returned from `offsets_from_biluo_tags`. However, setting the `doc.ents` from offsets currently requires the labels to be integers, which isn't consistent with the new API. So this should be fixed as well.

Fixing that is reasonable, yeah.

> (In general, I do prefer instantiating and using `Span` objects to overwrite the `doc.ents`, as opposed to offsets, even if it means creating a new object. The `doc.ents` are a sequence of spans, so it's much more intuitive to also set via a sequence of spans.)

What I hate is this tuples thing we have sometimes, `(start, end, label)` triples. Or is it `(label, start, end)`? It was a terrible idea; the `Span` is an object that's almost as lightweight as a tuple, honestly --- we should just use `Span` objects.

PR #3236, opened 2019-02-06
💫 Break up large matcher.pyx

ines:
## Description

As @honnibal  noted in #3173, `matcher.pyx` was getting quite large, so this PR breaks it up into its own subpackage. While doing that, I also tidied up a little.

@honnibal: The following helper function is currently unused. Is it supposed to be used somewhere? I removed it for now but we can put it back if needed.

```python
def _get_longest_matches(matches):
    '''Filter out matches that have a longer equivalent.'''
    longest_matches = {}
    for pattern_id, start, end in matches:
        key = (pattern_id, start)
        length = end-start
        if key not in longest_matches or length > longest_matches[key]:
            longest_matches[key] = length
    return [(pattern_id, start, start+length)
              for (pattern_id, start), length in longest_matches.items()]
```

### Types of change
enhancement

## Checklist


honnibal:
@ines I think you're right that that helper is unnecessary. I wrote it while there was a different problem, I think --- but then I fixed the underlying issue.

PR #3243, opened 2019-02-07
Update Example input JSON file to adhere to specification.

lauraBaakman:
## Description

I have changed the type of the `id` from a string to an int. According to the [json input spec ](https://spacy.io/api/annotation#json-input) the `id ` needs to be an `int` not a string. Using a string as `id` results in a `TypeError` when calling `spacy.gold.read_json_file()`.

### Types of change

change to the documentation.

## Checklist


ines:
Thanks! 👍

PR #3246, opened 2019-02-07
💫 Break up large pipeline.pyx

ines:
## Description

This PR breaks `pipeline.pyx` down into the following:

* `pipeline/entityruler.py` - entity ruler component
* `pipeline/functions.py` - simple components/helpers like `merge_entities`
* `pipeline/hooks.py` - maybe not the best name, but includes similarity hook, sentence segmenter
* `pipeline/pipes.pyx` – the built-in pipeline components like tagger, parser, NER etc. and the pipe base class

I tried breaking the built-in components down further, but it ended up being kinda frustrating because a lot of stuff depends on each other. So I ended up keeping them in one file.

## Checklist


spacy/pipeline/pipes.pyx
```diff
@@ -3,278 +3,47 @@
 # coding: utf8
 from __future__ import unicode_literals
 
-import numpy
```
DeNeutoy: If you didn't rename this file, the change would be backward compatible 😇

honnibal: How so? It's not like we documented `import numpy from spacy.pipeline` was supposed to work :p

honnibal:
Looks good to me! @ines go ahead and merge this if it's ready.

spacy/pipeline/pipes.pyx
ines: I think he meant the whole file in general? But in terms of user-facing API, it should be backwards compatible. Unless I forgot something here?

https://github.com/explosion/spaCy/blob/196f5e8a2bd68c8cfc2b882e2491a62621e9f5ca/spacy/pipeline/__init__.py#L4-L8

DeNeutoy: Oh my bad. It would just break if you imported components directly from the file, which maybe you care less about. I'll butt out!

ines: No worries – definitely appreciate the attention to detail, though!

PR #3253, opened 2019-02-08
Add split one token into several (resolves #2838)

### LINKED ISSUE #2838: Split one token into several
## Feature description
We now have a feature to merge tokens into one, implemented in the retokenizer. We're lacking the reverse feature: split one token into several. An API exist in _retokenize, but is not implemented yet.

grivaz:
Adds the capacity to split tokens

## Description
Implemented the feature as described in #2838. I added several error messages, and a new test suite for the functionality.

The tests passed with 554 warnings, but I think this is not related to this change.

Does this change require new documentation?

### Types of change
New feature and new tests. 

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -253,8 +253,8 @@ class Errors(object):
     E098 = ("Trying to set conflicting doc.ents: '{span1}' and '{span2}'. A token"
             " can only be part of one entity, so make sure the entities you're "
             "setting don't overlap.")
-    E099 = ("The splitted token can only have one root (head = 0).")
-    E100 = ("The splitted token needs to have a root (head = 0)")
+    E099 = ("The newly split token can only have one root (head = 0).")
+    E100 = ("The newly split token needs to have a root (head = 0)")
     E101 = ("All subtokens must have associated heads")
 
 @add_codes
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -253,6 +253,9 @@ class Errors(object):
     E098 = ("Trying to set conflicting doc.ents: '{span1}' and '{span2}'. A token"
             " can only be part of one entity, so make sure the entities you're "
             "setting don't overlap.")
+    E099 = ("The splitted token can only have one root (head = 0).")
+    E100 = ("The splitted token needs to have a root (head = 0)")
+    E101 = ("All subtokens must have associated heads")
 
 @add_codes
 class TempErrors(object):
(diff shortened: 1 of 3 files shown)


spacy/errors.py
```diff
@@ -253,6 +253,9 @@ class Errors(object):
     E098 = ("Trying to set conflicting doc.ents: '{span1}' and '{span2}'. A token"
             " can only be part of one entity, so make sure the entities you're "
             "setting don't overlap.")
+    E099 = ("The splitted token can only have one root (head = 0).")
+    E100 = ("The splitted token needs to have a root (head = 0)")
```
honnibal: Hmm. `splitted` isn't ideal English, but it's less ambiguous than `The split token...`, so I think I prefer it.

spacy/tokens/_retokenize.pyx
```diff
@@ -279,3 +279,106 @@ def _bulk_merge(Doc doc, merges):
 
     # Return the merged Python object
     return doc[spans[0].start]
+
+
+def _split(Doc doc, int token_index, orths, heads, deps, attrs):
+    """Retokenize the document, such that the token at
+    `doc.text[token_index]` is split into tokens with the orth 'orths'
```
honnibal:
```suggestion
    `doc[token_index]` is split into tokens with the orth 'orths'
```

spacy/tokens/_retokenize.pyx
```diff
@@ -43,12 +43,12 @@ cdef class Retokenizer:
         attrs = intify_attrs(attrs, strings_map=self.doc.vocab.strings)
         self.merges.append((span, attrs))
 
-    def split(self, Token token, orths, attrs=SimpleFrozenDict()):
+    def split(self, token_index, orths, heads, deps=[], attrs=SimpleFrozenDict()):
```
honnibal: Should we take a `Token` here, instead of the `token_index`? I think taking a `Token` is more consistent with `merge`, which takes a `Span`. I think some code paths can be simpler by taking a `Token` as well. If we take an index, do we end up doing the wrong thing if we make sequential changes, and the token indices end up incorrect?

honnibal:
Thanks a lot for this! It'll be a really useful change. The main question remaining is whether we want the API to take an index, or a `Token` object. I'm leaning towards `Token` object. Keen to get your thoughts.

spacy/errors.py
ines: This is referring to the resulting split tokens, right? So how about something like "The new tokens created by splitting a token..."?

honnibal: Maybe "the newly split tokens"?

honnibal:
I think there was a small problem before -- the `token.idx` attribute, which holds the character offset for the token, wasn't being updated. I made a change and tested for it, hopefully it goes green.

I also went ahead and made the change to the `retokenizer.split()` function, so that it takes a `Token` object instead of the index.

grivaz:
Sorry, I didn't see your input. Thank you for fixing it!

PR #3273, opened 2019-02-13
💫  Make handling of [Pipe].labels consistent 

### LINKED ISSUE #2179: ner/cfg has wrong extra_labels
Hello,
this is my ner/cfg:

```
{
  "token_vector_width":128,
  "nr_class":5,
  "pretrained_dims":0,
  "beam_density":0.0,
  "cnn_maxout_pieces":3,
  "beam_width":1,
  "hidden_width":200,
  "maxout_pieces":2,
  "hist_width":0,
  "extra_labels":[
    "CITIZENSHIP",
    "CITIZENSHIP",
    "CITIZENSHIP",
    "CITIZENSHIP"
  ],
  "hist_size":0,
  "hidden_depth":1
}
```
i have created this model with this code:
```
damiano@damiano:~$ python3
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import spacy
>>> nlp = spacy.blank('it')
>>> ner = nlp.create_pipe('ner')
>>> ner.add_label('CITIZENSHIP')
>>> nlp.add_pipe(ner)
>>> nlp.begin_training()
<thinc.neural.optimizers.Optimizer object at 0x7fcbed774cc0>
>>> nlp.to_disk('/home/damiano/ner')

```
as you can see i did not do anything special. Why my custom entity is repeated three times?

## Your Environment
    Info about spaCy

    Python version     3.5.2          
    Models                            
    Platform           Linux-4.13.0-37-generic-x86_64-with-Ubuntu-16.04-xenial
    spaCy version      2.0.10         
    Location           /home/damiano/.local/lib/python3.5/site-packages/spacy

### LINKED ISSUE #3130: Labels being created multiple times
Created a custom NER model the cfg file is showing my entities multiple times 

"extra_labels":[
    "PAY",
    "PAY",
    "PAY",
    "PAY",
    "CPTI",
    "CPTI",
    "CPTI",
    "CPTI",
    "CHG_Amts",
    "CHG_Amts",
    "CHG_Amts",
    "CHG_Amts"
  ],

## Your Environment

* Operating System: Windows 7
* Python Version Used:  3.6
* spaCy Version Used: 2.0.18
* Environment Information:

ines:
## Description

* Make sure all (trainable) built-in components expose a `labels` property.
* Make sure `labels` always returns the same data type (list).
* Make sure `labels` doesn't contain duplicates.
* Add tests and check that labels can be added via `add_label` and that behaviour is consistent.

Also actually resolves #2179 and resolves #3130. The problem here was that we were extracting the entity labels from the moves, e.g. `B-PERSON`, `L-PERSON` and so on. But we were not actually filtering out duplicates, so `EntityRecognizer.labels` would return every label 4 times and pass this down accordingly into the config etc.

## Checklist


honnibal:
Nice improvement!

Does it break backwards compatibility if we make it a tuple not a list though? `parser.labels.append` will otherwise "succeed" without doing anything useful.

honnibal:
Just looked at this again and I'm a little worried. I think the problems with the current API are a little deeper than this.

Basically, the parser and NER have this property `labels` that's returning stuff like `PERSON`, `nmod`, etc. That really only makes sense for the parser. For the tagger, the "labels" is just the tag set.

What does make sense to return for the pipeline components is the names of the classes the model is predicting. This is something many (but not all) pipeline components can have. In the parser this is currently at `parser.move_names`, which is crap.

I think we have two options:

1) Change the behaviour of `parser.labels` so it returns a list like `["U-PERSON", "B-PERSON", ...]` etc. This will be a breaking change, and some people's tests will probably pass while they get runtime errors, which is a pretty bad break. Like, I can easily imagine people just testing for the type not the content.

2) Keep the inconsistency of `component.labels`, and add a new property called something like `component.classes` or `component.class_names`. This component will hold the new consistent strings. We'd deprecate `parser.move_names` and `tagger.labels`. We might want to also deprecate `parser.labels`, and if you need that, you can write `parser.moves.labels`, to avoid confusing the top-level API.

More broadly, it's also fairly untrue that all components will have something like `component.labels` to return. The matcher doesn't have anything like that, neither do the merge functions. We could also have something like a sentiment model which predicts a score from -1 to 1. So, I don't know whether we want to promise this in the API. The main thing is to have the information in a consistent place for the components that do predict something like this.

ines:
> Does it break backwards compatibility if we make it a tuple not a list though? `parser.labels.append` will otherwise "succeed" without doing anything useful.

But then I guess they should all be tuples for consistency?

> More broadly, it's also fairly untrue that all components will have something like `component.labels` to return. The matcher doesn't have anything like that, neither do the merge functions.

The matcher isn't a pipeline component and the entity ruler does have a `.labels` property. I'm not suggesting that all pipeline components should have this – but the trainable components that expose a user-facing `add_label` method should also have a property exposing those labels, ideally in a consistent way and actually in the same format they were added.

> Basically, the parser and NER have this property `labels` that's returning stuff like `PERSON`, `nmod`, etc. That really only makes sense for the parser. For the tagger, the "labels" is just the tag set.

Yeah, but that's fine? You can do `Tagger.add_label`, so there should be a way to retrieve those labels.

> What does make sense to return for the pipeline components is the names of the classes the model is predicting. This is something many (but not all) pipeline components can have. In the parser this is currently at `parser.move_names`, which is crap.

I don't think this is so bad? Basically, what I'm suggesting is: keep the `move_names` as the attribute exposing the labels or "moves" the component is *actually* predicting (which isn't necessarily exposed to the user in an obvious way. For example, we're predicting `B-PERSON`, but the user mostly thinks about the label `PERSON`). Then have the `labels` expose just the readable labels that are used to annotate the objects (part-of-speech tags, dependency labels, entity types, text categories).

spacy/pipeline/pipes.pyx
```diff
@@ -358,7 +358,7 @@ class Tagger(Pipe):
 
     @property
     def labels(self):
-        return self.vocab.morphology.tag_names
+        return list(self.vocab.morphology.tag_names)
```
honnibal:
```suggestion
        return tuple(self.vocab.morphology.tag_names)
```

spacy/pipeline/pipes.pyx
```diff
@@ -1012,6 +1012,11 @@ cdef class DependencyParser(Parser):
         return (DependencyParser, (self.vocab, self.moves, self.model),
                 None, None)
 
+    @property
+    def labels(self):
+        # Get the labels from the model by looking at the available moves
+        return list(set(move.split("-")[1] for move in self.move_names))
```
honnibal:
```suggestion
        return tuple(set(move.split("-")[1] for move in self.move_names))
```

spacy/pipeline/pipes.pyx
```diff
@@ -1040,8 +1045,8 @@ cdef class EntityRecognizer(Parser):
     def labels(self):
         # Get the labels from the model by looking at the available moves, e.g.
         # B-PERSON, I-PERSON, L-PERSON, U-PERSON
-        return [move.split("-")[1] for move in self.move_names
-                if move[0] in ("B", "I", "L", "U")]
+        return list(set(move.split("-")[1] for move in self.move_names
```
honnibal:
```suggestion
        return tuple(set(move.split("-")[1] for move in self.move_names
```

spacy/tests/pipeline/test_pipe_methods.py
```diff
@@ -112,3 +112,15 @@ def test_add_lots_of_pipes(nlp, n_pipes):
...
+@pytest.mark.parametrize("component", ["ner", "tagger", "parser", "textcat"])
+def test_pipe_base_class_add_label(nlp, component):
+    label = "TEST"
+    pipe = nlp.create_pipe(component)
+    pipe.add_label(label)
+    if component == "tagger":
+        # Tagger always has the default coarse-grained label scheme
+        assert label in pipe.labels
+    else:
+        assert pipe.labels == [label]
```
honnibal:
```suggestion
        assert pipe.labels == (label,)
```

honnibal:
Okay. I've added a review to do the tuples change. I still think there's potential for confusion here, but you're right that if there's an `add_label` it makes sense to have a `labels`.

Issue #2926, opened 2018-11-14
spaCy tokenizer does not split correctly tokens separated by a slash (/) ending in a digit


arimbr:
spaCy tokenizer seems not to tokenize correctly tokens separated by slash (/) when some of them end with a digit.

## How to reproduce the behaviour

```
In [57]: import spacy
In [58]: nlp = spacy.load('fr')

In [59]: [t for t in nlp('Learn html5/css3/javascript/jquery')]
Out[59]: [Learn, html5/css3/javascript, /, jquery] # UNEXPECTED

In [60]: [t for t in nlp('Learn html/css/javascript/jquery')]
Out[60]: [Learn, html, /, css, /, javascript, /, jquery] # EXPECTED

```

## Your Environment
* **spaCy version:** 2.0.11
* **Platform:** Linux-4.15.0-36-generic-x86_64-with-Ubuntu-18.04-bionic
* **Python version:** 3.6.5
* **Models:** fr, en

Related issue https://github.com/explosion/spaCy/issues/891

ines:
Merging this with the master issue in #1642!

Issue #2822, opened 2018-10-04
Italian "po'" word is tokenized incorrectly


dvaosta:
## Your Environment
* **spaCy version:** 2.0.12
* **Platform:** Windows-10-10.0.16299-SP0
* **Python version:** 3.6.5
* **Models:** it

"Po'" (with the apostrophe) is the truncated form of "poco" (that means "a little") and must be considered as a single word, while currently spacy considers "po" and "'" as two tokens.
Notice that "Po", without the apostrophe, is the name of the longhest river in Italy. But in this case it always has the capital P and is never followed by the apostrophe.

dvaosta:
Thanks.

ines:
Merging this with the master issue in #1642!

Issue #3242, opened 2019-02-07
Tokenization speed too slow for French


svlandeg:
## Feature description
The evaluation on the UD corpora has shown that French tokenization is significantly slower (in words per second) than other languages. I'm opening this issue to experiment with a few ideas and discuss potential solutions.

#### Historical trail
Benchmarking the tokenizer has some flunctuations, but across runs this seems to be the general performance for the French blank model:
- early Jan 2019: 5-6K
- 30 Jan 2019 (from regex to re): 8-9K
- early Feb 2019 (after PR https://github.com/explosion/spaCy/pull/3218): 27-32K

![image](https://user-images.githubusercontent.com/8796347/52414271-25681900-2ae4-11e9-8f9f-8ea313e26af4.png)

svlandeg:
I tried simplifying the `ALPHA` character class, which brings a lot of complexities because of Latin letters that are included but irrelevant for French, e.g. https://en.wikipedia.org/wiki/Latin_Extended-D.
I assumed that the large list of tokenization exceptions using the complicated `ALPHA` class would run faster when limiting to a much smaller set of Latin characters relevant to French only.

Unfortunately, this didn't seem to have much effect on the tokenization speed.

svlandeg:
When I remove the [16K exceptions list](https://github.com/explosion/spaCy/blob/develop/spacy/lang/fr/_tokenizer_exceptions_list.py) entirely though, ofcourse some unit tests fail, but the UD evaluation accuracy remains almost the same on `fr_gsd-ud-train.txt`:
- Precision drops from 99% to 98%
- F-score stays at 99%

However, WPS rises to 73-74K!

svlandeg:
If we do remove the exceptions, one potential direction is to also remove the [hyphens as infix](https://github.com/explosion/spaCy/blob/develop/spacy/lang/punctuation.py#L43) for French. This makes a lot of the current French exceptions redundant (most of the list + additional regular expressions), but will ofcourse result in some undersegmentation instead of oversegmentation.
- orig state: 0.1% under, 0.4% over, 99% precision, 99% F
- without exception list: 0.09% under, 0.7% over, 98% precision, 99% F
- **without hyphen as infix**: 0.2% under, 0.2% over, 99% precision, 99% F

This will bring WPS at around 90K.

[...]

Issue #3002, opened 2018-12-02
Long sequence of digits and dots cause exponential time increase


WolfFinster:
## How to reproduce the behaviour

I discovered the following issue with a sentence from a wikipedia dump. Long sequences of digits and dots make the computation time explode.

```
In [3]: nlp = spacy.load('de_core_news_sm') 
   ...: nlp.remove_pipe('parser') 
   ...: nlp.remove_pipe('ner') 
In [13]: for x in nlp('Bei den Aufgaben der Rechenkünstler geht es natürlich um viel höhere Potenzen mehrstelliger Zahlen – zum Beispiel die Berechnung der 25. Wurzel aus 880.794.982.218.444.893.023.439.794.626.12
    ...: 0.190.780.624.990.275.329.063.400.179.824.681.489.784.873.773.249 (Lösung: 1729) und extremere Aufgaben.'): 
    ...:     print(x) 
```

The more digits-dot segments you add the worse it gets.

The following did not terminate for me. I waited several minutes:

```
In [31]: for x in nlp('880.794.982.218.444.893.023.439.794.626.120.190.780.624.990.275.671 ist eine lange Zahl'): 
    ...:     print(x) 
```

## Your Environment

* Operating System:

> Linux

* Python Version Used:

> 3.7

* spaCy Version Used:

> 2.0.16

* Environment Information:

>     spaCy version      2.0.16         
>     Location           /home/me/miniconda3/lib/python3.7/site-packages/spacy
>     Platform           Linux-4.13.9-1-default-x86_64-with-glibc2.9
>     Python version     3.7.0          
>     Models             de_core_news_sm

honnibal:
Thanks. This is related to the use of the `regex` library, and the fact that non-fixed-width lookarounds have crept into our expressions. The tokenizer regex needs to be refactored significantly.

ines:
I'll merge this with the regex issue in #1642 👍

PR #3293, opened 2019-02-19
Clean up of char classes, few tokenizer fixes and faster default French tokenizer

svlandeg:
## Description
This PR builds further upon PR https://github.com/explosion/spaCy/pull/3218 and finalizes Issue #1642. Further, it address Issue #3242.

- Split out the huge unicode ranges to allow using `LATIN_BASIC` or `LATIN_LOWER` instead of the large `ALPHA` classes. Note that for some classes, `upper+lower != all` because some unicode characters are a combination of both upper and lower letters and thus only included in the larger group (eg: `_latin_extendedB`)
- For French, removed the large `_tokenizer_exceptions_list.py` from the default model. This also means that `test_issue852` needs to be skipped.

### Small issues

- Added unit test for Issue #2656 (works)
- Added unit test for Issue #3002 (works)
- Removed xfail for Issue #2179 previously fixed by @honnibal 
- Fixed Issue #2822 by adding an Italian exception for `po'` + unit test
- Fixed Issue #2926 by allowing a number right before the infix `/` + unit test
- Added failing unit test for Issue #1235 (can not be fixed without variable-width look-behind)
- Added failing unit test for Issue #2070 (created partial fix, but can not be fully fixed without iterative looping between prefix/suffix and infix)
- Adjusted the documentation on `char_classes.py` in `adding-languages.md` to remove the reference to the  `regex` lib

### Types of change
Small enhancements

## Checklist


ines:
Thanks, this looks super good! 

I can't really make sense of the test failures 🤔 *Something* seems to segfault *somewhere* (where it happens in the tests usually isn't indicative of what actually causes it) and I'm not even 100% sure it's related to your changes?

svlandeg:
Yea, I'm pretty confused too. I'm not sure what could cause this, it's a sort of compiler error?

The only issue I had on my local system when compiling this branch, was that I had some outdated `.cpp` files (`matcher.cp` & `pipeline.cp`) that were moved to a subfolder in one of your recent commits. When trying to recompile from source, I got errors because it was still looking for the corresponding `.pyx` files in the old directories. Deleting those old `.cpp` files fixed that.

But I don't see how this could be a similar issue, as your commits did succeed on the tests before...

svlandeg:
Triggering the tests again with a new commit gives slightly different errors, including

> *** Error in `python': free(): invalid next size (fast): 0x00000000064d5b20 ***

Which seems to point to a memory error. I can't reproduce on my own system though, everything just runs...

ines:
Ok, so I just checked and this seems to be a problem on `develop` that's existed before your PR. It failed in a similar way before that, I just didn't notice.

There's definitely *something* wrong in the matcher, because I ended up with a weird segfault in a test the other day:

https://github.com/explosion/spaCy/blob/3b667787a932efdf2179bb8eb8a1654517e9a3e6/spacy/tests/regression/test_issue1971.py#L53-L67

The really frustrating and confusing thing here is that there's nothing wrong with the assertion it fails on – but *somewhere* previously, there's some memory corruption, so at some point down the line, possibly totally unrelated, it's trying to access something that's not there.

> Deleting those old `.cpp` files fixed that.

Yeah, unfortunately, manually deleting all `.cpp` and `.so` files is sometimes the only good solution.

ines:
Ok, confirmed that the matcher was causing the segfaults. Finally merging 👍

PR #3318, opened 2019-02-22
Clarify expected type; fix name

chbrown:
### Types of change

Documentation

Not sure if you periodically sync docs & (C|P)ython source docstrings, but `biluo_tags_from_offsets`'s source should also indicate that it reads through `entities` twice.

## Checklist
  * N/A? Is that required for documentation?
- N/A ~~I ran the tests, and all new and existing tests passed.~~
- N/A ~~My changes don't require a change to the documentation, or if they do, I've added all required information.~~

### DIFF
--- a/website/api/goldparse.jade
+++ b/website/api/goldparse.jade
@@ -151,7 +151,7 @@ p
 
     +row
         +cell #[code entities]
-        +cell iterable
+        +cell list
         +cell
             |  A sequence of #[code (start, end, label)] triples. #[code start]
             |  and #[code end] should be character-offset integers denoting the
@@ -185,7 +185,7 @@ p
         +cell The document that the BILUO tags refer to.
 
     +row
-        +cell #[code entities]
+        +cell #[code tags]
         +cell iterable
         +cell
             |  A sequence of #[+a("/api/annotation#biluo") BILUO] tags with


website/api/goldparse.jade
```diff
@@ -151,7 +151,7 @@ p
 
     +row
         +cell #[code entities]
-        +cell iterable
+        +cell list
```
honnibal: This change isn't right. The function accepts any iterable of entities, e.g. a tuple, generator, custom type. It's more correct to say "iterable" here than "list".

website/api/goldparse.jade
```diff
@@ -185,7 +185,7 @@ p
         +cell The document that the BILUO tags refer to.
 
     +row
-        +cell #[code entities]
+        +cell #[code tags]
```
honnibal: This is more debateable. People often think of tags as part-of-speech tags though, and the description does say what the function expects.

honnibal:
Really appreciate the thought, but there are some subtleties to how the documentation should be worded I think you might have missed. We definitely do appreciate small corrections to the docs --- it's just that I think these ones aren't quite right.

website/api/goldparse.jade
chbrown: You're right that "list" is too specific, but "iterable" is incorrect, and I actually ran into a bit of trouble the first time I used this function, prior to reading the source, because I assumed it could take any kind of iterable. It _accepts_ a generator, true, but it reads through the entire collection twice, so by the time it gets to the second loop, the generator will be empty. See https://github.com/explosion/spaCy/blob/386cec1/spacy/gold.pyx

website/api/goldparse.jade
chbrown: `tags` is the name of the argument that that function actually uses (see https://github.com/explosion/spaCy/blob/386cec1/spacy/gold.pyx#L550), and I think it's useful to distinguish between these "entities" (B/I/L/U-prefixed or O strings) and the 3-tuple "entities" expected by `biluo_tags_from_offsets`.

chbrown:
Thanks for taking a look. I agree you have more comprehensive view of the terminology used throughout the documentation, and that my changes might not be optimal. But at least with the iterable -> list change, assuming "iterable" was accurate actually caused me maybe half an hour of debugging. I've added more context and links to the source for both of the changes above.

PR #3340, opened 2019-02-27
Make doc[0].is_sent_start == True (closes #2869)

### LINKED ISSUE #2869: is_sent_start on first document token should be True instead of None
Hey everyone,

From the documentation of **Token.is_sent_start**:
"... It also now returns None if the answer is unknown, and _fixes a quirk in the old logic that would always set the property to 0 for the first word of the document_."

The value of is_sent_start for the first token in a document is now always **None** instead of 0. Is this the intended behavior? I would argue that **True** would be more reasonable and consistent. At least it caused some issues to my code when I concatenated several documents and the is_sent_start flag on each first document word was missing.

```
import spacy
nlp = spacy.load('en')
text = "This is the first sentence. This is the second sentence."
doc = nlp(text)
for token in doc:
    print(token.i, token.is_sent_start, token.text)
```

## Info about spaCy
* **spaCy version:** 2.0.16
* **Platform:** Linux-4.13.0-38-generic-x86_64-with-debian-stretch-sid
* **Python version:** 3.7.0
* **Models:** en, de

Kind regards,
Andreas

honnibal:
We had `doc[0].is_sent_start == None` for reasons, but on the whole, that's dumb. We know the first token should start a sentence, so just set it. We ensure that `doc.is_sentenced` remains `False` if `doc[0].is_sent_start` is the only value set.

### DIFF
--- a/spacy/tests/doc/test_token_api.py
+++ b/spacy/tests/doc/test_token_api.py
@@ -192,3 +192,9 @@ def test_tokens_sent(doc):
     assert doc[7].sent.text == "This is another sentence ."
     assert doc[1].sent.root.left_edge.text == "This"
     assert doc[7].sent.root.left_edge.text == "This"
+
+def test_token0_has_sent_start_true():
+    doc = Doc(Vocab(), words=["hello", "world"])
+    assert doc[0].is_sent_start is True
+    assert doc[1].is_sent_start is None
+    assert not doc.is_sentenced
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -188,13 +188,18 @@ cdef class Doc:
 
     @property
     def is_sentenced(self):
-        # Check if the document has sentence boundaries,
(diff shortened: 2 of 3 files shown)


mmunozm:
Hi,

I am afraid this change here created undesired behavior if we decide to use a custom "sentencizer" instead of a parser and the doc has just one sentence. E.g.:
```Python
import en_core_web_sm

nlp = en_core_web_sm.load(disable=["parser", "ner"])

# As an example of a simple custom sentencizer
def simple_sentencizer(doc):
    doc[0].is_sent_start = True
    return doc
nlp.add_pipe(simple_sentencizer, name="sentencizer", before="tagger")

doc = nlp("Just one sentence.")
print(doc.is_sentenced) # False

list(doc.sents)
# ValueError: [E030] Sentence boundaries unset. ...
```

The correct behavior used to work on previous versions of spacy. Is there anything else I should be doing  inside the sentencizer with this new version?

PR #3364, opened 2019-03-07
Relax regex requires version for Py3

javabrett:
## Description
In builds I've noticed `spaCy` is forcibly downgrading `regex`, which is a direct dependency in some projects, and a transient dependency in others.  This is caused by the locked version `regex==2018.01.10`.  Ideally this would be lower-bound only, and `spaCy` should probably have a `pip freeze` `requirements.txt` file for reproducible build/test.

### Types of change
This PR removes the upper-bound on `regex`.  The second commit makes sure this occurs for Py3 only, as there are compatibility issues with Py2 which showed-up in tests/Travis.

## Checklist


honnibal:
Unfortunately new versions of regex do indeed break spaCy, and regex doesn't semver, making life difficult.

For these and other reasons, on the develop branch we've actually removed the regex dependency entirely, so we won't be making another release that depends on regex.

javabrett:
If there won't be a maintenance release ahead of `develop`, then I agree this is not needed.  Closing.

Issue #2740, opened 2018-09-06
Ability to pass additional arguments to pipeline components


thomasopsomer:
## Feature description

I'm intensively using custom components :) And Sometimes I would like to pass them additional, non text/doc arguments for instance the date of a document, tags, options...

One way to allow that could be to do like scikit learn does with their `Pipeline`: one can provide a dict of parameters (or a list of tuple), with keys being the arguments prefixed with the name of the of the component, something like `pipeline_params = {<component_name>__<arg_name>: "value"}` and each component of the pipeline extract its arguments from it.

honnibal:
Finally got to this :tada: . I think it'll be a nice change. We went with the name `component_cfg`.

PR #3386, opened 2019-03-10
💫 Allow passing of config parameters to specific pipeline components

honnibal:
In `nlp.begin_training()`, we previously had a single group of `**cfg` keyword arguments which went out to all components. This makes it hard to configure the pipeline if different components have keyword arguments of the same name -- which will obviously be the case if we have two instances of the same class.

The solution's pretty simple: have a new keyword argument, `component_cfg`, that can take a dict keyed by the component name.

We should consider adding this to the `Language.update()`, `Language.__call__()` and `Language.pipe()` methods as well. See Issue #2740 .

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -106,6 +106,7 @@ class Language(object):
 
     DOCS: https://spacy.io/api/language
     """
+
     Defaults = BaseDefaults
     lang = None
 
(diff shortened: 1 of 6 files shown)


themantalope:
Hi! I'm curious, has this been added to `Language.__call__`? I saw that issue #2740 has been referenced but that one is now closed. 

I would like to be able to develop a custom extension to `Language.__call__` for processing output from `tesseract` :)

PR #2594, opened 2018-07-25
Changed conllu2json to be able to extract NER tags

katarkor:
## Description
conllu2json now checks the 10th column of the first word: if it's a NER tag it will extract NER tags from the whole conllu file and add them to the json file. If not, it creates the json file as before, without NER tags.

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [ ] I ran the tests, and all new and existing tests passed.
- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #3405, opened 2019-03-13
Merging conversion scripts for conll formats

### LINKED ISSUE #3373: 💫 Merge conllu converters into one function
I was just updating the `spacy convert` command and docs and noticed that the converters are currently a bit messy. We might be able to eliminate some code duplication and make them a bit nicer overall. It'd be nice if we could merge the `conll` converters (or at least `conll`/`conllu` and `conllubio`) into a single script.

(That said, we do have to accept that they'll always be kinda hacky, simply because the formats we're dealing with aren't always 100% consistent. There are several variations of the `.conll` format alone that we need to handle).

### Resources

* `spacy.cli.converters`: https://github.com/explosion/spaCy/tree/develop/spacy/cli/converters
* #2525: original PR and motivation for the `conllubio` converter

### FYI: Future plans

1. Change converters to create and return `Doc` objects instead of JSON objects. This means that the new `Doc.to_json` object will be the single source of truth for the JSON format and we don't end up with arbitrary data transformation logic all over the place.
2. Update the training data format to a more straightforward JSONL format. See #2928.

Poluglottos:
## Description

 Hopefully this fixes #3373 , by combining the conll/conllu and conllubio format converters into one script. It does not address the conll_ner format, but this PR is a work in progress and I may add this if what I am doing is going in the right direction

### Types of change

Enhancement to cli converter

## Checklist


Poluglottos:
@ohenrik does this still support your intended functionality?

ohenrik:
It looks correct, but please try to download this file and convert it. If the result correctly includes NER tags it works :)

https://raw.githubusercontent.com/ohenrik/nb_news_ud_sm/master/original_data/no-ud-dev-ner.conllu

ines:
@Poluglottos Thanks so much for taking this on, looks good 👍

@ohenrik Ah, nice – having a test file like that is really useful. When I was testing the converters, I was never really sure if I was doing it correctly, since there are so many subtle variations of the formats.

I wonder if we could add converter tests that only test the converter functions themselves and run on maybe one or two lines (hard-coded into the test – I wouldn't necessarily want to include the `.conllu` files in our test suite).

ohenrik:
Glad i could help :)

At the moment i have a bit too much work on my hands to write these tests, but i could potentially run them and look over the code to see that they are correct (given that the test document i provided is in norwegian).

I can alternatively write these tests at a later point, for example in a month or two.

The updated version of the files i used when creating these converters and the NER model i made is here: https://github.com/ltgoslo/norne/

ines:
@ohenrik No worries! I just added one simple test using a few rows from the data. This will hopefully also make it easier for others to add tests for different configurations.

@Poluglottos If there's anything else you want to test, you can add a test in a similar format to `tests/test_cli.py` 🙂

Poluglottos:
Hmm... I have read into this a little bit. It seems that although #2525 was necessary to even include biluo ner tags, this same functionality eventually got added to the original conllu conversion script in #2594.

ines:
Cool, I'll go ahead and merge this then 👍 With the test template I've added, it should hopefully be easier to add more in the future.

PR #3408, opened 2019-03-14
label in span not writable anymore

svlandeg:
Revert part of PR https://github.com/explosion/spaCy/pull/3031 which made `Span.label_` writable. This can get confusing when e.g. running something like 

```
for ent in doc.ents:
    ent.label_ = "GPE"
```
Which wouldn't actually write the "GPE" annotation to the underlying `Token` / `doc` objects. With this PR, the above code isn't executable anymore and will throw a `NotImplementedError` similar to the one as implemented for property `ent_id`.

### Types of change
Functionality change

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -358,6 +358,10 @@ class Errors(object):
             "arguments to exclude fields from being serialized or deserialized "
             "is now deprecated. Please use the `exclude` argument instead. "
             "For example: exclude=['{arg}'].")
+    E129 = ("Can not write the label of an existing Span object. "
+            "Instead, create a new Span object and specify the `label` keyword argument, "
+            "for example:\nfrom spacy.tokens import Span\n"
+            "span = Span(doc, start=3, end=5, label='GPE')")
 
 
 @add_codes
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -358,11 +358,11 @@ class Errors(object):
             "arguments to exclude fields from being serialized or deserialized "
             "is now deprecated. Please use the `exclude` argument instead. "
             "For example: exclude=['{arg}'].")
-    E129 = ("Can not write the label of an existing Span object because a Span "
-            "is only a read-only view of the underlying Token objects stored in the Doc. "
+    E129 = ("Cannot write the label of an existing Span object because a Span "
+            "is a read-only view of the underlying Token objects stored in the Doc. "
             "Instead, create a new Span object and specify the `label` keyword argument, "
             "for example:\nfrom spacy.tokens import Span\n"
-            "span = Span(doc, start=3, end=5, label='GPE')")
+            "span = Span(doc, start={start}, end={end}, label='{label}')")
 
 
 @add_codes
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -358,7 +358,8 @@ class Errors(object):
             "arguments to exclude fields from being serialized or deserialized "
             "is now deprecated. Please use the `exclude` argument instead. "
             "For example: exclude=['{arg}'].")
-    E129 = ("Can not write the label of an existing Span object. "
+    E129 = ("Can not write the label of an existing Span object because a Span "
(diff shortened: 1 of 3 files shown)


spacy/tests/doc/test_span.py
```diff
@@ -178,6 +178,7 @@ def test_span_string_label(doc):
     assert span.label == doc.vocab.strings["hello"]
 
 
+@pytest.mark.xfail(reason="label is not writable")
```
ines: Instead of xfailing, we should probably explicitly test for the error here. For instance:

```python
span = Span(doc, 0, 1)
with pytest.raises(NotImplementedError):
    span.label_ = "hello"
```

spacy/tokens/span.pyx
```diff
@@ -653,7 +653,7 @@ cdef class Span:
             return self.doc.vocab.strings[self.label]
 
         def __set__(self, unicode label_):
-            self.label = self.doc.vocab.strings.add(label_)
+            raise NotImplementedError(TempErrors.T007.format(attr="label_"))
```
ines: Maybe we want a different error message here? `T007` says "Vote for this feature on the issue tracker", but... do we want this? The implications of writing to a span are pretty subtle and I think we've now come up with a few good reasons why we can't make `Span.label` writable like this. So the error message might be misleading.

Instead, maybe we can add a custom error that explains why writing to the span's label isn't possible? Like, a `Span` is only a view and its label corresponds to the tokens entity labels. If you want to set entities, you should create a new `Span` like this: etc etc.

cc: @honnibal

svlandeg: All good points - made the changes accordingly!

spacy/errors.py
```diff
@@ -358,6 +358,11 @@ class Errors(object):
             "arguments to exclude fields from being serialized or deserialized "
             "is now deprecated. Please use the `exclude` argument instead. "
             "For example: exclude=['{arg}'].")
+    E129 = ("Can not write the label of an existing Span object because a Span "
+            "is only a read-only view of the underlying Token objects stored in the Doc. "
```
ines:
```suggestion
            "is a read-only view of the underlying Token objects stored in the Doc. "
```

spacy/errors.py
```diff
@@ -358,6 +358,11 @@ class Errors(object):
             "arguments to exclude fields from being serialized or deserialized "
             "is now deprecated. Please use the `exclude` argument instead. "
             "For example: exclude=['{arg}'].")
+    E129 = ("Can not write the label of an existing Span object because a Span "
```
ines:
```suggestion
    E129 = ("Cannot write the label of an existing Span object because a Span "
```

spacy/errors.py
```diff
@@ -358,6 +358,11 @@ class Errors(object):
...
+            "is only a read-only view of the underlying Token objects stored in the Doc. "
+            "Instead, create a new Span object and specify the `label` keyword argument, "
+            "for example:\nfrom spacy.tokens import Span\n"
+            "span = Span(doc, start=3, end=5, label='GPE')")
```
ines: If we wanted to, we could be *super* helpful here and pass in the `start`, `end` and `label` from the span / label the user was trying to add. So they'd be able to just copy-paste the code 😃

(Or is this overkill? Any potential bad side-effects of this that I'm missing? cc: @honnibal)

svlandeg: Good idea, it can hardly be overkill if it's only a minute's work :-)

svlandeg: The only edge case maybe that isn't covered now is passing "None", e.g.
`span.label_ = None`
will print, at the end of the error msg:
`span = Span(doc, start=0, end=1, label='None')`

honnibal: This is a great error message! Tonnes better than my suggestion of just removing the setter...

honnibal:
Really nice! Thanks for the attention to detail.

PR #3409, opened 2019-03-15
Improved Dutch language resources and Dutch lemmatization

yvespeirsman:
## Description
This pull request provides improvements to the Dutch language resources and adds Dutch lemmatization: 
- We added list of tokenizer exceptions for better sentence segmentation,
- expanded the Dutch stopword list to achieve parity with English and French stopword lists,
- added `punctuation.py`, which modifies the list of infixes, so that hyphenated words are not split, and filters `'s` (and its variants) from the list suffixes so that plural forms such as `auto's` are treated as one token,
- defined a new DutchLemmatizer class (as is the case for French), which does not subclass from the default Lemmatizer but mimics its interface. A lookup table provides basic lemmatization without part-of-speech tagging, but we recommend using lemmatization together with part-of-speech tagging for improved accuracy. This is because the lookup table only contains unambiguous cases: if the lemmatization of a word depends on its part of speech, it is included in the exception indexes. As in other lemmatizers, the suffix rules in `_lemma_rules.py` are treated as a last resort.

See code comments for more details. 

Tests can be run with `python -m pytest spacy/tests/lang/nl/`.

### Types of change
Enhancements: improved sentence segmentation (with tokenizer exceptions), more language-specific tokenization, longer Dutch stopword list
New feature: Dutch lemmatization

## Checklist


ines:
Thanks so much for this! I'll try to rebase this onto `develop`, since that branch includes @svlandeg's improvements to the tokenizer, plus lots of other stuff. There probably won't be another significant release from `master`. (Just a heads-up in case I temporarily break something in the progress. Fingers crossed 🤞)

svlandeg:
I ran a quick evaluation on `nl_alpino-ud-train.txt` comparing the `develop` branch with this PR:
- **tokenization performance** goes up from 98% to virtually 100%, mainly by reducing over-tokenization on hyphenated strings `xxxx-xxxx` 
- **lemmatization performance** goes up from 73% to 84%
- **sentence segmentation** goes up from 73% to 76%

All these nice performance improvements do come at a small speed reduction:
- **loading time** went up from 0.18s to 1.85s (ran each only once, so to take with a grain of salt but still gives some indication)
- **tokenization time** went from ca. 120K tokens per second to 100K (as reference: it used to be around 60K with `regex`)

yvespeirsman:
Hi @svlandeg, 

Thanks for that! One thing that strikes me is that our own evaluation of the lemmatizer gave a considerably higher performance on the UD data. Are you using spaCy's Dutch part-of-speech tagging together with the lemmatization, or are you working with a blank Dutch model? Are you lowercasing the UD lemmas?

svlandeg:
The numbers I quoted are using the blank Dutch model, using this script: https://github.com/svlandeg/spaCy/blob/develop/spacy/cli/ud/run_eval.py

yvespeirsman:
OK, that explains a lot:

- We've optimized lemmatization for working together with part-of-speech tagging. All tokens whose lemmatization depends on their part of speech are in the exception indexes, so they will not be lemmatized when you use a blank model. 
- We've chosen to always lowercase lemmas (as is the case, for example, when you lemmatize with an English model). We therefore also lowercased the UD lemmas in our evaluation.
If you apply both changes, accuracy should be more around 95-97%. 

Finally, could you check if the responsibility for the slower tokenization lies with the tokenizer exceptions? If this is the case, and if it's an issue, we can work on making that list shorter.

spacy/lang/nl/lemmatizer/_adjectives.py
```diff
@@ -0,0 +1,4584 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+
+ADJECTIVES = set(
```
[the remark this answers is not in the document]
svlandeg: This list seems to contain quite a lot of verbs that I can't imagine being adjectives, like `klussen` or `verwilderen` ?

spacy/lang/nl/lemmatizer/_adverbs_irreg.py
```diff
@@ -0,0 +1,23 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+
+ADVERBS_IRREG = {
+    "'ns": ('eens',),
+    "'s": ('eens',),
+    'acht': ('halfacht',),
```
svlandeg: I'm puzzled by `acht` -> `halfacht`

spacy/lang/nl/punctuation.py
```diff
@@ -0,0 +1,39 @@
...
+_quotes = CONCAT_QUOTES.replace("'", "")
+
+_infixes = (
+    LIST_ELLIPSES
+    + LIST_ICONS
+    + [
+        r"(?<=[{al}])\.(?=[{au}])".format(al=ALPHA_LOWER, au=ALPHA_UPPER),
+        r"(?<=[{a}])[,!?](?=[{a}])".format(a=ALPHA),
+        r"(?<=[{a}])[:<>=](?=[{a}])".format(a=ALPHA),
+        r"(?<=[{a}]),(?=[{a}])".format(a=ALPHA),
```
svlandeg: It looks like this rule with `,` is redundant given the second rule with `,!?`

spacy/lang/nl/punctuation.py
```diff
@@ -0,0 +1,39 @@
...
+        r"(?<=[{a}])([{q}\)\]\(\[])(?=[{a}])".format(a=ALPHA, q=_quotes),
```
svlandeg: IIRC, round brackets `()` are contained within the `CONCAT_QUOTES` var so could be removed from the regex here.

spacy/lang/nl/punctuation.py
```diff
@@ -0,0 +1,39 @@
...
+        r"(?<=[{a}])--(?=[{a}])".format(a=ALPHA),
+        r"(?<=[0-9])-(?=[0-9])",
```
svlandeg: Would it make sense to use `char_classes.HYPHENS` here instead of only `-` ?

spacy/tests/lang/nl/test_text.py
```diff
@@ -2,10 +2,19 @@
 from __future__ import unicode_literals
 
 import pytest
-from spacy.lang.nl.lex_attrs import like_num
 
 
-@pytest.mark.parametrize("word", ["elf", "elfde"])
```
svlandeg: why was the test for `like_num` removed ?

yvespeirsman:
All good points. We'll pick them up next week!

svlandeg:
@yvespeirsman : makes sense. 95-97% is impressive! And I guess it's good to see also the blank model going up by itself :-)
>     * We've optimized lemmatization for working together with part-of-speech tagging. All tokens whose lemmatization depends on their part of speech are in the exception indexes, so they will not be lemmatized when you use a blank model.
> 
>     * We've chosen to always lowercase lemmas (as is the case, for example, when you lemmatize with an English model). We therefore also lowercased the UD lemmas in our evaluation.
>       If you apply both changes, accuracy should be more around 95-97%.

svlandeg:
@ines : I think you may have accidentally removed/overwritten an existing unit test: https://github.com/explosion/spaCy/blob/develop/spacy/tests/lang/nl/test_text.py

ines:
@svlandeg Ah, damn, I thought this was part of the PR. Fixing!

spacy/lang/nl/lemmatizer/_adjectives.py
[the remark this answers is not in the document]
lemontheme: Going to pitch in here so as to keep momentum high. (I worked together with Yves on this pull request.)

That's an interesting point about sorting the strings. Looks like that's what the other languages with lemmatization data do. Beyond conforming to that convention, however, there isn't much to be gained from it. It's not recommended to edit the lemmatization data manually (neither the lemma indexes "_<pos>.py" nor the exceptions indexes in "_<pos>_irreg.py" ), since the data is in fact extracted entirely automatically (even the serialization/persistence as a python module happens automatically). So any improvements to the data should focus on the data extraction pipeline (which currently resides in an ugly Jupyter notebook, but will be shared once it's cleaned up).

That brings us to the second observation. I'll run some tests to see how those cases slipped in. In any case, it looks like they're also included correctly in "_verbs.py", so the impact on lemmatization accuracy should be minimal.

spacy/lang/nl/lemmatizer/_adverbs_irreg.py
lemontheme: That is indeed puzzling. I would suggest manually removing it. I'll investigate how we can avoid the same mistake in future updates.

spacy/lang/nl/punctuation.py
lemontheme: Yikes, that does indeed seem to be case!

spacy/lang/nl/punctuation.py
lemontheme: Wasn't aware that there was a character class for that. Seems like that would make sense!

spacy/lang/nl/punctuation.py
ines: Yeah, when I did the rebase to `develop`, I also just copied over the current rules from `de.punctuation` without really questioning them. So if it turns out there are improvements to be made here, we can also update the `de` rules accordingly 🎉

ines:
Quick update: We just merged `develop` into `master`, so I've changed the base of this PR to `master` again.

spacy/lang/nl/punctuation.py
svlandeg: @ines @lemontheme : It's nothing urgent or a big deal - so I'd be happy to look into this and check any other redundancies in other languages after this PR is merged! I did some initial cleanup during the regex PR but apparently didn't catch it all at the time ;-)

lemontheme:
We're almost done incorporating @svlandeg's feedback. Here's where we stand right now.

- I've removed the infix pattern that was identified as redundant. 
- By contrast, I've left the infix patterns for round brackets and dashes untouched. This is mostly because I'm not 100% confident of how those patterns are ultimately compiled into regexes behind the scenes. There are other considerations too, however, which I'll explain in a comment tacked onto @svlandeg's original posts.

Pending thorough evaluation, a second push will include an updated version of the lemmatization data. The changes to expect there:

- Lemma indexes are sorted alphabetically.
- Cases like `verwilderen` (see @svlandeg's original post) have been removed from the adjectives exceptions index.

Expect another update soon!

spacy/lang/nl/punctuation.py
lemontheme: So, I've just taken a closer look at this. Turns out some caution is required when using the `HYPHENS` class, depending on whether it's hyphens (`-`) or dashes (`–` (em-dash), `—`  (en-dash), `--`, `---`, etc.)  that you want to split or, by exclusion, preserve. This is because `HYPHENS` conflates hyphens and dashes into a single set. 

In the case of the pattern on line 23, it makes sense to use the `HYPHENS` character class because our goal is to split on both hyphens and dashes.

In the case of the penultimate pattern on line 22, however, we should use a variant of `HYPHENS` that (confusingly enough) leaves out the actual hyphen character -- the effect being that hyphens are *not* split on. Otherwise, compound words like `zwart-witconclusie` are tokenized as `['zwart', '-', 'witconclusie']`.

I'm having some trouble getting the regular expressions to compile though. Is this potentially something you would like to take a closer look at, @svlandeg?

svlandeg: You might have to introduce a new variable in `char_classes.py`:
`CONCAT_HYPHENS = group_chars(_hyphens)`
and then do something like 
`r"(?<=[0-9])[{h}](?=[0-9])".format(h=CONCAT_HYPHENS)`

For line 22 you can make a custom variant of `CONCAT_HYPHENS ` (which I would personally define in the `nl` file because it's lang-specific)

I haven't checked/run this code yet though, as I'm in the middle of a different PR ;-)

yvespeirsman:
I have now also updated the lemmatization data:

- we've sorted the strings alphabically,
- got rid of the incorrect verbs in the adjectives file, and
- removed the most obvious lemmatization errors. 

The lemmatization files are generated automatically from potentially noisy sources, so I don't expect all problems will be fixed. However, as most of these are extremely infrequent, I expect their impact on accuracy will be minimal, if any.

ines:
@lemontheme @yvespeirsman Just double-checking – is this PR ready to merge or are you still working on something that should be included here?

yvespeirsman:
Hi @ines, as far as we're concerned, this is ready to merge!

ines:
@yvespeirsman Okay, perfect! Sorry for letting this sit around, but for some reason, I thought it was still a work in progress. But I'll go ahead and merge then 🎉

yvespeirsman:
No worries, I could have been more explicit about the status :)

PR #3427, opened 2019-03-19
Tiny correction in french lookup dictionary

mhham:
## Description
Correct the lemmatization of "nombres",  which changed from "nombre" to "nombrer" from spacy 2.0 to spacy 2.1

## Checklist


ines:
cc: @amperinet @mauryaland – could you help review this? 🙂

amperinet:
Yes, sure! 
It looks fine to me. Thank you @mhham for spotting this error ; the lemma for "nombres" is "nombre".

mauryaland:
Looks legit to me! It is one of numerous examples in the lookup table where the lemma form for a plural noun is an infinitive verb. This is due to the fact that the plural/singular form of a noun is the same as one form of the conjugated verb. However, we often want the singular noun as a lemma because the verb is underused. Another example was 'mondes' -> 'monder'.

If I remember well, @amperinet  already did a great work about cleaning a lot of them.

ines:
Perfect, thank you so much! 👍

PR #3434, opened 2019-03-19
Raise Error for a narrow unicode build of Python

### LINKED ISSUE #3426: 💫 Raise error if user is running narrow unicode build
## Feature description

Because v2.1 includes various changes to the regular expressions used in the tokenizer (including replacing `regex` with `re`), loading a model with a narrow unicode build of Python will fail with `sre_constants.error: bad character range`. See #3421 for details.

spaCy should probably check `sys.maxunicode` on import and raise an error if it's `65535`, meaning the user is running a narrow build. (Even aside form spaCy, a narrow unicode build is pretty inconvenient when working with text in Python.)

I've already added a section on this to the troubleshooting guide, so we can use text from that for the error message: https://spacy.io/usage#narrow-unicode

Bharat123rox:
Fix #3426 by raising a `ValueError` if a narrow unicode build is detected.
## Description

A generic `ValueError` has been provided (Is it correct?). The error message is provided with a slight modification from [here](https://spacy.io/usage#narrow-unicode). IMO, I felt a `UnicodeDecodeError` or `UnicodeEncodeError` doesn't fit here. In case the error message has to be improved, do let me know!

### Types of change

- [x] I have submitted the spaCy Contributor Agreement (Along with this PR).
- [x] My changes don't require a change to the documentation.

### DIFF
--- a/.github/CONTRIBUTOR_AGREEMENT.md
+++ b/.github/CONTRIBUTOR_AGREEMENT.md
@@ -87,7 +87,7 @@ U.S. Federal law. Any choice of law rules will not apply.
 7. Please place an “x” on one of the applicable statement below. Please do NOT
 mark both statements:
 
-    * [ ] I am signing on behalf of myself as an individual and no other person
+    * [x] I am signing on behalf of myself as an individual and no other person
     or entity, including my employer, has or will have rights with respect to my
     contributions.
 
@@ -98,9 +98,9 @@ mark both statements:
 
 | Field                          | Entry                |
 |------------------------------- | -------------------- |
-| Name                           |                      |
+| Name                           | Bharat Raghunathan   |
 | Company name (if applicable)   |                      |
 | Title or role (if applicable)  |                      |
-| Date                           |                      |
-| GitHub username                |                      |
+| Date                           | 19 - 03 - 2019       |
+| GitHub username                | Bharat123rox         |
 | Website (optional)             |                      |
--- a/spacy/__init__.py
+++ b/spacy/__init__.py
@@ -1,6 +1,7 @@
 # coding: utf8
 from __future__ import unicode_literals
 import warnings
+import sys
 
 warnings.filterwarnings("ignore", message="numpy.dtype size changed")
 warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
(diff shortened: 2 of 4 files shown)


spacy/__init__.py
```diff
@@ -14,6 +15,12 @@
 from .errors import Warnings, deprecation_warning
 from . import util
 
+if __version__ >= '2.1.0' and sys.maxunicode <= 65535:
```
ines:
```suggestion
if sys.maxunicode == 65535:
```

We don't actually need to check the version here, since this code will only be included if the user is running a relevant version anyways. And we might as well do `==`, since there are really only two values. If the result was any other number, it'd be _super_ confusing anyways.

.github/CONTRIBUTOR_AGREEMENT.md
```diff
@@ -98,9 +98,9 @@ mark both statements:
 
 | Field                          | Entry                |
 |------------------------------- | -------------------- |
-| Name                           |                      |
+| Name                           | Bharat Raghunathan   |
```
ines: Quick note: Looks like you accidentally filled in the *template* instead of copying it over to a new file like `.github/contributors/Bharat123rox.md`. If you could fix that, that'd be great!

spacy/__init__.py
```diff
@@ -14,6 +15,12 @@
...
+    raise ValueError('''You are running a narrow unicode build,
```
ines: How about `SystemError`? That's the closest I can think of, because it expresses there's something wrong with the system configuration?

Also, we can move the error message text to [`spacy/errors.py`](spacy/errors.py), to keep the texts tidy and in one place. We introduced this a while ago and it made the code so much easier to read, because there wouldn't be big blocks of messages everywhere.

Using the error message would then look something like this:

```python
from .errors import Errors

raise SystemError(Errors.E123)  # or whatever the number is
```

Bharat123rox:
Looks like intermittent failure caused by 503 error in `git fetch` .  Should I do anything more for it?

ines:
Yay, thanks! 👍 Just restarted the build manually – looks like a temporary GitHub glitch!

bdewilde:
Coming to this late — sorry! — but I'm running into issues using the latest `spacy` on macOS, where Python 2 builds are almost always (and by default) narrow unicode builds. (No problems in Python 3, of course!) Building with `--enable-unicode=ucs4` apparently leads to serious compatibility issues with other packages; you can see a bunch of discussion around this at https://github.com/pyenv/pyenv/pull/726 and the linked issue. I don't know if there's anything you can do about this, tbh, but worth underscoring that it may be a blocker for any stragglers still using py2 on macOS.

Issue #3339, opened 2019-02-27
💫 Entity Linking in spaCy


svlandeg:
## Feature description
With @honnibal & @ines we have been discussing adding an Entity Linking module to spaCy. This module would run on top of NER results and disambiguate & link tagged mentions to a knowledge base. We are thinking of implementing this in a few different phases:

1. Implement an efficient encoding of a knowledge base + all APIs / interfaces, to integrate with the current [processing pipeline](https://spacy.io/usage/processing-pipelines#custom-components-extensions). We would take the following components of EL into account:
   - Candidate generation
   - Encoding document context
   - Encoding local context 
   - Type prediction
   - Coreference resolution / ensuring global consistency
2. Implement a model that links English texts to English Wikipedia entries
3. Implement a cross-lingual model that links non-English texts to English Wikipedia entries
4. Fine-tune WP linking models to be able to ship them as such
5. Implement support in Prodigy to perform custom EL annotations for your specific project
6. Test / implement the models on a different domain & non-wikipedia knowledge base

## Notes
As some prior research, we compiled some notes on this project & its requirements: https://drive.google.com/file/d/1UYnPgx3XjhUx48uNNQ3kZZoDzxoinSBF. This contains more details on the EL components and implementation phases.

## Feedback requested
We will start implementing the APIs soon, but we would love to hear your ideas, suggestions, requests with respect to this new functionality first!

wejradford:
This sounds really exciting!

I'm curious how this relates to a task I've used spaCy for in the past (others may have too). The use-case is you're a user with a _small KB_, which is a set of entities (possibly with aliases) that you want to link text to. Currently, you can roll-your-own system using the existing NER with rule-based patches when required, then matching and ranking candidates. But, if you already had a big Wikipedia model available for linking, I can imagine wanting to try to match into the small KB then back-off to Wikipedia (or is there some weird KB-composition operation???), ideally with the same interfaces.

This kind of thing is almost certainly an MVP non-goal, but I'm interested to see if it's something the team is thinking about.

turbolent:
Is there any plan on integrating this with Wikidata?

svlandeg:
@turbolent : as our focus will be on linking to Wikipedia in the first phases, I think an integration with Wikidata will come naturally. There are crosslinks between the two anyway, and I've seen some prior work where the Wikidata knowledge graph was used to tune the prior probabilities for P(entity|alias). So yea, I think it's an important resource to exploit.

[...]

PR #3459, opened 2019-03-22
Basic framework and APIs for entity linker

svlandeg:
This PR is the first to start adressing Issue https://github.com/explosion/spaCy/issues/3339 to implement Entity Linking functionality in spaCy. While most functionality is still "dummy", any feedback on the code structuring & API's is welcome before continuing with the next bit of work. I'd expect some of this code to still change quite a bit when we actually start implementing non-dummy algorithms.

### Basic architecture and APIs for Entity Linking functionality:
 - spans and their underlying tokens get a `kb_id` and `ent_kb_id` field respectively

 - added a pipe called `el` to `nlp`, referring to an instance of `EntityLinker`
 - The current implementation of `EntityLinker` simply selects the candidate entity with highest prior probability.

 - `kb.pxd` and `kb.pyd` hold the knowledge base (based on @honnibal's notes)
    - The KB works like this: first add all entities. Then per alias/mention, add each candidate entity with its prior probability
    - `Candidate` is an object generated from the candidate_generator and used as input to the actual `EntityLinker`
    - Entity ID (string) is translated to Entity hash (using `vocab`) and then to Entry index (using `_entry_index`). The entry index then points to an `_EntryC` struct (ID, name, features, ...) in the `_entries` vector
    - Alias (string) is translated to Alias hash (using `vocab`) and then to Alias index (using `_alias_index`). The alias index then points to an `_AliasC` struct (entry candidates + prior probs) in the `_aliases_table` vector

 - added some unit tests to check the above behaviour
    - reading and writing of `kb_id` in span
    - valid and invalid construction of a KB

 - `examples/pipeline/dummy_entity_linking.py`: Shows how the current functionality works

### Open questions
 - `get_candidates` is sort of inbetween the internal KB structure and the actual implementation of the Entity Linker --> not sure where to put it. Currently part of `kb.pyx`. Different algorithms should be able to overwrite its behaviour but then again that may also influence the actual storage in KB (e.g. case invariance: "Douglas", "DOUGLAS" and "douglas" all mapping to the same entries/candidates). Something to think about :-)

 - should we always assume there is a `vocab` from `nlp`, or should we have it as optional argument and create a new one if none was given?

### Issues not yet addressed
 - vectors & features & types in kb (seems more easy to start thinking about this in the context of a non-dummy algorithm)
 - documentation not yet updated (perhaps it makes sense to continue with the next phases first, to get to more stable APIs)

### Types of change
New feature

## Checklist


spacy/kb.pyx
```diff
@@ -0,0 +1,147 @@
...
+
+cdef class Candidate:
+
+    def __init__(self, KnowledgeBase kb, entity_id_hash, alias_hash, prior_prob):
+        self.kb = kb
+        self.entity_id_hash = entity_id_hash
+        self.alias_hash = alias_hash
+        self.prior_prob = prior_prob
+
+    property entity_id:
```
ines: Super nitpicky, but now that Cython supports the `@property` syntax, we recently came up with our own convention for this: If there's only a getter, we use `@property` with `def`, if there's a getter and setter, we use `property foo:`.

setup.py
```diff
@@ -40,6 +40,7 @@ def is_new_osx():
     "spacy.lexeme",
     "spacy.vocab",
     "spacy.attrs",
+    "spacy.kb",
```
ines: @honnibal Do you think `kb` is okay for a module name here? Alternative could be `spacy.knowledge`, which I like as well. (But might be one of those words that's annoying to spell / easy to misspell?)

examples/pipeline/dummy_entity_linking.py
```diff
@@ -0,0 +1,69 @@
+# coding: utf-8
+"""Demonstrate how to build a simple knowledge base and run an Entity Linking algorithm.
+Currently still a bit of a dummy algorithm: taking simply the entity with highest probability for a given alias
+"""
```
ines: Also nitpicky, but we should add `from __future__ import unicode_literals` here and all `.py` files, just in case and for consistency.

examples/pipeline/dummy_entity_linking.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # adding aliases
+    print()
+    alias_0 = "Douglas"
+    print("adding alias", alias_0)
+    kb.add_alias(alias=alias_0, entities=["Q1004791", "Q42", "Q5301561"], probabilities=[0.1, 0.6, 0.2])
+
+    alias_1 = "Douglas Adams"
+    print("adding alias", alias_1)
+    kb.add_alias(alias=alias_1, entities=["Q42"], probabilities=[0.9])
```
ines: Really like the user-facing API so far 😃 Fits in well with our other components!

spacy/kb.pyx
```diff
@@ -0,0 +1,147 @@
...
+        Return the hash of the entity ID at the end
+        """
+        if not entity_name:
+            entity_name = entity_id
+        cdef hash_t id_hash = self.vocab.strings.add(entity_id)
+        cdef hash_t name_hash = self.vocab.strings.add(entity_name)
+
+        # Return if this entity was added before
+        if id_hash in self._entry_index:
+            user_warning("Entity " + entity_id + " already exists in the KB")
```
ines: Small code style edit:

```suggestion
            user_warning("Entity {} already exists in the KB".format(entity_id))
```

Soo looking forward to the day when we can switch the entire code base over to f-strings. (I think it shouldn't be a problem in Cython already, but we want things to be consistent so regular string formatting it is for now.)

**Edit:** Ah, just realised that this will kinda solve itself once the messages are moved to `spacy.errors`. Nevermind.

spacy/kb.pyx
```diff
@@ -0,0 +1,147 @@
...
+
+    def add_alias(self, unicode alias, entities, probabilities):
+        """
+        For a given alias, add its potential entities and prior probabilies to the KB.
+        Return the alias_hash at the end
+        """
+
+        # Throw an error if the length of entities and probabilities are not the same
+        if not len(entities) == len(probabilities):
+            raise ValueError("The vectors for entities and probabilities for alias '" + alias
```
ines: Not super important at this point, but these messages should probably all move to `spacy.errors` later on.

examples/pipeline/dummy_entity_linking.py
```diff
@@ -0,0 +1,69 @@
...
+    print("adding entity", entity_0)
+    kb.add_entity(entity_id=entity_0, entity_name="Douglas", prob=0.5)
+
+    entity_1 = "Q42"
+    print("adding entity", entity_1)
+    kb.add_entity(entity_id=entity_1, entity_name="Douglas Adams", prob=0.5)
+
+    entity_2 = "Q5301561"
+    print("adding entity", entity_2)
+    kb.add_entity(entity_id=entity_2, entity_name="Douglas Haig", prob=0.5)
```
ines: What's the significance of the `entity_name` vs. the `entity_id`? Is it just the human-readable version? Is it supposed to map to the `Span.text`?

(I'm not an entity linking expert, so there's a chance I'm missing something obvious here.)

svlandeg: It's like the one official standard human-readable name for an entity, like entity_id="Q42" and entity_name="Douglas Adams". But it's not unique, while entity_id is. I was thinking about whether we need this specifically, but from experience I think we're going to need this when printing things etc, humans don't usually like reading unique machine IDs...

spacy/kb.pyx
svlandeg: Please be as nitpicky as you like, especially since it's my first big contribution in Cython, and I don't know all the conventions yet!

ines: We don't have so many code style conventions for Cython yet (except for [these](https://spacy.io/api/cython#conventions), which you've probably seen already) – also because auto-formatting and linting is still a bit difficult for Cython.

(For Python, `black` and `flake8` enforce most of the standards, so that's much easier. With the current version, you should be able to just run `flake8 spacy` and it shouldn't produce any warnings. And if you do have to break one of the rules on purpose, like, assign an unused variable, you can add `# noqa: F841` etc to the line.)

examples/pipeline/dummy_entity_linking.py
ines: Ah okay, yes, that makes sense. The only potential problem I see here is that a user might falsely assume that the `entity_name` has significance for the result, or confuse it with an alias. Because in this example, the entity name and alias are identical ("Douglas Adams"), but the alias is what actually matters (right?).

svlandeg: Yea, that's right. Worse even, currently the `entity_name` isn't even recorded as an alias because aliases are recorded alias-by-alias to make sure the sum of their prior probabilities don't exceed 1. So yes, it's confusing :-(

setup.py
honnibal: I dunno, `knowledge` is a bit vague. It sounds like it holds *all* the knowledge.

On the topic of naming, I wonder whether `KnowledgeBase` is actually a bit problematic? It sounds like the base class of the `Knowledge` hierarchy, which is quite emphatically untrue. We might disprefer `spacy.knowledge_base.pyx` as a module name for similar reasons.

Maybe we want `spacy.entities`, to contrast with `spacy.tokens`?

examples/pipeline/dummy_entity_linking.py
honnibal: I think we'd probably prefer to force `name` to be unique, and then use it as the ID. For consistency with the rest of the API, we're always going to need to accept both string and hash versions of these things. It gets pretty unwieldy if you have a name and an ID, and then the "integer ID of the name" (how we describe the hash in the docs) isn't the same as the ID.

It's a pity we don't have the `P(mention | candidate)` distribution. If we did we could just use that to get the canonical mention for the entity, i.e. whatever people usually call it, which could be used as the human-readable name.

svlandeg: ok, so I remove (non-unique) `entity_name`, and rename `entity_id` to (unique) `entity_name` ? or just `entity` (as it was initially, I think) ?

I was also thinking along those lines @honnibal. Those probabilities would be too detailed perhaps, but you could imagine each alias having a type, such as "abbreviation" or "standard_name". But then you would be inclined to add the aliases entity-by-entity instead of mention-by-mention.

svlandeg: Another option would be to sort of extract a human-readible name for each entity, given the final annotations in the document(s). So if I processed all these texts and Q42 is most often linked to the alias "A.D.", then that becomes the "standard" string of Q42 in those docs. It is at least closest to the actual processed text & its user.

setup.py
ines: Yeah, I know what you mean. I don't really have a better idea either, though.

> Maybe we want `spacy.entities`, to contrast with `spacy.tokens`?

Hmm, not sure – I think that'd also be misleading, because people would expect the entity recognizer and entity ruler to live there (and IMO those should stay in `spacy.pipeline`, because they're pipeline components).

svlandeg: In EL literature, `knowledge base` is by far the most often used term. Other options I can think of are `thesaurus`/`ontology` or `concepts`. Let me know what you guys decide :-)

examples/pipeline/dummy_entity_linking.py
```diff
@@ -0,0 +1,71 @@
...
+
+    entity_2 = "Q5301561_Douglas_Haig"
+    print("adding entity", entity_2)
+    kb.add_entity(entity=entity_2, prob=0.5)
+
+    # adding aliases
+    print()
+    alias_0 = "Douglas"
+    print("adding alias", alias_0)
+    kb.add_alias(alias=alias_0, entities=[entity_0, entity_1, entity_2], probabilities=[0.1, 0.6, 0.2])
```
ibeltagy: I find the name `alias` a bit confusing. I would call this `mention` or `mention_text` or even `candidate_mention_text`, and I would use `aliases` to describe the list of alternative names for the entity. While they might be mappable in certain situations, they come from different sources and can serve different purposes.

svlandeg: So the `alias` here is in fact an alternative name for a set of given entities. The reason why we initially opted not to add the aliases when defining an entity (cf. your other comment), is because we want to facilitate defining prior probabilities `P(entity|alias)`. By adding these alias-by-alias, we can enforce that the sum of these prior probabilities (over all possible entities for 1 alias/mention) does not exceed 1, which is more difficult when you add the aliases at the same time when defining one entity. Does that make sense?

svlandeg: I agree with you though that if you don't have prior probabilities, adding aliases when defining the entities will be more intuitive...

examples/pipeline/dummy_entity_linking.py
```diff
@@ -0,0 +1,71 @@
...
+from spacy.kb import KnowledgeBase
+
+
+def create_kb(vocab):
+    kb = KnowledgeBase(vocab=vocab)
+
+    # adding entities
+    entity_0 = "Q1004791_Douglas"
+    print("adding entity", entity_0)
+    kb.add_entity(entity=entity_0, prob=0.5)
```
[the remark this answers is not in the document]
svlandeg: - Alternative names/aliases: cf. other comment
- Types: for sure will be added to the KB and used during disambiguation
- Definition: @honnibal : I'm thinking we may not want to load that into memory (similar as the previous discussion around "standard name") - what do you think?

setup.py
honnibal: Let's stick with `KnowledgeBase`, which I think makes the module `kb`. We can always change it before we ship, so let's not get hung up on it for now.

examples/pipeline/dummy_entity_linking.py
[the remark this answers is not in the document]
honnibal: Agreed re definition. I think we want to keep the KB's in-memory representations focused on the stuff you need for disambiguation. Once you have your entity ID's, you can make your DB lookups or API calls to get the extra data. We don't want to require DB lookups or API calls for the disambiguation though, so we need to stay disciplined about what we put in memory here.

examples/pipeline/dummy_entity_linking.py
honnibal: The problem with `mention` to me is it sounds like a *token*, when we want a *type* here. I think we need to keep the term `mention` for an instance of an entity in text. An `alias` would be the text that anchors a mention. I agree that it might not be what you first think of, but at least it will let us keep consistent terminology.

> I agree with you though that if you don't have prior probabilities, adding aliases when defining the entities will be more intuitive...

Yeah, but we want to make sure there's only one way to do it. So I think this API is better overall.

examples/pipeline/dummy_entity_linking.py
svlandeg: update: simplified things: https://github.com/explosion/spaCy/pull/3459/commits/8814b9010d139f92bc817378eace25e24a817b7e

honnibal: > ok, so I remove (non-unique) entity_name, and rename entity_id to (unique) entity_name ? or just entity (as it was initially, I think) ?

Yes, I think so.

> Another option would be to sort of extract a human-readible name for each entity, given the final annotations in the document(s).

Well, maybe. I actually think the unique ID won't be too bad; I mean for Wikipedia pages it's fine, and for other knowledge bases, people basically are responsible for coming up with a scheme that's useful to them. If something's lacking we could develop it later.

PR #3465, opened 2019-03-22
Dependency tree pattern matcher

skrcode:
Improving the DependencyTreeMatcher
## Description

This commit fixes the issues with regard to the DependencyTreeMatcher. The earlier matching algorithm was only a simple heuristic to handle the time complexity, which gave spurious results in quite a few occasions. The algorithm has been changed to a "correct" brute force one which iterates over a reduced list of all subgraphs. The algorithm runs in a time lower-bound by O(length of text ^ cardinality of pattern ). I think this would be near optimum since in the worst case, the number of matched patterns would be exponential and so we would definitely have to be lower-bound by this number.
## Checklist


honnibal:
Thanks! I think you're right that algorithmically this seems best. We can do a lot to improve the constants by moving loops into Cython as well.

I think the rename `dependencytreematcher` instead of `dependencymatcher` isn't better, though. `DependencyTreeMatcher` sounds like there's separately also a `DependencyGraphMatcher`. I'd prefer to keep it as `DependencyMatcher`.

setup.py
```diff
@@ -58,7 +58,7 @@ def is_new_osx():
     "spacy.tokens._retokenize",
     "spacy.matcher.matcher",
     "spacy.matcher.phrasematcher",
-    "spacy.matcher.dependencymatcher",
+    "spacy.matcher.dependencytreematcher",
```
honnibal:
```suggestion
    "spacy.matcher.dependencymatcher",
```

spacy/matcher/__init__.py
```diff
@@ -3,6 +3,6 @@
 
 from .matcher import Matcher
 from .phrasematcher import PhraseMatcher
-from .dependencymatcher import DependencyTreeMatcher
+from .dependencytreematcher import DependencyTreeMatcher
```
honnibal:
```suggestion
from .dependencymatcher import DependencyMatcher
```

spacy/tests/matcher/test_matcher_api.py
```diff
@@ -288,42 +288,41 @@ def deps():
...
+        {"SPEC": {"NODE_NAME": "quick", "NBOR_RELOP": ".", "NBOR_NAME": "jumped"}, "PATTERN": {"ORTH": "fox"}}
+    ]
+
+    pattern3 = [
+        {"SPEC": {"NODE_NAME": "jumped"}, "PATTERN": {"ORTH": "jumped"}},
+        {"SPEC": {"NODE_NAME": "fox", "NBOR_RELOP": ">", "NBOR_NAME": "jumped"}, "PATTERN": {"ORTH": "fox"}},
+        {"SPEC": {"NODE_NAME": "r", "NBOR_RELOP": ">>", "NBOR_NAME": "fox"}, "PATTERN": {"ORTH": "brown"}}
     ]
+
     matcher = DependencyTreeMatcher(en_vocab)
```
honnibal:
```suggestion
    matcher = DependencyMatcher(en_vocab)
```

honnibal:
```suggestion
    matcher = DependencyMatcher(en_vocab)
```

spacy/matcher/__init__.py
```diff
@@ -3,6 +3,6 @@
...
 
 __all__ = ["Matcher", "PhraseMatcher", "DependencyTreeMatcher"]
```
honnibal:
```suggestion
__all__ = ["Matcher", "PhraseMatcher", "DependencyMatcher"]
```

skrcode:
@honnibal Thanks for the suggestions ! I've got the renaming done. For the loops, I'm not quite sure as to what level of optimization is required. For the recurse method(which is the bottleneck), I've only converted lists to numpy arrays in the loops and added types to a few other variables.

cyclecycle:
Would like to use this. Would be great if it could be merged. Thanks a lot for your work on this.

skrcode:
@cyclecycle Have you tried out this particular commit or the previous ones? It would be great if you could provide feedback on areas such as correctness, usability and performance.

cyclecycle:
@skrcode Apologies for the delayed response. This commit is now readily usable for me since it is present in the spaCy release 2.1.5 so I have started testing it.

The first thing I've noticed in updating my code to work with this version is that the return format seems to have changed from:

```python
[(15329811787164753587, [1, 0, 3])]
```

to:

 ```python
[(15329811787164753587, [[1, 0, 3]])]
```

Could you please explain this? I may be missing something.

Otherwise. everything seems to be working as expected so far. I plan to use this module a fair bit in the near future so I'll be sure to report any unexpected results if I encounter them.

Cheers

skrcode:
@cyclecycle Thanks for your response. Regarding the return format, the output is essentially a list of trees(subtrees) as opposed to a list of nodes in the traditional matcher. A tree is represented as a list of nodes and hence the return format would be a list of list of nodes. 

The module initially just returned the first match found but after looking at corenlp, I figured out that a list of tree matches is probably the right way to go about it. From a memory perspective, in order to be in-memory efficient, it would probably make sense later on to just store the roots of the tree matches  instead of the entire tree but I think that we'll need to see how memory usage plays out practically in order to consider changing this.

PR #3492, opened 2019-03-27
DOC: Include default value for batch_size in tokenizer pipe in docs

Bharat123rox:
Include the default value of 1000 in `spacy.Tokenizer` docs [here](https://github.com/explosion/spaCy/blob/master/spacy/tokenizer.pyx) and [here](https://github.com/explosion/spaCy/blob/master/website/docs/api/tokenizer.md)

## Checklist


ines:
Thanks! 👍

PR #3495, opened 2019-03-27
fix(util): fix decaying function output

SamuelLKane:
## Description
Primary change is moving from division to subtraction to yield the __nearly__ linear series. Cleaned up some unneeded code to simplify the method and made it match how `stepping` and `compounding` are implemented.

### Type of change
Bug fix

## Checklist


honnibal:
Thanks!

PR #3500, opened 2019-03-28
adds textpipe to universe

graus:
Adds textpipe to spaCy's universe  

## Checklist


ines:
Thanks for adding this so quickly! 👍

I slightly adjusted some of the formatting etc. and made one small change: I think it fits better into the `standalone` category, since it's a standalone package using spaCy, rather than a pipeline component or plugin that's added *to* spaCy.

PR #3519, opened 2019-03-31
Add ability to train from .txt files for `spacy pretrain`

tokestermw:
`spacy pretrain` currently requires all data to be in `.jsonl` files where each line is a JSON of the form: `{"text": "Blabla"}`. JSONL makes sense when there are labels in the data, but pretrain only uses the raw text. So this change makes it possible to train from plain text files if the file has a `.txt` suffix..

Small change, but I think it greatly increases the usability for first-timers. Also, a lot of language model data are plain text so you don't need to produce another ~GB file with essentially the same information.

## Description

Save this file to `sample_sents.jsonl`

```
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
```

You should be able to train from `.jsonl` per usual.

```
spacy pretrain sample_sents.jsonl en_core_web_md here -nw 1 -bs 1 -i 10
```

Save this file as `sample_sents.txt`:

```
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
Hello there.
```

You should be able to train as usual.

```
spacy pretrain sample_sents.txt en_core_web_md here -nw 1 -bs 1 -i 10
```

### Types of change

New feature for `spacy pretrain`.

#### TODO

* [ ] Change docs?

## Checklist

### DIFF
--- a/spacy/cli/pretrain.py
+++ b/spacy/cli/pretrain.py
@@ -86,8 +86,13 @@ def pretrain(
         texts_loc = Path(texts_loc)
         if not texts_loc.exists():
             msg.fail("Input text file doesn't exist", texts_loc, exits=1)
-        with msg.loading("Loading input texts..."):
-            texts = list(srsly.read_jsonl(texts_loc))
+        if texts_loc.suffix == ".txt":
+            with msg.loading("Loading input texts..."):
+                with open(texts_loc, "r") as f:
+                    texts = [{"text": line} for line in f]
+        else:
+            with msg.loading("Loading input texts..."):
+                texts = list(srsly.read_jsonl(texts_loc))
         msg.good("Loaded input texts")
         random.shuffle(texts)
     else:  # reading from stdin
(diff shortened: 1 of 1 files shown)


honnibal:
Well, something I've learned is that `.txt` is sort of a trap. It looks simple, but as soon as you need to delimit multiple documents, you have a problem. What if the texts contain the newline character? Then you need to escape them. If you say "well, documents can't contain the newline character", then what if you encounter newlines at runtime? The pretraining won't have had the chance to learn that.

If your data is in newline delimiter `.txt` and you really want to keep it that way, you can just have a script that reads it and prints the jsonl lines to stdout. Then you can pipe that forward into `spacy pretrain`.

tokestermw:
OK that makes sense. Will do that.

Issue #3521, opened 2019-04-01
Contraction "n’t" not tagged as stop word ("n't" with dumb quote is fine)


admackin:
## How to reproduce the behaviour

```
en = en_core_web_md.load(disable=['parser', 'ner', 'textcat'])
tok = en("don’t")[1]  # note that it is smartquote-like "’" not dumb "'"
print(tok.lemma_)   #  correctly returns "not"
print(tok.is_stop)   # incorrectly returns False
```

## Info about spaCy

* **spaCy version:** 2.1.3
* **Platform:** Darwin-17.7.0-x86_64-i386-64bit
* **Python version:** 3.6.4

ines:
Thanks for the report – looks like we might have to add `n’t` to the list of English stop words as well.

svlandeg:
I suggest to make a little loop in [stop_words.py](https://github.com/explosion/spaCy/blob/master/spacy/lang/en/stop_words.py) that goes through all hyphens for those stopwords that are hyphenated, then we'll catch all cases. I can quickly draft that PR if noone's currently working on it.

svlandeg:
Should be fixed with PR https://github.com/explosion/spaCy/pull/3530 !

Issue #3449, opened 2019-03-20
Tokenization regression


christian-storm:
## How to reproduce the behaviour
Each sentence in https://github.com/explosion/spaCy/issues/1266#issue-250718778 is tokenized with I. as one token.

## Your Environment

* **spaCy version:** 2.0.18
* **Platform:** Darwin-18.2.0-x86_64-i386-64bit
* **Python version:** 3.7.2

svlandeg:
Thanks for the report - I'll have a look!

svlandeg:
Ok I looked into it. Just a quick question: which version of spaCy are you using now, and was this ever working in a previous version? 

It does look like this is the intended default behaviour, though. https://github.com/explosion/spaCy/blob/master/spacy/lang/punctuation.py#L30-L31 defines a suffix rule for when **one lower-case** character comes before a dot, and https://github.com/explosion/spaCy/blob/master/spacy/lang/punctuation.py#L33 for when **two upper-case** characters are followed by a dot. 

It looks virtually impossible to change these default rules to accomodate your use-case for **one upper-case letter**, because it would break other cases such as "Tonight I saw A. Douglas." or "I live in the U.S. since 2 years." And it feels like you would not often have grammatically correct English sentences with `I` at the end?

It's virtually impossible to cover all use-cases entirely correct out of the box, but you can always write your own custom tokenizer to work around specific issues in your dataset ofcourse!

christian-storm:
Thanks for taking a look into this.  This behavior was with Spacy 2.0.18.  I had filed a previous bug (#1266) for 1.9.  @honnibal  replied a few months later that it had been fixed leading me to believe that it is worth fixing.  I also agree that this case is fairly unlikely given it usually arises with ungrammatical sentences.  However, a lot of text is ungrammatical.  My 12 year old twins, for instance, continually use I/me incorrectly.

I hear you with regards to the rules.  However, with 2.1, where the tokenization is no longer rule based, it seems the cue that "I." is a noun would hint at that it should be "I" PRON and "." PUNCT.  In your example, "Tonight I saw A. Douglas." spacy correctly finds "A." PROPN "Douglas" PROPN "." PUNCT leading me to believe that would work.  Maybe I'm barking up the wrong tree but it seems worth considering.  To that end perhaps this is actually fixed in 2.1?!

With 2.1 writing my own tokenizer to fix this is finally possible.  Before it was impossible to split a token.  Cheers to 2.1!

ines:
[...]

PR #3530, opened 2019-04-02
Allow English stopwords with any type of apostrophe

svlandeg:
Fixing Issue https://github.com/explosion/spaCy/issues/3521: Contraction `n’t` not tagged as stop word.

## Description

- I removed the English stop words with a hyphen from the generic stop list and added a small loop that for each such hyphenated stopword, adds all variants with all potential hyphens.
- I added a relevant unit test that succeeds after implementing above fix
- I added a failing unit test for Issue https://github.com/explosion/spaCy/issues/3449 (an issue I looked into before but can't be fixed generically)

### Types of change
Small fix

## Checklist

### DIFF
--- a/spacy/lang/en/stop_words.py
+++ b/spacy/lang/en/stop_words.py
@@ -69,7 +69,10 @@ yet you your yours yourself yourselves
 """.split()
 )
 
-for hyphen in ["'", "`", "‘", "´", "’"]:
-    for stopword in u"n't 'd 'll 'm 're 's 've".split():
-        STOP_WORDS.add(stopword.replace("'", hyphen))
+contractions = ["n't", "'d", "'ll", "'m", "'re", "'s", "'ve"]
+STOP_WORDS.update(contractions)
+
+for apostrophe in ["‘", "’"]:
+    for stopword in contractions:
+        STOP_WORDS.add(stopword.replace("'", apostrophe))
 
--- a/spacy/lang/en/stop_words.py
+++ b/spacy/lang/en/stop_words.py
@@ -39,7 +39,7 @@ made make many may me meanwhile might mine more moreover most mostly move much
 must my myself
 
 name namely neither never nevertheless next nine no nobody none noone nor not
-nothing now nowhere n't
+nothing now nowhere 
 
 of off often on once one only onto or other others otherwise our ours ourselves
 out over own
@@ -66,7 +66,10 @@ whereafter whereas whereby wherein whereupon wherever whether which while
 whither who whoever whole whom whose why will with within without would
 
 yet you your yours yourself yourselves
-
-'d 'll 'm 're 's 've
 """.split()
 )
+
+for hyphen in ["'", "`", "‘", "´", "’"]:
+    for stopword in "n't 'd 'll 'm 're 's 've".split():
+        STOP_WORDS.add(stopword.replace("'", hyphen))
+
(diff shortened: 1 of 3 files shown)


spacy/lang/en/stop_words.py
```diff
@@ -66,7 +66,10 @@
 whither who whoever whole whom whose why will with within without would
 
 yet you your yours yourself yourselves
-
-'d 'll 'm 're 's 've
 """.split()
 )
+
+for hyphen in ["'", "`", "‘", "´", "’"]:
```
ines: Hypen = apostrophe, right?

Also, maybe we should consider leaving out \` and ´? Those could possible produce unexpected results. Not that it matters much, but it could be confusing if a string like \`m is randomly declared a stop word.

And we can leave out the `"'"` as well, unless that's a different unicode character? Otherwise, the first loop would replace `'` with itself?

spacy/lang/en/stop_words.py
```diff
@@ -66,7 +66,10 @@
...
+    for stopword in u"n't 'd 'll 'm 're 's 've".split():
```
ines:
```suggestion
    for stopword in ["n't", "'d", "'ll", "'m", "'re", "'s", "'ve"]:
```

I'd actually write this as a list here, because it's more straightforward. The `split` syntax is easier for long lists like the stop words, but inline, I think a real list is easier to read. Also, no need for the `u` – we have the unicode declarations at the top 🎉

spacy/tests/regression/test_issue3449.py
```diff
@@ -0,0 +1,23 @@
+# coding: utf8
+import pytest
+
```
ines: For consistency:

```suggestion
from __future__ import unicode_literals

```

spacy/tests/regression/test_issue3521.py
```diff
@@ -0,0 +1,17 @@
+# coding: utf8
+import pytest
+
```
ines:
```suggestion
from __future__ import unicode_literals

```

spacy/tests/regression/test_issue3521.py
```diff
@@ -0,0 +1,17 @@
...
+
+@pytest.mark.parametrize(
+    "word",
+    [
+        u"don't",
```
ines: No big deal, but we can leave out the `u` here as well!

spacy/lang/en/stop_words.py
svlandeg: I'll address your first two points, but I think we need to keep `"'"`, otherwise `n't` wouldn't actually be added to the stopwords as it's not in the original list anymore. The replacement is indeed a bit dummy, but the code is compacter this way?

spacy/lang/en/stop_words.py
svlandeg: Yea, I forgot to add that declaration initially and ran into the unit tests for Python 2.7 ofcourse. Wonder how many more times I'm going to make the same mistake ;-)

spacy/lang/en/stop_words.py
ines: Ah, yes, makes sense 🤦‍♀️ 

I guess an alternative would be something like:

```python
contractions = ["n't", "'d", "'ll", "'m", "'re", "'s", "'ve"]
STOP_WORDS.update(contractions)
# and then go on with the for loop
```

PR #3538, opened 2019-04-03
Improved training and evaluation

kowaalczyk:
- early stopping option in `spacy train`
- return scores option in `spacy evaluate`
- fix bug: `spacy train` did not convert output path from string to Path

## Description

While I haven't seen any issues related to early stopping or returning evaluation scores,
I have implemented it because I needed it for my own use. The bug I have found obviously needs fixing, as for the two new features:

- Early stoppping stops training after the validation score did not improve for N iterations (a new optional parameter to `spacy train`) - it is much easier to impelment inside `spacy train` than anywhere else, it does not break API nor any conventions mentioned in contributing guide and should be useful for most people.

- `return_scores` option for `spacy evaluate` makes it return a dict of metric values when called from Python, and does not affect the script behavior when called from bash (or any other shell). Generally, spacy cli offers a nice abstraction over training, evaluation, etc. In my opinion, you should consider adding returns to more functions, as that will make the use of cli in python more _"pythonic"_

My use case of optimizing model hyperparametres would basically require me to re-write or copy code of `spacy.cli.train` and `spacy.cli.evaluate`, so I decided to make changes to spacy instead. I tried running the scripts on their own on Ubuntu as well as implemented a [hyperparameter grid search script that these changes were initially intended for](https://github.com/spacy-pl/utils/blob/master/training/models.py) and it all works as intended.

The only thing missing in my PR is the documentation, but I would first like to hear your opinion on these changes.

### Types of change

Bug fix + enhancement.

## Checklist

### DIFF
--- a/spacy/cli/evaluate.py
+++ b/spacy/cli/evaluate.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals, division, print_function
 import plac
 from timeit import default_timer as timer
 from wasabi import Printer
-import srsly
 
 from ..gold import GoldCorpus
 from .. import util
--- a/spacy/cli/evaluate.py
+++ b/spacy/cli/evaluate.py
@@ -4,6 +4,7 @@ from __future__ import unicode_literals, division, print_function
 import plac
 from timeit import default_timer as timer
 from wasabi import Printer
+import srsly
 
 from ..gold import GoldCorpus
 from .. import util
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks so much for this! I've been meaning to add some basic creature comforts to the training script for a long time now.

While you're there, my number one wish would actually be a simple bit of save/resume logic. This would allow cheaper training, as we'd be able to use preemptible instances on GCP.

I'm busy today but I hope to be able to go over your changes in the next day or two. But definitely keen to accept the feature.

kowaalczyk:
@honnibal I'm glad to hear that, I have also found and fixed another bug in `spacy.cli.train`:

For some reason, if training was performed on gpu there was a broken line of code in the cpu evaluation phase that was crashing entire process (fix is in [the last commit](https://github.com/explosion/spaCy/pull/3538/commits/b2609e9251791258f5d716153a85f7b302338054), I have also rebased the branch hence the force push).

As for the save/resume logic that you mentioned, I will certainly think about it but currently I am loaded with other tasks related to Polish language implementation that me & my team are working on.

Issue #3523, opened 2019-04-01
Training a new model using cli throws error `KeyError`


xssChauhan:
I am trying to train a new spacy model based on the [Tweebank](https://github.com/Oneplus/Tweebank) annotated data. 
For that I first tried using the training example given in the docs to familiarize myself with the procedure. 
Example and training on the Tweebank throw the same error.

## How to reproduce the behaviour
Follow the example [here](https://spacy.io/usage/training#spacy-train-cli)
For the sake of completeness:
```
git clone https://github.com/UniversalDependencies/UD_Spanish-AnCora
mkdir ancora-json
python -m spacy convert UD_Spanish-AnCora/es_ancora-ud-train.conllu ancora-json
python -m spacy convert UD_Spanish-AnCora/es_ancora-ud-dev.conllu ancora-json
mkdir models
python -m spacy train es models ancora-json/es_ancora-ud-train.jsonl ancora-json/es_ancora-ud-dev.jsonl
```

## Info about spaCy

* **spaCy version:** 2.1.3
* **Platform:** Linux-4.15.0-46-generic-x86_64-with-debian-buster-sid
* **Python version:** 3.6.7
* **Models:** en_core_web_md, en_core_web_sm

## The Error

```
>>> python -m spacy train es models es_ancora-ud-train.jsonl es_ancora-ud-dev.jsonl
Training pipeline: ['tagger', 'parser', 'ner']
Starting with blank model 'es'
Counting training words (limit=0)
Traceback (most recent call last):
  File "/home/shikhar/.conda/envs/factmata/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/shikhar/.conda/envs/factmata/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/shikhar/.conda/envs/factmata/lib/python3.6/site-packages/spacy/__main__.py", line 35, in <module>
    plac.call(commands[command], sys.argv[1:])
  File "/home/shikhar/.conda/envs/factmata/lib/python3.6/site-packages/plac_core.py", line 328, in call
    cmd, result = parser.consume(arglist)
  File "/home/shikhar/.conda/envs/factmata/lib/python3.6/site-packages/plac_core.py", line 207, in consume
    return cmd, self.func(*(args + varargs + extraopts), **kwargs)
[...]

PR #3542, opened 2019-04-04
fix typos in tag_map flagged by `python -m debug-data`

jeannefukumaru:
## Checklist


ines:
Thanks 👍

This is a really clever idea for creating the tag map, actually! Run `debug-data`, see what's missing and update the tags accordingly. I should add this to the adding models thread as a tip 🙂

jeannefukumaru:
Awesome, thanks! It came from trying to train a language model and seeing one of your comments on an issue thread actually https://github.com/explosion/spaCy/issues/3523#issuecomment-478535869 https://github.com/explosion/spaCy/issues/3523#issue-427642393 I wouldn't have known about it otherwise!

PR #3543, opened 2019-04-04
Add hyphen prefix for French proper nouns

mauryaland:
## Description

Add hyphen prefix for French proper nouns

### Types of change

I added several hyphen prefix for French proper nouns relative to those present in the _tokenizer_exceptions_list.py. I didn't add those which appear only one time in that list even if the list is not used anymore. Let me know if you think that it is a good idea to add more hyphen prefix relative to city names for example. I guess you do not want that this list become too large.

I also updated _tokenizer_exceptions_list.py taking into accounts the changes.

## Checklist


svlandeg:
Hi @mauryaland ! You're right that the `_tokenizer_exceptions_list.py` isn't used anymore, because we noticed it was slowing down the tokenizer too much with virtually no effect on accuracy (as measured on UD): https://github.com/explosion/spaCy/issues/3242#issuecomment-461420739

I think we have to be careful not to let the `tokenizer_exceptions.py` become this huge list again. But then on the other hand I agree with you that there are many names and cities with this `Xxxx-Xxxx` format. One suggestion at the time by @honnibal (https://github.com/explosion/spaCy/issues/3242#issuecomment-464447606) was to keep the oversegmentation and hope that the model will learn to rejoin these cases, but I'm not sure it's been tried yet though whether that works.

So ideally we should work towards a more generic/elegant solution instead of maintaining these huge lists. What do you think?

mauryaland:
Hello @svlandeg, you are totally right. Should we put in place a regex that match the pattern `Xxxx-Xxxx`  you mentioned ?

svlandeg:
I think that would be preferred above the long list of exceptions, considering it's such a common case in French. But then on the other hand I know @honnibal prefers oversegmentation so I'm not sure. Have you experimented with a statistical model to see whether it would learn to rejoin these cases by itself? Or do you have a use-case where you're using the blank tokenizer only?

mauryaland:
Yes, I agree with you for the case of hyphenated proper nouns, it make sense to catch them with the pattern `Xxxx-Xxxx`. Plus, if I understand well, @honnibal in the message you quoted say that he would like to keep the hyphen as an infix. So I guess it is the right solution.

If we go in that direction, we should consider for the pattern cases like "Amélie-les-Bains-Palalda" or "Auriac-l'Église"

I have not experimented with a statistical model because I am only using the French deterministic tokenizer.

honnibal:
We now have a new format for storing language data, that prevents the problem of the data files being too large on disk. It's unclear whether we want these exceptions by default, but we might at least provide an option now, without increasing the library footprint too much.

See here for details about the json data: https://github.com/explosion/spaCy/pull/4167

If you like, you could resubmit the PR with the exceptions in json. I'll close this in the meantime.

PR #3548, opened 2019-04-06
Added Ludwig among the projects

w4nderlust:
Just added Ludwig among the projects in Universe as it uses spaCy.
Also feel free to add Uber among the company that use spaCy if you want.


ines:
Thanks! 👍

PR #3565, opened 2019-04-09
Fix website docs for Vectors.from_glove

bryant1410:


ines:
Thanks for spotting and fixing this! 👍

PR #3573, opened 2019-04-11
Update jsonl file change

pickfire:
## Checklist


pickfire:
I already added myself to contributors, should be able to add this now.

ines:
Thanks! 👍 The underlying default was now changed in #3583, so we don't need this update to the docs anymore at this point (only when we "officially" switch over to JSONL as the default).

PR #3510, opened 2019-03-30
Add save after `--save-every` batches for `spacy pretrain`

tokestermw:
When using `spacy pretrain`, the model is saved only after every epoch. But each epoch can be very big since `pretrain` is used for language modeling tasks. So I added a `--save-every` option in the CLI to save after every `--save-every` batches.

## Description

To test...

Save this file to `sample_sents.jsonl`

```
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
```

Then run `--save-every 2` when pretraining.

```bash
spacy pretrain sample_sents.jsonl en_core_web_md here -nw 1 -bs 1 -i 10 --save-every 2
```

And it should save the model to the `here/` folder after every 2 batches. The models that are saved during an epoch will have a `.temp` appended to the save name.

At the end the training, you should see these files (`ls here/`):

```bash
config.json     model2.bin      model5.bin      model8.bin
log.jsonl       model2.temp.bin model5.temp.bin model8.temp.bin
model0.bin      model3.bin      model6.bin      model9.bin
model0.temp.bin model3.temp.bin model6.temp.bin model9.temp.bin
model1.bin      model4.bin      model7.bin
model1.temp.bin model4.temp.bin model7.temp.bin
```

### Types of change

This is a new feature to `spacy pretrain`.

🌵 **Unfortunately, I haven't been able to test this because compiling from source is not working (cythonize error).** 

```
Processing matcher.pyx
[Errno 2] No such file or directory: '/Users/mwu/github/spaCy/spacy/matcher.pyx'
Traceback (most recent call last):
  File "/Users/mwu/github/spaCy/bin/cythonize.py", line 169, in <module>
    run(args.root)
  File "/Users/mwu/github/spaCy/bin/cythonize.py", line 158, in run
    process(base, filename, db)
  File "/Users/mwu/github/spaCy/bin/cythonize.py", line 124, in process
    preserve_cwd(base, process_pyx, root + ".pyx", root + ".cpp")
  File "/Users/mwu/github/spaCy/bin/cythonize.py", line 87, in preserve_cwd
    func(*args)
[...]

PR #3586, opened 2019-04-12
Add `--latest` option in `spacy train` to save latest n epochs of CLI training

### LINKED ISSUE #3584: Add a parameter for saving only the n latest epochs while using CLI Training
## Feature description

Enable saving models of the `n_latest` epochs alone to disk (say `n_latest=5`, save only the last 5 epochs to disk) in spacy train cli [here](https://spacy.io/api/cli#train).
This would ideally be a parameter, `n_latest`, specified by the user, to save only that many epochs to disk at the end of training, for example:-

If `n_iter=9` and `n_latest=5`, then only model5 to model9 must be saved on disk, however to prevent losing information in the event of some crash/errors, a suggested implementation is model0 may be deleted only after model5 is saved, model1 may be deleted only after model6 is saved etc.

This will help users decide the appropriate tradeoffs for disk space vs ability to choose the "best" model.

In my case, I certainly require this feature as I'm running out of disk space.

@ines  @honnibal  I'm ok to submit a PR for this, if this feature seems ok. (Even if it is a breaking change, I feel it's essential moving forward and some steps could be taken to make it a v2.1 feature)

## Could the feature be a [custom component](https://spacy.io/usage/processing-pipelines#custom-components) or [spaCy plugin](https://spacy.io/universe)?
If so, we will tag it as [`project idea`](https://github.com/explosion/spaCy/labels/project%20idea) so other users can take it on.

No, since it is a parameter specific to Spacy's CLI Training

Bharat123rox:
Close #3584 by adding an optional parameter `--latest`
## Description

This parameter `--latest` saves only the latest `n` epochs trained by the model. This could be more beneficial since the model improves with training. `0` or a negative integer enables saving all epochs, preserving the defaults.

All tests pass except for this warning:
```python3
spaCy/bin/ud/ud_train.py:45
  /home/user/spaCy/bin/ud/ud_train.py:45: DeprecationWarning: invalid escape sequence \s
    space_re = re.compile("\s+")
```

Also refer #3510 for a modification of this idea (save model after every `n` epochs) in `spacy pretrain`  

## Checklist

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -36,6 +36,7 @@ from .. import about
     vectors=("Model to load vectors from", "option", "v", str),
     n_iter=("Number of iterations", "option", "n", int),
     n_examples=("Number of examples", "option", "ns", int),
+    latest=("Number of epochs to save from the latest epoch", "option", "l", int),
     use_gpu=("Use GPU", "option", "g", int),
     version=("Model version", "option", "V", str),
     meta_path=("Optional path to meta.json to use as base.", "option", "m", Path),
@@ -74,6 +75,7 @@ def train(
     pipeline="tagger,parser,ner",
     vectors=None,
     n_iter=30,
+    latest=0,
     n_examples=0,
     use_gpu=-1,
     version="0.0.0",
@@ -286,6 +288,8 @@ def train(
                             cpu_wps = nwords / (end_time - start_time)
                     acc_loc = output_path / ("model%d" % i) / "accuracy.json"
                     srsly.write_json(acc_loc, scorer.scores)
+            if latest > 0 and i >= latest:
+                shutil.rmtree(output_path / ("model%d" % (i - latest)))
 
                     # Update model meta.json
                     meta["lang"] = nlp.lang
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -288,8 +288,6 @@ def train(
                             cpu_wps = nwords / (end_time - start_time)
                     acc_loc = output_path / ("model%d" % i) / "accuracy.json"
                     srsly.write_json(acc_loc, scorer.scores)
-            if latest > 0 and i >= latest:
-                shutil.rmtree(output_path / ("model%d" % (i - latest)))
 
                     # Update model meta.json
                     meta["lang"] = nlp.lang
@@ -332,6 +330,10 @@ def train(
                         gpu_wps=gpu_wps,
                     )
                     msg.row(progress, **row_settings)
+
+            if latest > 0 and i >= latest:
+                shutil.rmtree(output_path / ("model%d" % (i - latest)))
+
     finally:
         with nlp.use_params(optimizer.averages):
             final_model_path = output_path / "model-final"
--- a/website/docs/api/cli.md
+++ b/website/docs/api/cli.md
@@ -198,7 +198,7 @@ will only train the tagger and parser.
 
 ```bash
 $ python -m spacy train [lang] [output_path] [train_path] [dev_path]
-[--base-model] [--pipeline] [--vectors] [--n-iter] [--n-examples] [--use-gpu]
+[--base-model] [--pipeline] [--vectors] [--n-iter] [--latest] [--n-examples] [--use-gpu]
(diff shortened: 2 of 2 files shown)


honnibal:
@Bharat123rox I definitely agree it's annoying the way the disk saving works at the moment. I wonder whether this is the best solution though.

Perhaps it would be better to retain the best model for each component? We can always write out the latest one, and then if the score is better for the parser, we retain the parser model, if the score is better for the tagger, we retain the tagger model, etc. This would prevent the disk from filling up, while being a bit more satisfying than only using the latest iterations, which might not be the best ones.

Bharat123rox:
@honnibal Can I implement this next week as I will be away from work for the rest of this week (and also, I'm still struggling to find how to extract the metrics for each epoch at the moment, but I will figure it out soon)

website/docs/api/cli.md
```diff
@@ -198,7 +198,7 @@ will only train the tagger and parser.
 
 ```bash
 $ python -m spacy train [lang] [output_path] [train_path] [dev_path]
-[--base-model] [--pipeline] [--vectors] [--n-iter] [--n-examples] [--use-gpu]
+[--base-model] [--pipeline] [--vectors] [--n-iter] [--latest] [--n-examples] [--use-gpu]
```
ines:
```suggestion
[--base-model] [--pipeline] [--vectors] [--n-iter] [--n-latest] [--n-examples] [--use-gpu]
```

website/docs/api/cli.md
```diff
@@ -213,7 +213,8 @@ $ python -m spacy train [lang] [output_path] [train_path] [dev_path]
...
 | `--pipeline`, `-p` <Tag variant="new">2.1</Tag>       | option        | Comma-separated names of pipeline components to train. Defaults to `'tagger,parser,ner'`.                                                                         |
 | `--vectors`, `-v`                                     | option        | Model to load vectors from.                                                                                                                                       |
-| `--n-iter`, `-n`                                      | option        | Number of iterations (default: `30`).                                                                                                                             |
+| `--n-iter`, `-n`                                      | option        | Number of iterations (default: `30`).|
+| `--latest`, `-l`                                      | option        | Number of epochs to save from the latest epoch (`0` or a negative integer to save all epochs, default: `0`).|
```
ines:
```suggestion
| `--n-latest`, `-l`                                      | option        | Number of epochs to save from the latest epoch (`0` or a negative integer to save all epochs, default: `0`).|
```

spacy/cli/train.py
```diff
@@ -36,6 +36,7 @@
     vectors=("Model to load vectors from", "option", "v", str),
     n_iter=("Number of iterations", "option", "n", int),
     n_examples=("Number of examples", "option", "ns", int),
+    latest=("Number of epochs to save from the latest epoch", "option", "l", int),
```
ines:
```suggestion
    n_latest=("Number of epochs to save from the latest epoch", "option", "l", int),
```

spacy/cli/train.py
```diff
@@ -74,6 +75,7 @@ def train(
     pipeline="tagger,parser,ner",
     vectors=None,
     n_iter=30,
+    latest=0,
```
ines:
```suggestion
    n_latest=0,
```

spacy/cli/train.py
```diff
@@ -328,6 +330,10 @@ def train(
                         gpu_wps=gpu_wps,
                     )
                     msg.row(progress, **row_settings)
+
+            if latest > 0 and i >= latest:
```
ines:
```suggestion
            if n_latest > 0 and i >= n_latest:
```

spacy/cli/train.py
```diff
@@ -328,6 +330,10 @@ def train(
...
+                shutil.rmtree(output_path / ("model%d" % (i - latest)))
```
ines:
```suggestion
                shutil.rmtree(output_path / ("model%d" % (i - n_latest)))
```

Also, if I remember correctly, `shutil.rmtree` takes a string, not a path? I think I fixed some issue recently where this caused a warning. We have a `path2str` helper in `spacy.compat` that you can use for this.

PR #3591, opened 2019-04-15
Add optional `id` property to EntityRuler patterns

kabirkhan:
Adds an  optional `id` property to EntityRuler patterns that gets added as a custom extension at _.entity_id  for entities that have it defined.

This solves the one downside of switching from an custom Entity Matcher implemenation to the EntityRuler component of providing a pattern identifier to do super basic "Entity Linking"

Ideally this will be replaced longer term with the Entity Linker work described here https://github.com/explosion/spaCy/issues/3339

### Types of change

Enhancement

No docs written yet, just want to get feedback first.

## Checklist


honnibal:
I agree with the general concept, but I think we need to be careful to avoid ever setting custom attributes in the main library. We want to make sure these are reserved for user data. It's confusing if the library starts setting some of these.

Currently tokens do have an `ent_id` attribute, which isn't a wonderful mechanism as the data should be attached to spans, but it's the best we have at the moment. Perhaps we should wait on this to make sure it integrates properly with how things are handled by the entity linking changes that @svlandeg has been working on?

svlandeg:
So for the entity linking functionality I added a `kb_id` field to Span objects which can be retrieved as `token.ent_kb_id_` or `ent.kb_id_`, with the ID referring back to entries in a `KnowledgeBase`. Dummy example here: [dummy_entity_linking.py](https://github.com/explosion/spaCy/blob/master/examples/pipeline/dummy_entity_linking.py). Not sure whether that would align with your use-case though.

kabirkhan:
@honnibal totally agree on not setting custom attributes, didn't see a great way to write this without doing that.

@svlandeg I think adding in a knowledge base and setting the ent.kb_id_ would probably work well. I'll work on that change.

kabirkhan:
@svlandeg What's the plan for handling Aliases moving forward? 
Will the exact string need to be present in the kb aliases to be identified?
Should we allow for ignoring case?
Or willl we allow for fuzzy matching/search capabilities for possible aliases based on your entity text


For instance if:
text = "nyc is a city" identifies nyc as a LOC entity it would not get recognized as NYC in a knowledge base unless the knowledge base had "nyc" and "NYC" both pointing to the entity "New York City".

From the user point of view this is easy to handle in the simple case of lowercase aliases by just running
kb.add_alias(alias, ...)
kb.add_alias(alias.lower(), ...)

but that could get out of hand quickly with more complex patterns.

svlandeg:
@kabirkhan : there's pro's and con's for case-invariant matching. Personally, I would like to preserve the original casing in the KB when entering aliases because the prior probabilities can be quite different. For instance: The mention _Plato_ has 97% prior probability to be linked to [`Q859`](https://www.wikidata.org/wiki/Q859), but _PLATO_ only 4% because this spelling is much more likely to refer to [`Q1755406`](https://www.wikidata.org/wiki/Q1755406).

The candidate generation step of the algorithm will have to deal with generating appropriate candidates with lexical variants etc.

kabirkhan:
@svlandeg That makes total sense, thanks.
Knowing that, it doesn't make sense to use a Knowledge Base since patterns in EntityRuler could be weird and it'd be impossible to infer the aliases that need to be added.

@honnibal I changed this PR to use the built in ent_id/ent_id_ token attribute which I'm a lot happier with.

kabirkhan:
@honnibal @svlandeg @ines Just checking on the status of review. I do think this change is valuable by itself and doesn't impact the current use/functionality of EntityRuler so it seems safe to me.

honnibal:
@kabirkhan Sorry for the lack of attention to this. I've been falling behind a bit the last week or two.

It's probably obvious if I read more carefully, but could you help me out for a second and explain a bit about the `|` separator parsing? I couldn't immediately see an example of this in the tests.

Is there any way to avoid having to do this string decomposition, e.g. by adding keyword attributes onto the relevant methods? Whenever there are composite fields like this, we end up with potential bugs from the code which parses or composes the formatted strings. It looks like there might be such a bug in your code: if we do need the `|` label parsing, when you do `string.split('|')`, make sure to do `string.rsplit('|', 1)`. This way the code will still work if there's a `|` in the text somewhere.

kabirkhan:
@honnibal My bad I thought I already responded to this.

No worries, thanks for getting back to me.
The idea behind the `|` separator is to provide a unique key for each entity to the Matchers but to hide that implementation from the user. 

For example:

```python
text = 'Apple is a company in Cupertino'
patterns = [{'label': 'TECH_ORG', pattern: 'Apple', 'id': 'a1'}]
```
will create a PhraseMatcher with a unique key `TECH_ORG|a1` for the entity Apple.

However when the user adds this to the `EntityRuler` and gets `doc.ents` they get "Apple", `TECH_ORG` with the ent_id_ set to 'a1'

I considered doing something in the on_match function for each Matcher or editing the Matcher API directly to set the ent_id_ attribute but neither of those are trivial. The easiest thing seems to be to create a unique match_id using the user provided id and the the label together.

Let me know if you have a better idea of how to do this but to me, this seems like the best implementation.

honnibal:
@kabirkhan Okay, I see how that makes sense. Merging, thanks again!

PR #3599, opened 2019-04-16
entity types for colors should be in uppercase

fizban99:
although the text indicates the entity types should be in lowercase, the sample code shows uppercase, which is the correct format.

## Checklist


ines:
Thanks a lot! 👍

PR #3630, opened 2019-04-24
Eb 2.0.18

ksureshkumar-onalytica:
## Description
Changing version to old version used in nlp-api


ksureshkumar-onalytica:
This was an accident

PR #3646, opened 2019-04-26
Fix inconsistant lemmatizer issue #3484

### LINKED ISSUE #3484: Inconsistent lemmatization across different python sessions

The following will either produce the output `['dose']` or `['dos']`.  The output will be the same in multiple runs during the same python session, but if you start a new python session you may instead see the other of the two (which is then consistent within that session).

```python
import spacy
nlp = spacy.load("en")

print([tok.lemma_ for tok in nlp('doses')])
```

After some poking around, I think I've traced the issue to [this line](https://github.com/explosion/spaCy/blob/9e14b2b69fa1e8a065eb451a06141a1dcdb0d902/spacy/lemmatizer.py#L122) the function `lemmatize` defined in `spacy/lemmatizer.py`.  Based on the comment there, it seems like the assumption is being made that calling `list()` with an argument of `set` type will produce a sorted list.  As far as I understand, this is not true (at least in Python 3.6.4) and instead the returned list has an arbitrary order.

Further up the call stack, the list reaches [this line](https://github.com/explosion/spaCy/blob/146dc2766a069c8fe9e91c801f9695c2e262f742/spacy/morphology.pyx#L180) in `Morphology.lemmatize`, which selects the first element of the returned list.  This seems sufficient to explain the behavior.

## Your Environment

* **spaCy version:** 2.1.3
* **Platform:** Darwin-18.2.0-x86_64-i386-64bit
* **Python version:** 3.6.4
* **Models:** en

bjascob:
This closes #3484 where the lemma returned for some words is different (randomly) between different python sessions.

## Description
The issue is that when the lemmatization rules are applied (in lemmatizer.py::lemmatize) a `set` is used to create a unique list of forms.  Only the first on the list is returned to the user but since `set` has no odering the returned lemma is random.

The fix is to use an OrderedDict instead of a set to preserve ordering.  This was suggested by @BramVanroy.

Note that the bug only occurs across python sessions and I was not able to reproduce it inside of the same interpreter instance.  Because of this, the added test case (regression/test_issue3484.py) can not run automatically with pytest.  It needs to be run as a simple script, multiple times, from the command line.

### Types of change
This is a very simple bug fix, only modifying 2 lines of code and has no documentation changes.

## Checklist


honnibal:
I think the bug fix is good, I definitely appreciate it.

We couldn't merge the test as it is though -- we really need all the tests in the regression suite to work the same way, and to be triggered by pytest.

Perhaps we could make the test use multiprocessing, or just call it from `subprocess`? We could also just not have the test...But as it is the testing file is problematic, just because it doesn't do what one would expect.

ines:
> Perhaps we could make the test use multiprocessing, or just call it from subprocess? We could also just not have the test...

Another problem with this particular test is that it loads a model and the test suite included with the library shouldn't rely on the models, because it'd mean that the models would have to be installed and loaded for each build. And most of the time, there's actually no need to load a model and the same behaviour can be tested by creating a `Doc` with the given tags manually, or by calling into the internals directly (e.g. the `Lemmatizer`). Tests that specifically verify a model's prediction all live in [`spacy-models`](https://github.com/explosion/spacy-models).

I'd be fine with leaving this test out for now. Alternatively, we could also try and rewrite it to call into `Lemmatizer.lemmatize` directly and use `subprocess` (although this sounds pretty hacky).

bjascob:
Given that it's a one-line change and the note above it already says "and preserve order", my preference would be to remove the test,   However I'm also willing to spend some time with subprocess (multiprocessing won't do this under Linux) and see if I can come up with something.

Let me know, your preference.  If you're willing to live without the test let me know if you want me to remove it or if you can just reject that file when you merge it in.

ines:
@bjascob Yep, let's just remove the test for now!

bjascob:
I removed the test.

PR #3652, opened 2019-04-26
Add resume logic to spacy pretrain

tokestermw:
## Description

Adds resume logic to `spacy pretrain` after some discussion here: https://github.com/explosion/spaCy/pull/3510#issuecomment-478334698

### Test

Save this file to `sample_sents.jsonl`

```
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
{"text": "hello there."}
```

Then run `spacy pretrain`.

```bash
spacy pretrain sample_sents.jsonl en_core_web_md here -nw 1 -bs 1 -i 10 -uv -se 2
```

There will be some files in the `here` directory. You can load one of the pre-trained weights by:

```bash
spacy pretrain sample_sents.jsonl en_core_web_md here -nw 1 -bs 1 -i 10 -uv -se 2 -t2v ./here/model2.bin
```

🛑 Warning: Currently, `_load_pretrained_tok2vec` is incompatible in some cases betweeen the `-uv` option and `vectors_model` arguments / options (https://github.com/explosion/spaCy/pull/3510#issuecomment-483733166).

| Pretrain | Load pretrained Tok2Vec | Result | 
| --- | --- | --- |
| None | `en_core_web_md` | No `vectors` error |
| None | `en_core_web_sm` | ✅ |
| `-uv` | `en_core_web_md` | ✅ |
| `-uv` | `en_core_web_sm` | Broadcast shape error |

### Types of change

This is a new option to `spacy pretrain` called `--init-tok2vec`. It works just like `--init-tok2vec` in `spacy train`.

## Checklist

### DIFF
--- a/spacy/cli/pretrain.py
+++ b/spacy/cli/pretrain.py
@@ -18,6 +18,7 @@ from ..attrs import ID, HEAD
 from .._ml import Tok2Vec, flatten, chain, create_default_optimizer
 from .._ml import masked_language_model
 from .. import util
+from .train import _load_pretrained_tok2vec
 
 
 @plac.annotations(
@@ -36,6 +37,12 @@ from .. import util
     seed=("Seed for random number generators", "option", "s", float),
     n_iter=("Number of iterations to pretrain", "option", "i", int),
     n_save_every=("Save model every X batches.", "option", "se", int),
+    init_tok2vec=(
+        "Path to pretrained weights for the token-to-vector parts of the models. See 'spacy pretrain'. Experimental.",
+        "option",
+        "t2v",
+        Path,
+    ),
 )
 def pretrain(
     texts_loc,
@@ -53,6 +60,7 @@ def pretrain(
     min_length=5,
     seed=0,
     n_save_every=None,
+    init_tok2vec=None,
 ):
     """
     Pre-train the 'token-to-vector' (tok2vec) layer of pipeline components,
@@ -70,6 +78,9 @@ def pretrain(
     errors around this need some improvement.
     """
     config = dict(locals())
+    for key in config:
+        if isinstance(config[key], Path):
+            config[key] = str(config[key])
     msg = Printer()
     util.fix_random_seed(seed)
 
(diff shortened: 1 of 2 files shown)


spacy/cli/pretrain.py
```diff
@@ -18,6 +18,7 @@
 from .._ml import Tok2Vec, flatten, chain, create_default_optimizer
 from .._ml import masked_language_model
 from .. import util
+from .train import _load_pretrained_tok2vec
```
tokestermw: Move to `util.py`?

honnibal: Probably should, yes. Not strictly necessary for this PR.

honnibal:
It's a neat implementation, but I wonder whether the argument name is a little confusing. I think people would find it easier if it were called something to do with resume, rather than the current naming, which makes people reach the conclusion that this is how they could resume?

tokestermw:
Right, I guess it's not exactly resume as it's not keeping track of epochs, eval metrics, etc.

I can still rename to `--resume-path` or something like that while keeping the other code as-is.

honnibal:
@tokestermw Merging to get this out of PR purgatory. We'll just add the rename later, possibly when we're doing the release notes. Thanks again for your help!

robertsonwang:
👋 There's been a breaking change in `train.py` for this functionality. Specifically, `_load_pretrain_tok2vec` now has a `base_components` argument. This argument is used to check that a pretrained base component is not combined with a pretrained tok2vec layer. I got around the error by setting `base_components=""` but the proper solution would probably be to pass a `base_components` argument somewhere in the `pretrain` call.

PR #3657, opened 2019-04-29
Remove dangling M

yaph:
I assume this is a typo. Sorry if it has a meaning that I'm not aware of.


ines:
Thanks! And yes, that's definitely a typo – no idea how this happened!

PR #3661, opened 2019-04-30
update response after calling add_pipe

henry860916:
component:print_info is appened in the last, so need show it at the end of  pipeline

Types of change
a change to the documentation

## Checklist

- [ X] I ran the tests, and all new and existing tests passed.
- [] My changes don't require a change to the documentation, or if they do, I've added all required information.


ines:
Good point, thanks! 👍

PR #3662, opened 2019-04-30
Update French example sents and add two German stop words

BreakBB:
## Description
I updated the French example sentences as suggested in [this comment](https://github.com/explosion/spaCy/issues/1107#issuecomment-352291273) a while back. And added the two German stop words "anderem" and "ihren" which where missing but similar to the existing ones.

### Types of change
Minor enhancement

## Checklist


ines:
Thanks! 👍

PR #3685, opened 2019-05-06
Request to include Holmes in spaCy Universe

richardpaulhudson:
## Description

Dear spaCy team, I would be grateful if you would consider my Python library Holmes for inclusion in the spaCy Universe. Holmes transforms the syntactic structures delivered by spaCy into semantic structures that, together with various other techniques including ontological matching and word embeddings, serve as the basis for information extraction. Holmes supports several use cases including chatbot, structured search, topic matching and supervised document classification. I had the basic idea for Holmes around 15 years ago and now spaCy has made it possible to build an implementation that is stable and fast enough to actually be of use - thank you! At present Holmes supports English and German (I am based in Munich) but could easily be extended to support any other language with a spaCy model.

## Checklist

### DIFF
--- a/.github/contributors/richardpaulhudson.md
+++ b/.github/contributors/richardpaulhudson.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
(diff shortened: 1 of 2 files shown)


honnibal:
Wow, Holmes looks *amazing*!

I worked on semantic extraction stuff for a while in a post-doctoral project, and actually a lot of it was never published, because I was trying to do this fancy joint model with a linguistically motivated grammar, and it never quite worked. I always wanted to have this sort of pipelined approach for spaCy, as I think it's more practical.

I look forward to digging into the details of this more in future, and giving it a try. I'm sure the named entity linking work that's currently under development will be helpful for this too.

Btw, I assume the choice of the GPL license was made consciously? I'm sure you'll have some folks hoping that it could be switched to MIT.

richardpaulhudson:
Thanks – I will look forward to your feedback! I was not aware of the named entity linking development and it will certainly be a great improvement to include disambiguated named entities in matches.

Neither Holmes nor NeuralGym (which was added to universe.json last week) is currently visible on the Universe web page. Is this what you would expect (is there a release cycle) or has something gone wrong with the tooling?

I’m aware that the GPL is not the most user-friendly of licenses, but unfortunately - owing to a combination of irritating legal issues – it’s the best I can do – see https://github.com/msg-systems/holmes-extractor/#a-note-on-the-license.

PR #3693, opened 2019-05-07
fix ThaiTokenizer bug

wannaphongcom:
fix word tokenize for PyThaiNLP 2.0+

## Description

Update api for PyThaiNLP 2.0+

### Types of change

bug fix

## Checklist


ines:
Thanks! 👍

Btw, we should probably add pythainlp to the `extras_require`, just like we're currently doing it for Japanese:

https://github.com/explosion/spaCy/blob/2663f4133c28a7d294e35d8fe2215ffb9b5a7dae/setup.py#L248-L249

This would let people install `spacy[th]`, and would also give us a place to define version compatibility.

PR #3711, opened 2019-05-08
Polish lemmatizer

Gizzio:
## Description
This PR introduces rule-based lemmatizer for polish language. We obtained the rules from polish dictionary "ispell", converted and divided them into different UD POS tags. We had several issues with the speed of existing lemmatization algorithm because of the number of lemmatization rules for the polish language. We decided to write a new algorithm based on tries and generated all possible matches for regexes using sre_yield library. Results are satisfying, so we are hoping it can be merged to spaCy.
This lemmatizer requires POS tags to work, we trained POS taggers for polish language on 3 different tagsets, and when they are good enough, we will publish them along with parser and NER.
We also add auxiliary tagmap with mapping from polish specific POS to UD POS.

## Checklist


spacy/tests/conftest.py
```diff
@@ -152,6 +152,10 @@ def pt_tokenizer():
 @pytest.fixture(scope="session")
 def ro_tokenizer():
     return get_lang_class("ro").Defaults.create_tokenizer()
+    
+@pytest.fixture(scope='session')
+def pl_tokenizer():
+    return util.get_lang_class('pl').Defaults.create_tokenizer()
```
BreakBB: This should be `return get_lang_class('pl').Defaults.create_tokenizer()`

kowaalczyk:
@honnibal @ines could you please let us know what is the status of this PR?

In case you need more information, the details of our project (including published models) can be found here: http://spacypl.sigmoidal.io/

If everything is fine, how soon can we expect it to be merged? 
If not, please let us know what else do you need and we'll be happy to help :)

honnibal:
@kowaalczyk Sorry for the delay on this, we've had some indecision about it.

We're trying to cut down the package size, so I'm worried about the number of lines in the change. We've been merging these lemmatization exception lists, and I think it's really not a viable solution. It's sort of arbitrary to call the cut-off point here, but it has to be somewhere.

Are you able to put the lemmatization data in your models?

Alternatively, we'd like to switch the package data into json files within the repo, so that we can convert it into msgpack files when we package the sdist and upload it to PyPi. This will greatly reduce the on-disk footprint of spaCy, and also improve the loading times. As an alternative to putting the lemmatization data into your models, perhaps we could make Polish the first language we convert to this new approach?

Here's how the steps would look:

1. Change the data to .jsonl files, instead of the current .py files
2. Add a util function where you pass it a resource path, and it tries to load first the `.msg` file, and then the `.jsonl` file, depending on file availability.
3. In the `spacy.lang.pl` module, load the resources from the `.jsonl` or `.msg` files. This can be done in the top-level of the module, to give the same experience we currently have of the variables being there defined within the `.py` files.
4. In the `setup.py`, convert the `.jsonl` files to `.msg` files.
5. Update the `MANIFEST.in` declaration, so that we ship the `.msg` files within the package, and not the `.jsonl` files.

dzieciou:
I was looking for a stemmer or lemmatizer for the Polish language with Python API and I could not find anything satisfactory.  Hence, I have decided to [port Stempel stemmer to Python](https://github.com/dzieciou/pystempel/tree/feature/1). It's very fast implementation written in Java, from Apache Lucene.

Do I understand correctly that in SpaCy there's no place for stemmers, because SpaCy model follows a different flow: POS tagging, lemmatization and and POS-based lemma disambiguation?

honnibal:
We now have a new format for the language data, to prevent the overly-large installation size issue. See here: https://github.com/explosion/spaCy/pull/4167

When you have time, could you resubmit as json data? I'll close this in the meantime.

tadeoos:
Hey @kowaalczyk @Gizzio 
Are you guys working on the updated version? If so, do you have any ETA?

kowaalczyk:
Yes we are, this is our priority right now - we hope to have it finished in two weeks :)

dzieciou:
How is release of lemmatizer going on?

PR #3719, opened 2019-05-09
Fixing Regex Matcher Examples (#3708)

### LINKED ISSUE #3708: error in rule-based matching documentation
The following code (taken from a couple of different places [here](https://spacy.io/usage/rule-based-matching#entityruler)) produces no matches:

```python
import spacy

matcher = Matcher(nlp.vocab)
# Add match ID "HelloWorld" with no callback and one pattern
pattern = [{"TEXT": {"REGEX": "([Uu](\.?|nited) ?[Ss](\.?|tates))"}},
           {"LOWER": "president"}]
matcher.add("HelloWorld", None, pattern)

doc = nlp(u"This is about a United States president")
matches = matcher(doc)
for match_id, start, end in matches:
    string_id = nlp.vocab.strings[match_id]  # Get string representation
    span = doc[start:end]  # The matched span
    print(match_id, string_id, start, end, span.text)
```
It seems that the regex cannot deal with spaces - i.e. each element of the pattern list must match one individual token.

aaronkub:
## Description

Small change to the rule-based-matching documentation to resolve #3708.

### Types of change

Change to the documentation.

## Checklist

### DIFF
--- a/.github/contributors/aaronkub.md
+++ b/.github/contributors/aaronkub.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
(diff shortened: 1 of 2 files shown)


ines:
Thanks so much! 👍

Issue #3694, opened 2019-05-07
Make "text" key in JSONL format optional when "tokens" key is provided


devforfu:
## Feature description
I would like to build a pre-trained model using an already tokenized dataset. I store my data in JSONL format as described in the docs.
```
{"tokens": ["my", "tokenized", "data", "."]}
{"tokens": ["one", "more", "example", "."]}
...
```
As I understand, the `text` attribute is obligatory. (The CLI tool raises error trying to access `text` key). However, it seems that [these lines](https://github.com/explosion/spaCy/blob/master/spacy/cli/pretrain.py#L184-L188) use exactly one of these two keys:
```python
def make_docs(nlp, batch, min_length, max_length):
    docs = []
    for record in batch:
        text = record["text"]
        if "tokens" in record:
            doc = Doc(nlp.vocab, words=record["tokens"])  # use tokens
        else:
            doc = nlp.make_doc(text)  # use "raw" text
        ... # the rest of the code
```
Is it possible to make the `text` key optional if tokens are provided? It is not a big deal to extend my data with something like `{"text": null, "tokens": [...]}` but it seems a bit clumsy taking into account that this fragment can be easily refactored to account both cases without explicitly asking for a missing key.

Or am I missing something and these keys are used in some other places also?

BreakBB:
In the code the `text` key is used only optional, since it is just used in the `else` part. So moving `text = record["text"]` in the `else` part should solve your issue and would make pre-training require either `text` or `tokens` instead of `text` and optional `tokens`.

Moreover the `heads` key (see [these lines](https://github.com/explosion/spaCy/blob/a1e07f0d147e470cd4f0fd99bb62d28ddba7dd8d/spacy/cli/pretrain.py#L189)) isn't documented at all in the [CLI docs](https://spacy.io/api/cli#pretrain-jsonl).

devforfu:
Yes, that's what I was talking about. The `text` attributed is forced while it is not really required and can be accessed only in the second branch of the conditional logic. Also, a good point about `heads` thing.

ines:
Yes, good point – moving `text = record["text"]` into the `else` should be fine. If you want to submit a PR with this, that'd be great! 👍

PR #3721, opened 2019-05-10
Make "text" key in JSONL format optional when "tokens" key is provided

devforfu:
## Description

As it was discussed in the issue #3694, the PR modifies `pretrain` function a bit to make it possible skip `text` attribute in the JSONL records when `tokens` attribute is provided.

### Types of change

The only change is moving dictionary key access into the `else` branch of conditional logic. Also, the docs were adjusted a bit to reflect the new behavior. Though probably this update should be a bit more clear and verbose than my version. Please let me know if this should be re-written to become more explicit, or if something else should be added.

## Checklist

I didn't run the tests because it seems that the [`test_cli.py`](https://github.com/explosion/spaCy/blob/master/spacy/tests/test_cli.py) doesn't test the function that was modified, and it will not show any changes in behavior even if they were introduced. I can check that everything works as expected (including the website) but a bit later because setting up `spacy`'s environment takes some time.

### DIFF
--- a/website/docs/api/cli.md
+++ b/website/docs/api/cli.md
@@ -327,7 +327,7 @@ tokenization can be provided.
 
 | Key      | Type    | Description                                  |
 | -------- | ------- | -------------------------------------------- |
-| `text`   | unicode | The raw input text.                          |
+| `text`   | unicode | The raw input text. Is not required if `tokens` available. |
 | `tokens` | list    | Optional tokenization, one string per token. |
 
 ```json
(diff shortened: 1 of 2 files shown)


website/docs/api/cli.md
```diff
@@ -327,14 +327,15 @@ tokenization can be provided.
...
-| `text`   | unicode | The raw input text.                          |
+| `text`   | unicode | The raw input text. Is not required if `tokens` available. |
 | `tokens` | list    | Optional tokenization, one string per token. |
 
 ```json
 ### Example
 {"text": "Can I ask where you work now and what you do, and if you enjoy it?"}
 {"text": "They may just pull out of the Seattle market completely, at least until they have autonomous vehicles."}
 {"text": "My cynical view on this is that it will never be free to the public. Reason: what would be the draw of joining the military? Right now their selling point is free Healthcare and Education. Ironically both are run horribly and most, that I've talked to, come out wishing they never went in."}
+{"tokens": ["If", "tokens", "are", "provided", "then", "we", "can", "skip", "the", "raw", "input", "text"]}
```
ines: Good example, thanks for adding that! 👍

ines:
Looks good, thanks for taking this on! 👍

PR #3790, opened 2019-05-29
Update tokenizer.md for construction example

demongolem:
Self contained example.  You should really say what nlp is so that the example will work as is

## Checklist

### DIFF
--- a/website/docs/api/tokenizer.md
+++ b/website/docs/api/tokenizer.md
@@ -15,11 +15,15 @@ Create a `Tokenizer`, to create `Doc` objects given unicode text.
 >
 > ```python
 > # Construction 1
+> import spacy
 > from spacy.tokenizer import Tokenizer
+> nlp = spacy.load("en_core_web_sm")
 > tokenizer = Tokenizer(nlp.vocab)
 >
 > # Construction 2
+> import spacy
 > from spacy.lang.en import English
+> nlp = spacy.load("en_core_web_sm")
 > tokenizer = English().Defaults.create_tokenizer(nlp)
 > ```
 
--- a/website/docs/api/tokenizer.md
+++ b/website/docs/api/tokenizer.md
@@ -15,16 +15,15 @@ Create a `Tokenizer`, to create `Doc` objects given unicode text.
 >
 > ```python
 > # Construction 1
-> import spacy
 > from spacy.tokenizer import Tokenizer
-> nlp = spacy.load("en_core_web_sm")
+> from spacy.lang.en import English
+> nlp = English()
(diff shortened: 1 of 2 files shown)


.github/CONTRIBUTOR_AGREEMENT.md
```diff
@@ -87,7 +87,7 @@ U.S. Federal law. Any choice of law rules will not apply.
 7. Please place an “x” on one of the applicable statement below. Please do NOT
 mark both statements:
 
-    * [ ] I am signing on behalf of myself as an individual and no other person
+    * [X] I am signing on behalf of myself as an individual and no other person
```
ines: It looks like you accidentally filled in the contributor agreement template, instead of copying it over to `.github/contributors/demongolem.md`. Could you update that?

website/docs/api/tokenizer.md
```diff
@@ -15,11 +15,15 @@ Create a `Tokenizer`, to create `Doc` objects given unicode text.
 >
 > ```python
 > # Construction 1
+> import spacy
 > from spacy.tokenizer import Tokenizer
+> nlp = spacy.load("en_core_web_sm")
```
ines: I probably wouldn't load a morel here. The API docs are really supposed to show the more abstract construction and API instead of very specific usage examples (like the [usage docs](https://spacy.io/usage/linguistic-features#tokenization)).

The main message here should be that the `Tokenizer` takes an instance of `Vocab`, and that `create_tokenizer` takes an instance of `Language`. So the most abstract case would be this one:

```python
from spacy.vocab import Vocab
tokenizer = Tokenizer(Vocab())
```

But since this is really far removed from actual usage, we made the example use a generic `nlp` object to stand for "whichever `nlp` object you've created and are using". But I agree that it's nicer if you can actually copy-paste the example. So I'd suggest using the `English` class instead:

```python

nlp = English()
tokenizer = Tokenizer(nlp.vocab)
```

website/docs/api/tokenizer.md
```diff
@@ -15,11 +15,15 @@ Create a `Tokenizer`, to create `Doc` objects given unicode text.
...
 > tokenizer = Tokenizer(nlp.vocab)
 >
 > # Construction 2
+> import spacy
 > from spacy.lang.en import English
+> nlp = spacy.load("en_core_web_sm")
```
ines: We already have the `English` class here, so the more consistent way of showing the creation would probably be something like this:

```python

nlp = English()
tokenizer = English().Defaults.create_tokenizer(nlp)
```

ines:
Made the small fixes, so this is ready to merge now!

PR #3797, opened 2019-05-31
Fix all references to BILUO annotation format

ramananbalakrishnan:
## Description
The current code base was a mix of BILUO and BILOU when referring to the popular NER annotation format. Was confusing especially when searching for the phrase in the documentation.

Since BILUO is more commonly used across the package, all references to BILOU have been replaced with BILUO.(_also made sense to put O (OUT) at the outside of the abbreviation)_

### Types of change
- Two edits to comments/documentation
- One edit to a function name

## Checklist


ines:
Thanks a lot, always love improvements to consistency 👍 And BILUO is definitely the spelling we prefer.

PR #3800, opened 2019-05-31
Incorrect Token attribute ent_iob_ description

nipunsadvilkar:
## Description

class `Token` attribute `ent_iob_` gives unicode value of IOB tags and not integers.
```python
import spacy
nlp = spacy.load('en')
doc = nlp(u"I work for Apple Inc. and live in San Fransisco")
for tok in doc:
    print(tok, tok.ent_iob, tok.ent_iob_)
```
```
I 2 O
work 2 O
for 2 O
Apple 3 B
Inc. 1 I
and 2 O
live 2 O
in 2 O
San 3 B
Fransisco 1 I
```
### Types of change

Change to documentation of `Token` class attribute `ent_iob_` description
## Checklist


ines:
Ahh, looks like this was a copy-paste mistake! Thanks for fixing 👍

Issue #3811, opened 2019-06-03
Pretrain CLI, different seeding not possible


intrafindBreno:
## How to reproduce the behaviour

Passing a parameter to `-s` or `--seed` exposes the error. E.g. `python -m spacy pretrain data/lm/de.train.jsonl de_core_news_md lm_models/de -s 42`.

The error is:
```
Traceback (most recent call last):
  File "mtrand.pyx", line 678, in mtrand.RandomState.seed
TypeError: 'float' object cannot be interpreted as an integer
```
The seed parameter should be typed as `int` here: https://github.com/explosion/spaCy/blob/e7033011299d0444bb08f115778b03aedfcb4a38/spacy/cli/pretrain.py#L36

## Info about spaCy
* **spaCy version:** 2.1.4
* **Platform:** Linux-4.9.0-9-amd64-x86_64-with-debian-9.9
* **Python version:** 3.7.3

PR #3815, opened 2019-06-03
Fix for #3811

intrafindBreno:
Corrected type of seed parameter.

## Description

Changed type of seed parameter for cli `pretrain` from `float` to `int`

### Types of change

Bug fix

## Checklist


honnibal:
Thanks! Not sure how that slipped in, but definitely appreciate the analysis and PR.

Issue #3056, opened 2018-12-16
🌍 Adding models for new languages master thread


ines:
This thread bundles discussion around **adding pre-trained models for new languages** (and improving the existing language data). A lot of information and discussion has been spread over various different issues (usually specific to the language), which made it more difficult to get an overview.

[See here](https://spacy.io/models) for the available pre-trained models, and [this page](https://spacy.io/usage/models#languages) for all languages currently available in spaCy. Languages marked as "alpha support" usually only include tokenization rules and various other rules and [language data](https://spacy.io/usage/adding-languages).

## How to go from alpha support to a pre-trained model

The process requires the following steps and components:

* **Language data**: shipped with spaCy, [see here](https://github.com/explosion/spaCy/tree/master/spacy/lang/). The tokenization should be reliable, and there should be a [tag map](https://spacy.io/usage/adding-languages#tag-map) that maps the tags used in the training data to coarse-grained tags like `NOUN` and optional morphological features.
* **Training corpus:** the model needs to be trained on a suitable corpus, e.g. an existing [Universal Dependencies treebank](https://github.com/UniversalDependencies). Commercial-friendly treebank licenses are always a plus. Data for tagging and parsing is usually easier to find than data for named entity recognition – in the long term, we want to do more data annotation ourselves using [Prodigy](https://prodi.gy), but that's obviously a much bigger project. In the meantime, we have to use other available resources (academic etc.).
* **Data conversion:** spaCy comes with a range of built-in converters via the `spacy convert` command that take `.conllu` files and output spaCy's JSON format. [See here](https://spacy.io/usage/training#spacy-train-cli) for an example of a training pipeline with data conversion. Corpora can have very subtle formatting differences, so it's important to check that they can be converted correctly.
* **Training pipeline:** if we have language data plus a suitable training corpus plus a conversion pipeline, we can run `spacy train` to train a new model.

With our new internal model training infrastructure, it's now much easier for us to integrate new pipelines and train new models. 

> ⚠️ **Important note:** In order to train and distribute "official" spaCy models, we need to be able to integrate and reproduce the full training pipeline whenever we release a new version of spaCy that requires new models (so we can't just upload a model trained by someone else).

## Ideas for how to get involved

Contributing to the models isn't always easy, because there are a lot of different things to consider, and a big part of it comes down to sourcing suitable data and running experiments. But here are a few ideas for things that can move us forward:

### 1️⃣ Difficulty: good for beginners

[...]

PR #3835, opened 2019-06-11
Update lex_attrs.py

Azagh3l:
Corrected typos, added french (from France) versions of some numbers.
#3056
#3836


ines:
Thanks a lot! 👍

PR #3843, opened 2019-06-13
Update error raising for CLI pretrain to fix #3840

### LINKED ISSUE #3840: pretrain with incorrect input file raises "cryptic" error

I ran into this issue with a JSONL file which lines are simple strings instead of real key-value JSON. Running `pretrain` with a file which includes neither the `text` nor the `tokens` attribute, e.g.

> "This is just some string with no attributes"

Such input leads to an error [right here](https://github.com/explosion/spaCy/blob/f35ce097762c3bfc7010cfe6ee8398c66d8223b0/spacy/cli/pretrain.py#L201) when `record["text"]` is called while `record` doesn't have the `text` key.

The error message is:

> TypeError: string indices must be integers

## Your Environment
* Python Version Used: Python 3.7.3
* spaCy Version Used: 2.1.4

## Possible solution

As far as I know a simple string is still valid JSON therefore [srsly](https://github.com/explosion/srsly) shouldn't raise an error [here](https://github.com/explosion/spaCy/blob/f35ce097762c3bfc7010cfe6ee8398c66d8223b0/spacy/cli/pretrain.py#L103) or [here](https://github.com/explosion/spaCy/blob/f35ce097762c3bfc7010cfe6ee8398c66d8223b0/spacy/cli/pretrain.py#L108).

Instead a simple check if the `text` key exists would prevent this and raising a good error message in a new `else` path could hint the user about the invalid input.

BreakBB:
## Description
This PR adds some checks for the input of the CLI pretrain command as discussed in #3840. Those are:

- empty input file
- input file contains something else than a JSON object (per line)
- neither the `text` nor the `tokens` key is provided

Moreover empty values for the keys are skipped, counted and logged as a warning after pretrain. The `make_docs` function is also tested now except for the undocumented `heads` key.

### Types of change
Enhancement

## Checklist

### DIFF
--- a/spacy/cli/pretrain.py
+++ b/spacy/cli/pretrain.py
@@ -152,16 +152,18 @@ def pretrain(
             with (output_dir / "log.jsonl").open("a") as file_:
                 file_.write(srsly.json_dumps(log) + "\n")
 
+    skip_counter = 0
     for epoch in range(n_iter):
         for batch_id, batch in enumerate(
             util.minibatch_by_words(((text, None) for text in texts), size=batch_size)
         ):
-            docs = make_docs(
+            docs, count = make_docs(
                 nlp,
                 [text for (text, _) in batch],
                 max_length=max_length,
                 min_length=min_length,
             )
+            skip_counter += count
             loss = make_update(
                 model, docs, optimizer, objective=loss_func, drop=dropout
             )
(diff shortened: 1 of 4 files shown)


spacy/cli/pretrain.py
```diff
@@ -195,20 +202,30 @@ def make_update(model, docs, optimizer, drop=0.0, objective="L2"):
...
+                skip_count += 1
             doc = nlp.make_doc(text)
+        else:
+            raise ValueError(Errors.E138.format(text=record))
         if "heads" in record:
             heads = record["heads"]
             heads = numpy.asarray(heads, dtype="uint64")
             heads = heads.reshape((len(doc), 1))
             doc = doc.from_array([HEAD], heads)
         if len(doc) >= min_length and len(doc) < max_length:
```
BreakBB: Should this case be added to `skip_count` as well?

honnibal:
Really nice! Thanks a lot for this.

PR #3855, opened 2019-06-17
Changed learning rate by its param name.

elbaulp:
I've been searching for a while how the parameter learning rate was named, with `beta1` and `beta2` its easy as they are marked as code, but learning rate wasn't. I think writing the actual parameter name would be helpful.

## Checklist


ines:
Thanks a lot! 👍

PR #3864, opened 2019-06-19
Entity linking using Wikipedia & Wikidata

svlandeg:
Implementation of the Entity Linker using Wikidata entities and Wikipedia training. Focus of this PR is on the general pipeline - further performance improvements can certainly be made.

**Note**: this PR temporarily reverts [this edit](https://github.com/explosion/spaCy/commit/a931d72459ab83c4583a995b8b433bd2470087f7) as it broke the parsing by `en_core_web_lg`.

### Added core functionality

- KB stores entity vectors for each entity 
- KB to/from bytes + unit test
- various improvements to KB code (nogil)
- implementation of Entity linking pipe, using entity encoder and context encoder
- Entity linker pipe to/from file (as part of writing nlp object)
- GoldParse has `links` property with (start_char, end_char, gold_kb_id) tuples

### Types of change
New feature

### General flow & implementation with Wikipedia & WikiData

- **create KB**: vocab + entities (freq, entity vector) + aliases (prior prob)
  - prior probabilities from intrawiki links on Wikipedia
  - total of 7.9M entities/titles filtered down to 1.1M by min. 20 incoming intrawiki links for an entity
  - trained entity encodings with simple encoder-decoder self-supervision to 64D vectors
  - alias defined by max. 10 candidates per entity and min. 5 occurrences of an alias-entity pair: 1.6M aliases
  - the result is a 345MB KB

- **obtain training data**: list of (`Doc`, `GoldParse`) tuples,  `GoldParse`.`links` contains (start_char, end_char, gold_entity) tuples
  - align intrawiki links with `en_core_web_lg` NER `ents`, store the sentence as local context
  - some custom filtering to keep memory usage down: articles below 30K chars and sentences between 5 and 100 tokens

- **run training** of the pipe similar to training NER or textcat:
  - the context encoder runs a `Doc` through a CNN and outputs 64D vectors that are compared to the entity vectors
  - Training loss is defined by calculating the partial derivative of cosine similarity between context and entity vectors (currently only pos examples)
  - Dev performance is measured as accuracy, comparing the gold IDs for the NER `ents` to their predicted entity IDs
```
other_pipes = [pipe for pipe in nlp.pipe_names if pipe != "entity_linker"] 
with nlp.disable_pipes(*other_pipes): 
    ...
    nlp.update(...)
   ...
```

### File overview

- in `examples/pipeline/` :
  - `wikidata_entity_linking.py` : shows the full pipeline for running the Entity Linking functionality

- in `/bin`:
  - `wikidata_processor.py`: Parse the JSON wiki data to get all entities (takes about 7h to parse 55M lines)
  - `wikipedia_processor.py`: Process the XML Wikipedia dump to calculate entity frequencies and prior probabilities (takes about 2h to parse 1100M lines)
  - `training_set_creator.py`: Process the XML Wikipedia dump to parse out raw text + gold-standard entities in offset format per article (takes about 15h to parse 60M articles on 1100M lines)
  - `train_descriptions.py`: Train the embeddings of entity descriptions to fit a fixed-size entity vector (e.g. 64D).
  - `kb_creator.py` : Pipeline to create the knowledge base from Wikidata entries

- in `tests/` :
  - `pipeline/test_entity_linker.py`: test correct & incorrect usage of the KB
  - `serialize/test_serialize_kb.py`: test correct writing & reading of KB

### First results

With 1.1M entities in the KB (14% of all), "oracle" KB accuracy is 84.2% on a dev set of 5K entities, and prior probability by itself gives 78.2% accuracy. Adding sentence context improves this marginally, but consistently. Training on 200K entities for 2h, the context encoder (by itself, but limited by the KB) gets to 73.9% accuracy and brings the combined accuracy with the prior probabilities to 79.0%. The "pick one at random" baseline on the other hand would only achieve around 54%, so the context encoder is doing relatively well on its own.

I experimented with more complex models than the ones in this PR, including e.g. an additional document encoder, but I figured that would bias the results too much towards the typical structure of Wikipedia documents with 2-3 introductory sentences. And preliminary tests with other models didn't improve the accuracy.

### Beyond this PR
There's multiple ways of further building on top of this result:

1.  bring in more structured data from wikidata (e.g. whether entities have a day of birth, and when)
2.  implement a compositional model for entity vectors
3.  refine the context encoder (currently simply the whole sentence and a CNN) 
4.  bring in document consistency either through coreference resolution or unification of the set of predictions per document
5.  implement a document encoder using the other entities in a document 
6.  set a threshold for NIL predictions. Currently not done because there are no gold "false" examples in the data, because WP annotation is incomplete
7.  learn (also) from incorrect examples by generating wrong candidates from the KB (preliminary results were not so encouraging)
8.  add in other information from the NLP pipeline such as NER type, POS, etc.
9.  further experiment with hyperparameters of KB construction & model implementation and/or other models
10.  k-best NER
11.  fuzzy matching for candidate generation

### Example output
```
for ent in doc.ents:
    print(ent.text, ent.label_, ent.kb_id_)
```

_In The Hitchhiker's Guide to the Galaxy, written by Douglas Adams, Douglas reminds us to always bring our towel, even in China or Brazil. The main character in Doug's novel is the man Arthur Dent, but Douglas doesn't write about George Washington or Homer Simpson._

>  The Hitchhiker's Guide WORK_OF_ART 
>  Douglas Adams PERSON Q42
>  Douglas PERSON Q156220
>  China GPE Q148
>  Brazil GPE Q155
>  Doug PERSON Q1251705
>  Arthur Dent PERSON Q613901
>  Douglas PERSON Q156220
>  George Washington PERSON Q23
>  Homer Simpson PERSON Q7810

## Checklist


svlandeg:
The prior probability baseline is a strong one to beat, partly because the KB is biased (during construction) towards more frequent entities/aliases. Additionally, there are also a lot of mistakes on Wikipedia where editors linked to the more frequent item because that is easier to do. During error analysis I came across "wrong" predictions which should have been right, like "France" in "Agnes Maria of Andechs-Merania (died 1201) was a Queen of France." being predicted as "kingdom in Western Europe from 987 to 1791" (Q70972) instead of "republic with mainland in Europe and numerous oversea territories" (Q142), but the latter was the "gold" annotation.

I've also noticed that the Date NER type confuses the model because the common years simply have the description "year" (https://www.wikidata.org/wiki/Q25274) so the context encoder tries to match something more fancy like "2019 Copa América" (https://www.wikidata.org/wiki/Q4630358)

honnibal:
Great work!

The results here really bring back my love/hate relationship with the entity linking task. It's a very important problem, but is so annoying to work on. The gap between the oracle and the baseline is typically small, as you've found, and the data is either too small, or automatically created.

I think it's important that we avoid getting hung up on improving the accuracy numbers on this data. You're completely right that optimising document context features on this training data are unlikely to give us a model that generalises well to other mentions, because we have such an unrepresentative (and often incorrect) set of mentions.

Now that we have the system in place, real improvements will be coming from getting better data, and especially a coreference tagged corpus so we can estimate `P(link | mention)` using the abbreviated mentions that come deeper in a coreference chain, which Wikipedia typically wouldn't link.

Once we have a hill worth climbing, we'll be able to revisit a lot of the questions you've set out as areas of improvement.

spacy/structs.pxd
```diff
@@ -72,3 +76,32 @@ cdef struct TokenC:
     attr_t ent_type # TODO: Is there a better way to do this? Multiple sources of truth..
     attr_t ent_kb_id
     hash_t ent_id
+
+
+# Internal struct, for storage and disambiguation of entities.
+cdef struct EntryC:
```
honnibal: If this will live in the `structs.pxd`, then unfortunately we need a less vague name than `EntryC`. `KBEntryC` is ugly, but I can't immediately do better.

spacy/pipeline/pipes.pyx
```diff
@@ -1063,51 +1064,205 @@ cdef class EntityRecognizer(Parser):
...
+            return loss
+        return 0
+
+    def get_loss(self, docs, golds, scores):
+        targets = [[1] for _  in golds]  # assuming we're only using positive examples
+        loss, gradients = self.get_cossim_loss_2(yh=scores, y=golds, t=targets)
+        loss = loss / len(golds)
+        return loss, gradients
+
+    def get_cossim_loss_2(self, yh, y, t):
```
honnibal: Is there a bug in the cosine loss function that `spacy pretrain` is using?

spacy/pipeline/pipes.pyx
```diff
@@ -1063,51 +1064,205 @@ cdef class EntityRecognizer(Parser):
...
+        util.from_disk(path, deserialize, exclude)
+        return self
+
+    def rehearse(self, docs, sgd=None, losses=None, **config):
         # TODO
         pass
 
     def add_label(self, label):
-        # TODO
         pass
```
honnibal:
```suggestion
        raise NotImplementedError
```

spacy/pipeline/pipes.pyx
```diff
@@ -1063,51 +1064,205 @@ cdef class EntityRecognizer(Parser):
...
+        deserialize = OrderedDict()
+        deserialize["cfg"] = lambda p: self.cfg.update(_load_cfg(p))
+        deserialize["model"] = load_model
+        exclude = util.get_serialization_exclude(deserialize, exclude, kwargs)
+        util.from_disk(path, deserialize, exclude)
+        return self
+
+    def rehearse(self, docs, sgd=None, losses=None, **config):
         # TODO
         pass
```
honnibal:
```suggestion
        raise NotImplementedError
```

spacy/pipeline/pipes.pyx
```diff
@@ -1063,51 +1064,205 @@ cdef class EntityRecognizer(Parser):
...
-                    best_candidate = max(candidates, key=lambda c: c.prior_prob)
-                    for token in ent:
-                        token.ent_kb_id_ = best_candidate.entity_
+            if len(doc) > 0:
+                context_encoding = self.model([doc])
+                context_enc_t = np.transpose(context_encoding)
+                for ent in doc.ents:
+                    candidates = self.kb.get_candidates(ent.text)
+                    if candidates:
+                        scores = list()
```
honnibal:
```suggestion
                        scores = []
```

spacy/pipeline/pipes.pyx
```diff
@@ -1063,51 +1064,205 @@ cdef class EntityRecognizer(Parser):
...
+                        for c in candidates:
+                            prior_prob = c.prior_prob * self.prior_weight
+                            kb_id = c.entity_
+                            entity_encoding = c.entity_vector
+                            sim = float(cosine(np.asarray([entity_encoding]), context_enc_t)) * self.context_weight
+                            score = prior_prob + sim - (prior_prob*sim)
+                            scores.append(score)
+
```
honnibal: Is this loop a performance problem? There are a couple of things we could do if so. Basically we can assemble arrays and then do the operations vectorised. I often find this harder to read then a good honest loop though, so I'd only want to make the change if there's actually a problem.

spacy/pipeline/pipes.pyx
```diff
@@ -1063,51 +1064,205 @@ cdef class EntityRecognizer(Parser):
...
+                        # TODO: thresholding
+                        best_index = scores.index(max(scores))
```
honnibal: We should probably make `scores` an array above, so we can use `scores.argmax()`

spacy/tests/serialize/test_serialize_kb.py
```diff
@@ -0,0 +1,73 @@
...
+    kb.add_entity(entity=u'Q53', prob=0.33, entity_vector=[0, 5, 3])
+    kb.add_entity(entity=u'Q17', prob=0.2, entity_vector=[7, 1, 0])
+    kb.add_entity(entity=u'Q007', prob=0.7, entity_vector=[0, 0, 7])
+    kb.add_entity(entity=u'Q44', prob=0.4, entity_vector=[4, 4, 4])
+
+    kb.add_alias(alias=u'double07', entities=[u'Q17', u'Q007'], probabilities=[0.1, 0.9])
+    kb.add_alias(alias=u'guy', entities=[u'Q53', u'Q007', u'Q17', u'Q44'], probabilities=[0.3, 0.3, 0.2, 0.1])
+    kb.add_alias(alias=u'random', entities=[u'Q007'], probabilities=[1.0])
+
+    return kb
```
honnibal: Great test here! I often struggle to build these small test cases.

spacy/tests/serialize/test_serialize_kb.py
```diff
@@ -0,0 +1,73 @@
...
+    kb1 = _get_dummy_kb(en_vocab)
+    _check_kb(kb1)
+
+    # dumping to file & loading back in
+    with make_tempdir() as d:
+        dir_path = ensure_path(d)
+        if not dir_path.exists():
+            dir_path.mkdir()
+        file_path = dir_path / "kb"
+        print(file_path, type(file_path))
```
honnibal:
```suggestion

```

honnibal:
Hm, I think both the `predict()` and `update()` methods will fail if the GPU is enabled, currently :(. There's some places where I think the code assumes all of the data is on CPU, when I think some of it could be on GPU? The model also could be GPU-allocated.

While fixing the GPU handling, we also need to be careful that we don't end up indexing into a GPU-allocated array during a loop. The `Tagger` class has an example of the dance of allocating stuff to CPU to calculate the gradient of the loss, then copying the gradients back to the GPU to do the backprop.

If it's difficult for you to test on GPU, we could merge this despite the problem, and I'm test and make the fix.

spacy/tests/serialize/test_serialize_kb.py
ines: We can totally leave out the `u` strings in our own code btw if the files all have the correct declarations:

```python
# coding: utf-8
```

spacy/language.py
```diff
@@ -809,6 +810,14 @@ def from_disk(self, path, exclude=tuple(), disable=None):
             # Convert to list here in case exclude is (default) tuple
             exclude = list(exclude) + ["vocab"]
         util.from_disk(path, deserializers, exclude)
+
+        # download the KB for the entity linking component - requires the vocab
```
ines: This code feels a little arbitrary and lost here? Maybe we could add it as a deserializer above? Since it's a few lines of code, maybe the `kb` module could expose a deserializer helper function that does all of this and that follows the same format as the other deserializers?

I think it's cleaner to do it all in `util.from_disk`, and it'd also handle the exclusion etc. automatically, so we don't have to reimplement all of this here.

spacy/structs.pxd
svlandeg: I don't mind either way - I can also move it back to kb.pxd?

spacy/pipeline/pipes.pyx
svlandeg: No, not that I know off :-) I was having this function here to experiment with the loss function for negative examples, but I ended up removing that again. I double checked, and the current code is giving the exact same output as your original function so I'll remove this again (should have done that before submitting)

svlandeg:
Thanks for the detailed reviews! I'm working on addressing them. Additionally, I just found that `Span.as_doc()` is not preserving the entity links because that's not properly defined as an attribute yet.
I guess I need to add an `ENT_KB_ID` or something such to `attrs.pyx` that links to `Token.ent_kb_id` ?

spacy/pipeline/pipes.pyx
svlandeg: Good point @honnibal. I looked into this, testing on a batch of 10K docs, the original code took about 20s. The biggest issue was actually running each doc through the model instead of just running `self.model(docs)` once (little silly of me). This shaves off 7s, and this call to the model then takes up the most time, with about 5s out of 13s left to the `candidates` loop. I tried rewriting it using arrays (cf. latest commit), then it goes to about 3s orso. I can run a test with bigger datasets, but for now this seems the best I can do - let me know if you have further comments/ideas !

spacy/structs.pxd
svlandeg: Renamed to `KBEntryC` for now

spacy/language.py
svlandeg: Yea, I agree the current solution is not ideal ;-) Not sure yet how to rewrite though, the main problem being that the KB needs the vocab, which is only available after running `util.from_disk`, right?

svlandeg: Yes, for sure. I turned to this ugliness because I somehow missed that the `deserializers` are in an `OrderedDict` so the `vocab` is available in time for the `kb` construction. Should be fixed in the latest commits.

svlandeg:
Update @ines @honnibal : I've addressed most comments - feel free to review again. It still needs to be adjusted & tested on GPU though. I'm happy to look into that, though it will take me a bit longer than when you do Matt, but I don't mind either way.

The main thing that changed since the publication of this PR is that originally, the model was looking at cosine similarity between the context (sentence) and entity (description) encoders, but now there's an additional EL model that takes the input of the context & entity encoders + additional features such as prior probability and NER types, and runs a net on top of that. This is a more generic setup and will allow for more flexibility, but it's a little awkward how that currently results in two models being trained in the `EntityLinker` pipe: the context encoder (`Tok2Vec`) and the actual EL model. I wasn't sure how to define the optimizer for the context encoder, so right now it's a field `sgd_context` in `EntityLinker`. Ideally I would have wanted to `chain` the context linker into the EL model, making the backprop easier too, but I'm not sure how to do that in this case because it needs to be concated first with an encoder that uses a different input.

ines:
Okay, I'll go ahead and merge this 🎉 It'll still be inofficial and undocumented, but it'd allow people who want to test it to do so. We can then keep adding to it until it's ready for "official" release 🙂

almoslmi:
I tried to implement the NEL on my computer. After long processing time of Wekidata, I got this error: 
"NameError: name 'EntityEncoder' is not defined"
Does anybody know how to solve this issue, please?

================
Here is the last lines of the log:

2019-07-19 19:28:05.160215 processed 57500000 lines of WikiData dump

 * _get_entity_frequencies 2019-07-19 19:30:03.913394

Kept 1908129 out of 7988394 titles with filter frequency 20

 * train entity encoder 2019-07-19 19:30:31.422478

  File "C:/Users/tareq/PycharmProjects/spaCy/examples/pipeline/wikidata_entity_linking.py", line 441, in <module>
    run_pipeline()
  File "C:/Users/tareq/PycharmProjects/spaCy/examples/pipeline/wikidata_entity_linking.py", line 105, in run_pipeline
    wikidata_input=WIKIDATA_JSON)
  File "C:\Users\tareq\PycharmProjects\spaCy\bin\wiki_entity_linking\kb_creator.py", line 64, in create_kb
    encoder = EntityEncoder(nlp, INPUT_DIM, DESC_WIDTH)
NameError: name 'EntityEncoder' is not defined

almoslmi:
I have run NEL and after it processed both Wikipedia and Wikidata dumps I got an error related to Cython. All cimport in all .pyx and .pxd files don't work although the necessary libs are installed and added to the environment too. I am getting a SyntaxError: invalid syntax with all of them. Any idea how to solve this issue, please?

I am using Pycharm, and Here is the log when I tried to run KB.pyx file

C:\Users\tareq\Anaconda3\python.exe "C:\Program Files\JetBrains\PyCharm 2019.1.3\helpers\pydev\pydevconsole.py" --mode=client --port=52781
import sys; print('Python %s on %s' % (sys.version, sys.platform))
sys.path.extend(['C:\\Users\\tareq\\PycharmProjects\\spaCy', 'C:\\Users\\tareq\\PycharmProjects\\spaCy\\spacy', 'C:/Users/tareq/PycharmProjects/spaCy'])
Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.6.1 -- An enhanced Interactive Python. Type '?' for help.
PyDev console: using IPython 7.6.1
Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)] on win32
runfile('C:/Users/tareq/PycharmProjects/spaCy/spacy/kb.pyx', wdir='C:/Users/tareq/PycharmProjects/spaCy/spacy')
  File "C:\Users\tareq\Anaconda3\lib\site-packages\IPython\core\interactiveshell.py", line 3325, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-819cb021c446>", line 1, in <module>
    runfile('C:/Users/tareq/PycharmProjects/spaCy/spacy/kb.pyx', wdir='C:/Users/tareq/PycharmProjects/spaCy/spacy')
  File "C:\Program Files\JetBrains\PyCharm 2019.1.3\helpers\pydev\_pydev_bundle\pydev_umd.py", line 197, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "C:\Program Files\JetBrains\PyCharm 2019.1.3\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "C:/Users/tareq/PycharmProjects/spaCy/spacy/kb.pyx", line 7
    from cymem.cymem cimport Pool
                                                 ^
SyntaxError: invalid syntax

svlandeg:
Hi @almoslmi , issues, bugs and problems are best reported on the [issue tracker](https://github.com/explosion/spaCy/issues) instead of as comments on a merged pull request ;-)
As a quick response to your message, `kb.pyx` shouldn't be run as such, but you can try  running [wikidata_entity_linking.py](https://github.com/explosion/spaCy/blob/master/examples/pipeline/wikidata_entity_linking.py). All of this is currently undocumented though.

almoslmi:
Thanks @svlandeg for your reply. I have deleted it from here and added it to the issue tracker. By the way, I run wikidata_entity_linking.py and was working well after fixing some issues, but when it successfully finished processing the two dumps of Wikipedia and Wikidata, it failed to write the KB. I have explained that in the issue tracker, and you may have a look at it, please.
https://github.com/explosion/spaCy/issues/4000

zainbnv:
I have successfully trained a spacy entity linking model(obviously by limiting the data).

my question is how to display the description of entity from kb as output?

import spacy
nlp = spacy.load(r"D:\el model\nlp")
doc = nlp("Amir Khan is a great boxer")
ents = [(e.text, e.label_, e.kb_id_) for e in doc.ents]
print(ents)

svlandeg:
Hi @zainbnv. The descriptions are currently not stored in the KB itself because of performance reasons. However, from the intermediary results during processing, you should have a file `entity_descriptions.csv` that maps the WikiData ID to its description in a simple tabular format.

PR #3895, opened 2019-07-02
Lithuanian language support

rokasramas:
## Description
Added Lithuanian language support & tests.

### Types of change
- Lemmatizer
- Lex Attrs
- Morph rules
- Stop words
- Tag map (based on [UD Lithuanian ALKSNIS](https://universaldependencies.org/treebanks/lt_alksnis/index.html))
- Tokenizer exceptions

## Checklist


spacy/lang/lt/__init__.py
```diff
@@ -1,20 +1,40 @@
...
+    tokenizer_exceptions = update_exc(BASE_EXCEPTIONS, TOKENIZER_EXCEPTIONS)
     stop_words = STOP_WORDS
+    tag_map = TAG_MAP
+    morph_rules = MORPH_RULES
+    lemma_lookup = LOOKUP
 
 
 class Lithuanian(Language):
-    lang = "lt"
+    lang = 'lt'
```
ines: I hate being nitpicky like this and it's totally not a big deal – but we now use [`black`](https://github.com/python/black) for auto-formatting, so our standard quotes are double-quotes. Looks like you have your editor configured differently, so if it's too annoying to change, let me know and I can just run the whole PR through Black to unify it.

ines:
Thanks so much, this is great! 🎉 Also thanks for adding tests.

zaibacu:
Also, on top of that, we have language model trained. How we should go forward providing it?

PR #3899, opened 2019-07-02
integrate SudachiTokenizer which is installed from github

hiroshi-matsuda-rit:
Please review our strategy for integrating SudachiPy and SudachiDict packages along with ja_core_web_md.tar.gz.

Current version requires non-automated pip installation for two packages below.

pip install -e git+git://github.com/WorksApplications/SudachiPy@develop#egg=SudachiPy
pip install "https://github.com/megagonlabs/UD_Japanese-PUD/releases/download/v2.4-NE-spacy/ja_core_web_md-1.4.0.tar.gz"

The former SudachiPy installation will be merged to the latter ja_core_web_md install step when SudachiPy would be registered to PyPI.

Thanks,
Hiroshi Matasuda @ Megagon Labs, Tokyo.

## Checklist


hiroshi-matsuda-rit:
I filled in SCA fields and pushed that file to this branch.

hiroshi-matsuda-rit:
The model file is trained on UD_Japanese-PUD v2.4-NE that is a fork of official PUD and contains additional OntoNotes 5 based gold NE labels with BI-CATEGORY format.
https://github.com/megagonlabs/UD_Japanese-PUD/blob/v2.4-NE-spacy/ja_pud-ud-test.train.json
https://github.com/megagonlabs/UD_Japanese-PUD/blob/v2.4-NE-spacy/ja_pud-ud-test.test.json

These jsons are created by my conllu-to-json converter which I implemented to handle my custom NE-BI labels and also to apply Japanese-specific Zenkaku/Hankaku character data-augmentation method.
https://github.com/megagonlabs/ginza/blob/develop/ginza_util/conllu_to_json.py

I used these json files for the arguments of spacy train command. You can view all the arguments of spacy train in the training shell script below.
https://github.com/megagonlabs/ginza/blob/develop/shell/train_pipeline.sh

hiroshi-matsuda-rit:
I release ja_core_web_md-1.4.1.tar.gz.
https://github.com/megagonlabs/UD_Japanese-PUD/releases/download/v2.4-NE-spacy/ja_core_web_md-1.4.1.tar.gz

From this version, I add a 'install_requires' entiry for SudachiPy.

By the way, how do I remove the testing errors?
I'll add some required packages for tests.

hiroshi-matsuda-rit:
I reverted LANGUAGES list in soacy/test/lang/test_initialize.py.
Still I'm considering the way to install some packages only for testing environment.
Actually, my language model uses SudachiPy and its dictionary packages but these guys are not so small hence I'd like to install them on demand.
Does anyone know good methodology for that?

hiroshi-matsuda-rit:
I added some requirement entries for setup.py but not reflected to the testing pipelines.
I'd try to configure test environment but it might require certain permissions which I do not have.

hiroshi-matsuda-rit:
I found that the testing environment only refers requirements.txt.
I added the line which specifies SudachiPy tar-ball URL and found that it's working on recent python versions.
I'm trying to solve setuptools compatibility problem around SudachiPy.

hiroshi-matsuda-rit:
I've resolved all the test blockers related to SudachiPy installation.
I set 'skipif' conditions for all the test cases related to SudachiPy because it supports only 3.5+.

I'm going to send a PR to merge ja_core_web_md-2.1.0.json for spacy-models.
https://github.com/megagonlabs/UD_Japanese-PUD/releases/tag/v2.4-NE-spacy

In addition, I discussed with SudachiPy develop team to release pip install archive (including SudachiPy and its dictionary) from their own github repository, and we agreed for that and releasing archives ASAP.
https://github.com/WorksApplications/SudachiPy/pull/44

Please wait. Thanks,

hiroshi-matsuda-rit:
I sent a PR for spacy-models.
https://github.com/explosion/spacy-models/pull/16

hiroshi-matsuda-rit:
#3945 is a simplified version of this PR.
Please use better one.

Thanks,

erip:
Should this be closed? It seems like Sudachi has been abandoned for fugashi.

hiroshi-matsuda-rit:
@erip Thanks for mentioning that. I'd like to make sure that Fugashi is the best option for the spaCy's official Japanese mode.

@polm I think your Fugashi's approach, calling MeCab from python, is one of the fastest solutions to tokenize the Japanese text from python. But I have some objections.

The reasons why I selected the SudachiPy and its dictionaries are:
- Easy to install
- Well maintained and high performance codes
- Well tuned and maintained dictionaries

By Fugashi, it became much easier to install and call the MeCab from the python environments.
But the advantage of Fugashi's tokenization speed will be dropped while using slower CNN parser and ner.
And actually, the dictionaries of the MeCab are not tuned for business domains, and not well maintained recently (previous release was in March 2016).

How do you think these pros and cons?

polm:
@hiroshi-matsuda-rit Thanks for your comments. I also want to get away from MeCab, and I would love to switch spaCy from using fugashi to SudachiPy, but right now I think SudachiPy is just too slow. In a [benchmark I wrote](https://github.com/polm/ja-tokenizer-benchmark) it is over 200 times slower than fugashi. 

To address that I plan to work on Cythonizing SudachiPy. Since most of the code in Sudachi has not been Cythonized I think we can expect a big performance improvement, and at that point I think it should be fine to integrate. 

For now, I think this PR should be closed, and we can re-open a new PR later. (Also, if #3945 is a better version then this should be closed anyway.)

hiroshi-matsuda-rit:
Thank you, @polm.
I believe your approach, cythonizing SudachiPy, is the best solution for the spaCy Japanese model.
Let us discuss about this in #3945 later.
I'd close this issue.

PR #3902, opened 2019-07-03
Update pretrain to prevent unintended overwriting of weight files

### LINKED ISSUE #3859: pretrain with -t2v to resume overwrites existing weight files
Using the same output directory for a continued run (`-t2v`) of `pretrain` will overwrite existing weight files, since the counting of `epoch` begins at 0 again, even though you might use a weight file which was created after 100 iterations. Then [_save_model](https://github.com/explosion/spaCy/blob/81c12640ab4f16ad2a9c718416f32409d0c457d2/spacy/cli/pretrain.py#L139) saves the next epoch as `model0.bin` instead of `model101.bin`.

## Your Environment
* Python Version Used: 3.7.3
* spaCy Version Used: latest master commit

## Possible solution
This should either be added to the docs or maybe some useful information is included in the weight file. Trying to parse the epoch from the `-t2v` value is not a good idea I think.

BreakBB:
## Description
This PR fixes #3859. It tries to parse the epoch number from the given weight file name from `--init-tok2vec`.

For the situation where the weight file has been renamed and does not match the ' modelX.bin' pattern the `--epoch-start` argument was introduced. This is an integer which has to be greater or equal to 0.

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/cli/pretrain.py
+++ b/spacy/cli/pretrain.py
@@ -5,6 +5,7 @@ import plac
 import random
 import numpy
 import time
+import re
 from collections import Counter
 from pathlib import Path
 from thinc.v2v import Affine, Maxout
@@ -65,6 +66,13 @@ from .train import _load_pretrained_tok2vec
         "t2v",
         Path,
     ),
+    epoch_start=(
+        "The epoch to start counting at. Only relevant when using '--init-tok2vec' and the given weight file has been "
+        "renamed. Prevents unintended overwriting of existing weight files.",
+        "option",
+        "es",
+        int
+    ),
 )
 def pretrain(
     texts_loc,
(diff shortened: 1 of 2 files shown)


BreakBB:
I am not sure about whether or not the "new in 2.1" tag in the docs should be added to the new argument or not

ines:
Thanks for this!

> I am not sure about whether or not the "new in 2.1" tag in the docs should be added to the new argument or not

I guess it should probably be 2.1.5, since that will be the next version?

honnibal:
Thanks again for your help @BreakBB ! We really appreciate all the replies and the patches you've been submitting.

Btw, I was looking for your email to reach out, basically just saying it was a shame you couldn't make it to spaCy IRL and I hope we can catch up in future. Perhaps you could email us? It'd be good to have your contact :).

BreakBB:
@honnibal Thanks for your kind words. Check your mails when you have time.

PR #3911, opened 2019-07-05
Evaluation of NER model per entity type, closes #3490

### LINKED ISSUE #3490: Evaluation of NER model per entity type
I evaluated my NER model (built on three entity types ORG, PERSON, and LOC) and got results like this for one of the folds

```
{'uas': 0.0,
 'las': 0.0,
 'ents_p': 94.22382671480143,
 'ents_r': 92.226148409894,
 'ents_f': 93.21428571428572,
 'tags_acc': 0.0,
 'token_acc': 100.0}
```

using this method

```
import spacy
from spacy.gold import GoldParse
from spacy.scorer import Scorer
def evaluate(ner_model, examples):
    scorer = Scorer()
    for input_, annot in examples:
        doc_gold_text = ner_model.make_doc(input_)
        gold = GoldParse(doc_gold_text, entities=annot['entities'])
        pred_value = ner_model(input_)
        scorer.score(pred_value, gold)
    return scorer.scores
```
However I know that these values are for all entities, but I wanted to see what the results were per entity. I search here and on stackoverflow, and I found one of the solutions someon posted as follows

https://stackoverflow.com/a/53345979/1462656

```
import spacy
from spacy.gold import GoldParse
from spacy.scorer import Scorer

def evaluate(nlp, examples, ent='PERSON'):
    scorer = Scorer()
    for input_, annot in examples:
        text_entities = []
        for entity in annot.get('entities'):
            if ent in entity:
                text_entities.append(entity)
        doc_gold_text = nlp.make_doc(input_)
        gold = GoldParse(doc_gold_text, entities=text_entities)
        pred_value = nlp(input_)
        scorer.score(pred_value, gold)
    return scorer.scores
```
But when I ran it on one of my folds, I got a result like this

```
{'uas': 0.0,
 'las': 0.0,
 'ents_p': 10.743801652892563,
 'ents_r': 90.27777777777779,
 'ents_f': 19.202363367799112,
 'tags_acc': 0.0,
 'token_acc': 100.0}
```

Which leads me to believe that the GoldParse method is using all entities in its calculation so the solution above won't work.

One work around I can think of would be to create individual gold standard files (one for each entity type) and use the above solution on each entity type. 

But is there a more automated way to use the GoldParse method so that it calculates the scores based on a specific entity type? (without having to eliminate entities manually from the input)

elbaulp:
## Description

Now each ent score is tracked individually in order to have its own Precision, Recall and F1 Score.

This issue is discussed here in #3490. 

~~This is an ongoing pull request, there is still a few things to do.~~

### Types of change

This is an enhancement to show scores per entity tag, in addition to the overall scores.

## Checklist

### DIFF
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -105,7 +105,6 @@ class Scorer(object):
             "ents_f": self.ents_f,
             "tags_acc": self.tags_acc,
             "token_acc": self.token_acc,
-
         }
 
     @property
@@ -113,7 +112,8 @@ class Scorer(object):
         """RETURNS (dict): Scores per NER entity
         """
         return {
-            k: {'p': v.precision * 100, 'r': v.recall * 100, 'f': v.fscore * 100} for k,v in self.ner_per_ents.items()
+            k: {"p": v.precision * 100, "r": v.recall * 100, "f": v.fscore * 100}
+            for k, v in self.ner_per_ents.items()
         }
 
     def score(self, doc, gold, verbose=False, punct_labels=("p", "punct")):
@@ -159,8 +159,8 @@ class Scorer(object):
                     cand_deps.add((gold_i, gold_head, token.dep_.lower()))
         if "-" not in [token[-1] for token in gold.orig_annot]:
             cand_ents = set()
-            current_ent = {k.label_:set() for k in doc.ents}
-            current_gold = {k.label_:set() for k in doc.ents}
+            current_ent = {k.label_: set() for k in doc.ents}
+            current_gold = {k.label_: set() for k in doc.ents}
             for ent in doc.ents:
                 if ent.label_ not in self.ner_per_ents:
                     self.ner_per_ents[ent.label_] = PRFScore()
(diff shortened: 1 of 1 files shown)


elbaulp:
I have one question, should I print the scores for each entities in the return value of `def scores(self)` or should I create other method for that purpose, like [def scores_per_ents(self)](https://github.com/explosion/spaCy/pull/3911/files#diff-9efd65585aea4ec44c25a33dabd547a4R112)?

honnibal:
@elbaulp Thanks very much for this! It's definitely something that should've been done long ago.

Regarding your question, I think it makes sense to just add another key into the dict returned by the `scores` function? I'm not sure what the key should be called, but maybe something like `ents_per_type`? @ines , what do you think?

Also, we've started using the Black autoformatter. Could you do `pip install black` and then run `black spacy/scorer.py`?

elbaulp:
> Also, we've started using the Black autoformatter. Could you do `pip install black` and then run `black spacy/scorer.py`?

Done!

Will wait for @ines to see her opinion regarding where to place the scores.

Cheers.

ines:
`scores_per_ents` looks great 👍

@matt The `scores` are also automatically added to the model meta after training, right? So are you sure we want the full per-label score in there by default?

PR #3916, opened 2019-07-07
Update example and sign contributor agreement

askhogan:
## Checklist


ines:
Thanks! Looks like I missed this when I replaced the filtering logic with `filter_spans` 👍

PR #3918, opened 2019-07-07
Added support for serializing overwrite and ent_id_sep

### LINKED ISSUE #3526: Saving model with EntityRuler pipeline doesn't preserve the overwrite flag

```python
    # Create the model
    nlp = spacy.load("xx")
    ruler = EntityRuler(nlp, overwrite_ents=True)
    ruler.add_patterns([{"label": "ORG", "pattern": "Apple"}])
    nlp.add_pipe(ruler)
    nlp.to_disk("test_model")

    print(nlp.pipeline)
    print(nlp.pipeline[-1][-1].patterns)
    print(nlp.pipeline[-1][-1].overwrite) # This prints True

    # Load the saved model
    nlp3 = spacy.load("test_model")
    print(nlp3.pipeline)
    print(nlp3.pipeline[-1][-1].patterns)
    print(nlp3.pipeline[-1][-1].overwrite) # This prints False
```

## Your Environment

* Operating System: Mac Mojave
* Python Version Used: 3.6
* spaCy Version Used: 2.1.3
* Environment Information:

I am trying to use `EntityRuler` by adding it to the existing xx model and saving the model to disk as described in the docs but I noticed after reloading the model from disk the `overwrite_ents` flag is not preserved. Any ideas?

kognate:
Added support for serializing `overwrite` and `ent_id_sep`

## Description
This PR fixes #3526 and serializes `overwrite` and `ent_id_sep`.   I tried to follow patterns I found in `nn_parser.pyx` and `Sentencizer` for the bytes and file-based serialization changes.  New tests include (similar) example from the issue itself, and a test of the `to_bytes`/`from_bytes` serialization.

The code handles restoring from serialized objects from the previous format and should be fully backwards compatible (serialized elements from before this change can be deserialized).  

### Types of change
This is a bug fix and does not change overall functionality.  

## Checklist
- [x] I have submitted the spaCy Contributor Agreement. (in this PR)

### DIFF
--- a/spacy/tests/regression/test_issue3526.py
+++ b/spacy/tests/regression/test_issue3526.py
@@ -6,10 +6,8 @@ from spacy.tokens import Span
 from spacy.language import Language
 from spacy.pipeline import EntityRuler
 from spacy import load
-from tempfile import mkdtemp
-from shutil import rmtree
 import srsly
-from pathlib import Path
+from ..util import make_tempdir
 
 @pytest.fixture
 def patterns():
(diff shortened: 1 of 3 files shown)


ines:
Thanks so much for taking this on and the extensive tests – much appreciated! 🙏

spacy/tests/regression/test_issue3526.py
```diff
@@ -0,0 +1,91 @@
...
+    assert len(new_ruler) == len(ruler)
+    assert new_ruler.patterns == ruler.patterns
+    assert new_ruler.overwrite is not ruler.overwrite
+
+
+def test_entity_ruler_from_disk_old_format_safe(patterns, en_vocab):
+    nlp = Language(vocab=en_vocab)
+    ruler = EntityRuler(nlp, patterns=patterns, overwrite_ents=True)
+    try:
+        tmpdir = mkdtemp()
```
ines: The implementation here is totally fine, but if we wanted to avoid a little bit of code duplication, I think we could use the tmpfile/tmpdir helpers in `tests.util` (`make_tempfile` and `make_tempdir`).

kognate: Thank you. I've updated the PR to use that util function

ines: That was quick, thanks! 👍

PR #3919, opened 2019-07-07
Added an argument to `EntityRuler` constructor to pass attrs to the `PhraseMatcher`

### LINKED ISSUE #3822: in EntityRuler, allow using case insensitive phrase matches
EntityRuler allow using both token patterns and phrase patterns

for token patterns it's possible to specify whether the match  is case sensitive or not.
for phrase patterns the option isn't available 

the workaround I use now is:
```
er = spacy.pipeline.EntityRuler(nlp)
# override the phrase matcher with case insensitive matcher
er.phrase_matcher = spacy.matcher.PhraseMatcher(nlp.vocab, attr='LOWER')
```

though it would be cleaner if this was an option when creating an EntityRuler object

kognate:
Added an argument to `EntityRuler` constructor to pass attrs to the `PhraseMatcher`

## Description
This adds an argument to the `EntityRuler` constructor to allow passing arguments
to the `PhraseMatcher` it contains.  This fixes #3822 and depends on #3918 

### Types of change
This is an enhancement that allows attributes to pass down rather than re-initializing
the `PhraseMatcher`.  

## Checklist


ines:
Thanks a lot – also really appreciate the docs update 👍 I merged #3918 first, I hope that wasn't wrong? In any case, if you want to resolve the merge conflicts to make sure it's all correct, just let me know. Otherwise I can also take a look at it later!

kognate:
I sorted out the merge-conflicts (I was working on both issues and my branches got sloppy).  Thank you.

ines:
Thanks so much! 👍

Btw, I'm so puzzled by the test flakiness on Python 3.5... it's not *always* happening, sometimes it passes. But sometimes it doesn't? See here: https://dev.azure.com/explosion-ai/Public/_build/results?buildId=902&view=logs Do you have an idea what might be going on here? 🤔

kognate:
I do not have an explanation for the flaky test.  I ran this locally on MacOS with python 3.5 and 3.7 and it was green.  it kindof looks like the deserialized patterns are in a different order than they started, but I haven't confirmed that.   I may modify this test to check differently to get around this, but that doesn't feel like a very good solution.

ines:
@kognate Yeah, good point, maybe this is a deserialization quirk and related to the ordering. I guess it'd be fine to work around this for now in the tests – eventually, we'll be able to drop Python 3.5 anyways 🙂

ines:
Okay, looks good, thanks for doing this so quickly. Merging now – fingers crossed! 🤞

PR #3939, opened 2019-07-10
Update tokenizer and doc init example 

### LINKED ISSUE #3892: Tokenizer fails to initilize correctly (or different initialization behaviors)
This is according to the following [tokenizer initialization example](https://spacy.io/api/tokenizer#init).

import spacy
nlp = spacy.load("en_core_web_lg") # for example
test_case = "I don't know."
tokenizer_lame = Tokenizer(nlp.vocab)
[token for token in tokenizer_lame(test_case)] # will produce first output
tokenizer_awesome = nlp.Defaults.create_tokenizer(nlp)
[token for token in tokenizer_awesome(test_case)] # will produce second output

**First output:**
[I, don't, know.] # BAD

**Second output:**
[I, do, n't, know, .] # GOOD

**Notes:**
1. Not sure if this is the expected behavior (?), but even so, presenting both initializations as the same is misleading (IMHO).
2. I came across this when I used [spacy_conll](https://spacy.io/universe/project/spacy-conll){ @BramVanroy  } (and passed the flag is_tokenized=False), as they use the first method I posted as their tokenizer initialization step.

## Your Environment
* **spaCy version:** 2.1.4
* **Platform:** Windows-10-10.0.17134-SP0
* **Python version:** 3.6.5

BreakBB:
## Description
This PR:
- updates the documentation examples of the tokenizer creation to fix #3892 
- updates the `Doc.__init__` example in the methods docstring
- fixes the hyperlink to `Doc.to_json`

### Types of change
Documentation changes

## Checklist

### DIFF
--- a/website/docs/api/tokenizer.md
+++ b/website/docs/api/tokenizer.md
@@ -25,7 +25,7 @@ Create a `Tokenizer`, to create `Doc` objects given unicode text. For examples o
 > from spacy.lang.en import English
 > nlp = English()
 > # Create a Tokenizer with the default settings for English
-> # including matching rules and exceptions
+> # including punctuation rules and exceptions
 > tokenizer = nlp.Defaults.create_tokenizer(nlp)
 > ```
 
--- a/website/docs/api/tokenizer.md
+++ b/website/docs/api/tokenizer.md
@@ -9,7 +9,10 @@ Segment text, and create `Doc` objects with the discovered segment boundaries.
 
 ## Tokenizer.\_\_init\_\_ {#init tag="method"}
 
-Create a `Tokenizer`, to create `Doc` objects given unicode text. For examples of how to construct a custom tokenizer with different tokenization rules, see the [usage documentation](https://spacy.io/usage/linguistic-features#native-tokenizers).
+Create a `Tokenizer`, to create `Doc` objects given unicode text. For examples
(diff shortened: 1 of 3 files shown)


website/docs/api/tokenizer.md
```diff
@@ -18,11 +18,14 @@ Create a `Tokenizer`, to create `Doc` objects given unicode text.
...
 > from spacy.lang.en import English
 > nlp = English()
+> # Create a blank Tokenizer with just the English vocab
 > tokenizer = Tokenizer(nlp.vocab)
 >
 > # Construction 2
 > from spacy.lang.en import English
 > nlp = English()
+> # Create a Tokenizer with the default settings for English
+> # including matching rules and exceptions
```
ines:
```suggestion
> # including punctuation rules and exceptions
```

I think this is a bit more straightforward – othwerwise, a user may think that it's actually rules for the `Matcher` and get confused.

ines:
Thanks so much! 🎉

website/docs/api/tokenizer.md
BreakBB: Hehe, "punctuation" was what I wrote in the first place but after another look at the code I changed it. But you're right it most likely would have lead to confusion.

Issue #3625, opened 2019-04-22
For hindi text period '.' punctuation not splitting text


gauravgr:
## How to reproduce the behaviour
from spacy.lang.hi import Hindi
nlp = Hindi()
doc = nlp(u"hi. how हुए. होटल, होटल")
print([token.text for token in doc])

**Output**
['hi', '.', 'how', 'हुए.', 'होटल', ',', 'होटल']

**Issue**
For 'हुए.'  '.' should be splitted, but it's part of 'हुए' after tokenization. Where as it works properly for ',' punctuation

## Your Environment

* **spaCy version:** 2.1.3
* **Platform:** Windows-10-10.0.10240-SP0
* **Python version:** 3.6.4

ines:
Hmmm, maybe the character classes used in the default punctuation rules currently don't include certain unicode characters, so the rules that say "split `.` after a letter" aren't applied? It should be okay, though – see here: 

https://github.com/explosion/spaCy/blob/master/spacy/lang/char_classes.py#L196-L200

This is definitely worth investigating. Maybe it even makes sense to implement more specific punctuation rules for Hindi – I'm not sure which puncutation characters are common and which aren't, but we might even be able to use a much simpler rule set in this case.

PR #3948, opened 2019-07-11
Fix default punctuation rules for splitting Hindi text

yash1994:
## Description

- Fix default punctuation rules for Hindi text

- Add Hindi Unicode character range to the character class

Please see issue #3625 for more details.

### Types of change

- Bug fix

## Checklist

### DIFF
--- a/.github/contributors/yashpatadia.md
+++ b/.github/contributors/yashpatadia.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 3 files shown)


honnibal:
Looks good, thanks!

We're about to release v2.1.5, so I want to hold off on merging this into a point release, just in case it causes problems (although I don't see how it could). We should be able to merge it soon after that though. Thanks again for the help!

ines:
I think it'll be fine to merge – it only affects Hindi and the patch looks good, so I'll just go ahead and merge it 🙂

PR #3957, opened 2019-07-12
Make Danish tokenizer split on forward slash

sorenlind:
## Description

I've implemented changes to make the Danish tokenizer split on forward slash to make it work properly with terms separated by forward slash, such as "Bus/tog/metro" (busses, trains, metro).  This makes it behave identically to the English tokenizer in this regard. Since there are several abbreviations that contain forward slash (and which should be treated as one token), I've added exceptions for (most of?) these.

Also, I ran Black on the Danish `punctuation.py` to convert a few remaining single quotes to double quotes.

### Types of change

This is an enhancement to the Danish tokenizer.

## Checklist

### DIFF
--- a/spacy/lang/da/punctuation.py
+++ b/spacy/lang/da/punctuation.py
@@ -14,10 +14,11 @@ _infixes = (
     + [
         r"(?<=[{al}])\.(?=[{au}])".format(al=ALPHA_LOWER, au=ALPHA_UPPER),
         r"(?<=[{a}])[,!?](?=[{a}])".format(a=ALPHA),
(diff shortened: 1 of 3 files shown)


ines:
Looks good, thanks! 👍

hafiz031:
It is creating problems and splitting abbreviations like: `a/c` (account), `b/l` (balance) etc.

adrianeboyd:
This PR is not related to the `xx` tokenizer settings. Please see the documentation on tokenizer exceptions ("special cases"): https://spacy.io/usage/linguistic-features#special-cases

The discussion board is a better spot for similar questions in the future: https://github.com/explosion/spaCy/discussions

PR #3964, opened 2019-07-15
Fix Links to Custom Attributes Documentation in Rule-Based Matching Documentation

pmbaumgartner:
##  Description
Two of the links to the documentation for custom extension attributes in the Rule-Based Matching page were missing the `/usage/` prefix in their url. This corrects those links by adding that prefix back in.

This does mess with the alignment of a markdown table containing one of the links. I assume that you have some automated way to fix that alignment, so I did not adjust that.

### Types of change
Documentation hyperlinks.


ines:
Thanks so much! 👍

> This does mess with the alignment of a markdown table containing one of the links. I assume that you have some automated way to fix that alignment, so I did not adjust that.

Yeah, the alignment isn't super important. I use Prettier, which fixes this automatically on save. (Tbh, I'm not even a huge fan of formatting tables like that because it creates unnecessary diffs, but the other Markdown auto-formatting that Prettier does is super useful.)

Issue #3962, opened 2019-07-13
Doc.to_json() method fails when called on doc generated via `Span.as_doc()`


JohnStuartRutledge:
## How to reproduce the behaviour

It would appear that calling `to_json()` on a Doc that was created via the `as_doc()` method is prone to failure. This may very well be connected to other issues with with `as_doc` like the that mentioned in [issue #3669 ](https://github.com/explosion/spaCy/issues/3669)

```python
import spacy

nlp = spacy.load('en')
doc = nlp('He jests at scars, that never felt a wound.')
span = doc[0:3]      # "He jests at"
doc2 = span.as_doc()
doc2.to_json()
# IndexError: [E040] Attempt to access token at 7, max length 3.
```

The above error seems to trigger when second token of the span (`jests`), calls [spacy.tokens.Token.head.\_\_get\_\_()](https://github.com/explosion/spaCy/blob/master/spacy/tokens/token.pyx#L609) and hits the `self.c.head`

## Environment
* **spaCy version:** 2.1.6
* **Platform:** Darwin-18.6.0-x86_64-i386-64bit
* **Python version:** 3.6.4
* **Models:** en

svlandeg:
Yep, I can replicate this bug with `en_core_web_md` (not with `en_core_web_lg`) - looks like it's definitely connected to the trouble caused by the dependency parsing, as discussed in that other issue...

svlandeg:
@JohnStuartRutledge : I think we can fix this, cf. https://github.com/explosion/spaCy/pull/3969. Basically what happens in your example, is that the second word `jests` refers to `felt` (index: 7) as its head word in the dependency graph, but after the `as_doc` operation that index is out of bounds. So we need to prevent copying that kind of out-of-span information ...

Issue #3669, opened 2019-05-03
Segmentation fault when using span.as_doc() method


apuranik1:
## How to reproduce the behaviour

I am trying to parse a number of reddit comments, dumped from pushshift.io. A file with a sample of these comments (about 130,000) can be found here: [comment_sample_100k.csv](https://drive.google.com/file/d/1eG4hNb7xAD_DcB4pcoUJIfZxxqeffNnL/view?usp=sharing).

The following script is a heavily shortened version of the one I'm using to do the parsing: [spacy_failure.py](https://gist.github.com/apuranik1/714440fee410e7f4356f417e6ac43125). There are two custom components in the pipeline: the tokenizer and the sentence boundary setter. The script builds the custom Language object, reads lines in csv format from stdin, parses them, and throws away the result. Running the script looks like:
```bash
$ python failure_script.py < comment_sample_100k.csv
```
More than half the time, this command will eventually segfault. It's nondeterministic - sometimes it crashes in 30 seconds, other times it runs for 10 minutes, and sometimes it finishes successfully.

This is about as short as I could make the script while reproducing the error. Lots of seemingly unrelated lines will prevent the error from occurring. Certain small details that prevent the error are marked with comments. Certain other lines that aren't critical to the error but seem to increase the crash rate are also marked. Occasionally, instead of a segfault, the CSV reader will be corrupted and raise an exception.

My best guess is the issue is related to resource cleanup of the object returned by `sent.as_doc().to_bytes()`. Changing where the `Sentence` object gets garbage collected seems to change the outcome.

I'm using spaCy 2.1.0, but I've replicated it on 2.1.3 as well.

## Info about spaCy

* **spaCy version:** 2.1.0
* **Platform:** Darwin-18.2.0-x86_64-i386-64bit
* **Python version:** 3.7.2
* **Models:** en

honnibal:
Thanks for the report.

Could you try avoiding the `span.as_doc()` call? I've had trouble with this before, and it doesn't work the way it was originally intended. Originally I wanted it to be a zero-copy operation, but that didn't work. I'm very suspicious that this could be where the bug is, as it's a quite untested method that had bugs previously.

You might find the serialization code here useful: https://github.com/explosion/spaCy/blob/master/spacy/tokens/_serialize.py This lets you serialize a collection of `Doc` objects in an efficient format. It should let you then reconstruct your sentences, given the `Doc` objects.

apuranik1:
[...]

PR #3969, opened 2019-07-15
Fix dependency copy for as_doc

svlandeg:
The `span.as_doc` functionality seems to have some issues (cf. Issue #3962 and Issue #3669) which I think are related to the dependency graph being cut into pieces. Some tokens may still refer to other tokens outside the `span`, thus resulting in an out of bounds error when that token is being accessed.

## Description
In `doc.from_array` I have added code that checks the offset of the head for each token, and does not copy the value unless that token is inside the doc : `if (i+value) in range(self.length)`. The token head defaults to 0 (i.e. itself) if the value is not copied/set.

This fix feels a little hacky, so happy to get feedback. Also currently the actual dependency link (e.g. `ccomp`) is being copied even if the head isn't, but I think we need that to ensure other cases keep working where a token refers to itself as head.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tests/regression/test_issue3962.py
+++ b/spacy/tests/regression/test_issue3962.py
@@ -1,0 +1,15 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+import spacy
+
+
+def test_issue3962():
+    nlp = spacy.load('en_core_web_md')
+    doc = nlp('He jests at scars, that never felt a wound.')
+    span = doc[0:3]
+    doc2 = span.as_doc()
+    assert doc2
+    doc2_json = doc2.to_json()
+    print(doc2_json)
+    assert doc2_json
--- a/spacy/tests/regression/test_issue3962.py
+++ b/spacy/tests/regression/test_issue3962.py
@@ -18,7 +18,7 @@ def doc(en_tokenizer):
 
 def test_issue3962(doc):
     """ Ensure that as_doc does not result in out-of-bound access of tokens.
-    This is achieved by setting the head to itself it it would lie out of the span otherwise."""
+    This is achieved by setting the head to itself if it would lie out of the span otherwise."""
     assert doc[1].dep_ == "ccomp"
     assert doc[1].head.text == "felt"
     assert doc[2].dep_ == "prep"
--- a/spacy/tests/regression/test_issue3962.py
+++ b/spacy/tests/regression/test_issue3962.py
@@ -17,8 +17,26 @@ def doc(en_tokenizer):
 
 
 def test_issue3962(doc):
-    span = doc[0:3]
+    """ Ensure that as_doc does not result in out-of-bound access of tokens.
+    This is achieved by setting the head to itself it it would lie out of the span otherwise."""
+    assert doc[1].dep_ == "ccomp"
+    assert doc[1].head.text == "felt"
+    assert doc[2].dep_ == "prep"
+    assert doc[2].head.text == "jests"
+    assert doc[3].dep_ == "pobj"
+    assert doc[3].head.text == "at"
+    assert doc[4].dep_ == "punct"
+    assert doc[4].head.text == "felt"
+    span = doc[1:5]
     doc2 = span.as_doc()
     assert doc2
     doc2_json = doc2.to_json()
     assert doc2_json
+    assert doc2[0].dep_ == "ccomp"
+    assert doc2[0].head.text == "jests"  # head set back to itself
+    assert doc2[1].dep_ == "prep"
+    assert doc2[1].head.text == "jests"
+    assert doc2[2].dep_ == "pobj"
+    assert doc2[2].head.text == "at"
+    assert doc2[3].dep_ == "punct"
+    assert doc2[3].head.text == ","      # head set back to itself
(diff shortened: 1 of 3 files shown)


honnibal:
Hmm, yes I see the problem. This is a pretty nasty case :(.

It does make sense to handle these as fragments, but the problem is we'll be getting the fragments with `doc.sents`, which feels pretty undesirable.

Perhaps we could do the following to reconcile the tree:

* Find all the words whose head is outside the span.
* For each of those words, look for an ancestor that's within the span (using the `token.ancestors` attribute.) Attach to the lowest within-span ancestor, using the label `dep`.
* If there is no within-span ancestor, attach to the previous `ROOT` token using the label `dep`.

The `dep` label is kind of the miscellaneous one, used for tree-shapes that the PSG-to-dependencies converter couldn't classify into a more specific construction. So it's the best one to use for these cases where we're just stitching things together into a tree, even if it doesn't make good linguistic sense.

spacy/tests/regression/test_issue3962.py
```diff
@@ -0,0 +1,42 @@
...
+    doc2_json = doc2.to_json()
+    assert doc2_json
+    assert doc2[0].dep_ == "ccomp"
+    assert doc2[0].head.text == "jests"  # head set back to itself
+    assert doc2[1].dep_ == "prep"
+    assert doc2[1].head.text == "jests"
+    assert doc2[2].dep_ == "pobj"
+    assert doc2[2].head.text == "at"
+    assert doc2[3].dep_ == "punct"
+    assert doc2[3].head.text == ","      # head set back to itself
```
ines: @svlandeg Sorry for the nitpicky comment, but just for consistency: Can you set your editor to automatically insert newlines at the end of files? Or I think if you set up [`black`](https://github.com/python/black) to auto-format on save, it should also do it by default (at least for Python).

svlandeg:
@honnibal : sounds like a good idea and I can definitely implement this. Good idea to use the generic `dep`.

Note however that we'll never recover from this specific example, with the sentence `He jests at scars, that never felt a wound.` where the `ROOT` is `felt` which is not in the span, and there's no common ancestors either, so you'll always end up with two "sentences", unless you introduce some kind of new artificial root (?)

svlandeg:
Actually, I'm not sure your third point (fall-back) makes sense: if there's no ancestor in the span, attaching it to the previous ROOT would effectively attach it to the previous sentence?

svlandeg:
Ok, so the latest push changes the behaviour to locate the first ancestor within the span. I moved the code to `span.pyx` instead of `doc.pyx` because we need access to the data outside of the span (i.e. the original array) to navigate the ancestors. I also changed the link text to `dep` when the head was changed either to the ancestor, or to itself (in case no ancestor was found).

I added a bit to the unit test that tests a case where this procedure does end up in 1 sentence instead of 2 (when taking `doc[6:9]`, "a" is attached to "felt" instead of to "wound"), though the first example (`doc[1:5]`) still remains two sentences as I explained above.

honnibal:
I'm mostly thinking of something like the heuristics in `span.root`: https://github.com/explosion/spaCy/blob/master/spacy/tokens/span.pyx#L490

I wonder whether we could keep it simple and use exactly this? For all words that point outside the span, attach them to the `span.root` with label `dep`? I think that actually does roughly what I described, but would be really easy to implement, and definitely easy to document.

svlandeg:
Ok, so right now the implementation goes as such, if a token's `head` is outside the span:
1. Navigate its ancestors by looping through `orig_array` (L256 in `Span`), take the first good ancestor, and set the `DEP` to `dep`  --> we can remove this step again if you like, but I think this keeps the tree as intact as possible

2. If still no suitable ancestor is found, set to the `span.root.i` and set the `DEP` to `dep`

This fixes the first unit test as `,` is now attached to `jests` (after `felt` was found to be out-of-span) and now we have 1 sentence there. Hooray :-)

However - I still think this may artificially merge sentences. I added a new example, taking span [1:7] in the sentence "He jests at scars. They never felt a wound." with `ROOT` tokens `jests` and `felt` respectively. After breaking the doc into the span "jests at scars. They never", both `They` and `never` will get attached to `jests` in the previous sentence. The final span will be one sentence (instead of, arguably, two).

honnibal:
I think you're right that we'd want two sentences in that example. Do you see an easy way to improve the logic.

svlandeg:
Perhaps we could introduce "artificial new heads", being the first token encountered in a sentence that got "cut up". Basically as soon as we encounter a token of which the root ancestor is outside of the span (and no other ancestors were appropriate), we link that token to itself, and keep track that this token replaces the old root ancestor. Then whenever we encounter another token with the same root (i.e. within the same sentence) we link it to the new artificial one. This should ensure that we'll keep the same number of sentences. What do you think?

svlandeg:
Ok, I implemented the above and it seems to work, let me know whether you think this is OK.

Then when testing, I stumbled on this: In the last unit test, the original text is `He jests at scars. They never felt a wound.` When taking span `[1:7]` which should be `jests at scars. They never` and then printing the sentences, the first sentence becomes `jests at scars .` cf [here](https://github.com/explosion/spaCy/blob/3949b4661eea2722dc0d8c8b11f891d09809676f/spacy/tests/regression/test_issue3962.py#L111), note the space before the dot. Is this ok, or does this point to some (other) trouble regarding the copying of the whitespaces?

Issue #3968, opened 2019-07-15
Invalid per entity NER accuracy


FallakAsad:
I trained a NER model with some new entity types, My model accuracy is as follows on a test set  which shows per entity accuracy as well as total accuracy. 
{
  "ents_f": 84.65116279069768,
  "ents_p": 95.28795811518324,
  "ents_per_type": {
    "label_1": {
      "f": 39.47368421052631,
      "p": 24.59016393442623,
      "r": 100.0
    },
    "label_2": {
      "f": 90.47619047619048,
      "p": 90.47619047619048,
      "r": 90.47619047619048
    },
    "label_3": {
      "f": 80.00000000000001,
      "p": 80.0,
      "r": 80.0
    },
    "label_4": {
      "f": 80.0,
      "p": 72.72727272727273,
      "r": 88.88888888888889
    },
    "label_5": {
      "f": 14.14141414141414,
      "p": 8.045977011494253,
      "r": 58.333333333333336
    },
    "label_6": {
      "f": 100.0,
      "p": 100.0,
      "r": 100.0
    },
    "label_7": {
      "f": 50.0,
      "p": 50.0,
      "r": 50.0
    }
  },
  "ents_r": 76.15062761506276,
  "las": 0.0,
  "tags_acc": 0.0,
  "token_acc": 100.0,
  "uas": 0.0
}

However, If I evaluate same model on the slightly different set which contains same docs as in previous test set but have few more annotation for label-4 and label-5 which were missing for some docs in previous test set, the model gives '0' f-score for label-3 for which no annotation were changed in new test set. The evaluation result on new test set is:
{
  "ents_f": 84.65346534653466,
  "ents_p": 95.53072625698324,
  "ents_per_type": {
    "label_1": {
      "f": 32.35294117647059,
      "p": 19.298245614035086,
      "r": 100.0
    },
    "label_2": {
      "f": 94.44444444444444,
      "p": 94.44444444444444,
      "r": 94.44444444444444
    },
    "label_3": {
      "f": 0.0,
      "p": 0.0,
      "r": 0.0
    },
    "label_4": {
      "f": 80.0,
      "p": 72.72727272727273,
      "r": 88.88888888888889
    },
    "label_5": {
      "f": 14.432989690721651,
      "p": 8.13953488372093,
      "r": 63.63636363636363
    },
    "label_6": {
      "f": 100.0,
      "p": 100.0,
      "r": 100.0
    },
    "label_7": {
      "f": 50.0,
[...]

PR #3982, opened 2019-07-17
Bugfix/issue 3968

FallakAsad:
## Description
Per entity calculations for  f-score, precision and recall were incorrect because of bug in spacy/scorer.py. See issue https://github.com/explosion/spaCy/issues/3968 for more details.

### Types of change
A bug fix in spacy/scorer.py for invalid per entity values for f-score, precision and recall. I ran tests with "pytest [spacy directory] --slow" command 

Test result before my changes:
1967 passed, 319 skipped, 60 xfailed, 1 warnings in 36.88 seconds

Test result after my changes:
1968 passed, 319 skipped, 60 xfailed, 1 warnings in 36.83 seconds 

## Checklist

### DIFF
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -172,12 +172,8 @@ class Scorer(object):
                     self.ner_per_ents[ent.label_].fp += 1
                 else:
                     cand_ents.add((ent.label_, first, last))
-                    current_ent[ent.label_].add(
-                        tuple(x for x in cand_ents if x[0] == ent.label_)
-                    )
-                    current_gold[ent.label_].add(
-                        tuple(x for x in gold_ents if x[0] == ent.label_)
-                    )
+                    [current_ent[ent.label_].add(x) for x in cand_ents if x[0] == ent.label_]
+                    [current_gold[ent.label_].add(x) for x in gold_ents if x[0] == ent.label_]
             # Scores per ent
             [
                 v.score_set(current_ent[k], current_gold[k])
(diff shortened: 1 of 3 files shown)


spacy/scorer.py
```diff
@@ -172,12 +172,8 @@ def score(self, doc, gold, verbose=False, punct_labels=("p", "punct")):
                     self.ner_per_ents[ent.label_].fp += 1
                 else:
                     cand_ents.add((ent.label_, first, last))
-                    current_ent[ent.label_].add(
-                        tuple(x for x in cand_ents if x[0] == ent.label_)
-                    )
-                    current_gold[ent.label_].add(
-                        tuple(x for x in gold_ents if x[0] == ent.label_)
-                    )
+                    [current_ent[ent.label_].add(x) for x in cand_ents if x[0] == ent.label_]
```
honnibal: Pretty minor style thing, but I try to avoid list comprehensions that aren't assigned to anything.

```suggestion
                    current_ent[ent.label_].update([x for x in cand_ents if x[0] == ent.label_])
```

spacy/scorer.py
```diff
@@ -172,12 +172,8 @@ def score(self, doc, gold, verbose=False, punct_labels=("p", "punct")):
...
+                    [current_gold[ent.label_].add(x) for x in gold_ents if x[0] == ent.label_]
```
honnibal:
```suggestion
                    current_gold[ent.label_].update([x for x in gold_ents if x[0] == ent.label_])
```

honnibal:
Sorry my suggested changes have showed up really confusingly! I was just trying to make a small style fix to the list comprehensions. I then tried to "batch" them, which is a new Github feature...But maybe I don't understand how it works. Anyway, you might need to just apply the change manually.

Thanks for the fix! Looks great.

spacy/scorer.py
FallakAsad: Thanks, made the suggested changes.

PR #4001, opened 2019-07-22
Update annotation docs for German

adrianeboyd:
## Description

Update annotation docs to match use in Tiger:

- minor formatting fixes
- remove STTS tags not used in Tiger
- update list of dependency relations to match tiger2dep


ines:
Thanks! 🙏

Issue #4000, opened 2019-07-22
Having an error with Entity Linking


almoslmi:
I have run NEL (wikidata_entity_linking.py) and was working well after fixing some issues, but when it successfully finished processing the two dumps of Wikipedia and Wikidata, it failed to write the KB and got an AssertionError. All necessary libs are installed and added to the environment too. Any idea how to solve this issue, please?
By the way, I tried to debug the file that causes the problem (kb.pyx) but didn't work.

I am using Pycharm, and Here is the last lines of the log 
===============
4 1907 0.07248597717285156
4 1908 0.07176494598388672
Trained on 9540645 entities across 5 epochs
Final loss: 0.07176494598388672

 * get entity embeddings 2019-07-20 18:33:18.329715

 * adding 1908129 entities 2019-07-20 19:30:44.652900

 * adding aliases 2019-07-20 19:31:01.172724

kb size: 1908129 1908129 2830524
done with kb 2019-07-20 19:32:18.446255
kb entities: 1908129
kb aliases: 2830524

STEP 3b: write KB and NLP 2019-07-20 19:32:36.823641
Traceback (most recent call last):
  File "C:/Users/tareq/PycharmProjects/spaCy/examples/pipeline/wikidata_entity_linking.py", line 441, in <module>
    run_pipeline()
  File "C:/Users/tareq/PycharmProjects/spaCy/examples/pipeline/wikidata_entity_linking.py", line 111, in run_pipeline
    kb_1.dump(KB_FILE)
  File "kb.pyx", line 208, in spacy.kb.KnowledgeBase.dump
  File "kb.pyx", line 346, in spacy.kb.Writer.__init__
AssertionError

Process finished with exit code 1

========================

Later, I tried to debug the file that causes the error (kb.pyx) and got this error:

C:\Users\tareq\Anaconda3\python.exe "C:\Program Files\JetBrains\PyCharm 2019.1.3\helpers\pydev\pydevconsole.py" --mode=client --port=52781
import sys; print('Python %s on %s' % (sys.version, sys.platform))
sys.path.extend(['C:\Users\tareq\PycharmProjects\spaCy', 'C:\Users\tareq\PycharmProjects\spaCy\spacy', 'C:/Users/tareq/PycharmProjects/spaCy'])
Python 3.7.1 (default, Dec 10 2018, 22:54:23) [MSC v.1915 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
[...]

PR #4003, opened 2019-07-22
API changes for Entity linking functionality

svlandeg:
Made some small adjustements that make the NEL functionality resemble the code for the other pipes more closely. This will be followed by another PR which has the (updated) documentation for all this functionality.

### GoldParse
* `gold.links` is now a dictionary with `(ent.start_char, ent.end_char)` keys and the values being dicts with `kb_id:value` entries, representing the KB IDs mapped to either 1.0 or 0.0, indicating positive and negative examples respectively. This is similar to the `gold.cats` field for text classification.
* For training, all the `kb_id: value` combinations are used, including the negative cases
* For testing (accuracy measurement), only the positive cases are checked

### EntityLinker
* `gold.links` now corresponds 1-to-1 to `doc.ents` which allows a more proper chaining of `entity_linker.predict()` and `entity_linker.get_loss()` functions etc (just like for the other pipes)
* `EntityLinker` also now returns tensors as part of its `predict()` output
* default `context_width` set to 128

### KB
* new methods in `kb`: `get_vector` and `get_prior_prob` to easily access information about entities & aliases, + unit tests 
* In `kb.pyx`, renamed entity `prob` to `freq` (entity frequency in a corpus) to avoid confusion with `prior_prob` (which refers to the probability of an alias being linked to an entity)
* `Errors.E144` when `entity_width` or `context_width` could not be found when building the NEL model
* Created KB directory if it didn't exist to address Issue #4000

### MISC
* `black` processing caused a lot of edits in the files
* small fix in `tokenizer.pyx` documentation

### Types of change
Enhancement

## Checklist

### DIFF
--- a/bin/wiki_entity_linking/training_set_creator.py
+++ b/bin/wiki_entity_linking/training_set_creator.py
@@ -1,7 +1,6 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
-import os
 import random
 import re
 import bz2
@@ -37,7 +36,7 @@ def _process_wikipedia_texts(wikipedia_input, wp_to_id, training_output, limit=N
 
     read_ids = set()
     entityfile_loc = training_output / ENTITY_FILE
-    with open(entityfile_loc, mode="w", encoding="utf8") as entityfile:
+    with entityfile_loc.open("w", encoding="utf8") as entityfile:
         # write entity training header file
         _write_training_entity(
             outputfile=entityfile,
@@ -301,8 +300,8 @@ def _get_clean_wp_text(article_text):
 
 
 def _write_training_article(article_id, clean_text, training_output):
-    file_loc = training_output / str(article_id) + ".txt"
-    with open(file_loc, mode="w", encoding="utf8") as outputfile:
+    file_loc = training_output / "{}.txt".format(article_id)
+    with file_loc.open("w", encoding="utf8") as outputfile:
         outputfile.write(clean_text)
 
 
@@ -330,7 +329,7 @@ def read_training(nlp, training_dir, dev, limit, kb=None):
     skip_articles = set()
     total_entities = 0
 
-    with open(entityfile_loc, mode="r", encoding="utf8") as file:
+    with entityfile_loc.open("r", encoding="utf8") as file:
         for line in file:
             if not limit or len(data) < limit:
                 fields = line.replace("\n", "").split(sep="|")
@@ -349,11 +348,8 @@ def read_training(nlp, training_dir, dev, limit, kb=None):
                         # parse the new article text
                         file_name = article_id + ".txt"
                         try:
-                            with open(
-                                os.path.join(training_dir, file_name),
-                                mode="r",
-                                encoding="utf8",
-                            ) as f:
+                            training_file = training_dir / file_name
+                            with training_file.open("r", encoding="utf8") as f:
                                 text = f.read()
                                 # threshold for convenience / speed of processing
                                 if len(text) < 30000:
@@ -364,7 +360,9 @@ def read_training(nlp, training_dir, dev, limit, kb=None):
                                         sent_length = len(ent.sent)
                                         # custom filtering to avoid too long or too short sentences
                                         if 5 < sent_length < 100:
-                                            offset = "{}_{}".format(ent.start_char, ent.end_char)
+                                            offset = "{}_{}".format(
+                                                ent.start_char, ent.end_char
+                                            )
                                             ents_by_offset[offset] = ent
                                 else:
                                     skip_articles.add(article_id)
--- a/bin/wiki_entity_linking/training_set_creator.py
+++ b/bin/wiki_entity_linking/training_set_creator.py
@@ -307,18 +307,8 @@ def _write_training_article(article_id, clean_text, training_output):
 
 
 def _write_training_entity(outputfile, article_id, alias, entity, start, end):
-    outputfile.write(
-        article_id
-        + "|"
-        + alias
-        + "|"
-        + entity
-        + "|"
-        + str(start)
-        + "|"
-        + str(end)
-        + "\n"
-    )
+    line = "{}|{}|{}|{}|{}\n".format(article_id, alias, entity, start, end)
+    outputfile.write(line)
 
 
 def is_dev(article_id):
--- a/bin/wiki_entity_linking/training_set_creator.py
+++ b/bin/wiki_entity_linking/training_set_creator.py
@@ -401,15 +401,13 @@ def read_training(nlp, training_dir, dev, limit, kb=None):
                                 gold_start = int(start) - found_ent.sent.start_char
                                 gold_end = int(end) - found_ent.sent.start_char
 
-                                # add both positive and negative examples (in random order just to be sure)
+                                # add both pos and neg examples (in random order)
                                 if kb:
                                     gold_entities = {}
-                                    candidate_ids = [
-                                        c.entity_ for c in kb.get_candidates(alias)
-                                    ]
-                                    candidate_ids.append(
-                                        wd_id
-                                    )  # in case the KB doesn't have it
+                                    candidates = kb.get_candidates(alias)
+                                    candidate_ids = [c.entity_ for c in candidates]
+                                    # add positive example in case the KB doesn't have it
+                                    candidate_ids.append(wd_id)
                                     random.shuffle(candidate_ids)
                                     for kb_id in candidate_ids:
                                         entry = (gold_start, gold_end, kb_id)
@@ -418,7 +416,8 @@ def read_training(nlp, training_dir, dev, limit, kb=None):
                                         else:
                                             gold_entities[entry] = 1.0
                                 else:
-                                    gold_entities = {}
+                                    entry = (gold_start, gold_end, wd_id)
+                                    gold_entities = {entry: 1.0}
 
                                 gold = GoldParse(doc=sent, links=gold_entities)
                                 data.append((sent, gold))
--- a/bin/wiki_entity_linking/training_set_creator.py
+++ b/bin/wiki_entity_linking/training_set_creator.py
@@ -397,33 +397,43 @@ def read_training(nlp, training_dir, dev, limit, kb=None):
                                 current_doc = None
                             else:
                                 sent = found_ent.sent.as_doc()
-                                # currently feeding the gold data one entity per sentence at a time
+
                                 gold_start = int(start) - found_ent.sent.start_char
                                 gold_end = int(end) - found_ent.sent.start_char
 
-                                # add both pos and neg examples (in random order)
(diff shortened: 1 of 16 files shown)


honnibal:
With the `prob` to `freq` rename, are we actually taking frequency counts? If the value is expected to be a probability (float value between 0 and 1) I think we'd better keep it as `prob`.

spacy/kb.pyx
```diff
@@ -343,7 +372,7 @@ cdef class Writer:
             loc = bytes(loc)
         cdef bytes bytes_loc = loc.encode('utf8') if type(loc) == unicode else loc
         self._fp = fopen(<char*>bytes_loc, 'wb')
-        assert self._fp != NULL
+        assert self._fp != NULL, "Could not access %s" % loc
```
honnibal: I'm a repeat offender at this, but we should try to avoid `assert` for expected errors. Unable to access a file is definitely something that can come up, so I think this should be a conditional and raise either `IOError` or `OSError`.

spacy/kb.pyx
```diff
@@ -414,18 +443,18 @@ cdef class Reader:
...
                 return 0  # end of file
             raise IOError("error reading entity hash from input file")
 
-        status = self._read(prob, sizeof(float))
+        status = self._read(freq, sizeof(float))
         if status < 1:
             if feof(self._fp):
                 return 0  # end of file
-            raise IOError("error reading entity prob from input file")
+            raise IOError("error reading entity freq from input file")
```
honnibal: We probably want to move this error message to `spacy/errors.py`

svlandeg:
> With the `prob` to `freq` rename, are we actually taking frequency counts? If the value is expected to be a probability (float value between 0 and 1) I think we'd better keep it as `prob`.

The Wikipedia pipeline  actually does take frequency counts:

> Marjory Stoneman Douglas  173      
                                                                                                                                                                                                                                          Neil Douglas  22            
                                                                                                                                                                                                                                                     Bruce Douglas-Mann  46

 Some example codes and unit tests set this `freq` field to a floating point value though, so I'd better change those as well to avoid confusion.
The thing is, this field is not actually being used in the EL algorithm right now, so we can still change it either way.

bin/wiki_entity_linking/training_set_creator.py
```diff
@@ -249,43 +279,57 @@ def _get_clean_wp_text(article_text):
...
 def _write_training_article(article_id, clean_text, training_output):
     file_loc = training_output / str(article_id) + ".txt"
-    with open(file_loc, mode='w', encoding='utf8') as outputfile:
+    with open(file_loc, mode="w", encoding="utf8") as outputfile:
         outputfile.write(clean_text)
 
 
 def _write_training_entity(outputfile, article_id, alias, entity, start, end):
-    outputfile.write(article_id + "|" + alias + "|" + entity + "|" + str(start) + "|" + str(end) + "\n")
+    outputfile.write(
```
ines: Not a big deal, but: `start` and `end` are ints, right? If so, something like this is probably cleaner:

```python
line = "{}|{}|{}|{}|{}\n".format(article_id, alias, entity, start, end)
outputfile.write(line)
```

I think there are a few other instances as well where calling `str` around stuff makes the code unnecessarily verbose and possibly more error-prone.

spacy/pipeline/pipes.pyx
```diff
@@ -1150,49 +1152,40 @@ class EntityLinker(Pipe):
...
-            for entity in gold.links:
-                start, end, gold_kb = entity
+            for entity, kb_dict in gold.links.items():
+                start, end = entity
                 mention = doc.text[start:end]
+                for kb_id, value in kb_dict.items():
+                    entity_encoding = self.kb.get_vector(kb_id)
+                    prior_prob = self.kb.get_prior_prob(kb_id, mention)
+
+                    gold_ent = ents_by_offset[str(ent.start_char) + "_" + str(ent.end_char)]
```
ines:
```suggestion
                    gold_ent = ents_by_offset["{}_{}".format(ent.start_char, ent.end_char)]
```

spacy/pipeline/pipes.pyx
```diff
@@ -1150,49 +1152,40 @@ class EntityLinker(Pipe):
...
+                    assert gold_ent is not None
```
ines: As @honnibal already mentioned, we always want to avoid `assert`s in non-test code, because they're not very user-friendly. So this should just raise a somewhat descriptive error `if gold_ent is None`.

spacy/pipeline/pipes.pyx
```diff
@@ -1150,49 +1152,40 @@ class EntityLinker(Pipe):
...
 
-                    if kb_id == gold_kb:
-                        cats.append([1])
-                    else:
-                        nr_neg += 1
-                        cats.append([0])
-
         if len(entity_encodings) > 0:
-            assert len(priors) == len(entity_encodings) == len(context_docs) == len(cats) == len(type_vectors)
+            assert len(priors) == len(entity_encodings) == len(context_docs) == len(type_vectors)
```
ines: Todo: replace `assert` with explicit error.

spacy/pipeline/pipes.pyx
```diff
@@ -1203,39 +1196,44 @@ class EntityLinker(Pipe):
...
-        d_scores = (prediction - golds)
+    def get_loss(self, docs, golds, scores):
+        cats = []
+        for gold in golds:
+            for entity, kb_dict in gold.links.items():
+                for kb_id, value in kb_dict.items():
+                    cats.append([value])
+
+        cats = self.model.ops.asarray(cats, dtype="float32")
+        assert len(scores) == len(cats)
```
ines: Todo: replace `assert` with explicit error.

bin/wiki_entity_linking/training_set_creator.py
```diff
@@ -27,21 +32,23 @@ def _process_wikipedia_texts(wikipedia_input, wp_to_id, training_output, limit=N
...
     """
-    title_regex = re.compile(r'(?<=<title>).*(?=</title>)')
-    id_regex = re.compile(r'(?<=<id>)\d*(?=</id>)')
+    title_regex = re.compile(r"(?<=<title>).*(?=</title>)")
+    id_regex = re.compile(r"(?<=<id>)\d*(?=</id>)")
 
     read_ids = set()
     entityfile_loc = training_output / ENTITY_FILE
-    with open(entityfile_loc, mode="w", encoding='utf8') as entityfile:
+    with open(entityfile_loc, mode="w", encoding="utf8") as entityfile:
```
ines: Not a big deal, but we typically want to use pathlib for opening and validating files:

```suggestion
    with entityfile_loc.open("w", encoding="utf8") as entityfile:
```

spacy/pipeline/pipes.pyx
svlandeg: yea? In this case they should really be the same if the code is well-written, it doesn't depend on user input. So I thought `assert` would be more appropriate?

bin/wiki_entity_linking/training_set_creator.py
```diff
@@ -249,43 +279,57 @@ def _get_clean_wp_text(article_text):
...
-    while '  ' in clean_text:
-        clean_text = clean_text.replace('  ', ' ')
+    while "  " in clean_text:
+        clean_text = clean_text.replace("  ", " ")
 
     return clean_text.strip()
 
 
 def _write_training_article(article_id, clean_text, training_output):
     file_loc = training_output / str(article_id) + ".txt"
```
ines:
```suggestion
     file_loc = training_output / "{}.txt".format(article_id)
```

bin/wiki_entity_linking/training_set_creator.py
```diff
@@ -249,43 +279,57 @@ def _get_clean_wp_text(article_text):
...
-    with open(file_loc, mode='w', encoding='utf8') as outputfile:
+    with open(file_loc, mode="w", encoding="utf8") as outputfile:
```
ines:
```suggestion
    with file_loc.open("w", encoding="utf8") as outputfile:
```

bin/wiki_entity_linking/training_set_creator.py
```diff
@@ -296,58 +340,103 @@ def read_training(nlp, training_dir, dev, limit):
     skip_articles = set()
     total_entities = 0
 
-    with open(entityfile_loc, mode='r', encoding='utf8') as file:
+    with open(entityfile_loc, mode="r", encoding="utf8") as file:
```
ines:
```suggestion
    with entityfile_loc.open("r", encoding="utf8") as file:
```

bin/wiki_entity_linking/training_set_creator.py
```diff
@@ -296,58 +340,103 @@ def read_training(nlp, training_dir, dev, limit):
...
+                    dev == is_dev(article_id)
+                    and article_id != "article_id"
+                    and article_id not in skip_articles
+                ):
                     if not current_doc or (current_article_id != article_id):
                         # parse the new article text
                         file_name = article_id + ".txt"
                         try:
-                            with open(os.path.join(training_dir, file_name), mode="r", encoding='utf8') as f:
+                            with open(
```
ines: We should be able to kill all calls to `os.path` this way:

```python
training_file = training_dir / file_name
with training_file.open("r", encoding="utf8") as f:
    # etc.
```

spacy/pipeline/pipes.pyx
ines: Okay, I guess if we're 100% sure that this should never happen and it's really just an invariant, then yes. And if there's nothing the user can possibly put in that would trigger this, and not really a way to introduce a bug ourselves in some other piece of code that would have an effect. (There are a few places in the code base where we raise errors like "XY went wrong. This is likely a bug in spaCy.")

So if you think there's not really a risk here, we can leave the assert. Maybe just add a comment then so it's clear what it means if the assertion failed in this case.

honnibal:
> The Wikipedia pipeline actually does take frequency counts:
> Some example codes and unit tests set this `freq` field to a floating point value though, so I'd better change those as well to avoid confusion.

If we do want it to be frequencies, there might be a disadvantage to changing it to `int`, because we can't then accept smoothed counts (which would be floats).

I'm surprised frequencies are a good input there though? Surely we'd need to smooth and normalize them later to use them?

svlandeg:
Yea, I agree we want to normalize/smooth them. But the reason I was originally thinking of keeping the raw frequency counts, is the fact that you add entity per entity to the KB. So you may not know beforehand how many entities you'll end up adding, so it's also difficult to normalize up-front.

spacy/pipeline/pipes.pyx
svlandeg: I like the idea of specifically mentioning this might be a bug in spaCy - refactoring with custom error messages.

PR #4012, opened 2019-07-23
Remove old comment about using French parser in Norwegian

polm:
## Description

Norwegian used to borrow from French but that doesn't appear to have
been true for a while now, so the comment that was here is no longer
relevant.

### Types of change

comment fix

## Checklist


ines:
Thanks! 👍

Issue #2928, opened 2018-11-15
💫 Proposal: New JSON(L) format for training and improved training commands


ines:
### Motivation

One of the biggest invonveniences and sources of frustration is spaCy's current [JSON format for training](https://spacy.io/api/annotation#section-training). It's weirdly specific, annoying to create outside of the built-in converters and difficult to read. Training the model with incomplete information is pretty unintuitive and inconvenient as well.

To finally fix this, here's my proposal for a new and simplfied training file format that is easier to read, generate and compose.

### Example

```json
{
  "text": "Apple Inc. is an American multinational technology company headquartered in Cupertino, California. It was founded by Steve Jobs, Steve Wozniak, and Ronald Wayne in April 1976.",
  "ents": [
    {"start": 0, "end": 10, "label": "ORG"},
    {"start": 17, "end": 25, "label": "NORP"},
    {"start": 76, "end": 85, "label": "GPE"},
    {"start": 87, "end": 97, "label": "GPE"},
    {"start": 117, "end": 127, "label": "PERSON"},
    {"start": 129, "end": 142, "label": "PERSON"},
    {"start": 148, "end": 160, "label": "PERSON"},
    {"start": 164, "end": 174, "label": "DATE"}
  ],
  "sents": [
    {"start": 0, "end": 98},
    {"start": 99, "end": 175}
  ],
  "cats": {
    "TECHNOLOGY": true,
    "FINANCE": false,
    "LEGAL": false
  },
  "tokens": [
    {"start": 0, "end": 5, "pos": "PROPN", "tag": "NNP", "dep": "compound", "head": 1},
    {"start": 6, "end": 10, "pos": "PROPN", "tag": "NNP", "dep": "nsubj", "head": 2},
    {"start": 11, "end": 13, "pos": "VERB", "tag": "VBZ", "dep": "ROOT", "head": 2},
    {"start": 14, "end": 16, "pos": "DET", "tag": "DT", "dep": "det", "head": 7},
    {"start": 17, "end": 25, "pos": "ADJ", "tag": "JJ", "dep": "amod", "head": 7},
    {"start": 26, "end": 39, "pos": "ADJ", "tag": "JJ", "dep": "amod", "head": 6},
    {"start": 40, "end": 50, "pos": "NOUN", "tag": "NN", "dep": "compound", "head": 7},
    {"start": 51, "end": 58, "pos": "NOUN", "tag": "NN", "dep": "attr", "head": 2},
[...]

PR #4013, opened 2019-07-23
Fix gold.docs_to_json() to match JSON train format

adrianeboyd:
The output of `gold.docs_to_json()` was missing the outer list brackets and
couldn't be loaded by GoldCorpus.

* added brackets
* added roundtrip test of `gold.docs_to_json()` / GoldCorpus

Also, `docs_to_json()` puts each doc in its own paragraph rather than its own document. Is this intentional?

### DIFF
--- a/spacy/tests/test_gold.py
+++ b/spacy/tests/test_gold.py
@@ -7,6 +7,7 @@ from spacy.gold import spans_from_biluo_tags, GoldParse
 from spacy.gold import GoldCorpus, docs_to_json
 from spacy.lang.en import English
 from spacy.tokens import Doc
+from .util import make_tempdir
 
 
 def test_gold_biluo_U(en_vocab):
@@ -81,12 +82,12 @@ def test_docs_to_json_roundtrip(tmpdir):
     for i in range(1, len(doc)):
         doc[i].is_sent_start = False
 
-    json_str = srsly.json_dumps(docs_to_json(doc))
-    json_file = tmpdir.mkdir("docs_to_json_roundtrip").join("flew.json")
-    with open(json_file, "w") as fileh:
-        fileh.write(json_str)
+    with make_tempdir() as tmpdir:
+        json_file = tmpdir / "flew.json"
+        srsly.write_json(json_file, docs_to_json(doc))
 
-    goldcorpus = GoldCorpus(str(json_file), str(json_file))
+        goldcorpus = GoldCorpus(json_file, json_file)
+
     (reloaded_doc, goldparse) = next(goldcorpus.train_docs(nlp))
 
     assert goldcorpus.count_train() == len(doc)
--- a/spacy/tests/test_gold.py
+++ b/spacy/tests/test_gold.py
@@ -1,8 +1,11 @@
 # coding: utf-8
 from __future__ import unicode_literals
 
+import srsly
 from spacy.gold import biluo_tags_from_offsets, offsets_from_biluo_tags
 from spacy.gold import spans_from_biluo_tags, GoldParse
+from spacy.gold import GoldCorpus, docs_to_json
+from spacy.lang.en import English
 from spacy.tokens import Doc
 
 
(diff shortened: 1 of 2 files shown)


spacy/tests/test_gold.py
```diff
@@ -68,3 +71,23 @@ def test_gold_ner_missing_tags(en_tokenizer):
...
+    text = "I flew to Silicon Valley via London."
+    nlp = English()
+    doc = nlp(text)
+    doc[0].is_sent_start = True
+    for i in range(1, len(doc)):
+        doc[i].is_sent_start = False
+
+    json_str = srsly.json_dumps(docs_to_json(doc))
+    json_file = tmpdir.mkdir("docs_to_json_roundtrip").join("flew.json")
+    with open(json_file, "w") as fileh:
```
ines: I think the tempdir stuff is what makes the tests fail on some platforms. We have a `tests.util.make_tempdir` contextmanager that should hopefully help with this and make the test more consistent.

adrianeboyd: Thanks, I didn't think to check that. I rewrote it to use make_tempdir and look more like some of the existing tests that write temp files.

ines: Thanks, looks like this did the trick! And no worries – most of that stuff is super arbitrary.

honnibal:
The training data does produce a `(Doc, GoldParse)` pair for each paragraph of text, rather than the whole "document".

I know the situation with the training data format is very confusing, so I want to explain a little bit about how we ended up here. Mostly what people do for parsing and NER is process the sentences individually. I wanted to avoid that, and instead work on larger units of text. Sentence boundary detection is often non-trivial, and I wanted to allow cross-sentence features for NER and potentially later coreference resolution.

But once you go above the sentence, what unit of text should we actually be working on? A document can be extremely long, e.g. a whole book. I figured paragraphs would be a good compromise: a paragraph isn't a real linguistic unit, but we can basically treat it as whatever unit it's easy to segment the text into.

However, if we're splitting up the document into these "paragraphs", I figured we'd still want to retain the information of how the paragraphs actually belong together into documents. There are a couple of reasons for this. First, coreference resolution is obviously at the document level, so if we want to do that task, we'd obviously need the whole document. The OntoNotes data I was working with when I came up with this scheme has coreference annotations, so this seemed very relevant. Second, if we're making train/dev/test splits, we really want to make sure the splits happen at document boundaries. We don't want one paragraph of a document to be in the training data, and another paragraph to be in the testing data. Incorrect splits like that lead to incorrect measurements, e.g. you'll have unrealistically few unseen words and entities in the test data, and generally higher performance than you should have.

The `docs_to_json` function accepts a list of `doc` objects that represent the paragraphs of a single logical document, and produces the json structure for that logical document. The files that the `GoldCorpus` read expect the outer-level to be a list of those documents, usually organised into a section of the corpus. That's why the outer list seems to be "missing".  You can see an example of the function's usage here: https://github.com/explosion/spaCy/blob/master/spacy/cli/converters/jsonl2json.py#L10

I'm not happy with the current training data format solution, and I find this all confusing too. I think when we were talking about this earlier, I even gave you some wrong information (e.g. about the paragraphs) -- because I also find it hard to keep all this straight.

The problem is, I still don't know what we should actually do to improve it. I've come back to this problem a lot over the years, and even when I was developing the library I hated the training format. But it replaced several previous efforts that I hated even more, because once the data loses some information, it's pretty hard to realign things to reinsert it.

I'm hoping you have some ideas? Other discussion on the training data format is here: https://github.com/explosion/spaCy/issues/2928

adrianeboyd:
Yeah, I wondered if there was reason for it to be structured as paragraphs in `docs_to_json()`. I think someone working just from the information in the API is going to be pretty confused that the output of `docs_to_json()` isn't actually the right format, though. (See: me, yesterday.)

I could see that modifying `docs_to_json()` didn't break anything in spacy.gold, but I missed that cli example when I was looking to see how it was used.

Let me put some thoughts in #2928.

honnibal:
I guess a lot of the evil here has come from trying to "future proof" the format to account for things we've never handled (like coreference). Maybe we should just:

* Make it jsonl, where the lines are what we currently have as the "paragraph" elements
* Add a field for `parent_id` to record that some of these elements belong together

adrianeboyd:
No, I think future-proofing is a good idea. I also think being able to reconstruct the original document is a good idea. (I don't think `parent_id` would be enough.) I do think having "paragraph" as a significant unit (but no other units?) in terms of organizing the annotation is a little odd and is going to cause problems.

adrianeboyd:
And I forgot to add the reason why I started this: I want to add doc.cats to this exported data. If docs are docs it's a lot more straightforward than if docs are paragraphs.

honnibal:
Closing, as I think we want to keep the current structure, where one `Doc` object is a paragraph.

Issue #4020, opened 2019-07-25
gold.biluo_tags_from_offsets should throw E103 if two entity labels are overlapping


RyanZHe:
## Feature description
`spacy.gold.biluo_tags_from_offsets` should throw `Error.E103` when the collection of entites passed in have overlapping tokens. 

For example,
```
from spacy.gold import biluo_tags_from_offsets

doc = nlp(u"I like California Pizza Kitchen.")
entities = [(7, 17, "LOC"), (7, 31, "BRAND") ]
tags = biluo_tags_from_offsets(doc, entities) # should throw Error.E103
```
Right now `tags` is being assigned as `["O", "O", "U-LOC", "I-BRAND", "L-BRAND"]`, which violates the biluo tagging schema.

pratapaprasanna:
I agree that an error has to be raised while forming tags only and not while giving input to gold corpus . It now throws error while feeding in data in goldcorpus. which I think a bad way.
-
Thanks

PR #4021, opened 2019-07-25
biluo_tags_from_offsets throw exception for overlapping entities

RyanZHe:
## Description

#4020 
biluo_tags_from_offsets used to return a malformed biluo tag sequence even if the passed in entites parameters have overlaps.

### Types of change

- biluo_tags_from_offsets now checks whether passed in entities have overlaps and throw E107 if they do
- also added unit tests for the exception

## Checklist

### DIFF
--- a/.github/contributors/ryanzhe.md
+++ b/.github/contributors/ryanzhe.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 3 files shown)


ines:
Thanks for the fix! Just FYI, we're not merging this yet for the next patch release, since it could potentially impact people's existing workflows and scripts. Raising the error is really good and definitely something we want – but we want to wait with releasing it until 2.2.

PR #4022, opened 2019-07-25
Fix error when Korean text contains regexp special characters.

b1uec0in:
## Description
* Fix error when Korean text contains regexp special characters (ex: ?, $).

### Types of change
Bug fix

## Checklist


ines:
Thanks! 👍

PR #4023, opened 2019-07-25
Update GoldParse attributes in API docs

adrianeboyd:
## Description

* add `words`
* update name of entity list to `ner`

I think it might be a bit more consistent to have `ner` named `entities` or `ents` (and `ents` is actually set somewhere to `None`, which is a bit confusing), but it looks like renaming it would be a non-trivial decision.

### Types of change

Documentation.

## Checklist


ines:
Thanks! 👍

PR #4062, opened 2019-08-01
Corrected imported function

ejarkm:
The example showed an incorrected import

## Description
Only on the documentation, the example showed an incorrect import


ines:
Thanks for fixing this! 👍

PR #4066, opened 2019-08-01
added missing punctuation following conventions.

mirfan899:
### Types of change
Added punction `۔` in proper place following conventions.

## Checklist


spacy/lang/char_classes.py
```diff
@@ -214,7 +214,7 @@
 # used in Chinese (see #1333, #1340, #1351) – unless there are cross-language
 # conflicts, spaCy's base tokenizer should handle all of those by default
 _punct = (
-    r"… …… , : ; \! \? ¿ ؟ ¡ \( \) \[ \] \{ \} < > _ # \* & 。 ？ ！ ， 、 ； ： ～ · । ، ؛ ٪"
+    r"… …… , : ; \! \? ¿ ؟ ¡ \( \) \[ \] \{ \} < > _ # \* & 。 ？ ！ ， 、 ； ： ～ · । ، ۔ ؛ ٪"
```
ines: I wonder if it makes sense to write this one as `\u06d4`? Because it looks so visually similar to `.`. When I first saw it, I was like, wait, is that a regular period? 😅 And then I checked the unicode representation.

(This is really interesting btw, I had no idea that character was used as well.)

mirfan899: It's a full stop in Urdu. Yes, it seems more like a regular full stop but actually is a bit lengthy.

ines: Okay, let's try this then?

```suggestion
    r"… …… , : ; \! \? ¿ ؟ ¡ \( \) \[ \] \{ \} < > _ # \* & 。 ？ ！ ， 、 ； ： ～ · । ، \u06d4 ؛ ٪"
```

mirfan899: I've checked it and tested it properly. Some punctuations I've already added are `،` and `؛` so we do not need Unicode for character (`\u06d4`).  I believe `۔` is better than `Unicode` representation.

Issue #4042, opened 2019-07-29
Adding EntityRuler before ner and saving model to disk crashes loading the model


psychosis448:
## How to reproduce the behaviour
I got a strange problem when loading a modified model, adding an `EnityRuler` to the pipeline with the `before='ner'` flag.

So I create my patterns, add them to my entity_ruler and the ruler to my pipe:

```python
nlp = spacy.load('de_core_news_sm')
ruler = EntityRuler(nlp)
ruler.add_patterns(patterns)
nlp.add_pipe(ruler, before='ner')
```

Then I check for the dir, create it if necessary and save the model to disk:

```python
output_dir = Path('custom_model')
if not output_dir.exists():
  output_dir.mkdir()
nlp.to_disk(output_dir)
```

When I now load the model with

```ptyhon
spacy.load(Path('custom_model'))
```

It throws the following error:

```bash
ValueError: [E109] Model for component 'ner' not initialized.
Did you forget to load a model, or forget to call begin_training()?
```

There is no problem doing this without `before='ner'`...

Extended Error:

```bash
Traceback (most recent call last):
  File "/spacy/__init__.py", line 27, in load
    return util.load_model(name, **overrides)
  File "/spacy/util.py", line 135, in load_model
    return load_model_from_path(name, **overrides)
  File "/spacy/util.py", line 173, in load_model_from_path
    return nlp.from_disk(model_path)
  File "/spacy/language.py", line 791, in from_disk
    util.from_disk(path, deserializers, exclude)
  File "/spacy/util.py", line 630, in from_disk
    reader(path / key)
  File "/spacy/language.py", line 787, in <lambda>
    deserializers[name] = lambda p, proc=proc: proc.from_disk(p, exclude=["vocab"])
  File "/spacy/pipeline/entityruler.py", line 183, in from_disk
    self.add_patterns(patterns)
  File "/spacy/pipeline/entityruler.py", line 138, in add_patterns
    self.phrase_patterns[label].append(self.nlp(pattern))
  File "/spacy/language.py", line 390, in __call__
    doc = proc(doc, **component_cfg.get(name, {}))
  File "nn_parser.pyx", line 205, in spacy.syntax.nn_parser.Parser.__call__
  File "nn_parser.pyx", line 238, in spacy.syntax.nn_parser.Parser.predict
  File "nn_parser.pyx", line 235, in spacy.syntax.nn_parser.Parser.require_model
ValueError: [E109] Model for component 'ner' not initialized.
[...]

PR #4072, opened 2019-08-02
WIP: attempt to fix nlp deserialization 

svlandeg:
Trying to fix Issue #4042 by creating and deserializing the components one by one.

Does not work yet...

## Description
* Rather than creating all (empty) pipe components first, and then adding all deserializers, this PR creates and deserializes the components one by one, to avoid using `nlp()` with empty components further down in the pipeline.

### Types of change
Bug fix

## Checklist


honnibal:
Closing this, as I think the idea of running the `nlp` object within the `PhraseMatcher` really needs to be rethought. It's something that we can otherwise design around, and it causes way too many problems.

PR #4075, opened 2019-08-03
WIP: attempt 2 to fix nlp deserialization

svlandeg:
Trying to fix Issue #4042 (alternative) by disabling the subsequent `nlp` pipeline components in the `EntityRuler`.

Does not work yet...

## Description
* In the EntityRuler, when calling `nlp()`, first disable all pipes that are coming after it by checking `self.nlp.pipe_names.index(self.name)`

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/pipeline/entityruler.py
+++ b/spacy/pipeline/entityruler.py
@@ -176,21 +176,28 @@ class EntityRuler(object):
 
         DOCS: https://spacy.io/api/entityruler#add_patterns
         """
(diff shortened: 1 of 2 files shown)


honnibal:
Closing this, as I think the idea of running the `nlp` object within the `PhraseMatcher` really needs to be rethought. It's something that we can otherwise design around, and it causes way too many problems.

Issue #4074, opened 2019-08-03
💫 Raise error if annotation dict in simple training style has unexpected keys


ines:
This is a really easy mistake to make and can lead to very frustrating debugging experiences. It just happened to me again and I've debugged people's training code in the past where this issue was the root cause of the model not learning anything.

Consider the following example:

```python
texts = ["hello world", "this is a text"]
cats = [{"LABEL": True}, {"LABEL": False}]
nlp.update(texts, cats)
```

Spot the problem? It's here:

```diff
texts = ["hello world", "this is a text"]
- cats = [{"LABEL": True}, {"LABEL": False}]
+ cats = [{"cats": {"LABEL": True}}, {"cats": {"LABEL": False}}]
nlp.update(texts, cats)
```

Requiring a dict like that makes sense because it allows you to train multiple things at once (entities and text categories for instance). However, we currently do not raise an error if no expected top-level keys are present and instead just quietly ignore the additional keys. Which means that in the first example, the model would have been updated with *nothing*.

Instead, spaCy should raise an error like "Trying to call nlp.update with annotation type 'LABEL'. Expected top-level keys 'words', 'tags', 'heads', 'deps', 'entities' or 'cats'. Got: {"LABEL": True}."

jenojp:
Hi @ines , I took a stab at this one. I put in a check that checks to see if any of the keys are in the list of expected top-level keys and raises an error if not. So this ensures:

annots = [{"LABEL": True}, {"LABEL": False}] **does not work**
annots = [{"cats": {"LABEL": True}}, {"cats": {"LABEL": False}}] **works**
annots = [{"cats": {"LABEL": True}, "other":{"POSITIVE": 1.0}}, {"cats": {"LABEL": False}, "other":{"POSITIVE": 1.0}}] **also works though**

Is this the behavior you were expecting? I can go ahead submit a pull request if so.

ines:
@jenojp Nice, thanks a lot for taking this on so quickly! 👍 The fix looks good, so yes, please go ahead and submit the PR.

> annots = [{"cats": {"LABEL": True}, "other":{"POSITIVE": 1.0}}, {"cats": {"LABEL": False}, "other":{"POSITIVE": 1.0}}] also works though

I think we could maybe even be stricter here and check for all keys that are not in our list of expected keys. And then raise an error if any unexpected values are found. For example, something like this:

```python
expected_keys = ("words", "tags", "heads", "deps", "entities", "cats")
[...]

PR #4079, opened 2019-08-04
Raise error if annotation dict in simple training style has unexpected keys #4074

jenojp:
# Raise error if annotation dict in simple training style has unexpected keys #4074

## Description
Added strict check to see if top level keys in simple annotation dictionary are in list of expected keys and if not, raise ValueError.

### Types of change
Ehancement from issue #4074 

## Checklist


spacy/language.py
```diff
@@ -445,6 +445,10 @@ def update(self, docs, golds, drop=0.0, sgd=None, losses=None, component_cfg=Non
             if isinstance(doc, basestring_):
                 doc = self.make_doc(doc)
             if not isinstance(gold, GoldParse):
+                expected_keys = ("words", "tags", "heads", "deps", "entities", "cats")
+                unexpected_keys = [k for k in gold if k not in expected_keys]
+                if unexpected_keys:
+                    raise ValueError(Errors.E151.format(unexpected_keys=unexpected_keys))
```
svlandeg: Lets add "links" to this list to ensure the upcoming entity linking functionality is already accounted for :-)

spacy/errors.py
```diff
@@ -417,6 +417,10 @@ class Errors(object):
             "component matches the model being loaded.")
     E150 = ("The language of the `nlp` object and the `vocab` should be the same, "
             "but found '{nlp}' and '{vocab}' respectively.")
+    E151 = ("Trying to call nlp.update without required annotation types. "
+            "Expected top-level keys 'words', 'tags', 'heads', 'deps', 'entities' or 'cats'."
+            " Got: {unexpected_keys}.")
+
```
svlandeg: Could we somehow make sure to define this list only once? Because also here we'd want to include "links" (cf other comment)

ines: Yes, that's a good idea! We can just pass in the `expected_keys` here.

jenojp: Good ideas! I put both of these in the latest commit

svlandeg: Looks good!

honnibal:
Thanks for your help! Looks good.

PR #4090, opened 2019-08-06
💫 Require downloaded model in pkg_resources so it can be used in same session

### LINKED ISSUE #3923: Trying to load a model in the same script that downloaded it raises OSError
Hi,

When downloading a model using `spacy.cli.download` and then directly using `spacy.load` (in the same interpreter), it raises an OSError because `spacy.utils.is_package` returns False: https://github.com/explosion/spaCy/blob/a7fd42d937d405088803da3fb3b301cc81cea719/spacy/util.py#L133-L134 
We currently need to start a new interpreter for the list of packages to be updated with the newly installed model.
Is this the expected behaviour?
Is there a way to download a model at run time and using it directly? (I'd like my package to download the model lazily only when needed and without having to execute the script twice).

Thanks,

```bash
$ pip uninstall -y en_core_web_sm; python -c "import spacy; spacy.cli.download('en_core_web_sm'); spacy.load('en_core_web_sm')"

OSError: [E050] Can't find model 'en_core_web_sm'. It doesn't seem to be a shortcut link, a Python package or a valid path to a data directory.
```

## Your Environment

* Operating System: Ubuntu 18.04
* Python Version Used: 3.7.3
* spaCy Version Used: 2.1.3
* Environment Information:

ines:
Resolves #3923.

## Description

This PR attempts to resolve an inconvenience of the `download` command. If a model is downloaded and then loaded within the same process, our `is_package` check currently fails, because `pkg_resources.working_set` is not refreshed automatically (see #3923). This can commonly happen in Jupyter notebooks if you do something like:

```
In [1]: !python -m spacy download de_core_news_sm
In [2]: import spacy
        nlp = spacy.load("de_core_news_sm")
```

To work around this, we can try and require the newly installed package explicitly using `pkg_resources.working_set.require`. 

The following test case works without the code proposed in this PR (if you want to try it, make sure the model isn't installed):

```python
from spacy.cli import download
from spacy.util import is_package
import pkg_resources

model = "de_core_news_sm"

download(model)
print(is_package(model))  # False
pkg_resources.working_set.require(model)
print(is_package(model))  # True
```

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/cli/download.py
+++ b/spacy/cli/download.py
@@ -6,6 +6,7 @@ import requests
 import os
 import subprocess
 import sys
+import pkg_resources
 from wasabi import Printer
 
 from .link import link
@@ -67,6 +68,16 @@ def download(model, direct=False, *pip_args):
                     "the model via its full package name: "
                     "nlp = spacy.load('{}')".format(model, model_name),
                 )
+        # If a model is downloaded and then loaded within the same process, our
+        # is_package check currently fails, because pkg_resources.working_set
+        # is not refreshed automatically (see #3923). We're trying to work
+        # around this here be requiring the package explicitly.
+        try:
+            pkg_resources.working_set.require(model_name)
+        except:  # noqa: E722
+            # Maybe it's possible to remove this – mostly worried about cross-
+            # platform and cross-Python copmpatibility here
+            pass
 
 
 def get_json(url, desc):


honnibal:
Obviously the naked `except` is unideal, but I agree that it's better to "first, do no harm" here. In the case that it errors, we'll just get the same results we have currently.

ines:
Yeah, I just don't want to risk breaking stuff. It should all work as expected, but there might be some edge case where it fails and in that case, the behaviour will just be the same as before.

> Perhaps we could warn though? Hard to write a warning that's not puzzling, since it's such a Python-packaging-minutiae thing.

We don't know whether the user cares about it, though. If you're running the script from the command line (probably still the most common use case), this part is completely irrelevant. It only matters if you want to load that model in the same session (e.g. if you're calling `download()` in a script or if you're doing `!python -m spacy download` in a notebook).

Issue #3959, opened 2019-07-12
Modified pos attribute not being serialized


aabbi:
## How to reproduce the behaviour
Since `pos` attribute of tokens is now writable (related [comment](https://github.com/explosion/spaCy/issues/2014#issuecomment-376642763) ), I ran into the issue that I changed the pos tag and serialized the doc but on deserializing, it retained the original pos tag. 
I believe this is happening because only the `tag` attributes are pickled

```
import spacy
nlp = spacy.load('en')
d = nlp("displaCy uses JavaScript, SVG and CSS to show you how computers understand language")
print(d[0].pos_) # prints ADJ
d[0].pos_ = "NOUN"
d.to_disk("t.txt")

d2 = nlp("")
d2.from_disk("t.txt")
print(d2[0].pos_) # prints ADJ
```
Wasn't entirely sure if this should be a bug or a feature request to make pos tags picklable or maybe documentation could be added specifying how serializing would not save changes in coarse grained attributes with pos tag as examples

## Your Environment
* **spaCy version:** 2.1.3
* **Platform:** Ubuntu-14.04-trusty
* **Python version:** 3.6.8
* **Models:** en_core_web_lg, en, en_core_web_md

svlandeg:
Thanks for the report, good catch! Fixed in PR #4092.

PR #4092, opened 2019-08-06
Serialize POS attribute when doc.is_tagged

svlandeg:
## Description
Bugfix for Issue #3959 : Ensure `pos` attributes are serialized (and not just `tags`).
The current fix only works when `doc.is_tagged` is `True`, but I think that makes sense.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tests/regression/test_issue3959.py
+++ b/spacy/tests/regression/test_issue3959.py
@@ -1,0 +1,29 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+from spacy.lang.en import English
+from ..util import make_tempdir
+
+
+def test_issue3959():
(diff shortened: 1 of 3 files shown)


honnibal:
It's kind of disappointing to serialize the almost-redundant field, but I guess we have to. I hope it at least compresses okay?

I'll hold off on merging this for the patch release, since it seems bad to make a modification to the serialisation in 2.1.7. The next release would probably be 2.2.0, which seems like a better time to do this.

svlandeg:
I know what you mean. I guess the other option is to only serialize those cases where the `pos` is "conflicting" with the `tag` annotation? It would be less clean in code than the current "brute force" solution, but would ensure we only store data we really have to?

Issue #3540, opened 2019-04-04
Problem with vector attribute after calling Retokenizer.split


klti:
When trying to use split and retokenize like in the docs (https://spacy.io/usage/linguistic-features/#retokenization), `displacy.render(list(doc.sents))` breaks with an error

## How to reproduce the behaviour

code to reproduce (nearly the same as example in docs, only made strings unicode and added displacy.render calls)
```
import spacy
from spacy import displacy

nlp = spacy.load("en_core_web_sm")
doc = nlp(u"I live in NewYork right now")
print("Before:", [token.text for token in doc])
displacy.render(doc)  # works
displacy.render(doc, style="dep")  # works
displacy.render(list(doc.sents), style="dep")  # works

with doc.retokenize() as retokenizer:
    heads = [(doc[3], 1), doc[2]]
    attrs = {"POS": ["PROPN", "PROPN"], "DEP": ["pobj", "compound"]}
    retokenizer.split(doc[3], [u"New", u"York"], heads=heads, attrs=attrs)
print("After:", [token.text for token in doc])
displacy.render(doc)  # works
displacy.render(doc, style="dep")  # works
displacy.render(list(doc.sents), style="dep")  # breaks
```

Output
```
('Before:', [u'I', u'live', u'in', u'NewYork', u'right', u'now'])
('After:', [u'I', u'live', u'in', u'New', u'York', u'right', u'now'])
Traceback (most recent call last):
  File "test_example.py", line 18, in <module>
    displacy.render(list(doc.sents), style="dep") # breaks
  File "/usr/lib/python2.7/site-packages/spacy/displacy/__init__.py", line 46, in render
    docs = [obj if not isinstance(obj, Span) else obj.as_doc() for obj in docs]
  File "span.pyx", line 224, in spacy.tokens.span.Span.as_doc
  File "span.pyx", line 410, in spacy.tokens.span.Span.vector.__get__
  File "span.pyx", line 187, in __iter__
  File "token.pyx", line 392, in spacy.tokens.token.Token.vector.__get__
IndexError: index 6 is out of bounds for axis 0 with size 6
```

## Your Environment

* **Python version:** 2.7.16
* **Platform:** Linux-5.0.3-arch1-1-ARCH-x86_64-with-glibc2.2.5
* **spaCy version:** 2.1.3

P.S.: great library, I love it :thumbsup:

ines:
[...]

PR #4097, opened 2019-08-08
Update lemma and vector information after splitting a token

svlandeg:
## Description
This PR fixes two related bugs that happened after running `retokenizer.split`:
* The `lemma` information was not updated and was still referring to the old `token.text` attributes. This is fixed/reset by calling `token.lemma = 0` on the split tokens.

* The `vector` attributes were left unchanged, causing an out-of-bounds error as described in Issue #3540. This is fixed by extending the `doc.tensor` and setting the vectors of the split tokens to an array of zeros (not sure what else to do).

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tests/regression/test_issue3540.py
+++ b/spacy/tests/regression/test_issue3540.py
@@ -1,14 +1,22 @@
-import spacy
+# coding: utf8
+from __future__ import unicode_literals
+
+from spacy.tokens import Doc
+
+import numpy as np
 
 
-def test_issue3540():
-    nlp = spacy.load("en_core_web_sm")
-    doc = nlp(u"I live in NewYork right now")
+def test_issue3540(en_vocab):
+
+    words = ["I", "live", "in", "NewYork", "right", "now"]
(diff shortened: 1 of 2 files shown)


ines:
Hmm, looks like the build failure is a problem with Azure Pipelines:

> ##[Error 1]
> The agent: Hosted Agent lost communication with the server. Verify the machine is running and has a healthy network connection. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610 

I'll restart! 🤞

ines:
Hmmm, now this random matcher (?) test keeps failing again randomly on only one configuration... wtf... 🤔

svlandeg:
Where do you see the failing test? I only still see the "Hosted Agent lost communication" error.

ines:
@svlandeg See here for all builds: https://dev.azure.com/explosion-ai/Public/_build?definitionId=8&_a=summary

Restarted the test again and it now passed 🤷‍♀️

svlandeg:
Weird 😬
I can push a dummy commit and see what happens? On one of the other PR's the other day I had to do the same because the PR wasn't updating itself properly...

honnibal:
Patch looks good, thanks!

ines:
I'll just go ahead and merge this! There really seems to be nothing wrong with this PR – there's likely *something* wrong  *somewhere*, though, that makes the tests flaky like this.

Issue #4100, opened 2019-08-09
PhraseMatcher does not match LEMMA


fizban99:
According to the documentation, the [PhraseMatcher](https://spacy.io/api/phrasematcher) allows setting a different attribute other than `ORTH`. It explicitely mentions `LOWER`, `POS` and `DEP`, but leaves the door open for other attributes. 
The `LEMMA` attribute does not seem to work with the English model. This was mentioned in the gitter chat room.

## How to reproduce the behaviour
If we take the exact example in the [usage guide](https://spacy.io/usage/rule-based-matching#phrasematcher) and just add `attr="LEMMA"` to the constructor, it does not find any match:
```
import spacy
from spacy.matcher import PhraseMatcher
​
nlp = spacy.load('en_core_web_sm')
matcher = PhraseMatcher(nlp.vocab, attr="LEMMA")
terms = [u"Barack Obama", u"Angela Merkel", u"Washington, D.C."]
# Only run nlp.make_doc to speed things up
patterns = [nlp.make_doc(text) for text in terms]
matcher.add("TerminologyList", None, *patterns)
​
doc = nlp(u"German Chancellor Angela Merkel and US President Barack Obama "
          u"converse in the Oval Office inside the White House in Washington, D.C.")
matches = matcher(doc)
for match_id, start, end in matches:
    span = doc[start:end]
    print(span.text)
```
we can verify that in this case the lemmas and the text should be the same for the matching terms and I would expect to get the same results with `ORTH` or with `LEMMA` in this specific example:
```
for tok in doc:
    print(tok.lemma_, tok.text)
```
If instead of the en_core_web_sm model we use an empty English model:
```
from spacy.lang.en import English
nlp = English()
```
We get that the terms match everything, which is also incorrect.

## Your Environment
* **spaCy version:** 2.1.8
* **Platform:** Linux-4.15.0-52-generic-x86_64-with-debian-buster-sid
* **Python version:** 3.7.3

svlandeg:
Hi @fizban99, thanks for the report! That does sound suspicious - let me look into it ;-)

adrianeboyd:
The problem is that `make_doc()` doesn't set lemmas. You can look up the lookup lemmas with `token.lemma_` from the returned `doc` but they're not actually set. It's a pretty confusing situation for users, who can't tell that the `lemma` isn't set.

[...]

Issue #4070, opened 2019-08-02
Improve handling of unsupported Matcher attributes


adrianeboyd:
## Feature description

I decided to separate this from #4069 (see also #4063 ) since it's not quite the same issue.

With Matcher, it's very confusing that "bad" parts of patterns (e.g., with attributes that aren't supported) are silently discarded and end up matching every token rather than no token.

`{'ASDF': True}` shouldn't be equivalent to `{}`.

How slow is validation? Would it make sense to make `validate=True` the default? Are there simpler checks that could be added for when attributes are discarded that don't require full validation so that you could provide warnings or errors in these cases?

I think that less confusing default behavior would be:

* `{'ASDF': True}` matches nothing
* there are always warnings or errors when it is known that patterns will match nothing due to this kind of problem

ines:
Yes, I agree! 👍

> How slow is validation? Would it make sense to make validate=True the default? Are there simpler checks that could be added for when attributes are discarded that don't require full validation so that you could provide warnings or errors in these cases?

Yes, we should probably just consider implementing simpler checks for just the key names and whether they map to a valid attribute or not. 

The current validation uses `jsonschema`, which we've made an optional dependency (we didn't want it to drag in too many other dependencies down the tree). JSON schema validation is nice because it lets you write pretty complex constraints and gives you helpful and detailed error messages. It's also cross-Python compatible. (Once we drop 2.7 and 3.5, we could probably implement something similar using native Python type hints and make that enabled by default.)

Issue #4063, opened 2019-08-01
Matcher and EntityRuler token patterns should use same language and use original Token attribute names


johann-petrak:
This is one of the most puzzling and not very well documented details about Spacy:

* the Matcher apparently only allows to match a subset of token attributes and requires that instead of the original names of those attributes, special all-uppercase identifiers have to be used. Why? Why not allow to use all Token attributes and by their original name?
* The entity ruler examples seem to indicate that lower case, original name attributes can be used. But the documentation does not really say much about this or if other features of the matcher like value ranges, operators and quantifiers can be used.

Ideally the token matching in the entity ruler would work identically to the matcher and both would allow to use all Token attributes with their original name.

BreakBB:
I already had a look at the code about this and the EntityRuler internally creates a Matcher (see [here](https://github.com/explosion/spaCy/blob/2cdf7d39e777b0b7f9a822be376ec8aef5cdacf6/spacy/pipeline/entityruler.py#L53)) and possibly a PhraseMatcher (see [here](https://github.com/explosion/spaCy/blob/2cdf7d39e777b0b7f9a822be376ec8aef5cdacf6/spacy/pipeline/entityruler.py#L56)) as well.

Moreover the `add_patterns` method of the EntityRuler delegates the pattern to those matchers (see [here](https://github.com/explosion/spaCy/blob/2cdf7d39e777b0b7f9a822be376ec8aef5cdacf6/spacy/pipeline/entityruler.py#L190)). So from my understanding the pattern format should be identical and the examples and docs should be updated.

ines:
Sorry if this was confusing. The `EntityRuler` doesn't do anything special – its token-based patterns are passed directly to the `Matcher`. So the attributes they support are the same.

The names being uppercase was more of a convention that was introduced early on to make them resemble the [symbol IDs](https://github.com/explosion/spaCy/blob/master/spacy/attrs.pyx) and make them stand out a bit more as "special" in Python. But spaCy normalises them internally, so if you try it, you should see that capitalisation doesn't matter.

To make this more clear in the docs, we could add the following:
* A note [here](https://spacy.io/usage/rule-based-matching#entityruler-patterns) mentioning that "token patterns" == `Matcher` patterns and "phrase patterns" == `PhraseMatcher` patterns.
* A note that capitalisation doesn't matter.
* We might as well make the attributes in the entity ruler patterns use uppercase attributes for consistency.

johann-petrak:
Thank you! 

[...]

PR #4105, opened 2019-08-12
 Improve token pattern checking without validation 

adrianeboyd:
## Description

Add more detailed token pattern checks without full JSON pattern validation and
provide more detailed error messages.

Addresses #4070 (also related: #4063, #4100).

* Check whether top-level attributes in patterns and attr for PhraseMatcher are
  in token pattern schema

* Check whether attribute value types are supported in general (as opposed to
  per attribute with full validation)

* Report various internal error types (OverflowError, AttributeError, KeyError)
  as ValueError with standard error messages

* Check for tagger/parser in PhraseMatcher pipeline for attributes TAG, POS,
  LEMMA, and DEP

* Add error messages with relevant details on how to use validate=True or nlp()
  instead of nlp.make_doc()

* Support attr=TEXT for PhraseMatcher

* Add NORM to schema

* Expand tests for pattern validation, Matcher, PhraseMatcher, and EntityRuler

### Types of change

Some combination of bugfix and enhancement?

## Checklist


adrianeboyd:
Concerns:

* `if attr not in TOKEN_PATTERN_SCHEMA["items"]["properties"].keys():` is potentially too hacky

* This will change the behavior for some users' existing faulty patterns that were silently discarding unsupported attributes

ines:
> `if attr not in TOKEN_PATTERN_SCHEMA["items"]["properties"].keys()`: is potentially too hacky

Yeah, I see what you mean – but on the other hand, it only really refers to the standard JSON schmea structure, and a data structure that we control. It will not have to deal with any user input.

The `.keys()` should be redundant here btw so that'd at least make it slightly shorter.

> This will change the behavior for some users' existing faulty patterns that were silently discarding unsupported attributes

Yes, we shouldn't ship this in a patch release and only in v2.2. There are a few other PRs pending for the exact same reason.

adrianeboyd:
Ah, I fixed the `.keys()` bits.

It seemed simpler to keep that check up-to-date with the structure than to extract a list, but it's still a bit hacky. I don't know how much the JSON schemas change across versions? But again, it's only 2 lines to keep in sync.

ines:
> I don't know how much the JSON schemas change across versions? But again, it's only 2 lines to keep in sync.

I'm no JSON schemas expert, but I don't think the structure is going to change. And I think it's fine to use the schema as the "source of truth" for the matcher attributes – we just need to remember to update it if we ever change the supported attributes. 

(Sorry for rebuilding the PR btw but I'm trying to get this new WIP auto-merge-prevention bot to work and I thought I'd test it here. It's supposed to respect the "soon" emoji in the tag 😄)

honnibal:
Looks good!

Only thing I noticed is that the wording on the errors seems a little different from the rest of the messages. I often avoid imperatives in the messages, but where we do use them, I think we usually don't use "please".

spacy/matcher/matcher.pyx
```diff
@@ -568,6 +571,8 @@ def _preprocess_pattern(token_specs, string_store, extensions_table, extra_predi
             # Signifier for 'any token'
             tokens.append((ONE, [(NULL_ATTR, 0)], [], []))
             continue
+        if not isinstance(spec, dict):
+            raise ValueError(Errors.E154.format())
```
honnibal:
```suggestion
            raise ValueError(Errors.E158.format())
```

spacy/matcher/matcher.pyx
```diff
@@ -581,21 +586,29 @@ def _get_attr_values(spec, string_store):
...
                 continue
-            elif attr.upper() == "OP":
+            elif attr == "OP":
                 continue
-            if attr.upper() == "TEXT":
+            if attr == "TEXT":
                 attr = "ORTH"
-            attr = IDS.get(attr.upper())
+            if attr not in TOKEN_PATTERN_SCHEMA["items"]["properties"]:
+                raise ValueError(Errors.E152.format(attr=attr))
```
honnibal:
```suggestion
                raise ValueError(Errors.E156.format(attr=attr))
```

spacy/matcher/matcher.pyx
```diff
@@ -581,21 +586,29 @@ def _get_attr_values(spec, string_store):
...
+            attr = IDS.get(attr)
         if isinstance(value, basestring):
             value = string_store.add(value)
         elif isinstance(value, bool):
             value = int(value)
         elif isinstance(value, dict):
             continue
+        else:
+            raise ValueError(Errors.E153.format(vtype=type(value).__name__))
```
honnibal:
```suggestion
            raise ValueError(Errors.E157.format(vtype=type(value).__name__))
```

spacy/matcher/matcher.pyx
```diff
@@ -581,21 +586,29 @@ def _get_attr_values(spec, string_store):
...
         if attr is not None:
             attr_values.append((attr, value))
+        else:
+            # should be caught above using TOKEN_PATTERN_SCHEMA
+            raise ValueError(Errors.E152.format(attr=attr))
```
honnibal:
```suggestion
            raise ValueError(Errors.E156.format(attr=attr))
```

spacy/matcher/matcher.pyx
```diff
@@ -755,11 +768,13 @@ def _get_operators(spec):
...
     else:
         keys = ", ".join(lookup.keys())
-        raise KeyError(Errors.E011.format(op=spec["OP"], opts=keys))
+        raise ValueError(Errors.E011.format(op=spec["OP"], opts=keys))
 
 
 def _get_extensions(spec, string_store, name2index):
     attr_values = []
+    if not isinstance(spec.get("_", {}), dict):
+        raise ValueError(Errors.E154.format())
```
honnibal:
```suggestion
        raise ValueError(Errors.E158.format())
```

spacy/matcher/phrasematcher.pyx
```diff
@@ -62,6 +63,11 @@ cdef class PhraseMatcher:
         if isinstance(attr, long):
             self.attr = attr
         else:
+            attr = attr.upper()
+            if attr == "TEXT":
+                attr = "ORTH"
+            if attr not in TOKEN_PATTERN_SCHEMA["items"]["properties"]:
+                raise ValueError(Errors.E152.format(attr=attr))
```
honnibal:
```suggestion
                raise ValueError(Errors.E156.format(attr=attr))
```

spacy/matcher/phrasematcher.pyx
```diff
@@ -123,6 +129,10 @@ cdef class PhraseMatcher:
             length = doc.length
             if length == 0:
                 continue
+            if self.attr in (POS, TAG, LEMMA) and not doc.is_tagged:
+                raise ValueError(Errors.E155.format())
```
honnibal:
```suggestion
                raise ValueError(Errors.E159.format())
```

spacy/matcher/phrasematcher.pyx
```diff
@@ -123,6 +129,10 @@ cdef class PhraseMatcher:
...
+            if self.attr == DEP and not doc.is_parsed:
+                raise ValueError(Errors.E156.format())
```
honnibal:
```suggestion
                raise ValueError(Errors.E160.format())
```

Issue #4102, opened 2019-08-09
Wrong lemmatization using only tokenization (en language)


ajrader:
## Observation
I noticed if I tried to use spacy for tokenization only (disabling 'ner', 'parser' and 'tagger') that i was getting an erroneous mapping for the english word 'spun'.  As long as tagger is not disabled, this past tense verb is correctly mapped to 'spin'. But when 'tagger' is disabled then it gets mapped to 'spin-dry' which in my estimation is wrong.

I think the source of this error is due to line 35297 in https://github.com/explosion/spaCy/tree/master/spacy/lang/en/lemmatizer/lookup.py
"spun": "spin-dry"

## How to reproduce the behavior
`import spacy`
`nlp = spacy.load('en_core_web_sm')`
`doc = nlp('spun', disable=['tagger'])`
`for tok in doc0:`
`    print(tok.text, tok.lemma_)`

The output of this is
`spun spin-dry`

I noticed the same behavior when using 'en_core_web_lg' and 'en_core_web_md' because I think this is the default lemmatizer mapping in all 'en' dictionaries.

## Your Environment

* Operating System: Windows 10
* Python Version Used: 3.7.3
* spaCy Version Used: 2.1.7
* Environment Information: installed with conda

svlandeg:
It looks like the look-up lemmatization indeed has two weird entries:

> "spun": "spin-dry",
> "dry": "spin-dry",

I guess it should be

> "dry": "dry",
> "spun": "spin",
> "spun-dry": "spin-dry",

instead. You could make the edit to [lookup.py](https://github.com/explosion/spaCy/blob/master/spacy/lang/en/lemmatizer/lookup.py) and create a pull request if you like. 

However, please also mind [Ines' recent comment about lookup lemmatizations](https://github.com/explosion/spaCy/issues/4100#issuecomment-519880187): they should be made redundant for languages where there is better lemmatization available.

PR #4110, opened 2019-08-12
Correction of default lemmatizer lookup in English (Issue # 4104)

ajrader:
Resolves [issue 4102](https://github.com/explosion/spaCy/issues/4102).

## Description
* made the following changes to lookup.py:
   - 'dry' : 'dry'
   - 'spun': 'spin'
   - 'spun-dry': 'spin-dry'
* created new test (test_issue4104.py) and verified it passed after the above changes.

### Types of change
mainly feat / lemmatizer bug fix.

Note that this change will become obsolete once using a default lookup for a language is implemented.

## Checklist


spacy/tests/regression/test_issue4104.py
```diff
@@ -0,0 +1,13 @@
...
+import pytest
+from ..util import get_doc
+
+@pytest.mark.parametrize('text', ['dry spun spun-dry'])
+
+def test_issue4104(en_tokenizer, en_vocab, text):
+    """Test that English lookup lemmatization of spun & dry are correct"""
+    doc = get_doc(en_vocab, [t for t in text.split(" ")])
+    expected = {'dry': 'dry', 'spun': 'spin', 'spun-dry': 'spin-dry'}
+    assert [token.lemma_ for token in doc] == list(expected.values())
```
ines: Thanks for adding the test! Looks like this one failed on Python 3.5 because dicts aren't ordered yet, so `values()` returns the key in a different order. (Totally not your fault btw, it's not exactly intuitive.) Calling `sorted` around both lists should resolve this and make sure the order is always the same.

spacy/tests/regression/test_issue4104.py
```diff
@@ -0,0 +1,13 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+import pytest
+from ..util import get_doc
+
+@pytest.mark.parametrize('text', ['dry spun spun-dry'])
+
```
ines:
```suggestion
```

spacy/tests/regression/test_issue4104.py
```diff
@@ -0,0 +1,13 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+import pytest
+from ..util import get_doc
+
+@pytest.mark.parametrize('text', ['dry spun spun-dry'])
```
ines: Not sure we need to parametrize here because the expected values are hard-coded into the test anyways. So there's not really a motivation to try out different words here. So feel free to move the string `'dry spun spun-dry'` into the function.

spacy/tests/regression/test_issue4104.py
ajrader: Thanks. I went ahead and streamlined the test to not parametrize the string and compare it directly to a list of expected results.

Issue #4096, opened 2019-08-08
Add check for cycles in dependency trees to debug-data CLI


adrianeboyd:
## Feature description

The debug-data CLI should also check for cycles in dependency trees.

Cycles should be identified elsewhere, too, to prevent the kind of error from #4083, but a check in debug-data would also be helpful.

adrianeboyd:
Wait, I already get the `ValueError [E069]` from `GoldCorpus` when the corpus is loaded, so maybe it is just a matter of reporting that more cleanly.

ines:
Yes, that's a good idea! It might also be nice to test the `debug-data` a bit more thoroughly to make sure it really works as expected – and then I can document it properly 🙂 (It's currently still considered experimental and thus undocumented – but it has held up really well and people who've tried it found it very useful.)

PR #4114, opened 2019-08-13
WIP: Extending debug-data

adrianeboyd:
## Description

Starting with the checks for cycles needed for #4096, some additional checks and enhancements for `debug-data`. So far:

* Modify debug-data to load with GoldCorpus to iterate over .json/.jsonl files within directories

* Add GoldCorpus iterator train_docs_without_preprocessing to load original train docs without shuffling and projectivizing

* Report number of misaligned tokens

* Add more dependency checks and messages

To do:

* Add additional checks (suggestions?)

* Adjust thresholds for dependency warnings?

* JSON validation in GoldCorpus?

### Types of change

Enhancement.

## Checklist


adrianeboyd:
Output for en v2.1 train and dev data:

[en-v2.1-debug-data-output.txt](https://github.com/explosion/spaCy/files/3496789/en-v2.1-debug-data-output.txt)

(`iobj` looks like a mistake. I think it'd be useful to normalize some of the quotes.)

spacy/cli/debug_data.py
```diff
@@ -397,3 +547,18 @@ def _get_labels_from_model(nlp, pipe_name):
         return set()
     pipe = nlp.get_pipe(pipe_name)
     return pipe.labels
+
+
+def _get_nonprojective_examples(data):
```
ines: Can we move some of these into the `_compile_gold` helper? I've been trying to limit the times we loop over the entire data and for overall counts, I do think we can take care of all of this upfront. (It won't work for cases like `_get_examples_without_label`, though, because that needs to run for specific labels.)

That said, speed definitely isn't a big concern here and it's fine if the user has to wait a little to have their whole corpus analysed.

spacy/cli/debug_data.py
```diff
@@ -292,17 +311,143 @@ def debug_data(
...
+        # multiple root labels
+        root_labels = set()
+        for doc, gold in train_docs_unpreprocessed:
+            for i, (dep, head) in enumerate(zip(gold.labels, gold.heads)):
+                if head == i:
+                    root_labels.add(dep)
+        if len(root_labels) > 1:
+            msg.warn(
+                "Multiple root labels ({}) ".format(", ".join(root_labels)) +
+                "found in training data. Spacy's parser uses a single root "
```
ines:
```suggestion
                "found in training data. spaCy's parser uses a single root "
```

Sorry 😛

ines:
This is great! 👍

> JSON validation in GoldCorpus?

I started writing a JSON schema for the [proposed new training data format](https://github.com/explosion/spaCy/blob/master/spacy/cli/_schemas.py), also to see how it'd look as a schema... but given the discussions we've had, I think it probably makes sense to rewrite it to reflect the current JSON format so we can actually use it.

This would let us handle the most basic validation and the `debug-data` command could exit early if either the training or dev data is formatted incorrectly or has significant errors.

spacy/cli/debug_data.py
adrianeboyd: Sure, and some of the things it's counting can be removed, since they really shouldn't happen. (Does projectivize introduce cycles? Only if spacy has done something really wrong, but I wanted to rule it out for the Chinese corpus discussion.)

adrianeboyd:
Here's the output for en v2.1 train/dev:

[en-v2.1-debug-data-output.txt](https://github.com/explosion/spaCy/files/3497890/en-v2.1-debug-data-output.txt)

(I think `iobj` is a mistake and a large number of the misaligned tokens would be fixed with some quote normalization.)

honnibal:
If you can fix the misaligned tokens with quote normalisation, that would be great. I'm confused about when that regression crept in, maybe when I updated the alignment code for v2.1?

Is this ready to merge? Great work!

ines:
> Btw one feature request I can think of: should we have a --json output mode for debug-data?

What would that do? Output the result as a JSON object?

adrianeboyd:
I suppose it could be merged. There will still be more things to adjust and add in the future, but that's probably fine.

ines:
Great, merging! 👍 Also, I think the feature is now tested enough that we can make it "official" in the next release.

Issue #3850, opened 2019-06-14
displacy span rendering enhancement


yanaiela:
## Feature description

Adding a customized html display option into the `render` function

Hi!
I have a usecase where I want to add a hover into the EntityRenderer spans in displacy.
For the meantime, I just copied the code and hard-coded into it the component I needed but I thought this might be useful for other people as well, and with other customized html options.
Therefore, I'll present here the implementation idea on changing the code before a PR, and I'd be happy to get a feedback if you think this would be a useful feature / how to make it more generic.

So the idea is to add into the `options` argument another dictionary parameter. This dictionary would consist of additional html code that would be inserted into the html place-holders. The keys of the additional dictionary are additional html code and their values are elements that can be inserted into the keys, is necessary. When that dictionary is not empty, the html place-holders - `TPL_ENT` and `TPL_ENT_RTL` would add these additional components.

How does this sound?
Also, I'm not sure if this is useful for the `DependencyRenderer`, let me know if you think it is.

ines:
Thanks for raising this – it's definitely something I've thought about before! I wonder if we could maybe just add an option to replace the templates used for the page, container and spans. That would allow full customisation, without the awkwardness of just inserting HTML fragments into the templates (and the error potential here).

Another, slightly more complex solution would be to change the rendering slightly and use templating functions instead of just template strings. So there would be a function along the lines of `get_entity_span_html` that'd take the text and label as an argument and return a string. The advantage here is that it'd let you compute something with those values and include it in your template – for instance, you could look up the labels in a dictionary and display their descriptions in hover, and stuff like that. But I'm not sure if that's woth the added complexity.

yanaiela:
That's a great idea.
I guess the first one is more customized right?
The issue with the second solution is that in a few days/weeks/months someone else would need a different visualization.

Maybe the solution is to combine the two?

Have a function (e.g. get_entity_span_html) that gets the text, label and an html template, where this template can be a prepared template (like the one that exists now) or a completely new one, which will provide the user complete freedom.

ines:
> Have a function (e.g. get_entity_span_html) that gets the text, label and an html template, where this template can be a prepared template (like the one that exists now) or a completely new one, which will provide the user complete freedom.

Ah yeah, sorry if this was confusing in my description. The idea here would be that you could override those getter functions with your own. Your function would need to take the same arguments and has to return a string of markup – everything else is up to you.

yanaiela:
[...]

PR #4117, opened 2019-08-14
Custom entity render

yanaiela:
## Description
This PR is in response to the proposal suggested at #3850.
Now one can supply a custom html template for the entities displayer, along with custom arguments to have more control on the display.

### Types of change

This is a new feature for displacy, specifically for the entities renderer.

## Checklist

Still not sure about this, so after the feature logic will be approved, will change also documentation

### DIFF
--- a/spacy/displacy/render.py
+++ b/spacy/displacy/render.py
@@ -247,6 +247,15 @@ class EntityRenderer(object):
         self.direction = DEFAULT_DIR
         self.lang = DEFAULT_LANG
 
+        template = options.get("template", "")
+        if template:
+            self.ent_template = template
+        else:
+            if self.direction == "rtl":
+                self.ent_template = TPL_ENT_RTL
+            else:
+                self.ent_template = TPL_ENT
+
     def render(self, parsed, page=False, minify=False):
         """Render complete markup.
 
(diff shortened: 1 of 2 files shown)


spacy/displacy/render.py
```diff
@@ -284,6 +293,7 @@ def render_ents(self, text, spans, title):
             label = span["label"]
             start = span["start"]
             end = span["end"]
+            additional_params = span.get("additional_params", {})
```
ines: That's a really nice idea! 👍 Maybe we could just call the key `"params"`, to make it a bit shorter?

```suggestion
            additional_params = span.get("params", {})
```

ines:
Thanks a lot – really like the clean and straightforward implementation 🙂

spacy/displacy/render.py
```diff
@@ -247,6 +247,15 @@ def __init__(self, options={}):
         self.direction = DEFAULT_DIR
         self.lang = DEFAULT_LANG
 
+        template = options.get("template", "")
```
ines: Not a big deal, but we might as well just make it default to `None` here:

```suggestion
        template = options.get("template")
```

Issue #4107, opened 2019-08-12
Improvements on setting attributes in merge_noun_chunks function


alaponin:
## Feature description

It would be nice if during the merging of noun chunks the lemma would also be properly set. Currently, the lemma of the merged token is the lemma of the first word in the noun chunk sequence, which I don't think is the desired behavior in most cases. 
Additionally, it would be nice to have more flexibility in setting the attributes of the merged token. For example, I would like to keep the entity type of the root token as the entity type of the whole merged token.

ines:
I think those both sound like reasonable defaults, so we should probably consider just adding them to the built-in function 🙂 Feel free to submit a PR btw.

We probably want to avoid introducing settings to the built-in component functions – it adds too much complexity for what they are (small wrappers around `doc.retokenize`), and once we add settings, we then also need to add serialization methods to preserve them. 

Btw, also in case others come across this issue later: the `merge_noun_chunks` function itself is tiny, so if you ever need fully custom settings that might not be good defaults, I'd recommend just copying it and writing your own:

https://github.com/explosion/spaCy/blob/250a54414b41e365dfe5f45a17ca1f2c245ca3a2/spacy/pipeline/functions.py#L7-L21

ines:
Resolved by #4219!

PR #4130, opened 2019-08-16
Add lemma and ent_type attributes to the merge_noun_chunks function

alaponin:
## Description

A PR for issue #4107 

### Types of change

I have included the following small enhancements:
The lemma of the merged token consists of the lemmas of all the tokens. 
The entity type of the merged token is the entity type of the root token.

## Checklist

### DIFF
--- a/spacy/pipeline/functions.py
+++ b/spacy/pipeline/functions.py
@@ -16,7 +16,8 @@ def merge_noun_chunks(doc):
         return doc
     with doc.retokenize() as retokenizer:
         for np in doc.noun_chunks:
-            attrs = {"tag": np.root.tag, "dep": np.root.dep}
+            lemma = doc[np.start:np.end].lemma_
+            attrs = {"tag": np.root.tag, "dep": np.root.dep, "lemma": lemma, "ent_type": np.root.ent_type}
             retokenizer.merge(np, attrs=attrs)
     return doc
 
--- a/.github/contributors/alaponin.md
+++ b/.github/contributors/alaponin.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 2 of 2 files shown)


svlandeg:
Hi @alaponin , thanks for the PR! Could you also write an appropriate unit test to check this behaviour? You could write it at `spacy/tests/regression/test_issue4107` and you could use `assert` statements to verify behaviour on a toy example that wasn't working before, but now does work after your fix.

For some inspiration, a vaguely related unit test checking whether vector information was properly copied after running `tokenizer.split` can be found [here](https://github.com/explosion/spaCy/blob/master/spacy/tests/regression/test_issue3540.py).

spacy/pipeline/functions.py
```diff
@@ -16,7 +16,8 @@ def merge_noun_chunks(doc):
         return doc
     with doc.retokenize() as retokenizer:
         for np in doc.noun_chunks:
-            attrs = {"tag": np.root.tag, "dep": np.root.dep}
+            lemma = doc[np.start:np.end].lemma_
+            attrs = {"tag": np.root.tag, "dep": np.root.dep, "lemma": lemma, "ent_type": np.root.ent_type}
             retokenizer.merge(np, attrs=attrs)
```
svlandeg: Is this behaviour specific to the `merge_noun_chuncks` function, or should this be fixed in the underlying `retokenizer.merge` functionality?

honnibal: @svlandeg I do think we want to take care of this in the retokenizer. Noun chunks are always going to be a subtree, but for other merges that work as a subtree it's good to keep the parse working properly. I actually thought we already did this?

ines:
This should now be resolved automatically by #4219, so I'm closing this! Thanks for the PR though – there was obviously nothing wrong with it, but it ended up inspiring a more general fix 🙂

Issue #3258, opened 2019-02-11
Package size issue: consider a different format for lemmatizer dictionaries


remram44:
## Feature description

The spacy package is almost 300 MB, which is a lot. This comes mostly from the `spacy/lang/*/lemmatizer` files which include huge lookup tables in uncompressed Python source files.

Storing this as package_data, in gzip'd form, could save about 90%.

## Could the feature be a [custom component](https://spacy.io/usage/processing-pipelines#custom-components) or [spaCy plugin](https://spacy.io/universe)?

No. This would mean a change to this repository's structure, and the need for the maintainers to run a little script to create the gzip'd files from the source data before release. But going from 300 MB to 30 MB I think is worth it.

ines:
Thanks for bringing this up and yes, I definitely agree. We've actually been thinking about solutions for this as well 👍

mitar:
I think it could also be useful if this could be made optional. So that there would be a package without this included and you would have to initialize it so that you point to those files you provide from outside.

ines:
> So that there would be a package without this included and you would have to initialize it so that you point to those files you provide from outside.

I wish there was something like `extras_require` but for **not installing** certain submodules. So we could do something like `pip install spacy[light]`. But unfortunately, this only works for *adding* stuff (unless I'm missing something?). We briefly thought about moving the language data out into a separate package but this would make the installation more frustrating, especially for beginners, because things wouldn't just work out-of-the-box anymore.

mitar:
Then you can have two Python package. `spacy-lite` and `spacy`. And `spacy` depends on `spacy-lite` and adds data.

remram44:
I think whether or not that it done, we should minify the data. The current size is unnecessary.

I do not know how crucial the data is to normal spacy operation. Perhaps a "lite" package can be made, but that seems to me like a separate endeavor.

ines:
Quick update on this – copying over my reply from #3983:

>  I definitely agree that the package size is a problem and it's actually something we're actively working on at the moment. 
> 
[...]

PR #4141, opened 2019-08-18
Reduce size of language data

polm:
## Description

This PR reduces the size of a spaCy installation from ~280MB to ~74MB by packaging language data into gzipped JSON rather than Python source files, addressing concerns collected in #3258. Following this change language data is slightly less than half of the total size of a spaCy installation. Note that this depends on explosion/srsly#9, and requirements.txt will need to be updated after that is merged and released.

Lemmatizer classes have basically two structures: a single lemmatizer.py, or a directory where files divided by part of speech and other considerations are unified in an __init__.py. This change deals with both kinds of lemmatizers, but only compressed files that were individually over ~500kb. Some of these files were dictionaries and some were lists. In languages that use a directory for lemmatizer data, it might be a good idea to put all language data in a single JSON file in the future.

Note that because the data has just been moved from Python to JSON, the spaCy source is not actually smaller. As part of a dist build JSON files are gzipped and only the gzipped files are included for distribution. This also means the actual wheel files or .tar.gz packages haven't changed in size (since they're already zipped).

Gzipped msgpack was also considered but resulted in larger files than JSON.

### Types of change

This is an enhancment.

## Checklist

This is a re-submission of #4140 without the git history issues. Fixing the git history issues also fixed the failing tests.


honnibal:
Yay! Thanks.

Issue #2935, opened 2018-11-16
Uploading spacy model


zkrhm:
Hi, how do I upload model so my model can be available for all spacy user and downloadable from spacy's cli?

* **spaCy version:** 2.0.17
* **Platform:** Darwin-18.2.0-x86_64-i386-64bit
* **Python version:** 3.7.0
* **Models:** en_core_web_sm, en, xx

ines:
Hi! What model are you interested in contributing? Is it something we can train, or does it rely on data that can't be distributed?

All models we distribute need to be trained by us, so we can guarantee that they're good and safe, and so we can re-train then whenever we publish a new version. However, we do appreciate contributions and the best way you can help is by providing training pipelines that we can run – e.g. a series of settings and commands and the dataset you used, so we can reproduce your results and provide an official spaCy model.

Also see #2404:

> The problem is that we want to make sure all the models can be reproduced, so that when we release a new version, we can provide updated models. I've been working for a long time on better training automation, which might allow us to offer some facilities to contribute scripts. But for now the process is tied into our other infrastructure, so there's no clean repo to contribute to.

PR #4151, opened 2019-08-20
Added RONEC to spaCy Universe

avramandrei:
This module offers a named entity recognition model for Romanian Language. The model has been
trained on a  corpus that holds 5127 sentences, annotated with 16 classes. More information about the corpus can be found at https://github.com/dumitrescustefan/ronec/blob/master/README.md.

## Description
We used a script to transform the corpus from CoNLLU-Plus format to spaCy's json. The training has been done using the cli.

### Types of change
We would like to integrate this model with spaCy download interface and easily load with with `spacy.load("ro_ner")`, as described in the sample code.

We have seen [here](https://github.com/explosion/spaCy/issues/2935) that the models must be trained by you. The pipeline of training can be found at https://github.com/dumitrescustefan/ronec/blob/master/spacy/README.md. 

## Checklist

### DIFF
--- a/.github/contributors/avramandrei.md
+++ b/.github/contributors/avramandrei.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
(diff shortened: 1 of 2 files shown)


ines:
Oh wow, this is cool! ✨

And just to confirm: the whole corpus is MIT-licensed? If so, we'd definitely be super intested in integrating it into our model training pipeline and publish a `ro_core_web_sm`.

avramandrei:
Yes, it is MIT-licensed: https://github.com/dumitrescustefan/ronec/blob/master/LICENSE

avramandrei:
I have talked with my colleague and it was a misunderstanding from my part. The only trainable part of the corpus is the Named Entity part. The DEP, POS and the others are (partially) generated by another model. We would like to stick with only this part until we come with an update. 

Thank you very much and sorry for the inconvenience.

ines:
@avramandrei Thanks for the clarification. We actually want to retrain all models for v2.2 anyways, so we'll check it out and see if we can get Romanian included.

PR #4174, opened 2019-08-22
Small retokenizer fix

svlandeg:
## Description

This doesn't address a current bug, but it's a patch to avoid future problems.

The `_split()` function in `_retokenize.pyx` doubles the length of a `doc` and then moves tokens to create space for the new tokens. When it does, it accesses `doc.c[token_to_move]` in a loop, but the `token_to_move` variable was being defined on the **new** length of the `doc` - meaning that it was occassionally accessing data out of memory. I patched it here by making sure it's only reading tokens that actually existed in the original `doc` by introducing a new variable `orig_length`.

### Types of change
small fix

## Checklist


svlandeg:
Oh weird, the tests don't start running?

Issue #3972, opened 2019-07-16
PhraseMatcher returns only 1 match while more than 1 rules are verified


cverluise:
Hello,

We are working on a project ([PatentCity](https://github.com/Antoberge/patent_city/tree/master)) to collect inventors location directly from early ages patent documents. spaCy is undoubtedly our 🥇 library. So, please, let me start by a big "Thank You" to the spaCy developing community for your awesome work 💌 .

## Issue

We try to complement statistical GPE recognition with administrative entity lists look-up. To do so, we use `PhraseMatcher` (from `spacy.matcher`). We also want to track which kind of entity we match (e.g CITY, COUNTY, STATE, etc). Hence, we have a different rule for each. 

Now, some entities have common ngrams. E.g, 'New York' is a STATE,  a COUNTY and a CITY. In these case, the PhraseMatcher only returns 1 match. E.g, 'New York' is matched _only_ by a single rule, let say the CITY rule. It seems that the last rule has the priority.

We find this behavior potentially misleading. Is it in the spirit of the `PhraseMatcher` or should it be changed ? 

Thanks in advance for help ! 

## How to reproduce the behaviour

```python
from spacy.matcher import PhraseMatcher

import en_core_web_sm
nlp = en_core_web_sm.load()

matcher = PhraseMatcher(nlp.vocab)

matcher.add('COUNTY', None, *[nlp('New York')])
matcher.add('CITY', None, *[nlp('New York')])

matcher._docs
# Check that the PhraseMatcher was properly populated
# {14532842148348552135: (New York,), 13852145969607952771: (New York,)}  # ok

matcher(nlp('I live in New York'))
# [(13852145969607952771, 3, 5)]  # Only 1 match
```
<details>

## Environment

* Operating System: Mac OS Mojave
* Python Version Used: Python 3.7.0
* spaCy Version Used: spacy 2.1.3 
</details>

ines:
Thanks for the kind words and the detailed report 👍 I think you're correct here and I would have also expected the matcher to return two identical matches here, one for each match ID.

svlandeg:
I've gotten to the bottom line of why this is happening, it's a bug in the conversion between `Matcher` IDs and `PhraseMatcher` IDs. Trying to fix it!

Issue #4098, opened 2019-08-08
Flaky matcher test and ghost IDs


svlandeg:
When merging PR #4097 we ran into an error for [test_issue_1971_4 ](https://github.com/explosion/spaCy/blob/master/spacy/tests/regression/test_issue1501-2000.py#L319) which was resolved after restarting. What [happened](https://dev.azure.com/explosion-ai/Public/_build/results?buildId=1276) is that test, which is supposed to produce 1 match like `[(1046765068364475028, 0, 3)]`, produced 1 for each token instead: `[(9223420625083070725, 0, 1), (9223420625083070725, 1, 2), (9223420625083070725, 2, 3)]`. I can't reproduce the error on my own system, even though I also run Windows and Python 3.6.

Note that the unit test has a comment:
`# Uncommenting this caused a segmentation fault` 

After discussing with @ines, this seems related to previous issues #3839 and #3291, where the matcher was producing "ghost matches" with IDs that weren't in the string store.
Possibly, the fix that was supposed to resolve the original issue (matcher segfault) didn't fully fix it...

Not very clear on how to proceed here because it's so difficult to test, but we wanted to record this for future reference.

svlandeg:
More background: https://github.com/explosion/spaCy/issues/1971

svlandeg:
Again [weird behaviour](https://dev.azure.com/explosion-ai/Public/_build/results?buildId=1458&view=logs) with the tests running Windows and Python 3.6, this time the flaky failing test is #3839 (noted earlier to be related) 

> Traceback (most recent call last):
  File "D:\a\1\s\spacy\tests\regression\test_issue3839.py", line 17, in test_issue3839
    assert matches[0][0] == en_vocab.strings[match_id]
AssertionError: assert 10304252664499679735 == 11920309760829426267

honnibal:
Do we believe this resolved by #4179, or should we keep this open?

ines:
Maybe assume it's resolved and reopen if it turns out it's not? 😅 Alternatively, we could wait for a week or two and if we don't get a random ghost match test failure, assume it's resolved?

svlandeg:
Sure :-)

svlandeg:
This flaky behaviour is still here :(
PR #4219 [failed ](https://dev.azure.com/explosion-ai/Public/_build/results?buildId=1602) on Windows with Python 3.6 with 

> ______________________________ test_issue_1971_4 ______________________________
> Traceback (most recent call last):
[...]

PR #4179, opened 2019-08-22
Matcher ID fixes

svlandeg:
Fixes Issue #3972 (`PhraseMatcher` not returning duplicate matches with different IDs) and attempts to prevent ghost-matcher-ID issues with a small change in the `Matcher`.

## Description
* Previously, the `PhraseMatcher` had a field called `phrase_ids` which was used to map hashes of matches, onto the original pattern identifiers. However, if those patterns were identical like those in Issue #3972, the second pattern ("B") would overwrite the first ("A") in the `phrase_ids` map, thus resulting in only a match for "B". I changed the behaviour by having `phrase_ids` map into a list/vector of IDs, instead - these are stored in `ent_id_matrix`.
* Made a small change to one of the calls to  `get_ent_id` in `Matcher.pyx`. I would be surprised if this would fix the ghost issues we've been seeing (cf Issue #4098), but it can't hurt, as it feels a little bit more safe not to do the `[1]` but have the loop in `get_ent_id` take care of that instead.

### Types of change
bug fix

## Checklist


honnibal:
Nice, I see the problem!

I was going to say, why can't we just make the values of `phrase_ids` a pointer to an array...But if we did that, we'd always have to be reallocating it, which gets annoying (especially when it's 0-terminated). And storing pointers to vectors is kind of bad, because then we have to have dealloc logic, to avoid leaking memory. So I do think your solution is best.

Issue #4170, opened 2019-08-22
Unexpected behavior during convert .iob to json file


F0rge1cE:
## How to reproduce the behaviour

I try to call `python -m spacy convert <path> ./ -t json -c iob` with this labeled NER training corpus:
[https://github.com/EuropeanaNewspapers/ner-corpora/blob/master/enp_DE.onb.bio/enp_DE.onb.bio](url)

After the conversion, you can get a json file with content:
```
[
  {
    "id":0,
    "paragraphs":[
      {
        "sentences":[
          {
            "tokens":[
              {
                "orth":"November",
                "tag":"-",
                "ner":"O"
              }
            ]
          }
        ]
      }
    ]
  },
  {
    "id":0,
    "paragraphs":[
      {
        "sentences":[
          {
            "tokens":[
              {
                "orth":"Heute",
                "tag":"-",
                "ner":"O"
              }
            ]
          }
        ]
      }
    ]
  },
......
```
Such behavior is really weired. Seems that every separate token is considered as a whole sentence and a paragraph. 

Another issue has already mentioned here: #4111 , which will cause exception when the token contains any non-word char ("[^\w\-]" in the regex scope).

## Your Environment

* Operating System: OSX 10.13.6
* Python Version Used: 3.7.3
* spaCy Version Used: 2.1.8
* Environment Information: N/A

adrianeboyd:
This looks like a bug. I'll take a look.

adrianeboyd:
For a temporary solution that uses `-c ner`, see: https://stackoverflow.com/q/57551479/461847

(The OCR errors are probably going to cause a lot of problems and I suspect that the lack of sentence/document boundaries is going to cause problems during training. Even if automatic sentence boundaries aren't perfect, it would be better to divide the data into sentence-ish segments for shuffling/batching during training.)

PR #4186, opened 2019-08-23
Updates/bugfixes for NER/IOB converters

### LINKED ISSUE #4111: spacy convert .iob fix.

spacy convert test.iob ./

It fails if the file have tokens that contain "? , $"  f.e. PRP$ i.e. explain('PRP$') => 'pronoun, possessive'

Here is the fix. Modify the regex in iob2json.py::read_iob() , from  : 

 tokens = [re.split("[^\w\-]", line.strip())]

to:

 tokens = [re.split("[^\w\-$,?.]", line.strip())]

or it will be probably better to split on white-space, rather than enumerating all the exception cases !!

Example :

Lily NNP B-SRC
has VBZ O
sung VBN B-PRED
the DT O
blues NNS B-DTGT
all PDT O
her PRP$ O
life NN O

adrianeboyd:
## Description

Updates/bugfixes for NER/IOB converters:

* Converter formats `ner` and `iob` use autodetect to choose a converter if possible

* `iob2json()` is reverted to handle sentence-per-line data like`word1|pos1|ent1 word2|pos2|ent2`

  * Fix bug in `merge_sentences()` so the second sentence in each batch isn't skipped

* `conll_ner2json()` is made more general so it can handle more formats with whitespace-separated columns

  * Supports all formats where the first column is the token and the final column is the IOB tag; if present, the second column is the POS tag

  * As in CoNLL 2003 NER, blank lines separate sentences, `-DOCSTART- -X- O O` separates documents

  * Add option for segmenting sentences (new flag `-s`)

  * Parser-based sentence segmentation with a provided model, otherwise with sentencizer (new option `-b` to specify model)

  * Can group sentences into documents with `n_sents` as long as sentence segmentation is available

  * Only applies automatic segmentation when there are no existing delimiters of this type in the data

* Provide info about settings applied during conversion with warnings and suggestions if settings conflict or might not be not optimal.

* Add tests for common formats

Fixes #4111 and #4170.

### Types of change

Bugfixes and enhancements.

## Checklist


adrianeboyd:
Concerns:

* Possibly a few too many options and too much option logic in `conll_ner2json()`, but I tried to make the defaults and the default warnings fairly reasonable, since many newer datasets are long lists of unsegmented tokens and turning them into one huge document with no sentence breaks is not a great idea.

* There aren't any tests for the option logic.

* How should tests including `msg` be handled in pytest? For now, there's a lot of `if msg:`

* Is there a good way to link to raw examples in the docs? I could move the test documents into external files and link to them in github as examples? (I would make more canonical versions rather than tests where every sentence is in a different format as it is currently.)

spacy/tests/test_cli.py
```diff
@@ -32,6 +32,91 @@ def test_cli_converters_conllu2json():
...
+        assert [t["orth"] for t in tokens] == ["I", "like", "London", "and", "New", "York", "City", "."]
+        assert [t["ner"] for t in tokens] == ["O", "O", "U-GPE", "O", "B-GPE", "I-GPE", "L-GPE", "O"]
+
+
+def test_cli_converters_conll_ner2json():
+    lines = [
+        "-DOCSTART- -X- O O",
+        "",
+        "I	O",
+        "like	O",
```
ines: Small suggestion for futureproofing the test: Maybe we should use `\t` here to make the tabs more visible/explicit?

adrianeboyd: That would be fine. Or should we move the tests to external files?

ines: I think in this case, it's short enough to have it in the same file. It makes the test a lot easier to read and to follow if it's all there.

adrianeboyd: What about having examples to include in the docs? Would you want separate ones?

ines:
Sorry, somehow I missed these questions!

> How should tests including msg be handled in pytest? For now, there's a lot of if msg:

That's the `wasabi.Printer` object we're passing around, right? Maybe it's easier to not pass that around and just have each converter declare its own instance at the top of the function, e.g. `msg = Printer()`. If we want to incorporate a "silent mode" logic, we could set `no_print=True` on the `Printer` ([see here](https://github.com/ines/wasabi#method-printer__init__)) and then pass that in instead.

> Is there a good way to link to raw examples in the docs? I could move the test documents into external files and link to them in github as examples? (I would make more canonical versions rather than tests where every sentence is in a different format as it is currently.)

Yes, for the docs framework, I actually wrote a component that automatically embeds example code from GitHub. We're doing something similar to embed the long JSON training data example [here](https://spacy.io/api/annotation/#json-input). All you need to do is create a regular Markdown code block and set the GitHub URL as its content. Like this:

```
```python
https://github.com/explosion/spaCy/tree/master/examples/pipeline/custom_component_countries_api.py
\```
```

(Had to add the `\` to prevent GitHub from messing up the formatting.) Also see the source of [`/usage/examples`](https://raw.githubusercontent.com/explosion/spaCy/master/website/docs/usage/examples.md).

honnibal:
Looking good! Is this ready to merge?

adrianeboyd:
Yes, I think so. The sample data was the last major thing I wanted to add.

ines:
Yes, looks good to me! One thing I'd probably still change is the `msg` logic. Instead of passing this around and doing the `if msg` checks, define `msg = Printer()` at the top of each converter function.

Issue #2339, opened 2018-05-17
Norwegian language code should be "no"


gurudevrao:
For Norwegian language spaCy currently uses "nb" as the language code.

However, "no" is the ISO 639-1 code for Norwegian.
"nb" is one of the two written standards of Norwegian ie Bokmål the other being Nynorsk (code "nn")

So, should "nb" be changed to "no" instead or change the language name to Bokmål?

ines:
Merging this with #2338!

PR #4187, opened 2019-08-23
💫 Change lang.nb to lang.no?

ines:
## Description
Also see #2339. Not sure if we want to change this, but if we do, v2.2 could be a good opportunity. Suggested implementation here includes a fallback, so the user will see a warning.

### Types of change
?

## Checklist


honnibal:
I feel quite under-informed on this. @jarib , do you have an opinion?

We also have the more general question about what to do about dialect variations. It might be good sometimes to have an additional qualifier.

Is it too crazy to extend the language codes to have a notation like `no.nb`? We could hen have something like `de.ch` for Swiss German.

jarib:
To me it makes sense to keep `nb` (Norsk Bokmål) and `nn` (Norsk Nynorsk) separate, as they are the two official written languages. 

Either separate at the top-level like now, or `no.nb` / `no.nn` would work. And simply `no` could be an alias for `no.nb`, which is the most used variant.

It is possible that a model trained on `nb` text would perform decently on `nn` text as well, but we shouldn't exclude the possibility of a future `nn` model. How would you name it if you merge this? Having both `no` and `nn` models would be weird.

The UD Norwegian treebanks appear to use `no_bokmaal` and `no_nynorsk` in their files: 

https://github.com/UniversalDependencies/UD_Norwegian-Bokmaal/tree/master
https://github.com/UniversalDependencies/UD_Norwegian-Nynorsk/tree/master

I like the idea of `no.nb` and `no.nn`, if possible.

alejandrojcastaneira:
I am in favor on the idea of extending languages clases for language dialects, because in some cases dialects could be very different from the original language like the de.ch by example.

svlandeg:
As a side note, "no" would be in line with the UD corpora: https://github.com/explosion/spaCy/blob/master/bin/ud/run_eval.py#L255. If we do make the change, we should remove that additional check too.

jarib:
@svlandeg How do they split `no` and `nn` then? 

From what I can see the UD treebanks use `no_bokmaal` and `no_nynorsk`:

```console
$ ll ud-treebanks-v2.4/UD_Norwegian-Bokmaal
total 40856
-rw-r--r--@ 1 jari  staff    68B May 11 09:44 LICENSE.txt
-rw-r--r--@ 1 jari  staff   6.5K May 11 09:44 README.md
-rw-r--r--@ 1 jari  staff   2.1M May 11 09:44 no_bokmaal-ud-dev.conllu
-rw-r--r--@ 1 jari  staff   196K May 11 09:44 no_bokmaal-ud-dev.txt
-rw-r--r--@ 1 jari  staff   1.8M May 11 09:44 no_bokmaal-ud-test.conllu
-rw-r--r--@ 1 jari  staff   161K May 11 09:44 no_bokmaal-ud-test.txt
-rw-r--r--@ 1 jari  staff    14M May 11 09:44 no_bokmaal-ud-train.conllu
-rw-r--r--@ 1 jari  staff   1.3M May 11 09:44 no_bokmaal-ud-train.txt
-rw-r--r--@ 1 jari  staff   9.6K May 11 09:44 stats.xml
$ ll ud-treebanks-v2.4/UD_Norwegian-Nynorsk
total 39520
-rw-r--r--@ 1 jari  staff    68B May 11 09:45 LICENSE.txt
-rw-r--r--@ 1 jari  staff   4.7K May 11 09:45 README.md
-rw-r--r--@ 1 jari  staff   1.8M May 11 09:45 no_nynorsk-ud-dev.conllu
-rw-r--r--@ 1 jari  staff   166K May 11 09:45 no_nynorsk-ud-dev.txt
-rw-r--r--@ 1 jari  staff   1.5M May 11 09:45 no_nynorsk-ud-test.conllu
-rw-r--r--@ 1 jari  staff   136K May 11 09:45 no_nynorsk-ud-test.txt
-rw-r--r--@ 1 jari  staff    14M May 11 09:45 no_nynorsk-ud-train.conllu
-rw-r--r--@ 1 jari  staff   1.3M May 11 09:45 no_nynorsk-ud-train.txt
-rw-r--r--@ 1 jari  staff   9.2K May 11 09:45 stats.xml
```

svlandeg:
Oh, you're right @jarib. I think back at the time when I was writing those UD eval scripts I didn't realise that they made a language distinction by doing `no_bokmaal` and `no_nynorsk`, because for other languages the additional information means other things, like `nl_alpino` and `nl_lassysmall` simply refer to 2 different datasets in the same language `nl`.

ines:
Thanks everyone for your input! It does sound like this is still a bigger question, so I guess we'll hold off on this plan until we've solved the "sub-sub" language classes.

I do like the idea of `no.nb` or `de.ch`. There are two main questions here:

* Is this going to mess up any assumptions we currently make about `nlp.lang` and the downstream usage of it (like, model names etc)? I do think we just treat it as an arbitrary unicode sequence and dynamically import from `"spacy.lang.{lang}"`, so it should all keep working as expected.
* How should we organise the file strucutre? In theory, we could just have a `lang/no/nb/__init__.py` defining a `NorwegianBokmal` class, and a `lang/no/__init__.py` that sets `Norwegian = NorwegianBokmal` (similar to what I did in the PR). But I am slightly worried about bloat and code duplication 🤔

PR #4202, opened 2019-08-27
Fix visualizer options linking for displaCy.

BreakBB:
## Description
There were two broken links in the displaCy docs which are fixed with this PR.

### Types of change
Docs

## Checklist


ines:
Thanks so much! 👍

PR #4219, opened 2019-08-30
Modify retokenizer to use span root attributes

### LINKED ISSUE #4015: How to convert OntoNotes 5 dataset to spaCy’s JSON format?

### LINKED ISSUE #4051: Retokenizer not setting POS tags correctly
I am trying to merge entities and noun chunks for a custom application without having to use merge_entities or merge_noun_chunks pipeline. But the pos tag information is inaccurate. Here's one example: 
```
import spacy 
from spacy import displacy
from spacy.util import filter_spans

nlp = spacy.load('en_2.1.0_md')
doc = nlp("While our net charge-off rate improved from a year ago, our provision expense increased due to a $150 million reserve build in the first quarter of 2019 compared with a $550 million reserve release a year ago.")

spans = list(doc.ents) + list(doc.noun_chunks)
spans = filter_spans(spans)

    for span in spans:
        retokenizer.merge(span)

for i, tkn in enumerate(doc):
    print("{:<2}{:<2}{:<32}{:<2}{:<8}{:<2}{:<8}{:<2}{:<2}".format(i, '|', tkn.text, '|', tkn.pos_, '|', tkn.dep_, '|', tkn.ent_type_))
```

But if I specifically set the pos attribute to be the same as the root's attribute, it works fine: 

```
    for span in spans:
        retokenizer.merge(span, attrs={"POS": span.root.pos_})
```
I remember this being the default behaviour in 2.0. And also, even after the fix, displacy would show the incorrect pos tag information: 
```
displacy.render(doc, style='dep')
```

## Your Environment
* Operating System: MacOS
* Python Version Used: 3.6
* spaCy Version Used: 2.1.4

adrianeboyd:
## Description

* tag/pos/morph are set to root tag/pos/morph initially (can be overridden by attrs later)

* lemma and norm are reset and end up as orth (not ideal, but better than orth of first token)

Fixes #4051.

### Types of change

Bugfix.

## Checklist

### DIFF
--- a/spacy/tokens/_retokenize.pyx
+++ b/spacy/tokens/_retokenize.pyx
@@ -171,6 +171,20 @@ def _merge(Doc doc, int start, int end, attributes):
     token.tag = doc.c[span.root.i].tag
     token.pos = doc.c[span.root.i].pos
     token.morph = doc.c[span.root.i].morph
+    token.ent_iob = doc.c[span.root.i].ent_iob
+    token.ent_type = doc.c[span.root.i].ent_type
+    # If span root is part of an entity, merged token is B-ENT
+    if token.ent_iob in (1, 3):
+        token.ent_iob = 3
+        # Unless start token is not B-ENT and previous token is of the same
+        # type, then I-ENT
+        if doc.c[start].ent_iob != 3 \
+                and doc.c[start].ent_type != token.ent_type \
+                and doc.c[start - 1].ent_type == token.ent_type:
+            token.ent_iob = 1
+    # If following token is part of a different entity, following token is B
+    if end < len(doc) and doc.c[end].ent_iob in (1, 3) and doc.c[end].ent_type != token.ent_type:
+        doc.c[end].ent_iob = 3
     # Unset attributes that don't match new token
     token.lemma = 0
     token.norm = 0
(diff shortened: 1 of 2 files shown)


honnibal:
Looks good! So this is the more general version of #4430 right?

ines:
Yes, this should resolve #4130. (Sorry, no idea why I assigned myself to this – I think I misclicked while labelling.)

@honnibal Do we still need modifications here to also cover the `Token.ent_type`?

adrianeboyd:
I didn't look at the ent bits because I thought they were already handled, but now that I look at that part, I'm not sure the ent merging logic is sufficient. The iob is sort of handled, but not the types.

With this test case:

https://github.com/explosion/spaCy/blob/bcd1b12f430140d5316741172a9eb505fd0c115c/spacy/tests/doc/test_retokenize_merge.py#L202-L218

The ent tags look like:

```
a 	 O
b 	 O
c d 	 O
e f 	 B-ent-de
g 	 I-ent-fg
h i 	 O
```

So `g` ends up being identified as part of `ent-de` by `doc.ents`.

Hmm, the easy cases are:

- `O+`: no changes
- `I+`: no changes
- `BI+`: becomes `B`

Everything else is tricky. As a better default, you could do `I` with the span root ent_type and then adjust it to `B` in some cases? Deciding when to adjust it to `B` is the tricky part. When `span[start]` is `B-same` (the current version)? Or detect when the whole span is `BI+`?

I'll submit one attempt in a second, but I think it still needs some work.

I also think it'd be much simpler to maintain this if you get rid of the `_merge()` exception for single cases. Is it common enough to be worth the performance improvement?

spacy/tokens/_retokenize.pyx
```diff
@@ -167,6 +167,26 @@ def _merge(Doc doc, int start, int end, attributes):
...
+    token.ent_type = doc.c[span.root.i].ent_type
+    merged_iob = token.ent_iob
+    # If span root is part of an entity, merged token is B-ENT
+    if token.ent_iob in (1, 3):
+        merged_iob = 3
+        # If start token is I-ENT and previous token is of the same
+        # type, then I-ENT (could check I-ENT from start to span root)
+        if doc.c[start].ent_iob == 1 \
+                and doc.c[start].ent_type == token.ent_type \
+                and doc.c[start - 1].ent_type == token.ent_type:
```
honnibal: Can `start` be 0 here? If so the `start-1` will introduce an out-of-bounds access.

spacy/tokens/_retokenize.pyx
```diff
@@ -480,3 +505,13 @@ def _validate_extensions(extensions):
...
         if not is_writable_attr(extension):
             raise ValueError(Errors.E119.format(attr=key))
+
+
+cdef make_iob_consistent(TokenC* tokens, int length):
+    cdef int i
+    if tokens[0].ent_iob == 1:
+        tokens[0].ent_iob = 3
+    for i in range(1, length):
+        print(tokens[i].ent_iob, tokens[i].ent_type)
```
honnibal:
```suggestion

```

honnibal:
I think this looks good. I agree that it's bad to have the two merge functions. Let's delete the single-merge case.

spacy/tokens/_retokenize.pyx
adrianeboyd: You mean you don't want my debugging code?!?

PR #4222, opened 2019-09-01
adds Croatian lemma_lookup.json, license file and corresponding tests

isaric:
I have added a Croatian lemmatizer with the corresponding tests and supporting test data.
## Description

It adds three new files : the lookup_lemma.json and corresponding licence, the corresponding test and the support fixture in terms of adding a Croatian tokenizer.
### Types of change

This is an enhancement that allows lemmatization of Croatian words. The lemmatizer was adapted from the Croatian lexicon v1.1 which is under a GPL v3 licence. GPL v3 should be compatible with 
the spaCy MIT licence. Let me know if I am mistaken.
## Checklist

### DIFF
--- a/spacy/lang/hr/__init__.py
+++ b/spacy/lang/hr/__init__.py
@@ -18,6 +18,7 @@ class CroatianDefaults(Language.Defaults):
     )
     tokenizer_exceptions = update_exc(BASE_EXCEPTIONS)
     stop_words = STOP_WORDS
+    resources = {"lemma_lookup": "lemma_lookup.json"}
 
 
 class Croatian(Language):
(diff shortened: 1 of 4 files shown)


isaric:
@ines Sorry I thought GPL was permissive enough. I will try to contact the authors of the lexicon through the following week. Can you tell me what format the permission should be? Is a copy of an email enough or do you need some sort of contract?

honnibal:
@isaric I understand that the licensing is confusing, because there's so many and the differences are a bit technical. But GPL definitely has different implications than MIT: you can only use GPL resources in other GPL projects, because it's sort of a "give to receive" system. So if we merged some GPL resources, we'd have to be GPL too, which would mean only GPL code could use spaCy.

Including the resources in spaCy would require the authors to give us an MIT license to the work. This is really the same as giving everyone an MIT license, because if it's included in spaCy under MIT, you anyone is free to redistribute it under that license.

So, it's not really a question of getting a permission with a certain format. You'd be asking the authors to change the license from GPL to MIT. It's possible they're not willing to do that, which is definitely their choice.

isaric:
@honnibal Thank you for taking the time to explain.
@honnibal @ines I found another source that is licensed under Apache 2.0. Is that ok to use? Let me know and I will make the changes.

ines:
> I found another source that is licensed under Apache 2.0. Is that ok to use? Let me know and I will make the changes.

Yes, Apache 2.0 would be compatible 🙂

isaric:
@ines Ok, I will close this PR and make the necessary changes.

PR #4226, opened 2019-09-03
Add textcat to train CLI

adrianeboyd:
## Description

Since paragraphs are currently treated as documents, I added support for `"cats"` in the current training format the paragraph level.

The two main components are adding support for reading/writing `"cats"` and for using the `"cats"` annotation to train textcat models with the train CLI.

The `Scorer` attempts to choose reasonable default metrics for the combinations of `exclusive_classes` and the number of labels in the training data. The positive label can be provided for a binary textcat component with two labels.

###  Add doc.cats to spacy.gold at the paragraph level

Support `doc.cats` as `"cats": [{"label": string, "value": number}]` in
the spacy JSON training format at the paragraph level.

* `spacy.gold.docs_to_json()` writes `docs.cats`
* `GoldCorpus` reads in cats in each `GoldParse`

### Add textcat to train CLI

* Add textcat options to train CLI
* Add textcat labels in `TextCategorizer.begin_training()`
* Add textcat evaluation to `Scorer`:
  * For binary exclusive classes with provided label: F1 for label
  * For 2+ exclusive classes (and no label if 2 classes): F1 macro average
  * For multilabel (not exclusive): ROC AUC macro average (currently relying on sklearn)
* Provide user info on textcat evaluation settings, potential incompatibilities
* Provide pipeline to Scorer in `Language.evaluate` for textcat config
* Customize train CLI output to include only metrics relevant to current pipeline
* Add textcat evaluation to evaluate CLI

TODO:

* Remove sklearn requirement
* Add Scorer tests for new metrics
* Improve validation/feedback for a wider range of textcat model settings
* Add support for sparse textcat labels?

### Types of change

Enhancement.

## Checklist


honnibal:
Looking good so far! Agree with the todos. The `sklearn` dependency is the main blocker.

adrianeboyd:
I think the train CLI needs at least one more check for multilabel vs. not because otherwise you get nonsense results. The sparse feature still aren't implemented, but that can wait for now, I think.

adrianeboyd:
Is it okay to have a small amount of sample data from CC BY-SA sources in `examples/`? If not, I can remove it and make fake examples.

honnibal:
@adrianeboyd It's okay to have that in the `examples/` directory and have a `LICENSE` file for it. Users don't need to install that data when they pip install, so it's no problem.

If you want examples for `CC-BY NC`, you could make a repo with them and ask users to clone the repo if they can agree to the license.

adrianeboyd:
I wouldn't be at all surprised if there are some edge cases I've missed in the config verifications, but I think this is ready to be de-WIP-ified.

honnibal:
Really good to have this in!

PR #4235, opened 2019-09-04
Changes to wiki_entity_linker

euand:
## Description

I have included various changes in this PR based on work I have done for [Schibsted Media Group](https://github.com/schibsted) adapting the code for working with Swedish wikipedia. In particular:
 - Adding the option of training entity linkers for languages other than English (using wikidata & wikipedia in those languages)
 - Get descriptions from wikipedia pages (as in [the following paper](https://cogcomp.seas.upenn.edu/page/publication_view/817)) rather than wikidata. In Swedish it is more common for entities to have a wikipedia page than a Swedish description in wikidata.
 - Pulling evaluation into another file to make it easier to read.
 - Evaluate precision and recall rather than accuracy, and count missed gold entities as false negatives.
 - Using the .gz wikidata rather than .bz2 because it was significantly faster.
 - When creating training data create a single json lines file with both article text and entities. This makes it much faster to read training data in and also doesn't create millions of individual files. 
 - Simplify the scripts by providing only a directory (I'm not sure if this is preferred since this used to be how the example pipeline was structured and that has been changed?)

Tested with python 3.7, the 2019-06-17 wikidata dump and the 2019-08-01 wikipedia dump.

### Types of change
Enhancement (the ability to train entity linkers for languages other than English, mainly).

## Checklist

- [x] I have submitted the spaCy Contributor Agreement. (Covered by [this agreement])(https://github.com/explosion/spaCy/blob/master/.github/contributors/Schibsted.png)
- [ ] I ran the tests, and all new and existing tests passed. (59 tests failed but they fail when I run them on `master` branch too.


svlandeg:
Hi @euand , thanks a lot for this great work! I'm happy to see this getting tested on a non-English language, as the code was always intented to be quite generic.

With respect to some of the functional changes you made, I propose to make those options regulated by parameters: e.g. whether to get the description from Wikidata or Wikipedia, or the choice between which performance numbers (acc/p/r) to use/output.

I'll definitely have a more detailed look / review later on!

svlandeg:
Also, with respect to the simplification of the CLI scripts: I made them so intentionally so that you could run the CLI scripts in different stages. For instance in `wikidata_pretrain_kb.py`, if you would provide the parameter `loc_prior_prob`, it would *read* from that location and the user is free to define those prior probabilities anyway/anywhere they want, but if that parameter is not provided, this information gets parsed and written to the default location within the output directory.

euand:
Hi @svlandeg, thanks for your comments! I will revert the CLI parameter name changes and parameterise my changes.

bin/wiki_entity_linking/wikidata_pretrain_kb.py
```diff
@@ -13,63 +13,73 @@
...
+        output_dir.mkdir(parents=True)
 
     # STEP 1: create the NLP object
-    print(now(), "STEP 1: loaded model", model)
-    nlp = spacy.load(model)
+    logger.info("STEP 1: loading model {}".format(model))
+    if model_is_path:
+        nlp = spacy.load(Path(model))
+    else:
+        nlp = spacy.load(model)
```
svlandeg: Are you sure this change is necessary? `spacy.load(model)` should work with a package name like `en_core_web_lg`, a string denoting a path, or an actual `Path`

bin/wiki_entity_linking/wikidata_pretrain_kb.py
```diff
@@ -13,63 +13,73 @@
...
 
 @plac.annotations(
-    wd_json=("Path to the downloaded WikiData JSON dump.", "positional", None, Path),
-    wp_xml=("Path to the downloaded Wikipedia XML dump.", "positional", None, Path),
+    wikidata_json=("Path to the downloaded WikiData JSON dump.", "positional", None, Path),
+    wikipedia_xml=("Path to the downloaded Wikipedia XML dump.", "positional", None, Path),
     output_dir=("Output directory", "positional", None, Path),
     model=("Model name, should include pretrained vectors.", "positional", None, str),
+    model_is_path=("Model is a path not a model name.", "flag", "mp"),
     max_per_alias=("Max. # entities per alias (default 10)", "option", "a", int),
```
svlandeg: cf other comment - I think this should have worked if you passed in a path to a model - perhaps we only need to update the docstring for the `model` parameter?

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,247 @@
+import logging
+import random
+
+from collections import defaultdict
+from tqdm import tqdm
+from typing import Dict, NamedTuple
+
+logger = logging.getLogger(__name__)
+
```
svlandeg: Thanks for updating all the `print` statements - definitely more clean solution to work with a proper logger instead :-)

bin/wiki_entity_linking/wikidata_pretrain_kb.py
euand: You're right! I'll remove it and update the docstring.

bin/wiki_entity_linking/__init__.py
```diff
@@ -0,0 +1,10 @@
+TRAINING_DATA_FILE = "gold_entities.jsonl"
+KB_FILE = "kb"
+KB_MODEL_DIR = "nlp_kb"
+OUTPUT_MODEL_DIR = "nlp_long_descr"
+
+PRIOR_PROB_PATH = "prior_prob.csv"
+ENTITY_COUNTS_PATH = "entity_freq.csv"
+ENTITY_DEFS_PATH = "entity_defs.csv"
+ENTITY_DESCR_PATH = "entity_descriptions.csv"
+FILTERED_ENTITY_DESCR_PATH = "entity_descriptions_filtered.csv"
```
svlandeg: I like the idea of centralizing some fixed names here, but the last 6 are not used in the code (anymore). So best to either remove them or adjust the corresponding code like in `wikidata_pretrain_kb.py` to use these.

bin/wiki_entity_linking/wikidata_pretrain_kb.py
```diff
@@ -79,64 +87,66 @@ def main(
         )
 
     # STEP 2: create prior probabilities from WP
-    print()
-    if loc_prior_prob:
-        print(now(), "STEP 2: reading prior probabilities from", loc_prior_prob)
-    else:
+    if not prior_prob_path.exists():
```
svlandeg: What's the motivation to remove these additional logging prints? I found them informative when I was running the scripts themselves, as they make sure that all steps are in the logging output and you can review which parts were newly created (the "writing" statements) and which were used as input (the "reading" statements)

euand: I'll add them back in

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+def measure_acc(data, el_pipe=None, error_analysis=False):
+    # If the docs in the data require further processing with an entity linker, set el_pipe
+    correct_by_label = defaultdict(int)
+    incorrect_by_label = defaultdict(int)
+
+    docs = []
+    golds = []
+    for d, g in tqdm(data, leave=False):
+        if len(d) > 0:
+            docs.append(el_pipe(d))
```
svlandeg: `el_pipe` could be None, in which cause you need to append `d` to the `docs` as such

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # measuring combined accuracy (prior + context)
+    el_pipe.cfg["incl_context"] = True
+    el_pipe.cfg["incl_prior"] = True
+    dev_precision_combo, dev_recall_combo, dev_acc_combo_d = measure_acc(dev_data, el_pipe)
+    print("dev precision combo avg:", round(dev_precision_combo, 3))
+    print("dev recall combo avg:", round(dev_recall_combo, 3))
+
+
+def measure_acc(data, el_pipe=None, error_analysis=False):
```
svlandeg: The `error_analysis` parameter is not doing anything anymore because that part of the code was removed. I agree it was kind of "hacky" to put it in the `measure_acc` method, but it was useful to me when debugging :p either way, if we want to leave it out, this parameter can go too.

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+                    else:
+                        incorrect_by_label[ent_label] += 1
+                        false_positives += 1
+                        false_negatives += 1
+
+        except Exception as e:
+            print("Error assessing accuracy", e)
+
+    _, acc_by_label = calculate_acc(correct_by_label, incorrect_by_label)
+    return (
```
svlandeg: I think your switch to `precision` and `recall` makes sense. `accuracy` is anyway confusing because you can't really define true negatives with this data. So in the definition we have here, accuracy equals precision, right? Should we go ahead then and remove the references to accuracy/`acc` and call all of it `precision`, also the `_by_label` dictionaries?

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+                        if c.entity_ == gold_entity:
+                            oracle_candidate = c.entity_
+
+                    best_index = scores.index(max(scores))
+                    best_candidate = candidates[best_index].entity_
+                    random_candidate = random.choice(candidates).entity_
+
+                random_correct_d[label] += (gold_entity == random_candidate)
+                random_incorrect_d[label] += (not gold_entity == random_candidate)
+
```
svlandeg: This code may be a little shorter than the original `if ... else` structure, but it certainly can't be faster: you're performing the same comparison 4 or 5 times instead of just once...

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+    for label in sorted(all_keys):
+        correct = correct_by_label.get(label, 0)
+        incorrect = incorrect_by_label.get(label, 0)
+        acc_by_label[label] = _calculate_precision(correct, incorrect)
+    return acc, acc_by_label
+
+
+def _calculate_precision(true_positives: int, false_positives: int):
+    if true_positives == 0:
```
svlandeg: Also here I'd remove the typing, because it's not really used anywhere else. But hopefully at some point we'll use typing everywhere ;-)

bin/wiki_entity_linking/entity_linker_evaluation.py
euand: > So in the definition we have here, accuracy equals precision, right?

That's right, precision here is equivalent to how accuracy was defined before.

> Should we go ahead then and remove the references to accuracy/acc and call all of it precision, also the _by_label dictionaries?

That sounds good to me!

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,247 @@
...
+    random_precision: float
+    random_accuracy_by_label: Dict
+    prior_recall: float
+    prior_precision: float
+    prior_accuracy_by_label: Dict
+    oracle_recall: float
+    oracle_precision: float
+    oracle_accuracy_by_label: Dict
+
+    def report_accuracy(self, model: str) -> str:
```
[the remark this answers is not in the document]
euand: Ah of course! I'll remove all type hints.

bin/wiki_entity_linking/entity_linker_evaluation.py
euand: Good point! The way I thought about it the comparison was fast enough to be negligible but I get that this may not be convincing. I'll change it to do only one comparison.

honnibal: I wouldn't have thought speed would be much of a factor in evaluation code? I usually like to keep the evaluation code shorter, because it's sort of a detail.

I'm a bit confused by `not gold_entity == random_candidate` though! Can this just be `gold_entity != random_candidate`?

bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -0,0 +1,182 @@
+import logging
+import random
+
+from tqdm import tqdm
```
svlandeg: We just recently ran into a bug with importing `tqdm`, cf https://github.com/explosion/spaCy/pull/4263. The (ugly and temporary) fix is to import it only within the functions themselves, could you update your code? We'll revert it back once the underlying bug is fixed in `tqdm`

bin/wiki_entity_linking/entity_linker_evaluation.py
svlandeg: so the _by_label dictionaries are completely removed now? I thought it was kind of informative to break up the performance by NER label, because it can vary quite a bit...

euand: I think I might have misunderstood the "also the _by_label dictionaries" in your earlier comment. I understood it to mean we remove all references to the `_by_label` dictionaries. I can add them back but with precision and recall 👍

svlandeg: Oh, sorry that that was unclear. I was indeed just proposing to change the terminology, following your initial change from acc to p, which made sense...

bin/wiki_entity_linking/entity_linker_evaluation.py
svlandeg: I agree with you @honnibal that speed doesn't really matter here, but in two pretty much equivalent ways of writing the same code I would still prefer the one where the same statement is not evaluated 5 times. Anyway these kind of rewritings are often a matter of taste. I think the last version committed is a nice compromise (and doesn't have the `not ... ==` anymore) ;-)

ines:
@svlandeg Is this ready to merge? 🙂

ines:
Ok cool, I'll go ahead and merge this then! Also seems like a low-risk PR anywas since it's just `bin` scripts. We can always add more in separate PRs later.

Issue #1061, opened 2017-05-16
tokenizer.add_special_case not working when special token not followed by whitespace


christian-storm:
When a special case token is followed by non-whitespace character the special token isn't recognized.  I wrote a quick test to demonstrate:

```python
import spacy
from spacy.symbols import ORTH

text = '...gimme...? that ...gimme...? or else ...gimme...?!'
nlp.tokenizer.add_special_case(u'...gimme...?', [{ORTH: u'...gimme...?'}])
#print([w.text for w in nlp(text)])
assert [w.text for w in nlp(text)] == ['...gimme...?', 'that', '...gimme...?', 'or', 'else', '...gimme...?', '!']
```
The last '...gimme...?'  is broken up into '...', 'gimme', '...', '?' by the presence of the '!'.

## Your Environment
* **spaCy version:** 1.8.2
* **Platform:** Darwin-16.5.0-x86_64-i386-64bit
* **Python version:** 3.6.0
* **Installed models:** en_depent_web_md

slavaGanzin:
This bug replicates only when you already used pipepline:
```
import spacy
from spacy.symbols import ORTH
nlp = spacy.load('en_depent_web_md')
text = '...gimme...? that ...gimme...? or else ...gimme...?!'
print([w.text for w in nlp(text)])
#['...', 'gimme', '...', '?', 'that', '...', 'gimme', '...', '?', 'or', 'else', '...', 'gimme', '...', '?', '!']
nlp.tokenizer.add_special_case(u'...gimme...?', [{ORTH: u'...gimme...?'}])
print([w.text for w in nlp(text)])
#['...gimme...?', 'that', '...gimme...?', 'or', 'else', '...', 'gimme', '...', '?', '!']
```

But:
```
import spacy
from spacy.symbols import ORTH
nlp = spacy.load('en_depent_web_md')
text = '...gimme...? that ...gimme...? or else ...gimme...?!'
nlp.tokenizer.add_special_case(u'...gimme...?', [{ORTH: u'...gimme...?'}])
print([w.text for w in nlp(text)])
#['...gimme...?', 'that', '...gimme...?', 'or', 'else', '...gimme...?', '!']
```

## Info about spaCy

* **spaCy version:** 1.8.2
* **Platform:** Linux-4.10.13-1-ARCH-x86_64-with-arch
* **Python version:** 3.6.1
* **Installed models:** en, en_depent_web_md

honnibal:
Both really helpful, thanks!

[...]

PR #4239, opened 2019-09-04
Add regression test for #1061 back to test suite

adrianeboyd:
## Description

Add regression test for #1061 back to test suite.

(Apparently lost in merge for 2.0.0rc1?)

### Types of change

Enhancement?

## Checklist


honnibal:
Hmm, really not sure how this got lost! Good catch.

PR #4245, opened 2019-09-06
Fix ValueError exception on empty Korean text.

b1uec0in:
## Description

Fixed check_spaces method to return a generator of length 0 instead of 1 on empty string.

This fixes the following error issue. 

> File ".../spacy/lang/ko/__init__.py", line 72, in \_\_call\_\_
doc = Doc(self.vocab, words=surfaces, spaces=list(check_spaces(text, surfaces)))
File "doc.pyx", line 209, in spacy.tokens.doc.Doc.\_\_init\_\_
ValueError: [E027] Arguments 'words' and 'spaces' should be sequences of the same length, or 'spaces' should be left default at None. spaces should be a sequence of booleans, with True meaning that the word owns a ' ' character following it.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/lang/ko/__init__.py
+++ b/spacy/lang/ko/__init__.py
@@ -58,7 +58,8 @@ def check_spaces(text, tokens):
             yield prev_end != idx
         prev_end = idx + len(token)
         start = prev_end
-    yield False
+    if start > 0:
+        yield False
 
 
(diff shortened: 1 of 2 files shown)


ines:
Thanks a lot! 👍

PR #4246, opened 2019-09-06
Make pos/tag distinction more clear in docs

### LINKED ISSUE #4242: Docs: difference between POS and TAG / origins of these
As shown in the table in https://spacy.io/usage/linguistic-features#pos-tagging, spacy differentiates two kinds of POS tags (POS and TAG). Probably I'm just a newbie, but is there some docs, which kind of philosophy these are based on? TAG seem to be based on Penn TreeBank tags, POS on a more universal set (as in Petrov, Slav, et al. "A Universal Part-of-Speech Tagset"). Any additional explanation on this would be helpful for the docs I guess, especially for people like me ;-)

svlandeg:
## Description
* Moved the "Part-of-speech tag scheme" infobox up on the page https://spacy.io/usage/linguistic-features
* Corrected one link to the dependencies doc (was pointing to POS doc)

### Types of change
Change to the documentation, closes #4242 

## Checklist

### DIFF
--- a/website/docs/usage/101/_pos-deps.md
+++ b/website/docs/usage/101/_pos-deps.md
@@ -3,7 +3,16 @@ the statistical model comes in, which enables spaCy to **make a prediction** of
 which tag or label most likely applies in this context. A model consists of
 binary data and is produced by showing a system enough examples for it to make
 predictions that generalize across the language – for example, a word following
-"the" in English is most likely a noun.
+"the" in English is most likely a noun. spaCy assigns tags on two different levels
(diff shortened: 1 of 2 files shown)


ines:
Thanks! 👍

poke1024:
Thank you.

ritwikmishra:
If someone is still looking for difference then 

> POS = The simple UPOS part-of-speech tag.
> TAG = The detailed part-of-speech tag.

taken from [here](https://spacy.io/usage/linguistic-features#pos-taggingl)

PR #4248, opened 2019-09-06
Allow period as suffix following punctuation

adrianeboyd:
## Description

Addresses rare cases (such as `_MATH_.`, see #1061) where the final period was not recognized as a suffix following punctuation.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/lang/punctuation.py
+++ b/spacy/lang/punctuation.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 
 from .char_classes import LIST_PUNCT, LIST_ELLIPSES, LIST_QUOTES, LIST_CURRENCY
 from .char_classes import LIST_ICONS, HYPHENS, CURRENCY, UNITS
-from .char_classes import CONCAT_QUOTES, ALPHA_LOWER, ALPHA_UPPER, ALPHA
+from .char_classes import CONCAT_QUOTES, ALPHA_LOWER, ALPHA_UPPER, ALPHA, PUNCT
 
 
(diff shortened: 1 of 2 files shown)


honnibal:
Do we have an accuracy check for this? I could imagine it messing some things up, e.g. on abbreviations.

adrianeboyd:
I don't think ellipses are affected by this. It shouldn't affect "standard" abbreviations that are in all caps. Can you think of a string that includes that includes sentence-final-ish punctuation followed by "." where you want to tokenize them together?

I'll see if I can test if there are any differences with misaligned tokens with the model training data, but I'm not sure there are going to be any instances.

honnibal:
Fair enough, I see what you mean now. @svlandeg do we have scripts for the tokenizer evaluations you did checked in somewhere?

svlandeg:
yes, here: https://github.com/explosion/spaCy/blob/master/bin/ud/run_eval.py

You can either `run_single_eval` on one specific language, or `run_all_evals` that takes a set of models and treebanks as input. Or just run the CLI script which takes a path to the UD corpora and evaluates all languages by default.

I haven't run this script in a few months so could be that it requires some updating though.

adrianeboyd:
Thanks, @svlandeg, that script looks perfect! I'll try to update it for UD 2.4 and new languages.

honnibal:
I'll go ahead and merge this, since I agree it seems pretty safe. If the evaluations show a problem we can always revert.

Issue #3944, opened 2019-07-10
tokenizer.add_special_case not working


bklippstein:
tokenizer.add_special_case is not working for me. 

```
import spacy

from spacy.symbols import ORTH, POS, X, LEMMA

nlp = spacy.load('en')

nlp.tokenizer.add_special_case('»', [{ORTH: '»', POS: X, LEMMA:'Test'  }])
nlp.tokenizer.add_special_case('«', [{ORTH: '«', POS: X, LEMMA:'Test' }])

doc = nlp('He said: »I am lying.«')

for token in doc:
    print('{:10}{:10}{:10}'.format(token.text, token.pos_, token.lemma_))
```
returns
```
He        PRON      -PRON-    
said      VERB      say       
:         PUNCT     :         
»         PUNCT     »         
I         PRON      -PRON-    
am        VERB      be        
lying     VERB      lie       
.         PUNCT     .         
«         PUNCT     «   
```

The reason why I try to change the POS tags is that they are wrong for '»'  and '«' in the german model. 

Even the code from [here](https://stackoverflow.com/questions/51766157/how-to-force-a-pos-tag-in-spacy-before-after-tagger) is not working:

```
import spacy

from spacy.symbols import ORTH, POS, NOUN, VERB

nlp = spacy.load('en')

nlp.tokenizer.add_special_case('{G}', [{ORTH: '{G}', POS: NOUN}])
nlp.tokenizer.add_special_case('{T}', [{ORTH: '{T}', POS: VERB}])

doc = nlp('This {G} a noun and this is a {T}')

for token in doc:
    print('{:10}{:10}'.format(token.text, token.pos_))
  
# Wanted output:
# This      DET       
# {G}       NOUN      
# a         DET       
# noun      NOUN      
# and       CCONJ     
# this      DET       
# is        VERB      
# a         DET       
# {T}       VERB

# My output:
# This      DET       
# {G}       NOUN      
# a         DET       
# noun      NOUN      
# and       CCONJ     
# this      DET       
# is        VERB      
# a         DET       
# {T}       NOUN   <<<<<<<<<<<<<<<<<<<
```

## Your Environment

* Operating System: Linux-5.0.0-20-generic-x86_64-with-debian-buster-sid
* Python Version Used: 3.6.8
* spaCy Version Used: 2.1.4  [cuda100]
* Environment Information:

svlandeg:
I think there's a few things going on here. 

[...]

PR #4249, opened 2019-09-06
Allow affixes as tokenizer special cases

adrianeboyd:
## Description

Modify the tokenizer to additionally check whether prefixes and suffixes are special cases.

Prefixes and suffixes are stored as LexemeC, so only the lexeme attributes are affected, other token attributes are ignored.

Related to #3944.

### Types of change

Bugfix.

## Checklist

### DIFF
--- a/spacy/tests/tokenizer/test_tokenizer.py
+++ b/spacy/tests/tokenizer/test_tokenizer.py
@@ -120,3 +120,18 @@ def test_tokenizer_add_special_case_tag(text, tokens):
     assert doc[0].tag_ == tokens[0]["tag"]
     assert doc[0].pos_ == "NOUN"
     assert doc[1].text == tokens[1]["orth"]
+
+@pytest.mark.parametrize(
+    "text,tokens",
+    [
+        ("»", [{"orth": '"'}]),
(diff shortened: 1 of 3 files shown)


adrianeboyd:
It's not satisfying that the non-affix special cases can set token attributes but the affix special cases can only set lexeme attributes.

This could be modified, of course, but I would agree that it would make sense to remove token attributes from the special cases since the token attributes are not tokenization issue.

honnibal:
I definitely want the token attributes removed from the special cases, that whole thing was a mistake. It's really confusing to have the tokenizer setting some attributes like the `POS`, `TAG` or `LEMMA`.

adrianeboyd:
After looking at this some more, I really don't think this is the way to go. I've rewritten how special cases are handled to make it more general (using Matcher internally) and I'll submit a PR soon. (The rewrite doesn't have an opinion on whether to remove the token attributes.)

The problem is that special cases only work sometimes. They only work with +/- 1 balanced prefixes and suffixes and not when separated by infixes. It's not obvious to the user when it doesn't match and it doesn't quite do what it's intended to do, I don't think.

PR #4251, opened 2019-09-07
Lemmatizer lookup dictionary for Serbian and basic tag set adde…

Pavle992:
Lemmatizer dictionary created using Apache 2.0 licenced source. Basic tagset created.
## Description

- Lookup dictionary for Serbian word-to-lemma translation added. Data is obtain from the source that is licenced under Apache 2.0. Licence is included together with owner description and description of data usage. 

- Basic tag set added to tag_map.py file

- Few tests added

## Checklist


isaric:
@Pavle992 Great minds think alike. :) I opened a PR for Croatian with the same lemmatizer source. Concerning the tag map : have you thought about mapping MSD tags to universal dependencies tags?

Pavle992:
> @Pavle992 Great minds think alike. :) I opened a PR for Croatian with the same lemmatizer source. Concerning the tag map : have you thought about mapping MSD tags to universal dependencies tags?

Good morning @isaric  hahah you are right Croatia and Serbia always go in one package 😃 
Yes definately, I plan to do more advance like msd tag-mappings at one point. 
For our use-case I found well described [msd-examples](http://nlp.ffzg.hr/data/tagging/msd-hr.htmll). However, we need to create our own mappings and and I found this [paper](https://pdfs.semanticscholar.org/c63d/2d323682070bbdec8208425b76046d795c22.pdf) as a good starging point :)

spacy/lang/sr/tag_map.py
```diff
@@ -0,0 +1,27 @@
+# coding: utf8
```
ines: This is the default tag map, right? In that case, we could probably also leave it out for now, because it's what spaCy should default to anyways.

isaric:
@Pavle992 The first link is broken for me. Could you repost, please? 
An idea that I am playing with is to write a script that would go through corpora and collect all of the MSD tags it uses and map them to the Universal dependency tags. The corpora from ffzg and ReLDI would work nicely.

spacy/lang/sr/tag_map.py
Pavle992: Yes, you're right. I've removed the file and updated PR.

Pavle992:
@isaric ofc, sorry for that, here is it: http://nlp.ffzg.hr/data/tagging/msd-hr.html

Pavle992:
@isaric I had similar idea. Basically we need to get all combinations correctly maped. I wanted also to scrape it, but this source is CC type of licence.
I will try to obtained them from Reldi's gitrepo which is Apache and maybe we can use couple of files combined to solve it. I'll ping you again if I came up with something :smile:

PR #4252, opened 2019-09-07
Adding a Croatian lemmatizer from Apache 2.0 source reldi-tager

isaric:
This PR adds a list of Croatian lemmas adapted from the reldi-tagger.
## Description

The PR adds lemma_lookup.json for Croatian, its licence and registers it. It also adds the corresponding test for the lemmatizer as well as setting up the test fixture for a Croatian tokenizer.
The list of lemmas was adapted from the reldi-tagger repository (https://github.com/clarinsi/reldi-tagger) which is licensed under Apache 2.0. I also made a fork of the repo on my account to cover future changes, if necessary.

### Types of change

This is an enhancement for the Croatian language module of spaCy.
## Checklist

### DIFF
--- a/spacy/lang/hr/__init__.py
+++ b/spacy/lang/hr/__init__.py
@@ -18,6 +18,7 @@ class CroatianDefaults(Language.Defaults):
     )
     tokenizer_exceptions = update_exc(BASE_EXCEPTIONS)
     stop_words = STOP_WORDS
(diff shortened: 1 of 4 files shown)


honnibal:
Looks good! Thanks for taking the time over the license issues. Previously we were reluctant to accept more lemmatizer data, but now that we have them separated into json files and zipped, it shouldn't be a problem.

PR #4258, opened 2019-09-08
Flush tokenizer cache when necessary

adrianeboyd:
## Description

Flush tokenizer cache when affixes, token_match, or special cases are
modified.

Fixes #4238, same issue as in #1250.

### Types of change

Bugfix.

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #4259, opened 2019-09-08
Generalize handling of tokenizer special cases

adrianeboyd:
## Description

Handle tokenizer special cases more generally by using the Matcher internally to match special cases after the affix/token_match tokenization is complete.

Instead of only matching special cases while processing balanced or nearly balanced prefixes and suffixes, this recognizes special cases in a wider range of contexts:

* Allows arbitrary numbers of prefixes/affixes around special cases
* Allows special cases separated by infixes

Existing tests/settings that couldn't be preserved as before:

* The emoticon '")' is no longer a supported special case
* The emoticon ':)' in "example:)" is a false positive again

When merged with #4258 (or the relevant cache bugfix), the affix and token_match properties should be modified to flush and reload all special cases so that they use the updated internal tokenization with the Matcher.

### Types of change

Bugfix.

## Checklist


honnibal:
Hmm I'm really not sure about this. I think this will be too slow?

adrianeboyd:
What you have is pretty broken, though, and there's no easy way to fix it within the affix tokenization functions.

honnibal:
Can you explain the problem with the imbalanced affixes? I don't really understand how it happens that the imbalanced affixes cause the special-cases to not match.

honnibal:
So this would allow special-cases that cross whitespace boundaries, right? So you can have a special case like `vice versa` to tokenize that as one word?

adrianeboyd:
That should have been more specific. It's a problem when the special case starts or ends with a prefix or suffix character. (There were examples in several issues with things like `_MATH_` and `_SPECIAL_`.)

Current behavior:

```
import spacy
nlp = spacy.load('en')
nlp.tokenizer.add_special_case("_PRODUCT_NAME_", [{"ORTH": "_PRODUCT_NAME_"}])
print([t.text for t in nlp('She asked, "What is the _PRODUCT_NAME_?"')])
# ['She', 'asked', ',', '"', 'What', 'is', 'the', '_', 'PRODUCT_NAME', '_', '?', '"']
```

In each loop, the tokenizer considers lopping off one prefix and/or one suffix, so it lops off the prefix `_` before it gets to a point with the suffixes that it would have a chance of identifying the special case.

adrianeboyd:
Yes, you currently can do `vice versa`, but that would be easy to filter out.

adrianeboyd:
Wait, I thought you meant that was bad. No, there's no restriction there as long as you're not changing the text.

adrianeboyd:
Special cases won't match in between existing token boundaries, so you can't match `ce ver` out of `vice versa`.

adrianeboyd:
I also haven't tried to test the doc reallocation code with a pathological example, I just copied most of it from the retokenizer.

adrianeboyd:
Aside from the fact that `")` is a terribly confusing emoticon within typical text, that case actually failed even though I was trying to match it and I am not 100% sure why.

honnibal:
Aaah, I see the problem now.

If the problem only occurs when we have special-cases that contain affixes, can we make it so that we only use this match logic for those special-cases?

It feels bad because it's quite tricky (similar stuff ends up handled in quite different ways), but it means there would be almost no performance penalty from the change, because only a few special cases (that would otherwise not work) get handled with the fallback matcher.

adrianeboyd:
There is still the problem that special cases can have infix boundaries:

```
import spacy
nlp = spacy.load('en')
nlp.tokenizer.add_special_case("SPECIAL_", [{"ORTH": "SPECIAL_"}])
print([t.text for t in nlp('a-SPECIAL_')])
# ['a', '-', 'SPECIAL', '_']
```

But it's possible if you also checked for special cases within the infixes and then used the matcher for any cases containing affixes, you could get it all to work.

adrianeboyd:
Detecting all special cases with the matcher, the overall tokenization time for the UD tests with most supported languages (exceptions: ko, vi, zh) is about twice as long as before. The differences vary a lot by language depending more or less on the number of tokenizer exceptions. The most extreme example is Indonesian, which has probably ~10000 exceptions and is 15x slower. I will try adding special cases without affixes back as before. (Indonesian, where most exceptions contain `-` may not get much faster.)

adrianeboyd:
I'm going to complicate this PR a bit by merging in the PhraseMatcher and UD script changes, plus the recent changes I've made.

I've optimized about everything I know how to optimize and the speed hasn't improved enough. It's still about 70% slower than before. I'm hoping there might be some improvements I have missed?

To run this on all the UD corpora to test:

```
$ bin/ud/run_eval.py -m -t out.csv ud-treebanks-v2.4
```

honnibal:
Okay so, as far as further optimisation goes, I think the next step is to break up the `_apply_special_cases` function so that we have a couple of subfunctions, and we'll be able to profile more accurately. This will tell us which bits to focus on for the optimisation.

You probably want to have three functions: one that does all of the setup and sets up the `span_data` data structure, a second that does that while loop, and a third which fiddles with the doc. 

The first function would be easy to optimise, as there's lots of Python in there. But if it's not taking much of the runtime there won't be any point. So, we'd need to profile to check.

adrianeboyd:
Profiling showed that the problem wasn't in the special cases handling at all, but that it was a caching bug. :cry:  (I moved where `has_special` was defined and then didn't check the value correctly in `_save_cached`, so nothing was getting cached.)

Now the performance looks totally acceptable. I'm running the UD scripts again now.

honnibal:
@adrianeboyd Sort of good news though! I thought the speed problems might end up kind of intractable.

adrianeboyd:
It's such a dumb kind of bug, but I guess I've optimized all kinds of other things kind of accidentally now?

I need to do another round of profiling, but I think it's closer to only being 5% slower for English and it can be optimized a bit more.

honnibal:
@adrianeboyd How do the speed numbers look on other languages now?

If this is ready I think we should consider getting it merged. It can be tempting to procrastinate on merging large changes, but it's worse to have them sitting in a PR.

adrianeboyd:
Okay, with the UD script it's about 11% slower overall. (I just learned about the `<details>` tag!) This not-very-exact clock time, not a careful timing test, although I tried to keep the environment very similar.

<details>
<summary>UD tokenization timing comparion</summary>

text_path | master | new | ratio new / master
---|---|---|---
fr_partut-ud-train.txt | 0.19 | 0.42 | 2.21
ur_udtb-ud-train.txt | 0.48 | 1 | 2.08
fa_seraji-ud-train.txt | 0.65 | 1.3 | 2.00
fr_sequoia-ud-train.txt | 0.61 | 1.19 | 1.95
it_partut-ud-train.txt | 0.45 | 0.78 | 1.73
fr_gsd-ud-train.txt | 4.46 | 7.55 | 1.69
en_partut-ud-train.txt | 0.36 | 0.59 | 1.64
pt_bosque-ud-train.txt | 2.03 | 3.22 | 1.59
pt_gsd-ud-train.txt | 2.56 | 4.05 | 1.58
nl_lassysmall-ud-train.txt | 0.95 | 1.4 | 1.47
it_isdt-ud-train.txt | 2.12 | 2.89 | 1.36
nl_alpino-ud-train.txt | 2.48 | 3.38 | 1.36
ca_ancora-ud-train.txt | 2.86 | 3.86 | 1.35
de_gsd-ud-train.txt | 3.59 | 4.7 | 1.31
da_ddt-ud-train.txt | 0.98 | 1.28 | 1.31
fr_spoken-ud-train.txt | 0.14 | 0.18 | 1.29
hu_szeged-ud-train.txt | 0.31 | 0.38 | 1.23
pl_lfg-ud-train.txt | 1.62 | 1.95 | 1.20
bg_btb-ud-train.txt | 1.54 | 1.82 | 1.18
ga_idt-ud-train.txt | 0.17 | 0.2 | 1.18
en_ewt-ud-train.txt | 1.61 | 1.89 | 1.17
fi_ftb-ud-train.txt | 2.31 | 2.7 | 1.17
pl_pdb-ud-train.txt | 4.64 | 5.39 | 1.16
tr_imst-ud-train.txt | 0.72 | 0.83 | 1.15
fi_tdt-ud-train.txt | 3.59 | 4.11 | 1.14
es_gsd-ud-train.txt | 4 | 4.57 | 1.14
sv_lines-ud-train.txt | 0.5 | 0.56 | 1.12
ar_padt-ud-train.txt | 2.6 | 2.91 | 1.12
it_vit-ud-train.txt | 1.92 | 2.13 | 1.11
ro_rrt-ud-train.txt | 2.36 | 2.59 | 1.10
es_ancora-ud-train.txt | 4.07 | 4.46 | 1.10
it_postwita-ud-train.txt | 1.25 | 1.36 | 1.09
et_ewt-ud-train.txt | 0.26 | 0.28 | 1.08
ta_ttb-ud-train.txt | 0.13 | 0.14 | 1.08
ru_gsd-ud-train.txt | 1.55 | 1.65 | 1.06
lt_alksnis-ud-train.txt | 0.34 | 0.36 | 1.06
ru_taiga-ud-train.txt | 0.38 | 0.4 | 1.05
el_gdt-ud-train.txt | 0.67 | 0.7 | 1.04
sv_talbanken-ud-train.txt | 1 | 1.04 | 1.04
ru_syntagrus-ud-train.txt | 15.41 | 16.02 | 1.04
no_nynorsk-ud-train.txt | 2.64 | 2.73 | 1.03
he_htb-ud-train.txt | 1.41 | 1.45 | 1.03
no_bokmaal-ud-train.txt | 3.52 | 3.61 | 1.03
en_lines-ud-train.txt | 0.4 | 0.41 | 1.03
ja_gsd-ud-train.txt | 2.87 | 2.94 | 1.02
sr_set-ud-train.txt | 0.95 | 0.97 | 1.02
cs_fictree-ud-train.txt | 1.67 | 1.68 | 1.01
uk_iu-ud-train.txt | 1.74 | 1.75 | 1.01
et_edt-ud-train.txt | 5.95 | 5.97 | 1.00
lv_lvtb-ud-train.txt | 3.19 | 3.2 | 1.00
af_afribooms-ud-train.txt | 0.26 | 0.26 | 1.00
en_gum-ud-train.txt | 0.01 | 0.01 | 1.00
lt_hse-ud-train.txt | 0.07 | 0.07 | 1.00
mr_ufal-ud-train.txt | 0.04 | 0.04 | 1.00
no_nynorsklia-ud-train.txt | 0.16 | 0.16 | 1.00
sk_snk-ud-train.txt | 1.2 | 1.2 | 1.00
sl_sst-ud-train.txt | 0.23 | 0.23 | 1.00
te_mtg-ud-train.txt | 0.07 | 0.07 | 1.00
cs_cac-ud-train.txt | 7.96 | 7.95 | 1.00
ro_nonstandard-ud-train.txt | 2.85 | 2.84 | 1.00
hr_set-ud-train.txt | 2.17 | 2.16 | 1.00
sl_ssj-ud-train.txt | 1.66 | 1.65 | 0.99
hi_hdtb-ud-train.txt | 1.31 | 1.3 | 0.99
ko_kaist-ud-train.txt | 8.92 | 8.74 | 0.98
ko_gsd-ud-train.txt | 1.79 | 1.74 | 0.97
cs_pdt-ud-train.txt | 22.45 | 21.64 | 0.96
cs_cltt-ud-train.txt | 0.57 | 0.54 | 0.95
id_gsd-ud-train.txt | 3.63 | 3.33 | 0.92
total | 157.55 | 174.87 | 1.11

</details>

The most noticeable differences are for languages with a lot of exception matches, typically involving apostrophe or period (pretty much all Germanic and Romance languages). It's enough slower for enough frequently used languages that I'm not quite happy with it yet. I also have no idea what's going on with Farsi and Urdu and should probably figure that out, too.

Here's the raw table:

[tokenization-timing-comparison.csv.txt](https://github.com/explosion/spaCy/files/3693274/tokenization-timing-comparison.csv.txt)

honnibal:
Okay, I'm happy with that performance 🎉 . How much should we worry about different behaviour? Maybe you should run it over a few million words of text and just see what comes out different?

@svlandeg and @ines , what are your thoughts, now that this is ready for a more detailed review? I think to me the `add_special_cases` function in `tokenizer.pyx` is looking a little long, so maybe that could be broken up.

adrianeboyd:
The average tokenization f-score goes from 0.9587 to 0.9588. (German goes from .99 to 1, everything else stays the same, at least to two decimal points.)

I wouldn't expect a whole lot of difference in performance for the default exceptions, but they're probably a smidge more consistent (maybe the German difference is a few `` or '' quotes?) and the user-added ones should behave more consistently when they contain affixes.

spacy/tokenizer.pyx
```diff
@@ -174,23 +211,124 @@ cdef class Tokenizer:
...
 
-    def _reset_specials(self):
+    def _flush_specials(self):
         for k in self._specials:
             cached = <_Cached*>self._specials.get(k)
             del self._specials[k]
             if cached is not NULL:
                 self.mem.free(cached)
 
+    cdef int _apply_special_cases(self, Doc doc):
```
svlandeg: I guess this can be `void` return type?

spacy/tokenizer.pyx
```diff
@@ -174,23 +211,124 @@ cdef class Tokenizer:
...
+        cdef stdset[int] seen_tokens
+        stdsort(original.begin(), original.end(), len_start_cmp)
+        cdef int orig_i = original.size() - 1
+        while orig_i >= 0:
+            span = original[orig_i]
+            if not seen_tokens.count(span.start) and not seen_tokens.count(span.end - 1):
+                filtered.push_back(span)
+            for seen_i in range(span.start, span.end):
+                seen_tokens.insert(seen_i)
+            orig_i -= 1
```
svlandeg: So if I understand correctly, this matches the biggest spans first, and if two spans are equally big, it gives preference to the **last** one? Because with equal length you're sorting on `a.start < b.start` but you're processing the list backwards. Just checking whether this was the intended behaviour. Personally I think I would have given preference to the first span in the `doc` (when equal length), but there's probably not a "best" answer, so maybe it doesn't matter either way.

spacy/tokenizer.pyx
```diff
@@ -174,23 +211,124 @@ cdef class Tokenizer:
...
+        if not modify_in_place:
+            memcpy(doc.c, tokens, max_length * sizeof(TokenC))
+        for i in range(doc.length + offset, doc.length):
+            memset(&doc.c[i], 0, sizeof(TokenC))
+            doc.c[i].lex = &EMPTY_LEXEME
+        doc.length = doc.length + offset
+        return True
+
+    cdef void _filter_spans(self, vector[MatchStruct] &original, vector[MatchStruct] &filtered, int doc_len) nogil:
+
```
svlandeg: This is a nifty function, I like it! Should we have some test cases with overlapping spans just to be sure?

spacy/tokenizer.pyx
```diff
@@ -174,23 +211,124 @@ cdef class Tokenizer:
...
+        for span in spans:
+            rule = self._rules.get(span.text, None)
+            span_length_diff = 0
+            if rule:
+                span_length_diff = len(rule) - (span.end - span.start)
+            if span_length_diff > 0:
+                modify_in_place = False
+            curr_length += span_length_diff
+            if curr_length > max_length:
+                max_length = curr_length
```
svlandeg: `max_length` will always end up holding the final value of `curr_length`, no? So I don't think we actually need two different variables ?

spacy/tokenizer.pyx
```diff
@@ -174,23 +211,124 @@ cdef class Tokenizer:
...
+            span_data[span.start] = (span.text, span.start, span.end, span_length_diff)
+        # Modify tokenization according to filtered special cases
+        # If modifications never increase doc length, can modify in place
+        if modify_in_place:
```
svlandeg: Alternatively, do without `modify_in_place` variable and just check `curr_length == doc.length`

spacy/tokenizer.pyx
adrianeboyd: No, it's possible that the document in temporarily longer in the middle of processing, like a bunch of tokens are split and then a bunch of tokens are joined.

spacy/tokenizer.pyx
adrianeboyd: The idea is that this function is identical to `util.filter_spans()`. We should add the same tests for this function, too.

I have to think about whether it's doing the right thing in those cases. Some of the details are simpler because it's not possible to have two spans over the same tokens. (There's only one special rule for any given token sequence.)

svlandeg:
@honnibal @adrianeboyd : I think it looks great! Sure there's some hit in speed, but I do think this is a more clean and consistent solution. I just added a few very minor comments as review. 

And I agree with Matt that perhaps `add_special_cases ` could be broken up in pieces. You have the first bit with `c_filtered` and `c_matches`, then looping over the `spans`, then the `while` with `span_data`, and then the post-processing. By splitting it up, those variables will have much smaller scope.

spacy/tokenizer.pyx
svlandeg: Gotcha - feel free to resolve and ignore :-)

spacy/tokenizer.pyx
honnibal: This is a good thing to highlight, as there's some subtleties to Cython signatures that are extremely non-obvious. It's why we often have `int` as the return type of functions and `except -1`. Maybe this is resolved in current versions of Cython, but my understanding is:

Cython uses the `except` value as the error code, so that you can raise errors within the function. Cython will call the Python error checking functions if the return value is equal to the exception value you set. If you have a C return type and don't specify a value as the exception value, when there's errors within the function, they're swallowed and you get that spam on your terminal about unraisable exceptions. Instead of specifying a particular value, you can also use `except *`. This forces Cython to check the Python error status after every function invocation. which introduces a bit of overhead.

If you use `void` as the return type, you have to use `except *`. So using `void` as a return type is slower --- which is extremely counter-intuitive, but there you go. The problem doesn't come up if you're making your function `nogil`, as a `nogil` function can't be raising any Python errors anyway (it can't call into any Python routines).

A similar problem occurs with the `bint` type. If you mark a function as `bint`, you don't really have a good value to use as an error. So `int` is usually better, so you can make it `except -1`.

adrianeboyd: I think I used `int` because all the other functions I'd seen were `int`, but I didn't understand the details.

So this should stay `int` and I should add `except -1`? And `_filter_spans()` can stay `void ... nogil` and the cmp functions can stay `bint ... nogil`?

spacy/tokenizer.pyx
adrianeboyd: You're right that it prefers the later longer span rather than the earlier one. If we change this here, we should also change `util.filter_spans()`. I don't think that this comes up a lot in practice with the tokenizer because the matches tend to be short and not overlapping (one real case might be `'''` becomes `' ''` rather than `'' '`), but it is relevant for `util.filter_spans()`, which is used for filtering things like entities and doesn't do exactly what it says, either.

spacy/tokenizer.pyx
svlandeg: Thanks for the explanation @honnibal ! I definitely did not realise all of that ...

spacy/tokenizer.pyx
svlandeg: Ah, you're right. So we should either change the sorting behaviour or the comment, at least. Any thoughts, @honnibal ? (https://github.com/explosion/spaCy/blob/master/spacy/util.py#L629 sorts on size small->big and start first->later, but then that order is reversed)

adrianeboyd: I think we should change the behavior for both to prefer the earliest span, as seems to be intended from the documentation. I think we can just swap the compare here and change `span.start` to `-span.start` in the other? I think @ines is the original author?

adrianeboyd:
I would like to convert `span_data` to a C++ data structure in order to make more of the refactored functions `nogil`, but my previous attempt ended up with a lot of windows/mac issues that were hard to debug. I'll give it another try tomorrow, but I don't think it's a huge bottleneck in the scheme of things.

spacy/tokenizer.pyx
ines: @adrianeboyd Yes, good point, I hadn't realised that the behaviour didn't actually match the documentation. (Sorry about that.) So yes, let's change that. 

Also, if there's anything we can combine here to avoid code duplication, that'd be good, too. I've also always wanted a more powerful `filter_spans` helper that implements different ways of filtering (prefer longest, prefer shortest, maybe prefer as many as possible?).

adrianeboyd: I didn't think about this problem when I was reimplementing it, either!

I originally used the method from `util` and only added this one to try to make a little faster. It would obviously be much nicer to just maintain one method, hmm. I'm using the `MatchStruct` struct, but they're really just spans underneath and there's no need for it to be part of `PhraseMatcher`.

I suppose we could create a top-level `SpanC` kind of struct instead of `MatchStruct` and move the `util` method to `Span` or `Doc`? The `util` method doesn't prevent you passing in spans from different docs, either, so you can get really weird output if you mix things up and that might be nice to check or prevent.

adrianeboyd: `Entity` is already there. Does it make sense to use it here, too? Rename it, though, maybe?

honnibal:
I think we should probably merge this onto develop?

adrianeboyd:
Yes, I think that would be okay. There are still a few weird slow cases, but nothing way out of line. I decided that replacing `span_data` would be too much of a pain, at least with other priorities right now.

Issue #2710, opened 2018-08-27
Spanish lemmatization


chozelinek:
I've been analysing a large amount of texts in Spanish and I've realised of several behaviours which are a bit weird regarding lemmatisation (in comparison with English at least).

Maybe, some issues could be avoided if the lemmatisation used the POS information to distinguish between word classes. If one loads a blank model, one gets lemmas.

See the examples below to reproduce the behaviour and my comments.

## How to reproduce the behaviour

```python
import spacy
nlp = spacy.load('es_core_news_md')  # same behaviour with spacy.blank('es')
```

```python
text1 = "Con estos fines, la Dirección de Gestión y Control Financiero monitorea la posición de capital del Banco y utiliza los mecanismos para hacer un eficiente manejo del capital."
for token in nlp(text1):
    print(token.text, token.lemma_, token.pos_)
```

Word | Lemma | POS | Comments
---|---|---|---
Con | Con | ADP
estos | este | DET
fines | finar | NOUN | instead of "fin" it uses the infinitive of a verb meaning to die
, | , | PUNCT
la | lo | DET
Dirección | Dirección | PROPN
de | de | ADP
Gestión | Gestión | PROPN
y | y | CONJ
Control | Control | PROPN
Financiero | Financiero | PROPN
monitorea | monitorea | VERB | not lemmatized, expected "monitorear"
la | lo | DET
posición | posición | NOUN
de | de | ADP
capital | capital | NOUN
del | del | ADP
Banco | Banco | PROPN
y | y | CONJ
utiliza | utilizar | VERB
los | lo | DET
mecanismos | mecanismo | NOUN
para | parir | ADP | "para" is a very frequent preposition in Spanish, here is lemmatized as the infinitive form of the verb to give birth, the parser got it right, the lemmatizer didn't
hacer | hacer | VERB
un | uno | DET
eficiente | eficiente | ADJ
manejo | manejar | NOUN | Manejo is an ambiguous word, because it could be the first person singular of present or a noun. In this case is the noun, as the POS information tells us, so the expected lemma would be manejo, and not the infinitive, the parser got it right, the lemmatizer didn't
del | del | ADP
capital | capital | NOUN
. | . | PUNCT

```python
text2 = "Incentivar la cultura de retroalimentación continua , nos dirige hacia el desarrollo de los talentos de los empleados, así como también, a la conformación de equipos de alto desempeño."
for token in nlp(text2):
    print(token.text, token.lemma_, token.pos_)
```

[...]

Issue #2668, opened 2018-08-14
💫 Improve rule-based lemmatization and replace lookups


ines:
This is an enhancement issue that's been high on our list for a while, and something we'd love to tackle for as many languages as possible. At the moment, spaCy only implements rule-based lemmatization for very few languages. Many of the other languages to support lemmatization, but only use lookup tables, which are less reliable and don't always produce good results. They're also large and add significant bloat to the library.

## Existing rule-based lemmatizers

| Language | Source |
| --- | --- |
| English | [`lang/en/lemmatizer`](https://github.com/explosion/spaCy/tree/master/spacy/lang/en/lemmatizer) |
| Greek | [`lang/el/lemmatizer`](https://github.com/explosion/spaCy/tree/master/spacy/lang/el/lemmatizer) |
| Norwegian | [`lang/nb/lemmatizer`](https://github.com/explosion/spaCy/tree/master/spacy/lang/nb/lemmatizer) |

## Most problematic lookup lemmatizers

Just my subjective selection, but these are the languages we want to prioritise. Of course, we'd also appreciate contributions to any of the other languages!

| Language | Related issues |
| --- | --- |
| German | #2486, #2368, #2120 |
| ~~French~~ | ~~#2659, #2251, #2079~~ |
| Spanish | #2710 |

mauryaland:
Hello,

I will do a PR soon to introduce a rule-based lemmatization for French language based on the following [lemme dictionary](http://infolingu.univ-mlv.fr/DonneesLinguistiques/Dictionnaires/telechargement.html) which is under this [license](http://infolingu.univ-mlv.fr/DonneesLinguistiques/Lexiques-Grammaires/lgpllr.html). I have created nouns, adjectives and adverbs files, largely inspired by the English lemmatizer structure, and would like to let the lookup table do the job for the verbs. 

However, since I introduced those files, the lemmatizer does not look at the lookup table and therefore does not return the infinitive form of the verb when calling the lemma_ method on a verb token.  I don't know where this behaviour comes from. I have checked several scripts from lemmatizer.py to language.py but I have no clue. 

Could you please take a look? Here is the link of [the concerned branch](https://github.com/mauryaland/spaCy/tree/patch-2/spacy/lang/fr). 

Thank you in advance for your help,

Amaury

sj660:
[...]

Issue #4255, opened 2019-09-07
Spanish is giving the infinitive verb as the lemma for a SCONJ


sontek:
## How to reproduce the behaviour
```
>>> import spacy
>>> nlp = spacy.load('es')
>>> sentence = "sencillo como habrían deseado."
>>> doc = nlp(sentence)
>>> for token in doc:
...     print(token.text, token.lemma_, token.pos_)
...
sencillo sencillo ADJ
como comer SCONJ
habrían haber AUX
deseado desear VERB
. . PUNCT
>>>
```

`como` in this case is conjunction in this case but the lemma is showing `comer` which is the infinitive verb.

## Info about spaCy

* **spaCy version:** 2.1.8
* **Platform:** Darwin-18.0.0-x86_64-i386-64bit
* **Python version:** 3.6.7
* **Models:** es

* Operating System: OSX
* Python Version Used: 3.6

ines:
Spanish currently only uses simple lookup tables for lemmatization, so it's limited in ambiguous cases like this. 

See #2668 for details on the planned changes to the lemmatization. You might also want to check out @guadi1994's talk on the planned rule-based lemmatization strategy for Spanish and German: https://www.youtube.com/watch?v=88zcQODyuko&list=PLBmcuObd5An4UC6jvK_-eSl6jCvP1gwXc&index=7

sontek:
When people start looking into this, I did find that the stanford NLP lemmatizer picks this up correctly. It uses this under the hood:

https://github.com/UniversalDependencies/UD_Spanish-AnCora

this is the code I used to test it:

```
>>> import stanfordnlp
>>> stanfordnlp.download('es')
>>> nlp = stanfordnlp.Pipeline("es")
>>> sentence = "sencillo como habrían deseado."
>>> doc = nlp(sentence)
>>> doc.sentences[0].print_tokens()
<Token index=1;words=[<Word index=1;text=sencillo;lemma=sencillo;upos=ADJ;xpos=ADJ;feats=Gender=Masc|Number=Sing;governor=0;dependency_relation=root>]>
<Token index=2;words=[<Word index=2;text=como;lemma=como;upos=SCONJ;xpos=SCONJ;feats=_;governor=3;dependency_relation=mark>]>
[...]

Issue #4254, opened 2019-09-07
Spanish is making abbreviations as the end of a sentence


sontek:
## How to reproduce the behaviour

```
>>> import spacy
>>> nlp = spacy.load('es')
>>> sentence = "¿Podría hacer el favor de comunicarme con Washington, EE. UU. ?"
>>> doc = nlp(sentence)
>>> len(list(doc.sents))
3
```

## Info about spaCy

* **spaCy version:** 2.1.8
* **Platform:** Darwin-18.0.0-x86_64-i386-64bit
* **Python version:** 3.6.7
* **Models:** es

* Operating System: OSX
* Python Version Used: Python3.6

ines:
Merging this with #3052. It's less useful to have these reports and test cases as separate issues – they're typically not immediately actionable, because these things come down to the statistical models. (In this case, the parser, which is used for sentence segmentation – you can always replace it with the rule-based `sentencizer` if that works better for your data.)

Maybe inaccuracies here point to a deeper problem, maybe they're just something that has to be accepted because of the training data that's available and the fact that models are statistical. (Even if a model is 90% accurate, every 10th prediction it makes will be wrong.)

PR #4265, opened 2019-09-09
Rule-based lemmatization in Spanish

pablodms:
Replaces lookup table based lemmatization in Spanish to rule-based lemmatization using Wiktionary ES  latest dump.

## Description
Parsing process is explained in the following [Jupyter Notebook](https://github.com/pablodms/wictionary_es_dump/blob/master/Dump%20lemmas%20from%20Wicktionary%20ES.ipynb). This notebook parses a downloaded dump file from [Wiktionary ES](https://dumps.wikimedia.org/eswiktionary/latest/), extracts metadata for each term in Spanish language and finally writes "lemma_exc.json" which contains mappings from variations to root terms for adjectives, adverbs, nouns, verbs, pronouns and determinants .

Then, previous lemma_exc.json and simple lemma_index.json and lemma_rules.json have been added to spacy/lang/es/lemmatizer.  _\_\_init\_\_.py_ file in spacy/lang/es has been modified to load these new files.

Also, I have modified lemmatizer in spacy to add support to DET and ADV Pos Tags, needed for a correct lemmatization for determinants and adverbs in Spanish.

This way issue #2668 for Spanish lang should have been addressed.

Changes have been tested against sentences written in #2710 producing more coherent lemmatization.

### Types of change
Enhacement

## Checklist

- [ ] I ran the tests, and all new and existing tests passed. -> I got 59 fails both in master and modified branch.


pablodms:
I have misspelled "pron" tag as "pronoun", latest commit fixes this issue.

honnibal:
@pablodms Sucks :(. You can always make a package that people can use as an extension?

sontek:
Has there been any progress towards splitting this out into its own extension?  I'm have a significant amount of issues with Spanish and this seems like it might help?  Here are the problems I'm having:

#4341
#4340
#4255
#4254
#4253

But it seems like a lot of them would address both #3052 and your original idea for fixing #2668

sontek:
Also, @pablodms Looks like you had some good stuff in here outside of the wiktionary stuff that might be worth merging?  Do you want to send a separate PR for some of your rules and the implementation of DET and ADV?

pablodms:
Hello @honnibal and @sontek,

Sorry for the delay,

These days I have been thinking about how to integrate wiktionary data without breaking the license terms.  I believe that if I only put the code to download, parse and finally generate lemmatizer files in the spacy project, I will not be breaking license since **spacy will not be distributing code based in Wiktionary data**, only code that is able to download this data and use it only if requested by user. The final user will then be able to locally configure and use the Spanish lemmatizer using Wiktionary data. 

Some issues related to this idea are: time (1-2 minutes) and disk space (around 1GB) taken  to parse dump file, and local privileges to download, create and copy needed files for lemmatizer.

What do you think @honnibal? Should I make a new pull request following this strategy?

sontek:
@honnibal Any chance we can consider re-opening this PR and getting it merged?   I just noticed that spacy already relies on this same dataset for other languages:

https://github.com/explosion/spaCy/blob/e0cf4796a505bd26e8fcb95d23b89fd7eca3be0a/spacy/lang/el/get_pos_from_wiktionary.py#L5

https://github.com/explosion/spaCy/blob/146dc2766a069c8fe9e91c801f9695c2e262f742/spacy/lang/sv/morph_rules.py#L7

versae:
+1 to merge this!

I've been running into similar issues and would love to have this merged, or at least living in its own package. Moreover, if GPLv2 or LGPL-LR were compatible, [UDLexicons 0.2](http://pauillac.inria.fr/~sagot/index.html#udlexicons) could be used, which include a couple of Spanish lexicons: UDLex_Spanish-Apertium (324925 words, GNU GPL v2) and UDLex_Spanish-Leffe (843426 words, LGPL-LR).

On the other hand, our project [spacy-affixes](https://github.com/linhd-postdata/spacy-affixes/) (a spaCy Pipeline to split clitics) uses Freeling rules and delegates the download of the data to the user so the license is honoured.

pablodms:
Hello @versae,

Since I think this request has been abandoned, I will try to develop a separate package following the idea of your project: implement lemmatizer as a spaCy pipeline stage. I will tell if I get any progress.

Thanks for your suggestions.

versae:
That's great news, @pablodms. A new pipeline would come in handy and if you need a hand with that just let me know. We could join forces!

pablodms:
I have developed a basic version and uploaded it to pip, @versae. In the [repository](https://github.com/pablodms/spacy-spanish-lemmatizer) there are simple instructions to deploy the package. I have no experience in developing Python packages, in fact, this is my first attempt, so any help would be greatly appreciated. Your proyect has been a great inspiration, by the way.

In addition, downloading (~65MB compressed file) and parsing (~900MB decompressed file) dump files is currently **SLOW**, so it can take several minutes.

versae:
That's awesome, @pablodms! Thanks for releasing it in such a short time :) Maybe in the future we could add the lemmas from UDLexicons too. I have a parser of the format and would not be a lot of work.

pablodms:
I can, @versae, add you as a contributor to the [spanish-lemmatizer](https://github.com/pablodms/spacy-spanish-lemmatizer) project so you can include your parser and your experience if you wish.

Issue #3971, opened 2019-07-16
💫 Proposal: API for efficient serializable dictionaries and lookup tables


ines:
This topic came up while talking to @polm at spaCy IRL about how to best incorporate the Unidic dictionary for Japanese. Since the licensing allows it, we ideally want to be shipping it within a model package – and we realised that we'd probably be able to provide an efficient API for this within the core library that could also be useful for many other tasks.

## Problem
Some languages and use cases may need to depend on large dictionaries or lookup tables for tokenization, lemmatization or other custom features. Including the linguistic resources with the spaCy core library itself is problematic because it creates a lot of bloat (see #3258). It's also not an option for very custom resources. For instance, a user may want to build a biomedical model and include a large dictionary to use in a custom pipeline component. 

Importing large dictionaries and serializing them with the model is currently possible out-of-the-box using the `EntityRuler`, or a custom pipeline component with [custom serialization methods](https://spacy.io/usage/saving-loading#serialization-methods). However, this is something a user would have to implement themselves – including how to make the lookup as fast and as efficient as possible.

## Proposed solution
spaCy should provide an API for adding, accessing and serializing large dictionaries and lookup tables.

The lookup methods and data should **live within the `Vocab` object** to make it  all accessible beyond pipeline components (e.g. in a custom tokenizer). It also seems like a much better fit conceptually: the vocab already holds the string store and vectors, so it makes sense for it to also hold the lookup tables. When the `nlp` object is serialized, the data will be written to the `/vocab` directory saved to disk.

### Advantages
* User lookups and dictionaries can take advantage of spaCy's built-in hashing. 
* Serialization out-of-the-box to make it easy to ship dictionaries and lookups with custom models.
* Since the API isn't tied to any specific internal use case, the lookups can be implemented anywhere – in a custom component, the tokenizer, the lemmatizer etc.
* We can finally consider moving the lemmatizer lookup tables and extensive tokenizer exception lists out of the library and provide them as separate files that can be loaded in, or as part of the statistical models we ship.

### Implementation details: set membership with Bloom filters

[...]

Issue #3248, opened 2019-02-08
pickling of PhraseMatcher not resolved yet in version '2.1.0a6' (nightly build)


Abhijit-2592:
#1971 Issue show that pickling is solved for PhraseMatcher. But, there still seems to be a bug in version **2.1.0a6**

Minimal reproducible example
```python
import spacy
from spacy.matcher import PhraseMatcher
import pickle

nlp = spacy.load("en")
matcher = PhraseMatcher(nlp.vocab, attr='LOWER')
matcher.add('president', None, nlp(u"Barack Obama"), nlp(u"Trump"))
matcher.add('health', None, nlp(u"health care reform"), nlp(u"hospital"))
print(len(matcher)) # shows 4
with open("./phrase_matcher.pkl", 'wb') as f:
    pickle.dump(obj=matcher,file=f)

with open("./phrase_matcher.pkl", 'rb') as f:
    loaded_matcher = pickle.load(f)
print(len(loaded_matcher)) # shows 0
```

## One more Bug:

According to documentation: `len(matcher)` should show 2 (only number of rule keys). But here it shows 4 (number of rules)

Abhijit-2592:
@ines  Does this commit #3252 resolve it? How can get the latest one?

ines:
Yes, my PR in #3252 should address my fix – we just need to review it and once it's merged, it'll be included in the next nightly. If you can't wait to test it, you can also build my `fix/phrasematcher-pickle-len` branch from source.

Abhijit-2592:
@ines thanks a lot for your quick response

PR #4268, opened 2019-09-10
Bloom-filter backed Lookup Tables

polm:
## Description

This is a working implementation of the lookup tables discussed in #3971. They're basically dicts that use the Bloom filters recently added to preshed, which should help with speed when misses are frequent, as is expected for language data lookup tables. There's a lot of room for improvements left,  such as changing the OrderedDict base to a PreshedMap.

One issue that came up during the implementation of this is that while the proposal indicates lookups can rely on only being accessed with integer ids, in practice many uses of lookup tables are in language-specific classes where a Vocab object (or StringStore) may not be available. For now I added methods to access the Table with a string and handle the hashing in the Table class directly, but it's kind of awkward, and likely results in repeated hashing that is best avoided. We should be able to get rid of it eventually, but for some of the more complicated Lemmatizers it might take a little time.

Currently four tests are failing. One is probably not a problem and the other three are not obviously related (I think I may have merged something poorly). The failing tests:

- [test_issue3248_2](https://github.com/explosion/spaCy/blob/e367864e59ed366adb8f1b416f91828c05eac3a0/spacy/tests/regression/test_issue3001-3500.py#L123)/ #3248 : This gives an error because the Bloom Filter can't be pickled. I don't think that's a problem, but what should be done with test? I guess we can exclude the lookups from serialization?
- `spacy/tests/doc/test_retokenize_merge.py`: Two tests failing related to token merging and the effect on UD/IOB tags. Not sure what's causing this.
- [test_issue_1061](https://github.com/explosion/spaCy/blob/e367864e59ed366adb8f1b416f91828c05eac3a0/spacy/tests/regression/test_issue1001-1500.py#L16) / #1061: This is about exceptions in the Tokenizer, which sounds related but doesn't seem to touch Lookups or Tables at all, not sure what's up.

Anyway, I will be looking into these failing tests more, but any feedback on the implementation in general or string-based access in particular would be great.

### Types of change

Enhancement

## Checklist

- [ ] I ran the tests, and all new and existing tests passed. (as noted above, there are some failing)
- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information. (I think updated docs aren't required immediately but should be added to encourage the uses proposed in #3971. They aren't in this PR right now.)


ines:
Thanks a lot! And I think you might have ended up with a slightly outdated `master` with inconsistent state. Just updated the PR so hopefully those failures are resolved now.

honnibal:
I think you probably want to base this off develop. I've got preshed 3.0.0 up now, and have released a new version of Thinc that's compatible with it. develop has the updated dependencies.

polm:
Merging in develop fixed all the tests on my machine, thanks! :tada:

spacy/lookups.py
```diff
@@ -143,15 +157,88 @@ def from_dict(cls, data, name=None):
         self.update(data)
         return self
 
-    def __init__(self, name=None):
+class Table(OrderedDict):
```
ines: The `Table` class above can be removed, right?

spacy/lookups.py
```diff
@@ -118,6 +117,21 @@ def to_disk(self, path, **kwargs):
...
+        path = ensure_path(path)
+        filepath = path / "lookups.bin"
+        if filepath.exists():
+            data = srsly.read_msgpack(filepath)
+            for key, value in data.items():
+                self._tables[key] = Table(key)
+                self._tables[key].update_raw(value)
+        return self
+
     def from_disk(self, path, **kwargs):
```
ines: Should this `from_disk` method go? There are two methods here, so it looks like the method added here isn't actually used?

spacy/lookups.py
polm: Ah, yes, that can go - this and the duplicated methods are leftovers from a bad merge. Thought I'd gotten rid of them...

PR #4276, opened 2019-09-11
💫 Add Language.pipe_labels

ines:
## Description
This is essentially just a shortcut to return the `[pipe].labels` for each component that has the property. Could be useful to get an overview of the label scheme of a model, and we could even consider including this in the `meta.json` (at least optionally).

This would also allow us to easily document the label schemes used by the individual models automatically (since the online model directory reads from the auto-generated metas in `spacy-models`). The [annotation specs](https://spacy.io/api/annotation) docs worked okay when we had only a few models. But it's becoming harder to maintain, and every model can potentially have its own, very different scheme.

(A downside of exposing this data via the `meta.json` is that it adds a lot more data to them.)

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -248,6 +248,18 @@ class Language(object):
         """
         return [pipe_name for pipe_name, _ in self.pipeline]
 
+    @property
+    def pipe_labels(self):
+        """Get the labels set by the pipeline components, if available.
+
+        RETURNS (dict): Labels keyed by component name.
+        """
+        labels = {}
+        for name, pipe in self.pipeline:
+            if hasattr(pipe, "labels"):
+                labels[name] = list(pipe.labels)
+        return labels
+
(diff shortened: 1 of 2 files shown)


spacy/language.py
```diff
@@ -248,6 +248,18 @@ def pipe_names(self):
         """
         return [pipe_name for pipe_name, _ in self.pipeline]
 
+    @property
+    def pipe_labels(self):
+        """Get the labels set by the pipeline components, if available.
+
+        RETURNS (dict): Labels keyed by component name.
+        """
+        labels = {}
```
honnibal: We probably want this to be `OrderedDict` to be kind to Python2 and Python3.5
```suggestion
        labels = OrderedDict()
```

ines: Yeah, fair enough!

PR #4284, opened 2019-09-13
Fix handling of spaces in Japanese

polm:
## Description

Before this change, half-width spaces in Japanese text were simply discarded. This wasn't immediately noticed because most Japanese text doesn't use spaces at all.

### Types of change

This is a bug fix.

## Checklist

- [x] I ran the tests, and all new and existing tests passed. (I added a new test.)

### DIFF
--- a/spacy/lang/ja/__init__.py
+++ b/spacy/lang/ja/__init__.py
@@ -37,6 +37,11 @@ def resolve_pos(token):
     in the sentence. This function adds information to the POS tag to
     resolve ambiguous mappings.
     """
+
+    # this is only used for consecutive ascii spaces
+    if token.pos == '空白':
+        return '空白'
+
     # TODO: This is a first take. The rules here are crude approximations.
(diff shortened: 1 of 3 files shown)


ines:
Great, thanks! 👍

tamuhey:
@polm How about treating full-width space in the same way as half-width space, not token?

polm:
@tamuhey That sounds like a good idea and, on inspection, actually seems to be how it's handled in the UD GSD corpus. I'll work on putting it in - thanks!

polm:
Just a followup, but turns out that internally spaces are stored as booleans, so supporting arbitrary whitespace is not a small task. I might come back to this later, but since full width spaces are currently consistent with non-ASCII-space whitespace such as tabs I think it's fine for now.

tamuhey:
@polm Thanks. Certainly it seems to be inconsistent to treat the full-width space differently only for the Japaneze tokenizer.

PR #4291, opened 2019-09-15
Reduce speed of merge command

Hazoom:
Making the `merge` command under `retokenize` session more efficient by changing the `head` indices from absolute position to relative position.

## Description
We've notices in our company that `merge` command takes a long time when performed on texts with a huge number of tokens since it converts all the `head` indices to absolute position at the beginning and change it back to relative position at the end of the method. Those changes are in fact two complete passes on the whole document tokens.
This PR removed those 2 passes and fixed the relevant code to work with relative indices, thus saving 2 passes on document's tokens and making the whole method takes less time to complete.

### Types of change
Performance -> speed

## Checklist
- [X ] I ran the tests, and all new and existing tests passed.
- [X ] My changes don't require a change to the documentation, or if they do, I've added all required information.

### DIFF
--- a/.github/contributors/Hazoom.md
+++ b/.github/contributors/Hazoom.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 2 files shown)


Hazoom:
Will open a new PR from master branch

PR #4292, opened 2019-09-15
Reduce speed of merge command

Hazoom:
Making the `merge` command under `retokenize` session more efficient by changing the head indices from absolute position to relative position.

## Description
We've notices in our company that `merge` command takes a long time when performed on texts with a huge number of tokens since it converts all the `head` indices to absolute position at the beginning and change it back to relative position at the end of the method. Those changes are in fact two complete passes on the whole document tokens.
This PR removed those 2 passes and fixed the relevant code to work with relative indices, thus saving 2 passes on document's tokens and making the whole method takes less time to complete.
Moreover, this PR makes the work with heads in the project unified and work with relative indices only. It's easier for maintenance and code understanding.

### Types of change
Performance -> speed

## Checklist

### DIFF
--- a/.github/contributors/Hazoom.md
+++ b/.github/contributors/Hazoom.md
@@ -101,6 +101,6 @@ mark both statements:
 | Name                           | Moshe Hazoom         |
 | Company name (if applicable)   | Amenity Analytics    |
 | Title or role (if applicable)  | NLP Engineer         |
-| Date                           | 2018-09-15           |
+| Date                           | 2019-09-15           |
 | GitHub username                | Hazoom               |
 | Website (optional)             |                      |
--- a/.github/contributors/Hazoom.md
+++ b/.github/contributors/Hazoom.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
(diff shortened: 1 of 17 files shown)


honnibal:
Thanks, looks interesting. I'm surprised it makes such a difference in speed though. When you do your merges, are you using the bulk retokenization, or merging one by one? Do you have numbers for the efficiency improvement?

Hazoom:
Hi @honnibal ,
Thanks for your comment.

At the end, it's the same amount of instructions, since I've added 2 more instructions in one of the tokens loop, so it shouldn't be a great change. The only thing that matters is that `for x in list` loops inside Cython code are very non efficient and I wanted to get rid from them as much as I can.

Moreover, the real purpose of this PR is to keep the work unified in all the project to work with relative heads and not absolution. The reason is, that in our company we are converting NLP syntactic rules (after parsing with spaCy) to C code and we developed our own `merge` command in C, so I believe it will be easier for others to understand the `merge` implementation with relative indices like everywhere else.

BTW, the reasons that we developed our own method are:
1. We are working in pure C (and not Cython).
2. We cannot delete tokens for the array since other previous rules can point to that token, so we perform the merge and mark this token as NULL token.

honnibal:
Hmm. I think there might be some confusion underneath all this. Apologies if you're already aware of this, but I do understand that Cython's a less well known technology, so it's possible to misunderstand things.

The loops you're replacing compile very directly into C loops, as the `doc.c` variable is a `TokenC*`, not a Python list. So this loop:

```python

Hazoom:
Thanks for the detailed explanation!!
I'm pretty familiar with Cython, but every day we learn something new :)

I've looked into the generated HTML and fixed all the places I could and now the execution time (average on 10 runs) on the 148K tokens document takes 23 milliseconds. A 50% reduction, I'm super happy :)

About the implementation in C - I've noted this just for a background on the reason that I've looked into the `_merge` implementation at the first place. We have a specific use case in our company that we work in C for it, so after looking into `_merge` implementation I was a little bit confused because I'm used to see implementations with `head` as relative index and not absolute. That's the reason I've decided to implement this PR for the first place.

So, for summary, this PR contains:
1) Reduction of 50% running time of `_merge` method
2) Change the absolute `head` indices to `relative` indices.

honnibal:
Glad the speed improved! Can you try just making the type declaration though? I think the other changes you've made are unnecessary and they make the implementation harder to maintain.

I'm also reluctant to merge changes to make a method 20ms faster even on really huge documents (150k tokens is a lot!)

Hazoom:
@honnibal Are you talking about the change from absolute index to relative index? or the changes in `for` loops to `while` loops?

honnibal:
@Hazoom I think it might be easier to just close this and commit the change separately. I'm still seeing changes from your previous efforts here, which I think has a high chance of leading to bugs, especially since this function is fairly complicated.

So, I'll close this and just commit the type declaration separately. Thanks for your attention to this :).

Hazoom:
Sure, will do. Thanks.

PR #4300, opened 2019-09-17
Improve speed of _merge method

Hazoom:
Making the `merge` command under `retokenize` session more efficient by declaring variables as Cython variables instead of Python variables.

## Description
This PR contains a change to the `_merge` method such that the Python variable that were in used during the method are now declared as Cython variables.
I've compared the average running time (10 runs) on a document with 148K tokens and the results are 47ms before the change and 23ms after the change.

### Types of change
Performance -> speed

## Checklist


Hazoom:
@honnibal I've seen that you already made the declaration fix, I thought that you planned that I'll do it. I've added some variables that you missed.
This PR doesn't contain my previous efforts, as you asked.
Thanks!

honnibal:
Thanks! And yeah since it was a small change I figured it was easier to just do it. But okay, thanks for the extra typings :).

Hazoom:
It's OK! Would love to help with other stuff as well.
Thanks!

PR #4303, opened 2019-09-17
add return_matches and as_tuples back to Matcher.pipe

SeanBE:
Adding `return_matches` and `as_tuples` back into `Matcher.pipe`. They were added in 1cf774bdc17b9f60e53c008d67fabfb164e84353 but were removed in issue https://github.com/explosion/spaCy/pull/3236.

## Description
I was following the documentation on the website when I ran into some unexpected keyword arguments when using `Matcher.pipe`. The alternative solution is updating the documentation but since I'm using `Matcher` in my implementation, I went for the code change. 

I could not find any tests on the pipe implementation for either `PhraseMatcher` or `Matcher`. I can create some if necessary. I also couldn't find an issue on this surprisingly. Maybe I'm missing something here. I can create an issue to link to this PR as well if wanted.

### Types of change
"new" feature

## Checklist

### DIFF
--- a/.github/contributors/seanBE.md
+++ b/.github/contributors/seanBE.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks! I think this looks good, @ines ?

ines:
Yes, looks good – thanks! No idea how this got lost in the refactoring 😞

PR #4305, opened 2019-09-18
remove redundant __call__ method in pipes.TextCategorizer

tamuhey:
Because the parent __call__ method behaves in the same way.

parent:
https://github.com/explosion/spaCy/blob/ee15fdfe88598b5bddf801d0d6700386efb4b7ab/spacy/pipeline/pipes.pyx#L61

textcat:
https://github.com/explosion/spaCy/blob/ee15fdfe88598b5bddf801d0d6700386efb4b7ab/spacy/pipeline/pipes.pyx#L935

If this is intentional, please close this PR.


svlandeg:
The failing unit tests actually indicate a bug in the original `__call__` method in the parent:
> self.set_annotations([doc], scores, tensor=tensors)

should be

> self.set_annotations([doc], scores, tensors=tensors)

tamuhey:
I think it is better to remove the duplication in terms of maintenance.
Like this time, duplications cause buggy code.

tamuhey:
This might cause backward incompatibility  for those who use `Pipe` interface and defines their own `set_annotation(..,**tensor**=..)`

svlandeg:
It is true that people may have copied the bug, which is unfortunate, but then I guess they'll have to correct that, too. Actually, this happened in the unit test `test_issue4278` :|

tamuhey:
Hahaha `test_issue4278` is what I created in (#4279).

tamuhey:
Sorry, removed the file.

honnibal:
Thanks!

I don't think there's much of a backwards incompatibility concern here, and if there is, v2.2 is the best time to make the fix.

PR #4307, opened 2019-09-18
Distinction between outside, missing and blocked NER annotations

### LINKED ISSUE #4267: Different ent_iob behavior after adding EntityRuler to pipeline
I'm not totally sure of the expected behavior but after adding an EntityRuler to a pipeline, non entities seem to get ```.ent_iob``` tags of 0 rather than 2 when just using the EntityRecognizer. This affects what ```doc.is_nered``` returns.

```python
import spacy

print(nlp.pipe_names)
## ['tagger', 'parser', 'ner']

doc = nlp("fgfgdghgdh")
print(doc.is_nered)
## True

for token in doc:
    print(token.ent_iob)
## 2

#addd entity ruler and run again
ruler = EntityRuler(nlp)
patterns = [{"label":"SOFTWARE", "pattern":"spacy"}]

ruler.add_patterns(patterns)
nlp.add_pipe(ruler)
print(nlp.pipe_names)
## ['tagger', 'parser', 'ner', 'entity_ruler']

doc = nlp("fgfgdghgdh")
print(doc.is_nered)
## False

for token in doc:
    print(token.ent_iob)
## 0
```

## Info about spaCy
* **spaCy version:** 2.1.8
* **Python version:** 3.6.8

svlandeg:
## Description
This PR attempts to process "empty" NER annotations more consistently. 

- Allow the NER algo to overwrite `O` (`ent_iob == 2`) annotations 
- Ensure that the NER algo preserves preset entities 
- Allow users to specify tokens that should never be in an entity. This "blocking" is done by setting `doc.ents` with a `Span` of tokens with empty `ent_type`. `ent_iob` is then set to `3`. In the transition system, these are recognized as `U-` actions, i.e. `UNIT` actions without a `label`.  
- As a result of the rewrite, `doc.ents = list(doc.ents)` now actually keeps the annotations on the token level consistent, instead of resetting `O` to empty string. It does this by checking previous annotations for each token: if it was `nered` before, we put it at `O`, otherwise empty string. This seems to be the most intuitive behaviour for a user inspecting the token-level data.
- Fixes #4267 

### Tests
 - I added some new unit tests in `test_ner.py` and for Issue 4267.
 - I tested the "preserving previous entities" functionality with statistical models, cf [here](https://github.com/svlandeg/spaCy/commit/fc779b31b90c26ee1d65776490a41f6ec48a9400), showing how that works properly. Removed those tests because they rely on the models to be installed.
 - `test_doc_add_entities_set_ents_iob` was in the repo twice so I removed one, and changed the other to have `O` annotations. 

### Open questions
 - Some old tests failed because `nn_parser.move_names` now contains `U-`. For now I removed it explicitely from `move_names`, but we could also adjust the unit tests. Depends on whether or not we want to keep that action internal.

### Caveat
For the "blocking" functionality to work with the statistical models, they'll have to be retrained.

### Types of change
Enhancement

## Checklist


spacy/tests/regression/test_issue1-1000.py
```diff
@@ -426,7 +426,7 @@ def test_issue957(en_tokenizer):
 def test_issue999(train_data):
     """Test that adding entities and resuming training works passably OK.
     There are two issues here:
-    1) We have to readd labels. This isn't very nice.
+    1) We have to read labels. This isn't very nice.
```
honnibal: Heh, sorry --- should've punctuated that better.

```suggestion
    1) We have to re-add labels. This isn't very nice.
```

honnibal:
Great work!

Quick question: do we add `U-` to the model? If we don't add it up-front, we'll end up having to add it later, which goes through the resize logic. That makes things much more complicated.

I think we should have this as one of the moves that we ensure is present, like we do with the `S` and `D` moves.

honnibal:
I'm going to go ahead and merge this, because it requires retraining the NER models for v2.2, and I want to get that triggered overnight.

svlandeg:
No, I haven't tried training a model yet. Once we do, we can use [these unit tests](https://github.com/svlandeg/spaCy/commit/08b06e483ac06c116850bdd4cbfaa72a0c7c4533) to test whether the new models work as expected.

PR #4309, opened 2019-09-19
Replace PhraseMatcher with trie-based search

### LINKED ISSUE #3922: Missing remove method in PhraseMatcher
## Feature description
In Matcher it is possible to remove an added pattern by calling the remove method but that's not possible with the PhraseMatcher. Any tips

### LINKED ISSUE #4308: PhraseMatcher starts missing matches if lots of new vocab has been added

You need this keywords file: https://github.com/mpuig/spacy-lookup/blob/master/data/keywords.txt

If you process the sample keywords before doing all the extra processing, the match is found. You can also process the sample keywords again after all the extra processing and the match is still found.

If you load the sample keywords after doing all the extra processing (as below) and `keyword_limit` is at least `10067`, the match is not found. Any lower, the match is found.

I think it's probably something related to the Vocab and not PhraseMatcher itself, but I really don't know?

```
import spacy
from spacy.matcher import PhraseMatcher

def run_matching(matcher, docs):
    matches = []
    for doc in docs:
        matches.extend([match for match in matcher(doc)])

    return (len(matches), matches)

def run(with_extra_keyword_processing=False):
    print("\n\nRunning test", "*with*" if with_extra_keyword_processing else "*without*", "extra document processing")

    keyword_limit = 10067

    nlp = spacy.load('en')

    if with_extra_keyword_processing:
        print("Processing a lot of keywords, then discarding...")
        unused_keywords = []
        count = 0
        with open("keywords.txt") as fileh:
            for line in fileh:
                unused_keywords.append(line.strip())
                if count > keyword_limit:
                    break
                count += 1
        unused_keywords = [nlp.make_doc(keyword) for keyword in unused_keywords]

    keywords = sample_keywords()
    keywords = [nlp.make_doc(keyword) for keyword in keywords]
    print("Loaded sample keywords")
    print("Keywords:", keywords)
    print("# Keywords:", len(keywords))

    texts = sample_texts()
    docs = list(nlp.tokenizer.pipe(texts))
    print("Loaded sample docs")
    print("# Docs:", len(docs))

    smatcher = Matcher(nlp.vocab)
    pmatcher = PhraseMatcher(nlp.vocab)

    for keyword in keywords:
        smatcher.add(keyword.text, None, [{"ORTH": token.text} for token in keyword])
        pmatcher.add(keyword.text, None, keyword)

    for matcher in [smatcher, pmatcher]:
        print("\nMatcher type:", type(matcher))
        (len_matches, matches) = run_matching(matcher, docs)
        print("# Matches", len_matches)
        print("Matches:", matches)

def sample_keywords():
    return ["Hanif Kureishi"]

def sample_texts():
    return [
"""
A real triumph for Roger Michell and Hanif Kureishi, and the rest of the team. A must see for serious film lovers.
""",
]

if __name__ == "__main__":
    run()
    run(with_extra_keyword_processing=True)
```

Output:

```
Running test *without* extra document processing
Loaded sample keywords
Keywords: [Hanif Kureishi]
# Keywords: 1
Loaded sample docs
# Docs: 1

Matcher type: <class 'spacy.matcher.matcher.Matcher'>
# Matches 1
Matches: [(4725172808025326492, 8, 10)]

Matcher type: <class 'spacy.matcher.phrasematcher.PhraseMatcher'>
# Matches 1
Matches: [(4725172808025326492, 8, 10)]


Running test *with* extra document processing
Processing a lot of keywords, then discarding...
Loaded sample keywords
Keywords: [Hanif Kureishi]
# Keywords: 1
Loaded sample docs
# Docs: 1

Matcher type: <class 'spacy.matcher.matcher.Matcher'>
# Matches 1
Matches: [(4725172808025326492, 8, 10)]

Matcher type: <class 'spacy.matcher.phrasematcher.PhraseMatcher'>
# Matches 0
Matches: []
```

## Your Environment

* **spaCy version:** 2.1.8
* **Platform:** Linux-4.19.0-5-amd64-x86_64-with-debian-10.0
* **Python version:** 3.7.3

adrianeboyd:
## Description

Replace PhraseMatcher with a trie-based search algorithm over numpy arrays of the hash values for the relevant attribute. The implementation is based on FlashText (https://github.com/vi3k6i5/flashtext).

The speed should be similar to the previous PhraseMatcher. It is now possible to easily remove match IDs and matches don't go missing with large keyword lists / vocabularies.

Fixes #3922, fixes #4308.

### Types of change

Bugfix? Enhancement? One of the two.

## Checklist


adrianeboyd:
Timing for matching 10-10000 phrases in 10000 Wikipedia documents (showing: # matches, seconds):

```
# Docs: 10000

# Keywords: 10
Matcher: 0 5.92
PhraseMatcher: 0 4.67

# Keywords: 100
Matcher: 340 23.69
PhraseMatcher: 340 5.63

# Keywords: 1000
Matcher: 1070 207.82
PhraseMatcher: 1070 4.79

# Keywords: 10000
Matcher: 26003 4224.96
PhraseMatcher: 26003 5.82

```

honnibal:
Great work! We'll easily be able to make this 10-20x faster by cythonising the match loop, although it sounds like that's not even a priority at the moment, as the speed seems fine.

A few small things:

* Why do we need to store the docs? That will make the pickling really big, and it should be unnecessary.
* Do we need the `Vocab` here, or just the string store? I guess we can keep passing the vocab in for backwards compatibility, but it's probably good to keep track of whether it's necessary. I think probably it isn't? We might not even need the string store, if we remember the strings for the match IDs.
* I suppose we should add `to_bytes()` and `from_bytes()` methods? I think we didn't have them on the old `PhraseMatcher`, but while we're here we may as well do it.

honnibal:
Actually maybe we should think about the speed here? If we get it fast enough, we could use it it the tokenizer, like you're planning in #4239.

adrianeboyd:
I originally thought the new version didn't need to store the docs, so I took it out, but then added it back to be like the old PhraseMatcher. I've never seen examples that use the PhraseMatcher as part of a pipeline and I'm not entirely sure how everything is expected to work in terms of saving/loading this as part of a pipeline or with pickling (which now just works like the old PhraseMatcher so I'm not removing functionality). So this definitely has a lot of room for improvement. As you said, I think the minimum vocab you would have to store is the strings for the match IDs.

> Actually maybe we should think about the speed here? If we get it fast enough, we could use it it the tokenizer, like you're planning in #4239.

If it were faster, it would make the tokenizer changes more straightforward. It's still too slow currently, about 40-50% slower than the current tokenizer on average, and still about 2x slower for English. (PhraseMatcher instead of Matcher improves the speed for cases like Indonesian with a very large number of patterns, but not for cases with a small number of patterns with frequent matches.)
 
> The main data structure, `keyword_trie_dict` is a `Dict[uint64, Trie]`, where the trie is a recursive data structure that terminates in a set of uint64, right? We could implement that with preshed, which should allow us to move the entire match loop into nogil code. It would be a `MapStruct` where the values are `MapStruct*`, and the values of the inner map are just 0 or 1.

Right now there's a special non-uint64 terminal symbol that points to the set, but basically, yes.

> Now that I'm looking at that data structure, is this line right?
> 
> https://github.com/explosion/spaCy/blob/3a4e1f5ca7ff550e90d0c309d8673a296e05e41c/spacy/matcher/phrasematcher.pyx#L195
> . It seems like the value there should be a set, so maybe you're missing a loop?

Yes, I missed that one, will fix in a second. (I originally had new match IDs replace old match IDs for the same phrase, but realized from the unit tests that that wasn't the desired behavior, so I swapped it to sets.)

I had to modify the FlashText matching loop quite a bit because the FlashText version only returns the longest match and because it didn't find adjacent matches, since it expects at least one whitespace character between the last and first letter of adjacent keywords. I should just double-check this whole loop again, since I would have expected this problem to cause errors and I never noticed any (in what must be nearly billions of tokens/matches in various tests at this point).

adrianeboyd:
The algorithm is not really Aho-Corasick, since there are no suffix/failure links. It's just a kind of trie-based search and I'm not sure what it's usually called.

We could implement add suffix links on the token level, but for the shorter patterns typically used with the PhraseMatcher, I don't think there would be a whole lot of benefit. It would complicate adding and removing matches and would make it more difficult to implement the matching loop with PreshMap.

adrianeboyd:
Timing for master vs. this, matching on 10000 Wikipedia documents, averaged over 10 runs:

\# keywords | # matches | master | this
--- | --- | --- | ---
10 | 0 | 0.439 | 0.027  
100 | 795 | 0.457 | 0.028
1000 | 11376 | 0.512 | 0.043
10000 | 105688 | 0.632 | 0.114

To do at this point:

- [ ] fix remove
- [ ] pickling
- [ ] `to_bytes`/`from_bytes`

honnibal:
Performance numbers look very nice! It does look like we're seeing a change in the runtime with respect to number of keywords, though. Our runtime is roughly doubling every order of magnitude increase in number of keywords, where previously it was very flat.

On projection it looks like we'd only cross over to being slower once we hit 10 million keywords. It would also depend a lot on the specifics of the keywords, as the PhraseMatcher performs poorly once enough common words end up being marked as potential entities. For instance, as soon as you have `a` and `the` beginning at least one phrase, it's a lot slower.

I definitely prefer the performance of the new implementation. If it does come out that people need the performance of the previous phrase matcher, we can help the people affected release it as an extension package, as a standalone component.

honnibal:
Getting down to the much smaller performance suggestions now:

Maybe try assigning the `attr` and `terminal` symbols to local variables? This will remove another pointer dereference from within the hot loop, which might help cache locality slightly.

adrianeboyd:
The current PhraseMatcher starts missing matches at ~10000 keywords so it gets hard to compare the performance, since both are sensitive to the number of matches, I think?

Here's what the new one looks like for larger keyword lists. It does seem to keep doubling on every order of magnitude increase in the number of keywords, which I think is largely due to the number of matches:

\# keywords | time
---|---
100000 | 0.204
450000 | 0.281

For reference, this is what I'm timing:

    matches = []
    for doc, doc_matches in matcher.pipe(docs, return_matches=True):
        matches.extend([doc[match[1]:match[2]] for match in doc_matches])

And if I don't do the span calculations it looks a little flatter, but not a whole lot:

\# keywords | time
--- | ---
10 | 0.0249
100 | 0.0270
1000 | 0.0316
10000 | 0.0371
100000 | 0.0727
450000 | 0.1253

The runtime should be something like O(doc_length * max_pattern_length + num_matches).

Do you know what typical cases look like in practice? Are people using lists with millions of patterns that start with `a` and `the`?

adrianeboyd:
On pickling: I'm tempted to punt and leave it as is. I don't think there's a simple way to pickle the MapStruct trie, is there? How often are people pickling PhraseMatchers and complaining about the size?

Pickle sizes:

\# keywords | size (MB)
---|---
10 | 17
100 | 17
1000 | 17
10000 | 22
100000 | 73
450000 | 79

honnibal:
Oh, true about the missing matches. So yeah, hard to compare.

> Do you know what typical cases look like in practice? Are people using lists with millions of patterns that start with a and the?

I'm particularly interested in using this for rule-based NER and entity linking. I do want to support use-cases where you have all the Wikipedia page titles in a single phrase-matcher. These would definitely start with common words in some cases.

About the pickling: Okay I guess we can punt on this and just make an improvement later. I agree that pickling the trie is too annoying, but ideally I'd prefer to have numpy arrays instead of doc objects stored.

adrianeboyd:
I can totally work on the pickling if it's needed, but I just wasn't sure it was much of a priority?

adrianeboyd:
The size of the vocab seems to be more of an issue than the size of the docs, but there's not any way to "clean" the vocab and just store the extra match ID strings, is there?

honnibal:
@adrianeboyd There's not, no, and there's no way to know whether that vocab object is shared with other objects (it probably is).

The main solution would be to just not store the doc objects. I'll go ahead and merge this though --- the pickle improvement is definitely something that could happen later.

PR #4331, opened 2019-09-27
Initialize low data warning for debug-data parser

adrianeboyd:
## Description

Initialize low data warning for debug-data parser, which resulted in an error if you only checked `-p parser` rather than all components.

### Types of change

Bugfix.

## Checklist


tataducs:
B07BVYPDJ4

PR #4339, opened 2019-09-29
Skip duplicate spans in Doc.retokenize

### LINKED ISSUE #3687: 💫 Retokenizer.merge should automatically skip duplicates
Inspired by #3681. 

If you try to merge two overlapping spans using the new `Retokenizer`, spaCy will now correctly raise an error, since it's impossible to achieve the intended result. For instance:

```python
# Impossible because there's no reality where both spans can exist as a separate merged token
    for span in [doc[0:3], doc[1:4]]:
        retokenizer.merge(span)
```

Duplicate spans can easily happen when you combine matches or things like entities and noun chunks. Because two identical spans are also "overlapping" in theory, an error is also raised when you try to merge them: 

```python
# Possible, but still raises an error
    for span in [doc[0:3], doc[0:3]]:
        retokenizer.merge(span)
```

My PR in #3686 adds a helper `filter_spans` that removes overlapping spans and duplicates, and alternatively, users can also write custom functions (e.g. to expand overlapping spans to the largest possible sequence of neighbouring tokens). But the basic case of duplicates should be handled internally and shouldn't require the user to write their own filter functions.

Retokenizer source: https://github.com/explosion/spaCy/blob/master/spacy/tokens/_retokenize.pyx

ines:
Closes #3687.

## Description
 
> Duplicate spans can easily happen when you combine matches or things like entities and noun chunks. Because two identical spans are also "overlapping" in theory, an error is also raised when you try to merge them

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/tests/doc/test_retokenize_merge.py
+++ b/spacy/tests/doc/test_retokenize_merge.py
@@ -414,3 +414,14 @@ def test_doc_retokenizer_merge_lex_attrs(en_vocab):
     assert doc[1].is_stop
     assert not doc[0].is_stop
     assert not doc[1].like_num
+
+
+def test_retokenize_skip_duplicates(en_vocab):
+    """Test that the retokenizer automatically skips duplicate spans instead
+    of complaining about overlaps. See #3687."""
+    doc = Doc(en_vocab, words=["hello", "world", "!"])
+    with doc.retokenize() as retokenizer:
+        retokenizer.merge(doc[0:2])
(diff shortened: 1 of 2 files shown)


honnibal:
Definitely a great change.

What do we want to do about the problem of overlapping or adjacent merges? For instance, if I characters (0, 10) and (5, 20), we could just resolve that to (0, 20). This is how it used to work.

The main problem I see is that it's unclear how to resolve the token attributes, if they're specified as well.

PR #4344, opened 2019-09-30
Tidy up and modernize setup and config

ines:
## Description

Let's see if this works. We need to make sure we support old enough versions of `setuptools`, so we can't fully move to `pyproject.toml` at this point.

* Move everything that doesn't have to be generated in Python from the `setup.py` to the `setup.cfg`.
* Remove unused old code from the `setup.py` and remove explicit `package_data` (so `include_package_data = true` works as expected when building wheels – long story 🙃)
* Update dependencies in `pyproject.toml` (forgot to update this and ended up with various confusing problems in the Docker containers building our wheels)
* Move all "static" meta that isn't used anywhere else from `about.py` to `setup.cfg`.
* Move `flake8` config to `setup.cfg`.

### Types of change
enhancement, install

## Checklist


setup.cfg
```diff
@@ -0,0 +1,90 @@
...
+# Language tokenizers with external dependencies
+ja =
+    mecab-python3==0.7
+ko =
+    natto-py==0.9.0
+th =
+    pythainlp>=2.0
+
+[bdist_wheel]
+universal = true
```
honnibal: This is wrong, right? I'm pretty sure we don't provide universal wheels

ines: Ah yeah, I guess. Sorry, just copy-pasted that.

PR #4353, opened 2019-10-01
Refactor lemmatizer and data table integration

ines:
## Description

* Get lookups table within the lemmatization methods to make sure it references the correct table (even if the table was replaced or modified, e.g. when loading a model from disk). This fixes a bug that'd cause data from a model to not be used correctly after deserialization (because the lemmatizer was still holding a reference to an old table).
* Deprecate other arguments on `Lemmatizer.__init__` and expect `Lookups` for consistency.
* Raise error on old and unsupported `Lemmatizer.load` classmethod.
* Refactor language-specific lemmatizers to inherit as much as possible from base class and override only what they need.
* Update docs accordingly.

### Types of change
bug fix

## Checklist


requirements.txt
```diff
@@ -15,7 +15,7 @@ pathlib==1.0.1; python_version < "3.4"
 jsonschema>=2.6.0,<3.1.0
 # Development dependencies
 cython>=0.25
-pytest>=4.0.0,<4.1.0
+pytest>=5.2.0,<5.3.0
```
honnibal: Apparently this is the last version that supports Python 2?

```suggestion
pytest>=4.6.5```

jshin49:
This is an old PR but I'm only encountering this problem now.

In [lemmatizer.py L46-47](https://github.com/explosion/spaCy/blob/a26f864ed3c227fab1d2a506e27cb4b5b5d831d2/spacy/lemmatizer.py#L46-L47)
what is the reasoning behind not using the lookup table when there are rules? 

I guess this is a design choice, but it does cause unexpected problems. Please look at Issue #6075  that I just opened,

PR #4359, opened 2019-10-02
Update Ukrainian lemmatizer with new lookups

adrianeboyd:
## Description

Update Ukrainian lemmatizer with new lookups (copied from Russian)

### Types of change

Bugfix.

## Checklist


ines:
Thanks! I remember thinking "need to remember to do Ukranian, too" while fixing Russian but then I forgot 🙃

PR #4361, opened 2019-10-02
Make the default NER labels less model-specific [ci skip]

adrianeboyd:
## Description

Make the default NER labels less model-specific

### Types of change

Bugfix.

## Checklist


ines:
Yeah, good point. I was trying to make the default selection less verbose by limiting it to some of the most common types – but it's true that this doesn't generalise well and there's not an easy way to make it work across models.

Issue #3697, opened 2019-05-07
Make most_similar work like Gensim's version


mjspeck:
## Feature description

Currently `nlp.vocab.vectors.most_similar()` returns only the single most similar vector for each vector passed to it. Gensim's `most_similar()` [returns the n-most similar vectors to a given vector](https://radimrehurek.com/gensim/models/keyedvectors.html#gensim.models.keyedvectors.BaseKeyedVectors.most_similar). This seems like a more useful implementation.

## Could the feature be a [custom component](https://spacy.io/usage/processing-pipelines#custom-components) or [spaCy plugin](https://spacy.io/universe)?
If so, we will tag it as [`project idea`](https://github.com/explosion/spaCy/labels/project%20idea) so other users can take it on.
This would not be a custom component or plugin.

honnibal:
I guess we could support this. If someone wants to take this on, I would suggest using the `libcpp.queue.priority_queue` class to implement it. The `Vectors` class is already in Cython and uses C++ data structures, so there's no problem with using a C++ class in this function. I've found the STL `priority_queue` class much better than the Python equivalents.

CartierPierre:
@mjspeck how to use this actual `most_similar` ? Is there any simple usage like
```
apples = nlp(u"I like apples")
oranges = nlp(u"I like oranges")
bananas = nlp(u"I like bananas")
most = apples.most_similarity(oranges, bananas)
```
or in a list ?

CartierPierre:
I tried this
```
vect = Vectors(data=nlp("This is sentence").vector)
most = vect.most_similar(nlp("This is other sentence").vector)
```
and an error appeared :
```
Traceback (most recent call last):
  File "nlp.py", line 79, in <module>
    test_nlp()
  File "nlp.py", line 57, in test_nlp
    print(vect.most_similar(nlp("This is other sentence").vector))
  File "vectors.pyx", line 322, in spacy.vectors.Vectors.most_similar
  File "/home/pierre/.local/lib/python3.6/site-packages/numpy/linalg/linalg.py", line 2481, in norm
    return sqrt(add.reduce(s, axis=axis, keepdims=keepdims))
numpy.AxisError: axis 1 is out of bounds for array of dimension 1
```
I also tried with more than 1 vector (using np.stack) and same error.

mjspeck:
[...]

PR #4364, opened 2019-10-02
most_similar() return the k most similar vectors

bintay:
## Description

Modified the `most_similar` method of `Vectors` so that it returns the top `n` most similar vectors rather than the single most similar vector.

The new value returned looks like `(keys, best_rows, scores)` where `keys`, `best_rows`, and `scores`  are arrays containing an array for each query. So `scores[0]` would be an array of scores corresponding to the `n` vectors most similar to the `0`th query.

In theory, performance should not be impacted much — `argmax` is replaced by `argpartition`, both of which run in linear time with respect to the number of elements. If the user chooses to sort the `n` most similar entries by setting the `sort` parameter to true, it will add an `O(n lg n)` step to sort the entries. However, I haven't tested the performance in practice yet and I might be missing something.

This could break existing code since the return type is changed from a tuple of three 1d arrays to a tuple of three 2d arrays. We could have it use the old method when `n = 1` and the new method otherwise, but then it could be confusing to use & documentation could get messy. Could also keep the old method and add a new `n_most_similar` method.

This is my first time contributing to an open source project, so let me know your thoughts!

### Types of change

Enhancement ([#3697](https://github.com/explosion/spaCy/issues/3697))

## Checklist

### DIFF
--- a/.github/contributors/bintay.md
+++ b/.github/contributors/bintay.md
@@ -87,7 +87,7 @@ U.S. Federal law. Any choice of law rules will not apply.
 7. Please place an “x” on one of the applicable statement below. Please do NOT
 mark both statements:
 
-    * [ ] I am signing on behalf of myself as an individual and no other person
+    * [x] I am signing on behalf of myself as an individual and no other person
     or entity, including my employer, has or will have rights with respect to my
     contributions.
 
--- a/.github/contributors/bintay.md
+++ b/.github/contributors/bintay.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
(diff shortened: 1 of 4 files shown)


ines:
Thank so much 🙏 This is really good timing btw, because I was just working on a [refactor of `sense2vec`](https://github.com/explosion/sense2vec/tree/refactor), where I'm want to use that method properly.

Also, I think that one test is failing because `Vocab.prune_vectors` calls into `Vectors.most_similar`, so we just need to adjust the method call there.

bintay:
Yeah it's because `scores[i]`, etc are arrays now but it was expecting a number. Changing them to `scores[i][0]` fixed that error. 

Now the test is actually failing — `neighbour` is `"dog"` instead of `"cat"`. Because all 3 of the vectors in the test are just scaled versions of each other, their cosine similarity is 1 so either should be valid. I guess `argmax` just happened to return `"cat"` whereas `argpartition` returns `"dog"`.

I updated the test with more "realistic" vectors for now. Let me know if we actually want to test the case where all the vectors point in the same direction.

honnibal:
Thanks! Yeah I think the change to the test is good, that makes sense. Merging 🎉

PR #4381, opened 2019-10-05
Add GluonNLP to spaCy Universe

szha:
Add [GluonNLP](https://gluon-nlp.mxnet.io) to spaCy Universe

### Types of change
documentation

## Checklist

### DIFF
--- a/.github/contributors/szha.md
+++ b/.github/contributors/szha.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks for the thought!

At the beginning, we did merge some things into the spaCy Universe that were a bit more tangentially related, especially since the Github Dependencies feature wasn't active yet --- so we wanted to highlight various projects that used spaCy.

As the Universe has grown, it's becoming more important that we curate it a bit more, to make sure it stays a useful resource. So we want to focus on things that are useful within the spaCy ecosystem, or projects that provide useful example usage of spaCy, so people can see what problems people are solving and how the library is being used. At some point we might divide the Universe up into these categories more explicitly.

I think Gluon doesn't really meet either definition, so I think it isn't so useful to list it. I do think Gluon is a cool library that more people should be aware of though! It would be nice to have a wrapper in case people wanted to use Gluon models within spaCy, if they already have code that uses spaCy's Doc object and pipeline.

szha:
@honnibal thanks for the reply. We will let you know if/when a wrapper for spaCy happens.

PR #4383, opened 2019-10-06
Simplify chained comparisons.

jbampton:
Less code and seems easier to read.


honnibal:
Hey,

Thanks for the thought, and we definitely are open to small improvements like this. But for this specific case, I'm not sure I agree I prefer your version. My system for comparisons is to always use >= and <. I try to avoid <= and >. This helps avoid off-by-one errors, because errors stand out better.

The ternary comparison breaks these rules, and to me takes a while to think about. It's also inconsistent with the rest of the code-base, so I don't think I prefer the change.

Issue #4382, opened 2019-10-05
model VersionConflict on upgrading to 2.2.1


Motorrat:
## How to reproduce the problem
Use virtual environment. Upgrade spacy 2.1.8 to 2.2.0 and upgrade the models. 
make symlink 
```
ln -s /home/user/venv/full/lib/python3.6/site-packages/de_core_news_md -->
/home/user/venv/full/lib/python3.6/site-packages/spacy/data/de_core_news_md
```

Then try loading the model with nlp=spacy.load('de_core_news_md')

and get the following error:

`pkg_resources.VersionConflict: (spacy 2.2.1 (/home/user/venv/full/lib/python3.6/site-packages), Requirement.parse('spacy<2.2.0,>=2.1.7'))`

So here are detailed steps (all run in virtual environment "full"):
```
pip install --upgrade spacy
Successfully installed blis-0.4.1 preshed-3.0.2 spacy-2.2.1 thinc-7.1.1 urllib3-1.25.6

python -m spacy validate
package   de-core-news-md            de_core_news_md            2.1.0   --> 2.2.0

python -m spacy download de_core_news_md
...
Successfully installed de-core-news-md-2.2.0

python -m spacy download de_core_news_md
Requirement already satisfied: de_core_news_md==2.2.0 from https://github.com/explosion/spacy-models/releases/download/de_core_news_md-2.2.0/de_core_news_md-2.2.0.tar.gz#egg=de_core_news_md==2.2.0 in /home/user/venv/full/lib/python3.6/site-packages
Requirement already satisfied: spacy>=2.2.0 in /home/user/venv/full/lib/python3.6/site-packages (from de_core_news_md==2.2.0)
Requirement already satisfied: preshed<3.1.0,>=3.0.2 in /home/user/venv/full/lib/python3.6/site-packages (from spacy>=2.2.0->de_core_news_md==2.2.0)
Requirement already satisfied: thinc<7.2.0,>=7.1.1 in /home/user/venv/full/lib/python3.6/site-packages (from spacy>=2.2.0->de_core_news_md==2.2.0)
Requirement already satisfied: blis<0.5.0,>=0.4.0 in /home/user/venv/full/lib/python3.6/site-packages (from spacy>=2.2.0->de_core_news_md==2.2.0)
Requirement already satisfied: requests<3.0.0,>=2.13.0 in /home/user/venv/full/lib/python3.6/site-packages (from spacy>=2.2.0->de_core_news_md==2.2.0)
[...]

PR #4387, opened 2019-10-06
Improve usage of pkg_resources and handling of entry points

ines:
## Description

* As we noticed in #4382, querying entry points via `pkg_resources` produces very cryptic errors if there's a dependency mismatch of the package exposing entry points (e.g. `spacy-pytorch-transformers`). The package name is not shown, which makes this really confusing for users. This PR replaces the current implementation with [`importlib_metadata`](https://importlib-metadata.readthedocs.io/en/latest/) (a backport of 3.8's standard library `importlib.metadata` module), which doesn't seem to have this issue.
* While researching this, I also found threads about `pkg_resources` being very slow on import. I couldn't find a "modern" replacement for `pkg_resources.working_set` (it's not in [`importlib_resources`](https://importlib-resources.readthedocs.io/en/latest)), so for now, I've moved the imports into the functions where they're needed (currently mostly related to model downloads and checking whether model packages are installed).

### Types of change
meh

## Checklist


honnibal:
Thanks for chasing this down! Hard to think of less fun details to resolve.

PR #4392, opened 2019-10-07
Fix --gold-preproc train cli command

honnibal:
The addition of the `cats` variable in `gold.pyx` caused problems on one of the conditional branches, when the `--gold-preproc` argument was passed for training. This flag tells spaCy to train on the gold-standard segments (words and sentences), instead of using the raw text or merged paragraphs.

The problem came up because of the way the cats variable was appended to the sentence tuples. Instead, I've placed it with the `brackets` variable, since most of the code doesn't use that --- so it's less disruptive.

We don't have many tests for the CLI, so I might have broken something. I'll run more commands, but in the meantime, does it look okay? 

## Checklist


adrianeboyd:
The problem is that `cats` is per paragraph and `brackets` is per sentence, so it doesn't make sense to store `cats` at that level.

Is there a performance reason for the complicated tuples? (I don't think so?) I think it would be a good idea to replace all of it with a class that's easier to maintain and extend. (It was really hard to keep track of what was going on with the tuples while adding `cats` and there were a few spots in `begin_training()` methods that I missed initially, too, since it's also really hard to search for the necessary places to update. I don't think they're handled correctly after this change, either?)

honnibal:
@adrianeboyd There's no performance advantage of the tuples, it's just bad code that's persisted.

I already had the `GoldParse` class but it's expensive to create because of the alignment, and it didn't seem so bad to have an informal data structure. But then that propagated through the code-base, and hasn't been cleaned up. I also wasn't sure what to name the intermediary structure.

Even if we don't want a class, it's plainly better to have a dict structure than the stupid tuples. The tuples mean that if we add a new field we break the receiving code, which is obviously bad.

The problem currently is that there may be backwards compatibility issues in changing this, if people have custom components that pass in the `get_gold_tuples` callback to the `begin_training()` method.

The `brackets` item corresponds to what will be created as one `Doc` object. It could be a single sentence or it could be multiple sentences concatenated together. Since that will be created as one `Doc`, it seems correct for that to be the level at which the cats are specified.

spacy/pipeline/pipes.pyx
```diff
@@ -1033,8 +1033,7 @@ class TextCategorizer(Pipe):
         return 1
 
     def begin_training(self, get_gold_tuples=lambda: [], pipeline=None, sgd=None, **kwargs):
-        for raw_text, annots_brackets in get_gold_tuples():
-            cats = annots_brackets.pop()
+        for raw_text, (_, (cats, _2)) in get_gold_tuples():
             for cat in cats:
                 self.add_label(cat)
```
adrianeboyd: This doesn't work because the cats-brackets are on each sentence. It needs to be something more like:

        for raw_text, annots_brackets in get_gold_tuples():
            for annots, (cats, brackets) in annots_brackets:
                for cat in cats:
                    self.add_label(cat)

Or possibly the problem is that `merge_sents` isn't called for some reason in some cases, but in my textcat test cases this doesn't currently work.

adrianeboyd:
In `gold.train_tuples` the cats are stored on every sentence:

```
import spacy
from spacy.gold import GoldCorpus
gc = GoldCorpus("examples/training/textcat_example_data/cooking.json", "examples/training/textcat_example_data/cooking.json")
for raw, annots_brackets in gc.train_tuples:
    print(raw)
    print(annots_brackets)
```

The last example is:

```
What is the difference between baking soda and baking powder?
And can I use one in place of the other in certain recipes?

[[[[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10], ['What', 'is', 'the', 'difference', 'between', 'baking', 'soda', 'and', 'baking', 'powder', '?'], ['-', '-', '-', '-', '-', '-', '-', '-', '-', '-', '-'], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10], ['', '', '', '', '', '', '', '', '', '', ''], ['O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O']], [{'baking': 0.0, 'not_baking': 1.0}, []]], [[[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14], ['\n', 'And', 'can', 'I', 'use', 'one', 'in', 'place', 'of', 'the', 'other', 'in', 'certain', 'recipes', '?'], ['-', '-', '-', '-', '-', '-', '-', '-', '-', '-', '-', '-', '-', '-', '-'], [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14], ['', '', '', '', '', '', '', '', '', '', '', '', '', '', ''], ['O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O', 'O']], [{'baking': 0.0, 'not_baking': 1.0}, []]], [[[0], ['\n'], ['-'], [0], [''], ['O']], [{'baking': 0.0, 'not_baking': 1.0}, []]]]
```

I think `merge_sents()` is only called for `train_docs...iter_gold_docs` not for `train_tuples`,  which is what is passed to `begin_training()` in the train CLI:

https://github.com/explosion/spaCy/blob/dd30d3ec998bfc70084b6dc096972343e3ea38fd/spacy/cli/train.py#L240

spacy/gold.pyx
```diff
@@ -68,9 +68,9 @@ def merge_sents(sents):
         m_deps[5].extend(ner)
         m_brackets.extend((b["first"] + i, b["last"] + i, b["label"])
                           for b in brackets)
+        m_cats.update(cats)
```
adrianeboyd: This makes me nervous because it relies on every sentence being labeled the same way in an intermediate step or there will unexpected bugs.

adrianeboyd:
An easy non-goldproc test case is:

```
spacy train en output cooking.json cooking.json -p textcat -tpl baking
```

adrianeboyd:
Other places that are affected:

https://github.com/explosion/spaCy/blob/dd30d3ec998bfc70084b6dc096972343e3ea38fd/spacy/syntax/arc_eager.pyx#L344-L345

https://github.com/explosion/spaCy/blob/dd30d3ec998bfc70084b6dc096972343e3ea38fd/spacy/syntax/ner.pyx#L75-L76

adrianeboyd:
Is there anything that `merge_sents()` does that causes problems for the data in terms of `gold_preproc`? Maybe there's something I'm not seeing, but I don't see any obvious problems? If there aren't any, then a solution could be to remove the `else:` here:

https://github.com/explosion/spaCy/blob/5cbe21700b43bd2562ba49751fe646983d79dff4/spacy/gold.pyx#L238-L239

honnibal:
> Edit: I see that Matt says that the idea is not to merge paragraphs, but why would you want that behavior? 

Mostly for experimentation. For direct comparison with other tools, we want to be able to train in the "normal" per-sentence sort of setting. Internally, it's also good to be able to check how running on multiple sentences affects accuracy. Finally, it also can be useful if the user expects that at runtime they'll be dividing up their data such that they have one sentence per document.

adrianeboyd:
Right, but nothing is stopping you from structuring your training data with one sentence per document to get the same effect. It's just going to be really hard to do this with `cats` in any sensible way.

ryszardtuora:
Any updates on this?

adrianeboyd:
We are working on improving how the raw data is stored internally so we can avoid this problem. If you don't need the sentence-per-document part of `-G` and just the gold tokenization, you can make the edit from my comment above to get the gold tokenization to work:

> Is there anything that `merge_sents()` does that causes problems for the data in terms of `gold_preproc`? Maybe there's something I'm not seeing, but I don't see any obvious problems? If there aren't any, then a solution could be to remove the `else:` here:
> 
> https://github.com/explosion/spaCy/blob/5cbe21700b43bd2562ba49751fe646983d79dff4/spacy/gold.pyx#L238-L239

You will have to compile/install from source to make this change, though.

(I wouldn't recommend the sentence-per-document part except in very special cases, because then the parser can't split sentences, so I hope that gold tokenization and split-into-sentence-per-document are two separate options in the future.)

svlandeg:
Alternative quick fix until we fix the format internally properly: PR #4516

ryszardtuora:
@adrianeboyd 
Does -G do the split-into-sentence-per-document also on the data files which have been converted into .json with the explicit command to include 10 sentences per document? I would also hope for these functions to be separate.

adrianeboyd:
Yes, `-G` in v2.1.8 would split each sentence into a document no matter what the paragraphs looked like in the JSON. There are valid reasons to need this setting sometimes, but I also think it should be separate from gold tokenization.

honnibal:
I think #4516 seems like a less problematic fix, so I'll close this.

honnibal:
Having more problems with #4516, so I'll take another look at this approach instead.

svlandeg:
Sorry to hear #4516 gave more problems @honnibal  :(

Fundamentally though I guess the introduction of the `cats` variable has also introduced a backwards incompatibility, because `get_gold_tuples` should now look different. I guess that's why the training code got broken originally? But won't it be broken with this fix, too ?

honnibal:
@svlandeg Yeah, the root of the problem is the tuples stuff, which left no good options for adding a new variable. So all the solutions are pretty error prone. I do think it's a bit more error-prone to add an unlike element to the end of the list, which is why I want to explore this way more.

I'll port across the unit tests and the training script fixes to this branch.

PR #4415, opened 2019-10-09
Unify matcher get_ent_id and get_pattern_key

honnibal:
This is basically stabbing blindly at the ghost match problem, but it at
least seems like there was a bug previously here --- so this should
hopefully be an improvement, even if it doesn't fix the ghost match
problem.

## Checklist


svlandeg:
Linking Issue #4098 for future reference.

PR #4422, opened 2019-10-10
Adding noun_chunks to the Swedish language model (sv)

gustavengstrom:
Adds a syntax_iterators.py file to the Swedish language model

## Description
Follows the discussion [https://github.com/explosion/spaCy/pull/1882#issuecomment-361409573](url) I suggest adding a syntax_iterators also to the Swedish language model. This request corresponds  to the pull request #1913 which made the corresponding changes to Norwegian Bokmål model. Similarly the Swedish tag_map is also of the universal dependencies type.

### Types of change
new feature

## Checklist


gustavengstrom:
Note: the request is based on the assumption that the french syntax_iterators.py is applicable to also to the Swedish language model given it also uses the same UD dependency labelling scheme. At @ines or @honnibal it would be very useful to know exactly where the rules for the noun_chunker are derived from in this case i.e. how are the labels list in the noun_chunks method decided upon?

honnibal:
Thanks!

If possible, it'd be great to have a test case for this. The test should hard-code the parse information rather than running the model; you can use the `spacy.tests.util.get_doc` utility function to assemble the annotated `Doc` object. After creating the `Doc`, you can test that the `noun_chunks` work as expected.

gustavengstrom:
@honnibal I added a test method as suggested with three examples and expected outcomes (which I found in some online lecture notes).  I also added the"dobj" to the noun_chunks method. This was needed in order to generate the expected noun_chunks. However, I am not a trained linguist and it would be good to get some guidence on this issue. After spending several hours researching the topic online I am not convinced I have a clearcut definition of exactly what constitutes a noun chunk and how these may be derived from the UD annotation scheme. If you have some paper or links to how these should be constructed that would be great. UD provides limited guidence on the subject which I do not fully understand (https://universaldependencies.org/workgroups/newdoc/simple_noun_phrases.html). I did however find the following paper that should be useful for English (see: Investigating NP-Chunking with Universal Dependencies for English, https://www.aclweb.org/anthology/W18-6010.pdf ).

honnibal:
Thanks for taking so much time on this, I agree that there's a lot of detail to go through in the UD schema. Your rules look good. Thanks for that paper link as well --- that's a useful reference.

I think this is ready to merge, but it looks like you've got a temp file dragged into your PR -- that iPython notebook checkpoint. Could you remove that?

ines:
Already took care of that – merging! 👍

PR #4429, opened 2019-10-10
prevent zero-length mem alloc

svlandeg:
Following the PR #4423, I went through the entire code base to locate other places where we might be allocating zero-length memory. These should be prevented or flagged to ensure we don't run into more flaky behaviour in the future.

## TODO
Requires detailed review by @ines or @honnibal, as for some of these functions I'm not that familiar with the intrinsic working of the code. 
* Especially the changes to initializing **GoldParse**. The GH comparison is confusing, but I really only moved code + added an `if` statement.
* Also pretty unsure about the `_save_cached` function in **`tokenizer.pyx`**. Is it OK to just abort the whole thing if n `<= 0` ?
* In other cases I used `assert` when the normal flow should not hit a zero-length case. My reasoning is that these are not user-triggered cases, but programming errors.

~~I would suggest to not merge this immediately, to isolate its potential effects from PR #4423. But wanted to get this out so we don't forget.~~ I think this is fine now !

### Types of change
bug prevention !

## Checklist

### DIFF
--- a/examples/information_extraction/entity_relations.py
+++ b/examples/information_extraction/entity_relations.py
@@ -7,7 +7,7 @@ dependency tree to find the noun phrase they are referring to – for example:
 $9.4 million --> Net income.
 
 Compatible with: spaCy v2.0.0+
-Last tested with: v2.1.0
+Last tested with: v2.2.1
 """
 from __future__ import unicode_literals, print_function
 
@@ -38,14 +38,17 @@ def main(model="en_core_web_sm"):
 
 def filter_spans(spans):
     # Filter a sequence of spans so they don't contain overlaps
-    get_sort_key = lambda span: (span.end - span.start, span.start)
+    # For spaCy 2.1.4+: this function is available as spacy.util.filter_spans()
+    get_sort_key = lambda span: (span.end - span.start, -span.start)
     sorted_spans = sorted(spans, key=get_sort_key, reverse=True)
     result = []
     seen_tokens = set()
     for span in sorted_spans:
+        # Check for end - 1 here because boundaries are inclusive
         if span.start not in seen_tokens and span.end - 1 not in seen_tokens:
             result.append(span)
-            seen_tokens.update(range(span.start, span.end))
+        seen_tokens.update(range(span.start, span.end))
+    result = sorted(result, key=lambda span: span.start)
     return result
 
 
--- a/examples/training/training-data.json
+++ b/examples/training/training-data.json
@@ -8,7 +8,7 @@
             {
               "tokens": [
                 {
-                  "head": 4,
+                  "head": 44,
                   "dep": "prep",
                   "tag": "IN",
                   "orth": "In",
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # Our libraries
-cymem>=2.0.3,<2.1.0
+cymem>=2.0.2,<2.1.0
 preshed>=3.0.2,<3.1.0
 thinc>=7.1.1,<7.2.0
 blis>=0.4.0,<0.5.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 # Our libraries
-cymem>=2.0.2,<2.1.0
+cymem>=2.0.3,<2.1.0
 preshed>=3.0.2,<3.1.0
 thinc>=7.1.1,<7.2.0
 blis>=0.4.0,<0.5.0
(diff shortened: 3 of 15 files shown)


honnibal:
It does feel like this should be something that gets fixed in cymem?

I wonder whether it's my `memset` that causes this. Or maybe it's just that `PyMem_Malloc` doesn't guarantee that 0 allocations return NULL...

svlandeg:
I was going through the docs earlier, and it looks like returning non-NULL is actually the intended behaviour. It should only return NULL when the action fails, and it'll never fail to allocate no bytes...

honnibal:
Oh man. I see.

I think returning NULL from `cymem.Pool.alloc` is less error prone?

svlandeg:
Oh, ofcourse! I temporarily forgot `cymem` is also from Explosion :-)
I'll do that and rework this PR.

svlandeg:
@honnibal : So fixing this in `cymem` makes sense, but I still think we need to make the code in spaCy robust to this too. If we have `cymem.Pool.alloc` return `NULL`, spaCy code will throw an exception when running code like this: https://github.com/explosion/spaCy/blob/master/spacy/gold.pyx#L603. The exception is kind of ugly "returned NULL without setting an error", so we might as well just throw a ValueException in `cymem.Pool.alloc` which mentions specifically the zero-length problem. 

Actually, by making it throw this error I found one more spot in the code base where it was allocating zero length for `predicate_cache` in the `matcher` (see commit before this comment). 

So I think we need to have both this PR + the change to `cymem` that will automatically flag problematic `alloc`'s in the future.

honnibal:
Wow, I didn't realise I'd typed `cymem.alloc` as `except NULL`. So yeah definitely can't return NULL for empty pointers.

I guess I really forgot how that code worked, and just assumed I knew instead of checking...So yeah we do need to fix this in spaCy. I'm not sure we should raise an error in cymem for zero length --- it should work mostly like malloc and calloc, where you can indeed allocate a zero size pointer. It's up to the calling code to handle that correctly.

So I think the solution is to stop checking pointers for NULL, thinking that means the entry is empty. We just need to look at the lengths.

honnibal:
Hmm okay.

Let's make it an option in Cymem, disabled by default? Cymem does actually have other users, so I'm nervous about breaking their stuff.

I think this will change the size of `Pool` again though :(. That always sucks...It means we have to increment the major version number, and then roll out versions of Preshed and Thinc.

svlandeg:
Ok, the optional flag sounds like a good idea!

svlandeg:
@honnibal : something like [this](https://github.com/explosion/cymem/pull/15/commits/db9dfacd8bf76cdd17be95e08108ae94f2915ab6)? 

Do we then change all the calls to `Pool()` to `Pool(error_alloc_zero=util.env_opt("error_alloc_zero", True))` (or too verbose ?)

I just tested it on the old code (before this PR, i.e. on master), creating a  `GoldParse` with an empty doc would throw an error, unless you set `os.environ["SPACY_ERROR_ALLOC_ZERO"] = "0"`. In/after this PR, neither settings would throw an error because now the code prevents zero-length allocation up-front. But by setting `error_alloc_zero` by default to `True` (in spaCy) we should prevent future errors. In cymem it's set to `False` by default, not to break other's code.

honnibal:
@svlandeg Hmm. I'll keep thinking about this -- I'm kind of nervous about the changes to `cymem.Pool`, and I feel like the API isn't great, as you're suggesting.

Maybe we should just have a global in `cymem.cymem`? We really only need to set it during development to debug, to make sure we don't have one of these allocations.

svlandeg:
Or maybe we just throw a warning?

honnibal:
Throwing a warning by default can also lead to a lot of spam. Maybe a global setting, and if true, warn?

svlandeg:
Ok, I see your point. To be fair, once you start googling, you can find plenty of cases where people ran into this, and several blogposts warning of this kind of behaviour from malloc, we certainly weren't the only ones who got bit by this ;-)

svlandeg:
Ok, see https://github.com/explosion/cymem/pull/15.

This now works on spaCy's master branch:

```
    cymem.WARN_ZERO_ALLOC = True
    doc1 = nlp("")
    with pytest.warns(UserWarning):
        gp1 = GoldParse(doc1)

    cymem.WARN_ZERO_ALLOC = False
    doc2 = nlp("")
    gp2 = GoldParse(doc2)
```

honnibal:
Nice. The `WARN_ZERO_ALLOC` assignment is even backwards compatible, because it will just do nothing on the older cymem version.

And, somewhat comforting to hear it wasn't just me who was holding it wrong :).

PR #4446, opened 2019-10-15
Most similar bug

### LINKED ISSUE #4443: Prune vectors error
Same issue as #4386, #4388 doesn't quite fix it, as the first element of the tuple here: https://github.com/explosion/spaCy/blob/2d249a9502bfc5d3d2111165672d964b1eebe35e/spacy/vectors.pyx#L340 still has the shape of the full array instead of the batch. Pretty sure it just needs to be turned into this:

```sorted_index = xp.arange(scores.shape[0])[:,None][i:i+batch_size],xp.argsort(scores[i:i+batch_size], axis=1)[:,::-1]```

Will open a PR in case you agree. I would also suggest not sorting if n=1 since its not necessary so I'll also add that in the PR.

## Info about spaCy

* **spaCy version:** 2.2.1 (master)
* **Platform:** Linux-4.4.0-112-generic-x86_64-with-debian-stretch-sid
* **Python version:** 3.6.7

### PR DESCRIPTION
**Pull request by danielkingai2**
## Description

Fixes #4443

In https://github.com/explosion/spaCy/blob/2d249a9502bfc5d3d2111165672d964b1eebe35e/spacy/vectors.pyx#L340, the first element of the tuple needs to have the correct shape relative to the batch size, so should be replaced by `sorted_index = xp.arange(scores.shape[0])[:,None][i:i+batch_size],xp.argsort(scores[i:i+batch_size], axis=1)[:,::-1]`. Also there is no need to run the sorting code if only one most similar vector is requested.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -337,7 +337,7 @@ cdef class Vectors:
             scores[i:i+batch_size] = xp.partition(sims, -n, axis=1)[:,-n:]
 
             if sort:
-                sorted_index = xp.arange(scores.shape[0])[:,None],xp.argsort(scores[i:i+batch_size], axis=1)[:,::-1]
+                sorted_index = xp.arange(scores.shape[0])[:,None][i:i+batch_size],xp.argsort(scores[i:i+batch_size], axis=1)[:,::-1]
                 scores[i:i+batch_size] = scores[sorted_index]
                 best_rows[i:i+batch_size] = best_rows[sorted_index]
 
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -336,7 +336,7 @@ cdef class Vectors:
             best_rows[i:i+batch_size] = xp.argpartition(sims, -n, axis=1)[:,-n:]
             scores[i:i+batch_size] = xp.partition(sims, -n, axis=1)[:,-n:]
 
-            if sort and n > 1:
+            if sort and n >= 2:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by svlandeg** [comment]
Thanks for the PR!
This might not be very straightforward - but is there a way to write a unit test / regression test to check for the correct behaviour ?

**Comment by danielkingai2** [comment]
@svlandeg Sure, added a little test. See if it makes sense/is in the right place.

**Comment by danielkingai2** [comment]
Any idea when this fix might make it into a released version? It is blocking us from updating [scispacy](https://github.com/allenai/scispacy) to be compatible with the latest version of spacy, thanks :)

File: spacy/vectors.pyx [hunk 1]
```diff
@@ -336,8 +336,8 @@ cdef class Vectors:
             best_rows[i:i+batch_size] = xp.argpartition(sims, -n, axis=1)[:,-n:]
             scores[i:i+batch_size] = xp.partition(sims, -n, axis=1)[:,-n:]
 
-            if sort:
-                sorted_index = xp.arange(scores.shape[0])[:,None],xp.argsort(scores[i:i+batch_size], axis=1)[:,::-1]
+            if sort and n > 1:
```
honnibal: Tiny nitpick, but house style is to use `>=` and `<`, to make off-by-one errors more visually obvious.

```suggestion
            if sort and n >= 2:
```

honnibal:
Thanks!

PR #4447, opened 2019-10-15
add meta to Defaults

tamuhey:
Add `nlp.meta` field to `language.Defaults` so that `Language` interface users can provide default meta information, e.g. requirements.

For example, `ja.Japanese` requires `mecab-python3` package, so this class should add the package name into `meta["requirements"]`. (`ja.Japanese` is also modified in this PR as an example.)

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1,10 +1,8 @@
 # coding: utf8
 from __future__ import absolute_import, unicode_literals
 
-import atexit
 import random
 import itertools
-from warnings import warn
 from spacy.util import minibatch
 import weakref
 import functools
@@ -123,6 +121,7 @@ class BaseDefaults(object):
     writing_system = {"direction": "ltr", "has_case": True, "has_letters": True}
     single_orth_variants = []
     paired_orth_variants = []
+    meta = {}
 
 
 class Language(object):
@@ -157,7 +156,7 @@ class Language(object):
     }
 
     def __init__(
-        self, vocab=True, make_doc=True, max_length=10 ** 6, meta={}, **kwargs
+        self, vocab=True, make_doc=True, max_length=10 ** 6, meta=None, **kwargs
     ):
         """Initialise a Language object.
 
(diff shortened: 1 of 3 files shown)


tamuhey:
I want the main contributors of `ja.Japanese` to review this change.

honnibal:
@polm , what do you think?

It seems to me like a good change, although we might have to rethink some of the assumptions in our model training process. Various command-line commands might be affected too, especially `spacy init-model`.

polm:
I haven't used `nlp.meta` before, but I'm in favor of anything that allows us to be more up-front about the requirements of Japanese support. We already have a check for when mecab-python3 fails to import normally, I guess this would help clarify issues with serialization?

spacy/language.py
```diff
@@ -157,7 +156,7 @@ class Language(object):
     }
 
     def __init__(
-        self, vocab=True, make_doc=True, max_length=10 ** 6, meta={}, **kwargs
+        self, vocab=True, make_doc=True, max_length=10 ** 6, meta=None, **kwargs
```
ines: I'm slightly worried that this could have unintended side-effects. Reconciling the meta defined in the language defaults and the meta passed in by the user could also end up being problematic and lead to unexpected results. So I'm not sure language subclasses should be allowed to define the full meta.

I do see the point of adding the requirements, though. (If a user saves out the `nlp` object and packages the model, the Python package will include these dependencies in its setup.) So an alternative solution could be to introduce a `requirements` attribute on the `Defaults` and then put this together in the `Language.meta` property?

honnibal: Agree that `requirements` seems like a good idea. Is there a way to avoid the duplication though? Can we populate it from `importlib_meta` using the language code?

ines: Ahh, that's a clever idea! And yes, I think this should be possible. I'll look into this! 👍

ines:
Okay, so the more "elegant" way of populating this automatically based on the `extras_require` works, but is too hacky (see my experiment in #4483).

I'm not sure we want to hard-code those dependencies *again* into the `Language` classes (also because we're hoping to eventually phase out all the third-party depdencies for basic tokenization etc.). And for user code like custom language subclasses, I guess the best solution would be to write straight to the meta, which is also how we're doing it in `spacy-transformers`.

I'll close this PR because there's just no satisfying, future-proof solution for the underlying problem at this point. So this isn't really about the PR or the idea itself. (Thanks for your thoughtful contributions btw!)

PR #4455, opened 2019-10-16
✨  project: pySBD - Python Sentence Boundary Disambiguation

nipunsadvilkar:
##  pySBD - Python Sentence Boundary Disambiguation

Hi @ines @honnibal, It was great to meet you at Pycon India 2019, Chennai. This PR is related to rules-based sentence boundary segmenter ported from [Pragmatic Segmenter](https://github.com/diasks2/pragmatic_segmenter) (ruby) to [pySBD](https://github.com/nipunsadvilkar/pySBD) (python). Currently, it only supports the English language and would be supporting other languages soon.

## Checklist


website/meta/universe.json
```diff
@@ -1800,6 +1800,24 @@
...
+                "text = 'My name is Jonas E. Smith. Please turn to p. 55.'",
+                "segmenter = pysbd.Segmenter(language='en', clean=False)",
+                "print(segmenter.segment(text))",
+                "# ['My name is Jonas E. Smith.', 'Please turn to p. 55.']"
+            ],
+            "author": "nipunsadvilkar",
+            "author_links": {
+                "github": "nipunsadvilkar"
+            },
+            "category": ["scientific"]
```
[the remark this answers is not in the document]
nipunsadvilkar: Pardon! have pushed necessary changes

website/meta/universe.json
```diff
@@ -1800,6 +1800,30 @@
             "author_links": {
                 "github": "microsoft"
             }
+        },
+        {
+            "id": "python-sentence-boundary-disambiguation",
+            "title": "pySBD - python Sentence Boundary Disambiguation",
+            "slogan": "a rule-based sentence boundary detection that works out-of-the-box",
+            "github": "nipunsadvilkar/pySBD"
```
honnibal: Missing comma:

```suggestion
            "github": "nipunsadvilkar/pySBD",
```

honnibal:
Do you have plans to add a spaCy-compatible component (with entry-point) for the project? We want to make sure the universe has stuff that's directly usable with spaCy, so we'd be reluctant to add it to the universe before it has that.

And yeah, great talking to you in Chennai :).

nipunsadvilkar:
@honnibal: That is a reasonable point.

In contrast to spaCy `sentencizer` which does segmentation with rules and setting `tok._is_sent_start` attribute, pySBD is purely regex-based and works on a full text and gives out sentences as an output. 

To make it spaCy-compatible I would need to figure out a way to set those token attribute. For that, I would like to hear your thoughts on how to go about it?

ines:
@nipunsadvilkar Maybe you could use an approach similar to the one shown here? https://spacy.io/usage/rule-based-matching#regex-text

Basically, if you have the start and end character offset of the sentence in the raw text, you can use `doc.char_span` to convert the offsets to a spaCy `Span` object (if they map to valid tokens). This gives you the sentence spans, and the first token of each span should have `is_sent_start` set to `True`, and all other tokens should set `False`.

nipunsadvilkar:
@ines Have been thinking to do the same. Check out WIP branch - https://github.com/nipunsadvilkar/pySBD/tree/sentence-char-span.

In `sentence-char-span`, I'm setting a boolean parameter `char_span=True` which returns sentences along with their character offsets.
```python
import pysbd
text = "My name is Jonas E. Smith. Please turn to p. 55."
seg = pysbd.Segmenter(language="en", clean=False, char_span=True)
print(seg.segment(text))
# [('My name is Jonas E. Smith.', 0, 26), ('Please turn to p. 55.', 27, 48)]
```
> [(sent_str, start_char, end_char)]

ines:
@nipunsadvilkar Looks good! 👍

So I guess the spaCy component could look something like this? Untested and could maybe be more elegant, though.

```python
def set_sentence_boundaries(doc):
    seg = pysbd.Segmenter(language="en", clean=False, char_span=True)
    segments = seg.segment(doc.text)
    char_spans = [doc.char_span(start, end) for _, start, end in segments]
    start_token_ids = [span[0].idx for span in char_spans if span is not None]
    for token in doc:
        token.is_sent_start = True if token.idx in start_token_ids else False
    return doc 
```

nipunsadvilkar:
@ines @honnibal Have released pySBD `v0.2.0` with `char_span` functionality. Also added example script - [pysbd_as_spacy_component.py](https://github.com/nipunsadvilkar/pySBD/blob/master/examples/pysbd_as_spacy_component.py)

ines:
@nipunsadvilkar Nice! 🎉 Would you consider shipping the spaCy component with your package? (You could even expose it as an [entry point](https://spacy.io/usage/saving-loading#entry-points-components).) Then all the user would have to do is import `pysbd_sentence_boundaries` and add it to their pipeline. 

And we could justify adding it to the universe since it's a package that includes a spaCy plugin (rather than just a package that you can integrate with spaCy yourself, which is true for most packages).

nipunsadvilkar:
@ines : Can you please review this PR regarding entrypoints - https://github.com/nipunsadvilkar/pySBD/pull/42

ines:
@nipunsadvilkar Done, looks great and works! 👍

website/meta/universe.json
```diff
@@ -1801,6 +1801,34 @@
...
+            "title": "pySBD - python Sentence Boundary Disambiguation",
+            "slogan": "a rule-based sentence boundary detection that works out-of-the-box",
+            "github": "nipunsadvilkar/pySBD",
+            "description": "pySBD is 'real-world' sentence segmenter which extracts a reasonable sentences when the format and domain of the input text are unknown. It is a rules-based algorithm based on [The Golden Rules](https://s3.amazonaws.com/tm-town-nlp-resources/golden_rules.txt) - a set of tests to check accuracy of segmenter in regards to edge case scenarios developed by [TM-Town](https://www.tm-town.com/) dev team. pySBD is python port of ruby gem [Pragmatic Segmenter](https://github.com/diasks2/pragmatic_segmenter).",
+            "pip": "pysbd",
+            "category": ["scientific"],
+            "tags": ["sentence segmentation"],
+            "code_example": [
+                "import spacy",
+                "import pysbd",
```
ines: You don't even need this line anymore 🙂

```suggestion
```

Although I'd probably show the straightforward construction first, using the factory directly? Makes it a bit more obvious what's going on:

```python
from pysbd.util import PySBDFactory

nlp.add_pipe(PySBDFactory(nlp))
```

nipunsadvilkar: Agree! making it more readable :)

nipunsadvilkar:
@ines Have released `pysbd` `v0.2.1`. You can merge this PR and yes tweet about it 😄

Issue #4452, opened 2019-10-15
docs_to_json output


akornilo:
The output of the `spacy.gold.docs_to_json` format is inconsistent with the function description. The function says that it `RETURNS (list): The data in spaCy's JSON format.`. In reality, the function combines all the documents into one big json. 

Either the description or the function itself needs to be adjusted. I think that it should do the later - because that would be in line with the input for the training functions. I am not sure how the ids should be handled.

## Which page or section is this issue related to?
https://github.com/explosion/spaCy/blob/master/spacy/gold.pyx#L740

svlandeg:
Thanks for the report, good catch! 

It used to be so that the output was a list, but this was rewritten, see [here](https://github.com/explosion/spaCy/pull/3419/files#diff-74efcfe2a206cfbaa90445ffc47c4a61). So the best fix is probably to adjust the inline documentation you referred to as well as the [API documentation](https://github.com/explosion/spaCy/blob/master/website/docs/api/goldparse.md#golddocs_to_json-docs_to_json-tagfunction). Doyou want to go ahead and create a PR for this doc change?

akornilo:
I can make a PR .. but I still find it confusing - whats the reasoning behind storing multiple Docs as a single "doc"?

adrianeboyd:
The docs should definitely be clarified here, but in effect it's more of a naming issue than anything: currently the things called "paragraphs" in the training format are treated as documents during training.

I thought this was confusing, too. Some relevant background is here, where I made a similar suggestion:

https://github.com/explosion/spaCy/pull/4013

svlandeg:
Fixed by  #4456

PR #4456, opened 2019-10-16
Fix documentation for the docs_to_json function

akornilo:
Update documentation for the docs_to_gold function.

## Description
The

### Types of change
The `docs_to_json` function returns a dict not a list - i've updated the description to reflext the actual function of the code.

This is a PR for: https://github.com/explosion/spaCy/issues/4452

## Checklist

- [x ] My changes don't require a change to the documentation, or if they do, I've added all required information.


honnibal:
Thanks! You're right that there was some confusion around this.

PR #4459, opened 2019-10-17
Replace Entity/MatchStruct with SpanC

adrianeboyd:
## Description

Create more general `SpanC` to replace `Entity` (used by NER) and `MatchStruct` (used by PhraseMatcher).

(Intended for future use in one general `filter_spans()` so that there are not duplicate functions in `util` and the updated tokenizer in #4259.)

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/matcher/phrasematcher.pxd
+++ b/spacy/matcher/phrasematcher.pxd
@@ -4,6 +4,7 @@ from cymem.cymem cimport Pool
 from preshed.maps cimport key_t, MapStruct
 
 from ..attrs cimport attr_id_t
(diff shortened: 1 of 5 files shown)


adrianeboyd:
I tried to align it with Span (adding all the simple elements), but I don't know whether it's possible to include the vector?

Do you want me to rewrite Span as part of this, or possibly separately?

honnibal:
I doubt we'd want the vector, since it'd have to be a pointer and we'd end up in a bunch of difficulties. I'll go ahead and merge this.

Issue #4426, opened 2019-10-10
Matcher doesn't match when lemmas are set by lemma_lookup in spaCy 2.2.1


zrlaida:
## How to reproduce the behaviour

In `spacy==2.2.1`, the following code snippet doesn't match. Whereas, in `spacy==2.1.8`, it matches correctly.

```
import spacy

nlp = spacy.load("de_core_news_sm")

nlp.Defaults.lemma_lookup["therapie"] = "Therapie"
nlp.Defaults.lemma_lookup["versuch"] = "Versuch"

doc = nlp("ein therapie versuch")

matcher = spacy.matcher.Matcher(nlp.vocab)

pattern = [{'LEMMA': 'Therapie'}, {'LEMMA': 'Versuch'}]

matcher.add("test_id", None, pattern)

for match in matcher(doc):
    print("Match: ", match, nlp.vocab.strings[match[0]])
```

## Your Environment

* Operating System: macOS 10.14.6
* Python Version Used: 3.7.0
* spaCy Version Used: 2.2.1
* Environment Information:

ines:
The lemma lookups work slightly differently in v2.2 (see [release notes](https://spacy.io/usage/v2-2#incompat)), so writing to `Defaults.lemma_lookup` won't make a difference (we should probably remove the properties and raise an error here).

Lemmatization tables are now powered by the new [`Lookups` API](https://spacy.io/api/lookups). Also see the `Lemmatizer` docs for details: https://spacy.io/api/lemmatizer

If I change your code like this, it works as expected:

```python
lemma_lookup = nlp.vocab.lookups.get_table("lemma_lookup")
lemma_lookup["therapie"] = "Therapie"
lemma_lookup["versuch"] = "Versuch"
```

PR #4468, opened 2019-10-18
Remove lemma attrs on BaseDefaults

ines:
## Description

Also see #4426: Because the attributes like `lemma_lookup` are still available on the base defaults, user code that writes to them still works, even though they're not used anymore (and have been replaced by the lookups via `nlp.vocab.lookups`).

Open questions: Should we raise an error here instead, and if so, how should this be implemented? Should we make the attributes properties with a setter that raises an error?

### Types of change
enhancements

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -112,10 +112,6 @@ class BaseDefaults(object):
     tag_map = dict(TAG_MAP)
     tokenizer_exceptions = {}
     stop_words = set()
-    lemma_rules = {}
-    lemma_exc = {}
-    lemma_index = {}
-    lemma_lookup = {}
     morph_rules = {}
     lex_attr_getters = LEX_ATTRS
     syntax_iterators = {}


honnibal:
It needs to be a class attribute, so just using the `@property` decorator wouldn't work. I think we need to use the "descriptor protocol", which is the mechanism underlying properties. This seems to work:


```python

class RaisesError:
    def __init__(self, name):
        self.name = name

    def __get__(self, obj, type=None):
        raise AttributeError(f"Access to {self.name} defunct")

class Defaults:
    lemma_rules = RaisesError("lemma_rules")

# Should raise
print(Defaults.lemma_rules)
```

honnibal:
Actually I guess raising on attribute access could cause problems, I think we had something where some tooling was confused by that? A lot of people also just use tab completion.

I think we're probably over-thinking it and just removing them is better.

PR #4483, opened 2019-10-20
Auto-add language-specific dependencies from extras_require to meta

ines:
## Description

Based on the suggestion in #4447. Use `importlib_metadata` to get all language-specific dependencies (via `extras_require`) and auto-populate `nlp.meta["requirements"]` if packages are found for the current language.

Mostly intended to show how this could work. I'm really not sure if it's worth it – this seems to be the only way to implement that kind of feature and it feels really hacky and brittle.

### Types of change
enhancement

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -82,6 +82,10 @@ ko =
     natto-py==0.9.0
 th =
     pythainlp>=2.0
+ru =
+    pymorphy2==0.8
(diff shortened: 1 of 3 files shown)


honnibal:
Okay yeah, I think you're right --- let's not bother with this.

Issue #3618, opened 2019-04-19
Memory issues for long-running parsing processes


oterrier:
## How to reproduce the behaviour
Hi, 
I'm suspecting a memory leak when using intensively nlp.pipe(), my process is growing in memory and it looks that it never garbage collect. Do you think that it is possible ?
Here is the little python I'am using to reproduce the problem:

```
import random
import spacy
import plac
import psutil
import sys


def load_data():
    return ["This is a fake test document number %d."%i for i in random.sample(range(100_000), 10_000)]


def parse_texts(nlp, texts, iterations=1_000):
    for i in range(iterations):
        for doc in nlp.pipe(texts):
            yield doc


@plac.annotations(
    iterations=("Number of iterations", "option", "n", int),
    model=("spaCy model to load", "positional", None, str)
)
def main(model='en_core_web_sm', iterations=1_000):
    nlp = spacy.load(model)
    texts = load_data()
    for i, doc in enumerate(parse_texts(nlp, texts, iterations=iterations)):
        if i % 10_000 == 0:
            print(i, psutil.virtual_memory().percent)
            sys.stdout.flush()


if __name__ == '__main__':
    plac.call(main)
```

For me the output looks like
```
0 28.3
10000 28.3
20000 28.7
30000 28.9
40000 28.9
50000 29.1
60000 29.3
70000 29.7
80000 29.9
90000 30.1
...
420000 37.0
430000 37.2
440000 37.6
450000 37.8
460000 37.6
470000 38.0
...
920000 53.4
930000 53.9
940000 54.0
950000 54.2
960000 54.4
970000 54.4
980000 54.8
990000 54.9
1000000 54.9
```
And at the end the process pmap output is:
` total          4879812K
`
Can you confirm ?

## Your Environment

* **Operating System:** Ubuntu 18.04 Bionic Beaver (64-bit)
* **spaCy version:** 2.1.3
* **Platform:** Linux-4.15.0-47-generic-x86_64-with-debian-buster-sid
* **Python version:** 3.6.7
* **Models:** en, xx, fr

Best regards and congrats for the amazing work on SpaCy

Olivier Terrier

honnibal:
Unfortunately it's really hard to tell whether a Python program is leaking memory or whether the Python interpreter is just refusing to give memory back to the operating system.

This blog post explains some good tips: https://medium.com/zendesk-engineering/hunting-for-memory-leaks-in-python-applications-6824d0518774

[...]

PR #4486, opened 2019-10-21
Free pointers in parser activations

adrianeboyd:
## Description

Free the remaining pointers in the `ActivationsC` activations at the end of `_parseC()` in `Parser`.

This is the only definite spacy-related memory leak I found with valgrind using examples from #3618. (I think there may be some remaining issues related to cython/numpy.)

### Types of change

Bugfix.

## Checklist

### DIFF
--- a/spacy/syntax/_parser_model.pxd
+++ b/spacy/syntax/_parser_model.pxd
@@ -36,7 +36,9 @@ cdef WeightsC get_c_weights(model) except *
 
 cdef SizesC get_c_sizes(model, int batch_size) except *
 
-cdef void resize_activations(ActivationsC* A, SizesC n) nogil
+cdef ActivationsC alloc_activations(SizesC n) nogil
+
+cdef void free_activations(const ActivationsC* A) nogil
 
 cdef void predict_states(ActivationsC* A, StateC** states,
         const WeightsC* W, SizesC n) nogil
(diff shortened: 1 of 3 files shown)


honnibal:
Oh awesome! Nice work. Would you consider writing up a tutorial on how to use Valgrind on a Cython project? As I mentioned in Slack, I've always wanted to use it but have had trouble getting started.

I feel sheepish about this memory leak, as it could have been discovered the old-fashioned way by just grepping for malloc and calloc. I normally try to be careful when I'm using them, because they break the normal pattern of `cymem.Pool` and take some vigilance. I'm not sure why I messed this up.

adrianeboyd:
I only kinda sorta know what I'm doing, but I can sketch out the basics of how I did this, sure!

honnibal:
Ah, I think I see how I missed this. The allocations and the frees are happening in different modules, which makes the mistake much more likely.

I suggest we make it a bit easier to see that the `_parseC` loop owns that memory. The smallest change would be to just call `spacy.syntax._parser_model.resize_activations()` at the start of the loop. Probably better is to add some setup and teardown functions to `_parser_model.pyx`:

```python

cdef ActivationsC alloc_activations(int n_states) nogil:
    cdef ActivationsC acts
    memset(&acts, 0, sizeof(acts))
    resize_activations(&acts, n_states)
    return acts

cdef void free_activations(const ActivationsC* acts) nogil:
    free(acts.token_ids)
    free(acts.scores)
    free(acts.unmaxed)
    free(acts.hiddens)
    free(acts.is_valid)
```

Also, the `resize_activations` function should probably not be exported from `_parser_model.pyx`.

adrianeboyd:
However you think is best...

PR #4509, opened 2019-10-22
Bugfix/fix wikidata train entity linker

ZhuoruLin:
Fix a NoneType iteration error in bin/wikidata_train_entity_linker.py

## Description
The None default value for argument 'labels_discard' would cause a NoneType iteration error in the predict method of EntityLinker Pipeline. Passing in an empty list instead solved the problem.

### Types of change
Bugfix

## Checklist

### DIFF
--- a/bin/wiki_entity_linking/wikidata_train_entity_linker.py
+++ b/bin/wiki_entity_linking/wikidata_train_entity_linker.py
@@ -77,6 +77,8 @@ def main(
     if labels_discard:
         labels_discard = [x.strip() for x in labels_discard.split(",")]
         logger.info("Discarding {} NER types: {}".format(len(labels_discard), labels_discard))
+    else:
+        labels_discard = []
 
     train_data = wikipedia_processor.read_training(
         nlp=nlp,
--- a/.github/contributors/zhuorulin.md
+++ b/.github/contributors/zhuorulin.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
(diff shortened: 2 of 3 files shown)


bin/wiki_entity_linking/wikidata_train_entity_linker.py
```diff
@@ -77,6 +77,8 @@ def main(
     if labels_discard:
         labels_discard = [x.strip() for x in labels_discard.split(",")]
         logger.info("Discarding {} NER types: {}".format(len(labels_discard), labels_discard))
+    else:
+        labels_discard = []
```
svlandeg: This fix is definitely fine to have here, thanks!

I'm actually surprised this throws an error. I'm assuming this happens because of [this line](https://github.com/explosion/spaCy/blob/e38a37e5a6dc1320053508879460defe05c196a6/spacy/pipeline/pipes.pyx#L1301) in `pipes.pyx` ? I was assuming that `self.cfg.get("labels_discard", [])` would return `[]` if the key for `labels_discard` is passed with value `None` in `cfg`, but I guess that's not the case then?

ZhuoruLin: Hi @svlandeg . I think that's due to the fact that at line 107 the config the key `labels_discard` got passed in already. Therefore the `self.cfg.get("labels_discard", [])` would return the value for existing key rather than the default. 

I am a big fan of SpaCy's NEL! Thank you for all the nice works!

svlandeg: Yep, you're right :-) Could you perhaps also fix the code in `pipes.pyx` itself? In case that gets called from a different script than the `wikidata_train_entity_linker.py` one?

We could rewrite that line to two lines:
```
to_discard = self.cfg.get("labels_discard", [])
if to_discard and ent.label_ in to_discard: 
```

Which would make it works with both `None` as value as well as `[]`. Or something such :-)

Issue #4514, opened 2019-10-24
spacy_factories in nlp.meta


tamuhey:
I made many custom components of spacy, and sometimes I want to put same components in one nlp pipeline.
I tried changing one of their name and they can be put in one pipeline, but that made it impossible to restore with `spacy.load`.

So it is great if `nlp.meta` can holds `spacy_factories` to avoid confliction.

christian-storm:
I agree with the need for this.  I ended up writing my own. To achieve this I ended up writing a wrapper around spacy.  With this I can also swap in different tokenizers and a preprocessing pipeline.  It required me to override mac_doc and pipeline factories.

ines:
See #4517 🙂 We'll be shipping this as part of v2.2.2, so you can start trying it out before the `@component` decorator becomes "official" in spaCy v3. Decorated components will automatically register their factories and provide both their name and factory name (so spaCy knows that `textcat_a` needs the `textcat` factory, and so on).

tamuhey:
@ines Thanks! But I want to restore the pipe when  `spacy.load("packaged_dir")`. 
I think it is required that `cli.package` generate `spacy_factories` entry in `setup.py`, and the mapping of model name to its actual implementation is retrieved from nlp.meta.

Issue #3943, opened 2019-07-10
💫 Proposal: component decorator for pipeline metadata and static analysis


ines:
As I mentioned in [our talk at spaCy IRL](https://speakerdeck.com/inesmontani/spacy-and-explosion-past-present-and-future), here's a rough proposal for enhanced pipeline component metadata.

## Overview
spaCy lets you add your own components to the pipeline that take a `Doc`, modify it and return it. You can assemble your components however you like. Their order doesn't always matter, but it can: for instance, you could write a custom lemmatizer that runs after the part-of-speech tagger, uses the tokens and part-of-speech tags and sets the `Token.lemma`/`Token.lemma_` attribute. However, your custom component will always be a black box, and there's currently no way to express that it requires the `Token.pos` and `Token.tag` attributes to be set and that it writes to `Token.lemma`. You have to know this, and make sure to add your components correctly and do your error handling in your components if necessary. It's also currently not really exposed to the user *where* many attributes are set and what they're based on. If pipeline components could define what they set and what they require, spaCy could output more helpful diagnostics.

In terms of the implementation, there's a challenge here: How do we add metadata to a component? Pipeline components need to be callables that take a `Doc`, modify it and return it. If a component is a class with a `__call__` method, adding more attributes is easy – but we don't want to require pipeline components to be classes, because the "just write a function" thing is actually pretty great.

### Advantages and use cases
* Output more details on the attributes on the `Doc` and `Token` and where in the pipeline they're set. For example, you could see that the `"parser"` component sets `Token.dep` and `Doc.sents`.
* Raise better error messages for custom components. If your custom component depends on the POS tags and they're not set, spaCy will be able to tell you this before it even runs.
* Generate pipeline overviews and visualizations of the `Doc` passing through the pipeline

### Pseudocode example implementation

```python
@component(
    assigns={"token": {"lemma": True, "_": {"extended_lemma": True}}},
    requires={"token": {"pos": True}}
)
def custom_lemmatizer(doc):
    return doc
```

| Name | Type | Description |
| --- | --- | ---|
| `assigns` | dict | Attributes set or overwritten by the component. |
| `requires` | dict | Attributes required by the component. |
| `name` ? | str | Alternative default component name. |

The above `assigns` definition means that the component will set two attributes: `Token.lemma` (and by proxy, `Token.lemma_`) and the custom [extension attribute](https://spacy.io/usage/processing-pipelines#custom-components-attributes) `Token._.extended_lemma`.

[...]

PR #4517, opened 2019-10-24
Component decorator and component analysis

ines:
Related issues: #3943, #4514

## Description 

The automated pipe analysis is only available behind a flag for now. To enable it:

```python
import spacy.language
spacy.language.ENABLE_PIPELINE_ANALYSIS = True
```

### Component decorator

The new `@component` decorator can be used to mark pipeline components. The first argument is the name of the component and its factory. If the decorator is used, components will be automatically registered in the `Language.factories`. The decorator can be used for simple function components:

```python
@spacy.component("my_component")
def custom_component(doc):
    # Do something to the doc here...
    return doc
```

It also works for more complex class components:

```python
@spacy.component("my_component")
class MyComponent:
    def __init__(self, nlp):
        self.nlp = nlp

    @classmethod
    def from_nlp(cls, nlp, **cfg):
        # If the component provides a from_nlp classmethod, it will
        # be used in the factories to give access to the nlp object and config
        return cls(nlp)

    def __call__(doc):
        # Do something to the doc here...
        return doc
```

Since the decorator can't really know (and shouldn't guess) how to initialize the component and how to pass in the `nlp` object and `**cfg` via the factories, class components can now provide a `from_nlp` classmethod that's used in the factory, if it exists. (This is btw how we're already doing it internally for our projects, e.g. in `spacy-transformers`.)

### Component analysis and depdendencies

Component decorators can also be used to declare the annotations a component assigns or expects. `assigns` and `requires` each take a list of attributes in "dot notation" format, e.g. `token.pos`. Attribute names should only be added without underscore for simplicity, for instance, `token.pos` instead of `token.pos_`.

```python
@spacy.component(
    "my_component", 
    requires=["token.pos", "token.tag"], 
    assigns=["token.lemma", "token._.extended_lemma"]
)
def custom_component(doc):
    # Do something to the doc here...
    return doc
```

Attributes are validated when they're added via the decorator. When the component is added to the pipeline, it is analysed with respect to the components that are already present and a warning is shown if the requirements aren't fulfilled. For example, if no components *before* the above component assign `token.pos` or `token.tag`.

> ⚠️ **Important note:** This mechanism isn't intended to verify that the annotations are *actually* assigned or *actually* required. So we need to take the declarations made by the components at face value and leave it up to the user to make sure that a component does what it promises. 

```python
@component("c1", assigns=["token.pos"])
def c1(doc):
    return doc

@component("c2", assigns=["token.lemma"], requires=["token.pos", "token.tag"])
def c2(doc):
    return doc

nlp = Language()
nlp.add_pipe(c1)
# This will show a warning because c2 requires token.tag to be set, but no
# component that runs before it in the pipeline (c1) assigns it
nlp.add_pipe(c2)
```

#### Helper: get components that assign or require

The `get_assigns_for_attr` and `get_requires_for_attr` helpers will return a list of `(name, component)` tuples for pipeline components that assign a certain attribute. For example:

```python
assigns_tensor = get_assigns_for_attr(nlp.pipeline, "doc.tensor")
```

This is useful for components that need to do things like backpropagate (see `spacy-transformers`). A component can now easily check which other components need an update, check if they have a specific method it needs and then update them. This is a much better solution than hard-coding the individual component names.

We still need to decide if we want to name those helper functions differently, and whether to expose them via the `nlp` object.

#### Helper: print pipeline analysis

The `print_summary` helper takes an `nlp` object and prints an overview of its pipeline components and the attributes they assign and require, as well as any problems found in the setup. Example output:

```
============================= Pipeline Overview =============================

#   Component        Requires         Assigns               Retokenizes
-   --------------   --------------   -------------------   -----------
0   tagger                            token.tag             False      
                                      token.pos                        
                                                                       
1   parser                            token.dep             False      
                                      token.is_sent_start              
                                      doc.sents                        
                                                                       
2   merge_entities   doc.ents                               True       
                     token.ent_iob                                     
                     token.ent_type                                    
                                                                       
3   ner                               doc.ents              False      
                                      token.ent_iob                    
                                      token.ent_type                   


================================ Problems (3) ================================
⚠ 'merge_entities' requirements not met: doc.ents, token.ent_iob,
token.ent_type
```

### Pipe factories

The `Language` object now also exposes a `pipe_factories` property that maps component names in the pipeline to their factory names (also see #4514). This fixes a problem that would otherwise occur when having multiple components of the same type in the pipeline with different names.

```python
nlp.add_pipe(nlp.create_pipe("tagger"))
nlp.add_pipe(nlp.create_pipe("textcat"), name="textcat_a")
nlp.add_pipe(nlp.create_pipe("textcat"), name="textcat_b")
print(nlp.pipe_factories)
# {"tagger": "tagger", "textcat_a": "textcat", "textcat_b": "textcat"}
``` 

### TODOS

- [x] Analyse pipes in other places and methods that modify the pipeline (e.g. `nlp.remove_pipe`, `nlp.replace_pipe`).
- [x] Actually use the `pipe_factories` when initializing components from the `meta.json`.
- [x] Move error messages to `spacy.errors`.
- [ ] Decide how to expose the helpers that get all components that assign/require a specific attribute.
- [ ] Where should we move what's currently in `spacy.analyis`? I don't necessarily want yet another top-level module, but it can't live in `spacy.pipeline`, since that would cause circular imports.
- [x] Should we even support `Span` in the assigns/requires declarations? Spans are only views and components typically don't set anything on them (especially not in a way that makes sense to express here – like, for NER, the important annotations are on the `Doc` and the tokens, the fact that it also sets `Span.label` by proxy is kind of irrelevant).

### Types of change
enhancement

## Checklist


spacy/language.py
```diff
@@ -999,6 +997,52 @@ def from_bytes(self, bytes_data, exclude=tuple(), disable=None, **kwargs):
...
+        obj.factory = factory_name
+        obj.assigns = self.assigns
+        obj.requires = self.requires
+        obj.retokenizes = self.retokenizes
+
+        def factory(nlp, **cfg):
+            if hasattr(obj, "from_nlp"):
+                return obj.from_nlp(nlp, **cfg)
+            elif isinstance(obj, class_types):
+                return obj()
```
honnibal: Is it expected that we discard the `cfg` in this case?

ines: Yes, if the component class doesn't specify a `from_nlp` classmethod, i.e. a factory and a way to receive the `nlp` object and config parameters, it doesn't get them and we're not going to make any assumptions about how the class might need to be initialized.

spacy/language.py
```diff
@@ -999,6 +997,52 @@ def from_bytes(self, bytes_data, exclude=tuple(), disable=None, **kwargs):
...
+
+    def __call__(self, *args, **kwargs):
+        obj = args[0]
+        args = args[1:]
+        factory_name = self.name or util.get_component_name(obj)
+        obj.name = factory_name
+        obj.factory = factory_name
+        obj.assigns = self.assigns
+        obj.requires = self.requires
+        obj.retokenizes = self.retokenizes
```
honnibal: It would be nice to have a design that didn't even modify the object. There could be unexpected results if people have an attribute in their class that conflicts. `name` seems especially likely to cause problems.

Perhaps if we only assigned one attribute, `component_info`, that might be better?

Alternatively we could store the information separately in a global variable, and have helper methods like `get_factory_name(component)` which would look it up. We're already using `getattr` to access the attribute anyway, so it doesn't really seem worse to lookup the value with a function.

ines: > `name` seems especially likely to cause problems.

This has already been the default since we've introduced the new style of components, though. We're not writing to it, but we're expecting class components to have an attribute `name`.

> There could be unexpected results if people have an attribute in their class that conflicts.

I get what you mean, but the idea here is that these are some of the few hard-coded attributes we expect. We also need class components to be able to provide this data independently from the decorator – see the `cdef` classes like `EntityRecognizer` for instance, which can't be decorated and defines these attributes and registers itself in the factories instead.

> We're already using getattr to access the attribute anyway, so it doesn't really seem worse to lookup the value with a function.

I guess we could have a function `register_component` that takes the factory and the other configuration (`requires`, `assigns` etc.). This can be called from within the decorator, but also easily outside of it (e.g. to add the Cython classes).

honnibal:
I agree it makes sense to not support spans, since they're only views.

I'm not really sure about where to put the analysis stuff. Another top-level module isn't necessarily terrible? We can always make it `_analysis` if we don't want imports directly from there.

Issue #2328, opened 2018-05-14
Thoughts around adding features to NER model


rbhambriiit:
## Feature description
NER model to use parser, tagger features.

Background:
I have watched Spacy's video to get an idea on how NER model is trained based on Embed, Encode, Attend, Predict.

I have trained custom blank NER models on my dataset.

Idea:
The NER model should rely on other pipelines such as parser, tagger.
Features for pos, dependency label, previous-word-pos etc. should be considered at the embed stage.

I would like to hear more from @honnibal if this makes sense?
Is it somewhere already done. If yes, please help get me to the right place.

honnibal:
The easiest way to run this sort of experiment will be to build from source and hack at the source code. Then once you know whether your ideas work, you can implement them more cleanly using the API.

The function you're looking for is in `spacy/_ml.py`, the `Tok2Vec` function, specifically. This function is called by `spacy/syntax/nn_parser.py::Parser.Model`, to define the shape of the neural network. The `Tok2Vec` function defines the embedding and CNN layers of the network; the input to these layers is a list of docs, and the output is a list of 2d tensors.

You'll want to create embedding tables for the attributes you're interested in. You can see how this is done in the current implementation, as we already concatenate the embeddings of several features (the normalized form, the prefix, suffix and word shape.

Specifically, we have:

```python
    cols = [ID, NORM, PREFIX, SUFFIX, SHAPE, ORTH]
    with Model.define_operators({'>>': chain, '|': concatenate, '**': clone,
                                 '+': add, '*': reapply}):
        norm = HashEmbed(width, embed_size, column=cols.index(NORM),
                         name='embed_norm')
        prefix = HashEmbed(width, embed_size//2, column=cols.index(PREFIX),
                           name='embed_prefix')
        suffix = HashEmbed(width, embed_size//2, column=cols.index(SUFFIX),
                           name='embed_suffix')
        shape = HashEmbed(width, embed_size//2, column=cols.index(SHAPE),
                          name='embed_shape')
        if pretrained_vectors is not None:
            glove = StaticVectors(pretrained_vectors, width, column=cols.index(ID))

            embed = uniqued(
                (glove | norm | prefix | suffix | shape)
                >> LN(Maxout(width, width*5, pieces=3)), column=cols.index(ORTH))
        else:
            embed = uniqued(
                (norm | prefix | suffix | shape)
                >> LN(Maxout(width, width*4, pieces=3)), column=cols.index(ORTH))
```

You need to make five changes:

[...]

PR #4518, opened 2019-10-24
Refactor Tok2Vec to use architecture registry

honnibal:
Our hyper-parameter config stuff is a mess, which makes it very difficult to customize the models. The architecture registry is a good step forward. Now we just have to progressively refactor everything to use it.

This PR starts with the `Tok2Vec` builder, which had accumulated a rats-nest of different options. We refactor these into different architecture functions, and start building up the new `spacy.ml`.

## Checklist

### DIFF
--- a/spacy/_ml.py
+++ b/spacy/_ml.py
@@ -3,16 +3,14 @@ from __future__ import unicode_literals
 
 import numpy
 from thinc.v2v import Model, Maxout, Softmax, Affine, ReLu
-from thinc.i2v import HashEmbed, StaticVectors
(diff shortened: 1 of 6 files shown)


bratao:
@honnibal this looks amazing. If I understood correctly this change, it means that we could be able now to train a NER using custom features? Like in #2328  ?
😍

honnibal:
@bratao That's where this is going, yes! We have to do the same sort of stuff elsewhere in the library before you'll be able to write your own features easily though. This is just the first step.

PR #4522, opened 2019-10-25
Implement new API for {Phrase}Matcher.add (backwards-compatible)

ines:
## Description

Another not so great API. Managed to implement the new syntax in a backwards-compatible way.

```diff
patterns = [[{"TEXT": "hello"}, {"text": "world"}], [{"TEXT": "hi"}]]
- matcher.add("KEY", None, *patterns)
+ matcher.add("KEY", patterns)

- matcher.add("KEY", on_match, *patterns)
+ matcher.add("KEY", patterns, on_match=on_match)
```

### TODO

- [x] add docs
- [x] update internals (everywhere we're calling the methods that's not user-facing examples – this prevents our own code base from generating deprecation warnings etc. later on)
- [x] should we rewrite all internal tests to already use the new API?
- [x] add very basic validation check for `patterns` to make sure it's list of patterns

### Types of change
enhancement, matcher

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -504,6 +504,13 @@ class Errors(object):
     E175 = ("Can't remove rule for unknown match pattern ID: {key}")
     E176 = ("Alias '{alias}' is not defined in the Knowledge Base.")
     E177 = ("Ill-formed IOB input detected: {tag}")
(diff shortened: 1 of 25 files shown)


honnibal:
Nice! And yes, I think updating the tests is a good idea.

PR #4524, opened 2019-10-25
Fix GoldParse init

svlandeg:
The `GoldParse` init constructor has a catch-all `**_` which seems to have led to a subtle bug in `nn_parser.pyx`.

Personally I would prefer removing `**_` from the constructor, but this may break people's scripts (though the breaking might be informational). We can also remove it as part of the v.3.0 release. Unless there is a reason to keep this?

### Types of change
bug fix

## Checklist


honnibal:
I think we had the catch-all there for some backwards compatibility at some point. Removing it seems fine.

Issue #4525, opened 2019-10-26
gold.align seems to be broken


tamuhey:
## How to reproduce the behaviour

```python
>>> from spacy.gold import align
>>> tokens_a=["ab","c"]
>>> tokens_b=["a","b","c"]
>>> align(tokens_a, tokens_b)
(2, array([-1,  2], dtype=int32), array([-1, -1,  1], dtype=int32), {}, {})
```
I think `j2i_multi` (the last of returned items) should be `{0:0:,1:0}`

## Your Environment

* Operating System: mac OSX
* Python Version Used: 3.7
* spaCy Version Used: master

tamuhey:
created PR #4526

PR #4526, opened 2019-10-26
[#4525] fix gold.align

tamuhey:
from #4525 

- Fixed the bug described in #4525
- Made the `gold.align` code simpler and more efficient

### Types of change

bug fix and refactor.

## Checklist

### DIFF
--- a/spacy/gold.pyx
+++ b/spacy/gold.pyx
@@ -146,44 +146,6 @@ def align(tokens_a, tokens_b):
     return cost, a2b, b2a, a2b_multi, b2a_multi
 
 
-
-def old_align(tokens_a, tokens_b):
-    """Calculate alignment tables between two tokenizations.
-
-    tokens_a (List[str]): The candidate tokenization.
-    tokens_b (List[str]): The reference tokenization.
-    RETURNS: (tuple): A 5-tuple consisting of the following information:
-      * cost (int): The number of misaligned tokens.
-      * a2b (List[int]): Mapping of indices in `tokens_a` to indices in `tokens_b`.
-        For instance, if `a2b[4] == 6`, that means that `tokens_a[4]` aligns
-        to `tokens_b[6]`. If there's no one-to-one alignment for a token,
-        it has the value -1.
-      * b2a (List[int]): The same as `a2b`, but mapping the other direction.
-      * a2b_multi (Dict[int, int]): A dictionary mapping indices in `tokens_a`
-        to indices in `tokens_b`, where multiple tokens of `tokens_a` align to
-        the same token of `tokens_b`.
-      * b2a_multi (Dict[int, int]): As with `a2b_multi`, but mapping the other
-            direction.
-    """
-    if tokens_a == tokens_b:
-        alignment = numpy.arange(len(tokens_a))
-        return 0, alignment, alignment, {}, {}
-    tokens_a = [w.replace(" ", "").lower() for w in tokens_a]
-    tokens_b = [w.replace(" ", "").lower() for w in tokens_b]
-    cost, i2j, j2i, matrix = _align.align(tokens_a, tokens_b)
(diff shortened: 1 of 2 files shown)


tamuhey:
Below is a notebook on profiling the old and new `gold.align` functions.
https://gist.github.com/tamuhey/1f16a107030659ff75781889fbc37fd1

spacy/gold.pyx
```diff
@@ -73,9 +73,15 @@ def merge_sents(sents):
     return [(m_deps, m_brackets)]
 
 
+_NORM_MAP = {"``": '"', "''": '"'}
```
ines: I'm not sure we want to be hard-coding the normalization here – I guess this should at least be an argument the user can change? Also, if we do normalization, do we want to handle this differently and actually substitute the characters? @honnibal

adrianeboyd: I thought the same thing, but it turns out that this hard-coded in the existing _align code, so that's why it's implemented here. I think it would be a good opportunity to remove this since we're handling it more generally elsewhere now.

ines: Ahh I hadn't seen that (and it actually substitutes the characters, whereas this norm map change just substitutes the tokens?). But yes, in any case, I think we should definitely fix this then.

tamuhey: It seems to substitutes the tokens in this line (https://github.com/explosion/spaCy/blob/118030444902f4696a5ace360d7854a3218dadb7/spacy/_align.pyx#L189)
If this normalization is unnecessary, I remove this.

honnibal:
Really appreciate your help @tamuhey ! Thanks so much. I've had a fair bit of trouble with the alignment code over time.

honnibal:
I merged a bit hastily here, and missed something. Will make a commit to master to address it.

spacy/gold.pyx
honnibal: @adrianeboyd I think it's better to avoid mixing behavioural changes into refactors, because it makes it harder to go back and debug the refactor later. It's much easier if the refactor is supposed to give identical output.

spacy/gold.pyx
```diff
@@ -92,23 +98,52 @@ def align(tokens_a, tokens_b):
...
+        elif a.startswith(b):
+            cost += 1
+            if offset_b == 0:
+                b2a_multi[j] = i
+            j += 1
+            offset_b = 0
+            offset_a += len(b)
+        else:
+            assert "".join(tokens_a) != "".join(tokens_b)
+            raise ValueError(f"{tokens_a} and {tokens_b} is different texts.")
```
tamuhey: @honnibal Would you please change `is different` in this line to `are different`?

adrianeboyd:
This has fixed some of the subtok problems I was having (this is great!) but now I have cases where texts aren't being aligned like they were before. At least one problem is when the raw text starts with a whitespace character. A minimal example:

```
[
  {
    "id":0,
    "paragraphs":[
      {
        "raw":" a",
        "sentences":[
          {
            "tokens":[
              {
                "head":0,
                "dep":"ROOT",
                "tag":"A",
                "orth":"a",
                "ner":"U-DATE",
                "id":0
              }
            ],
            "brackets":[

            ]
          }
        ]
      }
    ]
  }
]
```

adrianeboyd:
Ah, wait, I think I managed to test this at just the wrong point and the failing overall alignment was what Matt fixed above.

Now it doesn't throw an error but I get incorrect results with the GoldParse for the same example:

```
# loaded through GoldCorpus:
gold.words  # ['a', 'a']
gold.ner    # ['B-DATE', 'U-DATE']
gold.labels # ['ROOT', 'ROOT']
```

The difference is `i2j_multi`:

```
cost, i2j, j2i, i2j_multi, j2i_multi = align([" ", "a"], ["a"])
i2j_multi   # {0: 0}
```

tamuhey:
@adrianeboyd Trying to fix this bug in (#4537)

PR #4527, opened 2019-10-26
adds support for merging Docs

### LINKED ISSUE #2229: feature: Merge multiple `Doc()` objects into one
When processing large documents, I usually process sentence by sentence. Then I have numerous `Doc()` objects per document. It'll be great if I could merge those objects into one then serialize/save to disk.

kognate:
Adds a method in Language to concat docs into a new Doc.

## Description

Fixes #2229 and adds `concat` method to merge multiple docs into one doc.  

### Types of change

This is a new (very small) feature.  

## Checklist

- [ X ] I have submitted the spaCy Contributor Agreement.
- [ X ] I ran the tests, and all new and existing tests passed.
- [ X ] My changes don't require a change to the documentation, or if they do, I've added all required information.

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -727,7 +727,7 @@ class Language(object):
                 pass
 
     def concat(self, *docs, join_delimiter=None):
-        """Concatinate documents. Yields a new `Doc` object.
+        """Concatenate documents. Yields a new `Doc` object.
 
         *docs (Doc): `Doc` objects to be concatenated
         join_delibiter (str): the delimiter to join the text of the
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -726,10 +726,10 @@ class Language(object):
             except StopIteration:
                 pass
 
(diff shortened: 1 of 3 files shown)


adrianeboyd:
Thanks for taking the time to get started on this! Unfortunately this isn't quite the approach we would like to support. If this is implemented, we would like not only the text but also the annotations to be copied into the new `Doc` so that it's not necessary to run the pipeline again.

One reason is that running the pipeline again is time-consuming, but the other main reason is that the annotation for the new combined text may not be identical to the annotation for the shorter texts. One example that came up recently in #3052 was that `nlp("spy")` and `nlp("I saw a spy.")` would not produce the same tags for `spy` with v2.2.

If you're interested in working on this as new PR, we would be happy to provide a bit of guidance about how to implement it using `Doc.to_array()` and `Doc.from_array()` to copy the annotations.

Issue #4529, opened 2019-10-27
[Bug] GoldParse instantiation with Misaligned tokens.


tamuhey:
## How to reproduce the behaviour

1. IndexError

```python
text = "A'B C"
words = ["A", "'", "B", "C"]
doc = tokenizer(text)
GoldParse(doc, words=words)
```
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-7-45b5f2ec7502> in <module>
      2 words = ["A", "'", "B", "C"]
      3 doc = tokenizer(text)
----> 4 GoldParse(doc, words=words)

gold.pyx in spacy.gold.GoldParse.__init__()

IndexError: list index out of range
```

2. TypeError

```python
text = "A-B"
words = ["A-B"]
doc = tokenizer(text)
GoldParse(doc, words=words)
```
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-9ae6959d65fe> in <module>
      2 words = ["A-B"]
      3 doc = tokenizer(text)
----> 4 GoldParse(doc, words=words)

gold.pyx in spacy.gold.GoldParse.__init__()

TypeError: list indices must be integers or slices, not NoneType
```

## Your Environment

* Operating System: mac OSX
* Python Version Used: 3.7
* spaCy Version Used: master

PR #4530, opened 2019-10-27
[#4529] fix: gold pyx

tamuhey:
from #4529

### DIFF
--- a/spacy/tests/regression/test_issue4529.py
+++ b/spacy/tests/regression/test_issue4529.py
@@ -1,0 +1,10 @@
+import pytest
+from spacy.gold import GoldParse
+
+
+@pytest.mark.parametrize(
+    "text,words", [("A'B C", ["A", "'", "B", "C"]), ("A-B", ["A-B"])]
+)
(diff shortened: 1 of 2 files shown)


spacy/tests/regression/test_issue4529.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from spacy.compat import is_python2
+from spacy.gold import GoldParse
+
+
+@pytest.mark.parametrize(
+    "text,words", [("A'B C", ["A", "'", "B", "C"]), ("A-B", ["A-B"])]
+)
+@pytest.mark.skipif(is_python2, reason="I don't want to spend time on python2 support")
```
ines: What's the problem on Python 2? I think you just need to add unicode declarations at the top of the file? The code changes in this PR shouldn't make a difference, since it's all Cython anyways and gets compiled in a cross-compatible way.

tamuhey: Thanks!

honnibal:
Thanks!

PR #4516, opened 2019-10-24
Match pop with append for training format

svlandeg:
## Description
PR #4226 added a `cats` field to the training format, which needs to be `pop()`'d off before iterating over the sentences and their annotations. Afterwards however, the field needs to be added back in with `append`, in case multiple pipeline components would process the same `TRAIN_DATA`. If we don't do this, we could be losing training data throughout the pipeline.

Caught this error while trying to run `ner_multitask_objective.py`, which was crashing. It runs again now, though I'm not sure whether it's still functional OK? I think it's been a while since this script was tested :-)

EDIT: Also now fixes #4402

Meanwhile, I'm working on rewriting the training tuple-format (as discussed) which would be quite an overhaul (+ not backwards compatible). But I thought this quick fix would be good to make sure the current `master` works properly in the meantime.

### Types of change
bug fix

## Checklist
- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #4533, opened 2019-10-27
Revert "Match pop with append for training format"

honnibal:
Reverts explosion/spaCy#4516

I'm trying to train now and having more problems with this :(. It's kind of hard to track down because there's interactions with other PRs.

Appending and popping the cats is really pretty bad. I think we should just not do that.


svlandeg:
If you revert - let's at least keep the unit test for `test_issue4402.py` (might be failing then)  ?

honnibal:
I should probably just resolve the merge conflict on the other branch rather than reverting. I'll see what's easier.

PR #4541, opened 2019-10-28
Put new alignment behind flag for v2.2.2 release

honnibal:
I merged @tamuhey 's alignment changes from #4525  to master, but we probably shouldn't make that change on a point release, especially before we've done more testing. I've re-added the previous function, and added a feature flag to `spacy.gold` that can be toggled to enable the new alignment.

We can switch over to the new alignment on v2.3 or v3.0.

### DIFF
--- a/spacy/gold.pyx
+++ b/spacy/gold.pyx
@@ -21,6 +21,7 @@ from .util import minibatch, itershuffle
 from libc.stdio cimport FILE, fopen, fclose, fread, fwrite, feof, fseek
 
 
+USE_NEW_ALIGN = False
 punct_re = re.compile(r"\W")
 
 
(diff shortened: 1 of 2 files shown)


spacy/gold.pyx
```diff
@@ -86,6 +86,47 @@ def _normalize_for_alignment(tokens):
...
+            i2j[i] = j
+            i2j_multi.pop(i)
+    for j, i in list(j2i_multi.items()):
+        if j2i_multi.get(j+1) != i and j2i_multi.get(j-1) != i:
+            j2i[j] = i
+            j2i_multi.pop(j)
+    return cost, i2j, j2i, i2j_multi, j2i_multi
+
+
+USE_NEW_ALIGN = False
```
ines: I think we should have this at the top of the file? It can easily get lost if it's all the way down on line 128.

PR #4542, opened 2019-10-28
Free pointers in parser activations (#4486)

### LINKED ISSUE #4538: Backport memory leak fix to patch spacy 2.0 and 2.1
## Feature description

My team (and I assume a lot of other companies) are using spacy 2.0 and 2.1 in various places. We have A LOT of trained models in spacy 2.0 and 2.1. It would be great to have the fix @adrianeboyd made here https://github.com/explosion/spaCy/pull/4486 for the memory leak ported back to the older versions as well since it's a really small change with a pretty huge impact.

i.e. It's much easier to upgrade to spacy 2.1.9 from 2.1.8 rather than 2.2.2 and retrain all the models we have.

Thanks!

And not sure this counts as a feature request but didn't seem like a bug either haha.

adrianeboyd:
## Description

Backport 3dfc764 to v2.1.x:

* Free pointers in ActivationsC

* Restructure alloc/free for parser activations

* Rewrite/restructure to have allocation and free in parallel functions
in `_parser_model` rather than partially in `_parseC()` in `Parser`.

* Remove `resize_activations` from `_parser_model.pxd`.

Fixes #4538.

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ pathlib==1.0.1; python_version < "3.4"
 jsonschema>=2.6.0,<3.1.0
 # Development dependencies
 cython>=0.25
-pytest>=4.0.0,<4.1.0
+pytest>=4.6.5
 pytest-timeout>=1.3.0,<2.0.0
 mock>=2.0.0,<3.0.0
 flake8>=3.5.0,<3.6.0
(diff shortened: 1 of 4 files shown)


ines:
Ugh, the test failure here is some `pytest` version compatibility stuff... I remember this coming up before. Let me see how I fixed this. I think all we need to do is upgrade the `pytest` requirement.

adrianeboyd:
Yeah, I ran the tests probably but didn't downgrade pytest.

honnibal:
I wouldn't worry about the failure on 3.5 OSX. It's because our wheels don't work there, and it's failing to compile Thinc on that platform. It's fine to release with that test failing.

kabirkhan:
Haha, our time difference is pretty evident here. Thank you all for the quick responses and for getting this out. Really appreciate the fast turnaround!

PR #4543, opened 2019-10-28
Example class for training data

svlandeg:
Refactor of the gold data format. 

**Introduces backwards incompatibility** because of 
* a different format for the `get_examples` parameter to `begin_training` (previously: `get_gold_tuples`)
* `rehearse()` now takes `examples` instead of `docs`
* `update()` and `get_loss` now take `examples` instead of `docs, golds`

### v2
* Because the tuples got too error-prone, this PR introduces an `Example` class that holds the annotations on sentence level (ids, words, tags, heads, deps, ents, morph and brackets) with a list of `TokenAnnotation` and an additional `DocAnnotation` class that keeps track of document-level annotations such as `cats` (extensible to e.g. `links` etc) 
* the format for `begin_training()`, `update()`, `rehearse()` and `get_loss()` is now a list of `Example` objects
* `call()` and `pipe()` can either take in `Example` or `Doc` objects, and can return either of the two (Doc by default)

### v1
* ~~Because the tuples got too error-prone, this PR introduces a `RawAnnot` class that holds the tuples on sentence level (ids, words, tags, heads, deps, ents and brackets) and an additional `DocAnnot` class that keeps a list of such `raw_annots`, together with the `cats` (extensible to e.g. `links` etc)~~
* ~~the format for `begin_training` is now a list of `(text, doc_annot)` tuples~~
* ~~renamed some fields to reflect the new data format~~

### TODO
- [x] check the `ud_train` script
- [x] check the `conllu` script
- [x] check the `ner_multitask_objective` script
- [x] check the `wikidata_train_entity_linker` script
- [x] writer proper errors    
- [ ] document format change              

### Types of change
enhancement / refactor

## Checklist


spacy/gold.pxd
```diff
@@ -36,6 +36,19 @@ cdef class GoldParse:
 
     cdef readonly list cand_to_gold
     cdef readonly list gold_to_cand
-    cdef readonly list orig_annot
 
 
+cdef class RawAnnot:
```
honnibal: We could just make this a normal class? We would lose the compile-time checking, but gain a bit of flexibility.

spacy/gold.pxd
```diff
@@ -36,6 +36,19 @@ cdef class GoldParse:
...
+    cdef readonly list ids
+    cdef readonly list words
+    cdef readonly list tags
+    cdef readonly list heads
+    cdef readonly list deps
+    cdef readonly list ents
+    cdef readonly list brackets
+
+
+cdef class DocAnnot:
```
honnibal: I don't have a good solution, but as a set of three names, `GoldParse`, `RawAnnot` and `DocAnnot` aren't great. I'm sure we'll be able to do better.

spacy/gold.pyx
```diff
@@ -54,23 +51,25 @@ def tags_to_entities(tags):
     return entities
 
 
-def merge_sents(sents):
-    m_deps = [[], [], [], [], [], []]
-    m_cats = {}
+def merge_sents(raw_annots):
```
honnibal: Would it be metter to make this a classmethod constructor of `RawAnnot`?

honnibal:
I can already feel the improvements from this! Will be great to have this straightened out.

The confusion of having different classes for these related purposes does worry me though. Is there a way we could introduce one class to replace the need for these things? Maybe it could own or produce the `GoldParse` class, so that the `GoldParse` becomes an implementation detail that people don't really see?

spacy/gold.pxd
svlandeg: Yea, I agree. I started with the name "RawAnnot" to reflect the field that is stored in GoldParse with the original (non-aligned) annotations (previously called `self.orig_annot`, now just `self.orig`), but it's also used at times in the codebase where its function is more like the GoldParse itself... There's a kind of weird dependency. I think you're right that `GoldParse` should become an implementation detail.

spacy/gold.pyx
adrianeboyd: I think it should to be a method in the `DocAnnot`, which has an internal list of `RawAnnot`s, which this function should merge together into one long interinal `RawAnnot`.

spacy/gold.pxd
adrianeboyd: I think it could be something like`GoldInstance` that stores what is a document (=paragraph) in the current format. A `GoldInstance` contains a raw text, a list of the sentence/token level info as `RawAnnot`, and also some data structure that stores doc-level annotation like cats. A `GoldInstance` could provide aligned/preprocessed `GoldParses` with options to split sentences, use gold tokenization, etc. (Do you want to keep the name `GoldParse`?)

spacy/syntax/nonproj.pyx
```diff
@@ -77,39 +78,42 @@ def decompose(label):
...
                     if is_decorated(label):
                         freqs[label] = freqs.get(label, 0) + 1
-            prepro_sents.append(
-                ((ids, words, tags, proj_heads, deco_labels, iob), ctnts))
-        preprocessed.append((raw_text, prepro_sents))
+            proj_annot = RawAnnot(ids=raw_annot.ids, words=raw_annot.words, tags=raw_annot.tags,
+                                  heads=proj_heads, deps=deco_deps, ents=raw_annot.ents,
+                                  brackets=raw_annot.brackets)
+            prepro_annots.append(proj_annot)
+        prepro_doc = DocAnnot(raw_annots=prepro_annots, cats=doc_annot.cats)
```
adrianeboyd: I don't like having `cats` as a named keyword here. I think we should be able to support various doc-level annotation without API modifications.

svlandeg: with a `dict` ?

svlandeg: actually, in this spot, we can just modify the `DocAnnot.raw_annots` in place (if they're not read-only)

adrianeboyd: I'm not a huge fan of dicts, either. They get very confusing/undocumented very quickly and a named class with named properties would be easier to understand and validate. Everything needs to be renamed, but something like `RawAnnot` -> `TokenAnnot` and this would be `DocAnnot` (and `DocAnnot` would be something else more container-sounding).

honnibal: A counter-point to consider is that APIs which require non-basic objects as arguments are pretty miserable. The misery goes up exponentially the deeper the tree of objects you require.

If we're going to ask the user to pass a non-basic object for the annotations to `begin_training()` or `update()`, I'm pretty reluctant to have that object take a non-basic object to instantiate. A third level --- where it takes an object that *also* takes an object --- is out of the question.

adrianeboyd: Maybe I didn't explain that well. I didn't mean for it to be nested, just renamed. I think it's weirder to have a class (`RawAnnot`, potentially renamed to `TokenAnnot`) for token-level annotations and not a similar kind of class for doc-level annotations. Both would be one level deep in the `AnnotationThingy`.

svlandeg: We currently have a [convertor function](https://github.com/explosion/spaCy/pull/4543/files#diff-b6b3e43f85a70fd78f0b9e8395857783R454) for `nlp.update()` that transforms a tuple+dict format into the new objects (cf [test_gold.py#L248](https://github.com/explosion/spaCy/blob/533308b0719478c2d533a9f8647580e04238be71/spacy/tests/test_gold.py#L248)) - I guess we'd want something similar for begin_training(), to avoid some of these complexities? (this part should actually be backwards compatible)

svlandeg:
Intermediate update:

- introduced `Example` class as the one anchor class
- `Example` stores a `Doc` or `raw_text` in its `doc` field (there's also a `text` property to get the text specifically)
- `Example` stores annotations on the level of tokens (list of `TokenAnnotation`) and on the level of docs (`DocAnnotation`)
- the internals can be kept hidden from the user by providing `to_dict` and `from_dict` methods, also convenient for serialization with `srsly`
- to build an `Example` object, the methods `example.add_token_annotation()` and `example.add_doc_annotation()` can be used that hide the complexity of the nested classes
- an `Example` can provide `GoldParse` objects with `get_gold_parses()`, specifying whether the sentences should be merged or not

TODO:
- update `nlp.pipe` and `call` to also use `Example`
- rerun CLI scripts to check nothing broke
- more unit tests

spacy/gold.pyx
```diff
@@ -639,21 +644,246 @@ def _consume_ent(tags):
...
+        return {"cats": self.cats, "links": self.links}
+
+
+cdef class Example:
+    def __init__(self, doc_annotation=None, token_annotations=None, doc=None,
+                 make_projective=False, ignore_misaligned=False, goldparse=None):
+        """ Doc can either be text, or an actual Doc """
+        self.doc = doc
+        self.doc_annotation = doc_annotation if doc_annotation else DocAnnotation()
+        self.token_annotations = token_annotations if token_annotations else []
```
adrianeboyd: I think it would be better to go ahead and represent sentences differently rather than using a list of `TokenAnnotation`s here. We could have `self.token_annotation = TokenAnnotation()` to be consistent with `DocAnnotation`, a `sent_start` (or similar) attribute as a list within the `TokenAnnotation` with 0/1 or True/False values.

This would make the sentence annotation less privileged and more explicit. It's also a good bit of the work in getting a statistical sentence segmenter working, so I'd be doing it anyway in the near future, and then we don't have the awkward list here.

(I am happy to make the changes myself, but it might be easier to follow the development if this is merged into a branch here first?)

svlandeg: Yea I think it's a good idea to merge this in a branch of its own so you can add your stuff on top of it. I think you have a better idea of what is needed for the statistical sentence segmenter. 

I just need to fix a few more unit tests first before we can merge this - and it'll require some more experimentation with the CLI scripts etc too, to make sure nothing broke. There's ample opportunity for bugs in this overhaul ...

adrianeboyd: As a note to myself so I don't forget: this should actually work like `token.c.sent_start` with ternary values. It will be necessary to handle conflicting `sent_start` vs dependency parse segmentation.

svlandeg:
I think this can be merged into `develop` to keep experimenting with it a bit. I tried out several scripts but probably a bit more toying around with train scripts + some additional unit tests would help.

adrianeboyd:
There are a few bugs that need to be fixed before this is merged into `develop`. Could you merge it into a separate branch so I can work on it more easily? Problems so far:

* `merge_sents()` turns 0 heads into None (easy to fix)
* train sentences aren't being projectivized correctly (I'm still working on it)

svlandeg:
Referring to follow-up PR #4624 for future reference in case anyone's following the thread of logic :-)

PR #4553, opened 2019-10-30
Add "spaCy Server" to spaCy Universe

neelkamath:
## Description
**Rationale behind making a new project**

I _really_ wanted to use one of the existing two HTTP APIs for spaCy, but they didn't fit my needs due to poor documentation and lacking infrastructure. I had thought for several days about patching the original repositories or forking them, but they had serious issues such as security and poor routing (e.g., used an HTTP GET request for large text which may not fit in the URL instead of an HTTP POST request where the request body can have arbitrarily large amounts of text which doesn't require percent encoding) which made it faster and more sustainable to make my own solution.

### Types of change
Added a project to the spaCy Universe JSON file

## Checklist

### DIFF
--- a/.github/contributors/neelkamath.md
+++ b/.github/contributors/neelkamath.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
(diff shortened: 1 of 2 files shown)


ines:
Thanks, this looks very useful!

Our REST services were mostly designed for our demos and pretty basic (still need to switch them over to use [FastAPI](https://github.com/tiangolo/fastapi) btw). So it totally makes sense to have this as a separate project intended for general-purpose use 🙂

neelkamath:
Thanks!
For any potentially confused future viewer of this PR, the other two spaCy HTTP APIs I was referring to were third party since the official one is great but is for demos and not general NLP.

PR #4579, opened 2019-11-04
Fix conllu script

svlandeg:
## Description
* `Token.set_extension` needs `force=True` because both `ud_train.py` and `conllu.py` set this global property
* Fix order of arguments in documentation
* Set default encoding of file reader to UTF-8 (needed for Windows)
* Added example `conllu-config.json`

### Types of change
bug fix

## Checklist


examples/training/conllu.py
```diff
@@ -290,9 +289,9 @@ def get_token_conllu(token, i):
     return "\n".join(lines)
 
 
-Token.set_extension("get_conllu_lines", method=get_token_conllu)
-Token.set_extension("begins_fused", default=False)
-Token.set_extension("inside_fused", default=False)
+Token.set_extension("get_conllu_lines", method=get_token_conllu, force=True)
```
ines: Can we just move these inside the `main` function? Or does this not solve the problem? We typically avoid `force=True`, since this can lead to unexpected results and should really just be a last resort.

svlandeg: Sure. The other script already had `force` set to `True`, but I can change both!

PR #4583, opened 2019-11-04
Add load_from_docbin example

honnibal:
This example adds about 700k words of parsed Reddit data, which loads in about 11s. This should be pretty helpful for writing more examples of using the Doc object. For instance, we can add some more realistic examples using the parse tree navigation, or perhaps even use the parsed data to make examples of training.

Perhaps the example could be linked in the docs as well?


ines:
Closing this as discussed, since it's probably not a good idea to add 5mb of data to the examples. One file doesn't matter, but we want to avoid accumulating repo bloat here. So the data should be uploaded somewhere else.

PR #4584, opened 2019-11-04
Replace function registries with catalogue

ines:
## Description

This PR replaces our function registries (e.g. `@register_architecture`) and entry point handling with our new library, [`catalogue`](https://github.com/explosion/catalogue) that was designed for exactly this purpose.

We're able to remove a lot of code and helper functions, and streamline the way users can register custom functions. 

```python
@registry.architectures.register("xy")
class my_custom_architecture():
    # ...

@registry.languages.register("xy")
class MyCustomLanguage(Language):
    # ...

registry.displacy_colors.register("colors", func={"XY": "green"})
```

```python
# Will check both registered functions and entry points
all_architectures = registry.architectures.get_all()
custom_architecture = registry.architectures.get("custom_architecture")
```

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,12 +6,12 @@ blis>=0.4.0,<0.5.0
 murmurhash>=0.28.0,<1.1.0
 wasabi>=0.4.0,<1.1.0
 srsly>=0.1.0,<1.1.0
+catalogue>=0.0.7,<1.1.0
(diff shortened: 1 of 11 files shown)


ines:
Merging this as internals so we can see how it goes. If there are no problems, I'll document it for the next release.

PR #4587, opened 2019-11-05
Update example for __setting entity annotations__

Quemoy:
## Description
On the documentation page for [linguistic feature](https://github.com/explosion/spaCy/blob/master/website/docs/usage/linguistic-features.md), in the section __Setting entity annotations__, the example uses "FB" as an unrecognized entity that needs to be set in the doc. However, it seems that with the `en_core_web_sm` model, FB is now recognized as an entity, therefore breaking the example. 

Output with the error from running the example in Binder:
```
Before [('FB', 0, 2, 'ORG')]
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-5dfcd9e76b96> in <module>
      9 
     10 fb_ent = Span(doc, 0, 1, label="ORG") # create a Span for the new entity
---> 11 doc.ents = list(doc.ents) + [fb_ent]
     12 
     13 ents = [(e.text, e.start_char, e.end_char, e.label_) for e in doc.ents]

doc.pyx in spacy.tokens.doc.Doc.ents.__set__()

ValueError: [E103] Trying to set conflicting doc.ents: '(0, 1, 'ORG')' and '(0, 1, 'ORG')'. A token can only be part of one entity, so make sure the entities you're setting don't overlap.
```

There needs to be another commonly used symbol for a company that does not get correctly tagged by the small model. "AAPL" is sometimes used for Apple Inc. but does not get tagged by the small model.

### Types of change
Change to documentation

## Checklist

### DIFF
--- a/.github/contributors/Quemoy.md
+++ b/.github/contributors/Quemoy.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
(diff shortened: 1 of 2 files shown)


ines:
Thanks for flagging this! 👍

Maybe it's a better idea to not use the model in those examples and do `nlp = English()` instead? Having examples depend on a specific model's predictions is not that great, because it causes problems like this and doesn't generalise well.

PR #4596, opened 2019-11-06
Add tokenizer explain() debugging method

### LINKED ISSUE #4573: prefix_search overriding token_match in tokenizer
This may be working as intended, but I haven't seen it documented anywhere and for me it was unexpected behavior, so I figured I'd raise it as an issue. 

If you add a new token_match to the tokenizer that conflicts with something in the prefix_search then the prefix_search will win. So, for instance, if you compile something like r"#\w+" as your token match (to tokenize hashtags) it won't work because '#' is included in the default prefix_search. You can remedy this by explicitly removing '#' from the prefix_search. 

This looks like it might be new behavior because I've found examples where folks just used token_match to tokenize hashtags without changing prefix_search and it seemed to work for them. At any rate, whether this behavior is intended or not, I think the precedence of these functions should be clearly documented in the tokenizer section of the reference.
 
I'm on Python 3.6.5 with spaCy 2.2.2

This code:
```
import spacy
import re
from spacy.tokenizer import _get_regex_pattern
re_token_match = _get_regex_pattern(nlp.Defaults.token_match)
re_token_match = f"({re_token_match}|#\w+)"
nlp.tokenizer.token_match = re.compile(re_token_match).match
text = "I like #fresh #hashtags"
doc = nlp(text)
print([e for e in doc])
```
Produces this output:
`[I, like, #, fresh, #, hashtags]`

But if you modify it as follows:
```
import spacy
import re
from spacy.tokenizer import _get_regex_pattern
re_token_match = _get_regex_pattern(nlp.Defaults.token_match)
re_token_match = f"({re_token_match}|#\w+)"
nlp.tokenizer.token_match = re.compile(re_token_match).match

# begin new stuff
_prefixes = list(nlp.Defaults.prefixes)
_prefixes.remove('#')
nlp.tokenizer.prefix_search = spacy.util.compile_prefix_regex(_prefixes).search
# end new stuff

text = "I like #fresh #hashtags"
doc = nlp(text)
print([e for e in doc])
```

It produces this output: 
`[I, like, #fresh, #hashtags]`

### LINKED ISSUE #4645: Custom tokenizer token_match doesn't seem to take proper precedence
So I'm using one of the non-english models to do some tokenization. Specifically I'm interested in using the french tokenizer. While it works pretty well out-of-the-box, there are a few extra cases I'd like to add as special rules. Using [this doc](https://spacy.io/usage/linguistic-features#how-tokenizer-works), I added a custom tokenizer as follows:
```
import re
import spacy

nlp = spacy.load("fr_core_news_sm", disable=["tagger", "parser", "ner"])
hash_re = re.compile(r"#[\w]+")
def custom_tokenizer(nlp):
         return Tokenizer(nlp.vocab,
                   rules=nlp.Defaults.tokenizer_exceptions,
                   prefix_search=spacy.util.compile_prefix_regex(nlp.Defaults.prefixes).search,
                   suffix_search=spacy.util.compile_suffix_regex(nlp.Defaults.suffixes).search,
                   infix_finditer=spacy.util.compile_infix_regex(nlp.Defaults.infixes).finditer,
                   token_match=hash_re.search)

nlp.tokenizer = custom_tokenizer(nlp)
doc = nlp("C'est une j'ai #blah")
print([t.text for t in doc])
```
However, this produces:
```
["C'", 'est', 'une', "j'", 'ai', '#', 'blah']
```
thereby not keeping the #blah token intact. My understanding given the docs is that the `token_match` field provides a regex that takes precedence over other rules (in this case, the trouble-maker i think is the `prefix_search` default regex). However, it seems like that regex precedence isn't being respected here.

Am I doing something wrong? If this is not the way to add custom regex rules on top of existing tokenizers, then what is the proper way?

## Your Environment

* **spaCy version:** 2.1.8
* **Python version:** 3.6.9

Thanks for the help!

adrianeboyd:
## Description

This is a series of updates to the tokenizer and tokenizer docs in order to sync the docs and the current tokenizer implementation. A new tokenizer function makes debugging the tokenizer a bit easier by implementing a slow debugging version that returns labels for each token about the pattern or rule matched, based on the pseudo-code in the docs.

There are some minor changes to tokenizer to bring it in sync with the intended behavior based on the docs. While updating the pseudo-code in the docs, it made sense to make a working version to check the details, so I expanded that just a bit to add some better debugging functionality.

I think keeping this in sync with the actual tokenizer will be a similar amount of effort to keeping the pseudo-code in the docs in sync. Instead of being in the tokenizer itself, it could also be an example/demo script that takes `nlp.tokenizer` as an argument, but I liked being able to implement unit tests that compared its behavior to the actual tokenizer. I wish that it were easier to add tests for more of the tricky cases, since I haven't formally tested that the behavior is identical for all of the unusual test cases.

It's also kind of unsatisfying that it doesn't handle whitespace tokenization, but this isn't usually a source of confusion for users and I think adding it would make the pseudo-code harder to read, and it's already a bit too long. I wouldn't be opposed to adding it, though.

Fixes #4573, fixes #4645.

**Edited**: implements `explain()` method that returns a list of `(pattern_string, token_string)` tuples to avoid the hacky displacy usage. Easy displacy integration is postponed to a future PR.

----

### Outdated description of displacy integration:

In the initial version, the information is stored in a `Doc` with the debugging information saved on the tags. This is a bit hacky since there's no way to mark the `Doc` as **not-for-actual-use**, but it makes for easy visualization. Alternatively, the labels could also be stored on a custom attribute and the visualization would require a few extra steps.

debug-tok

I also tried visualizing the labels as entity spans, but the default entity visualization was pretty hard to read with labels on every token. With custom templates, it could be made a bit more readable and having the information as spans might be better than as hacky tags. (Or there could be a "tag" visualization where you can specify which attribute to display?)

### Types of change

Docs, enhancement.

## Checklist


ines:
I really like the visualization and doc debugging!

I wonder if we could separate the visualizer hacks from the tokenizer debug tags and have the method (or helper function) only return a list of tags, like `["PREFIX", "SPECIAL1"]` and so on. Creating the debug `Doc` could then happen within `displacy` in a new `tokens` visualizer/renderer, since it's really mostly a hack for the visualization. The debug doc itself isn't really that useful itself, but the debug *tags* are.

Also, if we ever want to make it nicer and more custom, it'll be easy to replace and won't require changing user-facing API. (I could try building a renderer that produces something [like this](https://spacy.io/tokenization-57e618bd79d933c4ccd308b5739062d6.svg), which might be cool?)

adrianeboyd:
I originally had it return `[("prefix", "'"), ("token", "a"), ...]` and it would be easy to do that instead. I do think you need the tokens because I want to confirm that it's doing the same thing as the normal tokenizer, otherwise it's not particularly useful.

I think a general-purpose `tags` kind of visualizer would be more useful. It's a pain to have to set up a parsed document just to view the POS tags, for instance.

ines:
> I originally had it return [("prefix", "'"), ("token", "a"), ...] and it would be easy to do that instead. 

Sure, that'd work!

> I think a general-purpose tags kind of visualizer would be more useful. It's a pain to have to set up a parsed document just to view the POS tags, for instance.

The central user-facing API takes a `Doc`, because that's often main use case, and the visualizers are task-specific, since that's how people typicall think about them. But you can always use displaCy in ["manual mode"](https://spacy.io/usage/visualizers#manual-usage) and pass in a dict.

The "general-purpose tags" visualizer would be something we could implement as a new renderer (first just calling into the dependency renderer, later I could implement a nicer and more custom version). And then we could expose that as the `tokens` visualizer and possibly reuse it for other things in the future?

website/docs/api/tokenizer.md
```diff
@@ -128,6 +128,25 @@ and examples.
 | `string`      | unicode  | The string to specially tokenize.                                                                                                                                        |
 | `token_attrs` | iterable | A sequence of dicts, where each dict describes a token and its attributes. The `ORTH` fields of the attributes must exactly match the string when they are concatenated. |
 
+## Tokenizer.explain {#explain tag="method"}
```
ines: This is great! 👍

ines:
If it's not too much of a hassle, maybe we could split this into two PRs? One that fixes the current documentation and one that adds `Tokenizer.explain`? Then I could already merge and publish the docs changes now.

adrianeboyd:
The docs changes will diverge slightly, since I modified one minor precedence thing in the tokenizer in order to be closer to the current docs, but it does make sense to split this up now so the docs aren't out-of-date for longer.

adrianeboyd:
This should be ready to go whenever you want to merge it. The "new" version in the docs may need to be updated. The `explain()` method and pseudo-code will need to be updated in v3.0 because of the new tokenizer special case handling with the Matcher.

Issue #4590, opened 2019-11-05
on_match functions with DependencyMatcher 


prilopes:
## How to reproduce the behaviour
I'm trying to create a method to use on_match in DependencyMatcher. I've done this successfully for Matcher and follow much of the same logic.

When I get the matches using something like this:
```python
matches = dep_matcher(doc)
for match in matches:
     print(match)
```
If I understand correctly `match` is a tuple with `match_id` (the rule that was triggered) and a list of nodes that match my pattern.

When defining an on_match method, I would expect `matches[i]` to return the same tuple, but that is not the case. Am I missing something??

I created the following script to help illustrate the problem

```python
from spacy.matcher import DependencyMatcher
from spacy import load


def my_on_match(matcher, doc, i, matches):
    print('inside my_on_match')
    match = matches[i]
    print(match)
    print('================')


nlp = load('en_core_web_sm')
doc = nlp('The quick brown fox jumped over the lazy dog.')

dep_matcher = DependencyMatcher(nlp.vocab)

dep_pattern = [
    {"SPEC": {"NODE_NAME": "jumped"}, "PATTERN": {"ORTH": "jumped"}},
    {"SPEC": {"NODE_NAME": "fox", "NBOR_RELOP": ">", "NBOR_NAME": "jumped"}, "PATTERN": {"ORTH": "fox"}},
    {"SPEC": {"NODE_NAME": "quick", "NBOR_RELOP": ".", "NBOR_NAME": "jumped"}, "PATTERN": {"ORTH": "fox"}},
]

dep_matcher.add('no_on_match', None, dep_pattern)
dep_matcher.add('with_on_match', my_on_match, dep_pattern)

matches = dep_matcher(doc)

print('all matches')
for match in matches:
     print(match)
```

## Your Environment

* **spaCy version:** 2.1.8
* **Platform:** Linux-4.15.0-66-generic-x86_64-with-Ubuntu-18.04-bionic
* **Python version:** 3.6.7

## I might have found the problem

I was reading the code for DependencyMatcher and found these lines on `dependencymatcher.pyx` (start at line 240 on master branch and line 237 for tag 2.1.8):

```python
    for i, (ent_id, nodes) in enumerate(matched_key_trees):
                on_match = self._callbacks.get(ent_id)
                if on_match is not None:
                    on_match(self, doc, i, matches)
```
So, if my line of thought is correct, on the last line of the excerpt (243 on master branch), should it be `matched_key_trees` instead of `matches`?

[...]

Issue #4433, opened 2019-10-11
DependencyMatcher documentation


fabio-reale:
I would like to use the DependencyMatcher, which I learned to exist reading [this issue](https://github.com/explosion/spaCy/issues/3275#issuecomment-471168539). There, I also learned there is no documentation for it.

I figured I would learn it from the code and testing, which I might end up doing, but I also figured I could try and ask for some help. I'm having some difficulty understanding what all of the operators are supposed to do (these ones for example: ">>", ".", "$+").

Once I get it, I fully intend to help with this documentation the best I can. So, any help or advice about either the DependencyMatcher or how to contribute to documentation are welcome
## Which page or section is this issue related to?

I assume the correct place for this documentation would be [rule based matching](https://github.com/explosion/spaCy/blob/master/website/docs/usage/rule-based-matching.md) page

svlandeg:
Hi @fabio-reale : it would be awesome to get some help to get this properly documented. I haven't used the DependencyMatcher myself yet either, but I'd be happy to dig through this code together with you.

As I understand it, the different operators refer to the kind of grammatical relations that can exist between tokens. E.g. `>`, referring to `gov`, refers to all the `children` of a node, while `>>`, referring to `gov_chain`, refers to the full `subtree` of a node. You can find the definitions of e.g. `subtree` in `token.pyx`, which is what `doc[node]` refers to (a `doc` is a list of tokens).

svlandeg:
I also saw that there's a `dependency_matcher` pytest fixture defined [here](https://github.com/explosion/spaCy/blob/master/spacy/tests/matcher/test_matcher_api.py#L287), which could be useful to look into as a first example. I agree with you that the patterns are a little hard to read those, with the various operators. Maybe there is a way to simplify those or make it more intuitive.

We should probably also look into the reason why `test_dependency_matcher` (in that same file) has been commented out. 

A little bit more background is here : https://github.com/explosion/spaCy/pull/2836 and https://github.com/explosion/spaCy/pull/3465

skrcode:
@fabio-reale You could have a look at these pull requests and go over the referenced links within them. 
1. https://github.com/explosion/spaCy/pull/2732
2. https://github.com/explosion/spaCy/pull/2836
3. https://github.com/explosion/spaCy/pull/3465
These would help clarify almost all of the questions that you may have with regard to the DependencyMatcher functionality. You could have a look at this https://nlp.stanford.edu/software/tregex.html to be more informed about the associated theory.

[...]

PR #4601, opened 2019-11-06
Bugfix/dep matcher issue 4590

prilopes:
## Description

Fixes bug #4590 
* adds test for issue
* updated on_match call in DependencyMatcher

### Types of change

bugfix

## Checklist


ines:
Thanks a lot – really appreciate the test! 👍

Btw, if you're using the dependency matcher and have some example patterns we could use for the documentation, that'd be super helpful. See #4433 for details. (The feature is currently still somewhat inofficial and undocumented, but we'd love to change that.)

PR #4612, opened 2019-11-08
Replace old gold alignment with new gold alignment

adrianeboyd:
## Description

Replace old gold alignment that allowed for some noise in the alignment between `raw` and `orth` with the new simpler alignment that requires that the `raw` and `orth` strings are identical except for whitespace and capitalization.

* Replace old alignment with new alignment
* Remove all quote normalizations
* Enable test
    * Modify test case for quote normalization
* Move relevant tests from `test_align.py` to `test_gold.py`

### Types of change

Enhancement?

## Checklist


honnibal:
Can we merge this onto master, or should we put it on develop?

If we wouldn't want to release this onto a patch release, we should probably put this on develop.

adrianeboyd:
I think you could go either way, but develop is probably the better choice. Should I rebase it?

adrianeboyd:
Let me start over with a new PR...

PR #4614, opened 2019-11-08
Adding support for Yoruba Language

Olamyy:
## Description

This PR introduces support for the Yoruba language. Current range of support includes:

1. Stopwords
2. Lexical Attributes Checkers i.e like_num
3. Tokenization using https://github.com/niger-Volta-LTI/iranlowo
3. Lemmatizer index

### Types of change
New Feature

## Checklist


adrianeboyd:
Thanks for your pull request! It would be great to add support for Yoruba!

A few small things to start:

* Instead of modifying `CONTRIBUTOR_AGREEMENT.md` directly, make a copy called `Olamyy.md` with your details filled in and add that renamed file in `.github/contributors`. Then undo your edits to the original template.

* Please restore the test files that were accidentally deleted.

* Do you have a source (even just a URL) for the stop words list?

Are you sure that `iranlowo` is the right tool for tokenization? I only looked at the most recent version, but it doesn't look like it includes a tokenizer, at least not one that's easy to use in this context. Taking a look at some Yoruba texts and the very small UD Yoruba corpus, my first guess would be that a tokenizer with prefix/suffix regexes similar to spacy's default English tokenizer might work fairly well, but that's really just a guess and I don't know any of the details. What kinds of special cases are there in Yoruba tokenization?

Spacy uses external libraries for languages like Chinese or Japanese where word segmentation is a very different task than for writing systems with mostly space-separated words, but for languages that use letters and spaces, we can usually develop prefix/suffix/etc. regexes to build a fast, accurate tokenizer with spacy's built-in tokenizer. We can provide some assistance in customizing the regexes for a new language, too.

Olamyy:
Hi @adrianeboyd,

- The contributor agreement changes has been effected.

- I can't seem to find the deleted tests in any of my commit history. Would you mind pointing me to it?

- Yes. The stopword list is sourced from https://raw.githubusercontent.com/dohliam/more-stoplists/master/yo/yo.txt . A comment about this was also included in the stop_words.py file.

- On Iranlowo : Iranlowo is a library we've been working on for a while now to provide extensive support for text preprocessing and analysis in yoruba. While you're right about the language not having any unusual word segmentation logic, I felt it'll be a good input to somehow link both works together. I have, however gone ahead to remove iranlowo as the default tokenization method. I hope that clears the confusion around it.

adrianeboyd:
>     * The contributor agreement changes has been effected.

That looks good, thanks!

>     * I can't seem to find the deleted tests in any of my commit history. Would you mind pointing me to it?

Click on [Files changed](https://github.com/explosion/spaCy/pull/4614/files) at the top of the page here to see all the edits in your PR. You can see that two files in `spacy/tests/lang` were deleted. (There's still an edit to `CONTRIBUTOR_AGREEMENT.md`, too.)

>     * Yes. The stopword list is sourced from https://raw.githubusercontent.com/dohliam/more-stoplists/master/yo/yo.txt . A comment about this was also included in the stop_words.py file.

Sorry I missed this! (We have so many languages without a reference, so it's hard to track down where things came from.)

>     * On Iranlowo : Iranlowo is a library we've been working on for a while now to provide extensive support for text preprocessing and analysis in yoruba. While you're right about the language not having any unusual word segmentation logic, I felt it'll be a good input to somehow link both works together. I have, however gone ahead to remove iranlowo as the default tokenization method. I hope that clears the confusion around it.

Could you update `yo/__init__.py` to remove the extra tokenizer settings? It looks like you started with `zh/__init__.py` as an example, but it has some settings that don't make sense for writing systems with an alphabet. A better example start from might be `ca/__init__.py`:

https://github.com/explosion/spaCy/blob/4d85f67eee1944379a9a2ccf883ce124fde3d04d/spacy/lang/ca/__init__.py#L1-L34

You can see how it's mostly using spacy's defaults, but with a few custom settings for lexical attributes, tokenizer exceptions, stop words, and the infix pattern. (There's nothing particularly special about Catalan as the example here, and many other languages would provide a good starting point, too.)

Olamyy:
Hi @adrianeboyd,
I've made the fixes you suggested. Everything should work well now.

ines:
Thanks! I just made a few small fixes to formatting and code organization 🙂

The only thing I had to take out was the lemmatization rules. Those should now live in the `spacy-lookups-data` repo: https://github.com/explosion/spacy-lookups-data

PR #4618, opened 2019-11-09
Set extensions when write_conllu() is called in UD train script

adrianeboyd:
## Description

`run_eval.py` uses the `write_conllu()` function from `ud_train.py` by itself, so it needs to set the token extensions if necessary.

### Types of change

Bugfix.

## Checklist

### DIFF
--- a/bin/ud/ud_train.py
+++ b/bin/ud/ud_train.py
@@ -221,12 +221,12 @@ def evaluate(nlp, text_loc, gold_loc, sys_loc, limit=None):
 
 
 def write_conllu(docs, file_):
-    try:
+    if not Token.has_extension("get_conllu_lines"):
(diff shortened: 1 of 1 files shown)


bin/ud/ud_train.py
```diff
@@ -221,6 +221,13 @@ def evaluate(nlp, text_loc, gold_loc, sys_loc, limit=None):
 
 
 def write_conllu(docs, file_):
+    try:
```
ines: We usually try to avoid `try`/`except` blocks, espeically in cases where we "know the answer". 

So maybe we could do something like this instead:

```python
if not Token.has_extension("get_conllu_lines"):
    Token.set_extension("get_conllu_lines", method=get_token_conllu)
# etc.
```

adrianeboyd: Sure!

PR #4621, opened 2019-11-10
Replace python-mecab3 with fugashi for Japanese

polm:
## Description

Currently Japanese uses python-mecab3 as a wrapper of MeCab for tokenization. However we use an old version of python-mecab3 because the latest versions make using Unidic (required by Universal Dependencies) tricky, in addition to having [an unresolved issue](https://github.com/SamuraiT/mecab-python3/issues/34) around memory usage. 

This change replaces python-mecab3 with [fugashi](https://github.com/polm/fugashi), a Cython wrapper for MeCab I wrote recently, which doesn't have the above issues. The API was also written with spaCy in mind, which allowed for simplifying the tokenizer code here a bit.

### Types of change

This is an enhancement.

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,7 @@ cuda100 =
     cupy-cuda100>=5.0.0b4
 # Language tokenizers with external dependencies
 ja =
-    mecab-python3==0.7
+    fugashi>=0.1.3
 ko =
     natto-py==0.9.0
 th =
--- a/spacy/lang/ja/__init__.py
+++ b/spacy/lang/ja/__init__.py
@@ -86,14 +86,14 @@ class JapaneseTokenizer(DummyTokenizer):
         dtokens, spaces = get_words_and_spaces(self.tokenizer, text)
         words = [x.surface for x in dtokens]
         doc = Doc(self.vocab, words=words, spaces=spaces)
-        mecab_tags = []
+        unidic_tags = []
         for token, dtoken in zip(doc, dtokens):
-            mecab_tags.append(dtoken.pos)
+            unidic_tags.append(dtoken.pos)
             token.tag_ = resolve_pos(dtoken)
 
             # if there's no lemma info (it's an unk) just use the surface
             token.lemma_ = dtoken.feature.lemma or dtoken.surface
-        doc.user_data["mecab_tags"] = mecab_tags
+        doc.user_data["unidic_tags"] = unidic_tags
         return doc
 
 
--- a/spacy/lang/ja/__init__.py
+++ b/spacy/lang/ja/__init__.py
@@ -12,21 +12,23 @@ from ...tokens import Doc
 from ...compat import copy_reg
 from ...util import DummyTokenizer
 
+# Handling for multiple spaces in a row is somewhat awkward, this simplifies
+# the flow by creating a dummy with the same interface.
+DummyNode = namedtuple("DummyNode", ["surface", "pos", "feature"])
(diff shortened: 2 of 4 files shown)


adrianeboyd:
I was looking at Japanese as an example for how to better set up the Chinese tokenizer and I noticed that the handling of whitespace is inconsistent with other languages. With the default tokenizer, consecutive whitespace tokens are merged into one token, so `I like cheese` with three spaces after `I` is `'I'`, `'  '`, `'like'`, `'cheese'` with two spaces as one token. (The spaces aren't getting displayed right here.)

Would your library make a difference for Korean? (And do you know if there's an easy way to set spacy up so that both Japanese and Korean work with mecab at the same time? The only way I could figure out was to set the environment variable `MECABRC` in `__init__.py` to point to the right configuration file for each, but hard-coded paths are obviously not a general solution. I mainly want to be able to run all the unit tests locally (especially since they're not run in the CI tests) and also do things like process all UD corpora in one batch.)

polm:
> I was looking at Japanese as an example for how to better set up the Chinese tokenizer and I noticed that the handling of whitespace is inconsistent with other languages. With the default tokenizer, consecutive whitespace tokens are merged into one token, so I like cheese with three spaces after I is 'I', ' ', 'like', 'cheese' with two spaces as one token. (The spaces aren't getting displayed right here.)

The way MeCab handles whitespace is weird and has caused issues before. I think we have a test and "I [three spaces] like cheese" should result in five tokens, two of which are spaces. Either way this PR shouldn't change the behavior of spaces, and if there's a problem or inconsistency it might be best to make a new issue.

> Would your library make a difference for Korean?

Based on a [benchmark I made](https://github.com/polm/ja-tokenizer-benchmark) when creating my library, the natto-py wrapper Korean uses is roughly four times slower than fugashi, presumably because it uses the dynamic cffi library for bindings. Currently fugashi makes some assumptions about the dictionary format but I would be glad to change that to support Korean if someone could give me a guide to the format (I don't speak Korean).

> And do you know if there's an easy way to set spacy up so that both Japanese and Korean work with mecab at the same time?

I don't think there's a very easy way... using the `MECABRC` env var is probably one of the better options. Another option would be allowing users to specify a dictionary path when initializing the language. That's desirable in any case for use with custom tokenizer dictionaries, which is pretty common.

Another option that's more work is to distribute a MeCab wrapper with a built-in dictionary. `mecab-python3` did some work on this but it's using the obsolete IPADic dictionary.

adrianeboyd:
Thanks for the info! I was hoping there might be a nicer method I didn't know about, but no such luck.

Since the CI testing doesn't help much here, I ran the unittests and the UD eval scripts and didn't run into any issues.

setup.cfg
```diff
@@ -73,7 +73,7 @@ cuda100 =
     cupy-cuda100>=5.0.0b4
 # Language tokenizers with external dependencies
 ja =
-    mecab-python3==0.7
+    fugashi>=0.1.3
```
honnibal: We always want to have closed ranges, or eventually this version of the library would be uninstallable:
```suggestion
    fugashi>=0.1.3,<0.2.0
```

honnibal: I'll fix this post-merge, I don't want to block the PR.

honnibal:
Thanks! Sorry it took a bit of time to get to this.

setup.cfg
polm: Thanks, and understood, I'll keep that in mind next time.

PR #4624, opened 2019-11-11
Fix Example details for train CLI / pipeline components

adrianeboyd:
## Description

Various fixes to get the new `Example` API working with the train CLI and for all pipeline components. Main changes:

* Training examples are projectivized correctly (including a new test)
* Head of `0` is a valid value, not treated as unset
* `pipe()` in all components is updated to accept `as_examples` argument and to yield the annotated examples without clobbering the provided data
* Use `Example.goldparse` in `iter_gold_docs()` instead of `Example.gold` because a `None` `GoldParse` is generated with `ignore_misaligned` and generating it on-the-fly can raise an unwanted `AlignmentError`

### Types of change

Bugfix.

## Checklist

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -341,7 +341,7 @@ def train(
         iter_since_best = 0
         best_score = 0.0
         for i in range(n_iter):
-            train_data = corpus.train_data(
+            train_data = corpus.train_dataset(
                 nlp,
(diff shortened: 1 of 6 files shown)


adrianeboyd:
(I moved a patch from #4632 to this PR because it should have been in this PR, but it didn't go as cleanly as I hoped.)

PR #4632, opened 2019-11-12
Restructure Example with merged sents as default

adrianeboyd:
## Description

An `Example` now includes a single `TokenAnnotation` that includes all  the information from one `Doc` (=JSON `paragraph`). If required, the individual sentences can be returned as a list of examples with `Example.split_sents()` with no raw text available.

* Input/output a single `Example.token_annotation`

* Add `sent_starts` to `TokenAnnotation` to handle sentence boundaries

* Replace `Example.merge_sents()` with `Example.split_sents()`

* Modify components to use a single `Example.token_annotation`
  * Pipeline components
  * conllu2json converter

* Rework/rename `Example.add_token_annotation()` and `Example.add_doc_annotation()` to `Example.set_token_annotation()` and `Example.set_doc_annotation()` as functions that set rather then appending/extending.

* Rename `morphology` to `morphs` in `TokenAnnotation` and `GoldParse`

* Add getters to `TokenAnnotation` to supply default values when a given attribute is not available

* `Example.get_gold_parses()` in `spacy.gold._make_golds()` is only applied on single examples, so the `GoldParse` is returned saved in the provided `Example` rather than creating a new `Example` with no other internal annotation

* Update tests for API changes and `merge_sents()` -> `split_sents()`

### Types of change

Enhancement.

## Checklist


adrianeboyd:
This PR is based on #4624, which should be merged first.

adrianeboyd:
(And after redoing this to have a single `TokenAnnotation` per `Example` there may not be as much of a strong argument for maintaining the `TokenAnnotation` and `DocAnnotation` classes as opposed to putting everything at the top level under `Example`. As long as you can easily add new attributes, I don't think it makes a big difference.)

spacy/gold.pyx
```diff
@@ -673,9 +694,58 @@ cdef class TokenAnnotation:
...
+            return self.morphs[i]
+        except IndexError:
+            return {}
+
+    def get_sent_start(self, i):
+        try:
+            return self.sent_starts[i]
+        except IndexError:
+            return None
+
```
honnibal: Many people in Python argue for this use of try/except, but personally I strongly disprefer it. i think try/except shouldn't be used for control flow, it should only be used for error handling. So I'd like to see these replaced by conditionals, probably just using the inline ternary, e.g.

```
return self.sent_starts[i] if i < len(self.sent_starts) else None
```

spacy/gold.pyx
```diff
@@ -691,35 +761,33 @@ cdef class DocAnnotation:
 
 
 cdef class Example:
-    def __init__(self, doc_annotation=None, token_annotations=None, doc=None,
-                 make_projective=False, ignore_misaligned=False, goldparse=None):
+    def __init__(self, doc_annotation=None, token_annotation=None, doc=None,
+                 goldparse=None):
```
honnibal: Naming here is tricky. These names are a bit long, because we'll want to be calling the constructor with keyword arguments. In fact I'd probably want to make the function keyword only. I also don't know whether it should be `doc_annotation` or `doc_annotations`. `goldparse` should definitely be `gold_parse`.

I'll accept `doc_annotation` and `token_annotation` if we can't think of anything better --- I can't right now. But I do wish we had something shorter.

adrianeboyd: As I said above, I kinda think we could drop both and just have `Example.words` and `Example.cats`, etc. They are a bit too long and it's not quite as useful to have them as when each sentence was a separate unit by default. I think this change should be in a different PR, though, to just focus on the split/merge distinction here.

honnibal: Fair enough, I think that's probably reasonable. But yeah let's get that merged.

PR #4648, opened 2019-11-14
Fix basic language support for Luxembourgish (by adding punctuation.py)

questoph:
## Description

- Added "punctuation.py" to the language package for Luxembourgish. 
- Changes to "__init__.py" and "tokenizer_exceptions.py" to match the new setup.

With these changes it is now possible to split tokens like "d'Kanner" correctly into 2 tokens [d'] and [kanner]. The "punctuation.py" file is a replication of the version for Italian.

Changed the long text in "test_text.py" because all tests were failing. Had to keep the text in one line to successfully pass the tests.

### Types of change
Fix for the basic language support for Luxembourgish

## Checklist

### DIFF
--- a/.github/contributors/questoph.md
+++ b/.github/contributors/questoph.md
@@ -87,7 +87,7 @@ U.S. Federal law. Any choice of law rules will not apply.
 7. Please place an “x” on one of the applicable statement below. Please do NOT
 mark both statements:
 
-    * [ ] I am signing on behalf of myself as an individual and no other person
+    * [x] I am signing on behalf of myself as an individual and no other person
     or entity, including my employer, has or will have rights with respect to my
     contributions.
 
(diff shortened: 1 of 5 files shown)


ines:
Thanks!

> With these changes it is now possible to split tokens like "d'Kanner" correctly into 2 tokens [d'] and [kanner].

It'd be cool to have some tests for these cases – they're pretty specific and we want to make sure that changes in the future still preserve them correctly. So if you can think of a few useful test cases and want to submit a separate PR, that'd be great.

questoph:
great, thanks! we will definitely provide more resources for lb including tests, also because there are other cases, e.g., lemmatized multi-word tokens or outdated orthographic variants, where it would be good to not split the word into two tokens. i just wanted to get the basis straight first.

PR #4664, opened 2019-11-17
Modified Scorer for custom textcat pipe

tamuhey:
I changed the textcat part of `Scorer.__init__` so that users can use this with their custom textcat pipe.
For example:
```python
@spacy.component("my_textcat")
def my_textcat(doc):
    ...

scorer = score.Scorer()
scorer.set_textcat_scorer(nlp.pipeline, name_textcat_pipe="my_textcat")
...
```

### DIFF
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -91,20 +91,23 @@ class Scorer(object):
         self.textcat_per_cat = dict()
         self.textcat_positive_label = None
         self.textcat_multilabel = False
+        self.set_textcat_scorer(pipeline)
 
-        if pipeline:
-            for name, model in pipeline:
-                if name == "textcat":
-                    self.textcat_positive_label = model.cfg.get("positive_label", None)
-                    if self.textcat_positive_label:
-                        self.textcat = PRFScore()
-                    if not model.cfg.get("exclusive_classes", False):
-                        self.textcat_multilabel = True
-                        for label in model.cfg.get("labels", []):
-                            self.textcat_per_cat[label] = ROCAUCScore()
-                    else:
-                        for label in model.cfg.get("labels", []):
-                            self.textcat_per_cat[label] = PRFScore()
+    def set_textcat_scorer(self, pipeline, name_textcat_pipe="textcat"):
+        if not pipeline:
+            return
+        for name, model in pipeline:
+            if name == name_textcat_pipe:
+                self.textcat_positive_label = model.cfg.get("positive_label", None)
+                if self.textcat_positive_label:
+                    self.textcat = PRFScore()
+                if not model.cfg.get("exclusive_classes", False):
+                    self.textcat_multilabel = True
+                    for label in model.cfg.get("labels", []):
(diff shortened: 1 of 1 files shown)


ines:
Thanks, I like this idea! 👍

I wonder if we could generalise this a bit more and maybe allow pipeline components to define how they want to be "scored", instead of introducing more keyword arguments.

One idea could be:

1. Use function registries (see [`catalogue`](https://github.com/explosion/catalogue)) to define the different scorers and assign them names (`textcat`, `ner` etc.).
2. Let the `@component` decorator define `scorer="textcat"` etc.

```python
# Simple usage with built-in scorer
@spacy.component("my_textcat", scorer="textcat")
def my_textcat(doc):
    ...

# More complex usage with fully custom scorer
@spacy.component("my_textcat", scorer="my_custom_textcat")
def my_textcat(doc):
    ...

@spacy.registry.scorers.register("my_custom_textcat")
def my_custom_textcat_scorer(doc, gold):
    return {"some_custom_metric": get_some_custom_score()}
```

(The function registries are currently mostly internals, but we do want to transition more parts of the library to use them and also expose this to the users. I think it's a great way to easily customise functionality without having to pass around functions or introduce too many settings.)

Alternatively, another idea:

1. Try and be clever and check what the component `assigns`. 
2. If the component assigns to `doc.cats`, we evaluate it with the textcat scorer, and so on.

This does make sense for most use cases but maybe we're beeing *too* clever here and making too many assumptions. It also doesn't address the custom scoring, e.g. if your component assigns to `doc._.custom_cats`.

cc: @adrianeboyd @honnibal What do you think?

adrianeboyd:
I don't think that we want to be clever and guess what kind of evaluation to do.

As a recent example that should also be easier: I wanted to add an additional tagger evaluation to evaluate just the `tag` part of `tag__morph` tags and it involved editing `Scorer` and also these parts of the train CLI below, and it'd be nice to be able to do this in a more general/modular way.

Basically, this kind of function from the train CLI could be implemented more generally:

https://github.com/explosion/spaCy/blob/5ad5c4b44a95a9264a3018c5f73a23639451db9a/spacy/cli/train.py#L587-L594

And it's clunky that you have to sync this in two spots in the train CLI (to be clear, I like it better than the hard-coded version with all the metrics from before, but I am responsible for some of this clunkiness):

https://github.com/explosion/spaCy/blob/5ad5c4b44a95a9264a3018c5f73a23639451db9a/spacy/cli/train.py#L597-L614

I think it would help if the `Scorer` were more modular rather than having this huge jumbled function:

https://github.com/explosion/spaCy/blob/5ad5c4b44a95a9264a3018c5f73a23639451db9a/spacy/scorer.py#L220-L343

I'd have to think about how to define/evaluate everything efficiently for lots of components...

honnibal:
I agree that there are a number of ways the scorer could be improved, and we expect to make some of these improvements for spaCy 3. But I think we should consider the redesign carefully, so I don't want to merge this PR, as it makes the current code more complex.

PR #4670, opened 2019-11-19
Bugfix/fix entity ruler from disk

GuiGel:
Solve a bug in the EntityRuler from_disk method

##### issue4651
The PhraseMatcher isn't initialized correctly in the EntityRuler from_disk method when the phrase_matcher_attr attribute is specified. 
This is because the from_disk function is used before the recovery of the phrase_matcher_attr attribute.
A way to solve this bug is to split the deserializer in a 2 keys dict, the patterns and the cfg. By this way we can first deserialize the cfg and initialize the PhraseMatcher with it, and secondly use the patterns part in the from_disk function, in order to add the patterns the PhraseMatcher that have been correctly initialized.

##### My environnement:
spaCy version: 2.2.2
Platform: Linux-4.9.0-11-amd64-x86_64-with-debian-9.11
Python version: 3.7.3

##### Tests run
test_issue4651_with_phrase_matcher_attr --> test that failed without modification of the EntityRuler from_disk method and work after.
test_issue4651_without_phrase_matcher_attr --> test that work all the time. 

### Types of change
bug fix

## Checklist


adrianeboyd:
Hi, thanks for submitting this PR! Your changes and tests look good, the main problems look like a few quirky things related to temporary files in the CI. (I ran into the same problems in my first PR with tests involving temporary files, it's not obvious!)

Try serializing the `EntityRuler` by itself (just to minimize the functionality that you're testing) and handling the temporary directories for `to_disk()` like this:

```
from ..util import make_tempdir

...
    nlp_reloaded = English()
    with make_tempdir() as d:
        file_path = d / "entityruler"
        ruler.to_disk(file_path)
        ruler_reloaded = EntityRuler(nlp_reloaded).from_disk(file_path)

    nlp_reloaded.add_pipe(ruler_reloaded)
    loaded_doc = nlp_reloaded(text)
```

You can see some examples in other tests like this one:

https://github.com/explosion/spaCy/blob/5ad5c4b44a95a9264a3018c5f73a23639451db9a/spacy/tests/serialize/test_serialize_pipeline.py#L48-L60

ines:
Thanks a lot! 👍 Also really appreciate the extensive tests.

GuiGel:
Happy to contribute a little to your great library and thanks for you help!

PR #4671, opened 2019-11-19
Add option for improved NER feature extraction

honnibal:
The parser and NER components use transition-based models, that build a state representation by combining the vectors from some tokens in the context. The previous NER feature set was:

* Current word
* Next word
* Previous word
* First word of the current entity
* Word before first word of the current entity
* Word after first word of the current entity

This PR adds a feature set that seems to perform a bit better, while using fewer tokens:

* Current word
* First word of current entity, if it exists
* Last word of current entity, if it exists

This is less redundant, and runs faster. It also prepares us better for the wider token vectors we'll be expecting from transformer models.

You can change the feature set by passing `nr_token_features` to the `begin_training` method. Care should be taken with this, as not all numbers are valid --- they're basically shorthand for particular feature sets. There's also currently no error handling for invalid values, so this is basically internals until we have a better config system in place.

```
nlp.begin_training(component_cfg={"ner": {"nr_feature_tokens": 3}})
```

## Checklist

### DIFF
--- a/spacy/syntax/_state.pxd
+++ b/spacy/syntax/_state.pxd
@@ -100,10 +100,30 @@ cdef cppclass StateC:
         free(this.shifted - PADDING)
 
     void set_context_tokens(int* ids, int n) nogil:
-        if n == 2:
+        if n == 1:
+            if this.B(0) >= 0:
+                ids[0] = this.B(0)
+            else:
+                ids[0] = -1
+        elif n == 2:
             ids[0] = this.B(0)
             ids[1] = this.S(0)
-        if n == 8:
+        elif n == 3:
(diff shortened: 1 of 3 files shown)


alejandrojcastaneira:
Hello, I would like to ask something, I think it's a good idea to have less redundant features, it should be possible to define custom features? It could be the case that there it's one entity (type or name) in the document that if it's present, it could (but not necessary) modify the behavior of other entities eg: If we extracted as company IKEA at the beginning of the paragraph we are more likely to extract furniture_products in the rest of the document. It thought should be the right moment to comment on this, as usual grateful and impressed with all your releases.

PR #4676, opened 2019-11-20
Add a tagger-based sentence segmenter

adrianeboyd:
## Description

Add a `SentTagger` component that segments sentences by tagging tokens with two tags: beginning of a sentence or not.

* Add `sent_starts` to `GoldParse`
* Add sentence segmentation evaluation to `Scorer` (also works for the parser and other components)
* Add evaluation to CLI evaluate
* Add metrics/scoring to train CLI

This component has higher precision and lower recall than the parser, but with these model settings the f-score is only 0.5% less than the parser and it's ~10x faster. It could be 10-20% faster and ~1% worse with even more minimal model settings. With these settings it's 275K on disk.

| Model | P | R | F | Words/Sec |
|---|---|---|---|---|
| en_core_web_sm parser | 93.86 | 95.10 | 94.48 | 5656 |
| senttagger | 97.63 | 90.65 | 94.01 | 66380 |

To do:

- [ ] Needs a better name / differentiation from `Sentencizer`
- [ ] Pipeline-related tests
- [ ] Docs
- [ ] Need to figure how to manage loading this component vs. a parser in the same model

## Checklist


adrianeboyd:
This is based on #4632, which should be merged first.

honnibal:
Nice! This will be really valuable, especially if we want to use more self-attention, which requires sentences as a preprocess. I'm sure we can improve the accuracy too --- we should make some Prodigy recipes, and make some annotated data for other genres.

I think for spaCy v3 we should rename the current `Sentencizer` to `SentenceRuler`, and then this can become `SentenceRecognizer`. Unfortunately this involves deprecating the arguably better name, `Sentencizer` -- but repurposing the same name for a new component is bad. We could consider reviving the name in a future version, but for the initial change I think deprecating it is best.

adrianeboyd:
`SentenceRuler` sounds like a good rename for `Sentencizer`, although I might go as far as something like `PunctSentenceRuler` or `PunctSentencizer` to really emphasize that it's not doing much analysis.

For this, though, what about `SentenceMarker` or something else that wouldn't get people confused with tagging/NER? `SentenceTagger` is maybe a bit nerdview-y and it sounds like it might be a thing that does POS tagging for a sentence. And `SentenceBoundaryTagger` is a bit too long.  (`SentTagger` was just intended to be a temporary / internally descriptive name.)

I was also thinking that maybe it would simplest if it just became part of the default pipeline. I'd have to compare to be sure, but since the parser seems to primarily improve the recall, I think it might basically be okay to have both in the pipeline and just allow the parser to add a few more boundaries (if it adds boundaries at all), and it's not like marking the sentence boundaries is the time-consuming part of parsing.

honnibal:
I think I'd prefer `SentenceRuler` to `PunctSentenceRuler`, as I could imagine non-punctuation rules such as whitespace. I think the name `Ruler` makes it clear enough, and makes a pair with `EntityRuler`.

I see your point about `SentenceTagger`. I think `SentenceRecognizer` is probably a fine choice then --- it's a good contrastive pair with `EntityRuler` and `EntityRecognizer`.

honnibal:
@adrianeboyd Could you just double-check my merge, https://github.com/explosion/spaCy/pull/4676/commits/a0fff5fa0463f95741f3e8848f9a45b7009ae0d8 ?

adrianeboyd:
I like the parallel to `EntityRuler/Recognizer` so I think `SentenceRecognizer` sounds good!

It might be easier if I just redid this PR, too. (I should have thought more about the squash merging before setting things up like this.)

PR #4679, opened 2019-11-20
Update to Holmes Universe entry

richardpaulhudson:
Hi there, I released Holmes v2.1.0 yesterday and would like to update the Universe entry. Thanks and best wishes from Munich, Richard


ines:
Thanks for the update – and I love the new demo! 💯

Issue #4673, opened 2019-11-19
vector_norm throws error for unusual text in sentence with more than one word.


mmaybeno:
## How to reproduce the behaviour
I found this bug that I'm not sure if it resides in `spacy` or `cupy`. It only appears on GPU instances and when you try to get vectors from a multiple word document containing non standard words. Any help tracking it down with a potential fix would be fantastic.

```
import en_core_web_md
import spacy
spacy.prefer_gpu()
nlp = en_core_web_md.load()

doc = nlp("somerandomword")
doc.vector_norm
# works

doc = nlp("somerandomword.")
doc.vector_norm
# throws type error

doc = nlp("The somerandomword")
doc.vector_norm
# throws type error
```

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-96-d19c7b8f8943> in <module>()
----> 1 doc.vector_norm
doc.pyx in spacy.tokens.doc.Doc.vector_norm.__get__()
doc.pyx in __iter__()
cupy/core/core.pyx in cupy.core.core.ndarray.__add__()
cupy/core/_kernel.pyx in cupy.core._kernel.ufunc.__call__()
cupy/core/_kernel.pyx in cupy.core._kernel._preprocess_args()
TypeError: Unsupported type <class 'numpy.ndarray'>
```

## Your Environment

* Operating System: Ubuntu 18.04.3
* Python Version Used: 3.6.8
* spaCy Version Used: 2.2.2
* Environment Information: Running on Google Colab but also experienced it on other GPU instances.

```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 430.50       Driver Version: 418.67       CUDA Version: 10.1     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla K80           Off  | 00000000:00:04.0 Off |                    0 |
| N/A   34C    P8    26W / 149W |      0MiB / 11441MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|  No running processes found                                                 |
[...]

PR #4680, opened 2019-11-20
Agnostic vocab array fix

mmaybeno:
Fixes a bug where we were not getting the correct array module when returning a vector in Vocab.

## Description

Fixes issue related to #4673. This issue causes a `TypeError` when trying to do any calculations related to the vectors where a word is not in the Vocab list. This leads to mixed array types (in our case `numpy` and `cupy` arrays). 
### Types of change

The change is doing what is done in other parts of the code where we first get the appropriate array module prior to any calculations. Removing `numpy` with `get_array_module`.

## Checklist

### DIFF
--- a/.github/contributors/mmaybeno.md
+++ b/.github/contributors/mmaybeno.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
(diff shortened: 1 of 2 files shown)


mmaybeno:
I would like to create a regression but would need some insight on how to do it with with `cupy`. Since it's a slight edge case, not sure how to force the vectors into a different type other than `numpy.ndarray`. I was not able to find any other tests that used a similar pattern.

mmaybeno:
Building a new spacy wheel and testing this patch on colab does indeed fix the `TypeError`

rjurney:
I really need this patch, when will it be merged?

mmaybeno:
Same. Not sure if a test needs to be created in order to merge this @honnibal? I can try and work on it more today. Otherwise, you could always pull from my fork'd branch and build the wheel locally (which is what I did).

rjurney:
@mmaybeno I can't figure out how to build the package with cuda100 support. How do I do that?

mmaybeno:
`docker run -it --rm --entrypoint bash nvidia/cuda:10.0-runtime`

#### While in the container run:
```
apt-get update \
&& apt-get install -y --no-install-recommends gcc g++ build-essential ca-certificates python3 python3-pip python3-dev python3-venv git \
&& pip3 install wheel \
&& pip3 install cupy-cuda100 \
&& git clone https://github.com/mmaybeno/spaCy.git --branch agnostic-vocab-array-fix \
&& cd spaCy \
&& make
```

#### Check the dist for the wheel. You can pip install it like normal
`pip install dist/spacy-2.2.2-cp36-cp36m-linux_x86_64.whl`

rjurney:
@mmaybeno Wow, that's more complicated than I would have thought :) Thanks!

mmaybeno:
If you want to keep it as a wheel and install it in a container. That is what I did at least :). I'm sure you can do it another way.

rjurney:
@mmaybeno Your branch works fine for me. I now have `spaCy[cuda100]` working with `spacy.prefer_gpu()`. Thanks!

honnibal:
Thanks!

If you need patches urgently, you can build yourself a dev version by checking out the repo. We do try to get simple fixes like this released quickly though.

The testing on GPU is a persistent problem, which is how these errors creep in. Unfortunately we don't have CI support for GPU.

kadu9:
Can you please specify instructions on how to use this patch

mmaybeno:
@kadu9 look at the docker build I posted above. Now that things are merged into master though you can target the `spacy` main branch. This will only be an issue till they release the new version with the patch.

PR #4685, opened 2019-11-20
Restructure Sentencizer to follow Pipe API

adrianeboyd:
## Description

Restructure Sentencizer to follow Pipe API so that it can be scored with `nlp.evaluate()`.

### Types of change

Enhancement?

## Checklist

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -341,7 +341,7 @@ def train(
         iter_since_best = 0
         best_score = 0.0
         for i in range(n_iter):
-            train_data = corpus.train_data(
+            train_data = corpus.train_dataset(
                 nlp,
                 noise_level=noise_level,
                 orth_variant_level=orth_variant_level,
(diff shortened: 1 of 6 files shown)


adrianeboyd:
This PR is based on #4632, which should be merged first.

adrianeboyd:
This should probably be done against master instead (I just did it this way so I could run direct evaluations with the proposed `SentTagger`) and after looking at the parser more, I also think that it shouldn't set `sent_start` to `-1` anywhere, just like the parser, but that should be a separate issue/PR, too. (Or is there a reason it sets all tokens? Is there a reason to want to prevent later components from adding more boundaries?)

honnibal:
@adrianeboyd Setting 1 vs -1 is a tricky question. I could imagine not wanting the parser to split sentences the sentencizer hasn't split. It should probably be a setting?

Merging this onto master seems like a fine choice.

adrianeboyd:
It could be a setting, or maybe having an extra pipeline component would be more flexible? The other components only set `1` and you add something like `finalize_sentences` at a particular point if you want the rest to become `-1`?

adrianeboyd:
Replaced by #4721.

PR #4691, opened 2019-11-21
Make vectors.find() return keys in correct order

honnibal:
`vectors.find()` accepts several input types. One of them is a sequence of rows, which are mapped into a sequence of corresponding keys. This function returned the keys out of order.


spacy/vectors.pyx
```diff
@@ -265,17 +265,13 @@ cdef class Vectors:
             rows = [self.key2row.get(key, -1.) for key in keys]
             return xp.asarray(rows, dtype="i")
         else:
+            row2key = {row: key for key, row in self.key2row.items()}
             targets = set()
```
honnibal:
```suggestion

```

Issue #3637, opened 2019-04-25
Add is_sent_end to token attributes


michelleful:
## Feature description
Tokens already have an attribute called `is_sent_start` which can be used in the Matcher. Having an equivalent `is_sent_end` would be useful, so that we can say "only match this token if it's at the end of a sentence".

## Could the feature be a [custom component](https://spacy.io/usage/processing-pipelines#custom-components) or [spaCy plugin](https://spacy.io/universe)?

It seems more at home in the SpaCy library itself.

kutoxinyi:
Any update?

honnibal:
I think I like this idea. Some implementation notes:

Several files need to be changed to support this. We need the computed property in the `spacy/tokens/token.pyx` file, and we also need the attribute name in the `spacy.attrs` file. Then we need to support the attribute at the Cython level, in the `spacy/tokens/token.pxd` file's `Token.get_struct_attr` method. And of course, we need to test the feature and add it to the docs.

The `is_sent_end` attribute should not be a writeable property. Instead, we'll want to raise an error that points people towards the right way to specify the sentence boundaries (`is_sent_start`).

When implementing the `token.is_sent_end` property, care should be taken to account for the `None` case. `is_sent_start` is ternary valued, to support cases where the sentence boundaries are unspecified. If the next token's `is_sent_start` value is `None`, we want `is_sent_end` to be `None` as well.

The `is_sent_end` value for the final token in the `Doc` should always be `True`.

igor-susic:
@honnibal If this is available and no one is working on it, I would give it a try :)

jackmen:
@igor-susic 
Any update on this issue? Would be interested in this feature. Thanks!

MiniLau:
I am interested in this feature, is anyone working on this ?

adrianeboyd:
No, I don't think anyone is working on this right now. (If you'd like to work on it, you are welcome to submit a PR. If you have questions or need feedback along the way, it's fine to submit a PR that's still in progress, just mark it as a WIP at first.)

polm:
It looks like code for this was merged in May but the PR didn't get linked here.

https://github.com/explosion/spaCy/pull/5375

svlandeg:
Ah, thanks @polm!

PR #4697, opened 2019-11-22
Add is_sent_end to token attributes

MiniLau:
## Description

Added #3637.
I added some basic tests in the tokenizer and the senten 

### Types of change

Added a new feature.
Added test for Tokenizer and Sentencizer to test the new attribute

## Checklist


spacy/symbols.pxd
```diff
@@ -83,6 +83,7 @@ cdef enum symbol_t:
     ENT_TYPE
     HEAD
     SENT_START
+    SENT_END
```
honnibal: Unfortunately the order of the symbols and attributes matters -- it influences which ID will be assigned to that attribute. So we need new attributes to go to the end, and we can only reorder them when we're ready to break model compatibility.

If we insert the `SENT_END` symbol here, the subsequent symbols will be mapped to different values, which could break models.

spacy/structs.pxd
```diff
@@ -75,6 +75,7 @@ cdef struct TokenC:
     uint32_t r_edge
 
     int sent_start
+    int sent_end
```
honnibal: We shouldn't add this to the struct, that's a waste of space and it's data duplication that's hard to keep in-sync with the `sent_start` attribute. It's okay to have `sent_end` as an attribute, but it has to be computed dynamically by checking whether the following token is `sent_start`.

spacy/attrs.pxd
```diff
@@ -86,6 +86,7 @@ cdef enum attr_id_t:
     ENT_TYPE
     HEAD
     SENT_START
+    SENT_END
```
honnibal: We need to place `SENT_END` at the end of the list for now -- see the comment on the `symbols.pyx`.

MiniLau:
Thank you for the implementation approval. So, to finalize the pull request I should just edit the 3 files the way you explained above ?

honnibal:
The changes would need to be a bit bigger than those three files. You'll also need to change the implementation of the `sent_end` property in `token.pyx`, because you'll need to compute it based on the following token's `sent_start`.

MiniLau:
I don't understand, I thought the implementation of the property was right. Why should I change it ? I can't use "sent_start" property in the implementation in token.pyx ?

honnibal:
@MiniLau My mistake, sorry --- yes the property's implementation is correct.

MiniLau:
Is is okay with those modifications ?

adrianeboyd:
Sorry, I think this got overlooked because it looked like there will still outstanding changes requested. Let me see if I can resolve the conflicts and let the CI run the tests again.

adrianeboyd:
@honnibal: Can you have another look at this?

There are still some minor changes to make to the `Matcher` to make `IS_SENT_END` work like `IS_SENT_START`, but I could do them in a separate PR.

adrianeboyd:
Since the original repo is no longer available, most of this PR is reconstructed in #5375.

PR #4738, opened 2019-11-30
Improvements to the Finnish language data

aajanki:
# Description

I enabled the like_num attribute on Finnish and improved the tokenization rules (using the same tokenizer infixes and suffixes as Danish). Added tests for these.

### Types of change

Enhancement

## Checklist


ines:
Thanks a lot! 👍

spacy/lang/fi/__init__.py
```diff
@@ -3,20 +3,26 @@
...
 from .stop_words import STOP_WORDS
+from .lex_attrs import LEX_ATTRS
 
 from ..tokenizer_exceptions import BASE_EXCEPTIONS
 from ..norm_exceptions import BASE_NORMS
 from ...language import Language
 from ...attrs import LANG, NORM
 from ...util import update_exc, add_lookups
 
+from ..da.punctuation import TOKENIZER_INFIXES, TOKENIZER_SUFFIXES
```
ines: I think it might be better to copy-paste the rules over instead of importing them here. It duplicates code, but otherwise there's a risk that arbitrary and very specific changes to Danish have unintended side-effects for Finnish.

aajanki: Makes sense! I created fi/punctuation.py file with tokenizer rules for Finnish.

PR #4767, opened 2019-12-04
Update token.md

Tclack88:
documentation is confusing: A '?' is a right punct, but '¿' is a left punct

## Checklist


adrianeboyd:
The '?' is part of the question in the text, not intended to an example of a left punct (compare to the right punct example with close parenthesis in monospace font with a gray background), but I can see how the gray on gray makes it hard to distinguish. Perhaps it could be rephrased instead?

```
Is the token a right punctuation mark, e.g. ")" ?
```

Then it looks more like the `like_num` explanation.

Tclack88:
Ah, this makes sense. I can change the pull request to reflect that

ines:
Thanks! I also moved the quotes into the code block as well to make it consistent with how we usually format strings in code.

PR #4783, opened 2019-12-08
Improve Arabic language class

abal2051:
*tag_map.py*: I added a tag set corresponding to the Prague Arabic Treebank. The tag set might look dreadfully huge but that's unfortunately the cost of Arabic being a highly inflectional language.

*tokenizer_exceptions.py*: I changed the *lemma* designation to *norm* since none of the expansions for the given orthographic units are lemmas. The "lemmas" were either inflected forms,  or expand like so:
```python
{NORM: "قبل الميلاد", ORTH: ".ق.م"}
```
In this example, the orthographic unit is "b.c." and the expansion is (obviously) "before christ". So it makes much more sense for them to be considered the norms. I also removed the following two exceptions:
https://github.com/explosion/spaCy/blob/c208eb6e4d26ae874a84dd4485bd809599748552/spacy/lang/ar/tokenizer_exceptions.py#L29-L30
These were really strange to find, because the first orthographic unit "ثنى" means "he bent [something]" but the lemma paired with it strangely means "tell me", and the second orthographic unit "أنا" means the singular pronoun *I*, but the lemma is "tell us [a prediction]". It looks like based on the following comment
https://github.com/explosion/spaCy/blob/c208eb6e4d26ae874a84dd4485bd809599748552/spacy/lang/ar/tokenizer_exceptions.py#L20
that the pairings make sense in the context of Arabic academic literature. However, even if the tokenizer isn't required to be as *domain agnostic* as possible, the fact remains that the above tokenizer exceptions overload very common words with incorrect lemmas. So I removed them.

### Types of change
Enhancement

## Checklist


adrianeboyd:
Thanks for your contribution!

A lot of other languages specify the lemmas of abbreviations or contractions as the full forms, so that may be why some of this looks like it does. Inflected lemmas definitely don't make sense and should be corrected, but it might make sense to keep the expanded lemmas for abbreviations. I'm not sure there's a "right" answer for how to define exceptions for all cases/languages, but it's not bad to keep some consistency across languages if we can.

A note on tag maps:

Since tag maps are specific to the treebank / training data, I don't think it makes sense to maintain tag maps in the source, except possibly for the ones that correspond to the models provided by spacy. It's still just a PR, but I would like to add a `tag-map` option to the train CLI so that you can provide a tag map while training without modifying the source code. I think it makes more sense to the store the tag maps as part of the training data, not as part of spacy. When you train a new model, the tag map is serialized with the vocab so you don't need it to be in spacy itself.

adrianeboyd:
Hi, sorry we didn't get back to this sooner! The release for v3 is coming up soon and I'm trying to get this ready to merge. If modifying your `master` branch is causing problems, my apologies, you can close this PR and open a new one from a branch, or I can redo this PR myself if you'd prefer.

abal2051:
Sorry for the hold up, If it's okay with you I'd appreciate it if you could do it. You can close this PR. Looking forward to the next release.

PR #4799, opened 2019-12-12
Fix alignment of character entity spans to words in GoldParse

### LINKED ISSUE #4791: Getting Error: gold.pyx in spacy.gold.GoldParse.__init__()  IndexError: list index out of range
Hello,
I'm training spacy english blank model with entities, tags, and parser in the normal training data format(not json).

Following is the code that I'm using for training.
```
for i in tqdm(range(300)): 
    random.shuffle(Train_data) 
    losses = {} 
#     for text, annotations in Train_data:
#         nlp.update([text], [{'words': annotations['words'], 'entities': annotations['entities'], 'heads': annotations['heads'], 'deps': annotations['deps'], 'tags': annotations['tags']}], sgd=optimizer, losses=losses, drop=0.5)
#         nlp.update([text], [annotations], sgd=optimizer, losses=losses, drop=0.5)

    batches = minibatch(Train_data, size=compounding(4.0, 32.0, 1.001)) 
    for batch in batches: 
        texts, annotations = zip(*batch) 
        nlp.update(texts, annotations, sgd=optimizer, losses=losses, drop=0.5) 
        
    print("Losses: ", losses) 
```
While training, if I use USD symbol('$') in training data, everything works fine but as I replace the dollar symbol with Indian Rupee symbol('₹') in the training data, I get following error:
```
IndexError   Traceback (most recent call last)
<ipython-input-12-4dc4a5e9998c> in <module>
     12     for batch in batches:
     13         texts, annotations = zip(*batch)
---> 14         nlp.update(texts, annotations, sgd=optimizer, losses=losses, drop=0.5)
     15 
     16     print("Losses: ", losses)

~/Documents/venv/spacy3.6/lib/python3.6/site-packages/spacy/language.py in update(self, docs, golds, drop, sgd, losses, component_cfg)
    494             sgd = self._optimizer
    495         # Allow dict of args to GoldParse, instead of GoldParse objects.
--> 496         docs, golds = self._format_docs_and_golds(docs, golds)
    497         grads = {}
    498 

~/Documents/venv/spacy3.6/lib/python3.6/site-packages/spacy/language.py in _format_docs_and_golds(self, docs, golds)
    466                     err = Errors.E151.format(unexp=unexpected, exp=expected_keys)
    467                     raise ValueError(err)
--> 468                 gold = GoldParse(doc, **gold)
    469             doc_objs.append(doc)
    470             gold_objs.append(gold)

gold.pyx in spacy.gold.GoldParse.__init__()

IndexError: list index out of range

```
Please help and thanks in advance
## System Environment
* Operating System: 16.04 Ubuntu
* Python Version: 3.6.9
* SpaCy Version: 2.2.2

adrianeboyd:
## Description

Fix alignment of character entity spans in cases where spacy's tokenization is different from the provided gold words. Create a temporary doc from the text and words in order to be able to use `biluo_tags_from_offsets()` to generate entity tags for those words. In case anything goes wrong, warn the user and discard all entity annotation for the instance.

In order to create a temporary doc, add an option to initialize a `Doc` from a text and list of words where the words may or may not include all whitespace tokens. If the text and words are mismatched, raise an error.

Fixes #4791.

### Types of change

Bugfix.

## Checklist


adrianeboyd:
Do you think the `Doc` options should be in the docs?

honnibal:
> In order to create a temporary doc, add an option to initialize a Doc from a text and list of words where the words may or may not include all whitespace tokens. If the text and words are mismatched, raise an error.

Really not sure I like this --- even if we don't document it, it'll show up in people's entry-helpers and code completion, and will be very confusing. I also don't think I understand the point of it? Is it just to raise an error if something goes wrong?

adrianeboyd:
No, it's not just to raise an error. You need something if you want to support character offsets for entities and people provide `words` for `GoldParse` (typically in the annotation format in `TRAIN_DATA`). (It's not an issue with the IOB tags in the normal training format.)

If you want to support `words` in the `TRAIN_DATA` format, you need something like it somewhere, even if it's not as a `Doc` option. I originally put it in `GoldParse`, but then thought it might be more generally useful, so I moved it to `Doc`.

As an example, you have some gold annotation:

```
text = "It is #December."
words = ["Is", "is", "#December", "."]
entities = [(8, 17, "DATE")]
```

The default tokenization is:

```
nlp = spacy.load('en')
doc = nlp("It is #December.")
assert [t.text for t in doc] == ['It', 'is', '#', 'December', '.']
```

You run `biluo_tags_from_offsets()` as in `GoldParse()`, using the doc generated from the text with the default tokenization:

```
from spacy.gold import biluo_tags_from_offsets
biluo_tags = biluo_tags_from_offsets(doc, words=words, entities=entities)
assert biluo_tags == ['O', 'O', 'B-DATE', 'L-DATE', 'O']
```

And then `GoldParse` tries to align `biluo_tags` with `words`, resulting in:

```
assert zip(words, biluo_tags) == assert list(zip(words, biluo_tags)) == [('Is', 'O'), ('is', 'O'), ('#December', 'B-DATE'), ('.', 'L-DATE')]
```

Basically, if there are any tokenization differences, you can't use `biluo_tags_from_offsets()`. And it gets worse with any extra whitespace in the text, so I thought I'd handle all the possibilities.

The other option would be to discard entity annotation for any instances where the default tokenization differs from `words`.

spacy/gold.pyx
```diff
@@ -716,7 +716,25 @@ cdef class GoldParse:
...
+                    entities = []
+                    words_offset = 0
+                    entdoc_words = [t.text for t in entdoc]
+                    for i in range(len(entdoc_words)):
+                        if words[i + words_offset] == entdoc_words[i]:
+                            entities.append(entdoc_entities[i])
+                        else:
+                            words_offset -= 1
+                    if len(entities) != len(words):
+                        user_warning(Warnings.W027.format(text=doc.text))
```
honnibal:
```suggestion
                        user_warning(Warnings.W028.format(text=doc.text))
```

adrianeboyd:
And let me make another argument for including this: it makes it much easier to deal with external tokenizers that strip whitespace or tokenize whitespace differently.

It would reduce this kind of approach:

https://github.com/explosion/spaCy/blob/1d6aec805d5c03ad8a039466e98ed3a619e650c4/spacy/lang/zh/__init__.py#L38-L67

to something like:

```
jieba_words = list([x for x in self.jieba_seg.cut(text, cut_all=False) if x])
return Doc(self.vocab, words=words, text=text)
```

Similarly, incorporating `pkuseg`, which strips all whitespace, is also a matter of two lines if this is supported.

adrianeboyd:
Closing to split into two separate PRs, one for the doc init (#5251) and one for the character entity spans in GoldParses (to come).

Issue #4200, opened 2019-08-26
text color dont working after spacy in jupyter notebook


phayllia:
## How to reproduce the behaviour

In a new jupyter notebook, print '\033[34mtest\033[0m' produces the text in blue as expected.

>>import spacy
>>print '\033[34mtest\033[0m'
After importing, the text color print doesn't work anymore

## Your Environment

* Operating System: Windows
* Python Version Used: 2.7
* spaCy Version Used: 2.0.9
* Environment Information:

svlandeg:
Yes, I can confirm that this happens indeed, something in the `spaCy` installation is preventing the color to print. @ines : any idea why this happens?

ines:
Just testing it – I really hope this won't be too difficult to track down. The thing is, this could happen in any dependency. In spaCy and our libraries, we typically avoid setting anything in the global scope (and of course never monkeypatch built-ins) for reasons like this.

The most related dependency is my package [`wasabi`](https://github.com/ines/wasabi), but it shouldn't be doing anything on import 🤔

svlandeg:
same here but on a windows:
![image](https://user-images.githubusercontent.com/8796347/63755653-590b6480-c8b7-11e9-9cc1-6a415c7f6e6b.png)

svlandeg:
I can take this on if you can't reproduce!

ines:
Ahh, maybe it's a Windows thing? Okay, here are some things to try:

```python
import wasabi
print("\033[34mtest\033[0m")
```

If the above works fine:

```python
from wasabi import Printer
msg = Printer()
print("\033[34mtest\033[0m")
```

If those work, it's most likely not in `wasabi` where we actually handle the related printing stuff.

svlandeg:
It's not `wasabi`  - the color stays blue. Let me try out a few other `spaCy` dependencies ;-)

svlandeg:
@phayllia : are you running this on a Windows?

ines:
(Sorry, misclicked!) Thanks for checking. And damn, I was hoping it'd be `wasabi`, since it's a super small zero-dependency codebase and ours, so it would have been easy to resolve. Still hope it's in one of our dependencies so we can just fix it, and not in a third-party package.

svlandeg:
I can't pin it on one of the dependencies :(

ines:
[...]

PR #4804, opened 2019-12-13
Restore tqdm imports

svlandeg:
## Description
Issue #4200 revealed a [bug](https://github.com/tqdm/tqdm/issues/678) in `tqdm` that was messing up colors for Windows users in Jupyter. We made an ugly temp hack [fix](https://github.com/explosion/spaCy/pull/4263) by moving our `tqdm` imports to their respective functions. Meanwhile, the original has been [fixed](https://github.com/tqdm/tqdm/commit/ae6de71c3460123944aa7bb80d7cb0b5bdcafb7b) over at `tqdm`, so we can revert our hack as well.

`tqdm`==4.38.0 is the first version with the fix, so I pinned that in our dependencies.

### Types of change
enhancement

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ numpy>=1.15.0
 requests>=2.13.0,<3.0.0
 plac>=0.9.6,<1.2.0
 pathlib==1.0.1; python_version < "3.4"
-tqdm>=4.38.0
+tqdm>=4.38.0,<5.0.0
 # Optional dependencies
 jsonschema>=2.6.0,<3.1.0
 # Development dependencies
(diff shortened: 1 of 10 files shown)


requirements.txt
```diff
@@ -12,6 +12,7 @@ numpy>=1.15.0
 requests>=2.13.0,<3.0.0
 plac>=0.9.6,<1.2.0
 pathlib==1.0.1; python_version < "3.4"
+tqdm>=4.38.0
```
ines: Maybe add an upper range pin, just to be safe?

```suggestion
tqdm>=4.38.0,<5.0.0
```

svlandeg: Done!

PR #4811, opened 2019-12-16
Reduce mem usage in training Entity Linker

svlandeg:
## Description
Optimizing the EL pipeline to be less memory-heavy. In practice this means that the same documents will have to be reprocessed with `nlp()` in each epoch, but storing in memory or on file seems worse. There's always a memory/cpu/disk trade-off but this change should hopefully make the pipeline runnable on most machines.

Additional minor edits:
- provide more informative progress bars
- warn up-front if output directory is not specified
- limit train/dev data in number of articles instead of number of entities (more intuitive)

### Types of change
enhancement

## Checklist

### DIFF
--- a/bin/wiki_entity_linking/entity_linker_evaluation.py
+++ b/bin/wiki_entity_linking/entity_linker_evaluation.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 import logging
 import random
 from tqdm import tqdm
-
 from collections import defaultdict
 
 logger = logging.getLogger(__name__)
--- a/bin/wiki_entity_linking/entity_linker_evaluation.py
+++ b/bin/wiki_entity_linking/entity_linker_evaluation.py
@@ -91,9 +91,9 @@ class BaselineResults(object):
         self.random.update_metrics(ent_label, true_entity, random_candidate)
 
 
-def measure_performance(dev_data, kb, el_pipe, baseline=True, context=True):
+def measure_performance(dev_docs, dev_golds, kb, el_pipe, baseline=True, context=True):
     if baseline:
-        baseline_accuracies, counts = measure_baselines(dev_data, kb)
+        baseline_accuracies, counts = measure_baselines(dev_docs, dev_golds, kb)
         logger.info("Counts: {}".format({k: v for k, v in sorted(counts.items())}))
         logger.info(baseline_accuracies.report_performance("random"))
         logger.info(baseline_accuracies.report_performance("prior"))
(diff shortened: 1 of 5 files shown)


bin/wiki_entity_linking/entity_linker_evaluation.py
```diff
@@ -92,133 +94,110 @@ def update_baselines(
...
 
-    return results
+    if context:
+        logger.info(context_results.report_metrics("context only"))
+        logger.info(combo_results.report_metrics("context and prior"))
 
 
-def measure_baselines(data, kb):
+def _add_eval_result(results, doc, correct_ents, el_pipe):
     """
```
svlandeg: I wish Github hadn't messed up the comparison so badly, but basically `get_eval_results` got rewritten to `_add_eval_result` and `measure_baselines` to `_add_baseline` so that the iteration over the generator happens in the main function `measure_performance` and the rewritten functions now take one `doc` at a time. Not much changed otherwise.

svlandeg:
Update: this PR allows training the full WP dataset (150K articles) on CPU with 16GB RAM, with one epoch taking 20h.

cmenguy:
I was trying to train a model on a 50k / 10k split with this PR for 1 epoch, but the progress bar seems to report a very long training time for the training set (seems to be processing at ~10s/it and expected to complete in 150hrs)

Could you share @svlandeg if you had set any parameter to get it under 10hrs for the same train/test set sizes, and your hardware configuration? I was using an AWS instance c5d.9xlarge for this.

svlandeg:
@cmenguy: Here's an overview of the settings I've used and the resulting output:

1. **wikidata_pretrain_kb.py**
    - run simply on my laptop
    - use `en_core_web_lg` to process `enwiki`
    - STEP 2 prior probabilities (default settings): _processed 1094894680 lines of Wikipedia XML dump_
    - STEP 4 entity defs (default settings): _Processed 55075951 lines of WikiData JSON dump_
    - STEP 5: `-lt 200000` --> _reading only 200000 lines of Wikipedia dump_
    - STEP 6: `-f 15` --->
         - _Filtering entities with fewer than 15 mentions_
         - _Kept 1250555 entities from the set of 5516028_

    - Resulting KB:
         - File size of `kb` is about 381MB
         - kb entities: 1250 555
         - kb aliases: 1 696 838

    - Resulting gold dataset:
         - File size of `gold_entities.jsonl` is 2,76 GB

2. **wikidata_train_entity_linker.py**
    - run on `r5.large` instance
    - `-d 1000` --> evaluate on 1000 dev articles (it does this every epoch, takes about half an hour)
   - `-l CARDINAL,DATE,MONEY,ORDINAL,QUANTITY,TIME,PERCENT` 

This takes about 20h to train 165.000 articles.

cmenguy:
Thank you @svlandeg it does seem that filtering down to 200k lines to reduce the size of the gold entities in pre-training made a huge difference in runtime and now it seems to be similar to what you wrote, this is helpful. 

Would you say that 200k lines is enough to get good overall accuracy, or would you recommend using more or perhaps the entire dump for some real-world applications?

Very excited to be able to use the EL feature from spaCy!

svlandeg:
I personally think that using the entire dump is probably overkill. I haven't done an exhaustive search yet, but the model built on 165K articles seemed to perform well.

amitbcp:
@svlandeg  I tried creating a full Wikidata KB and followed the above discussions as well as https://github.com/explosion/spaCy/issues/4870  comments. Also tried testing it on a small corpus.

The issue I am facing is continuously gold_entities.jsonl is empty. I deleted it ans re-run the wikidata_pretrain_kb.py .  This makes it run the step 5 again. But still an empty gold_entities.jsonl is created.

Do you have any hunch on what might be going wrong ?

PR #4828, opened 2019-12-21
Drop Python 2.7 and 3.5

ines:
## Description

🎉🎉🎉

- [x] remove unicode declarations
- [x] remove versions from CI
- [x] remove pathlib dependency
- [x] update compat and remove unused helpers
- [x] remove `OrderedDict` wherever possible
- [x] use f-strings wherever possible

### Types of change
enhancement, compat

## Checklist


spacy/cli/debug_data.py
```diff
@@ -437,110 +357,93 @@ def debug_data(
...
-        msg.warn(
-            "{} {}".format(warn_counts, "warning" if warn_counts == 1 else "warnings")
-        )
-    if fail_counts:
-        msg.fail("{} {}".format(fail_counts, "error" if fail_counts == 1 else "errors"))
-
+        msg.warn(f"{warn_counts} {'warning' if warn_counts == 1 else 'warnings'}")
     if fail_counts:
+        msg.fail(f"{fail_counts} {'error' if fail_counts == 1 else 'errors'}")
         sys.exit(1)
```
svlandeg: as in the cases you've corrected above, I guess these could also just simply be "check(s"), "warning(s)" and "error(s)"

website/docs/api/lookups.md
```diff
@@ -215,12 +215,11 @@ the file doesn't exist.
...
-more consistent and unified API and includes a Bloom filter to speed up missed
-lookups. Supports **all other methods and attributes** of `OrderedDict` /
-`dict`, and the customized methods listed here. Methods that get or set keys
-accept both integers and strings (which will be hashed before being added to the
-table).
+A table in the lookups. Subclass of `dict` that implements a slightly more
+consistent and unified API and includes a Bloom filter to speed up missed
+lookups. Supports **all other methods and attributes** of `dict`, and the
+customized methods listed here. Methods that get or set keys accept both
+integers and strings (which will be hashed before being added to the table).
```
svlandeg: Table is still implementing OrderedDict, but I guess maybe that's an implementation detail that you want to keep out of the docs ?

ines: Ahh, good catch – I had initially replaced `OrderedDict` with `dict`, but it didn't work for some really confusing reason (or maybe there's something I missed, not sure). So I ended up changing it back to `OrderedDict` for now.

PR #4831, opened 2019-12-21
Add better schemas and validation using Pydantic

ines:
## Description

Depends on #4828.

We're now able to validate our data much better, using [`pydantic`](https://github.com/samuelcolvin/pydantic). Here are some first example schemas (can probably be improved). We can probably also replace a lot of arbitrary conditional logic for data validation across the library with this, and we're also planning on using it to create and validate the new training config stuff. So I think it's reasonable to adopt a new dependency 🙂

**Note on the matcher validation:** The errors we get back now are more detailed than the previous JSON schema errors. For instance, if a value can be a dict or a string and it's an invalid dict, we'll get 2 errors. I've adjusted the tests accordingly. Maybe there's a way to make it behave more like the previous validation – although, I'm not even sure if we'd want that.

### Types of change
enhancement, ux

## Checklist


spacy/schemas.py
```diff
@@ -0,0 +1,188 @@
...
+        for error in errors:
+            err_loc = " -> ".join([str(p) for p in error.get("loc", [])])
+            data[err_loc].append(error.get("msg"))
+        return [f"[{loc}] {', '.join(msg)}" for loc, msg in data.items()]
+
+
+# Matcher token patterns
+
+
+def validate_token_pattern(obj):
```
ines: Quick note on the conversion part below: the problem here is that patterns may use symbols (ints) instead of the string values, e.g. `{ORTH: "foo"}` instead of `{"ORTH": "foo"}`. This trips up the validation, because the object seems to be spread into the schema as keyword arguments and you end up with non-string keyword argument names, which causes an error.

spacy/schemas.py
```diff
@@ -0,0 +1,188 @@
...
+from .attrs import NAMES
+
+
+def validate(schema, obj):
+    """Validate data against a given pydantic schema.
+
+    obj (dict): JSON-serializable data to validate.
+    validator (pydantic.BaseModel): The schema to validate against.
+    RETURNS (list): A list of error messages, if available.
+    """
```
svlandeg: Docstring: `validator` --> `schema` ? (+ switch order)

spacy/schemas.py
```diff
@@ -0,0 +1,188 @@
...
+    IN: Optional[List[StrictStr]]
+    NOT_IN: Optional[List[StrictStr]]
+
+    class Config:
+        extra = "forbid"
+
+    @validator("*", pre=True, whole=True)
+    def raise_for_none(cls, v):
+        if v is None:
+            raise ValueError("None / null is not a allowed")
```
svlandeg: typo (occurs three times in this file)

PR #4840, opened 2019-12-27
stop rendering mathjax by default in displacy

AlJohri:
By default, Jupyter renders MathJax between any two dollar signs. When running displacy through some news documents I was seeing very odd rendering due to this behavior. Using this [SO answer](https://stackoverflow.com/questions/16089089/escaping-dollar-sign-in-ipython-notebook), I added a span tag to stop rendering mathjax when `jupyter=True`.

signed contributor agreement: https://github.com/explosion/spaCy/pull/4839

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/displacy/__init__.py
+++ b/spacy/displacy/__init__.py
@@ -55,9 +55,10 @@ def render(
         html = RENDER_WRAPPER(html)
     if jupyter or (jupyter is None and is_in_jupyter()):
         # return HTML rendered by IPython display()
+        # See #4840 for details on span wrapper to disable mathjax
         from IPython.core.display import display, HTML
 
-        return display(HTML(f'<span class="tex2jax_ignore">{html}</span>'))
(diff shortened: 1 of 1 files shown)


AlJohri:
Added some visual "test" cases:
Screen Shot 2019-12-27 at 10 29 11 AM

Alternatively, this can be implemented by escaping the dollar sign character but I think this is an easier and less error-prone approach.

AlJohri:
Here's a small work around until this gets merged in case anyone else is running into this:

```python
from spacy import displacy
displacy.RENDER_WRAPPER = lambda html: f'<span class="tex2jax_ignore">{html}</span>'
```

ines:
Thanks – I had no idea this was a thing! Thanks for the fix 🙂

I just need to make one small edit and remove the f-string, since this goes on `master` and we're still testing for Python 2 / 3.5 on there.

PR #4843, opened 2019-12-28
Fixes typos

iurshina:


spacy/_align.pyx
```diff
@@ -30,7 +30,7 @@ S[:i]   -> T[:j]   (at D[i,j])
 S[:i+1] -> T[:j]   (at D[i+1,j])
 S[:i]   -> T[:j+1] (at D[i,j+1])
     
-Further, we now we can tranform:
+Further, we now we can transform:
```
svlandeg: we could also remove the duplicate "we" here !

iurshina: done :)

Issue #4748, opened 2019-12-03
Memory usage of `debug-data` with a huge training set


sfragis:
Hi, I'm using Spacy 2.2.2 to train new tagger and parser models for the Italian language.
My training data set is quite big (about 2.3 GB for the train and 580 MB for the dev) and is saved in two JSONL files.
I'm experiencing an unexpected memory usage when running the `debug-data` command: memory usage starts low and then grows up to consuming my 32GB of RAM as well as the whole swap (about the same size).
Before upgrading my RAM to 128 GB (which I suspect might be useless), I'm interested in your opinion about:
* hints about data set structure: for instance, comments in issue #4700 suggested to reduce the sentence length on average, but I've no clue about what values might be optimal; is there any rule of thumb to properly dimension the data set?
* possible optimizations to the source code to reduce memory footprint (for instance by improving the lazy loading of the data set); I'm willing to contribute to Spacy if anyone would kindly point me to the problematic parts (if any, of course)

## Info about spaCy

* **spaCy version:** 2.2.2.maps
* **Platform:** Linux-4.4.0-112-generic-x86_64-with-debian-stretch-sid
* **Python version:** 3.7.4

ines:
Thanks for the report! 

> My training data set is quite big (about 2.3 GB for the train and 580 MB for the dev) and is saved in two JSONL files.

You probably want to split these into multiple files. spaCy can also read from directories instead of single JSON files, so there's really no need to have a 2.3 GB file. This could easily cause other problems down the line.

About `debug-data`:  Since the `debug-data` command is really mostly a debugging utility, we didn't particularly focus on optimising it for efficiency. For instance, I'm pretty sure we're just loading the whole corpus into memroy (e.g. by calling `list(corpus.train_docs(nlp))`), and I think we're also making at least one additional pass over the data to compute the stats. That's typically okay, because you're usually just running the debugging manually a few times and even if you have to wait for a few minutes, that's usually not a big deal.

However, if it's not memory-efficient and you can't use it with large data files, that's obviously bad.

We could probably refactor the logic to only process the data as a stream, make one single pass over each corpus and compute all the stats that way. You can find the source here if you want to give it a try and see if it improves things for you: https://github.com/explosion/spaCy/blob/master/spacy/cli/debug_data.py

sfragis:
Hi Ines, thank you for your quick reply.
I successfully managed to read the whole dataset from JSONL and have it saved into smaller MessagePack files. 
[...]

Issue #4700, opened 2019-11-23
Huge RAM consumption during NER training


mosynaq:
I am trying to train a NER model from scratch. I have successfully built `train.json`, `dev.josn`, and `test.json`. I issue the below command to start training:
```python -m spacy train xx out train.json dev.json -p ner```
I have a PC with +60GB of RAM. spaCy occupies all but ~500MB. After some time and doing nothing, it stops and spits something like "not enough memory"!
I had this very problem with spaCy 2.2.2 too.
I use python v3.7.5, spaCy 2.2.3, and Windows 10

honnibal:
You must have an instance that's really long. We should probably put the length cap back in, or possibly raise an error. What's the longest document in your data?

mosynaq:
Thank you @honnibal! After I looked into the training data, I found multiple tool-long sentences. After they were split into smaller ones, the training works like a charm!
Thanks!

Issue #3698, opened 2019-05-07
Show doc['id'] or at least some other infos for errors when training with CLI


Arakkun:
I'm training a model, and obtain the 
"ValueError: [E069] Invalid gold-standard parse tree. Found cycle between word IDs:  ..." 
Knowing which doc is causing the error or which phrase, or at least a bit more infos about the error would be really useful. Especially as the JSON input format for training has already an 'id' field meaning "# ID of the document within the corpus"

sairampillai:
@ines If no one is working on it, I would like to work on this issue

ines:
@sairampillai Thanks! And yes, feel free to go ahead and prepare the PR! 😃

rbhambriiit:
Hitting this on v2.0.18
is someone working on this?

rbhambriiit:
can't we just skip cases with bad training data somehow? @ines

ines:
The error is raised here:

https://github.com/explosion/spaCy/blob/26c37c5a4dc8acf113e9b74f3162ed2a3bc74aea/spacy/gold.pyx#L568-L569

So in the meantime, you can always build spaCy from source and hack in a print statement here if you need it 🙂

rbhambriiit:
Thanks. I tried to build from source as you mentioned by following: https://spacy.io/usage#source

I hit "RuntimeError: Python version >= 3.5 required." - perhaps documentation needs to be updated with this for venv to require certain python version

Now i am running python3.6 and spacy v2.1.4
(the original version was v2.0.18 - where i hit the cycle issue and reached here.)

I did another run to reproduce the issue in this new spacy build - hitting a different scenario now

"""
(.env36) -bash-4.2$ python -m spacy train -n 5 en models_expt1 train_expt1_july1.json valid_expt1_july1.json
Training pipeline: ['tagger', 'parser', 'ner']
Starting with blank model 'en'
Counting training words (limit=0)
Traceback (most recent call last):
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/app/data/rahul/playground/spacy_2.1/spaCy/spacy/__main__.py", line 35, in <module>
    plac.call(commands[command], sys.argv[1:])
[...]

PR #4844, opened 2019-12-28
Debug data memory footprint and document identifier printed on errors 

sfragis:
## Description
This pull request is an attempt to address issues #4748, #4700 and #3698.
Changes are against version 2.2.3 of spaCy and pass all tests.

### Types of change
Concerning the memory footprint (the first two issues), my proposal is to go lazy whenever is possible in `debug_data`. From my understanding, the problem with the original implementation was that the whole training and dev datasets where eagerly moved into a list in order to catch any exceptions. This caused severe memory usage with huge datasets. I changed this logic and deferred the `try-except` logic until the `_compile_gold` is invoked (here is where the iteration over the dataset really happens).
There was also another point that required refactoring, that is when `_get_examples_without_label` is invoked; in this case, a second read of the dataset was necessary.

In order to address issue #3698, I introduced a new class `DocTuple` that mimics the behavior of a list of two elements but can optionally be used to retrieve the document identifier. The main point here is to not lose the document identifier when `json_to_tuple` is invoked while behaving the same in every line where the document tuple is destructured into two variables, like in this excerpt from `iter_gold_docs` in `gold.pyx`:

```py
        for raw_text, paragraph_tuples in tuples:
            if gold_preproc:
                # ...
```

Command `debug-data` can be invoked with a new `IE` flag; if passed, errors in the training/dev set will be collected into a list to be printed when iteration is completed. Otherwise, iteration stops as usual. In case of errors, failure messages are printed along with the document identifiers.

I've also added some timers to track dataset loading times; log messages can be turned off using the new input flag `NT`.
 

## Checklist

- [x] My changes don't require a change to the documentation, or if they do, I've added all the required information.


spacy/gold.pyx
```diff
@@ -216,9 +236,11 @@ class GoldCorpus(object):
         RETURNS (GoldCorpus): The newly created object.
         """
         self.limit = limit
+        self._max_sent_length = {'train': 0, 'dev': 0}
+        self._avg_sent_length = {'train': [0, 0, 0], 'dev': [0, 0, 0]}
```
svlandeg: This feels a little bug-prone to me. Perhaps we should have a small datastructure instead, which keeps track of `max`, `len_sum`, `count` and `len2_sum` for one dataset, and then have one such datastructure for train and dev each ?

spacy/gold.pyx
```diff
@@ -494,9 +553,47 @@ def read_json_object(json_corpus_section):
...
     """
     for json_doc in json_corpus_section:
-        tuple_doc = json_to_tuple(json_doc)
-        for tuple_paragraph in tuple_doc:
-            yield tuple_paragraph
+        yield from json_to_tuple(json_doc)
+
+
+class DocTuple(object):
+
```
svlandeg: Unfortunately we have been working on a similar idea on `develop`: a new [`Example`](https://github.com/explosion/spaCy/blob/develop/spacy/gold.pyx#L711) class that keeps track of annotations as well as of the original doc/text. Here, we could potentially add the `doc_id`. You would then define it when you call `json_to_examples` (renamed from `json_to_tuple`) in a similar way as you're doing it now.

spacy/gold.pyx
```diff
@@ -355,25 +402,37 @@ class GoldCorpus(object):
...
             except AlignmentError:
                 if ignore_misaligned:
                     gold = None
                 else:
                     raise
+            except ValueError as e:
+                if errors is not None and not error_occurred:
+                    t = (doc_id, e)
+                    errors.append(t)
+                    error_occurred = True
```
svlandeg: is the purpose of `error_occurred` to only record one error per doc? If so, shouldn't it be set to `False` as the first line within the `for` loop rather than before it ?

sfragis:
Hi Sofie, thanks for your review, I agree on all of your remarks.
I'm happy to contribute to spaCy, so I will try to create another pull request starting from the develop branch so to take advantage of the `Example` class and any other change.
This will probably take me some time, but it's the best option for sure.
We can close this pull request in the meantime.

svlandeg:
Ok, great! I'll close this one, but feel free to comment here if you run into issues preparing the PR against `develop`. Some function & variable names have changed, usually from XXX_tuples_XXX to XXX_examples_XXX but if you need any help I'd be happy to look into it with you !

Issue #4554, opened 2019-10-30
Issues with gold alignment


adrianeboyd:
## How to reproduce the behaviour

The new alignment method from @tamuhey is very nice (#4526) but it doesn't allow substitutions, which is problem for some of our corpora with minor differences between the raw texts and token orths in things like quotes. Ideally we would require that the token orths exactly match the texts, but since we do have some slightly noisy corpora, I think we need to keep using the old align, or at least a method that allows some noise.

Comparing the new and old align functions was very useful because it brought our attention to the old alignment code, which has some bugs we can fix:

* `j2i_multi` is incorrect (from #4525). The problem is that `_get_missing()` isn't symmetric, modifies `miss` regions as it's working, and doesn't quite backtrack enough when alignments aren't found (I think, anyway). It could probably be rewritten, but a quick solution is to call it twice with reversed arguments to get `i2j_multi` and `j2i_multi` separately, since the first returned alignment `i2j_multi` (which is all we use in `spacy.gold`, as far as I know) is correct.

* The Levenshtein distance costs are different with and without substitutions (old vs. new). An example:

```
words1 = ["a", "b", "cd"]
words2 = ["ab", "c", "d"]
```

The old alignment code says the cost is 3 (3 substitutions) and the new alignment code says it's 6 (3 insertions and 3 deletions).

The old alignment produces this:

```
cost:         3
i2j:          [-1  1 -1]
j2i:          [-1  1 -1]
i2j_multi:    {}
j2i_multi:    {}
```

New alignment:

```
cost:         6
i2j:          [-1 -1 -1]
j2i:          [-1 -1 -1]
i2j_multi:    {0: 0, 1: 0}
j2i_multi:    {1: 2, 2: 2}
```

* Looking at the previous example, I think that the results for `i2j` are a little unexpected for the old alignment. I think it's odd that tokens of the same length count as substitutions while tokens of different lengths don't and that's the only thing taken into consideration.

https://github.com/explosion/spaCy/blob/c2f5f9f5727fd854a6147e71ed6625b7c57c4150/spacy/_align.pyx#L107-L112

This might (nearly) work for the kind of punctuation noise present in the English data, but I wouldn't be surprised if there are some strange alignments for other corpora.

I think this could be solved in a few ways, either by increasing the cost of substitution in the Levenshtein calculation in most cases or modifying the "unaligned" filtering for `i2j` to be a little more sophisticated.

[...]

PR #4878, opened 2020-01-05
Update gold.align with pytokenizations

tamuhey:
From #4526 , #4554 
I created new `gold.align` in #4526 , but this algorithm assumes both toknizations add up to the same string and doesn't allow any noise (described in #4554).
Thus, I updated the `gold.align` based on [pytokenizations](https://github.com/tamuhey/tokenizations), which is I created for this purpose.
This library allows noisy tokenizations, and is very fast because it is implemented in Rust.

### Algorithm overview

1. Normalize tokens in the unicode normalization form "NFKD", then lowercase all characters.
2. Concatenate all tokens `tokens_a` and `tokens_b` to generate `ta` and `tb` respectively
3. Calculate shortest path on edit graph of `ta` and `tb` by [Myers' algorithm](http://www.xmailserver.org/diff2.pdf)
4. Get character mapping`c_ab` and `c_ba` from the edit graph
5. Get`al_ab` and `al_ba` from the character alignments `c_ab` and `c_ba`
6. Get `a2b`, `b2a`, `a2b_multi`, `b2a_multi`, and `cost` from `al_ab` and `al_ba`

### Types of change

an enhancement

## Checklist

I think the [doc](https://spacy.io/api/goldparse#align) should be modified.


tamuhey:
Note:

- [pytokenizations](https://github.com/tamuhey/tokenizations) doesn't support Python2.7, so I defined `_USE_PYTOKENIZATIONS`  so that `gold.align` can work even if pytokenizations is not installed.

tamuhey:
srsly build fails?

honnibal:
I really appreciate the effort, but I think we'd be pretty reluctant to accept a new third-party dependency on a compiled module, especially written in a language we don't already maintain anything in.

The thing is, making sure wheels get built correctly as new versions come out, and making sure everything stays working across platforms as software stacks change, and ensuring everything stays working across both pip and conda, is a fairly big commitment. For our own libraries, we can take responsibility for this and it's mostly worth the effort for us, because we're doing this professionally. 

But we couldn't ask a third-party volunteer to do what we need. Even if you're happy to do it today, it's totally reasonable for your plans to change, and it shouldn't be your responsibility to make sure a bunch of companies get a smooth experience using some free code you wrote some time ago.

adrianeboyd:
I also appreciate the effort, but I'd prefer to keep the constraint that the tokens have to match the text because it simplifies so many things.

tamuhey:
Thanks, I got it. I think Cython is harder to maintain than Rust (debugging, memory management, and so on), but it can be also implemented in Cython.

I think an align function that can handle noisy tokens is necessary. Actually it is useful for [spacy-transformers](https://github.com/explosion/spacy-transformers/pull/123).

honnibal:
@tamuhey Yes I agree that better handling of noisy alignments will be useful, especially in the `spacy-transformers`. I hope we can study your implementation later.

We've been working on a big update to the transformers support, which I hope we can show you soon :). If you want to test something early, feel free to email me at matt@explosion.ai 

In the meantime, I'll close this particular PR, because we want to avoid the Rust dependency, so we'll need to take a different approach.

PR #4882, opened 2020-01-06
Fix and improve URL pattern

### LINKED ISSUE #4862: Tokenization of URLs is different between English language model 2.2.0 and 2.2.5
Hi,

There was a change in tokenizing URLs between sPacy language model en_core_web_sm 2.2.0 and 2.2.5

This code:
```
import spacy
doc = nlp('the url is https://www.example.com.ai/foo/bar/)
for t in doc:
  print(t.lower_)
```

Using en_core_web_sm 2.2.0, the output is:
```
the
url
is
https://www.example.com.ai/foo/bar/
```

Using en_core_web_sm 2.2.5, the output is:
```
the
url
is
https://www.example.com.ai
/
foo
/
bar/
```

But the weird thing is when using just `https://www.example.com/foo/bar/` without `.ai` in domain, the output is the same for both model versions
```
the
url
is
https://www.example.com/foo/bar/
```

Any ideas? Thanks.

## Your Environment
* Operating System: macOS Catalina 10.15.2
* Python Version Used: 3.7.6
* spaCy Version Used: 2.2.3
* Environment Information: Language model en_core_web_sm 2.2.0 and 2.2.5

adrianeboyd:
## Description

* match domains longer than `hostname.domain.tld` like `www.foo.co.uk`
* expand allowed characters in domain names while only matching lowercase TLDs so that "this.That" isn't matched as a URL and can be split on the period as an infix (relevant for at least English, German, and Tatar)

Fixes #4862.

### Types of change

Bugfix and enhancement.

## Checklist

### DIFF
--- a/spacy/lang/tokenizer_exceptions.py
+++ b/spacy/lang/tokenizer_exceptions.py
@@ -3,14 +3,18 @@ from __future__ import unicode_literals
 
 import re
 
+from .char_classes import ALPHA_LOWER
 from ..symbols import ORTH, POS, TAG, LEMMA, SPACE
 
 
 # URL validation regex courtesy of: https://mathiasbynens.be/demo/url-regex
-# A few minor mods to this regex to account for use cases represented in test_urls
+# and https://gist.github.com/dperini/729294 (Diego Perini, MIT License)
+# A few mods to this regex to account for use cases represented in test_urls
 URL_PATTERN = (
+    # fmt: off
     r"^"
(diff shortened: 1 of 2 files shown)


adrianeboyd:
The old pattern "worked" on longer domains because this line matched too much:

https://github.com/explosion/spaCy/blob/7f82a1fe1bee156f82a59a58d961ec5770407f8f/spacy/lang/tokenizer_exceptions.py#L48

PR #4883, opened 2020-01-06
Mark most Hungarian tokenizer test cases as slow

adrianeboyd:
## Description

Mark most Hungarian tokenizer test cases as slow to reduce the runtime of the test suite in ordinary usage:

* for normal tests: run default tests plus 10% of the detailed tests
* for slow tests: run all tests

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/tests/lang/hu/test_tokenizer.py
+++ b/spacy/tests/lang/hu/test_tokenizer.py
@@ -296,9 +296,8 @@ WIKI_TESTS = [
     ("cérium(IV)-oxid", ["cérium", "(", "IV", ")", "-oxid"]),
 ]
 
-TESTCASES = (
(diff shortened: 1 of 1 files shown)


adrianeboyd:
The Hungarian tokenizer tests were nearly 20% of the running time of the non-slow test suite. This reduces the time for non-slow Hungarian tokenizer tests from ~30 s to ~5 s on my laptop.

adrianeboyd:
And there might be a better way to define the test cases without creating two lists and repeating the test function, but I couldn't figure it out from the pytest docs.

honnibal:
Was running the test cases today and totally see what you mean about it being a bit impractical.

PR #4890, opened 2020-01-08
Add sentrec shortcut to Language

adrianeboyd:
## Description

Add sentrec shortcut to `Language`. (I still think it needs to be renamed.)

### Types of change

Enhancement.

## Checklist


ines:
(I'm labelling these `feat / sentencizer` for now – we should probably rename that label to `feat / sentences` or something like that to make it more general.)

PR #4900, opened 2020-01-10
Fix ent_ids and labels properties when id attribute used in patterns

kabirkhan:
When using the "id" attribute in EntityRuler patterns right now, the labels property gives different strings for the same label because of the way patterns are stored.

## Description

Currently labels returns something like:
e.g. 

```python
[
    ...
    "TECH_ORG||apple",
    "TECH_ORG||microsoft",
    "TECH_ORG||google",
    ...
]
```

This PR fixes this so that the labels property only includes

```python
[
    ...
    "TECH_ORG",
    ...
]
```

and the ent_ids property has

```python
[
    ...
    "apple",
    "microsoft",
    "google",
    ...
]
```

## Checklist

### DIFF
--- a/spacy/pipeline/entityruler.py
+++ b/spacy/pipeline/entityruler.py
@@ -131,15 +131,15 @@ class EntityRuler(object):
         """
         keys = set(self.token_patterns.keys())
         keys.update(self.phrase_patterns.keys())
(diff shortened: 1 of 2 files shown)


ines:
Thanks! 👍

PR #4904, opened 2020-01-13
Adding a note on retrieving the string rep of the match_id

maknotavailable:
## Description
Stolen from here: https://stackoverflow.com/questions/47638877/using-phrasematcher-in-spacy-to-find-multiple-match-types

### Types of change
Change to documentation.


ines:
Thanks! 🙏

PR #4911, opened 2020-01-15
Improve train CLI with base model

### LINKED ISSUE #4342: Simplify adding new component to existing model with CLI
## Feature description

It looks like these kinds of cases aren't easy to handle with the train CLI:

1. I want to train a model from scratch given vectors, one corpus for the tagger, and one corpus for the parser. (See example in #4306.)

2. I have a model with a tagger and want to add a parser trained on a separate corpus.

For the internal spacy models, it looks like each component is trained separately and then they are combined using custom scripts.

Could it make sense to have a CLI component that combines models/components (with compatibility checks, of course)?

Case (2) is really just a minor variant of (1), but the train CLI might be able to handle it relatively easily by combining the components in `model-final`, for example. I think case (1) could be handled by the train CLI in theory, but the command-line options would get too complicated and it would be easier to handle it with a separate CLI command.

(What is the right way to combine the `vocab` directories (aside from `vectors`) for multiple models?)

adrianeboyd:
## Description

Improve train CLI with a provided base model so that you can:

* add a new component
* extend an existing component
* replace an existing component

When the final model and best model are saved, disabled components are reenabled and the meta information is merged to include the full pipeline and accuracy information for all components in the base model plus the newly added components if needed.

Fixes #4342.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -30,6 +30,7 @@ from .. import about
     raw_text=("Path to jsonl file with unlabelled text documents.", "option", "rt", Path),
     base_model=("Name of model to update (optional)", "option", "b", str),
     pipeline=("Comma-separated names of pipeline components", "option", "p", str),
+    replace_components=("Replace components from base model", "flag", "R", bool),
     vectors=("Model to load vectors from", "option", "v", str),
     n_iter=("Number of iterations", "option", "n", int),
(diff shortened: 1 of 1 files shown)


honnibal:
Agree this has been a problem; I hope the `train-with-config` stuff that's coming can make this better as well. The `train` CLI function has been getting pretty out of control.

Happy to merge this in the meantime though.

Issue #4924, opened 2020-01-20
`nlp.evaluate` with empty `docs_golds` causes a bug


tamuhey:
## How to reproduce the behaviour

```python
nlp = spacy.blank("en")
docs_golds = [("", {})]
nlp.evaluate(docs_golds)
```

Causes the error:

```
        gold_deps = set()
        gold_deps_per_dep = {}
        gold_tags = set()
>       gold_ents = set(tags_to_entities([annot[-1] for annot in gold.orig_annot]))
E       TypeError: 'NoneType' object is not iterable

spacy/scorer.py:239: TypeError
```

## Your Environment

* Operating System: osx
* Python Version Used: 3.7.4
* spaCy Version Used: master
* Environment Information:

tamuhey:
created PR in #4925

svlandeg:
Just as a small tip, if you put something like `Fixes #4924` in your PR description, it will automatically close the original issue when merging the PR, and on top will mention something like `Fixed by #4925` ;-)

PR #4925, opened 2020-01-20
fix nlp.evaluate (#4924)

tamuhey:
## Description

From #4924.

I found `nlp.evaluate` fails with empty `doc` and `gold`.
This seems to be unintended, because the error is somewhat confusing:

```
        gold_deps = set()
        gold_deps_per_dep = {}
        gold_tags = set()
>       gold_ents = set(tags_to_entities([annot[-1] for annot in gold.orig_annot]))
E       TypeError: 'NoneType' object is not iterable

spacy/scorer.py:239: TypeError
```

I thought the following three ways to correct this:

1. Informs the users that `nlp.evaluate` will not accept an empty `gold`, and fails.
2. Modify `GoldParse` to set `gold.orig_annot = []` instead of `None`, if `gold` is empty.
3. Modify `nlp.evaluate` to properly handles empty `gold`

I adopted the second way.
In the first way, users should write a boilerplate to check `gold` is empty.
In the third way, it is necessary to introduce multiple `if`s, which complicates the code.

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/gold.pyx
+++ b/spacy/gold.pyx
@@ -694,6 +694,11 @@ cdef class GoldParse:
         self.cats = {} if cats is None else dict(cats)
         self.links = links
 
+        # orig_annot is used as an iterator in `nlp.evalate` even if self.length == 0,
+        # so set a empty list to avoid error.
+        # if self.lenght > 0, this is modified latter.
+        self.orig_annot = []
+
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks!

Issue #4119, opened 2019-08-14
Adding patterns to EntityRuler and deserializing EntityRuler very slow


jdprusa:
Adding a large number of patterns to the EntityRuler, or loading a saved EntityRuler that has a large number of patterns is extremely slow. 

This is easily reproduced with:

`nlp = spacy.load('en_core_web_sm')`
`entityruler = EntityRuler(nlp)`
`patterns = [{"label": "TEST", "pattern": str(i)} for i in range(100000)]`
`entityruler .add_patterns(patterns)`

and

`nlp = spacy.load('en_core_web_sm')`
`patterns = [nlp.make_doc(str(i)) for i in range(1000000)]`
`phrasematcher= PhraseMatcher(nlp.vocab)`
`phrasematcher.add("TEST", None, *patterns)`

the EntityRuler code takes around 10 minutes to execute on an m5.4xlarge AWS SageMaker Notebook instance, while the PhraseMatcher code takes 20 seconds. Changing `nlp.make_doc(str(i))` to `nlp(str(i))` slows the PhraseMatcher down to the speed of the EntityRuler.

Looking through the code for the EntityRuler, it employs the PhraseMatcher and should be similar in speed, but is using `nlp(pattern)` with the full nlp pipeline (tagger, parser, ner)  instead of using `nlp.make_doc(pattern)` as recommended for the PhraseMatcher in https://spacy.io/usage/rule-based-matching#phrasematcher

Since EntityRuler uses `add_patterns()` when deserializing this also slows down `from_bytes()` and `from_disk()` by a considerable amount. 

This is fortunately an easy fix, just change the offending line 
**Current Line 187** in https://github.com/explosion/spaCy/blob/master/spacy/pipeline/entityruler.py:
`self.phrase_patterns[label].append(self.nlp(pattern)) `
**Updated Line 187:**
`self.phrase_patterns[label].append(self.nlp.make_doc(pattern)) `

This puts pattern adding and loading times in line with the PhraseMatcher as expected and doesn't appear to break anything.
 
## Your Environment
- **spaCy version:** 2.1.8
- **Platform:** Amazon Linux AMI 2018.03
- **Python version:** 3.6.5

adrianeboyd:
[...]

PR #4931, opened 2020-01-21
Use nlp.pipe in EntityRuler for phrase patterns in add_patterns

kabirkhan:
## Description

The EntityRuler will now use nlp.pipe for all phrase patterns. This can drastically reduce time for adding a large number of phrase patterns. 

### Types of change

Enhancement

## Checklist


adrianeboyd:
Take a look at the discussion in #4119. I don't think that we can make the assumption that if an attribute isn't in `(DEP, POS, TAG, LEMMA)` that `nlp.make_doc()` is sufficient.

kabirkhan:
Sorry, could have found that myself I'm sure. Based on that issue, I've refactored the add_patterns function to use nlp.pipe for add_patterns. 

It gets a 10x-20x speedup from the old function in my tests of 5000-100000 phrase patterns.

It's obviously still faster to disable pipeline components when running add_patterns but that's still extra for the user to remember and non-trivial for loading patterns from disk. I've added some docs explaining this issue in more detail. Not at all attached to my writing so feel free to request changes in that explanation.

honnibal:
@kabirkhan @adrianeboyd Sorry for the lack of attention to this. It looks good to me but I wanted to double check with you both, is it safe to merge now?

kabirkhan:
Not yet, shoot I accidentally pushed a commit to this PR based on this WIP PR: https://github.com/explosion/spaCy/pull/4988. I'll revert that commit here.

kabirkhan:
Ok should be all good now, sorry about that. 

Should be safe to merge

PR #4953, opened 2020-01-30
Fix for Issue 4665 - conllu2json

### LINKED ISSUE #4665: conll2json: Underscore not allowed in head column
I got a conllU file, from my university, where the head column is filled with _. 
Processing such file with the cli.convert method will result in a int cast error in 
https://github.com/explosion/spaCy/blob/master/spacy/cli/converters/conllu2json.py line 73 
in the read_conllx method (head = (int(head) - 1) if head != "0" else id_).

In the format documentation on https://universaldependencies.org/format.html
there is stated that: the allowed values for head are the same as for id (so _ should not be allowed)
"7. HEAD: Head of the current word, which is either a value of ID or zero (0)."

But some lines later, underscore "_" is only forbidden for the id column.
"Underscore (_) is used to denote unspecified values in all fields except ID. Note that no format-level distinction is made for the rare cases where the FORM or LEMMA is the literal underscore – processing in such cases is application-dependent. Further, in UD treebanks the UPOS, HEAD, and DEPREL columns are not allowed to be left unspecified."

So, the university guys who gave me the problematic conll file insits on beeing right in allowing _ in the head column and asked me to write a bug report here.

onlyanegg:
Allowing the HEAD field in CoNLL-U format to be an underscore

## Description
Added a check for underscores and added a regression test
Fixes #4665 

### Types of change
Bug Fix

## Checklist

### DIFF
--- a/.github/contributors/onlyanegg.md
+++ b/.github/contributors/onlyanegg.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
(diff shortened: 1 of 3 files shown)


ines:
Thanks for the fix and appreciate the test! 🙏

PR #4960, opened 2020-02-01
Customize entity color in displaCy

snosrap:
## Description
Currently, the color of an entity in displaCy is determined by its `label`. These colors can be overridden (at the entity label level) by setting the `colors` key in the `options` parameter for `displacy.serve`.

I would like the color of an entity to be determined by an arbitrary attribute of the `entity`. Consider negation with [negspaCy](https://spacy.io/universe/project/negspacy) -- a term like "hypertension" could be red (indicating a negated entity) in the phrase "no history of hypertension" but green (no negation) in "patient does have hypertension". Another use case would be to return a different intensity of a color, based on a confidence score.

I made a one-line change to the `parse_ents` function in `displacy/__init__.py` to add a `params` key to each entity dictionary sent to the renderer. The renderer is already looking for a params key (although I was never able to determine where it is set).

This change allows me to write the following in my code to change the color based on an entity's span._.negex value:
`spacy.tokens.Span.set_extension("params", getter=lambda span: { "bg": span._.negex and "#ff8197" or "#bfeeb7" })`

I'm totally open to feedback. This may be the wrong approach, but I thought it was an interesting use case.

### Types of change
enhancement

### DIFF
--- a/spacy/displacy/__init__.py
+++ b/spacy/displacy/__init__.py
@@ -173,7 +173,7 @@ def parse_ents(doc, options={}):
     RETURNS (dict): Generated entities keyed by text (original text) and ents.
     """
     ents = [
-        {"start": ent.start_char, "end": ent.end_char, "label": ent.label_}
+        {"start": ent.start_char, "end": ent.end_char, "label": ent.label_, "params": ent.has_extension('params') and ent._.params or {}}
         for ent in doc.ents
     ]
     if not ents:


ines:
Thanks, this is a good point!

I think we might need a more general way of handling this. Custom extension attributes are "user space" and spaCy itself should never actively set or expect them. One solution would be to allow users to pass in a color getter function that receives the `Span` and should return a color value. This would you implement colors via extension attributes, or any other custom logic. For example:

```python
get_color1 = lambda ent: ent._.entity_color
get_color2 = lambda ent: MY_COLOR_TABLE.get(ent.label_, "grey")
```

snosrap:
Hi Ines, you're definitely right -- that is a much better solution!

This new `color_getter` would ideally be set in the `options` dictionary parameter to displacy.render/serve, right? Then passed on to EntityRenderer? I'm not as familiar with DependencyRenderer to know if color would be useful there too.

I'll try to put together another pull request (with passing tests!) but I won't be offended if someone else tackles this.

ines:
Closing this for now to keep things tidy, but feel free to open a new PR if you like and are ready 🙂

Issue #4981, opened 2020-02-07
EntityLinker: linking from value of a custom extension 


kabirkhan:
## Feature description

I've been thinking about Entity Linking on custom knowledge bases that don't have as many available aliases as Wikidata. There's a bit of a gap in the current implementation where the `ent.text` is assumed to be an alias in the `KnowledgeBase` and looked up directly. While this works fine for Wikidata it would be nice to allow the user to define a custom attribute they would like the linker to use instead of just the text.

As an example, if my NER model extracts an Entity with an Alias that is slightly different from that of any in my `KnowledgeBase`, the linking step won't work. Or I'll have to add a lot of variations of the alias to my `KnowledgeBase` to make the linking step work.

For example, the below code doesn't set the kb_id for "Machine learning" because the "l" in "learning" is not capitalized.

```python
nlp = spacy.load('custom_model')
print(nlp.get_pipe('entity_linker').kb.get_alias_strings())
["Machine Learning", "Artificial Intelligence", "AI"]
# 
text = "Machine learning is a subcategory of AI."
doc = nlp(text)

print([(e.text, e.label_, e.kb_id_) for e in doc.ents])
# [('Machine learning', 'SKILL', None), ('AI', 'SKILL', 'a1')]
```

This is a pretty obvious example but you could extend it to more generic and likely cases.

For example:

```python
print(kb.get_alias_strings())
# ["bioinformatics", "research"]
```

if the NER model extracts "researched" and "bioinformatic", they cannot be properly linked.

My proposal is allowing the user to define what attribute they want to use for looking up the alias in the `kb` using the `kb.get_candidates` method. This could be exposed as a pipeline config variable similar to how the EntityRuler accepts a `phrase_matcher_attr` variable.

This would allow the user to have a custom component before entity linking that can compute some string similarity function on the entity text to all the aliases in the `kb` to get all aliases above a threshold and then run the `kb.get_candidates` method for all of them.

So when we get candidates in the `predict` method we can check the alias attribute if the user set it in the config.

Something like:
```python
...
alias = getattr(ent, self.cfg.get(alias_attr, ''))
if not alias:
    alias = getattr(ent._, self.cfg.get(alias_attr, ''), ent.text))
candidates = self.kb.get_candidates(alias)
...
```

Happy to contribute this change but I thought I'd start a discussion first in case I'm overlooking something. (I think it would work fine to keep the training code the same but not 100% sure)
[...]

PR #4988, opened 2020-02-10
[WIP] KB Registry for customized `get_candidates` function

kabirkhan:
WIP idea of allowing custom `get_candidates` function for KnowledgeBase based on this issue https://github.com/explosion/spaCy/issues/4981

@svlandeg I don't really like the idea of adding a full new registry for the kb itself.
And I'm not sold on the implementation here but this is just an idea of how to provide the custom functionality for methods of arbitrary spaCy objects. 

It would be interesting to expand the idea to a more generic solution but I assume that's a lot of what you're thinking about for spaCy v3.

Would love to discuss further

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/kb.pyx
+++ b/spacy/kb.pyx
@@ -2,7 +2,7 @@
 # cython: profile=True
 # coding: utf8
 from spacy.errors import Errors, Warnings, user_warning
-from spacy.tokens import Span
+from spacy.tokens.span import Span
 from spacy.util import registry
 
 from pathlib import Path
@@ -21,7 +21,7 @@ from libcpp.vector cimport vector
 
 
 @registry.kb.register("get_candidates")
-def get_candidates(KnowledgeBase kb, Span ent):
+def get_candidates(kb: KnowledgeBase, ent: Span):
     """Registered function to get_candidates from a `KnowledgeBase`
     given an ent from doc.ents
     """
--- a/spacy/kb.pyx
+++ b/spacy/kb.pyx
@@ -277,7 +277,6 @@ cdef class KnowledgeBase:
         and the prior probability of that alias resolving to that entity.
         If the alias is not known in the KB, and empty list is returned.
         """
-
         cdef hash_t alias_hash = self.vocab.strings[alias]
         if not alias_hash in self._alias_index:
             return []
@@ -285,11 +284,11 @@ cdef class KnowledgeBase:
         alias_entry = self._aliases_table[alias_index]
 
         return [Candidate(kb=self,
-                        entity_hash=self._entries[entry_index].entity_hash,
-                        entity_freq=self._entries[entry_index].freq,
-                        entity_vector=self._vectors_table[self._entries[entry_index].vector_index],
-                        alias_hash=alias_hash,
-                        prior_prob=prior_prob)
+                          entity_hash=self._entries[entry_index].entity_hash,
+                          entity_freq=self._entries[entry_index].freq,
+                          entity_vector=self._vectors_table[self._entries[entry_index].vector_index],
+                          alias_hash=alias_hash,
+                          prior_prob=prior_prob)
                 for (entry_index, prior_prob) in zip(alias_entry.entry_indices, alias_entry.probs)
                 if entry_index != 0]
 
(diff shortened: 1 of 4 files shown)


kabirkhan:
So I'm not in a huge hurry to get a quick fix in at the moment. This was mostly to get the conversation started. I'm more interested in aligning to the longer term vision of spaCy v3. 

I have a couple concerns with this approach currently

1. Is kb too specific a namespace for a function registry? Maybe, maybe not. This could be a slippery slope of allowing any arbitrary part of the library to be overridden. Not sure if that's good or bad but could cause some confusion for sure.
2. Is it going to slow down the current EntityLinker that doesn't implement a custom `get_candidates` function to make arbitrary call into a pure python package to check if there is a custom `get_candidates` function?

For sure this does work but yeah I can add a simple unit test for it if we do want to go down this route. I've tested it with the project I'm working on for Approximate Nearest Neighbors based Entity Linking here: https://microsoft.github.io/spacy-ann-linker/ but I'll make a simpler test case for the unit test.

svlandeg:
@honnibal @ines : would like to get your perspective on this, too!

kabirkhan:
Added a couple test cases for demonstration of this functionality.
The package for ANN linking I'm working on linked above would implement this function through entry points using this custom `get_candidates` function https://github.com/microsoft/spacy-ann-linker/blob/fe2219279b5f43638759aa11356473853494d15e/spacy_ann/__init__.py#L11

kabirkhan:
@svlandeg @honnibal @ines just pinging this issue. Would love to discuss more

svlandeg:
@kabirkhan : apologies for the late follow-up! We've been busy with various other tasks/projects but I think now would be a good time to pick this up again.

I've been exploring a bit how the NEL module could be integrated in the upcoming spaCy v.3 and how that could look like in a config file, see also PR https://github.com/explosion/spaCy/pull/5460, but it's still work in progress. Basically what I want, is that you can create a new `entity_linking` component in spaCy using some KB you specify in a config file like so:

```
[nlp.pipeline.entity_linker]
factory = "entity_linker"

[nlp.pipeline.entity_linker.kb]
@assets = "spacy.KBFromFile.v1"
nlp_path = ${nlp:name}
kb_path = "/myprojectdir/my_kb"
```

I've been looking into your ANN linker and I think the idea is really cool and useful. And I've been thinking about what the best way to deal with this, could be. I guess I'm struggling to find a final answer because there are various bits here that are so closely connected. The current `get_candidates` implementation works the way it does, because it is tightly coupled to the way the KB is implemented internally. In your extension however, you kind of take a step back and you require the aliases and entities to be defined (again) in a different format. I get why you need to do that, but I think this will feel like a redundant step for the user? Building the index and the KB should probably be happening at the same time, no?

Would it be possible to actually subclass the KB somehow and intercept the calls to `add_entity` & `add_alias` etc for this purpose? So you wouldn't have a custom `get_candidates` function but a whole custom `KnowledgeBase` object that tries to do things internally consistent & efficient?

I'm not sure - happy to discuss further!

fcggamou:
Hi, I'm actually looking for this exact feature. I'd need the EL to link entities even if they are not added as aliases in the KB with the exact same words. I'm not familiar with the inner workings of spaCy so I'm a little bit lost when looking at this pull request. Is there currently an actual way I could achieve something like this? 

Thanks a lot in advance.

kabirkhan:
@svlandeg Thanks for the response on this. I'd considered the subclass approach but didn't really like it without some way of loading a custom kb configurably. If you're planning to add the KBFromFile config approach I think that'll be perfect and won't require another change. 

@fcggamou if you're ok with some churn on development and overall API I'm working on a solution for this that we'll be using in the Microsoft Cognitive Services Entity Linking API. You can view the alpha development of that project with docs on getting started here:
https://microsoft.github.io/spacy-ann-linker/

fcggamou:
Thanks a lot @kabirkhan , this is brilliant. I will give it a shot.

svlandeg:
Thanks for starting this discussion @kabirkhan, I think having the option to swap in a different candidate generator like your [ANN linker](https://microsoft.github.io/spacy-ann-linker/) will be useful in many use-cases.

With spaCy 3 becoming more and more crystalized on the `develop` branch, I think we can now support this properly through the new config system as shown in PR #5876. Let's continue the discussion there? I'll tentavily close this PR (I didn't continue from it because there would be too many conflicts as `master` and `develop` have diverged a lot). But thanks again for your contributions, much appreciated!

PR #4990, opened 2020-02-10
Use int only in key2row for better performance

adrianeboyd:
## Description

Cast all keys and rows to `int` in `vectors.key2row` for more efficient access and serialization.

After pruning vectors, `key2row` ended up as `Dict[numpy.uint64, Union[int, numpy.int32]]` instead of `Dict[int, int]`, which is much larger when serialized with msgpack for (4x larger for English `md` vs. `lg`) and much slower to load and access.

## Checklist

### DIFF
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -283,7 +283,11 @@ cdef class Vectors:
 
         DOCS: https://spacy.io/api/vectors#add
         """
-        key = get_string_id(key)
+        # use int for all keys and rows in key2row for more efficient access
+        # and serialization
+        key = int(get_string_id(key))
+        if row is not None:
+            row = int(row)
         if row is None and key in self.key2row:
             row = self.key2row[key]
         elif row is None:


honnibal:
Does that work though? The keys really are uint64. Does Python `int` represent them faithfully?

adrianeboyd:
Yes, it's fine. From https://docs.python.org/3/library/stdtypes.html#typesnumeric: "Integers have unlimited precision." `key2row` is `Dict[int, int]` for the `lg` models with no problems.

I started looking into numpy memmaps, but then I couldn't figure out why the `md` models were taking nearly 50% longer to load than the `lg` models. It took me quite a bit of profiling to figure out that this was the main problem. With the `key2row` dict in the `md` models, things like this were also much much slower:

https://github.com/explosion/spaCy/blob/99a543367dc35b12aad00c4cd845ddd1f4870056/spacy/_ml.py#L288-L292

(And this is always run twice when a model is loaded. Do we actually need to store `lex_id` or `rank` anywhere? You could return it dynamically from `key2row` instead.)

honnibal:
That's very interesting. I don't know why the `numpy.uint64` type is so inefficient? Anyway, merging

Issue #4903, opened 2020-01-13
Multiprocessing in pipe() not working with custom attributes


BramVanroy:
I am using custom attributes and a custom pipeline for the first time, so it might very well be that there is a mistake on my part. However, the code works fine when not using the `n_process` argument.

**The problem**: using `nlp.pipe(text, n_process=2)` will throw an AttributeError complaining that I am assiging a value to an unregistered extension attribute. The error is not thrown without the `n_process` argument. 

Trace:
```bash
Process Process-1:
Traceback (most recent call last):
  File "C:\Python\Python37\Lib\multiprocessing\process.py", line 297, in _bootstrap
    self.run()
  File "C:\Python\Python37\Lib\multiprocessing\process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\bmvroy\.virtualenvs\spacy_conll-PtwHJ_vN\lib\site-packages\spacy\language.py", line 1124, in _apply_pipes
    sender.send([doc.to_bytes() for doc in docs])
  File "C:\Users\bmvroy\.virtualenvs\spacy_conll-PtwHJ_vN\lib\site-packages\spacy\language.py", line 1124, in <listcomp>
    sender.send([doc.to_bytes() for doc in docs])
  File "nn_parser.pyx", line 248, in pipe
  File "C:\Users\bmvroy\.virtualenvs\spacy_conll-PtwHJ_vN\lib\site-packages\spacy\util.py", line 481, in minibatch
    batch = list(itertools.islice(items, int(batch_size)))
  File "C:\Users\bmvroy\.virtualenvs\spacy_conll-PtwHJ_vN\lib\site-packages\spacy\language.py", line 1106, in _pipe
    doc = proc(doc, **kwargs)
  File "C:\Users\bmvroy\.PyCharm2019.2\config\scratches\scratch_33.py", line 15, in __call__
    sent._.set('my_ext', sent_ext)
  File "C:\Users\bmvroy\.virtualenvs\spacy_conll-PtwHJ_vN\lib\site-packages\spacy\tokens\underscore.py", line 71, in set
    return self.__setattr__(name, value)
  File "C:\Users\bmvroy\.virtualenvs\spacy_conll-PtwHJ_vN\lib\site-packages\spacy\tokens\underscore.py", line 63, in __setattr__
    raise AttributeError(Errors.E047.format(name=name))
[...]

PR #5006, opened 2020-02-12
load Underscore state when multiprocessing

svlandeg:
Fixes https://github.com/explosion/spaCy/issues/4903

## Description
- Define a `state` for the `Underscore` class
- Load that state when creating multiple works with `n_process`>1.

On Linux the current state is transferred as `fork` is being called (by default, can be changed), but this is not the case on Windows or macOS, where `spawn` is called instead. So it's always safer to make sure the necessary state is defined explicitely. See also https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods

If we have other global variables/state that should be retained when multiprocessing, we should make a similar fix for them.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -15,6 +15,7 @@ import multiprocessing as mp
 from itertools import chain, cycle
 
 from .tokenizer import Tokenizer
+from .tokens.underscore import Underscore
 from .vocab import Vocab
 from .lemmatizer import Lemmatizer
 from .lookups import Lookups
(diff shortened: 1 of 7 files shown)


svlandeg:
Got some interesting test failures because of this. https://github.com/explosion/spaCy/blob/master/spacy/tests/doc/test_underscore.py defines a lambda function as token_extension, which can't be pickled, causing other tests with `n_process`>1 to fail.

So, to use extension attributes and multiprocessing, the extension attributes need to be picklable.

ines:
> And tests really shouldn't be influencing eachother, but that's the global nature of this... :|

For the underscore tests module, we could use a fixture with `autouse=True` that runs around each test and cleans up the underscore space afterwards. See here:  https://stackoverflow.com/a/22638709/6400719 This also means that we can make sure that they don't actually depend on each other.

svlandeg:
@ines : now using a cleanup fixture :-)

PR #5007, opened 2020-02-12
Add warning for misaligned character offset spans

adrianeboyd:
## Description

Add warning for misaligned character offset spans

### Types of change

Enhancement.

## Checklist


adrianeboyd:
This gets extremely noisy with the example training script style because a new `GoldParse` is created each iteration.

I thought something like `SPACY_WARNING_FILTER="once"` might help filter things, but it didn't seem to do what I expected. There's not another mechanism to limit repeated warnings or cut it off at 3 of the same warning, is there?

ines:
@adrianeboyd I find Python's warnings system a bit unintuitive, but I think what might be happening is that the env var is overriding the filter on a per-warning basis (as described [here](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings)). So maybe that's why `"once"` has no effect, because it's applied every time the warning is raised?

I guess we could write our own logic for this and keep a dict of bools for each warning ID to keep track of whether we've warned before. It's not elegant, but it would let us handle the more fine-grained warning setting we want (e.g. only see each individual warning type once).

adrianeboyd:
I think I'd add a warnings filter to the example scripts as part of this. What do you think?

Issue #4829, opened 2019-12-21
please add the build dependencies back to pyproject.toml


AlJohri:
Please add the build dependencies back to pyproject.toml or remove the file altogether. I need to install the package from source and I want to do it in a single step using `pip install git+https://...`.

As you can see below, I forked the repo and added the build dependencies back in which fixed the installation via pip.

## How to reproduce the problem

FAILS:
```
.venv/bin/pip install git+https://github.com/explosion/spaCy
```

WORKS:

```bash
.venv/bin/pip install git+https://github.com/AlJohri/spaCy
```

My fork has a single commit which adds the build dependencies back to pyproject.toml: https://github.com/AlJohri/spaCy/commit/8156172ef4d14ef7bca14792a394a88fe5ab0a69

## Your Environment

`$ .venv/bin/python -m spacy info --markdown`

* **spaCy version:** 2.2.3
* **Platform:** Darwin-18.7.0-x86_64-i386-64bit
* **Python version:** 3.7.5

cc @ines

ines:
I'd be happy to remove the `pyproject.toml` alltogether for now. I'll need to find the original issue, but I think we initially added it because a user requested it for a specific use case.

But it turned out to not really be feasible to list the build dependencies in *yet another* place. A single out-of-sync version pin can easily cause a lot of headache, so ideally, we'd only want to list dependencies in the `setup.cfg` and `requirements.txt`.

AlJohri:
@ines sure I think that works. let me try removing the `pyproject.toml` file in my fork and seeing if that facilitates the direct pip install from git

b-long:
@AlJohri Any luck?

AlJohri:
sorry, not yet- will do that this weekend

AlJohri:
@ines Unfortunately, running `.venv/bin/pip install git+https://github.com/AlJohri/spaCy` doesn't install the build dependencies. I think build dependencies can only be specified using PEP 518 (using the pyproject.toml).

To get it to work _without_ the `pyproject.toml` file, I had to run the following commands. This still requires running two separate commands and doesn't inside of a single requirements.txt file.

```bash
.venv/bin/pip install -r <(curl -s https://raw.githubusercontent.com/explosion/spaCy/master/requirements.txt)
.venv/bin/pip install git+https://github.com/AlJohri/spaCy
```

[...]

PR #5008, opened 2020-02-12
Re-add pyproject.toml and add tests for dependency version consistency

svlandeg:
## Description
Fixes https://github.com/explosion/spaCy/issues/4829: bring back the build dependencies in `pyproject.toml`.

On @ines' suggestion, written a unit test that checks the versions of the libraries in `requirements.txt`, `setup.cfg` and `pyproject.toml`, and if not throws an error such as 
```
> assert (lib+v) == (lib+req_v)   # if fail: pyproject.toml & requirements.txt have conflicting versions
E  AssertionError: assert 'cymem>=2.1' == 'cymem>=2.0.2,<2.1.0'

test_requirements.py:54: AssertionError
```

The unit test is kind of ad-hoc and the parsing with regular expressions kind of crude, but I think it does the trick.

### Types of change
enhancement

## Checklist


svlandeg:
Ah, hm, `requirements.txt` is found by the CLI :|

spacy/tests/test_requirements.py
```diff
@@ -0,0 +1,59 @@
...
+    for line in lines:
+        line = line.strip()
+        if not line.startswith("#"):
+            lib, v = _parse_req(line)
+            if lib and lib not in libs_ignore_requirements:
+                req_dict[lib] = v
+
+    # check setup.cfg and compare to requirements.txt
+    # also fails when there are missing or additional libs
+    setup_file = root_dir / "setup.cfg"
```
ines: Not sure how much this helps, but Python's [configparser](https://docs.python.org/3/library/configparser.html) should be able to read the `.cfg` files. I think this might let us drop some of the hack code here?

spacy/tests/test_requirements.py
```diff
@@ -0,0 +1,59 @@
...
+    with setup_file.open() as f:
+        lines = f.readlines()
+    setup_keys = set()
+    for line in lines:
+        line = line.strip()
+        if not line.startswith("#"):
+            lib, v = _parse_req(line)
+            if lib and not lib.startswith("cupy") and lib not in libs_ignore_setup:
+                req_v = req_dict.get(lib, None)
+                assert req_v is not None    # if fail: setup.cfg contains a lib not in requirements.txt
```
ines: I guess we might as well add a message here (and the other asserts) so the output is easier to read and understand if it fails?

```suggestion
                assert req_v is not None, "in setup.cfg but not in requirements.txt: {}".format(req_v)
```

ines:
Can we add the `requirements.txt` and `pyproject.toml` to the `MANIFEST.in` so it's included in the package?

svlandeg:
> 
> Can we add the `requirements.txt` and `pyproject.toml` to the `MANIFEST.in` so it's included in the package?

Sure, if that doesn't have any other undesired effects? Will have a go. Thanks for the review!

spacy/tests/test_requirements.py
svlandeg: You mean you're not a huge fan of the hacky regular expressions? :-)

ines: Well, not if they're already built into the standard library 😛

spacy/tests/test_requirements.py
```diff
@@ -0,0 +1,59 @@
...
+from pathlib import Path
+
+
+def test_build_dependencies(en_vocab):
+    # Check that library requirements are pinned exactly the same across different setup files.
+    libs_ignore_requirements = ["pytest", "pytest-timeout", "mock", "flake8", "jsonschema"]
+    libs_ignore_setup = ["fugashi", "natto-py", "pythainlp"]
+
+    # check requirements.txt
+    root_dir = Path(__file__).parent.parent.parent
```
ines: I think if the file is included in the package, it will be at the package root, so:

```suggestion
    root_dir = Path(__file__).parent.parent
```

svlandeg: Hm. But then it wouldn't run locally ?

spacy/tests/test_requirements.py
```diff
@@ -0,0 +1,77 @@
...
+            lines = f.readlines()
+    except FileNotFoundError as e:
+        try:
+            # for local usage
+            root_dir = Path(__file__).parent.parent.parent
+            req_file = root_dir / "requirements.txt"
+            with req_file.open() as f:
+                lines = f.readlines()
+        except FileNotFoundError as e:
+            # where areth thou ?
```
ines: 😂

If it turns out we need these conditions, we could probably simplify this part and avoid the `try`/`excepts`:

```python
roots = [Path(__file__).parent.parent, Path(__file__).parent.parent.parent]  # or whatever
for root_dir in roots:
    req_file = root_dir / "requirements.txt"
    if req_file.exists():
        # read file etc.
        break
```

spacy/tests/test_requirements.py
svlandeg: I will clean this code up once we figure out the trouble with the CI ;-)

PR #5017, opened 2020-02-14
Update streamlit app

chkoar:


examples/streamlit_spacy.py
```diff
@@ -79,7 +77,7 @@ def process_text(model_name, text):
     st.header("Named Entities")
     st.sidebar.header("Named Entities")
     label_set = nlp.get_pipe("ner").labels
-    labels = st.sidebar.multiselect("Entity labels", label_set, label_set)
+    labels = st.sidebar.multiselect("Entity labels", options=label_set, default=None)
```
adrianeboyd: Streamlit doesn't handle these options symmetrically, argh. I think what we actually want is:

```
label_set = list(nlp.get_pipe("ner").labels)
labels = st.sidebar.multiselect("Entity labels", options=label_set, default=label_set)
```

chkoar: My first thought it was that we may not interested on all labels. Although streamlit needs a list not a set.

adrianeboyd: For the demo I think it should start out with all labels.

`label_set` is just a confusing variable name. `nlp.get_pipe("ner").labels` returns a tuple. `options` gets cast internally to a list but `defaults` doesn't.

chkoar: Yeap. Although, in their doc's they explicitly describe that they need a list of strings or None. I may send a PR.

adrianeboyd: Yeah, it's just kind of unexpected. (In any case, thanks for taking the time to update this!)

ines:
Thanks a lot! 👍 (Only made a few small formatting changes.)

PR #5019, opened 2020-02-14
Optimizing tokenization for Luxembourgish (dealing with apostrophe infixes)

questoph:
## Description

This update fixes some special cases for the Luxembourgish tokenizer, i.e., the correct handling of apostrophe infixes in words. Forms like "d'Saach" (definite article + word) should be split while others should remain as one token "entsche'den" (indication of long vowel in old orthography).

### Types of change

- Fix for tokenizer rules in punctuation.py
- Addition to exceptions in tokenizer_exceptions.py

## Checklist


ines:
Thanks! 👍

PR #5024, opened 2020-02-16
Fix/Improve german stop words

Jan-711:
Two stop words ("einige" and  "einigen") are sticking together.

Remove three nouns that may serve as stop words in a specific context (e.g. religious or news) but are not applicable for general use.


ines:
Thanks! 👍

PR #5030, opened 2020-02-18
Add `idx` as an attribute available via to_array

MisterKeefe:
Closes 5029. Make a token's idx property as one of the things that can be included in the output of a `Doc`'s `to_array` method. 

## Description

- Add `IDX` as an entry in `attrs.pxd` 
- Add clause in `elif` chain to make `IDX` result in `token.idx` lookup
- Add unit test

### Types of change

Minor enhancement

## Checklist

### DIFF
--- a/.github/contributors/MisterKeefe.md
+++ b/.github/contributors/MisterKeefe.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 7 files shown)


honnibal:
Looking good! The `spacy.attrs` enum is tricky but you did it correctly and added to the end, thanks!

There's one more place that needs to be changed though: the `spacy/symbols.pyx` and `spacy/symbols.pxd` files also have an enum, that needs to list the attrs. This will be cleaned up in v3, but for now we need to have it in both places.

honnibal:
Looks good, thanks!

PR #5058, opened 2020-02-24
Add missing comma in a dependency specification

bryant1410:
Conda is complaining that it can't parse that line otherwise.

- [x] My changes don't require a change to the documentation, or if they do, I've added all the required information.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,7 @@ install_requires =
 
 [options.extras_require]
 lookups =
-    spacy_lookups_data>=0.0.5<0.2.0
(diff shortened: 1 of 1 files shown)


ines:
Thanks! 👍 (And that's strange, surprised this hasn't come up before! I guess it's because we're defining the meta for our conda packages separately.)

bryant1410:
Yeah, I was wondering the same. I check pip and it works fine with it. Maybe it's because I have the option `pip_interop_enabled: true` enable in my Conda setup, so Conda reads the Pip package metadata I guess.

bryant1410:
(and the error doesn't pop up with `python setup.py check`)

PR #5060, opened 2020-02-25
update thinc 

svlandeg:
thinc from `7.4.0.dev0` to `7.4.0.dev2`


ines:
I'm just rebuilding this Thinc version with `wheelwright` with wheels for 3.5 and then upload those to PyPi. This should hopefully solve the issue.

Issue #5075, opened 2020-03-01
Checking sentence attributes of Tokens of unit test Doc causes a segfault sometimes


swartchris8:
I came across this issue while trying to write a complex spacy Doc based unit test. It seems checking the sent attribute of tokens of the doc created below causes a segfault sometimes. In my legacy spacy==2.0.12 dev env it always causes a segfault in the newer 2.2.3 only sometimes.

## How to reproduce the behaviour

Run below script 4-10 times and you should get a segfault:

```
from spacy.tests.util import get_doc
from spacy.vocab import Vocab

# "A simple method is used to compare the efficacy of ropinirole and rasagiline for Parkinson's disease."
words = ["A", "simple method"] + "is used to compare the efficacy of ropinirole and rasagiline for".split() + ["Parkinson's disease"]
vocab_words = """"A simple method is used to compare the efficacy of ropinirole and rasagiline for Parkinson's disease""".split()
vocab = Vocab(strings=vocab_words)
dependencies = [
    "det",  # A
    "nsubjpass",  # simple method
    "auxpass",  # is
    "ROOT",  # used
    "aux",  # to
    "xcomp",  # compare
    "det",  # the
    "dobj",  # efficacy
    "prep",  # of
    "pobj",  # ropinirole
    "cc",  # and
    "conj",  # rasagiline
    "prep",  # for
    "pobj",  # Parkinson's disease
]
tags = ["DT", "JJ", "VBZ", "VBN", "TO", "VB", "DT", "NN", "IN", "NN", "CC", "NN", "IN", "NNP"]
pos_tags = [
    "DET", "ADJ", "VERB", "VERB", "PART", "VERB", "DET", "NOUN", "ADP", "NOUN", "CCONJ", "NOUN", "ADP",
    "PROPN"]
heads = [1, 3, 3, 3, 5, 3, 7, 5, 7, 8, 9, 9, 9, 12]

en_doc = get_doc(vocab, words, pos_tags, heads, dependencies, tags)

for t in en_doc:
    print(t, t.sent)
```

## Your Environment

* **spaCy version:** 2.2.3
* **Platform:** Darwin-19.3.0-x86_64-i386-64bit
* **Python version:** 3.6.8
* Environment Information: Clean virtualenv with only spacy in it installed with `pip install spacy`

adrianeboyd:
[...]

PR #5079, opened 2020-03-02
Improve token head verification

adrianeboyd:
## Description

Improve the verification for valid token heads when heads are set:

* in `Token.head`: heads come from the same document
* in `Doc.from_array()`: head indices are within the bounds of the document

Addresses segfault in #5075 by making it impossible to create invalid heads in a document with `Doc.from_array()`.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -790,7 +790,7 @@ cdef class Doc:
 
         if SENT_START in attrs and HEAD in attrs:
             raise ValueError(Errors.E032)
-        cdef int i, col
+        cdef int i, col, abs_head_index
         cdef attr_id_t attr_id
         cdef TokenC* tokens = self.c
         cdef int length = len(array)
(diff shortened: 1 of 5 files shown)


adrianeboyd:
I'm not sure which head value to show in `E190`. I'm currently using the actual value from the array, but it makes the error message kind of hard to understand.

spacy/tokens/doc.pyx
```diff
@@ -804,6 +804,14 @@ cdef class Doc:
             attr_ids[i] = attr_id
         if len(array.shape) == 1:
             array = array.reshape((array.size, 1))
+        # Check that all heads are within the document bounds
+        if HEAD in attrs:
+            col = attrs.index(HEAD)
+            for i in range(length):
+                # cast index to signed int
+                abs_head_index = numpy.int32(array[i, col]) + i
```
honnibal: Is there a reason we need to do this in numpy, rather than in C? The C equivalent would be `(<int32>array[i, col]) + 1` right?

I'm not really worried about performance here but the code just looks a little odd to me.

adrianeboyd: No good reason, I'll change it.

adrianeboyd: Wait, I remember that I did try this first. With `<int32_t>array[i, col]` I get errors:

```
OverflowError: Python int too large to convert to C long
```

Am I doing something obvious wrong here? I guess the cast in `Token.set_struct_attr()` is simpler because it's just C?

honnibal: Aaah right because it's a uint64. Yeah okay let's keep it like this.

Issue #5083, opened 2020-03-02
Streaming Data Memory Growth (reprise)


dancsalo:
## How to reproduce the behaviour
According to https://github.com/explosion/spaCy/pull/1424, `nlp.vocab.strings` in the snippet below should not increase in length beyond 10k. Yet when I run the code, `nlp.vocab.strings` increases without bound, well beyond 10k.

```
from __future__ import unicode_literals
import spacy
import numpy.random

def generate_strings():
    while True:
        yield str(numpy.random.random())

def main():
    nlp = spacy.blank('xx')
    nlp.tokenizer.infix_finditer = None
    for i, doc in enumerate(nlp.pipe(generate_strings())):
        if not i % 1000:
            print(i, len(nlp.vocab.strings))

if __name__ == '__main__':
    main()
```

I'm running spaCy in a web app to process streaming data and have encountered a memory leak, similar to https://github.com/explosion/spaCy/issues/285. I believe this issue is causing it.

## Your Environment
* **spaCy version:** 2.2.3
* **Platform:** Darwin-17.7.0-x86_64-i386-64bit
* **Python version:** 3.7.5

adrianeboyd:
Thanks for the report!

First, I found a memory leak due to an extraneous iterator that should be fixed by #5101.

Second, on the `StringStore`: you're right that the `StringStore` is allowed to keep growing. Let's see, #1424 is an older version, here's the current version, which doesn't seem to work:

https://github.com/explosion/spaCy/blob/2281c4708cc3dfa68ffcdff5554c18d8fae0c9de/spacy/language.py#L807-L835

This is checking every 10K docs for older strings, not limiting the size to 10K strings. There are a couple of problems, the main one that `self.vocab._reset_cache(keys, strings)` isn't implemented, so you can't run this out-of-the-box at all. When I tried to run it without the vocab cache resetting, it was very slow and sometimes hung, so I wouldn't recommend trying to use this right now. (I think it was implemented for a much older version of spacy.)

In general, the vocab size should be limited to basically `max(size of vocab for a loaded model, 10K)`, but the `StringStore` is allowed to grow unbounded, since it's impossible to know which saved docs you may want to inspect in the future that were processed with this pipeline.

If you know for sure that you won't need to inspect older docs, you can periodically reduce the `StringStore` back to the minimum, which is: all strings required by the model + all items in the vocab. Here's an updated version of the script that keeps track of the minimum strings, with slightly more realistic test documents:

```
import spacy
import random
[...]

PR #5085, opened 2020-03-02
set vector of merged entity

### LINKED ISSUE #5082: Option to average vectors when using "merge_entity" pipeline.
Would it be possible to have an option for averaging the vectors of merge entities when using the "merge_entities" pipeline? For example, when named entities are represented as a Span in doc.ents, you can have a vector representation which is the average over the vectors in the Span. When using the merge_entities pipeline, the vector representation of the entity is a zero vector.

svlandeg:
Closes #5082 

## Description
When using the `merge_entities` pipe, the vector of the merged `token` is now appropriately set to the vector of the original `span` (i.e. the average of the original tokens). Before it was just a zero vector.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/pipeline/functions.py
+++ b/spacy/pipeline/functions.py
@@ -45,6 +45,9 @@ def merge_entities(doc):
         for ent in doc.ents:
             attrs = {"tag": ent.root.tag, "dep": ent.root.dep, "ent_type": ent.label}
             retokenizer.merge(ent, attrs=attrs)
+            # add the vector of the (merged) entity to the vocab
+            if not doc.vocab.get_vector(ent.text).any():
+                doc.vocab.set_vector(ent.text, ent.vector)
     return doc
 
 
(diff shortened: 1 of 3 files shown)


adrianeboyd:
@svlandeg: do you think this is something the retokenizer should handle more generally?

svlandeg:
Yea, perhaps? I was doubting where to put it. You think it would be better moved to the `_merge` function?

adrianeboyd:
I think it might make more sense to have it near `.tensor` here:

https://github.com/explosion/spaCy/blob/2281c4708cc3dfa68ffcdff5554c18d8fae0c9de/spacy/tokens/_retokenize.pyx#L197-L202

I was a little worried initially that it would be tricky to add new vectors if the table is full, but I see that it gets resized automatically.

One downside if it's not optional is that the size of the vectors could keep growing and growing, e.g., in a situation like #5083? I'm not sure this is a major concern, but I was just looking at all the places that can grow...

svlandeg:
Ok I moved it. The issue of the growing `Vocab` probably needs a more generic solution that is probably out-of-scope for this PR...

PR #5091, opened 2020-03-03
fix minibatch_by_words  with int

svlandeg:
## Description
Bug crept in - `minibatch_by_words` with `size` an `int` did not work on `develop` anymore, because the `else` statement would overwrite the first `if`.

### Types of change
bug fix

## Checklist


svlandeg:
I pulled this edit into https://github.com/explosion/spaCy/pull/5102, so closing this one.

PR #5099, opened 2020-03-04
Update Makefile

adrianeboyd:
## Description

Main changes:

* improve dependencies around variable filenames with timestamp file
* execute tests within venv

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,36 +1,37 @@
 SHELL := /bin/bash
-WHEELHOUSE := ./wheelhouse
 PYVER := 3.6
 VENV := ./env$(PYVER)
 
-version = $(shell "bin/get-version.sh")
+version := $(shell "bin/get-version.sh")
 
-dist/spacy-$(version).pex : wheelhouse/spacy-$(version)-*.whl
-	pex -f ./wheelhouse --no-index --disable-cache -m spacy -o dist/spacy-$(version).pex spacy==$(version) jsonschema
-	chmod a+rx dist/spacy-$(version).pex
+dist/spacy-$(version).pex : wheelhouse/spacy-$(version).stamp
+	$(VENV)/bin/pex -f ./wheelhouse --no-index --disable-cache -m spacy -o $@ spacy==$(version) jsonschema
+	chmod a+rx $@
 
 dist/pytest.pex : wheelhouse/pytest-*.whl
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,36 +1,37 @@
...
 VENV := ./env$(PYVER)
 
-version = $(shell "bin/get-version.sh")
+version := $(shell "bin/get-version.sh")
 
-dist/spacy-$(version).pex : wheelhouse/spacy-$(version)-*.whl
-	pex -f ./wheelhouse --no-index --disable-cache -m spacy -o dist/spacy-$(version).pex spacy==$(version) jsonschema
-	chmod a+rx dist/spacy-$(version).pex
+dist/spacy-$(version).pex : wheelhouse/spacy-$(version).stamp
+	$(VENV)/bin/pex -f ./wheelhouse --no-index --disable-cache -m spacy -o $@ spacy==$(version) jsonschema
```
honnibal: Ah not running pex within the virtualenv was was definitely an oversight. Nice idea with the timestamp too.

Makefile
```diff
@@ -1,36 +1,37 @@
...
+$(VENV)/bin/pex :
 	python$(PYVER) -m venv $(VENV)
 	$(VENV)/bin/python -m pip install pex wheel
 
 .PHONY : clean test
 
 test : dist/spacy-$(version).pex dist/pytest.pex
-	PEX_PATH=dist/spacy-$(version).pex ./dist/pytest.pex --pyargs spacy -x
+	( . $(VENV)/bin/activate ; \
+	PEX_PATH=dist/spacy-$(version).pex ./dist/pytest.pex --pyargs spacy -x ; )
```
honnibal: Why is this necessary? Pex shouldn't require the virtualenv to run I don't think? Is it just that it's picking the wrong Python for you?

adrianeboyd: I think the problem is that `pex` isn't necessarily installed in the local venv? I got `ModuleNotFoundError: No module named 'pex'` otherwise.

Makefile
adrianeboyd: (Officemates who are better at Makefiles are helpful.)

adrianeboyd:
One other thing that still isn't quite ideal is that `wheel` or `pip` doesn't seem to recythonize/rebuild correctly, so this is fine for a clean repo but didn't work well for me a normal working directory. Even `make clean; make` didn't seem to clean enough and rebuild enough. I had to rebuild separately with `python setup.py build_ext --inplace` and then it packaged the version I expected it to package.

PR #5102, opened 2020-03-04
Tok2Vec: extract-embed-encode

svlandeg:
## Description
Cleaning up the `Tok2Vec` code. Enable calling the `spacy.Tok2Vec.v1` architecture with `extract` - `embed` - `encode` parts to break up the functionality and allow more customizability than the big chunck `build_Tok2Vec_model` function.

Added two example configs that showcase this for `HashEmbed` as well as `CharEmbed`. The difference between / definition of "extract" and "embed" are maybe a bit fuzzy still, but the current version fits the generic  `spacy.Tok2Vec.v1` model that defines it as 
`tok2vec = extract >> with_array(embed >> encode, pad=field_size)`

~~This PR requires the presence of a `core` layer for the `Maxout` model, cf. https://github.com/explosion/thinc/pull/317 (happy to discuss alternatives!)~~ 

### Types of change
enhancement

## Checklist


svlandeg:
There's some interaction with the recently merged `senter` PR, weird. Looking into it.

PR #5107, opened 2020-03-05
Change "pass" to "continue"

tamuhey:
Probably accidentally used "pass" instead of "continue"


tamuhey:
ok, I agree with you!

PR #5108, opened 2020-03-05
Update morphologizer

adrianeboyd:
## Description

Update morphologizer:

* switch to tagger-based morphologizer
* add `Doc.is_morphed` flag
* add `pos` and `morph` scoring to `Scorer`
* modify/extend evaluate CLI to show tag / pos / morph instead of tag as "pos"
* add `morphologizer` to train CLI
* add basic initialization and overfitting morphologizer pipeline tests

### Types of change

Enhancement

## Checklist


spacy/ml/models/defaults/morphologizer_defaults.cfg
```diff
@@ -9,6 +9,6 @@ depth = 4
 embed_size = 7000
 window_size = 1
 maxout_pieces = 3
-subword_features = true
+subword_features = false
```
svlandeg: Actually, now that I think about it, this is probably wrong in the definition of `HashCharEmbedCNN`: `subword_features` should not be a parameter of [hash_charembed_cnn](https://github.com/explosion/spaCy/blob/develop/spacy/ml/models/tok2vec.py#L73 ) but instead it should call `build_Tok2Vec_model` with `subword_features=False`. Then the morphologizer and sentrec defaults can remove this line for `subword_features`. Right now it's just an empty promise, because it's being set to `False` downstream anyway.

svlandeg: (just saw the other PR - never mind about `sentrec`)

adrianeboyd:
Yeah, I didn't want to change anything on that end, just make it clearer in the definition for now. I'm not sure it makes sense to have two separate configs for this, though? It's a little weird that only the charembed config settings are optional?

svlandeg:
The reasoning is that otherwise, you'll have an architecture that always requires `nM` and `nC` even when `char_embed=False` and they're not being used, which makes things confusing all around. And I want to avoid having default settings as that got us in trouble before (defaults being set in random places).

I do think that it makes sense to further divide the `Tok2Vec` function in separate parts, giving you more control to each specific part instead of creating all these different configs, that's what PR https://github.com/explosion/spaCy/pull/5102 is about. But for now I wanted to stay as close as possible to the "original" code so that we can measure the model's performances first.

But please feel free to just go ahead and make the change to `HashCharEmbedCNN`, I think that makes sense in the short term at least ;-)

spacy/ml/models/tok2vec.py
```diff
@@ -118,7 +117,7 @@ def hash_embed_bilstm_v1(
 
 @registry.architectures.register("spacy.HashCharEmbedBiLSTM.v1")
 def hash_char_embed_bilstm_v1(
-    pretrained_vectors, width, depth, embed_size, subword_features, nM, nC, maxout_pieces
+    pretrained_vectors, width, depth, embed_size, maxout_pieces, nM=0, nC=0
```
svlandeg: I don't think we should have default values here? I thought the point was to have it all in the configuration file, so there are no "hidden" defaults set in potentially different places anymore.

spacy/ml/models/tok2vec.py
```diff
@@ -76,9 +76,8 @@ def hash_charembed_cnn(
     embed_size,
     maxout_pieces,
     window_size,
-    subword_features,
-    nM,
-    nC,
+    nM=0,
+    nC=0,
```
svlandeg: Also here I wouldn't have default values

PR #5113, opened 2020-03-06
Matcher support for Span as well as Doc

### LINKED ISSUE #5056: Matcher support for searching on a Span, as well as a Doc
## Feature description

It would be nice to have support for running pattern matching on Span, as well as Doc.
Sometimes it could be not necessary to perform the pattern searching in the whole document, but in some sentences or, more generally, spans only.

I already implemented and partially tested a solution locally, any downside I'm not aware of in terms of policies or code design constraints for this kind of feature?

Here are some of my changes in the matcher.pyx:
```
def __call__(self, doc_or_span):
        """ ... """
        if isinstance(doc_or_span, Doc):
            doc = doc_or_span
        elif isinstance(doc_or_span, Span):
            doc = doc_or_span.doc
        else:
            raise ValueError()
        
        ...

        length = (
            (doc_or_span.end - doc_or_span.start)
            if isinstance(doc_or_span, Span) else len(doc)
        )
        matches = find_matches(&self.patterns[0], self.patterns.size(), doc_or_span, length, 
                                extensions=self._extensions, predicates=self._extra_predicates)
```
where `find_matches` changes accepting `doc_or_span` and `length` as new parameters:
```
cdef find_matches(TokenPatternC** patterns, int n, doc_or_span, int length, extensions=None, predicates=tuple()):
        ...
```

paoloq:
## Description
Matcher support for searching on a Span, as well as a Doc. Fixes #5056
Missing of an appropriate error message at line 228 in `matcher.pxy`

### Types of change
Enhancement/feature for the Matcher.

## Checklist


spacy/matcher/matcher.pyx
```diff
@@ -211,22 +212,31 @@ cdef class Matcher:
...
             raise ValueError(Errors.E155.format())
         if DEP in self._seen_attrs and not doc.is_parsed:
             raise ValueError(Errors.E156.format())
-        matches = find_matches(&self.patterns[0], self.patterns.size(), doc,
-                               extensions=self._extensions,
-                               predicates=self._extra_predicates)
+        length = (
+            (doc_or_span.end - doc_or_span.start)
+            if isinstance(doc_or_span, Span) else len(doc)
+        )
```
svlandeg: As there's already an if-else structure with `isinstance` a few lines up, I would define `length` there as well

paoloq: Makes sense!

spacy/tests/matcher/test_matcher_api.py
```diff
@@ -458,3 +458,11 @@ def test_matcher_callback(en_vocab):
...
     matches = matcher(doc)
     mock.assert_called_once_with(matcher, doc, 0, matches)
+
+
+def test_matcher_span(matcher):
+    text = "JavaScript is good but Python is better"
+    doc = Doc(matcher.vocab, words=text.split())
+    span = Span(doc, 0, 3)
+    matches = matcher(span)
+    assert len(matches) == 1
```
svlandeg: I wonder whether we shouldn't just do `matcher(span.as_doc())` to support this use-case, then we wouldn't need to accomodate for `Span`'s specifically throughout the `Matcher` code?

Also, as a minor comment, this unit test would make more sense if the full `doc` would match more than once, e.g. by changing the sentence into "JavaScript is good but Java is better" --> full `doc` matches twice, the `span` only once.

paoloq: I see your point. Using `span.as_doc()` is a good solution for sure, but when scaling on the number of spans that is a heavier operation than using the Span created from slicing directly. I also feel a little redundancy in a use-case like matching really many spans from the same `Doc`.
Anyway, I'll adapt my use-cases to the current `matcher` if it's not seen a real need for this kind of feature.

I'll fix the tests in any case, thanks for your hint.

svlandeg: Yep, I agree with the efficiency argument!

svlandeg:
Hi @paoloq : apologies for the late follow-up. Could you merge the latest version of `master` into this branch and resolve the conflicts? Your contributor agreement is there now twice, and a different PR made some edits to `test_matcher_api`: just need to make sure that both PRs are kept intact.

Let me know if you need help merging!

.github/contributors/paoloq.md
```diff
@@ -5,7 +5,11 @@ This spaCy Contributor Agreement (**"SCA"**) is based on the
 The SCA applies to any contribution that you make to any product or project
 managed by us (the **"project"**), and sets out the intellectual property rights
 you grant to us in the contributed materials. The term **"us"** shall mean
+<<<<<<< HEAD
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+=======
 [ExplosionAI UG (haftungsbeschränkt)](https://explosion.ai/legal). The term
+>>>>>>> master
 **"you"** shall mean the person or entity identified below.
```
svlandeg, line 13: I think you still need to clean these merge annotations:
```
The term **"us"** shall mean
[ExplosionAI GmbH](https://explosion.ai/legal). The term
**"you"** shall mean the person or entity identified below.
```

paoloq: Damn it!

svlandeg:
Thanks a lot for your contribution! It's now merged, and will be included in the next release of spaCy, probably 2.3 (or otherwise 3.0) 👍

PR #5121, opened 2020-03-09
Revert token_match priority changes from #4374 and extend token match options

### LINKED ISSUE #5122: Tokenizer default URL pattern may be slow

Tokenizer may be extremely slow due to positive lookbehind. See example in #4362.

Tokenizer performance could be improved by including with multiple `token_match` options, one with the highest priority and one that allows prefixes and suffixes.

adrianeboyd:
## Description

* Revert changes to priority of `token_match` so that it has priority over all other tokenizer patterns as in <v2.2.2

* Add `url_match` option for URL pattern without slow lookbehind

* Expand character classes in URL pattern to improve matching around lookaheads and lookbehinds related to #4882

* Revert changes to Hungarian tokenizer

* Update `tokenizer.explain()` and docs accordingly

Fixes #5122.

### Types of change

Bug fix / enhancement.

## Checklist


adrianeboyd:
Updated with implementation to address both the token_match priority and the slow URL pattern issue. I don't see a way around two options to handle both problems, but maybe there's an alternative I haven't thought of?

API is still temporary and clearly terrible, suggestions are welcome!

PR #5127, opened 2020-03-09
is_XXX is True if doc is empty

### LINKED ISSUE #5117: doc.is_nered returns True for empty doc
```
nlp = spacy.load('en')
empty_doc = nlp("")
print(list(empty_doc.ents)) # --> Returns []
empty_doc.is_nered   # --> Returns True
```

According to **is_nered** definition, atleast 1 token must have a named entity tag set. **doc.is_nered** should therefore return False for an empty doc.

```
def is_nered(self):
        """Check if the document has named entities set. Will return True if
        *any* of the tokens has a named entity tag set (even if the others are
        unknown values).
        """
        if len(self) == 0:
            return True         # <----- SHOULD BE FALSE
        for i in range(self.length):
            if self.c[i].ent_iob != 0:
                return True
        return False
```

## Your Environment

* **spaCy version:** 2.2.2
* **Platform:** Windows-10-10.0.17134-SP0
* **Python version:** 3.8.0
* **Models:** en

svlandeg:
## Description
Update documentation to reflect the decision that `is_tagged`, `is_nered` etc are `True` if the document is empty.
Closes #5117

### Types of change
change to the documentation

## Checklist


Tiljander:
Great, thanks for the clarity!

PR #5197, opened 2020-03-24
Describing priority rules for overlapping matches

Tiljander:
## Description
Describing priority rules for overlapping matches in API and usage section.

### Types of change
Change to the documentation

## Checklist

### DIFF
--- a/website/docs/api/entityruler.md
+++ b/website/docs/api/entityruler.md
@@ -83,7 +83,8 @@ Find matches in the `Doc` and add them to the `doc.ents`. Typically, this
 happens automatically after the component has been added to the pipeline using
 [`nlp.add_pipe`](/api/language#add_pipe). If the entity ruler was initialized
 with `overwrite_ents=True`, existing entities will be replaced if they overlap
-with the matches.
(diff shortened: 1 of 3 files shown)


website/docs/api/entityruler.md
```diff
@@ -83,7 +83,8 @@ Find matches in the `Doc` and add them to the `doc.ents`. Typically, this
 happens automatically after the component has been added to the pipeline using
 [`nlp.add_pipe`](/api/language#add_pipe). If the entity ruler was initialized
 with `overwrite_ents=True`, existing entities will be replaced if they overlap
-with the matches.
+with the matches. When matches overlap in a Doc, entity ruler prioritize longer
```
svlandeg: I would write "the entity rules prioritizes"

ines: Like this?

```suggestion
with the matches. When matches overlap in a Doc, the entity ruler prioritizes longer
```

Tiljander: Yes that is better! 👍

PR #5207, opened 2020-03-25
Port over changes: master -> develop

ines:
## Description

Library compiles and tests pass, so I hope I didn't mess anything up or port over things that shouldn't have been ported 🤞

### Types of change
I guess we need a label for stuff like this? Code organisation, tidying up etc.?

## Checklist


adrianeboyd:
I worry a bit about losing the granularity of the original commits?

ines:
@adrianeboyd I thought GitHub would port them over if I do it as a merge but... apparently not? 😞 I guess I could try again and do it as a pull request and then resolve the merge conflicts again?

svlandeg:
Ah that's a good point. I usually work my way back through source code -> blame -> find commit -> PR number -> check what happened there...

PR #5213, opened 2020-03-26
Try master -> develop sync again (part 2)

ines:
## Description
See #5207.

### Types of change
meta

## Checklist


spacy/tests/regression/test_issue4725.py
```diff
@@ -0,0 +1,25 @@
+# coding: utf8
+from __future__ import unicode_literals
+
```
svlandeg: I think we can drop these two lines on `develop`, right? (same comment for all the new code files in this PR)

ines: Ah, didn't realise that so many of those were left. Fixing.

PR #5216, opened 2020-03-27
Modify Vector.resize to work with cupy and improve resizing

adrianeboyd:
## Description

Modify `Vectors.resize` to work with cupy. Modify behavior when resizing to a different vector dimension so that individual vectors are truncated or extended with zeros instead of having the original values filled into the new shape without regard for the original axes.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/vocab_vectors/test_vectors.py
+++ b/spacy/tests/vocab_vectors/test_vectors.py
@@ -84,7 +84,6 @@ def test_init_vectors_with_resize_shape(strings, resize_data):
 
 def test_init_vectors_with_resize_data(data, resize_data):
     v = Vectors(data=data)
-    print("test", data.shape, resize_data.shape)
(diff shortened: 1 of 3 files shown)


spacy/tests/vocab_vectors/test_vectors.py
```diff
@@ -84,22 +84,34 @@ def test_init_vectors_with_resize_shape(strings, resize_data):
 
 def test_init_vectors_with_resize_data(data, resize_data):
     v = Vectors(data=data)
+    print("test", data.shape, resize_data.shape)
```
honnibal:
```suggestion
```

adrianeboyd:
Ack, I added the WIP tag because I didn't want it merged yet. The in-place resizing still does the wrong thing with a different `shape[1]`.

PR #5234, opened 2020-03-31
Remove duplicated branch in if/else-if statement

leicmi:
## Description

This will remove a duplicated if/else-if branch.

### Types of change
Refactor/Quickfix

## Checklist


svlandeg:
Nice catch! :-)

PR #5236, opened 2020-03-31
fix NEL overfitting test for GPU

svlandeg:
## Description
Ran [`test_overfitting_IO`](https://github.com/explosion/spaCy/blob/develop/spacy/tests/pipeline/test_entity_linker.py#L264) in `test_entity_linker.py` with `require_gpu`.
It looks like only this one little change is needed to make it run properly?

@honnibal : I'm a little surprised by this. Considering [your original comment](https://github.com/explosion/spaCy/pull/3864#issuecomment-504370194) I was assuming I'd need to do more work. But perhaps I still should, to make it more efficient on GPU? Or do you think it's fine if it runs (which it does) ?

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/pipeline/pipes.pyx
+++ b/spacy/pipeline/pipes.pyx
@@ -1443,7 +1443,7 @@ class EntityLinker(Pipe):
                                     scores = prior_probs + sims - (prior_probs*sims)
 
                                 # TODO: thresholding
-                                best_index = scores.argmax()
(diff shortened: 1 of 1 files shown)


honnibal:
@svlandeg Yes I do think we'll have efficiency problems with the code as it is, but I'll go ahead and merge this.

PR #5238, opened 2020-03-31
Reduce stored lexemes data, move feats to lookups

adrianeboyd:
## Description

* Move non-derivable lexemes features (`norm / cluster / prob`) to `spacy-lookups-data` as lookups
  * Get/set `norm` in both lookups and `LexemeC`, serialize in lookups
  * Remove `cluster` and `prob` from `LexemesC`, get/set/serialize in lookups only
* Remove serialization of lexemes data as `vocab/lexemes.bin`
  * Remove `SerializedLexemeC`
  * Remove `Lexeme.to_bytes/from_bytes`
* Modify normalization exception loading:
  * Always create `Vocab.lookups` table `lexeme_norm` for normalization exceptions
  * Load base exceptions from `lang.norm_exceptions`, but load language-specific exceptions from lookups
  * Set `lex_attr_getter[NORM]` including new lookups table in `BaseDefaults.create_vocab()` and when deserializing `Vocab`
* Remove all cached lexemes when deserializing vocab to override existing normalizations with the new normalizations (as a replacement for the former step that replaced all lexemes data with the deserialized data)

### Types of change

Enhancement

## Checklist


adrianeboyd:
Sample lookups data for English here: https://github.com/adrianeboyd/spacy-lookups-data/tree/feature/lexeme-demo-data

honnibal:
Looks good! Go ahead and merge it if it's safe.

adrianeboyd:
Good! I think it needs to be coordinated with a release of `spacy-lookups-data` that moves all the features there, at least all the norm exceptions.

The remaining problem is that `spacy.blank()` (or similar) would then load all the prob and cluster data by default and this isn't great. I'm not sure how best to handle this?

adrianeboyd:
The norm exceptions and their test have been moved to spacy-lookups-data.

How to handle prob/cluster is still outstanding, but I think once there's an updated release of spacy-lookups-data, this will be ready to go.

I've bumped the version for spacy-lookups-data, but since the CI doesn't include optional packages, I think the CI won't test those dependencies.

(Whoops, I realized I was thinking the dependency was the other way around. The spacy tests will be fine. The spacy-lookups-data tests will require an updated version of spacy to pass.)

PR #5251, opened 2020-04-03
Add Doc init from list of words and text

adrianeboyd:
## Description

Add an option to initialize a `Doc` from a text and list of words where the words may or may not include all whitespace tokens. If the text and words are mismatched, raise an error.

This makes it possible to address #4791 and to create docs more easily from external tokenizers that do not preserve whitespace.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -555,6 +555,8 @@ class Errors(object):
     E193 = ("Unable to resize vectors in place if the resized vector dimension "
             "({new_dim}) is not the same as the current vector dimension "
             "({curr_dim}).")
+    E194 = ("Unable to create Doc from mismatched text '{text}' and words "
+            "'{words}'")
 
 
 @add_codes
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -555,8 +555,7 @@ class Errors(object):
     E193 = ("Unable to resize vectors in place if the resized vector dimension "
             "({new_dim}) is not the same as the current vector dimension "
             "({curr_dim}).")
-    E194 = ("Unable to create Doc from mismatched text '{text}' and words "
-            "'{words}'")
+    E194 = ("Unable to aligned mismatched text '{text}' and words '{words}'.")
 
 
 @add_codes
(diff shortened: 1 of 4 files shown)


honnibal:
I'm nervous about the API design here: it looks weird to have `Doc(vocab, text=text)` and then that has to raise an error if `words` isn't provided. Perhaps we could make this a function or a factory method instead?

I would probably just have a function that gets the spaces from the `(words, texts)` pair. So you do something like `Doc(vocab, words=words, spaces=get_spaces(words, text))`

adrianeboyd:
I can see that it could be a confusing API change. I just want the functionality somewhere and don't really have any strong opinions about where.

I think it is a slight issue that `words` could contain partial or non-standard space tokens, so I think a function should return `words` and `spaces`, but then it becomes kind of hard to name, like: `util.words_and_spaces_from_words_and_text()`. Or the `align_spaces` function just doesn't handle input with space tokens in `words`? I don't really have a good answer...

adrianeboyd:
How about something like this?

honnibal:
Code looks good! Regarding the name, maybe `util.get_words_and_spaces` is slightly better?

danielwachter:
That is super useful, thanks @adrianeboyd ! However it would be great if `util.get_words_and_spaces` could also return a mapping that allows to easily align the external tokenizer tokens with the `Doc` tokens after initialising the `Doc`? This would be super useful e.g. for external joint tokenizers and sentencizers or other taggers, that don't allow to split the tokenization from the tagging.

Unfortunately that seems a bit tricky to me to add, since you remove the possible whitespace tokens of `words` in the first place, what makes it difficult to map the tokenizer whitespace tokens to `Doc` whitespace tokens afterwards. Could you please have a look if you could maybe integrate such a mapping? Thanks!

adrianeboyd:
@danielwachter This method was only intended to produce words/spaces with the whitespace representations typically used by spacy and I'd like to keep it focused on just this task, since it will affect the speed of the tokenization step in the components it was developed for (currently only for the Chinese tokenizer, but it might make sense to use it for all the languages using external tokenizers).

For aligning tokenizations more generally, I'd recommend a library like this one: https://github.com/tamuhey/tokenizations. I'm not sure exactly how it handles whitespace and whitespace tokens, but even if it's not exactly what you're looking for, it should definitely point you in the right direction in terms of the kinds of algorithms that handle this.

Issue #5230, opened 2020-03-30
Warnings when saving a spacy model during unittest


lfiedler:
When saving a model during unit testing with `unittest` I get warnings on unclosed files.
This happens with custom models as well as with models downloaded from public repository.
```python
import spacy
from tempfile import TemporaryDirectory
from unittest import TestCase

class TestSaveSpacyModel(TestCase):
    def test_save(self):
        nlp = spacy.load("de_core_news_sm")
        with TemporaryDirectory() as temp_dir:
            nlp.to_disk(temp_dir)
```
Running with unittest results in:
```
Process finished with exit code 0
C:\***\lib\site-packages\spacy\language.py:900: ResourceWarning: unclosed file <_io.TextIOWrapper name='C:\\***\\Temp\\tmpdintxtwz\\meta.json' mode='w' encoding='cp1252'>
  srsly.json_dumps(self.meta)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
C:\***\lib\site-packages\spacy\util.py:645: ResourceWarning: unclosed file <_io.BufferedWriter name='C:\\***\\Temp\\tmpdintxtwz\\tagger\\model'>
  writer(path / key)
ResourceWarning: Enable tracemalloc to get the object allocation traceback
C:\***\lib\site-packages\spacy\util.py:645: ResourceWarning: unclosed file <_io.BufferedWriter name='C:\\***\\Temp\\tmpdintxtwz\\vocab\\vectors'>
  writer(path / key)
ResourceWarning: Enable tracemalloc to get the object allocation traceback


Ran 1 test in 1.457s

OK
```
This seems to be a bug similar as to #1706, which was related to loading models and has been fixed. However, it seems that saving a model also does not properly close files. 

Since this can be muted easily, this is not much of an issue for me. Still though, might as well report it.
```

## Your Environment
* **spaCy version:** 2.2.4
* **Platform:** Windows-10-10.0.18362-SP0
* **Python version:** 3.7.6
* Environment Information: conda 4.8.2
* models: de, custom

svlandeg:
Thanks for the report! Definitely something to look into. If you feel like helping out with a PR, that would be most welcome :-)

lfiedler:
Sure, I'll have a look.

lfiedler:
As a side note, the reason this may not have been seen yet appears to be this bug in pytest:
https://github.com/pytest-dev/pytest/issues/5676

lfiedler:
[...]

PR #5264, opened 2020-04-06
Fix ResourceWarnings during unittest

lfiedler:
This PR fixes the issue where files were not closed in a timely manner, after some objects were writing to them during tests using `unittest`  #5230 

## Description
I ran through possible sources for the `ResourceWarnings` being raised. Classes directly pointed out by the warnings were `Language`, `Tagger`, `Vectors`. There the methods `to_disk` used write functions that did not close properly after write when used within tests (`unittest` and `pytest` as it turned out). Tests were implemented by catching the warnings and checking whether they still occur.

Having had a closer look, the classes `Pipes` and `EntityLinker` seemed to have implemented similar versions of `to_disk`, so I took care of them as well.

For each case I wrote unit tests (`unittest`, `pytest`) and fixed the affected code in a minimal possible way, so that existing unit tests still run green.

Side notes:
- In `Vectors` the numpy method `save` is used to persist numpy arrays to disk. Reading the source code (numpy versions 1.18 and 1.15) this should close the file descriptor. However, after wrapping the affected line into a context manager, the `ResourceWarnings` vor `Vectors` disappeared. I left a comment about this at the corresponding line I changed.
- As mention in the discussion of #5230  `ResourceWarnings` are swallowed by `pytest`. They can be made visible in `pytest` as again simply by catching them using `warnings`. 

Tests were run using both, `pytest` and `unittest`.

Looking forward for your feedback. If there is anything that should be improved in this PR I'd be happy to do so.

Thanks for your great work!

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/pipeline/pipes.pyx
+++ b/spacy/pipeline/pipes.pyx
@@ -625,7 +625,7 @@ class Tagger(Pipe):
         serialize = OrderedDict((
             ("vocab", lambda p: self.vocab.to_disk(p)),
             ("tag_map", lambda p: srsly.write_msgpack(p, tag_map)),
-            ("model", self.model.to_disk),
+            ("model", lambda p: self.model.to_disk(p)),
             ("cfg", lambda p: srsly.write_json(p, self.cfg))
         ))
         exclude = util.get_serialization_exclude(serialize, exclude, kwargs)
@@ -1394,7 +1394,7 @@ class EntityLinker(Pipe):
         serialize["vocab"] = lambda p: self.vocab.to_disk(p)
         serialize["kb"] = lambda p: self.kb.dump(p)
         if self.model not in (None, True, False):
-            serialize["model"] = self.model.to_disk
+            serialize["model"] = lambda p: self.model.to_disk(p)
         exclude = util.get_serialization_exclude(serialize, exclude, kwargs)
         util.to_disk(path, serialize, exclude)
 
--- a/spacy/pipeline/pipes.pyx
+++ b/spacy/pipeline/pipes.pyx
@@ -202,7 +202,7 @@ class Pipe(object):
         serialize["cfg"] = lambda p: srsly.write_json(p, self.cfg)
         serialize["vocab"] = lambda p: self.vocab.to_disk(p)
         if self.model not in (None, True, False):
-            serialize["model"] = self.model.to_disk
+            serialize["model"] = lambda p: self.model.to_disk(p)
         exclude = util.get_serialization_exclude(serialize, exclude, kwargs)
         util.to_disk(path, serialize, exclude)
 
--- a/spacy/pipeline/pipes.pyx
+++ b/spacy/pipeline/pipes.pyx
@@ -202,7 +202,7 @@ class Pipe(object):
         serialize["cfg"] = lambda p: srsly.write_json(p, self.cfg)
         serialize["vocab"] = lambda p: self.vocab.to_disk(p)
         if self.model not in (None, True, False):
-            serialize["model"] = lambda p: p.open("wb").write(self.model.to_bytes())
+            serialize["model"] = self.model.to_disk
         exclude = util.get_serialization_exclude(serialize, exclude, kwargs)
         util.to_disk(path, serialize, exclude)
 
@@ -625,7 +625,7 @@ class Tagger(Pipe):
         serialize = OrderedDict((
             ("vocab", lambda p: self.vocab.to_disk(p)),
             ("tag_map", lambda p: srsly.write_msgpack(p, tag_map)),
-            ("model", lambda p: p.open("wb").write(self.model.to_bytes())),
+            ("model", self.model.to_disk),
             ("cfg", lambda p: srsly.write_json(p, self.cfg))
         ))
         exclude = util.get_serialization_exclude(serialize, exclude, kwargs)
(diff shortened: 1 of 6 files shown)


lfiedler:
Ok, the pipelines for python 3.5 both failed but from python 3.6 onwards they succeeded.

I think this is what happened: https://docs.python.org/3/library/os.path.html#os.path.exists

> Changed in version 3.6: Accepts a path-like object.

First of all the logs produced by the broken Azure Pipelines:
```
/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/spacy/tests/regression/test_issue5230.py:94: in write_obj_and_catch_warnings
    obj.to_disk(d)
pipes.pyx:1399: in spacy.pipeline.pipes.EntityLinker.to_disk
    ???
/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/site-packages/spacy/util.py:645: in to_disk
    writer(path / key)
pipes.pyx:1395: in spacy.pipeline.pipes.EntityLinker.to_disk.lambda34
    ???
kb.pyx:318: in spacy.kb.KnowledgeBase.dump
    ???
kb.pyx:449: in spacy.kb.Writer.__init__
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

path = PosixPath('/tmp/tmpiu_j7xu1/kb')

    def exists(path):
        """Test whether a path exists.  Returns False for broken symbolic links"""
        try:
>           os.stat(path)
E           TypeError: argument should be string, bytes or integer, not PosixPath

/opt/hostedtoolcache/Python/3.5.9/x64/lib/python3.5/genericpath.py:19: TypeError
```
For Windows it is almost the same message (`WindowsPath` instead of `PosixPath`).
The error was thrown during running my unit tests (both, pytest and unittest).

When looking at class `EntityLinker` this error seems to be raised when trying to dump its `KnowledgeBase` object (which happens in the line just before my edits). From the same line one can see that a Path object is passed to `KnowledgeBase.dump` which ends up in the `__init__` of some `Writer` object. This reads:
```
cdef class Writer:
    def __init__(self, object loc):
        if path.exists(loc):
            assert not path.isdir(loc), "%s is directory." % loc
        if isinstance(loc, Path):
            loc = bytes(loc)
        cdef bytes bytes_loc = loc.encode('utf8') if type(loc) == unicode else loc
        self._fp = fopen(<char*>bytes_loc, 'wb')
        if not self._fp:
            raise IOError(Errors.E146.format(path=loc))
        fseek(self._fp, 0, 0)

    def close(self):
    ...
```
I guess, swapping the first two if-statements could fix this.
Please confirm.

For the record: 
I did not try compiling spacy in my local copy for python 3.5 (it took me too long to install another VS). I.e. I did not directly check, if I could reproduce this without the changes I did. So I could as well be wrong about this.

svlandeg:
Sure, it's not always possible to test on any platform! That's why we have the CI :-)

I think swapping the two statements would make sense - feel free to try. Commiting to your original branch will trigger the CI to run the tests again.

lfiedler:
Ah ok, I see :)

svlandeg:
I've added the "Work in progress" label, so feel free to "play around" ;-)
You can ping me here once the tests succeed and you feel like you're done - then we can review :-)

lfiedler:
I think the changes can be reviewed now. These are the things I changed in addition those mentioned above:
- both `kb.Writer.__init__` and `kb.Reader.__init__` did instance checks of parameter `loc` *after* using them as input in `os.path`. As it happens, `loc` can be an instance of `Path` and this throws an exception in Python 3.5 (and even segfaults `pytest`). Fixed this by pulling the instance checks upfront in the according methods.
- added unit test for instantiating `kb.Writer`
- added unit test for loading and dumping an instance of `kb.KnowledgeBase`

Unfortunately I was not able to provide a unit test for this for `kb.Reader` that works under Windows. For some reason a file that was written to could not be released properly in order to open it with the `Reader` instance within the same scope provided by the context manager. I.e. the following did not work under Windows using Python 3.7:
```python
with make_tempdir as d:
    path = d / "test" 
    with path.open("w") as f:
        f.write("test")
    reader = Reader(path)
```
But maybe I was using it wrong. Instead I provided a unit test using a `Writer` instance (`KnowledgeBase`), hoping that this is sufficient.

spacy/pipeline/pipes.pyx
```diff
@@ -202,7 +202,7 @@ class Pipe(object):
         serialize["cfg"] = lambda p: srsly.write_json(p, self.cfg)
         serialize["vocab"] = lambda p: self.vocab.to_disk(p)
         if self.model not in (None, True, False):
-            serialize["model"] = lambda p: p.open("wb").write(self.model.to_bytes())
+            serialize["model"] = self.model.to_disk
```
svlandeg, line 205: While it's technically the same, personally I would prefer to stick to the `lambda` style of notation just for consistency (i.e. `lambda p: self.model.to_disk(p)`)

lfiedler: ok, changed it :)

lfiedler: Sure, I pushed the changes in a second commit.

PR #5282, opened 2020-04-09
Comparison predicate handling for `!=` 

### LINKED ISSUE #5281: Missing comparison predicate handling for `!=` in Matcher
```
import spacy
matcher = Matcher(nlp.vocab)
pattern = [{"LENGTH": {"!=": 2}}]
matcher.add("LENGTH_COMPARE", [pattern])
doc = Doc(nlp.vocab, words=["a", "aa", "aaa"])
matches = matcher(doc)
assert len(matches) == 2
```

## Your Environment

* **spaCy version:** 2.2.4
* **Platform:** Linux-4.19.0-6-amd64-x86_64-with-glibc2.28
* **Python version:** 3.8.1

paoloq:
## Description

Fix issue #5281.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/matcher/test_matcher_api.py
+++ b/spacy/tests/matcher/test_matcher_api.py
@@ -266,24 +266,24 @@ def test_matcher_regex_shape(en_vocab):
 
 
 @pytest.mark.parametrize(
-    "cmp, ngood, bad", 
+    "cmp, bad", 
     [
-        ("==", 1, "a aaa"),
-        ("!=", 2, "aa"),
-        (">=", 2, "a"),
-        ("<=", 2, "aaa"),
-        (">", 1, "a aa"),
-        ("<", 1, "aa aaa")
+        ("==", ["a", "aaa"]),
+        ("!=", ["aa"]),
+        (">=", ["a"]),
+        ("<=", ["aaa"]),
+        (">", ["a", "aa"]),
+        ("<", ["aa", "aaa"])
     ]
 )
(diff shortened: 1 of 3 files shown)


spacy/tests/matcher/test_matcher_api.py
```diff
@@ -265,14 +265,25 @@ def test_matcher_regex_shape(en_vocab):
...
+@pytest.mark.parametrize(
+    "cmp, ngood, bad", 
+    [
+        ("==", 1, "a aaa"),
+        ("!=", 2, "aa"),
+        (">=", 2, "a"),
+        ("<=", 2, "aaa"),
+        (">", 1, "a aa"),
+        ("<", 1, "aa aaa")
+    ]
```
svlandeg, line 277: I don't think you need `ngood` here: it's always just `3 - length(bad)`, right? Can `bad` be a list (e.g. `[a, aa]`) to make its meaning a bit more clear (and remove the `split()`)

svlandeg: I do like how you've generalized this unit test, but it took me a second to understand what was going on :-)

svlandeg:
Thanks for the PR, I think this would be a good additional feature to have!

spacy/tests/matcher/test_matcher_api.py
paoloq: Thank you for your feedback, you're definitely right on the optimisation and the ugliness!
I'll fix the test with your hints.

PR #5303, opened 2020-04-14
Use max(uint64) for OOV lexeme rank

adrianeboyd:
## Description

Use max(uint64) for OOV lexeme rank

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,6 @@ requires = [
     "cymem>=2.0.2,<2.1.0",
     "preshed>=3.0.2,<3.1.0",
     "murmurhash>=0.28.0,<1.1.0",
-    "thinc==7.4.0"
+    "thinc==7.4.0",
 ]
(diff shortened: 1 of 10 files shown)


adrianeboyd:
The tests will fail because the version of thinc isn't available.

Is there a better way to define OOV_RANK in a way that's fast in Lexeme and elsewhere?

honnibal:
I think we can do:

```
from cython.stdint cimport UINT64_MAX as `OOV_RANK`
```

We can do that in `spacy.lexeme`. We'd probably need a Python function or Python variable with the value as well to make it accessible to modules like `spacy.util`

adrianeboyd:
`util` isn't a great place to put it but nothing else seemed better and I was trying not to call a function if possible. If you want to call a function in the python parts, I think you may as well use `numpy.iinfo(numpy.uint64).max`. How about this version with no local math?

honnibal:
Okay, looks good!

PR #5315, opened 2020-04-16
Fixes  #5413

### LINKED ISSUE #5314: wikidata_pretrain_kb.py fails to process wiki articles due to change {lang}wiki-latest-pages-articles.xml.bz2
The format of {lang}wiki-latest-pages-articles.xml.bz2 has changed.

The text tag previous looked like this `<text xml:space="preserve">` but now look like this `<text bytes="{NUMBER}" xml:space="preserve">`

Thus, the function `bin.entity_linking.wikipedia_processor._process_wp_text` fails to parse any wikipedia article.

elben10:
(Fixes #5314 )

Fixes change of format `{lang}wiki-latest-pages-articles.xml.bz2` which makes the `bin.entity_linking.wikipedia_processor._process_wp_text` not parse any article.

## Description
The code removes everything within the `<text >` tag before extracting the text.

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/tests/regression/test_issue5314.py
+++ b/spacy/tests/regression/test_issue5314.py
@@ -1,0 +1,10 @@
+from bin.wiki_entity_linking.wikipedia_processor import _process_wp_text
+
+title = "Arkæologi"
+
+text = """<text bytes="11456" xml:space="preserve">[[Fil:Archäologie schichtengrabung.jpg|thumb|Arkæologisk [[udgravning]] med profil.]] '''Arkæologi''' er studiet af tidligere tiders [[menneske]]lige [[aktivitet]], primært gennem studiet af menneskets materielle levn.</text>"""
+
+
+def test_issue_xxx():
+    clean_text, entities = _process_wp_text(title, text, {})
+    assert clean_text is not None
--- a/spacy/tests/regression/test_issue5314.py
+++ b/spacy/tests/regression/test_issue5314.py
@@ -1,10 +1,18 @@
+import pytest
+
 from bin.wiki_entity_linking.wikipedia_processor import _process_wp_text
 
-title = "Arkæologi"
+old_format_text = """<text bytes="11456" xml:space="preserve">[[Fil:Archäologie schichtengrabung.jpg|thumb|Arkæologisk [[udgravning]] med profil.]] '''Arkæologi''' er studiet af tidligere tiders [[menneske]]lige [[aktivitet]], primært gennem studiet af menneskets materielle levn.</text>"""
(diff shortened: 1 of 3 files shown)


elben10:
Yes, it should works. The idea is to remove everything within the `<text>` tag. So it should work with future changes the text dump as well as with previous dumps

spacy/tests/regression/test_issue5314.py
```diff
@@ -0,0 +1,10 @@
+from bin.wiki_entity_linking.wikipedia_processor import _process_wp_text
+
+title = "Arkæologi"
+
+text = """<text bytes="11456" xml:space="preserve">[[Fil:Archäologie schichtengrabung.jpg|thumb|Arkæologisk [[udgravning]] med profil.]] '''Arkæologi''' er studiet af tidligere tiders [[menneske]]lige [[aktivitet]], primært gennem studiet af menneskets materielle levn.</text>"""
+
+
+def test_issue_xxx():
+    clean_text, entities = _process_wp_text(title, text, {})
+    assert clean_text is not None
```
svlandeg, line 10:
```suggestion
    assert clean_text.strip() == "Arkæologi er studiet af tidligere tiders menneskelige aktivitet, primært gennem studiet af menneskets materielle levn."
```

spacy/tests/regression/test_issue5314.py
```diff
@@ -0,0 +1,10 @@
+from bin.wiki_entity_linking.wikipedia_processor import _process_wp_text
+
+title = "Arkæologi"
+
+text = """<text bytes="11456" xml:space="preserve">[[Fil:Archäologie schichtengrabung.jpg|thumb|Arkæologisk [[udgravning]] med profil.]] '''Arkæologi''' er studiet af tidligere tiders [[menneske]]lige [[aktivitet]], primært gennem studiet af menneskets materielle levn.</text>"""
+
+
+def test_issue_xxx():
```
svlandeg, line 8:
```suggestion
def test_issue5314():
```

spacy/tests/regression/test_issue5314.py
```diff
@@ -0,0 +1,10 @@
+from bin.wiki_entity_linking.wikipedia_processor import _process_wp_text
+
+title = "Arkæologi"
+
+text = """<text bytes="11456" xml:space="preserve">[[Fil:Archäologie schichtengrabung.jpg|thumb|Arkæologisk [[udgravning]] med profil.]] '''Arkæologi''' er studiet af tidligere tiders [[menneske]]lige [[aktivitet]], primært gennem studiet af menneskets materielle levn.</text>"""
```
svlandeg, line 5: Actually, to ensure this works also for the previous version, you could run the tests on a few different `text` versions, by using `parameterize` and providing a list of texts. Then we can explicitely test once with and once without the `bytes` parameter. See https://github.com/explosion/spaCy/blob/e7ba0c4e76d316e795d97d5a2729b83adeb51b43/spacy/tests/regression/test_issue1501-2000.py#L115 for an example of using `parameterize`.

PR #5319, opened 2020-04-16
Amend documentation to Language.evaluate

laszabine:
### Problem 
The documentation of `Language.evaluate` was unclear to my understanding. It did not specify whether the `Doc`s and `GoldParse`s that are passed to this method should contain predictions or correct annotations. This information was only present in the documentation of `Scorer.score`. (The code example to `Language.evaluate` does not specify the correct usage either, and there are also no examples of model evaluations in the usage documentation.) It was conceivable that the model would be applied to `Doc.text` to generate the predictions, and that the `Doc` argument needed to be supplied for comparison purposes. 

### Solution
I amended the documentation of `Language.evaluate` to make it clear that the `Doc` argument should contain the predicted annotations and the `GoldParse` argument the correct annotations.

### Types of change

My changes only concern the documentation. 

## Checklist

- [ ] I ran the tests, and all new and existing tests passed. **(Not applicable)**

### DIFF
--- a/.github/contributors/laszabine.md
+++ b/.github/contributors/laszabine.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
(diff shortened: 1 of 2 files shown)


svlandeg:
Sorry that this was confusing, and thanks for contributing the PR!
Looks good to me :-)

PR #5331, opened 2020-04-20
Only run backprop once when shared tok2vec weights

honnibal:
Previously, pipelines with shared tok2vec weights would call the
tok2vec backprop callback multiple times, once for each pipeline
component. This caused errors for PyTorch, and was inefficient.

Instead, accumulate the gradient for all but one component, and just
call the callback once.


svlandeg:
Ah, was this the cause of that PyTorch error "Trying to backward through the graph a second time" ?

PR #5344, opened 2020-04-23
Prevent pip from installing spaCy on Python 3.4

michael-k:
## Description

Prevent pip from installing spaCy on Python 3.4 (which is unsupported since v2.2.0) by extending `python_requires`.

Missed in 80d554f2e2813aea41b0889b39d8f30f648af1ad

### Types of change

enhancement

## Checklist


svlandeg:
Thanks for the fix, very much appreciated !

PR #5361, opened 2020-04-27
Fix problems with lower and whitespace in variants

adrianeboyd:
## Description

* Initialize lower flag explicitly

* Handle whitespace words from GoldParse correctly when creating raw
text with orth variants

* Return the text with original casing if anything goes wrong

### Types of change

Bug fix.

## Checklist


adrianeboyd:
Brought to you by Greek NER data with numerous non-breaking whitespace tokens.

PR #5365, opened 2020-04-28
shortcut functionality `enable_only_pipes`

svlandeg:
## Description
I propose to have a new "shorthand" function 
```
with nlp.enable_only_pipes(["ner"]):
    ...
```
Instead of the long way of having to do 
```
pipe_exceptions = ["ner"]
other_pipes = [pipe for pipe in nlp.pipe_names if pipe not in pipe_exceptions]
with nlp.disable_pipes(*other_pipes):  # only train NER
    ...
```
which I think is just kind of unnecessary verbose and potentially confusing. I think the shorthand will be more clear also for example code snippets on the website etc.

If OK (and we can discuss about the exact name), I'll extend this PR by updating the docs, too.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -453,6 +453,17 @@ class Language(object):
             names = names[0]  # support list of names instead of spread
         return DisabledPipes(self, *names)
 
+    def enable_only_pipes(self, names):
+        """Disable all pipeline components except the ones given as argument.
+        If used as a context manager, the pipeline will be restored to the initial
+        state at the end of the block. Otherwise, a DisabledPipes object is returned,
+        that has a `.restore()` method you can use to undo your changes.
+
+        DOCS: https://spacy.io/api/language#enable_only_pipes
+        """
+        other_pipes = [pipe for pipe in self.pipe_names if pipe not in names]
+        return self.disable_pipes(*other_pipes)
+
     def make_doc(self, text):
         return self.tokenizer(text)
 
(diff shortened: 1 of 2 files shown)


ines:
> which I think is just kind of unnecessary verbose and potentially confusing. I think the shorthand will be more clear also for example code snippets on the website etc.

Yeah, I've actually been thinking about this as well, because the `disable_pipes` is kinda frustrating if you use it to disable _all other_ pipes.

I'm not sure if we should call this `enable_only_pipes` or `enable_pipes`, though? I do like the explicitnis of the `only`, but it is more verbose and less consistent.

honnibal:
What if we had `nlp.toggle_pipes`? It would have `enable` and `disable` keyword arguments. `disable_pipes` would then be deprecated.

I think if we have two methods they really need to have an obvious contrastive naming, and it can't be "enable" to match "disable".

svlandeg:
Ok, in that case maybe we want to put this on `develop` for 3.0?

ines:
Okay, so the syntax would then be:

```python
with nlp.toggle_pipes(enable=["ner"]):
   ...
```

```python
with nlp.toggle_pipes(disable=["parser", "tagger"]):
   ...
```

We'd then just have to decide how to handle it if users use both keyword arguments. Do we allow this and just ignore it? So we raise? Do we only raise for conflicts (e.g. `disable=["ner"], enable=["ner"]`)?

svlandeg:
If in this syntax, "enable" means "enable only this and disable everything else", then I would raise if a user specified both, and the outcome would not be exactly the same.
So if the pipeline has a tagger, parser and ner, I would raise with 
``` 
nlp.toggle_pipes(disable=["parser"], enable=["ner"])
```

svlandeg:
I'll close this here and work on a new draft for `develop`.

Issue #5137, opened 2020-03-11
cfg parameter is empty when run spacy.load()


alessandrobokan:
## How to reproduce the behaviour

My component class:

```
class MyComponent(object):
    name = "my_component"

    def __init__(self, nlp, **cfg):
        self.nlp = nlp
        self.categories = cfg.get("categories", "all_categories")  # cfg is EMPTY
```

I used entry points to add the component `my_component` to the pipeline. I updated the `spacy_factories` from `setup.py`. Then I run `python setup.py sdist`, generate the lib `en_core_web_test_sm`, and installed it by `pip install`.

However, when I run following code:

```
import spacy

nlp = spacy.load("en_core_web_test_sm", categories=["category1", "category2"])
```

the parameter `cfg` from `__init__(self, nlp, **cfg):` is EMPTY, so I can't get `categories`. 

Why does this happen? Did I do something wrong? Is this a bug? I just followed the [documentation](https://spacy.io/usage/saving-loading#advanced-cfg).

**OBS:** `**overrides` is not passed here https://github.com/explosion/spaCy/blob/master/spacy/util.py#L209

## Your Environment

* Operating System: Linux
* Python Version Used: 3.6
* spaCy Version Used: 2.2.3
* Environment Information: virtualenv

adrianeboyd:
Thanks for the report, this does look like a problem for `spacy.load()`.

@ines may have a better idea, but my first suggestion would be that `pipeline_args` in `overrides` should be used to update the `config` that's passed to the component:

```
    for name in pipeline:
        if name not in disable:
            config = meta.get("pipeline_args", {}).get(name, {})
            config.update(overrides.get("pipeline_args", {}).get(name, {})
            factory = factories.get(name, name)
            component = nlp.create_pipe(factory, config=config)
            nlp.add_pipe(component, name=name)
```

instead of:

https://github.com/explosion/spaCy/blob/26a90f011b8c21dfc06940579479aaff8006ff74/spacy/util.py#L205-L210

I think a second possibility would be to store the `overrides` in `nlp` somewhere so it's accessible to pipeline components from this point on:

https://github.com/explosion/spaCy/blob/26a90f011b8c21dfc06940579479aaff8006ff74/spacy/util.py#L197

alessandrobokan:
Hello guys, there is no answer from @adrianeboyd ... I think this issue is kinda important. It appear in the spacy documentation also.

Thanks :+1:

svlandeg:
[...]

PR #5374, opened 2020-04-29
Fix passing of component configuration

svlandeg:
Fixes Issue #5137 + unit test

## Description
As described in the [documentation](https://spacy.io/usage/saving-loading#advanced-cfg), 
```
nlp = spacy.load("en_core_snek_sm", snek_style="cute")
```
should pass on the custom configuration (`snek_style`) all the way down to the `init` `**cfg` of a custom component.

This was not actually happening though. Added a quick fix by putting `config.update(overrides)` to pass that `config` on to `create_pipe`.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tests/regression/test_issue5137.py
+++ b/spacy/tests/regression/test_issue5137.py
@@ -1,0 +1,33 @@
+import spacy
+from spacy.language import Language
+from spacy.lang.en import English
+from spacy.tests.util import make_tempdir
+
+
+def test_issue5137():
+    class MyComponent(object):
+        name = "my_component"
+
+        def __init__(self, nlp, **cfg):
+            self.nlp = nlp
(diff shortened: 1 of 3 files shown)


adrianeboyd:
I'm a little worried that this might cause problem if the overriding isn't component-specific? What if I have a setting with the same name in two components?

svlandeg:
The way it's written in the documentation, does assume the additional configuration is non-specific. The alternative would get kind of clunky I think ?

adrianeboyd:
Okay, it's probably used rarely enough that it's not a big deal either way.

honnibal:
This stuff was really messy in v2. It should be a bit more consistent in v3, but yeah there have definitely been bugs on this.

PR #5377, opened 2020-04-29
Add cuda101 and cuda102 options to setup

adrianeboyd:
## Description

Add cuda101 and cuda102 options to setup

### Types of change

Enhancement?

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,6 +72,10 @@ cuda92 =
     cupy-cuda92>=5.0.0b4,<9.0.0
 cuda100 =
     cupy-cuda100>=5.0.0b4,<9.0.0
+cuda101 =
+    cupy-cuda101>=5.0.0b4,<9.0.0
+cuda102 =
+    cupy-cuda102>=5.0.0b4,<9.0.0
 # Language tokenizers with external dependencies
 ja =
     fugashi>=0.1.3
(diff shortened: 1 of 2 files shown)


DROMZEE:
I have the WARNING

> spacy 2.2.4 does not provide the extra '102'

after this command

> pip install -U spacy[cuda102]

adrianeboyd:
v2.2.4 was released before this change was added. Installing with `pip install spacy[cuda102]` will be supported starting in the next release (most likely v2.3.0).

In the meanwhile, you can install `cupy-cuda102` directly and it should be detected automatically by spacy.

MarveM:
Hi! :) I am currently having the same issue, and have installed copy-cuda101 independently. But I still get the same error "space 2.2.4 does not provide the extra cuda101". With all other requirements already satisfied. So it doesn't seem to auto detect my copy installation? Any way I can force it to?

adrianeboyd:
You don't need to install cupy with the extra option. You can just install them separately and spacy will detect cupy if it's installed.

```
pip install spacy
pip install cupy-cuda101
```

You just need to call `spacy.require_gpu()` to switch to the GPU. If it can't enable it (because there's a problem with your CUDA or cupy setup), you'll see an error.

MarveM:
Thanks for your quick reply! Sorry for misunderstanding. spacy.require_gpu() still failed after the cup-cuda101 install! But it must be a local problem then.

PR #5378, opened 2020-04-29
Feature toggle_pipes

svlandeg:
As discussed in PR #5365

## Description
* `nlp.disable_pipes(pipes)` now becomes `nlp.toggle_pipes(disable=pipes)`
* we can also do `nlp.toggle_pipes(enable=pipes)` which is often shorter/easier in the training loop
* a custom warning is thrown when people still use `nlp.disable_pipes()`
* updated the documentation.

On an unrelated note, also removed the `bin/wiki_entity_linking` scripts (cf PR #5370) to avoid having to update scripts that will be removed anyway.

### Types of change
enhancement
 
## TODO
Ensure this doesn't disable components that enabled components rely on (like a `tok2vec` component). Will address this in separate PR.

## Checklist


adrianeboyd:
I think `toggle` is potentially a pretty confusing name because it sounds like it switches the status of everything to the opposite. I think the API `toggle(enable=, disable=)` is really unexpected. Let me see if I can come up with alternatives...

svlandeg:
I do also associate "toggle" with a binary state, but here I interpret `toggle_pipes` as you're setting each pipe to either 0 or 1, and you do so by either defining the ones (`enable`) or the zeros (`disable`). Actually, would `on` and `off` work better instead of `enable` and `disable`? Like `nlp.toggle_pipes(on="ner")` or `nlp.toggle_pipes(off=["tagger", "parser"])` ?

adrianeboyd:
I still think "toggle" means "switch current setting to opposite" when this is intended for something more like "configure-pipeline-for-updating-pipe(s)" (also a catchy name!) especially if you consider the tok2vec side effects. Hmm...

```
nlp.reduce_pipes(enable)
nlp.restrict_pipes(enable)
nlp.select_pipes(enable=, disable=)
```

`select` seems the most generic? And doesn't sound like it necessarily rules out tok2vec side effects? I'll keep thinking...

honnibal:
@ines Thoughts? I guess `select` is probably better than toggle, I do see the objection.

ines:
Ah sorry, only saw this now. Yes, I see the point and out of the options, I definitely prefer `select` 👍

svlandeg:
Ok, renamed to `select_pipes` !

PR #5436, opened 2020-05-14
Fix ErrorsWithCodes().__class__ return value

ilivans:
## Description
Fix `ErrorsWithCodes().__class__` value - it should be a class rather than a string like `"[__class__] .."` before.
It's critical for functions which try to inference the class of an object, i.e. `isinstance`
I personally encountered the issue using `rollbar` which collects all the visible variables and tries to serialize those. In case of ErrorsWithCodes() it failed during `isinstance(o, collections.abc.Mapping)` check.

I ran all the tests using Python 3.8, Ubuntu 19.04, no spacy models preinstalled.
`spacy` is installed with `pip install -e path_to_repo`.
```
FAILED spaCy/spacy/tests/lang/fa/test_noun_chunks.py::test_noun_chunks_is_parsed_fa - Failed: DID NOT RAISE <class 'ValueError'>
======================== 1 failed, 1709 passed, 890 skipped, 49 xfailed, 1 xpassed, 4 warnings in 69.88s (0:01:09) =========================
```
*The failed test case fails without my change too 

### Types of change
A bug fix

## Checklist

### DIFF
--- a/.github/contributors/ilivans.md
+++ b/.github/contributors/ilivans.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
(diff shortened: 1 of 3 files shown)


adrianeboyd:
Huh, I don't know why the CI didn't fail before on that test, very weird, but thanks for the quick catch!

PR #5449, opened 2020-05-18
Adapt parser and NER for transformers

honnibal:
The parser and NER models use a transition-based approach that's rather more intricate and indirect than the most obvious way people implement something like an NER model on top of a transformer. The most obvious way is more analogous to a tagger, with either a minimal modification to ensure valid sequences are produced, or perhaps a more involved CRF layer doing viterbi decoding with previous-tag features.

This PR adds a new component, provisionally titled `SimpleNER`, that uses this more baseline-ish strategy. The logic of applying the validity constraints is performed in a Thinc layer named `BILUO`.

Bug-fixes and some tweaks are also made to the existing parser model to make it more suitable for usage with transformers. In particular, the option to not use an output hidden layer receives bugfixes (it had been broken on develop), and instead a dimensionality reducing linear layer is added to the `tok2vec` component prior to the feature-extraction for the state.

These changes make the parser and NER faster and better for transformer models, which naturally use a wide tok2vec layer, e.g. 768 dimensions instead of the 96 or 128 dimensions we typically use in the CNN. The parser and NER use 8 and 3 features respectively, which would produce state vectors of 6144 and 2304 dimensions. These vectors are too wide, we want fewer parameters in the non-pretrained components. To address this, we simply downcast the token vectors to e.g. 128 prior to feature extraction. Now the state vectors are the same size we were using in the CNN models (1024 and 384). Dropping the last hidden layer further reduces the parameters and increases speed. I haven't tested whether this helps or hurts the model properly yet.

I haven't tested extensively, but so far the existing NER model does beat the `SimpleNER` baseline, in both speed and accuracy. I do think it makes sense: the NER model has an imitation-learning objective that's really a better way to decode than the dumb approach in the `SimpleNER` tagger.

With this approach we're now getting really good results on a multi-task transformer pipeline. With the `feature/spacy-v3` branch of the new `spacy-transformers` and the latest Thinc alpha, we get the following dev accuracies on OntoNotes 5:

| Speed (wps GPU) | TAGS | UAS | LAS | ENTS F |
| ------------------- | ------ | ---- | ---- | -------- |
| 2900                     | 97.9   | 95.9 | 94.5 | 89.6    |

These development accuracies are getting close to the highest published test accuracies on the data, e.g. Flair's OntoNotes 5 test accuracy is 89.7. The hyper-parameters aren't especially well-tuned, and the model uses `bert-base-cased`, which is definitely not the hottest transformer these days. Training was on a single GTX 2080 for about 7 hours. Full settings can be found here: https://github.com/explosion/spacy-transformers/blob/feature/spacy-v3/examples/listen/joint-core-bert.cfg

The nice thing here is that the parser, NER and tagger are all reading from and writing to a single copy of the transformer. No special tricks were required to make the multi-task learning work well.

The one thing I haven't gotten to work is the training from raw documents, which is the setting we use to produce the current core models. Nobody else trains from long documents, but I do think it ought to work. I suspect I have a bug in the alignment code or something. Anyway, if we can't get that setting to work, we'll just have to train on the gold preprocessing and use a sentence boundary detection model as a preprocess at runtime.

## Checklist

### DIFF
--- a/examples/experiments/tok2vec-ner/multihashembed_tok2vec.cfg
+++ b/examples/experiments/tok2vec-ner/multihashembed_tok2vec.cfg
@@ -4,12 +4,18 @@ limit = 0
 dropout = 0.2
 patience = 10000
 eval_frequency = 200
-scores = ["ents_f"]
+scores = ["ents_p", "ents_r", "ents_f"]
 score_weights = {"ents_f": 1}
 orth_variant_level = 0.0
 gold_preproc = true
 max_length = 0
-batch_size = 25
+
+[training.batch_size]
+@schedules = "compounding.v1"
+start = 3000
+stop = 3000
+compound = 1.001
+
 
 [optimizer]
 @optimizers = "Adam.v1"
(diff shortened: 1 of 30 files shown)


svlandeg:
The test failure due to flake etc should be fixed if this PR is merged first: https://github.com/explosion/spaCy/pull/5441

ines:
Ah, I just manually made this commit on `develop` because I couldn't find it in any of the PRs 😅

svlandeg:
Oh, my bad, should have labeled that PR with 3.0 !

Issue #4912, opened 2020-01-15
How do I train sentence splitter without training DEP parser?


buriy:
UD dep corpus is rather small, is made of single sentences per paragraph, and I'd like to train sentence splitter from DEP parser on arbitrary sentences, say, from other data sources containing sentence breaks.
How do I do it?

## Which page or section is this issue related to?
Training? Training DEP?

adrianeboyd:
For training data without real paragraphs there's a `spacy convert` option that lets you group sentences into fake paragraphs so that the parser has a chance to learn sentence boundaries:

```
spacy convert -n 10
```

We use this option for most of the UD corpora behind spacy's provided models. In theory the converter could also support the UD document and paragraph markers, but there are so many UD/CoNLL-U corpora that don't have them and it doesn't seem like something that spacy necessarily needs to support.

The next major version of spacy (v3?) will have a new component called `SentenceRecognizer` that you can train just from sentence-segmented data. Underneath it's just a tagger that tags words as either sentence start or not. It's much smaller and faster than the parser, with slightly higher precision and slightly lower recall in the tasks I've tried so far. (Prodigy will also support sentence boundary annotation as one option for gathering custom training data for this task.)

If you'd like to experiment with it, it should be working in the `develop` branch, and feedback is very welcome! If you have data in spacy's training format (you just need `"orth"` for each token at a minimum), you can train a model with `spacy train -p sentrec`. (Please keep in mind that `develop` is not going to be stable! At the very least `sentrec` is a terrible name that will change in the future...)

buriy:
That new component system (Morphologizer, SentRec) is really looking great! Thanks, I'll try!
Backporting to 2.1/2.2 also looks trivial for SentRec -- just update the .is_sent_start from this new model.

adrianeboyd:
I think that backporting is a little more challenging than it looks because of how the gold annotations are passed around as a tuple during training in v2. This has been improved a lot with a new class called `Example` in `develop`, which is why a number of new components have only been added in `develop` and there aren't plans to backport any of them.

buriy:
Well, first of all, I want to support existing v2.1/v2.2 right now, that's why considering using existing components and maybe several tricks, and backporting just could be one of them.
Can't we just add it as an additional model in the pipeline making a copy from one of POS/NER/DEP? We just need binary yes/no output per token.
And reg training backporting... you know, CLI training is broken anyway in v2.1/v2.2, so I had to write my own wrappers for almost everything.
[...]

PR #5456, opened 2020-05-19
Remove peeking from Parser.begin_training

adrianeboyd:
## Description

Inspect all instances in `Parser.begin_training` rather than only the first 1000.

(Reported in #4912.)

### Types of change

Bug fix.

## Checklist


adrianeboyd:
This may potentially both solve and cause problems for training on very large corpora...

PR #5457, opened 2020-05-19
Update website models for v2.3.0

adrianeboyd:
## Description

Update website models for v2.3.0

### Types of change

Docs.

## Checklist

### DIFF
--- a/website/meta/languages.json
+++ b/website/meta/languages.json
@@ -1,6 +1,36 @@
 {
     "languages": [
         {
+            "code": "zh",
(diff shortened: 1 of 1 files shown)


adrianeboyd:
How hard would it be to include the license and authors for individual sources? I think it should be acknowledged on the website and not just buried in `meta.json`.

ines:
FYI, created a new branch `docs/v2.3.0` that we can use to collect docs updates for v2.3. Then we only need one open PR to `master`.

PR #5462, opened 2020-05-20
Extend lemmatizer rules for all UPOS tags

### LINKED ISSUE #5459: lemmatizer does not take adverb as a parameter currently
## Feature description
Lemmatizer just takes NOUN, ADV, and VERB as the parameter. Can you please add ADVERB to it as well - https://github.com/explosion/spaCy/blob/develop/spacy/lemmatizer.py?

adrianeboyd:
## Description

Extend lemmatizer rules for all UPOS tags

Allow rules for any UPOS.

Fixes #5459.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/lemmatizer.py
+++ b/spacy/lemmatizer.py
@@ -56,6 +56,11 @@ class Lemmatizer(object):
         index_table = self.lookups.get_table("lemma_index", {})
         exc_table = self.lookups.get_table("lemma_exc", {})
         rules_table = self.lookups.get_table("lemma_rules", {})
+        if not any((index_table.get(univ_pos), exc_table.get(univ_pos), rules_table.get(univ_pos))):
+            if univ_pos == "propn":
+                return [string]
+            else:
+                return [string.lower()]
         lemmas = self.lemmatize(
             string,
             index_table.get(univ_pos, {}),
--- a/spacy/lemmatizer.py
+++ b/spacy/lemmatizer.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 from .symbols import NOUN, VERB, ADJ, PUNCT, PROPN
 from .errors import Errors
 from .lookups import Lookups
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Hmm, this is going to affect lowercasing of UPOS tags without rules, so sentence-initial tokens won't be lowercased.

The current version is way too English-specific. I'm not sure there's really a good language-independent way to handle this?

honnibal:
Definitely agree we need to do something about this, but that lower-casing problem will be bad. I also don't immediately see the best solution.

adrianeboyd:
Initially: use lowercase as the default if there aren't rule tables for that POS? This is pretty close (identical?) to what it did before.

luffycodes:
@adrianeboyd thanks for looking into it ! Just curious since the code using lemma_rules table, however, for english, rules for adverbs are not defined in https://github.com/explosion/spacy-lookups-data/blob/master/spacy_lookups_data/data/en_lemma_rules.json ! I may be wrong, and there may exist a separate table for adverb rules - like, removing "ly" from words like "similarly". thanks !

adrianeboyd:
This change would allow languages to define rules for any UPOS, but English does not currently have any adverb rules.

adrianeboyd:
I compared this proposal to the current lemmatizer on 300K tokens (from English wikipedia). I saw two minor differences related to adverbs (affecting 224 tokens)

* the adverb exceptions are actually used
* adverbs are not just lowercased by default because there are adverb exceptions (although not rules)

You can get adverbs lowercased by adding a dummy rule:

```
nlp.vocab.lookups.get_table("lemma_rules")["adv"] = [['', '']]
```

You could do this for English, Norwegian, and Dutch (the languages that have adv exceptions) and then the results would be nearly identical, just now with adverb exceptions. (I haven't actually tested with Dutch or Norwegian text at this point.)

PR #5463, opened 2020-05-20
Rename argument: doc_or_span/obj -> doclike

ines:
## Description

Mostly internals, but I don't think we want to have argument names like `doc_or_span`, because this just makes it messy. Maybe it's just personal preference but I do think it's more consistent with our overall code style.

### Types of change

code style

## Checklist

### DIFF
--- a/spacy/lang/es/syntax_iterators.py
+++ b/spacy/lang/es/syntax_iterators.py
@@ -5,8 +5,8 @@ from ...symbols import NOUN, PROPN, PRON, VERB, AUX
 from ...errors import Errors
 
 
-def noun_chunks(obj):
-    doc = obj.doc
+def noun_chunks(doclike):
+    doc = doclike.doc
 
     if not doc.is_parsed:
         raise ValueError(Errors.E029)
--- a/spacy/lang/es/syntax_iterators.py
+++ b/spacy/lang/es/syntax_iterators.py
@@ -21,7 +21,7 @@ def noun_chunks(doclike):
     np_right_deps = [doc.vocab.strings.add(label) for label in right_labels]
     stop_deps = [doc.vocab.strings.add(label) for label in stop_labels]
     token = doc[0]
-    while token and token.i < len(doc):
+    while token and token.i < len(doclike):
         if token.pos in [PROPN, NOUN, PRON]:
             left, right = noun_bounds(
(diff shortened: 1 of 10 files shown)


honnibal:
Hmm I dunno. I think the messiness is that the argument can be a doc or a span. I don't think it's an advantage to lie about that behind `obj`, which is even vaguer.

svlandeg:
FYI, I was just looking at the `noun_chunks` code which also uses `obj` for `Doc` or `Span`: https://github.com/explosion/spaCy/blob/master/spacy/lang/en/syntax_iterators.py#L8

ines:
Yeah, using `obj` has definitely been our convention for arguments that take different types of objects, like `Doc` and `Span` or `Doc`/`Token`/`Span`.

honnibal:
Hm, digging through the history, the `obj` name on that function was introduced by a user a few years ago and it's just sort of stuck around. I don't think it's the best policy. Are we doing this in lots of other places?

ines:
Yeah, we also use that convention in the underscore/custom extensions logic. Because writing `doc_or_span_or_token` would be kinda awkward. 

Not sure about `doclike`, that sounds more like some other object following the `Doc` API (like "`Path`-like" etc.).

honnibal:
In the underscore function, `obj` is kind of meaning `self`: it's the object the method is bound to, so it makes sense to give it a sort of pronominal name. That's actually an argument *against* using the name `obj` in other bits of the codebase.

And yeah that's just what's intended with `doclike`: it's indicating that the object follows the doc API, which could be a `doc` or a `span`. If we have a function that takes `token`, `span` or `doc` I think I've been calling that `tokens`, even though the plurality is a bit misleading. `tokenish` is probably better.

spacy/lang/es/syntax_iterators.py
```diff
@@ -21,7 +21,7 @@ def noun_chunks(obj):
     np_right_deps = [doc.vocab.strings.add(label) for label in right_labels]
     stop_deps = [doc.vocab.strings.add(label) for label in stop_labels]
     token = doc[0]
-    while token and token.i < len(doc):
+    while token and token.i < len(doclike):
```
ines, line 24: This is correct, right?

PR #5465, opened 2020-05-20
Disallow creation of 0-length spans

adrianeboyd:
## Description

Disallow creation of 0-length spans

0-length spans don't seem to make a lot of sense and accessing some of their attributes (like `vector` and `sentiment`) raises division by zero errors due to the length. In addition, retokenizing 0-length spans also leads to ill-formed docs.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -566,6 +566,7 @@ class Errors(object):
     E197 = ("Row out of bounds, unable to add row {row} for key {key}.")
     E198 = ("Unable to return {n} most similar vectors for the current vectors "
             "table, which contains {n_rows} vectors.")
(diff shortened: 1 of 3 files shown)


honnibal:
`list[0:0]` is valid though, and a `Doc` object can already be empty. I think we break the slice semantics too much if we don't allow 0-length spans.

adrianeboyd:
I'm not sure whether this is the right thing to do? It does kind of go against default slice behavior.

If there's some reason to keep something more like default slice behavior or to support 0-length spans, then the retokenizer needs to reject 0-length spans for merging.

adrianeboyd:
Jinx?

Issue #4497, opened 2019-10-21
token.left_edge/right_edge gets out of sync when manually modifying doc heads


jessecoleman:
I'm doing some post processing on the output from the `parse` pipe. As part of this process, I generate multiple document roots by setting `doc[i].head = doc[i]`. I would expect this to recreate the dependency tree and recompute all the associated attributes. It works as expected when calling `.subtree`, `.ancestors`, `.lefts`, and `.rights`, but `.left_edge` and `.right_edge` use the old tree structure.

## How to reproduce the behaviour
```
import spacy
en = spacy.load('en_core_web_sm')
doc = en('apple pie')
assert doc[1].left_edge == doc[0]
doc[0].head = doc[0]
assert list(doc[1].subtree) == [doc[1]] # fine
assert list(doc[1].lefts) == [] # fine
assert list(doc[0].ancestors) == [] # fine
assert doc[1].left_edge == doc[1] # AssertionError
```

## Your Environment
* Operating System: Ubuntu 16.04
* Python Version Used: 3.6.8
* spaCy Version Used: 2.2.1

jessecoleman:
Maybe related? https://github.com/explosion/spaCy/issues/251

honnibal:
Interesting, this must be quite an old bug!

I would've expected `Token.head.__set__` to call into the `set_children_from_heads` function that's in `spacy.tokens.doc`. Instead it seems to have its own code in there.

I haven't looked in detail, but my hunch is that the code in the `Token.head.__set__` assumes the parse tree is projective, and so does the wrong thing.

adrianeboyd:
I'm not sure it's related to projectivity. I think it's not dealing with adding a new root / separate tree correctly. The sent starts are also not updated correctly.

I can hackily fix this particular problem by detecting whether this is setting a new root and using `set_children_from_heads()` instead for that case, but what I don't understand yet is why replacing most of the code here with `set_children_from_heads()` doesn't work for the very non-projective test case in the test suite.

adrianeboyd:
I did figure this out, I think, in #6046, which replaces this code with `set_children_from_heads`.

I hope this is fixed in v3.0, please open a new issue if you run into problems!

PR #5470, opened 2020-05-20
Bugfix in noun chunks

### LINKED ISSUE #5393: How to integrate util.filter_spans in nlp.pipe() ?  - ValueError: [E102] Can't merge non-disjoint spans.
Hi, I've added nlp.create_pipe("merge_noun_chunks") to my nlp pipeline as described here: https://spacy.io/api/pipeline-functions. 
When I run the nlp pipeline on large amounts of text, I sometimes get the following error. (for some corpuses I get the error, for others I dont - probably depending on random sentences). 

```
ValueError: [E102] Can't merge non-disjoint spans. 'online' is already part of tokens to merge. If you want to find the longest non-overlapping spans, you can use the util.filter_spans helper:
https://spacy.io/api/top-level#util.filter_spans
```

I saw in other issues (e.g. https://github.com/explosion/spaCy/issues/3687), that this can be solved with the util.filter_spans function, but I don't understand how to integrate this helper function in an nlp.pipe pipeline. 

Thanks for your advice :) 

```
nlp.add_pipe(nlp.create_pipe("merge_noun_chunks"))
        
docs = []
for doc, context in nlp.pipe(context_tpl_lst, as_tuples=True, n_process=1):
                doc._.Date = context["Date"]
                doc._.Category = context["Category"]
                doc._.ID = context["ID"] 
                docs.append(doc)
```

(Unfortunately I can't give you a specific string or context_tpl_lst object, because I don't know which sentence in my corpus is causing the error)

## Your Environment

* **spaCy version:** 2.2.3
* **Platform:** Darwin-19.4.0-x86_64-i386-64bit
* **Python version:** 3.7.6

### LINKED ISSUE #5458: Merge_noun_chunk trying to merge disjoint spans
` There appears to be a problem where overlapping noun_chunks are being created for a particular text. This seems to be a bug within in-built functionality. Reproduction of the problem as follows:

    nlp = spacy.load("en_core_web_md")
    nlp2 =  spacy.load("en_core_web_md"

    merge_nps = nlp.create_pipe("merge_noun_chunks")
    nlp.add_pipe(merge_nps)
      
    merge_ents = nlp.create_pipe("merge_entities")
    nlp.add_pipe(merge_ents)

    text taken from: 
    https://www.americanrhetoric.com/speeches/gwbushcubaindependence100th.htm
     
    doc = nlp(text)

    error message: ValueError: [E102] Can't merge non-disjoint spans. 'markets' is already part of 
    tokens to merge. If you want to find the longest non-overlapping spans, you can use the 
    util.filter_spans helper: https://spacy.io/api/top-level#util.filter_spans
    
    for chunk in nlp(text).noun_chunks:
        if str(chunk).find("markets") != -1:
            print(chunk.start, '|', chunk)

    Output: 
    1232 | markets      # <= the start of this chunk overlaps with the next
    1231 | where markets have brought prosperity`

I should be able to temporarily resolve the problem by creating a custom add-on using ` filterspans`, but thought you would want to know this.

## Your Environment
spaCy version: 2.2.4
Platform: Windows-10-10.0.18362-SP0
Python version: 3.7.6

svlandeg:
## Description 
There was a bug in the code that was supposed to prevent overlapping noun chunks. It was checking whether a token's subtree was not yet in a chunk, and if not, would create a new chunk from `[word.left_edge.i, word.i + 1]`.

But every once in a while, depending on the model and its predicted POS tags / deps, the tokens in the `subtree` actually would be entirely different from the tokens in that span, so the check would fail. 

Replaced the code with a more straightforward check on the actual tokens in the span that are being defined as a noun chunk.

Fixes #5458
Fixes #5393

### Types of change
bug fix

## Checklist


adrianeboyd:
I will add that couldn't reproduce this with `master` with the exact same models, but I haven't figured out what might have changed. I guess it would have to be the parse trees somehow given your test case?

honnibal:
Hmm there's some nested looping here that makes me nervous. I think this is another place where we might be introducing accidentally quadratic behaviour. On long documents it could get really slow, and I think it's unnecessary?

Noun chunks are contiguous, right? Like, they have to form a span. So we only need to test the boundaries of two chunks to ensure they don't overlap. We also know that we're iterating in order. So why can't we just check `chunk[0].i > prev[-1].i`?

honnibal:
Does this work? I'm assuming that `word.left_edge.i <= word.i` and that everything in `word.rights` is indeed to the right of the word. If these aren't true that we've definitely got other problems, right?

```python


def noun_chunks(obj):
    """
    Detect base noun phrases. Works on both Doc and Span.
    """
    # It follows the logic of the noun chunks finder of English language,
    # adjusted to some Greek language special characteristics.
    # obj tag corrects some DEP tagger mistakes.
    # Further improvement of the models will eliminate the need for this tag.
    labels = ["nsubj", "obj", "iobj", "appos", "ROOT", "obl"]
    doc = obj.doc  # Ensure works on both Doc and Span.

    if not doc.is_parsed:
        raise ValueError(Errors.E029)

    np_deps = [doc.vocab.strings.add(label) for label in labels]
    conj = doc.vocab.strings.add("conj")
    nmod = doc.vocab.strings.add("nmod")
    np_label = doc.vocab.strings.add("NP")
    prev_end = 0
    for i, word in enumerate(obj):
        if word.pos not in (NOUN, PROPN, PRON):
            continue
        # Prevent nested chunks from being produced
        if word.left_edge.i <= prev_end:
            continue
        if word.dep in np_deps:
            if word.pos == NOUN:
                for potential_nmod in word.rights:
                    if potential_nmod.dep == nmod:
                        yield word.left_edge.i, potential_nmod.i + 1, np_label
                        prev_end = potential_nmod.i + 1
                        break
                else:
                    yield word.left_edge.i, word.i + 1, np_label
                    prev_end = word.i + 1
        elif word.dep == conj:
            # covers the case: έχει όμορφα και έξυπνα παιδιά
            head = word.head
            while head.dep == conj and head.head.i < head.i:
                head = head.head
            # If the head is an NP, and we're coordinated to it, we're an NP
            if head.dep in np_deps:
                yield word.left_edge.i, word.i + 1, np_label
                prev_end = word.i + 1
```

adrianeboyd:
I wish I could figure out the `set_children_from_heads` problem from #4497. Something in there with lefts/rights/edges is probably related.

honnibal:
The whole parse representation stuff went to hell when it added support for non-projective parses :(. We can't not support non-projective, but it makes life way harder.

svlandeg:
> 
> 
> I will add that couldn't reproduce this with `master` with the exact same models, but I haven't figured out what might have changed. I guess it would have to be the parse trees somehow given your test case?

I could reproduce the issue from #5458 with `en_core_web_md` and the latest code from `master` - that's what the unit test replicates. But yea it's very much dependent on the exact pos & dep tags that are being predicted.

svlandeg:
> So we only need to test the boundaries of two chunks to ensure they don't overlap. We also know that we're iterating in order. So why can't we just check `chunk[0].i > prev[-1].i`?

You're right, that's much better. And your code snippet makes sense. I'll adjust accordingly!

PR #5474, opened 2020-05-21
Docs for Chinese word segmentation

adrianeboyd:
## Description

Docs for Chinese word segmentation

### Types of change

Docs.

## Checklist


adrianeboyd:
The formatting of the sub-sub-sub-sections may not the best option?

ines:
I'll have a look at it in the browser and see if we can maybe tighten it up a bit! We also have the option to use an "accordion" (like [here](https://spacy.io/usage#troubleshooting)) and only show certain information on click, like maybe the `pkuseg` specifics.

PR #5489, opened 2020-05-22
Support training of connected components

honnibal:
During `nlp.update`, components can be passed a boolean set_annotations
to indicate whether they should assign annotations to the `Doc`. This
needs to be called if downstream components expect to use the
annotations during training, e.g. if we wanted to use tagger features in
the parser.

Components can specify their assignments and requirements, so we can
figure out which components have these inter-dependencies. After
figuring this out, we can guess whether to pass set_annotations=True.

We could also call `set_annotations=True` always, or even just have this
as the only behaviour. The downside of this is that it would require the
`Doc` objects to be created afresh to avoid problematic modifications.
One approach would be to make a fresh copy of the `Doc` objects within
`nlp.update()`, so that we can write to the objects without any
problems. If we do that, we can drop this logic and also drop the
`set_annotations` mechanism. I would be fine with that approach,
although it runs the risk of introducing some performance overhead, and
we'll have to take care to copy all extension attributes etc.

The current PR should be fine for experimenting during `spacy-nightly`. We
can decide to opt for the always-set-annotations approach prior
to release if we want.

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -545,13 +545,14 @@ class Language(object):
 
         if component_cfg is None:
             component_cfg = {}
+        component_deps = _count_pipeline_inter_dependencies(self.pipeline)
         # Determine whether component should set annotations. In theory I guess
         # we should do this by inspecting the meta? Or we could just always
         # say "yes"
-        for name, proc in self.pipeline:
+        for i, (name, proc) in enumerate(self.pipeline):
             component_cfg.setdefault(name, {})
             component_cfg[name].setdefault("drop", drop)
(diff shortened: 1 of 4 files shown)


spacy/language.py
```diff
@@ -1159,6 +1160,25 @@ def restore(self):
         self[:] = []
 
 
+def _count_pipeline_inter_dependencies(pipeline):
```
ines, line 1163: Maybe this function should live in `analysis.py`? I have a bunch of similar-ish logic in there already.

honnibal: Yeah okay, I'll move it.

PR #5516, opened 2020-05-28
Improve model version dependency handling

ines:
## Description

- [x] accept `spacy_version` in meta to define any version range
- [x] use `setuptools`/`packaging` ([relevant docs](https://packaging.pypa.io/en/latest/)) to parse version ranges so we know whether a model is compatible with the current version (or any version) of spaCy
- [x] generate default version ranges: current version that the model was created with, up to next minor
- [x] tidy up requirements
- [x] show warning in model loading helpers if loaded model is not compatible with current version (according to model's version specs)

### Open questions

- [ ] How do we include our backwards compatibility hack that also considers models specifying stuff like `>=2.2.2` as incompatible, even though they're _technically_ compatible?

#### Example warning

```
UserWarning: [W095] Model 'en_model' (0.0.0) requires spaCy >=3.1.0,<3.2.0 and 
is incompatible with the current version (3.0.0.dev9). This may lead to 
unexpected results or runtime errors. To resolve this, download a newer 
compatible model or retrain your custom model with the current spaCy version. 
For more details and available updates, run: python -m spacy validate
```

### Types of change

ehnancement

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,8 +13,11 @@ numpy>=1.15.0
 requests>=2.13.0,<3.0.0
 plac>=0.9.6,<1.2.0
 tqdm>=4.38.0,<5.0.0
-importlib_metadata>=0.20; python_version < "3.8"
 pydantic>=1.3.0,<2.0.0
+# Official Python utilities
+setuptools
+packaging
+importlib_metadata>=0.20; python_version < "3.8"
 # Development dependencies
 cython>=0.25
 pytest>=4.6.5
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -330,6 +330,16 @@ def get_model_meta(path):
     for setting in ["lang", "name", "version"]:
         if setting not in meta or not meta[setting]:
             raise ValueError(Errors.E054.format(setting=setting))
+    if "spacy_version" in meta:
+        if not is_compatible_version(about.__version__, meta["spacy_version"]):
+            warnings.warn(
+                Warnings.W095.format(
+                    model=f"{meta['lang']}_{meta['name']}",
+                    model_version=meta["version"],
+                    version=meta["spacy_version"],
+                    current=about.__version__,
+                )
+            )
     return meta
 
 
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -13,8 +13,11 @@ numpy>=1.15.0
 requests>=2.13.0,<3.0.0
 plac>=0.9.6,<1.2.0
 tqdm>=4.38.0,<5.0.0
-importlib_metadata>=0.20; python_version < "3.8"
 pydantic>=1.3.0,<2.0.0
+# Official Python utilities
+setuptools
```
honnibal, line 18: I guess we really do have to leave these unpinned upwards, but should we set a minimum lower version? I don't know what sort of ancient setuptools someone might have.

spacy/util.py
```diff
@@ -236,33 +238,46 @@ def get_package_version(name):
...
+        version or constraint are invalid.
+    """
+    # Handle cases where exact version is provided as constraint
+    if constraint[0].isdigit():
+        constraint = f"=={constraint}"
+    try:
+        spec = SpecifierSet(constraint)
+        version = Version(version)
+    except (InvalidSpecifier, InvalidVersion):
+        return None
```
honnibal, line 260: What sort of version constraints might be invalid? Trying to figure out whether `None` is better than raising.

ines: Something weirdly formatted, something that happens to not be a valid version. It's not about the version constraint, it's about whether the string is considered a valid version.

honnibal: Yeah but should we raise there? I guess if you download a model with a bad version string it's okay to let you load it.

ines: I think raising here may cause too many problem – if a version is formatted wrong, we can still load the model and we will report an incompatibility warning. I think it's better to be cautios here.

(Also, I currently have some logic in `spacy validate` that sets the version to `"n/a"` and then expects a `None` compatibility result. But of course I can adjust that if we decide we do want to raise for non-versions.)

PR #5517, opened 2020-05-29
Remove MorphAnalysis __str__ and __repr__

adrianeboyd:
## Description

The current version raises an error because `to_json()` returns a list, so remove it for v2 for now.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
In v3 this will be a UFEATS-style string.

As alternatives for now, it would be possible to return `str(self.to_json())` or some derivation of this, but I'm worried this would be more confusing than helpful since it's not really a list underneath. `list(token.morph)` is more useful, plus obviously the individual features.

PR #5521, opened 2020-05-30
fix deserialization order

svlandeg:
## Description

The pretrain command didn't work anymore when giving it a vectors model:

`spacy pretrain my_data.jsonl en_vectors_web_lg my_output -i 20`

gave a
```
  File "vectors.pyx", line 420, in spacy.vectors.Vectors.from_disk.load_keys
  File "vectors.pyx", line 309, in spacy.vectors.Vectors.add
ValueError: [E197] Row out of bounds, unable to add row 0 for key 64.
```

It looks like the order of deserialization was wrong - the vectors (`self.data`) needs to be loaded first?

This code hasn't changed in ages though, so why did it stop working? My best guess is that this edit https://github.com/explosion/spaCy/pull/5430/files#diff-2e6a697a4ce7f5268000e28254546b00L306 in a recent PR by @adrianeboyd fixed a bug that was hiding the current bug ?

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -425,9 +425,9 @@ cdef class Vectors:
                 self.data = xp.load(str(path))
 
         serializers = OrderedDict((
-            ("key2row", load_key2row),
-            ("keys", load_keys),
             ("vectors", load_vectors),
+            ("keys", load_keys),
+            ("key2row", load_key2row),
         ))
         util.from_disk(path, serializers, [])
         self._sync_unset()


honnibal:
This is a bit creepy, as I don't really get what's going on. Your change looks right though, so cautiously accept? I guess we should just keep these vectors changes in the back of our minds if weird stuff goes wrong.

adrianeboyd:
The problem is that `en_vectors_web_lg` is in the `2.1.0` format and doesn't include the vector metadata in `meta.json`, so it gets initialized as `(0,0)`. I think you're right that the other change revealed this given the fact that it uses `keys` instead of `key2row`. (Do we still want to support `keys`?)

I created an updated `en_vectors_web_lg` 2.3.0 last week (by hand) that will be published with the other 2.3.0 models.

PR #5533, opened 2020-06-02
add oversize examples before StopIteration returns

svlandeg:
## Description
~~Add `oversize` examples to the final `minibatch_by_words` result before the `StopIteration` returns.~~ 

Entire rewrite of the `minibatch_by_words` function, avoiding cases of very small batches by working with an "overflow" buffer. Added unit tests - many of which wouldn't succeed with the old code. It's not always easy to decide what the optimal solution would be, but at least it looks like we have less singletons now.

### Types of change
bug fix & enhancement

## Checklist


honnibal:
Ah there's a danger here I was hitting in the transformer training.

It should be possible to make sure no batch exceeds a certain size, to allow for more consistent memory planning. The risk with this is that we'll generate an overside batch and hit OOM. On the other hand we potentially discard data if not.

I struggled with this function and I think it should probably be rewritten, the current version is a hacked-on mess.

svlandeg:
The risk of discarding data is that if every example is larger than the batch size, you end up with an infinite loop when training.

I think that's the second time I've run into this infinite loop, so we probably need some checks & warnings, as this is not easy to debug (for users).

But yea, I get your point about preventing OOM. I'll have a look at rewriting if you want, and including appropriate warnings.

honnibal:
Well we definitely need to avoid an infinite loop, under any set of conditions. So there's definitely a bug here.

Unclear how documents longer than the batch size should be handled. One response is that it's the user's problem to set `max_length` lower than the batch size? But if a document is longer than the batch size it should definitely be the only document in the batch.

leomrocha:
Hi @honnibal @svlandeg (sorry for tagging you directly as I don't know which is the best) I'm working language statistical analysis on the UD v2.6 dataset about token and char length that might help on choosing a good compromise of transformer length (in fact I started the analysis for that). I can give you early access to my private repo and the notebook if it can help you now.
FYI I'm working on a complete analysis of the Gutenberg project (that one will take still a few weeks as I'm working on a more detailed analysis)

Please let me know if I have to delete this comment or if something of this can be useful for you

honnibal:
Hi @leomrocha ,

Thanks for the offer, although I think we'll have the transformer lengths as an option because people may have different sized GPU (and the options change a lot). There's no reason to delete your comment though :)

svlandeg:
@honnibal : I rewrote the function. I think it's more clean now? Oversized batches can be explicitely discarded by setting `discard_oversize=True`. Otherwise, they are yielded in a separate batch.

honnibal:
@svlandeg I think you've missed a push :)

spacy/util.py
```diff
@@ -656,42 +656,47 @@ def decaying(start, stop, decay):
...
-                return
-            n_words = count_words(example.doc)
-            if n_words < (batch_size + tol_size):
-                batch_size -= n_words
-                batch.append(example)
+            current_size += n_words
+        else:
+            # if the current example exceeds the batch size, it is returned separately
+            # but only if discard_oversize=False.
+            if current_size > target_size:
```
honnibal, line 684: Should this be `if n_words > target_size`?

honnibal: I think a case you might not have thought about is something like:

Target size: `1000`
Sizes: `[400, 400, 199, 3]`

We don't want to have a batch with just the `3`.

svlandeg: yea, you're right on all fronts, will fix.

svlandeg: RE the [400, 400, 199, 3]: let me add some tests.

svlandeg: I added a test. So with target size `1000` and tol `0.2`, you'd want `[400, 400, 199, 3]` to stick together in one batch (of `[4]`), but `[400, 400, 199, 3, 250]` should be two batches, with `3` in the second batch, i.e. `[3,2]`, right? Currently the algorithm does work on the example you gave (because it always includes the tolerance in the check) but it's probably not behaving as expected on the latter example (moving `3` into the first batch instead of the second). Right?

svlandeg: Note that the original implementation would have also failed on the last unit test for `[400, 400, 199, 3, 250]`, it would have also put the first 4 in the first batch.

honnibal: Ah yeah I forgot about the `tol`.

I think it's okay if the batching is greedy, i.e. I don't think it needs to move things around to make the batching as even as possible. It should fill up the current batch and then emit it.

But the problem is the batch isn't necessarily *full* when you first hit an example that doesn't fit in the current batch. And just because an example doesn't fit, doesn't mean that example should be in a batch by itself.

honnibal: I guess a fully general algorithm would look something like:

* Keep a queue of open batches, sorted inversely by length
* When you hit a new example, sort it into the largest batch it fits into
* If a batch is above the minimum size, emit it

svlandeg: Ha, shoot. I had just implemented a more straightforward solution that at least satisfies the tests (if you agree with the tests!): it keeps the examples that are in the "tolerance margin" separate, and only adds them to the (previous) batch if no new batch was started ...
The added advantage being that we don't need the sort. I feel like we may have quite some added complexity on our hands with sorting?

svlandeg: (Note that the old solution would have failed for the three lasts unit tests, it would have always taken the first 4 into the first batch)

honnibal: If you have a version that works, great! My algorithm definitely felt a bit painful.

spacy/tests/test_util.py
```diff
@@ -0,0 +1,23 @@
...
+        ([400, 400, 199, 3, 1], [5]),
+        ([400, 400, 199, 3, 200], [3, 2]),
+        ([400, 400, 199, 3, 1, 200], [3, 3]),
+    ],
+)
+def test_util_minibatch(doc_sizes, expected_batches):
+    docs = [get_doc(doc_size) for doc_size in doc_sizes]
+    examples = [Example(doc=doc) for doc in docs]
+    batches = list(minibatch_by_words(examples=examples, size=1000))
+    assert [len(batch) for batch in batches] == expected_batches
```
honnibal, line 23: Probably wise to add an extra check here:

```
tol = 0.2
max_size = 1000 + 1000 * tol
for batch in batches:
    assert len(batch) < max_size
```

honnibal:
Okay looking good! It's good to make the test that little bit tighter, but otherwise seems working (and definitely better than it was...)

spacy/tests/test_util.py
svlandeg: I've added: `assert sum([len(example.doc) for example in batch]) < max_size`

svlandeg:
I've added some more checks to cater for edge cases, if the overflowing buffer and the current example would be too large together. The code is a bit longer now and less clean, but I think it covers many cases quite well now, as shown by the tests. I don't think we can do much better without introducing additional loops for searching/sorting.

leomrocha:
Just in case, the study I was talking about is now available [here](https://leomrocha.github.io/ud_conllu_v2.6/index.html)

PR #5535, opened 2020-06-02
Add rudimentary version checks on model load

adrianeboyd:
## Description

Since v2.2 models will be loadable but with degraded performance for v2.3, add a rudimentary version check that tries to compare the major.minor version in `meta['spacy_version']` to the current spacy version. If there is a version mismatch or the meta can't be parsed to determine compatibility, show warnings.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
The test is kind of slow, though, so maybe it would be better to just call `get_model_meta` with the path?

ines:
@adrianeboyd Yeah, good point, let's do that then. We _know_ that we're calling `get_model_meta` during load (and if not, something else is fundamentally broken).

PR #5536, opened 2020-06-02
Add warning for loose version constraints

ines:
## Description

Show a warning if a model specifies a "loose" version constraint / unconstrained version / however we want to call it. Basically things like `>=3.0.0`, which means that the model will inevitably become incompatible (while still officially reporting that it is compatible). This case should only happen if a user manually edits the `meta.json` or if the version is old (v2.x). 

Wording could be improved here... communicating the point is a bit difficult.

### Example warning

```
UserWarning: [W094] Model 'en_model' (0.0.0) requires a spaCy version with very 
loose constraints: >=2.1.0. This can lead to compatibility problems as new 
spaCy versions are released, because the model may say it's compatible when it's 
not. Consider changing the "spacy_version" in your meta.json to a version range, 
for example: >=3.0.0.dev9,<3.1.0
```

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -104,6 +104,12 @@ class Warnings(object):
             "string \"Field1=Value1,Value2|Field2=Value3\".")
 
     # TODO: fix numbering after merging develop into master
+    W094 = ("Model '{model}' ({model_version}) requires a spaCy version with "
+            "very loose constraints: {version}. This can lead to compatibility "
+            "problems as new spaCy versions are released, because the model "
+            "may say it's compatible when it's not. Consider changing the "
+            '"spacy_version" in your meta.json to a version range, '
+            "for example: {example}")
     W095 = ("Model '{model}' ({model_version}) requires spaCy {version} and is "
             "incompatible with the current version ({current}). This may lead "
             "to unexpected results or runtime errors. To resolve this, "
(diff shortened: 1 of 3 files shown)


honnibal:
Maybe `no upward constraint`?

ines:
@honnibal You'd also see that warning if you don't specify a downward constraint, though. And technically `==2.0.0` isn't a true "upward constraint" either but of course valid.

honnibal:
Do we need the warning if there's no constraint downward, though? I think that's a less common problem, and it's also less problematic in general.

ines:
> Do we need the warning if there's no constraint downward, though? I think that's a less common problem, and it's also less problematic in general.

I was thinking about this but if we make this a "best practices" warning, I do think it makes sense to include versions with no downward constraints. I agree it's less common but we might as well cover it. If you specify `<=3.1.0`, that's bad and we might as well warn because your model will report it's compatible with versions when it's not.

honnibal:
Okay how about we say "under-constrained"

ines:
@honnibal Just updated the wording, does this sound okay?

honnibal:
I made a small change? I think `requires an under-constrained spaCy version` is misleading, people will think "Why does it need the version to be under-constrained?"

PR #5543, opened 2020-06-04
pretrain from config

svlandeg:
Rewriting `pretrain` to use config

## Description
* added example `pretrain` config building on the `onto-joint.defaults.cfg` one
* `pretraining` section in the config refers to `tok2vec_model` to know its config from the training part
* `init_tok2vec` and `epoch_start` were renamed to `resume_path` and `epoch_resume` for clarity
* `discard_oversize` setting in config
* make `dropout` configurable. This was previously removed (set to `None`) because it gave slightly better performance in experiments, but I still think that's weird.
* `pretraining` and `training` each can have a different `optimizer`
* the `train` command now refers to the `train-from-config` file (further WIP will be submitted as separate PR)
* prevent infinite loop in `train` when `minibatch_by_words` returns an empty generator + custom error
* removed `pretrain_textcat.py` and `Tensorizer`

### TODO
* Will do docs in different PR when CLI scripts for 3.0 have stabilized

### Questions
* The current code builds the full `nlp` training pipeline which is probably unnecessary ?
* Not entirely sure about the implementation of `build_masked_language_model` but it seems to work in preliminary experiments?

### Types of change
enhancement

## Checklist


honnibal:
Looks good!

> The current code builds the full nlp training pipeline which is probably unnecessary ?

I don't think it will matter and let's keep it simple for now. Who knows whether someone relies on some side-effect of loading a pipeline component or something.

> Not entirely sure about the implementation of build_masked_language_model but it seems to work in preliminary experiments?

I'll take a look.

spacy/ml/models/multi_task.py
```diff
@@ -24,6 +26,80 @@ def build_cloze_multi_task_model(vocab, tok2vec):
...
 
-def build_masked_language_model(*args, **kwargs):
-    # TODO cf https://github.com/explosion/spaCy/blob/2c107f02a4d60bda2440db0aad1a88cbbf4fb52d/spacy/_ml.py#L828
-    raise NotImplementedError
+def build_masked_language_model(vocab, wrapped_model, mask_prob=0.15):
+    """Convert a model into a BERT-style masked language model"""
+
+    random_words = _RandomWords(vocab)
+
+    def mlm_forward(model, docs, is_train):
```
honnibal, line 34: Yeah this seems fine. The random state in the outer closure is a bit awkward and may inconvenience someone at some point, but let's leave it alone for now.

PR #5544, opened 2020-06-04
Add Japanese Model

polm:
## Description

This adds changes so that a Japanese model can be trained. See #3756 for the full details.

The one change that isn't mainly internal is that fugashi has been replaced with SudachiPy. The training data has been created with SudachiPy, so we need to use it for consistency.

For training data, use [UD Japanese GSD](https://github.com/megagonlabs/UD_Japanese-GSD) from Megagon Labs. (I expect the updates will be merged into the official UD repo soon, though I'm not sure what the schedule for that is. Maybe @hiroshi-matsuda can clarify?) I think the JSON data in the repo should be usable as-is. 

### Types of change

This is a model addition.

## Checklist


honnibal:
Awesome, thanks a lot for this!

I felt bad that we let the previous Japanese PR sit around so long but I wasn't sure what to do on it. I think there should be no problem in merging this, and I hope we can have the model trained for v2.3. We're providing support for a lot of extra languages in that version.

PR #5546, opened 2020-06-04
Added Langauge Examples

aj-medianet:
I added language examples for Vietnamese, Czech, Irish, Icelandic, Afrikaans, Latvian and Slovenian.

## Description
I based my translations off of lang/en/examples.py and used google translate.  I was unsure if I needed to run tests for this pull request, I ran 'python3 -m pytest tests/lang' and got the same results before and after I added my examples.

### Types of change
Enhancement - adding example sentences for testing languages 

## Checklist

### DIFF
--- a/.github/contributors/aj-medianet.md
+++ b/.github/contributors/aj-medianet.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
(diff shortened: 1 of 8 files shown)


aj-medianet:
Hi @svlandeg 

Thank you for the great feedback. I’m sorry, I didn’t know we couldn’t include google translate. Unfortunately, I don’t speak any of the languages fluently and I’m a native English speaker.

Would crowd sourcing the correct translations using Reddit be acceptable? There is a translator subreddit that I think would be perfect for this situation. Please let me know if this would work and I will make a post asking for confirmations of my translations, or better translations.

Thanks!

honnibal:
> Thank you for the great feedback. I’m sorry, I didn’t know we couldn’t include google translate. 

Please stop and think carefully about this. Why would it *not* be a problem to include translated data? It's difficult for us to review the examples ourselves and so we have to trust they are correct at the point of inclusion, and other users who don't speak those languages be able to trust that the examples are correct. Anyone who wants data that may or may not be correct can fetch it from Google Translate themselves.

It's also worth considering for a moment how it would come across to speakers of these languages if the examples were simply machine translations. I think a reasonable person would conclude that this implied that we saw their language as unimportant and expected that nobody would notice the difference. It would also be reasonable for someone seeing machine translations in the examples to immediately doubt that the library could do any useful work for them in processing their language. 

> Unfortunately, I don’t speak any of the languages fluently and I’m a native English speaker. 

That's fine but it means you should not attempt to do this work. Many people are in fact native speakers of these languages, and we've been pleased to get their help with the example sentences. 

> Would crowd sourcing the correct translations using Reddit be acceptable? There is a translator subreddit that I think would be perfect for this situation. Please let me know if this would work and I will make a post asking for confirmations of my translations, or better translations.

The example sentences aren't necessarily translations, although many people are using the initial example sentences when they commit examples for their language. The examples often highlight an interesting range of constructions, and we suggest examples that don't translate neatly be excluded. So this isn't really a good task for crowd-sourcing.

PR #5547, opened 2020-06-04
Refactored util.py Registry class to use CamelCase for class name

aj-medianet:
I refactored the util.py file because the Registry class was not using CamelCase.

## Description
I used PyCharm IDE to refactor util.py and all of the references to the registry class. I ran the full test suite before and after I refactored and got the same results. 

### Types of change
Style Enhancement? It's not really a bug or a problem, it just wasn't adhering to the style guidelines of Python.

## Checklist

### DIFF
--- a/.github/contributors/aj-medianet.md
+++ b/.github/contributors/aj-medianet.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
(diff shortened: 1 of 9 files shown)


honnibal:
Naming decisions in the documented API (including spelling and casing) are always reviewed by multiple people and it is unlikely that they are accidental. This decision to make `registry` lower-cased was definitely intentional, as it operates more like a grouping than a class (people will mostly use the class attributes and not create subclasses).

In general we unlikely to accept breaking changes like this. Note that your entry in the submission form is wrong: this change *definitely* would need to be documented, and could only be included in a major version change because it's backwards incompatible.

PR #5548, opened 2020-06-05
Added Parameter to NEL to take n sentences into account

theudas:
I am dealing with a lot of short sentences. For disambiguation, it is sometimes needed to also include the neighbor sentences.

## Description
I changed the predict method of NEL to include n_sents neighbour sentences when generating the vector for the model. There was no change to the training needed, since training ignores sentence scopes.

### Types of change
Enhancement

## Checklist
- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information. (It probably needs some description.)

### DIFF
--- a/spacy/pipeline/pipes.pyx
+++ b/spacy/pipeline/pipes.pyx
@@ -1243,24 +1243,24 @@ class EntityLinker(Pipe):
                         try:
                             # find the sentence in the list of sentences.
                             sent_index = sentences.index(ent.sent)
-                            
-                            # get n previous sentences, if there are any
-                            start_sentence = max(0, sent_index - self.n_sents)
-
-                            # get n posterior sentences, or as many < n as there are
-                            end_sentence = min(len(sentences) -1, sent_index + self.n_sents)
-
-                            # get token positions
-                            start_token = sentences[start_sentence].start
-                            end_token = sentences[end_sentence].end
-
-                            # append that span as a doc to training
-                            sent_doc = doc[start_token:end_token].as_doc()
-                            sentence_docs.append(sent_doc)
 
                         except AttributeError:
                             # Catch the exception when ent.sent is None and provide a user-friendly warning
                             raise RuntimeError(Errors.E030)
+
+                        # get n previous sentences, if there are any
+                        start_sentence = max(0, sent_index - self.n_sents)
+
+                        # get n posterior sentences, or as many < n as there are
+                        end_sentence = min(len(sentences) -1, sent_index + self.n_sents)
+
+                        # get token positions
+                        start_token = sentences[start_sentence].start
+                        end_token = sentences[end_sentence].end
+
+                        # append that span as a doc to training
+                        sent_doc = doc[start_token:end_token].as_doc()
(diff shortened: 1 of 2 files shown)


svlandeg:
Thanks for your contribution! This is a great idea.

> There was no change to the training needed, since training ignores sentence scopes.

I'm not sure why you say that? I don't think the EL functionality could work if that's the case...?

In the `update` method, the code loops through the entities in the `gold.links`, and for each `ent` fetches the context with `sentence_docs.append(ent.sent.as_doc())`. So I guess you'll have to make the appropriate changes there too ?

theudas:
Good catch.

Did the changes there as well and added a minor optimization during inference.

spacy/pipeline/pipes.pyx
```diff
@@ -1228,13 +1234,30 @@ class EntityLinker(Pipe):
...
+
+                            # get token positions
+                            start_token = sentences[start_sentence].start
+                            end_token = sentences[end_sentence].end
+
+                            # append that span as a doc to training
+                            sent_doc = doc[start_token:end_token].as_doc()
+                            sentence_docs.append(sent_doc)
+
                         except AttributeError:
```
honnibal, line 1261: Can this `except` block be moved above? Ideally I would want:

```
try: 
    sent = ent.sent
except AttributeError:
    ...
```

try/except blocks should be scoped as narrowly as possible, ideally over a single statement so that they don't catch accidentally.

spacy/pipeline/pipes.pyx
```diff
@@ -1228,13 +1234,30 @@ class EntityLinker(Pipe):
...
+
                 ent = ents_by_offset[(start, end)]
 
                 for kb_id, value in kb_dict.items():
                     # Currently only training on the positive instances
                     if value:
                         try:
-                            sentence_docs.append(ent.sent.as_doc())
+                            # find the sentence in the list of sentences.
+                            sent_index = sentences.index(ent.sent)
```
honnibal, line 1245: The try/except block here should be changed so that it only covers the `ent.sent` part, like this:

```
try: 
    sent = ent.sent
except AttributeError:
    ...
```

We want try/except blocks to be scoped as narrowly as possible, usually over only one statement, because we want the exception to apply to a specific error. If we have a whole block under the `try`, there are lots of places that an `AttributeError` could be raised, so we don't know whether our error message is correct.

svlandeg: Yea, point taken @honnibal, my bad. 
@theudas, could you make those changes in this PR as well? Then we can kill two birds with one stone ;-)

honnibal:
Thanks for this, it looks good! Just made a small style comment :)

theudas:
To be honest I don't like that except block as well.
It could be better to throw a warning, and don't append that sentence to training.
I adjusted the try except block to only include ent.sent.

svlandeg:
I do think we need that custom error there, and not just a warning and skip, because it shouldn't really ever happen that some `ent` doesn't have an associated sentence. The main reason it does happen if there is no sentence segmentation at all in the pipeline, in which case all examples will fail and there's no point continuing.

PR #5554, opened 2020-06-06
fix japanese tokenizer

hiroshi-matsuda-rit:
Fix a misalignment bug and split mode setting from meta.json

## Description

This bug was triggered by space tokens.
I wrote a specialized version of `spacy.util.get_words_and_spaces()` to align the words and tokenizer outputs.

To ensure version compatibility in future design changes, we need to set SudachiPy split_mode from meta.json. I added an argument to create_tokenizer() to receive the split mode setting described in meta.json

I ran `spacy.test.lang.ja`, `spacy train` and `spacy evaluate` to make sure that this new token alignment logic works correctly. And I set "split_mode": "C" in "tokenizer" section in meta.json and tested this setting logic by running train and evaluate with the model which contains this modified meta.json.

### Types of change

- bug fix
- improve model version compatibility

## Checklist


adrianeboyd:
Let's split this into two PRs, one for the alignment and one for the tokenizer settings.

The tokenizer needs to save the split mode setting as part of the model or unexpected things will happen when models are saved and reloaded. The setup for Chinese is very similar, so I'll remove the settings part from this PR and open a second one with the tokenizer settings + serialization similar to Chinese.

hiroshi-matsuda-rit:
Okay. I'm going to revert the previous commit.

adrianeboyd:
How does the dictionary loading with split mode A at the very top interact with the split mode setting when the tokenizer is called?

hiroshi-matsuda-rit:
> How does the dictionary loading with split mode A at the very top interact with the split mode setting when the tokenizer is called?

It's nice idea.
We can avoid the serialization problems by removing a member holding a value coming from the supplemental library .
Even it is not formal but the users are still able to get split_mode by `nlp.tokenizer.tokenizer._mode`.
I'd like to submit a PR with this initial split_mode setting strategy.

hiroshi-matsuda-rit:
By the way, this branch became less readable.
I'd like to re-submit two PRs separately.
Now I close this issue.

adrianeboyd:
Could you send me an example that gets segmented differently in all three modes?

hiroshi-matsuda-rit:
I'd like to send the examples in #5560. @adrianeboyd

PR #5555, opened 2020-06-06
Improve spacy.gold (no GoldParse, no json format!)

honnibal:
~I'm doing some exploratory work on improving the `GoldCorpus` stuff. The first thing I tried was to break it up into a subpackage to make it a bit easier to work with. I'm still exploring other changes but if we want we could merge this first. It shouldn't affect anything, but I think it will decrease compile times a bit.~

Edit: Okay this turned into bigger changes! But very nice...

* Use the `Doc` object for annotations. This means the `Example` class just holds two `Doc` objects and an alignment.
* Remove the `GoldParse` class. Equivalent functionality is now just internals inside the parser transition systems.
* Train from the `DocBin` binary format. This is smaller, faster, and removes the need for the JSON format.
* Update the converters to be able to output `DocBin`.
* Simplify the `GoldCorpus` class and rename to just `Corpus`.
* Improve performance of the `doc.from_array` method, to further increase data reading speed.
* Update and improve efficiency of the parser oracles

There are still some things left to do.

* [x] Fix segfault in parser training when `gold_preproc`
* [x] Restore the `gold_preproc` functionality, if we want that.
* [x] Finish updating the NER oracle
* [x] Finish updating tests
* [x] Double-check NER alignment and training
* [ ] Restore the data augmentation functionality

## Checklist

### DIFF
--- a/examples/experiments/onto-joint/defaults.cfg
+++ b/examples/experiments/onto-joint/defaults.cfg
@@ -5,7 +5,7 @@
 # data is passed in sentence-by-sentence via some prior preprocessing.
 gold_preproc = false
 # Limitations on training document length or number of examples.
-max_length = 0
+max_length = 5000
 limit = 0
 # Data augmentation
 orth_variant_level = 0.0
@@ -14,7 +14,7 @@ dropout = 0.1
 patience = 1600
 max_epochs = 0
 max_steps = 20000
-eval_frequency = 400
+eval_frequency = 50
 # Other settings
 seed = 0
 accumulate_gradient = 1
@@ -57,8 +57,6 @@ vectors = null
 [nlp.pipeline.tok2vec]
 factory = "tok2vec"
 
-[nlp.pipeline.senter]
-factory = "senter"
 
 [nlp.pipeline.ner]
 factory = "ner"
@@ -73,16 +71,9 @@ factory = "tagger"
 [nlp.pipeline.parser]
 factory = "parser"
 learn_tokens = false
-min_action_freq = 1
+min_action_freq = 30
 beam_width = 1
 beam_update_prob = 1.0
-
-[nlp.pipeline.senter.model]
-@architectures = "spacy.Tagger.v1"
-
-[nlp.pipeline.senter.model.tok2vec]
-@architectures = "spacy.Tok2VecTensors.v1"
-width = ${nlp.pipeline.tok2vec.model:width}
 
 [nlp.pipeline.tagger.model]
 @architectures = "spacy.Tagger.v1"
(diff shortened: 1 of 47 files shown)


adrianeboyd:
I really don't think `GoldParse` should be under `spacy.syntax`. It could stand to be renamed, but I think it really belongs under `spacy.gold`?

honnibal:
I want to try to get it out of the API and make it just something that the parser and NER use internally. In the meantime it does import from `spacy.gold` so it can be found where it used to be.

But my idea is that if we just add let the `Example` class have the alignment, and add a method `example.get_aligned(field)` to get aligned data for a single field, we'll avoid a lot of problems. Each component could control how it wants to adjust the gold-standard for stuff like misalignments, and if you only have a text categorizer the alignment won't need to be calculated at all.

We'd also eliminate all the mess from the `GoldCorpus`. The `GoldCorpus` will just return `Example` objects, and it'll be up to the components to decide what they want to do.

honnibal:
I suppose you're right that in the absence of other changes it should be under `spacy.gold`

honnibal:
The idea now is to use the `Doc` class for the annotations as well, and store the alignment with it in the `Example` class. This will avoid the whole json format nastiness.

spacy/tokens/token.pyx
```diff
@@ -787,8 +791,7 @@ cdef class Token:
 
         RETURNS (str): IOB code of named entity tag.
         """
-        iob_strings = ("", "I", "O", "B")
-        return iob_strings[self.c.ent_iob]
+        return self.iob_strings()[self.c.ent_iob]
```
honnibal, line 794: This stuff is kind of frustrating every time. Maybe we should have a `token.iob_tag` attribute?

svlandeg: I'm not sure I understand what you mean? `token.ent_iob_` does give the IOB tag. But I guess the question is, who should be responsible of making the conversion from int to string?

spacy/tokens/_serialize.py
```diff
@@ -9,6 +9,19 @@
...
 
+ALL_ATTRS = (
+    "ORTH",
+    "TAG",
+    "HEAD",
+    "DEP",
+    "SENT_START",
+    "ENT_IOB",
+    "ENT_TYPE",
+    "LEMMA",
```
adrianeboyd: Add "POS" here, too.

adrianeboyd: Nevermind...

spacy/tokens/token.pyx
honnibal: I just meant it's annoying to assemble the tag from the IOB component and the ent_type component. It'd be kind of nice to have an attribute that just did it.

PR #5560, opened 2020-06-08
Fix Japanese Tokenizer (revised)

hiroshi-matsuda-rit:
Fix token misalignment bug

## Description

This bug was triggered by space tokens.
I wrote a specialized version of `spacy.util.get_words_and_spaces()` to align the words and tokenizer outputs.

I ran `spacy.test.lang.ja`, `spacy train` and `spacy evaluate` to make sure that this new token alignment logic works correctly.

### Types of change

- bug fix

## Checklist


adrianeboyd:
Thanks for this, especially for responding so quickly after the bug was discovered!

PR #5561, opened 2020-06-08
set split_mode from meta.json

hiroshi-matsuda-rit:
Set split_mode setting from meta.json

## Description

Apply "split_mode" entry under "tokenizer" in meta.json to set SudachiPy mode from model side to ensure the backward version compatibility. I added an argument to create_tokenizer() to receive the split mode setting described in meta.json

I ran `spacy.test.lang.ja`.
And I put "split_mode": "A" or "B" or "C" in "tokenizer" section in meta.json and tested this setting by running `spacy evaluate` with this modified model.
Also empty "tokenizer" entry and no "tokenizer" entry are tested.

(I submitted hiroshi-matsuda-rit.md in another PR. https://github.com/explosion/spaCy/pull/5560/files)

### Types of change

- improve model version compatibility

## Checklist
- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #5562, opened 2020-06-08
Update Japanese tokenizer config and add serialization

adrianeboyd:
## Description

* Use `config` dict for tokenizer settings
* Add serialization of split mode setting
* Add tests for tokenizer split modes and serialization of split mode setting

Based on #5561

### Types of change

Enhancement.

## Checklist


hiroshi-matsuda-rit:
I think this works fine and thank you for adding test cases.

PR #5573, opened 2020-06-11
Japanese model: add user_dict entries and small refactor

hiroshi-matsuda-rit:
Add user_dict entries (reading_forms, inflections, sub_tokens) and improving the code readability around the token alignment procedure.

## Description
- I added and reorganized the user_dict fields which are commonly used in Japanese NLP applications.
- I deleted separate_sentences() because we should use parser's sentencizer.
- I added the test cases which covers new user_dict entries and executed `py.test spacy/tests/lang/ja/test_tokenizer.py` and found no error.
- I refactored the codes and wrote comments around _get_dtokens() and get_dtokens_and_spaces() because they were complicated.

### Types of change
- enhancement

## Checklist


adrianeboyd:
Thanks for this contribution! In case there are any minor differences between these updates and the current version (which we've already used to train models), I would like to hold off on merging these updates for 2.3.0, but they can be included in the next releases (which would be 2.3.1 and 3.0.0).

hiroshi-matsuda-rit:
@adrianeboyd Fortunately, there is no difference in the Doc contents except the `user_data[]`. We do not need to retrain the model. I think we can merge this to v 2.3.0 and the users can use these features from the beginning of official Japanese model. Could you please consider merging to v2.3.0 again?

hiroshi-matsuda-rit:
In addition, @polm -san is still improving the SudachiPy.
I'd like to ask whether we need to retrain the model or not if we upgrade SudachiPy to v0.4.6.

polm:
Thanks for the ping. 

My work on Sudachi shouldn't change the output in any way so shouldn't require retraining the model. However, the ellipsis-related issue that came up while we were working on the Japanese model had a fix released around the same time. I think the fix won't change the actual tokens spaCy sees (since it discards empty tokens), but I'm not sure if tokenizations changes are possible. If there were changes, I would assume they would only be with rare/weird tokens. 

@sorami might know more about what the possible issues with the ellipsis change. What version of SudachiPy was used for training the model?

sorami:
The "ellipsis change" was included in [SudachiPy v.0.4.5] (released Tuesday last week).

There will be no change in the tokenization analysis result; Only the order of empty tokens will change.

adrianeboyd:
The models were trained with sudachipy v0.4.5 and this is the minimum required version listed in the `spacy[ja]` requirements.

This refactor is too large for me to feel comfortable including it in spacy v2.3.0. The models have already been trained and released and we're in the final stages of preparing the spacy library release, so we don't want to merge any more significant changes at this point.

I would expect the first patch release to be relatively soon as bugs are fixed, so I don't think you'll have to wait too long for v2.3.1.

hiroshi-matsuda-rit:
@adrianeboyd Okay. I understand where are you coming from and agree to wait until release v2.3.1. Thanks for taking time for releasing the Japanese model!

hiroshi-matsuda-rit:
@adrianboyd Thanks a lot!

PR #5575, opened 2020-06-11
train is from-config by default

svlandeg:
## Description
The old `train` command is removed and now `train` always means `train_from_config`. Note that I kept the original file name `train_fom_config.py` only for this PR, to view the changes better. After this, we probably just want to rename it back to `train.py`

### Changes to `train` script
* `init_tok2vec` to refer to previously pretrained layer (cf PR #5543)
* `verbose` logging (as before)
* `tag_map_path` and `omit_extra_lookups` (as before)
* `nlp.resume_training()` if `[training]` has `resume = true`
* `nlp.rehearse` if `raw_text` is provided
* Add evaluation results back to the meta file (?)
* Added more informative warnings & errors for when the config is not OK & for when the data can't be read 
* There's a `replace` functionality in `load_model_from_config`. Unclear whether we want to specifically offer this functionality, but the additional safety check on `if name in nlp.pipe_names:` doesn't hurt.
* Functionality that is not ported from the old script is the `eval_beam_widths` - this feels more like a hyperparameter search and should probably be done differently ?

### Textcat
* Verify `exclusive_classes` & `positive_label` are set correctly (as before)
* You can set which evaluation you want in the config: 
    * `textcat_f` will be the "normal" F-score for binary classification if the positive label is defined, in all other cases it will be the macro-avg F-score
```
scores = ["speed", "textcat_f", "textcat_auc"]
```
### Defaults moved out of the code and into the config
* `noise_level` 
* `omit_extra_lookups` 

The following defaults for parser & ner:
* `learn_tokens`
* `min_action_freq`
* `beam_width`
* `beam_update_prob`

To allow for component defaults and `nlp.create_pipe()` to still work as before, I added a `default_config` to the `component` decorator. Unfortunately this doesn't work for the parser, so the defaults needed to be added at the end of `pipes.pyx`. 

### Multitask
* Tried to get the multitask stuff working again with the config, but it's not fully functional yet because of the nestedness of components. Ideally it should look something like this (but allow a list of them):
```
[nlp.pipeline.parser.multitasks]
factory = "nn_labeller"
target = "dep"

[nlp.pipeline.parser.multitasks.model]
@architectures = "spacy.MultiTask.v1"
maxout_pieces = 3
token_vector_width = 96
```

### Types of change
enhancement, UX

## Checklist


svlandeg:
Test 4725 hangs.

PR #5606, opened 2020-06-19
Reduce memory usage of Lookup's BloomFilter

PluieElectrique:
## Description

When loading a model's vocab, `Lookups` creates several `Table`s. Each `Table` creates a `BloomFilter` with a default size of 1 million items. For small models, this size is too big. (This is assuming that the lookup tables are read-only and that we don't need to allocate extra space--I may be wrong about this.)

 For example, here are the sizes (`len(data)`) of `en_core_web_sm-2.3.0`'s vocab:
```
lookups.bin
    lemma_lookup: 41582
    lexeme_norm: 3506
    lemma_rules: 5
    lemma_index: 4
    lemma_exc: 4
lookups_extra.bin
    lexeme_cluster: 0
    lexeme_prob: 0
    lexeme_settings: 0
```
With a default error rate of `1e-4`, a `BloomFilter` of 1 million items takes up about 18.3 MiB, which results in a total of 146.4 MiB for the eight tables above. If we instead use the sizes of the data to hint the `BloomFilter`, we only use about 0.82 MiB.

Note: I used `max(len(data), 1)` because a divide by zero error is thrown in [`preshed.bloom.calculate_size_and_hash_count`](https://github.com/explosion/preshed/blob/985cf43445b8ab7ecb0ca78856c9446f03915253/preshed/bloom.pyx#L14) if we pass in `0` for `members`. We could do `BloomFilter(0, 0)` to use as little memory as possible, but it only saves 16 bytes over `BloomFilter.from_error_rate(1)`.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/lookups.py
+++ b/spacy/lookups.py
@@ -121,7 +121,6 @@ class Lookups(object):
         self._tables = OrderedDict()
         for key, value in srsly.msgpack_loads(bytes_data).items():
             self._tables[key] = Table(key, value)
-            self._tables[key].update(value)
         return self
 
     def to_disk(self, path, filename="lookups.bin", **kwargs):
--- a/spacy/lookups.py
+++ b/spacy/lookups.py
@@ -120,7 +120,7 @@ class Lookups(object):
         """
         self._tables = OrderedDict()
         for key, value in srsly.msgpack_loads(bytes_data).items():
-            self._tables[key] = Table(key)
+            self._tables[key] = Table(key, value)
             self._tables[key].update(value)
         return self
 
@@ -192,7 +192,7 @@ class Table(OrderedDict):
         self.name = name
         # Assume a default size of 1M items
         self.default_size = 1e6
-        size = len(data) if data and len(data) > 0 else self.default_size
+        size = max(len(data), 1) if data is not None else self.default_size
         self.bloom = BloomFilter.from_error_rate(size)
         if data:
             self.update(data)
--- a/.github/contributors/PluieElectrique.md
+++ b/.github/contributors/PluieElectrique.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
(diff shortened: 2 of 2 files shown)


adrianeboyd:
Thanks for this contribution! That's a good catch in `from_bytes` and I didn't think to profile the memory usage for the new empty tables.

I think the default size is too high overall for typical uses of these tables. The tables aren't read-only, but most users never modify them, so the size when loaded is a good default.

spacy/lookups.py
```diff
@@ -120,7 +120,7 @@ def from_bytes(self, bytes_data, **kwargs):
         """
         self._tables = OrderedDict()
         for key, value in srsly.msgpack_loads(bytes_data).items():
-            self._tables[key] = Table(key)
+            self._tables[key] = Table(key, value)
             self._tables[key].update(value)
```
adrianeboyd, line 124: This line can be removed because the data is added in the line above.

spacy/lookups.py
```diff
@@ -192,7 +192,7 @@ def __init__(self, name=None, data=None):
         self.name = name
         # Assume a default size of 1M items
         self.default_size = 1e6
```
adrianeboyd, line 194: I'm not entirely sure how the performance degrades if the size is way too small, but for the lookups tables we have, the average size is 250K and the median size is 5K. Maybe something like 10K is a better default? (@matt, what do you think?)

spacy/lookups.py
PluieElectrique: Thanks, I've removed the line.

spacy/lookups.py
PluieElectrique: 10k seems good to me. Would that be a minimum or a just a default if no `data` is provided? I'm thinking about the empty tables in `lookups_extra.bin`—if 10k isn't a minimum, then the Bloom filters will still have a size of 1, which will render them useless if the user adds to the tables.

But, if we make 10k a minimum, does that mean that we should initialize the size to `len(data) + 10000` when `data` is provided? That is, if we assume the tables aren't read-only, then should we allocate extra space? If we don't, the Bloom filter will still work, but the error rate will be higher. I'm not sure if we need to account for that or not.

adrianeboyd: That's a good point. Matt knows more about the performance details and I'm still not sure what he thinks about the default size here. Hopefully if performance is crucial people will look into the details a bit here like you have, but it is pretty buried in the code.

The main use case I could imagine where the extra lookups tables grow is that someone loads a provided model with an empty prob table and wants to add their own probabilities, which could easily be for 1M+ entries. But the memory usage for the much more common case where the empty probs table isn't touched should be reasonable and should be the main priority, I think.

adrianeboyd: Maybe `size = max(len(data), self.default_size)` with a default size of 10k?

honnibal:
Why don't we just turn off the bloom filter if the data is small?

PR #5613, opened 2020-06-19
Update universe.json

nlptechbook:
This update includes a Telegram bot that uses spaCy to find the most relevant articles about stock market and the channel filled with this bot.

## Checklist

- [x ] I ran the tests, and all new and existing tests passed.

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -1775,6 +1775,34 @@
             "category": ["standalone", "conversational"]
         },
         {
+            "id": "stocknewstip_bot",
+            "title": "stocknewstip_bot",
+            "slogan": "Get latest information about a company's stock.",
+            "description": "This bot can find and bring to you the latest information about a company's stock, using scraping and NLP techniques available in spaCy." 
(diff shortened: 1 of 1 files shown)


honnibal:
Thanks for letting us know about this, and glad you were able to build it with spaCy.

While we don't strictly demand that the universe entries are open-source, it's much harder for users to learn from the example or use it as a reference if there's no source and not much information about what the project does. Are you able to provide a bit more detail? I understand if not, but it would mean we couldn't really include this in the universe.

nlptechbook:
I can include link to my medium article
https://medium.com/swlh/getting-straight-to-the-point-with-scraping-and-natural-language-processing-1a62aba65586,
which provides information about the project and explains how it
works, including code snippets. If it suits you, please let me know.
In that case, please also suggest where I could include this link.

honnibal:
Thanks for the link but I still feel like it's not quite enough for people to really use as a resource for spaCy. Congratulations on launching the product, I just don't think it's a good candidate for inclusion on the universe page.

PR #5614, opened 2020-06-19
added example of ignoring a specific warning

MartinoMensio:
## Description

This PR includes an example for ignoring a specific warning that needs to be done differently from version 2.2 (breaking change). Since I never used the `warnings` package before and probably many spaCy users too, I think that an example is needed.
I also hope that this is the most appropriate way of ignoring one of the warnings declared and used by spaCy.

### Types of change

It is a change in the documentation, adding a code example

## Checklist

### DIFF
--- a/website/docs/usage/v2-3.md
+++ b/website/docs/usage/v2-3.md
@@ -159,6 +159,17 @@ debugging your tokenizer configuration.
 
 #### Warnings configuration
 
+> #### Ignore one specific warning
+>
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Thanks, this is a helpful addition! I'd like to modify it slightly, but don't want to edit it here since it would push to your master branch (in general it's good idea to use a branch for PRs), so I'll create a new PR instead.

PR #5619, opened 2020-06-20
Update develop from master

ines:
## Checklist


spacy/lang/en/tokenizer_exceptions.py
```diff
@@ -396,6 +396,14 @@
...
         {ORTH: "'s", LEMMA: PRON_LEMMA, NORM: "us"},
     ],
+    "c'mon": [
+        {ORTH: "c'm", NORM: "come", LEMMA: "come"},
+        {ORTH: "on"}
+    ],
+    "C'mon": [
+        {ORTH: "C'm", NORM: "come", LEMMA: "come"},
+        {ORTH: "on"}
+    ]
```
svlandeg, line 406: @honnibal : do we want to keep this exception in?

spacy/tests/regression/test_issue5152.py
```diff
@@ -1,3 +1,6 @@
+# coding: utf8
+from __future__ import unicode_literals
+
```
svlandeg, line 3: @ines : having dropped Python 2, we don't need these coding annotations and future imports anymore, right? Checking across the `develop` branch, it looks like these have crept back in from the merges from `master`...

website/docs/api/cli.md
```diff
@@ -455,7 +455,7 @@ improvement.
 
 ```bash
 $ python -m spacy pretrain [texts_loc] [vectors_model] [output_dir]
-[--width] [--depth] [--cnn-window] [--cnn-pieces] [--use-chars] [--sa-depth]
+[--width] [--conv-depth] [--cnn-window] [--cnn-pieces] [--use-chars] [--sa-depth]
 [--embed-rows] [--loss_func] [--dropout] [--batch-size] [--max-length]
```
svlandeg, line 459: This is outdated (with the configs), we'll need to thoroughly revise the CLI documentation for v.3.

ines: Yeah, I've pretty much ignored the docs for now, especially the CLI. I also have a PR up tomorrow that refactors the CLI (using @tiangolo's `typer`). This mostly affects the internals, but it might also open up some new possibilities for CLI arguments.

spacy/tests/regression/test_issue5152.py
ines: Good catch, thanks! And yeah, I think if it's new files, the merge will just port them over as-is and then we have to go hunt for them ourselves. But those should definitely all go!

ines:
@svlandeg Omg, you fixed it and made it go green! ✨🙏 ✨🙏 ✨🙏 ✨🙏 ✨🙏 ✨

spacy/tests/regression/test_issue5152.py
ines: I can take care of that 🙂

spacy/lang/en/tokenizer_exceptions.py
ines: I'll go ahead and merge the PR with it so it's in sync – but we should revisit this when we update the exceptions for v3 (we'll also need to touch this part anyways when we kill the `PRON_LEMMA`).

PR #5626, opened 2020-06-22
Refactor CLI

ines:
## Description

- [x] Rewrite CLI commands to separate functions called from the CLI from the functions called from Python (the CLI commands call into the regular Python functions). Also refactor some of the CLI functions to work better as standalone Python functions, e.g. `info`.
- [x] Update to latest `wasabi`, which will raise errors for messages with non-zero exit codes if `no_print=True` is set, so we can make the functions called from Python "silent" by default while still having them raise errors if something goes wrong, and have pretty-printed output if called via the CLI.
- [x] Clean up, add types, move relevant code into `spacy.cli` and clean up `__main__.py`.
- [x] ~Add `spacy project` command (WIP) with `clone`, `get-assets` and `run` subcommands.~ **Edit:** Moving this to another branch.
- [x] Update to latest `srsly` so we can parse YAML (project config).
- [x] Don't make `ml_datasets` an install dependency. It's only used in the `profile` command and it is also a dev dependency. IMO we shouldn't have to it for every user who installs spaCy?
- [x] Make `spacy package` run `python setup.py sdist` by default, so it outputs the final gzipped package.

### CLI setup

Instead of `plac`, we now use `typer` by @tiangolo (based on `click`, inspired by developing the CLI for Prodigy Teams). The way it works is pretty similar, but it will give us some additional useful features, including:

* subcommands: used for the new `project` CLI
* CLI option and flag names don't have to match the argument names (we do have to specify the flag names separately, but this also means they don't depend on whatever we called the function argument)
* CLI options and flags can specify multiple versions: this is useful for phasing out some inconsistent shortcuts in a backwards-compatible way, like `-f` for `--force` instead of `-F` (which is more consistent, since we typically use uppercase for boolean flags). We can now support both.
* autocomplete on the command line: haven't tested this yet, but it could be pretty cool
* callbacks to run something before every command/subcommand: I'm currently using this for the `project` subcommands to always check that DVC is installed and available
* settings to define more option constraints and raise at the CLI level: for instance, you can define that an argument needs to be a `Path` that exists and not a directory. We probably still want our own checks for the functions called from Python that don't have the CLI take care of the data conversion (e.g. to handle users passin in invalid paths).
* if there are bugs or problems, we have the expert on our team and a direct way to influence what gets fixed 😅

> ⚠️ **Important note:** The only thing that's still not working as intended is the help messages for positional arguments. But @tiangolo is working on implementing this for `typer`. (It's a deliberate [design decision](https://github.com/pallets/click/issues/587#issuecomment-222634050) of `click` but I've never liked this.)

### Project CLI

**Edit:** Moving this to another branch.

While refactoring the CI, I also tested the subcommands and added a simple preliminary version of `spacy project` to clone project templates from a repo, fetch assets with DVC (data version control) and run scripts defined in the `project.yml`. The new version of `srsly` now supports parsing YAML, btw. I'm just working on an end-to-end example and will keep making more updates to the commands as I go. It's already working, though, and pretty exciting 🎉

### Types of change
enhancement

## Checklist


honnibal:
I think we can merge this because it's definitely a step in the right direction and we don't have to wait for it to be perfect. But I have two suggestions for the CLI code:

1. I'd really like to have validity checks happening in separate (pure) functions as much as possible, so that the main function shows the happy path.

2. We've discussed a few ideas for the error handling. It would be good to avoid having the `msg.fail` within the inner function that might be called from Python. I've just had another idea for this:

What if the Python function raises a distinct error type with the message passed as an argument, and it uses `raise` with this error class to do its error handling. In the CLI function, we catch this exception and then do `msg.fail`. So it would be something like:

```
try:
    convert(...)
except SpacyCLIError as e:
    msg.fail(e.text)
```

ines:
> We've discussed a few ideas for the error handling. It would be good to avoid having the msg.fail within the inner function that might be called from Python. I've just had another idea for this:

That's pretty much what `wasabi` now does under the hood if you set `no_print=True`. It's a system exit but with a traceback of the error. It's as close as we can get.

I guess the alternative could work, too, and maybe it's a bit "cleaner" in that sense. But it does also introduce a slightly weird convention.

honnibal:
I think using `raise` within the Python function is a bit more explicit/standard, but the Wasabi mechanism sounds fine, okay. We can always change it later if we want.

ines:
> I think using raise within the Python function is a bit more explicit/standard, but the Wasabi mechanism sounds fine, okay. We can always change it later if we want.

I'll experiment with some ideas for this. One would be to have `wasabi` expose an exception you can raise, and depending on the context, it's either raised or shown as a pretty error. For instance:

```python
msg = Printer(native=True)
if path is None:
    raise msg.Exception("Some error")
```

PR #5631, opened 2020-06-23
Create entrypoints for distributed training with Ray

richardliaw:
## Description

Note: This PR is a WIP and I will need to run the spaCy tests to make sure things are working properly.

TODO:
 - [ ] Finalize User API.
 - [ ] Documentation
 - [ ] Testing

### Types of change

This is an enhancement to enable users to scale out their training on spaCy.

## Checklist


spacy/util.py
```diff
@@ -898,11 +898,17 @@ def escape_html(text):
...
+        cupy.cuda.runtime.getDeviceCount()
+        return True
+    except cupy.cuda.runtime.CUDARuntimeError:
+        return False
 
 def fix_random_seed(seed=0):
     random.seed(seed)
     numpy.random.seed(seed)
-    if cupy is not None:
+    if cupy is not None and gpu_is_available():
```
richardliaw, line 911: I added this because I had a GPU machine with cupy installed but was trying to use CPU training.

spacy/cli/train_from_config.py
```diff
@@ -385,6 +407,8 @@ def create_train_batches(nlp, corpus, cfg):
         )
         if len(train_examples) == 0:
             raise ValueError(Errors.E988)
+        for _ in range(randomization_index):
```
zhuohan123, line 410: what is this line used for? do we need some comments here?

spacy/cli/train.py
```diff
@@ -192,6 +201,10 @@ def train(
     msg.info("Creating nlp from config")
     nlp = util.load_model_from_config(nlp_config)
     optimizer = training["optimizer"]
+    # TODO: is there a cleaner way of doing this, instead of creating
+    # the optimizer twice? are there any problems when doing this?
+    if remote_optimizer:
+        optimizer = remote_optimizer
```
svlandeg, line 207: it seems unideal to overwrite the optimizer from the config with one from a function argument. Is there a way to get the remote_optimizer into `@registry.optimizer`, so it can be called from the config with `@optimizers = "RemoteOptimizer.v1"` (or whatever name)? (cf Adam here: https://github.com/explosion/thinc/blob/master/thinc/optimizers.py#L58)

If you're calling `train` from the CLI, you can pass it a `--code-path` (`-c`) flag with a file defining such custom registered functions. If you're calling `train()` directly, I think you should be able to access the registry from the calling code?

richardliaw: @svlandeg thank you for the review! Yeah, I think what you are suggesting is possible. 

The RemoteOptimizer actually is a wrapper around an arbitrary existing optimizer:
```
class RemoteOptimizer:
    def __init__(self, config_path, communicator, reporter=None):
	    msg.info(f"Loading config from: {config_path}")
	    config = util.load_config(config_path, create_objects=False)
	    util.fix_random_seed(config["training"]["seed"])
	    config = util.load_config(config_path, create_objects=True)
	    training = config["training"]
	    self.optimizer = training["optimizer"]
        ...
```
this means that the user still needs to provide configuration for the underlying optimizer (like Adam). Further, it is probably preferable for the user to not need to modify their existing configuration to "RemoteOptimizer.v1" when doing multi-GPU training. 

One option is to do:

```
@registry.optimizers("RemoteOptimizer.v1")
def RemoteOptimizer(
    underlying_optimizer: str = "Adam",
    opt_kwargs: dict,
):
    opt = get_optimizer(underlying_optimizer, opt_kwargs)
    return RemoteOptimizer(opt)
```

Do you have any thoughts as to how to best expose this?

honnibal:
Now being handled by the forthcoming `spacy-ray` package :tada:

PR #5634, opened 2020-06-23
Fix polarity of Token.is_oov and Lexeme.is_oov

adrianeboyd:
Fix `Token.is_oov` and `Lexeme.is_oov` so they return `True` when the lexeme does **not** have a vector.

## Checklist


adrianeboyd:
(Writing tests for your own code is a bit of liability!)

honnibal:
Damn, sorry I didn't catch that on review!

PR #5641, opened 2020-06-24
Disregard special tag  _SP in check for new tag map

### LINKED ISSUE #5648: part of speech tagger training example from documentation doesn't seem to work after update

https://github.com/explosion/spaCy/blob/master/examples/training/train_parser.py
It worked fine for version 2.0.16, now I have tried it using v2.3.0 and faced the following problem: on every iteration loss is zero, the resulting model assigns '_SP' tag to every token.

## Your Environment
* **spaCy version:** 2.3.0
* **Platform:** Windows-10-10.0.18362-SP0
* **Python version:** 3.8.3

adrianeboyd:
## Description

In `Tagger.begin_training()` check for new tags aside from `_SP` in the new tag map initialized from the provided gold tuples when determining whether to reinitialize the morphology with the new tag map.

This fixes a regression introduced by #5527 when `Tagger.begin_training()` is called without gold tuples.

Fixes #5648.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/pipeline/pipes.pyx
+++ b/spacy/pipeline/pipes.pyx
@@ -528,10 +528,10 @@ class Tagger(Pipe):
                         new_tag_map[tag] = orig_tag_map[tag]
                     else:
                         new_tag_map[tag] = {POS: X}
-        if "_SP" in orig_tag_map:
-            new_tag_map["_SP"] = orig_tag_map["_SP"]
         cdef Vocab vocab = self.vocab
-        if set(new_tag_map) - set(["_SP"]):
+        if new_tag_map:
+            if "_SP" in orig_tag_map:
+                new_tag_map["_SP"] = orig_tag_map["_SP"]
             vocab.morphology = Morphology(vocab.strings, new_tag_map,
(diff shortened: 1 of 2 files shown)


spacy/pipeline/pipes.pyx
```diff
@@ -531,7 +531,7 @@ class Tagger(Pipe):
         if "_SP" in orig_tag_map:
             new_tag_map["_SP"] = orig_tag_map["_SP"]
         cdef Vocab vocab = self.vocab
-        if new_tag_map:
+        if set(new_tag_map) - set(["_SP"]):
```
svlandeg, line 534: Right, came across this on the new WIP branch for `develop`, didn't realise the same bug would be on `master`. Anyway solved it slightly differently there: https://github.com/explosion/spaCy/blob/whatif/arrow/spacy/pipeline/pipes.pyx#L372 basically do the "_SP" stuff only when you have a `new_tag_map`?

adrianeboyd: That seems simpler, thanks! (I wish this could all be a bit less hacky...)

PR #5652, opened 2020-06-26
Convert custom user_data to token extension format for Japanese tokenizer

adrianeboyd:
## Description

Convert the user_data values so that they can be loaded as custom token extensions for `inflection`, `reading_form`, `sub_tokens`, and `lemma`.

## Checklist


Techno-coder:
Was there a reason why this PR was reverted? I can't seem to find any discussions on it 😅. I would have thought accessing inflections by `Token` would be more natural.

adrianeboyd:
Yes, I agree that it seems more natural, but we have a policy that the core `spacy` library never sets custom extensions so that there's no chance for it to introduce conflicts with users' custom extensions, which can potentially have the same names.

PR #5661, opened 2020-06-29
Provide lang to lemmatizer to limit is_base_form

adrianeboyd:
## Description

Provide `lang` to the lemmatizer so that the English-specific `Lemmatizer.is_base_form()` can be restricted to English.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
Nevermind, should do this properly.

PR #5669, opened 2020-06-29
Website: fixed the token span in the text about the rule-based matching example

hertelm:
## Description
The result of the code example at https://spacy.io/usage/rule-based-matching#adding-patterns is `15578876784678163569 HelloWorld 0 3 Hello, world`, which does not match with the text below, where the span is (0, 2). I think the code outcome is the expected result, since the pattern comprises three tokens (_Hello_, _,_, and _world_).

### Types of change
A change to the website.

## Checklist
- [ ] I ran the tests, and all new and existing tests passed. --> No code or tests were changed.

### DIFF
--- a/.github/contributors/hertelm.md
+++ b/.github/contributors/hertelm.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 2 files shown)


adrianeboyd:
That does look incorrect, thanks for the PR!

PR #5674, opened 2020-06-30
Fixes to project CLI

svlandeg:
Some small fixes:

## Description
- `shlex.split` needs the `posix` flag to work on Windows
- added some more informative Error messages here and there, because debugging wasn't always straightforward
- cloning a project in the working dir (no `dest` parameter) didn't work for me, because it would always say the working dir already exists. I added 
```
if dest == Path.cwd():
    dest = dest / name
```
so that it would create the new dir in the subdir of the working dir. Hopefully that doesn't break anything on Linux?

## Remaining trouble
After cloning the GH repo in the temp dir and moving it, on my system (running as admin), the script can't clean up the temp directory:
>  UserWarning: [W091] Could not clean/remove the temp directory at ...\Temp\tmp889ccvy9: [WinError 5] Access is denied: '...\\Temp\\tmp889ccvy9\\.git\\objects\\pack\\pack-661d6ea746862ec352be736c4ec67d142cb230c3.idx'.

This could be a Windows thing. I changed the `make_tempdir` context manager to throw this warning instead of crashing though.

### Types of change
bug fixes

## Checklist

### DIFF
--- a/spacy/cli/project.py
+++ b/spacy/cli/project.py
@@ -243,8 +243,8 @@ def project_clone(
             raise RuntimeError(f"Could not clone the repo '{repo}' into the temp dir '{tmp_dir}'.")
         with (tmp_dir / ".git" / "info" / "sparse-checkout").open("w") as f:
             f.write(name)
-        run_command(["git", "-C", tmp_dir, "fetch"])
-        run_command(["git", "-C", tmp_dir, "checkout"])
+        run_command(["git", "-C", str(tmp_dir), "fetch"])
+        run_command(["git", "-C", str(tmp_dir), "checkout"])
         shutil.move(str(tmp_dir / Path(name).name), str(project_dir))
     msg.good(f"Cloned project '{name}' from {repo}")
     for sub_dir in DIRS:
--- a/spacy/cli/project.py
+++ b/spacy/cli/project.py
@@ -14,6 +14,7 @@ import tqdm
 
 from ._app import app, Arg, Opt, COMMAND, NAME
 from .. import about
+from ..compat import is_windows
 from ..schemas import ProjectConfigSchema, validate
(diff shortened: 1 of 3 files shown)


spacy/cli/project.py
```diff
@@ -236,13 +239,16 @@ def project_clone(
     # We're using Git and sparse checkout to only clone the files we need
     with make_tempdir() as tmp_dir:
         cmd = f"git clone {repo} {tmp_dir} --no-checkout --depth 1 --config core.sparseCheckout=true"
-        run_command(shlex.split(cmd))
+        try:
+            run_command(shlex.split(cmd, posix=not is_windows))
```
ines, line 243: Maybe it's worth it to move the `shlex.split` call out to a `util` function like `split_command`? This way, we can ensure that the `posix` flag is always set correctly (because it's easy to miss).

(I initially considered making the `run_command` helper take both a `List[str]` and `str` and then automatically split if the input is a string. It would make some things a bit cleaner, but on the other hand, I'm typically trying to avoid writing functions that take different input types.)

spacy/util.py
```diff
@@ -467,7 +467,10 @@ def make_tempdir():
     """
     d = Path(tempfile.mkdtemp())
     yield d
-    shutil.rmtree(str(d))
+    try:
+        shutil.rmtree(str(d))
+    except PermissionError as e:
```
ines, line 472: This is strange indeed... Maybe there's something we can set here? Or maybe it's just how Windows works? I vaguely remember something like this coming up in our test suite when working with temp files and directories. I wonder how we fixed that... or did we just skip those on Windows? 🤪

svlandeg: I don't remember recently running into issues with temp files on Windows for the tests. We use it all the time. This particular issue seems to be triggered by a specific file that is "open" or something according to Windows. Dunno.

PR #5679, opened 2020-06-30
Working dir by default

svlandeg:
## Description
Proposal: for all `spacy project` commands, use the working dir by default. This is really convenient when experimenting on the command line in your project folder, and you can type just `python -m spacy project run-all` or `python -m spacy project run visualize`

Note that for this to work, the positional `subcommand` argument needs to be moved one up.

### Types of change
enhancement proposal

## Checklist

### DIFF
--- a/spacy/cli/project.py
+++ b/spacy/cli/project.py
@@ -149,8 +149,8 @@ def project_run_all_cli(
 def project_run_cli(
     # fmt: off
     ctx: typer.Context,
-    project_dir: Path = Arg(Path.cwd(), help="Location of project directory. Defaults to current working directory.", exists=True, file_okay=False),
     subcommand: str = Arg(None, help="Name of command defined in project config"),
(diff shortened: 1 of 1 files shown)


svlandeg:
Also added a custom warning to `run_command`.

spacy/errors.py
```diff
@@ -539,6 +539,7 @@ class Errors(object):
     E199 = ("Unable to merge 0-length span at doc[{start}:{end}].")
 
     # TODO: fix numbering after merging develop into master
+    E970 = ("Can not execute command '{str_command}'. Do you have '{tool}' installed?")
```
ines, line 542: This is a nice idea! 💯

Maybe we should rephrase this slightly to account for cases like `./non_existing_script.sh`? Although, I really like the "Do you have X installed?" message, which will likely be accurate in th majority of cases.

Issue #5643, opened 2020-06-25
KeyError [E018] when using nlp.pipe with n_process > 1


cordone:
## How to reproduce the behaviour

Hi, I'm trying to use the new **ja_core_news_sm** model to stream process a collection of sentences with `nlp.pipe(list_of_sentences)`. I'd like to be able to set `n_process` > 1 to increase speed, but when I do that I encounter KeyError [E018]. I'm using WSL through VSCode.

I'll try something like this...

```python

Word = collections.namedtuple("Word", ["surface", "lemma", "upos", "xpos", "dep"])

nlp = spacy.load("ja_core_news_sm", disable=["ner", "entity_linker"])

for doc in nlp.pipe(sentences, batch_size=150, n_process=2):
   for token in doc:
      word = Word(surface=token.text, lemma=token.lemma_, upos=token.pos_, xpos=token.tag_, dep=token.dep_)
```

and get this error.

```
word = Word(surface=token.text, lemma=token.lemma_, upos=token.pos_, xpos=token.tag_, dep=token.dep_)
  File "token.pyx", line 894, in spacy.tokens.token.Token.lemma_.__get__
  File "strings.pyx", line 136, in spacy.strings.StringStore.__getitem__
 KeyError: "[E018] Can't retrieve string for hash '17260935250788936050'. This usually refers to an issue with the `Vocab` or `StringStore`."
```

The failing token wasn't the first one in the sentence, so I counted the number of tokens throwing Exceptions and in one collection of sentences I have, 397/53597 iterated tokens cause an Exception (so far the number of failures has stayed constant on re-runs varying `batch_size` and `n_process`).

Just to sanity check, a bare `nlp.pipe()` or `[nlp(s) for s in sentences]` work with no issues. Possibly a model-specific issue?

## Your Environment

* **Operating System:** Windows 10 (Build 18363)
* **Python Version Used:** 3.8.1
* **spaCy Version Used:** 2.3.0
* **Environment Information:** WSL (Linux-4.4.0-18362-Microsoft-x86_64-with-glibc2.27)

adrianeboyd:
The Japanese models assign lemmas in the tokenizer rather than in the tagger, which is probably the source of the difference. I wouldn't have actually expected this to matter in the multiprocessing setup (since the token text is also added to the stringstore in the tokenizer step), but I can reproduce this and we'll look into it.

adrianeboyd:
[...]

PR #5684, opened 2020-07-01
Save lemmas in user_data for Japanese

adrianeboyd:
## Description

As a workaround for multiprocessing (#5643), save the lemmas in `user_data` along with the other additional features from sudachipy.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
Nah, also a case where it's better to do this properly.

PR #5686, opened 2020-07-01
♻ Refactor CLI setup for completion

tiangolo:
### Types of change

This PR does the following:

* Pin Typer to a version range safe from breaking changes (there could be breaking changes between `>=0.3.0` and `>=0.4.0`).
* Add `shellingham` as a dependency, a small library used by Typer to detect the current shell and be able to install completion automatically with:

```console
$ spacy --install-completion
```

Without `shellingham` it would have to be:

```console
$ spacy --install-completion zsh

// or

$ spacy --install-completion bash

// or

$ spacy --install-completion fish

// etc.
```
* Update the way the standalone `spacy` command is created, from an integrated shell script to a pure Python "entrypoint" in setuptools:
    * This fixes shell completion because with the current shell script the actual command called ends up being `python -m spacy some-args` instead of `spacy some-args`. And shell completion systems depend on that name of the command (otherwise, it triggers completion for `python`).

Here's how completion looks like afterwards:

[asciicast](https://asciinema.org/a/GTlCdFuDLI0VnSoqotwqIH3NK)

## Checklist

- [x] I have submitted the spaCy Contributor Agreement. ...I think so? I actually don't know :thinking: 
- [x] I ran the tests, and all new and existing tests passed. (Didn't change any internals).

### DIFF
--- a/.github/contributors/tiangolo.md
+++ b/.github/contributors/tiangolo.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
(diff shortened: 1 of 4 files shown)


ines:
Wowwww, this looks cool 🤩

> Add shellingham as a dependency, a small library used by Typer to detect the current shell and be able to install completion automatically with:

I think we'd be a bit reluctant to adopt a new third-party dependency in spaCy just for this feature (it's convenient, but it probably doesn't justify the extra dependency). @honnibal, what do you think?

tiangolo:
Sure! Should I remove it right away?

honnibal:
Yeah I think people can just write their shell? The detection is one more thing to go wrong, too

tiangolo:
Perfect, done! :heavy_check_mark:

PR #5688, opened 2020-07-01
Remove deprecated methods: Doc.print_tree, Doc.merge, Span.merge

ines:
## Description

These have been deprecated for a while and shouldn't be used anymore, so we might as well remove them for v3.

### Types of change

enhancement

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -1123,50 +1123,6 @@ cdef class Doc:
                 remove_label_if_necessary(attributes[i])
                 retokenizer.merge(span, attributes[i])
 
-    def merge(self, int start_idx, int end_idx, *args, **attributes):
-        """Retokenize the document, such that the span at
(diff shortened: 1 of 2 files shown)


svlandeg:
It looks like the `collu.py` script still uses `doc.merge` ... 
and `custom_component_entities.py` and `custom_component_countries_api.py` still use `span.merge` :|

adrianeboyd:
I think the `conllu.py` script can probably go. I don't think it works 100% with `master` and it would definitely need a ton of updates for v3. Using `spacy convert` is probably better in general. We can add an example project for this kind of thing?

svlandeg:
It would definitely be good to clean out the examples dir. I don't always know the history of those and whether or not they are still useful.

ines:
I've added a task for cleaning out the examples directory to the v3 board  👍 And yes, I think for v3, we should want fewer of these arbitrary scripts and move the more complex stuff out to projects.

PR #5694, opened 2020-07-02
Include git commit in package and model meta

adrianeboyd:
## Description

Include git commit in package and model meta

* save current git commit in `spacy/git_info.py`
* include current git commit in `meta.json` when models are saved
* git version code in `setup.py` is adapted from numpy, add `licenses/3rd_party_licenses.txt` with relevant license

### Types of change

Enhancement

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,7 @@ Pipfile.lock
 *.egg
 .eggs
 MANIFEST
+spacy/git_info.py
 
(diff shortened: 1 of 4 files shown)


adrianeboyd:
I didn't want to call it `version.py` because our version is in `about.py` and I didn't want it to be confusing, but `git_info.py` is easy to change here.

PR #5697, opened 2020-07-02
Record whether Doc objects are built from known spacing

honnibal:
Now that we don't have the `GoldParse`, it's important that the `Doc` object be able to track whether it was created from actual text, or whether we filled in the `spaces` variable with just `True`. This addresses a bug in the conversion where incorrectly spaced text was being tokenized.

The name of the field is kind of janky, but I don't know what else to call it.

We should create a flags variable on the `Doc` to hold these flag values, it's getting a bit bad to have all this cruft on the object. We can just have accessors in Python.

### DIFF
--- a/spacy/gold/example.pyx
+++ b/spacy/gold/example.pyx
@@ -15,7 +15,7 @@ from ..syntax import nonproj
 
 
 cpdef Doc annotations2doc(vocab, tok_annot, doc_annot):
-    """ Create a Doc from dictionaries with token and doc annotations. Assumes ORTH & SPACY are set. """
+    """ Create a Doc from dictionaries with token and doc annotations. """
     attrs, array = _annot2array(vocab, tok_annot, doc_annot)
     output = Doc(vocab, words=tok_annot["ORTH"], spaces=tok_annot["SPACY"])
     if "entities" in doc_annot:
@@ -414,7 +414,7 @@ def _parse_links(vocab, words, links, entities):
 
 def _guess_spaces(text, words):
     if text is None:
-        return [True] * len(words)
+        return None
     spaces = []
     text_pos = 0
     # align words with text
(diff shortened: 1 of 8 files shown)


spacy/gold/example.pyx
```diff
@@ -15,7 +15,7 @@ from ..syntax import nonproj
 
 
 cpdef Doc annotations2doc(vocab, tok_annot, doc_annot):
-    """ Create a Doc from dictionaries with token and doc annotations. Assumes ORTH & SPACY are set. """
+    """ Create a Doc from dictionaries with token and doc annotations. """
     attrs, array = _annot2array(vocab, tok_annot, doc_annot)
     output = Doc(vocab, words=tok_annot["ORTH"], spaces=tok_annot["SPACY"])
```
svlandeg, line 20: The assumption of ORTH and SPACY being set is still present in this line though ...

honnibal: The `SPACY` value can be `None` by `_guess_spaces`, so it felt weird to say "are set". I guess we could clarify?

svlandeg: Right, with "set" I meant "they are in the dictionary". I think it makes sense to expect that before this function is called, but like you said the value can still be `None`, which I think works fine further downstream.

adrianeboyd:
I think you need to make a lot of parallel changes to `DocBin`, too. (I think that `Doc.to_bytes()` and `DocBin` should share more code at some point.)

honnibal:
Ah, great point about the `DocBin`.

The two bits of code are a bit different, maybe `doc.to_bytes()` could one day go through `DocBin`, but I think it's better to let the two situations optimize separately. I like the idea of using Arrow one day, which would solve the problem.

PR #5700, opened 2020-07-02
Experimental character-based pretraining

honnibal:
This PR extracts a pretraining option from an experiment branch. I did some heavy rebasing to clean up hacks that this was interwoven with while I was experimenting, so 🤞 it doesn't break anything.

The idea behind the character-based objective is to predict only a few characters of the word, for instance the first and last three characters. This avoids the need for the vector training, and may be better for inflectional languages.

Edit: Okay, it runs. I think this can be merged, I'll port for develop.

## Checklist


spacy/pipeline/pipes.pyx
```diff
@@ -861,6 +866,8 @@ class ClozeMultitask(Pipe):
         self.vocab = vocab
         self.model = model
         self.cfg = cfg
+        self.cfg.setdefault("objective", "characters")
+        self.cfg.setdefault("nr_char", 4)
```
svlandeg, line 870: *twitch*

spacy/pipeline/pipes.pyx
```diff
@@ -1069,6 +1094,7 @@ cdef class DependencyParser(Parser):
     assigns = ["token.dep", "token.is_sent_start", "doc.sents"]
     requires = []
     TransitionSystem = ArcEager
+    nr_feature = 8
```
svlandeg, line 1097: isn't this in the config as `nr_feature_tokens`?

PR #5701, opened 2020-07-02
fix bug in Korean language, resulting in 100x speedup by reducing overhead of mecab

mikeizbicki:
**The Bug:** The original code in the `Korean` class called the `MeCab` constructor once for each document being parsed.  This code has a large overhead, however, because it must launch and kill a system process for each document.  (Each `MeCab` class communicates with the `mecab` executable via a system process).

**The Fix:** This pull request changes the behavior so that the `MeCab` constructor is called exactly once when the `Korean` class is executed, and the constructor is reused for each of the documents.  This is how `MeCab` was intended to be used, and the reduced overhead results in a 100x speedup.  On the original code, running
```
>>> import timeit
>>> import spacy.lang.ko
>>> nlp = spacy.lang.ko.Korean()
>>> timeit.timeit(lambda: nlp('이거은 테스트 케이스를 생성하기 위해 Google Translate에 넣은 예제 테스트 문장입니다.'), number=1000)
```
gives a runtime of
```
37.5136206799998
```
but on the new code, the runtime is only
```
0.34927031499955774
```

**Note:** The original call to the `MeCab` class was within a `with` block, which is the recommended way to call the class per the `natto-py` documentation.  Inspecting the `__exit__` function of the `MeCab` class, however, we can see that all that it does is call the `__del__` function directly.  Therefore, to prevent resource leaks, the new code also calls `MeCab`'s `__del__` function within `Korean`'s `__del__` function.

### DIFF
--- a/spacy/lang/ko/__init__.py
+++ b/spacy/lang/ko/__init__.py
@@ -42,7 +42,10 @@ def check_spaces(text, tokens):
 class KoreanTokenizer(DummyTokenizer):
     def __init__(self, cls, nlp=None):
         self.vocab = nlp.vocab if nlp is not None else cls.create_vocab(nlp)
-        self.Tokenizer = try_mecab_import()
+        self.MeCab = try_mecab_import()("-F%f[0],%f[7]")
+
+    def __del__(self):
+        self.MeCab.__del__()
 
     def __call__(self, text):
         dtokens = list(self.detailed_tokens(text))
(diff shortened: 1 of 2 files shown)


mikeizbicki:
I would like you to know that I strongly dislike having this contributor agreement for a project licensed under MIT since the agreement doesn't give either of us any legal protection and is a pain in the ass.  If I had known I would have had to do this before fixing this bug, I honestly wouldn't have bothered.

There's no need to reply to this, I just wanted you to know how I felt.

honnibal:
@mikeizbicki Sincerely, thanks for the feedback. It's a reasonable reaction and it's the sort of feeling we can expect to be under-reported. We're thinking aobut this.

honnibal:
Btw regarding your PR:

I'm disappointed that I didn't catch this in review, as it's a classic type of problem I'm usually quite aware of. I'm really pretty bad at reviewing PRs in Github, I need to think of ways to improve.

Thanks for the fix.

spacy/lang/ko/__init__.py
```diff
@@ -42,7 +42,10 @@ def check_spaces(text, tokens):
 class KoreanTokenizer(DummyTokenizer):
     def __init__(self, cls, nlp=None):
         self.vocab = nlp.vocab if nlp is not None else cls.create_vocab(nlp)
-        self.Tokenizer = try_mecab_import()
+        self.MeCab = try_mecab_import()("-F%f[0],%f[7]")
```
honnibal, line 45: So what we're assigning here now is the tokenizer instance, right?

Can we write this as:

```python

MeCab = try_mecab_import()
self.mecab_tokenizer = MeCab("-F%f[0],%f[7]")
```

It seems a bit clearer, otherwise it looks like we're assigning the module or the class instead.

spacy/lang/ko/__init__.py
```diff
@@ -42,7 +42,10 @@ def check_spaces(text, tokens):
...
+
+    def __del__(self):
```
honnibal, line 47: Is this definitely necessary? I would expect the object's cleanup code to be executed as normal during cleanup when it's ref-counted away.

I'm always wary of defining `__del__` methods of Python objects unnecessarily, it's a tricky thing that can have unexpected performance penalties as it can require Python to use garbage collection instead of reference counting.

honnibal: Oops just saw your note.

honnibal: Ugh, I hate SWIG wrapped code. I can't tell whether the `__del__` is really necessary. Best to just assume it is, okay.

spacy/lang/ko/__init__.py
mikeizbicki: I've added a new commit fixing this.

svlandeg:
FYI - the tests seem to be temporarily broken due to something with distutils/setuptools, but unrelated to this specific PR.

PR #5707, opened 2020-07-03
Feature/example only

svlandeg:
## Description

- Removed the option to pass in `(text, annot)` tuples to `nlp.update`, `nlp.evaluate` and `nlp.rehearse` - instead always require a list of `Example` objects, created with `Example.from_dict(nlp.make_doc(text), annot)`
- Added few more hopefully helpful warnings
- Tried to do a thorough sweep in tests and example scripts 

### Types of change
enhancement?

## Checklist

### DIFF
--- a/examples/training/train_morphologizer.py
+++ b/examples/training/train_morphologizer.py
@@ -86,9 +86,9 @@ def main(lang="en", output_dir=None, n_iter=25):
     nlp.add_pipe(morphologizer)
 
     # add labels and create the Example instances
-    examples_train_data = []
+    train_examples = []
     for text, annotations in TRAIN_DATA:
-        examples_train_data.append(Example.from_dict(nlp.make_doc(text), annotations))
+        train_examples.append(Example.from_dict(nlp.make_doc(text), annotations))
         morph_labels = annotations.get("morphs")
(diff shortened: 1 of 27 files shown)


examples/training/train_morphologizer.py
```diff
@@ -84,8 +85,10 @@ def main(lang="en", output_dir=None, n_iter=25):
     morphologizer = nlp.create_pipe("morphologizer")
     nlp.add_pipe(morphologizer)
 
-    # add labels
-    for _, annotations in TRAIN_DATA:
+    # add labels and create the Example instances
+    train_examples = []
+    for text, annotations in TRAIN_DATA:
+        train_examples.append(Example.from_dict(nlp.make_doc(text), annotations))
```
honnibal, line 91: It's probably better to have the loop fetch the tags and morphology out of the `Example.reference` `Doc` object, so that it's easier to use the example on converted data.

I don't want to request changes here though, especially since we'll be changing the examples more later anyway and removing a bunch.

svlandeg: Yea I just pretty much did a ctrl-F sweep and replaced whatever I found, but yea some of these examples are probably redundant in v.3.

PR #5710, opened 2020-07-04
Remove dead and/or deprecated code

ines:
## Description

This PR removes a bunch of old and/or deprecated code, including:

* Backwards compatibility support for long unsupported v1.x formats and syntax.
* Backwards compatibility errors that have been raised since early v2.x versions (unsupported arguments and formats).
* Functions that are now available in Thinc (schedules, `fix_random_seed` etc). They're still provided as imports from Thinc to prevent people's code from breaking.
* Unused error and warning messages.

## Checklist


spacy/language.py
```diff
@@ -232,32 +232,6 @@ def meta(self, value):
     def config(self):
         return self._config
 
-    # Conveniences to access pipeline components
-    # Shouldn't be used anymore!
-    @property
-    def tagger(self):
```
honnibal, line 238: Are we sure we want to remove these? I don't think they're very harmful, and it keeps old example code working.

spacy/language.py
```diff
@@ -313,10 +287,7 @@ def create_pipe(self, name, config=dict()):
         DOCS: https://spacy.io/api/language#create_pipe
         """
         if name not in self.factories:
-            if name == "sbd":
-                raise KeyError(Errors.E108.format(name=name))
```
honnibal, line 317: Why remove these errors?

spacy/tokens/token.pyx
```diff
@@ -237,11 +237,6 @@ cdef class Token:
         index into tables, e.g. for word vectors."""
         return self.c.lex.id
 
-    @property
-    def string(self):
```
honnibal, line 241: I'm pleased to finally be rid of this. It's an example of something that I _do_ feel really gets in the way.

spacy/language.py
ines: I think they're kinda harmful, because they produce confusing code that's hard to follow and very intransparent. It also perpetuates this idea of "_the_ tagger" (just like _the_ English model) that we want to move away from. `nlp.get_pipe("tagger")` isn't much more verbose, but much more explicit.

spacy/language.py
ines: Because they're only for backwards-compatibility handling of single very specific cases that haven't worked for a long time now. No user should have `sbd` in their pipeline anymore at this point, because it would have errored anyway. I don't think this specific case warrants extra treatment anymore – we might as well show the same error we show for all other non-existent factories.

spacy/language.py
honnibal: Okay fair enough.

PR #5711, opened 2020-07-05
Improve gold-standard alignment

honnibal:
* Replace the janky alignment data structures with Thinc's `Ragged` array
* For now, use the pytokenizations alignment library. We can probably just implement the function ourselves and remove the dependency, but let's just get this working for now.
* Use the "if all the tags are the same" logic suggested by @svlandeg 
* Use somewhat more permissive NER alignment. Might be bad? Testing.


spacy/tests/test_gold.py
```diff
@@ -298,7 +296,7 @@ def test_gold_biluo_one_to_many(en_vocab, en_tokenizer):
     gold_words = ["I", "flew", "to", "San", "Francisco", "Valley", "."]
     example = Example.from_dict(doc, {"words": gold_words, "entities": entities})
     ner_tags = example.get_aligned_ner()
-    assert ner_tags == ["O", None, "U-LOC", "O"]
+    assert ner_tags == ["O", "U-ORG", "U-LOC", "O"]
```
svlandeg, line 300: The second "word" is "flew to", of which only "flew" is annotated as "ORG", so isn't `None` more appropriate for a mixture of "ORG" and "O" ?

spacy/tests/test_gold.py
```diff
@@ -339,7 +336,7 @@ def test_gold_biluo_misaligned(en_vocab, en_tokenizer):
     gold_words = ["I", "flew to", "San", "Francisco Valley", "."]
     example = Example.from_dict(doc, {"words": gold_words, "entities": entities})
     ner_tags = example.get_aligned_ner()
-    assert ner_tags == [None, None, "B-LOC", "L-LOC", "O"]
+    assert ner_tags == ["B-ORG", "L-ORG", "B-LOC", "L-LOC", "O"]
```
svlandeg, line 340: Also here I think the `None`'s are more appropriate as the gold entity can not be mapped back to the original words.

honnibal: Can we update the example so it's more realistic? I keep getting confused by the text

svlandeg: Yea, I know, let me push a quick update.

svlandeg: So basically in `predicted`, the first two tokens are "Mr. and Mrs." and "Smith", but the NER annotation reads "Mrs. Smith". I think the first token should probably be `None`, as "Mr" was never annotated. You could argue about the second token "Smith", whether it should inherit the label from "Mrs. Smith", but I can imagine contexts where half of the entity doesn't make sense to be an entity by itself.

spacy/tests/test_gold.py
svlandeg: Updated: the `predicted` token "Mrs. Smith" consists of "Mrs." being annotated as a PERSON and "Smith" as "O", so shouldn't the final result be None instead?

spacy/tests/test_gold.py
honnibal: Thanks, I see the problem. I suggested a fix.

We could go a bit further and strip internal whitespace from the spans, to account for stuff like extra newlines. But if the underlying strings match I thing the code is correct.

PR #5731, opened 2020-07-08
Refactor the Scorer to improve flexibility

adrianeboyd:
## Description

Refactor the `Scorer` to improve flexibility for arbitrary pipeline components.

* Individual pipeline components provide their own `evaluate` methods that score a list of examples and return a dictionary of scores
* `Scorer` is initialized either:
  * with a provided pipeline containing components to be scored
  * with a default pipeline containing the built-in statistical components (senter, tagger, morphologizer, parser, ner)
* `Scorer.score` evaluates a list of examples and returns a dictionary of scores referring to the scores provided by the components in the pipeline
* `Scorer` includes new general methods for scoring:
  * `score_tokenization`: `token_acc` and new token-as-char-span PRF scores
  * `score_token_attr`: accuracy for a token attribute
  * `score_token_attr_per_feat`: PRF per-feat for a label in UFEATS format
  * `score_spans`: PRF + per-type PRF for labeled spans
  * `score_deps`: UAS, LAS, and LAS per type
  * `score_cats`: PRF or ROC AUC for a doc attribute with label scores like `Doc.cats`

Significant differences:

* Scoring is no longer cumulative: `Scorer.score` scores a list of examples rather than a single example and does not retain any state about previously scored examples
* `tags_acc` is renamed to `tag_acc` to be consistent with `token_acc` and the new `morph_acc`, `pos_acc`, and `lemma_acc`
* `uas`, `las`, and `las_per_type` are renamed to `dep_uas`, `dep_las`, `dep_las_per_type`
* `Doc.sents` is scored as the set of spans rather than on the set of sentence-initial tokens
* PRF values in the returned scores are no longer multiplied by 100

### Types of change

Enhancement.

## Checklist


honnibal:
Nice, really glad to have this broken up! I have a couple of suggestions for reorganization.

I think we should name this method on the pipeline components `get_scores` rather than `evaluate`. The reason is that we've already used `evaluate` on the `Language` object, and it actually runs the components. So we should use a different name for this method. I think `get_scores` matches `get_loss`, so it makes sense to me.

Second, I'd prefer most of the scoring logic to be moved out into standalone functions that are then called by the methods. We might want to reuse those bits of scoring logic, and if they're inside a method we can only reuse it via inheritance, which introduces a bunch of incidental problems. For instance, someone might make a component that does two tasks jointly. Their scoring method should be able to just call the two scoring functions, rather than having to think about multiple inheritance or something.

adrianeboyd:
I see the parallel to `get_loss` (the kind of ad hoc return dicts are similar, although the `get_loss` naming is more consistent) and see the difference between `evaluate` and `pipe` (I was thinking it was closer to `pipe`), but I agree with Sofie that the term `score` is pretty ambiguous. This isn't a function to get the same thing called `scores` in `predict` or `update`. Hmm...

I thought about having the returned scores keyed on the components like losses, but it didn't make sense since you're scoring the doc at the end of the pipeline, not the particular component individually, so `senter` will run a `sents` evaluation that is actually `senter` + `parser`. The evaluation is duplicated, but duplicating it is the only way to make this relatively flexible.

Standalone functions are a good idea, I will implement that. It will make it a bit slower, but that's probably not a huge concern.

adrianeboyd:
Okay, I think it can be keyed relatively well on the attribute name. It won't work if you have `Doc.cats` and `Token.cats`, but let's hope you have better names. Then the only additional changes in addition to `tag_acc` from above would be that you would have `dep_uas`, `dep_las`, and `dep_las_per_type`.

honnibal:
Yes I see your point about the name `get_scores`. That's bad.

Perhaps `get_evaluation`? `get_metrics`?

adrianeboyd:
I don't really like `get_` because it seems like it's calculating, not getting. How about `score` in parallel to `Scorer.score`? Then it would be more like `Language.pipe` and `Pipe.pipe`?

honnibal:
I see what you mean about `get_`, but our getters are usually properties, and we have `get_loss` for analogy. I would prefer it to `score`, which sounds very nouny to me.

adrianeboyd:
I really don't like `get_scores` because it's so ambiguous. Which scores from where?

If you really want to it be `get_`, then I think `get_evaluation` is better to make it sound more similar to `get_loss` and to make clear it's related to `Language.evaluate`.

On the other hand, I think the typical user isn't aware of `get_loss` at all, so I'm not sure basing it on that makes sense for the API overall. And you have a verb-y `Scorer.score` already?

I don't like `score_examples` because it seems like it's setting up a contrast that's not there. There's no other `score_texts` or anything.

honnibal:
Okay fair enough. Let's go with `component.score()`.

adrianeboyd:
Hmm, I realized I've mixed the scorer config into the `component_cfg`. Should I move it into a separate config?

honnibal:
@adrianeboyd I think it's good in the component CFG, because it means new components will own that configuration. When you invent a new component, you shouldn't have to update some central other place where the scorer config is defined.

PR #5736, opened 2020-07-09
Refactor Chinese tokenizer configuration

adrianeboyd:
## Description

Refactor `ChineseTokenizer` configuration so that it uses a single `segmenter` setting to choose between character segmentation, jieba, and pkuseg.

* replace `use_jieba`, `use_pkuseg`, `require_pkuseg` with the setting `segmenter` with the supported values: `char`, `jieba`, `pkuseg`
* make the default segmenter plain character segmentation `char` (no additional libraries required by default)

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -112,6 +112,8 @@ class Warnings(object):
             "string \"Field1=Value1,Value2|Field2=Value3\".")
     W101 = ("Skipping `Doc` custom extension '{name}' while merging docs.")
     W102 = ("Skipping unsupported user data '{key}: {value}' while merging docs.")
+    W103 = ("Unknown {lang} word segmenter '{segmenter}'. Supported "
+            "word segmenters: {supported}. Defaulting to {default}.")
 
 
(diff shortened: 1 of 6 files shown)


honnibal:
This looks much better! The docs change is quite nice.

@ines Do we have to update the examples in the Chinese course for this?

adrianeboyd:
I'm hoping to standardize this a bit across languages, so any API changes/suggestions are welcome at this point.

I tried using an enum instead just a tuple of strings of supported segmenters, but it seemed like too much hassle.

ines:
@honnibal At the moment, the examples are still all in English. And I don't think this would change much, because it's not like the actual output of the `pkuseg` tokenization will be different, right?

honnibal:
@ines Okay, great. I just wasn't sure whether we had usage examples that loaded the Chinese model.

I can see why the enum was tempting here, unfortunately I don't think I've ever found it helpful. So I'll go ahead and merge this.

PR #5747, opened 2020-07-10
Refactor config validation and train CLI args

ines:
## Description

This PR makes use of a pydantic schema for validating the training config, so we can show detailed errors for missing or incorrect values. I have removed the defaults from the schema, since this otherwise gets very messy – instead, we should include a `.cfg` with the defaults that we can use to give people a "starter config". Our test suite can then validate the default config against the schema to make sure it's always up-to-date and correct.

The PR is currenty draft, because the strict (but correct) validation will likely cause errors for some of our experimental configs.

Other changes:

* Update `pretrain` and `debug-data` to take config and make them read all settings from config.
* Move debug-related commands (`debug-data`, `debug-model`, `profile`, plus new `debug-config`) under `debug` as subcommands. This keeps the command overview tidy (when you run `--help`) and groups related commands together. Commands are also added under their previous name to the main CLI so people's commands don't break.

## CLI command overrides

The PR also introduce updates to the training commands and hopefully a solution for our "CLI argument creep" and "paths in the config" problems via an update I'm working on for Thinc. Relevant PR: explosion/thinc#360

#### Problem summary

Our previous `spacy train` command hat tons of CLI arguments that could easily get messy, and a bunch of hidden defaults all across the components. We already solved this by introducing the configs that build up the object tree, validate the arguments and are considered the "single source of truth" for the training process. It also keeps the command simple and straightforward.

However, there's one remaining problem: the config-only solution would mean that the user has to provide local paths in their config – like, paths to vectors, tok2vec weights, raw text, tag maps, _plus_ any other paths their custom components may need (which we can't anticipate). There are also other config settings users might want to override without having to edit the config every time – like `use_gpu`.

One approach would be to just keep _some_ arguments on the CLI, e.g. everything that needs a path or that's frequently toggled. But the big worry here is that over time, it will inevitable cause more and more CLI arguments to keep creeping back in as we add more features and need more options. Then we'd end up at the worst of both worlds: a config that doesn't quite cover _everything_ and that's conditional on arbitrary CLI arguments, with a distinction that's not always obvious to the user. The approach also doesn't address custom architectures and components: for those, the user would always end up with paths in their config again.

#### Proposed solution

Our training commands should only take the training and dev data, a config and an optional code file. That's it. But we can give the user a general-purpose unified CLI API for overwriting any values in their config. For instance, `--nlp.vectors /path` or `--training.batch_size.start 200`. The nested values refer to nested section in the config and would cover built-in and custom blocks. We should be able to implement this with a comnbination of Typer and an option in Thinc that allows passing in overrides to `registry.make_from_config`.

We can also consider adding an "update config" option to our training commands that will output a filled config with the overrides, so the user always has a record of the exact config they ran, including any overwritten values.

### Types of change
enhancement

## Checklist


svlandeg:
> Our training commands should only take the training and dev data, a config and an optional code file. That's it. But we can give the user a general-purpose unified CLI API for overwriting any values in their config. 

I really like this idea. I think it's the best solution against the otherwise inevitable parameter-creep in the CLI commands, while giving the user all the flexibility.

> We can also consider adding an "update config" option to our training commands that will output a filled config with the overrides, so the user always has a record of the exact config they ran, including any overwritten values.

I would definitely do this. As a user running ML experiments, I always want to keep detailed track of the inputs for any run. Storing the command AND the config that gets overridden is definitely less clean than storing the final, merged/filled config.

ines:
@svlandeg Glad you like it! With the latest Thinc update I just pulled in, it should all work as expected already in this PR. The schemas are now enforced, so we might have to update a few configs here and there to make sure the specify all defaults (but the errors should be pretty straightforward here).

> I would definitely do this. As a user running ML experiments, I always want to keep detailed track of the inputs for any run. Storing the command AND the config that gets overridden is definitely less clean than storing the final, merged/filled config.

Now that I think about it, we already store the training config with the model we export after training, right? That config would then be the final config with the updated values from the CLI 🙂

PR #5759, opened 2020-07-13
Refactor pipeline components, config and language data

ines:
## Description

Relevant tests to see expected behaviour in action:
* `pipeline/test_pipe_factories.py`
* `serialize/test_serialize_config.py`

### Pipeline components

Pipeline components can be marked using the following decorators:
	* `@Language.component`: for stateless function components
	* `@Language.factory`: for stateful components that are created using a function or class
	
The decorator takes an argument for the string name of the factory, an optional instance name, arguments for the assigned/required attributes and a `default_config`, a dict of the arguments passed to the function. This should follow the regular config format and not include any Python objects, just references to registered functions. Under the hood, the decorator takes care of registering the factory (using our function registry).

A factory should at least have two named arguments: `nlp` and `name` (the display name of the component instance, can be used in the component to write to the `losses` etc.). All other arguments will be passed in via the config. Type hints will be used for validation later, if available. 

```python
@Language.component("my_component")
def my_component(doc):
    return doc

@Language.factory("my_factory", default_config={"foo": "hello", "bar": 12})
def my_factory(nlp: Language, name: str, foo: str, bar: int):
    return my_component(nlp.vocab, name, foo=foo, bar=bar)
```

The `nlp.add_pipe` method now takes the **string name** of the component factory and an optional `config`. It returns the created pipe, so you can use it further, e.g. to add labels to it.

```python
ner = nlp.add_pipe("ner")
my_pipe = nlp.add_pipe("my_factory", config={"foo": "world", "bar": 15})
```

`nlp.add_pipe` is able to **create a pipe from config**, by resolving the pipeline component factory via the registry. The config passed in will be deep merged with the defaults (otherwise, the workflow would be too annoying, because you'd always have to specify every parameter). Under the hood, the object needed to create a pipeline component looks like this:

```json
{"@factories": "my_factory", "foo": "world", "bar": 15}
```

Or, represented in the exported config:

```ini
[nlp]
lang = "en"

[nlp.pipeline]

[nlp.pipeline.ner]
@factories = "my_factory"
foo = "world"
bar = 15
```

This means that:

* The regular `config.cfg` we use for training can also construct an `nlp` object and its pipeline components, out-of-the-box 🎉 `Language.from_config` now has to do very little, and can be our go-to way for deserializing models, with the `config.cfg` as the "single source of truth" for the pipeline.
* We can use type hints to validate the inputs and users can easily add them to their custom components to get instant feedback and validation. Same principle as Thinc configs.
* We can easily have multiple instances of the same component in the pipeline and the component can be serialized/deserialized, add custom losses etc.
* The config required by a factory can define pretty much anything and map to any custom function arguments. There's no single default model anymore – a component could take multiple models, or any other nested objects that can be expressed via the config. 
* **Bonus:** Custom language subclasses out-of-the-box! I didn't plan for this but since we're not also using function registries for language classes like `en` etc., any custom language class you register will automatically be available if you run `get_lang_class` because all it does is look up the code in the registry!

### Component meta under the hood

When a pipeline component is added, we create a `FactoryMeta` dataclass that stores metadata about it – like the factory name, requires/assigns and whatever else we may need in the future. This dataclass is stored for each factory and for each component instance – so given a component name, we can always fetch its info. We can also use this for static analysis.

### Config

We also store the final filled component configs when a component is added, so `nlp.config` is able to reproduce the config object of the current pipeline. This is the full config file, so `nlp.config.to_disk()` will let you export a usable config with all defaults set that you can use during training (still need to add the training defaults, see todos). This also answers many questions we had around the the "getting started with a config" workflow and means that users can also just set up a `nlp` pipeline and then export the training config for it.

Of course, this solution means that we have to be able to trust components to serialize their config that can be changed by the user. If you just overwrite some arbitrary component attribute, you can't expect it to be reflected in the config you export – but that's defensible IMO. If you want to initialize a new component with different config, re-add it to the pipeline – that's the safest solution anyways.

The `config.cfg` serialized with a model is able to reproduce the full complex pipeline and training. The `meta.json` is now only used to provide _actual_ meta information. Some of the properties are currently duplicated across meta and config (pipeline info etc.) and we could consider removing some of them from the meta to avoid confusion.

### Default config

The default training config is currently included in `spacy/default_config.cfg`. It's loaded by the `Language` class and used as the default config in `Language.config`. The `nlp` and `nlp.pipeline` config is added on top of it. This means that if you run `nlp.config.to_disk(...)`, it will export a training-ready config. And the `nlp.config` of an existing model you load in will always reflect the training config of that model.

To ensure that our default config is always up-to-date, we can run our validation, loaders and training commands against it in the test suite.

### Notes on config loading (mostly internals)

The one tricky part about loading the config is the `[nlp.pipeline]` block: we can't just call `registry.make_from_config` on that block, because the registered `@factories` all require two additional arguments: `nlp` and `name`. Those are filled in automatically by `Language.add_pipe` and `Language.create_pipe`. 
The classmethod `Language.from_config` takes care of filling and constructing the whole config.

### Language data

Instead of the messy `LanguageDefaults` and its classmethods to create a vocab, lemmatizer and tokenizer, languages can now define a `default_config`. This is used when you initialise a language class, e.g. `English()`. To pass in a custom config, you can use `English.from_config` instead.

This approach also makes it easy to define custom tokenizers for specific languages, with custom settings. For instance, the Chinese default config currently looks like this:

```ini
[nlp]
lang = "zh"

[nlp.tokenizer]
@tokenizers = "spacy.ChineseTokenizer.v1"
use_jieba = false
use_pkuseg = false
require_pkuseg = false
pkuseg_model = null
pkuseg_user_dict = "default"

[nlp.writing_system]
direction = "ltr"
has_case = false
has_letters = false
```

Using more explicitly named entry points that provide data via registered functions, it also becomes immediately clear where the lookups data is loaded from. To change the data sources, a user can override the argument in their config. Going forward, the lemmatizer should be replaced by a simple pipeline component, registered on the respective language class, e.g. `@Greek.factory("lemmatizer")`.

```ini
[nlp]
lang = "el"

[nlp.lemmatizer]
@lemmatizers = "spacy.GreekLemmatizer.v1"

[nlp.lemmatizer.data_paths]
@assets = "spacy-lookups-data"
lang = ${nlp:lang}
```

#### Open questions

* How do we deal with the `base_model` setting? I've moved this to the `[training]` section, along with the `vectors` setting, because these settings really define how the model is initialised for training – not the current state of the `nlp` object. In `train`, it's currently just loading the base model instead if it's provided – but it's unclear how we should deal with inconsistencies here? What if the components don't match? Should we check the `nlp.config` of the base model against the actual config's `[nlp]` block?
* Where do the vectors come in? How are we adding/loading them during training?
* How are we going about _filling_ a config, i.e. taking a config with incomplete values and filling in all defaults, if possible? (It's important to note that filling a conifg also always requires to create the objects, like registered functions. Otherwise we can't know their arguments.) At the moment:
  * you run `spacy debug config` with `--auto-fill` to auto-fill a config in place (or output it somewhere)
  * load a model with an incomplete config and have it auto-filled (just to be safe)

### Todos

- [x] Add default training config to `Language.config` and deep merge with whatever is provided on load, so the config that can be exported actially fully usable in `train` etc.
- [x] Two entity linker tests currently segfault because they pass the `kb` object into the config and it somehow causes a segfault during `json.dumps` under the hood. There are two problems here:
	* the segfault, obviously 😅 no idea why it does this, I wasn't able to reproduce it by dumping a KB
	* Python objects being passed into the conig: the config passed to components should be JSON-serializable. If we want objects to be created, we can either do this within the component/factory, or use a function registry.
- [x] I removed the base model / replace logic from `load_model_from_config` because this should be something we should handle separately in `spacy train`. The `base_model` is an argument of `[training]`, and I don't think the regular model loading should allow hacking together a frankenmodel. Way too messy and too many opportunities for stuff to be out of sync (vectors etc.).
- [x] Clean up entries in `_component_meta` and `_pipe_configs` when a component is removed or renamed.
- [x] ~~Re-add `W098` and related warnings around default config.~~
- [x] Add more tests for config serialization and deserialization, especially:
	* examples of configs with multiple instances of components (e.g. `[nlp.pipeline.ner1]` and `[nlp.pipeline.ner2]`)
	* edge cases, tricky configs, common errors to make sure they're handled
- [x] Fix `debug model`: we somehow need it to be able to provide the "path" to config section containing the model to debug, like `nlp.pipeline.tagger.model`. (See the `dot_to_dict` helper in `pipe_analysis.py` if we need to convert the dot notation to a nested dict.)
- [x] Update docstrings.
- [x] Update branch from latest develop.
- [x] Decide how we want the pipe analysis to work and clean up related code (only works partially at the moment).
- [x] Decide how we want to integrate the other language data into the config and replace `Language.Defaults`.
- [ ] Add more tests to run training etc. with `default_config.cfg` to make sure it always works.
- [ ] Figure out why that Russian lemmatizer test is failing.
- [ ] Clean up `Vocab.from_config`.
- [ ] Document everything.

### Todos in Thinc
https://github.com/explosion/thinc/pull/369

- [x] Move config copying to Thinc as `Config.copy`.
- [x] Make sure `Config.from_str` orders the sections in a reasonable way.
- [x] Serialize refrences to registered functions with no args as dict

```diff
[section]
+ subsection = {"@some_registry": "SomeFunction.v1"}

- [section.subsection]
- @some_registry = "SomeFunction.v1"

```

### Types of change
enhancement

## Checklist


honnibal:
I do like how this comes out! We've talked about this a bunch offline so I'll just add some more notes about the design decisions here.

## Why `add_pipe` can't just take objects anymore

We need to know how to serialize and deserialize the object, which means we need to be able to find a constructor function. Previously we accepted that serialization might not work properly if you added stuff to your pipeline, that's something we want to fix in v3. But that means being a bit more strict about how the pipeline works.

## Why `Language.factory` and `Language.component` are classmethods

We want to allow subclasses to set their own defaults, so that different languages or user classes can be configured properly.

## Why does the `Language.factory` decorator do so much magic? Can't you just take the config and let the function do the work?

The `Language.factory` decorator uses the signature of the function to create a schema, so that we can go via the Thinc `make_from_config` logic, which gives us validation and good errors. It also saves the user from a bunch of tricky code in their factory functions.

spacy/language.py
```diff
@@ -265,7 +293,112 @@ def pipe_labels(self):
...
+    @classmethod
+    def has_factory(cls, name: str) -> bool:
+        """RETURNS (bool): Whether a factory of that name is registered."""
+        return name in registry.factories
+
+    @classmethod
+    def get_factory_meta(cls, name: str) -> "FactoryMeta":
+        """RETURNS (FactoryMeta): The meta for the given factory name."""
+        if name not in cls._pipe_meta:
+            raise ValueError(Errors.E967.format(name=name))
```
svlandeg, line 305: Should we specify "factory" vs "component" in the error message of E967? Because right now it says "component meta", also when calling`get_factory_meta` ?

spacy/language.py
```diff
@@ -265,7 +293,112 @@ def pipe_labels(self):
...
+        if cls.has_factory(name):
+            raise ValueError(Errors.E004.format(name=name))
+
+        def add_factory(factory_func: Callable) -> Callable:
+            argspec = inspect.getfullargspec(factory_func).args
+            if "nlp" not in argspec or "name" not in argspec:
+                raise ValueError(Errors.E964.format(name=name))
+            # Officially register the factory so we can later call
+            # registry.make_from_config and refer to it in the config as
+            # @factory = "spacy.Language.xyz". We use the class name here so
```
svlandeg, line 346: drop the "@" ?

spacy/language.py
```diff
@@ -278,49 +411,57 @@ def get_pipe(self, name):
...
+        if not srsly.is_json_serializable(config):
+            raise ValueError(Errors.E961.format(config=config))
+        if not self.has_factory(factory_name):
+            raise ValueError(Errors.E002.format(name=factory_name))
+        pipe_meta = self.get_factory_meta(factory_name)
+        config = config or {}
+        # This is unideal, but the alternative would mean you always need to
+        # specify the full config settings, which is not really viable.
+        if pipe_meta.default_config:
+            config = util.deep_merge_configs(config, pipe_meta.default_config)
```
svlandeg, line 434: Should we add a check here, and warn if default values have been merged/filled in?

spacy/language.py
ines: Ah, good point, yes! 

(I initially had only one dict for factories but then I realised that we also need to be able to fetch meta for any given component _instance_, so I added the second.)

spacy/language.py
ines: Yes, I think this part could definitely benefit from more warnings 👍 We should also add `W097` and `W098` back here. 

We could probably just add a simple check that compares the default config to the regular config and only attempt to deep merge if they're different at all.

Not sure what the most efficient way is to check whether _some_ values have been added... like, we only care about block that don't reference registered functions or blocks that reference the same registered functions. Maybe the `deep_merge_configs` helper could just keep a count? Or maybe I'm overthinking this 😅

spacy/language.py
ines: Shouldn't this be `@factories`?

```suggestion
            # @factories = "spacy.Language.xyz". We use the class name here so
```

spacy/pipeline/entityruler.py
```diff
@@ -1,16 +1,49 @@
...
+)
+def make_entity_ruler(
+    nlp: Language,
+    name: str = "entity_ruler",
+    phrase_matcher_attr: Optional[Union[int, str]] = None,
+    validation: bool = False,
+    overwrite_ents: bool = False,
+    ent_id_sep: str = DEFAULT_ENT_ID_SEP,
+    patterns: Optional[List[PatternType]] = None,
+):
```
svlandeg, line 35: Should we remove the default values from `make_entity_ruler` - as we got them in the `default_config`?

spacy/pipeline/entityruler.py
```diff
@@ -22,7 +55,16 @@ class EntityRuler:
...
+    def __init__(
+        self,
+        nlp: Language,
+        name: str = "entity_ruler",
+        phrase_matcher_attr: Optional[Union[int, str]] = None,
+        validate: bool = False,
+        overwrite_ents: bool = False,
+        ent_id_sep: str = DEFAULT_ENT_ID_SEP,
+        patterns: Optional[List[PatternType]] = None,
+    ):
```
svlandeg, line 67: we have the same set of default values here...

svlandeg: Oh, this happens in the PR throughout. Like `incl_prior=True` for the `EntityLinker` is defined three times?

spacy/pipeline/entityruler.py
ines: Yeah, I've been thinking this, too... I agree would be cleaner to not use defaults in our code and code examples to prevent duplication and make it clear that the _config_ is where the defaults come from. I think it's a reasonable best practice.

spacy/pipeline/entityruler.py
ines: I think it's reasonable to not use function defaults on factories – they won't be used and it communicates the wrong thing. So I'll remove them.

Not sure how we should go about default arguments on the `Pipe` subclasses etc. In theory, those could still be instantiated separately, and the factory arguments don't _have to_ match the subclass arguments. So I think it makes sense to keep them there, even if it causes some duplication. But I guess it's not too weird to have a class with default args and then a function that instantiates it and passes in args.

svlandeg:
> Should we deprecate and remove the "pipeline" and "pipe_factories" setting in the meta.json/nlp.meta?
> 
>     Pros: Now unused (pipeline is defined in the config), can get out-of-sync and be confusing because changing the values in the meta actually won't do anything
>     Cons: May break people's code?

I would do, yes. We can safely assume people's code is going to be broken anyway, switching from v.2 to v.3, so I think now would be the correct time to do this... It would be good to cut away any redundancy we can, it's going to be less confusing in the long run.

spacy/cli/train.py
```diff
@@ -277,29 +260,26 @@ def evaluate():
...
-    dropout,
-    eval_frequency,
-    accumulate_gradient=1,
-    patience=0,
-    max_steps=0,
+    dropout: float,
+    eval_frequency: int,
+    accumulate_gradient: int = 1,
+    patience: int = 0,
+    max_steps: int = 0,
```
svlandeg, line 282: I don't think we need any default values here, we should assume we've gotten them from the config

spacy/default_config.cfg
```diff
@@ -0,0 +1,84 @@
...
+max_steps = 20000
+eval_frequency = 200
+eval_batch_size = 128
+# Other settings
+seed = 0
+accumulate_gradient = 1
+use_pytorch_for_gpu_memory = false
+# Control how scores are printed and checkpoints are evaluated.
+scores = ["speed", "tags_acc", "uas", "las", "ents_f"]
+score_weights = {"las": 0.4, "ents_f": 0.4, "tags_acc": 0.2}
```
svlandeg, line 25: For (human) readability, let's just keep the order of `scores` and `score_weights` the same?
```suggestion
score_weights = {"tags_acc": 0.2, "las": 0.4, "ents_f": 0.4, }
```

ines: Ah yes! There are probably more thing we want to adjust in the default config – I think this is copied over from one of Matt's Onto Notes experiment configs.

spacy/tests/test_misc.py
```diff
@@ -138,3 +138,22 @@ def test_is_compatible_version(version, constraint, compatible):
...
+            {"token.pos": True, "token._.xyz": True},
+            {"token": {"pos": True, "_": {"xyz": True}}},
+        ),
+        (
+            {"training.batch_size": 128, "training.optimizer.learn_rate": 0.01},
+            {"training": {"batch_size": 128, "optimizer": {"learn_rate": 0.01}}},
+        ),
+    ],
+)
+def test_dot_to_dict(dot_notation, expected):
```
svlandeg, line 156: This is a neat little function!

spacy/cli/debug_data.py
```diff
@@ -46,17 +47,26 @@ def debug_config_cli(
...
     msg.good("Config is valid", show=not is_stdout)
+    if auto_fill:
+        orig_config = config.to_str()
+        filled_config = nlp.config.to_str()
+        if orig_config == filled_config:
+            msg.text("Config is up-to-date, no values were auto-filled")
+        else:
+            msg.info("Updated config values")
+        if diff:
+            print(diff_strings(config.to_str(), nlp.config.to_str()))
```
svlandeg, line 64: Great idea!

spacy/cli/debug_data.py
```diff
@@ -46,17 +47,26 @@ def debug_config_cli(
...
     if is_stdout:
         print(nlp.config.to_str())
-    elif output_path is not None or auto_fill:
-        filled_output = output_path if output_path is not None else config_path
-        nlp.config.to_disk(filled_output)
-        msg.good("Saved updated config", filled_output)
+    elif output_path is not None:
+        nlp.config.to_disk(output_path)
+        msg.good("Saved updated config", output_path)
```
svlandeg, line 70: I was going through your commits sequentially and I agree that we probably should avoid overwriting stuff (was going to make a comment otherwise). Better to be on the safe side here ;-)

spacy/default_config.cfg
```diff
@@ -22,7 +22,7 @@ accumulate_gradient = 1
 use_pytorch_for_gpu_memory = false
 # Control how scores are printed and checkpoints are evaluated.
 scores = ["speed", "tags_acc", "uas", "las", "ents_f"]
-score_weights = {"las": 0.4, "ents_f": 0.4, "tags_acc": 0.2}
+score_weights = {"tags_acc": 0.2, "las": 0.4, "ents_f": 0.4, }
```
svlandeg, line 25: damnit, sorry
```suggestion
score_weights = {"tags_acc": 0.2, "las": 0.4, "ents_f": 0.4}
```

spacy/cli/debug_data.py
ines: Yeah, I got bitten by that immediately, like 10 minutes after writing the code to do it, so I was like... nope, that needs to go again 😅

spacy/tests/test_misc.py
ines: Actually, I think this (or something like it) might be the missing piece for the oopen TODO in `debug model`: now that it takes the whole config and loads it, there needs to be a way for the user to specify which model to debug. So I thought on the CLI you could set `--section nlp.pipeline.tagger.model` and then it will select that component, if it exists.

spacy/errors.py
```diff
@@ -502,10 +550,12 @@ class Errors:
...
-    E985 = ("The pipeline component '{component}' is already available in the base "
-            "model. The settings in the component block in the config file are "
-            "being ignored. If you want to replace this component instead, set "
-            "'replace' to True in the training configuration.")
+    E984 = ("Invalid component config for '{name}': no @factories key "
+            "specifying the registered function used to initialize the "
+            "component. For example, @factories = \"ner\" will use the 'ner' "
+            "factory and all other settings in the block will be passed "
+            "to it as arguments.\n\n{config}")
+    E985 = ("Can't load model from config file: no 'nlp' section found.\n\n{config}")
```
svlandeg, line 558: I'm not sure about all these `\n\n{config}` lines in the error messages. I guess it could be useful for debugging, but I'm not a huge fan of spamming the terminal with such a blob of text...

svlandeg: Also, the same messages get used when running `debug config`, at which point surely the user knows which config they are debugging and it feels annoying to see it printed again...

ines: Maybe we can truncate them? I do think for the smaller sub-blocks, it can be useful, because the origin of the config may not be immediately obvious to the user. It's one thing if you're loading in a `config.cfg` you wrote, but if you have an error somewhere down the line when creating/adding components, I think it makes sense to show the data the error is based on.

spacy/tests/test_misc.py
svlandeg: Done!

spacy/cli/debug_model.py
```diff
@@ -39,27 +41,46 @@ def debug_model_cli(
...
+        msg.info(f"Fixing random seed: {seed}")
+        fix_random_seed(seed)
 
-    debug_model(
-        config_path, print_settings=print_settings,
-    )
+    component = config
+    parts = section.split(".")
+    for item in parts:
+        try:
```
ines, line 65: Ah, this is a practical solution! I was initially overthinking this in my head and was only thinking about more complex recursion 😅

We should consider moving this out to a helper function in util? I think it fits in well with our suite of dot notation <-> dict helpers and it totally seems like something that'll come up in different places. Like, selecting some property down the tree, checking if a config has a certain setting etc.

spacy/language.py
```diff
@@ -922,7 +1223,70 @@ def _link_components(self):
...
+                    lang_code=cls.lang,
+                    lang=util.get_object_name(cls),
+                )
+            )
+        nlp_config["lang"] = cls.lang
+        # This isn't very elegant, but we remove the [pipeline] block here to prevent
+        # it from getting resolved (causes problems because we expect to pass in
+        # the nlp and name args for each component). If we're auto-filling, we're
+        # using the nlp.config with all defaults.
+        config = Config().from_str(config.to_str())  # deepcopy config!
```
honnibal, line 1258: So the obvious mistake people will make as you've pointed out before, is people will pass actual instances in the config, like e.g. as a `Vocab`instance or a model or something. What will the failure look like here if they do that?

Also, can we just implement a `Config.copy` method in Thinc?

ines: Ah, I should add my `is_json_serializable` check here and do a proper deep copy. Forgot to add that here. See error `E961` for the message we raise.

spacy/language.py
```diff
@@ -922,7 +1223,70 @@ def _link_components(self):
...
+        config["nlp"]["pipeline"] = {}
+        non_pipe_overrides, pipe_overrides = _get_config_overrides(overrides)
+        resolved, filled = registry.resolve(
+            config, validate=validate, schema=ConfigSchema, overrides=non_pipe_overrides
+        )
+        filled["nlp"]["pipeline"] = orig_pipeline
+        config["nlp"]["pipeline"] = orig_pipeline
+        create_tokenizer = resolved["nlp"]["tokenizer"]
+        lemmatizer = resolved["nlp"]["lemmatizer"]
+        nlp = cls(create_tokenizer=create_tokenizer, lemmatizer=lemmatizer)
```
honnibal, line 1269: I do see why you were eager to avoid this approach and instead let the make from config handle it, but I think in the end this is quite okay. Yeah it's a bit fiddly, but it's one bit of code and you've made it quite easy to understand imo. I like that you can see the `add_pipe` loop in there, just like you'd have in a script.

ines: Yeah, I think this is okay 🙂 Also check out the `Vocab.from_config` – if we adjust that and improve the handling of defaults, move the lemmatizer out etc., this can actually be pretty elegant.

spacy/language.py
```diff
@@ -1149,13 +1472,15 @@ def _fix_pretrained_vectors_name(nlp):
...
         self.nlp = nlp
         self.names = names
         # Important! Not deep copy -- we just want the container (but we also
         # want to support people providing arbitrarily typed nlp.pipeline
         # objects.)
         self.original_pipeline = copy(nlp.pipeline)
+        self.metas = {name: nlp.get_pipe_meta(name) for name in names}
+        self.configs = {name: nlp.get_pipe_config(name) for name in names}
         list.__init__(self)
         self.extend(nlp.remove_pipe(name) for name in names)
```
honnibal, line 1485: Does `nlp.remove_pipe` change the config?

I think either answer will be a bit painful. If it does, it will be a rare time when the config updates from state changes, which will be a bit inconsistent.

But we do need to change the config here, don't we? `nlp.add_pipe` changes the config so that we load back in the updated pipeline. And disable pipes is how we've asked people to change the pipeline before saving previously.

spacy/language.py
```diff
@@ -417,11 +680,20 @@ def remove_pipe(self, name):
         if name not in self.pipe_names:
             raise ValueError(Errors.E001.format(name=name, opts=self.pipe_names))
         removed = self.pipeline.pop(self.pipe_names.index(name))
+        # We're only removing the component itself from the metas/configs here
+        # because factory may be used for something else
+        self._pipe_meta.pop(name)
+        self._pipe_configs.pop(name)
```
honnibal, line 686: Removing the config is definitely right here, as it's the inverse of what we do in `nlp.add_pipe`.

It's worth noting a small inconsistency here though. In most other places, the config doesn't update when you change the object state. But this does, because we're removing the pipeline from the meta.

What we're doing is right and any alternative would be worse, but the inconsistency is unfortunate. I'm not sure whether users will notice/care, but some might be confused.

spacy/lang/fr/lemmatizer.py
```diff
@@ -13,7 +16,9 @@ class FrenchLemmatizer(Lemmatizer):
     the lookup table.
     """
 
-    def __call__(self, string, univ_pos, morphology=None):
+    def __call__(
+        self, string: str, univ_pos: str, morphology: Optional[Morphology] = None
```
adrianeboyd, line 20: `morphology` is a dict of morphological features, not a `Morphology` object.

spacy/lang/en/lemmatizer.py
```diff
@@ -0,0 +1,38 @@
+from typing import Optional
+
+from ...morphology import Morphology
+
+
+def is_base_form(univ_pos: str, morphology: Optional[Morphology] = None) -> bool:
```
adrianeboyd, line 6: Also a `Dict` here.

spacy/lang/fr/lemmatizer.py
```diff
@@ -133,3 +95,41 @@ def lemmatize(self, string, index, exceptions, rules):
         if not forms:
             forms.append(string)
         return list(set(forms))
+
+
+def is_base_form(univ_pos: str, morphology: Optional[Morphology] = None) -> bool:
```
adrianeboyd, line 100: And again. It should probably be changed to be a `MorphAnalysis`.

spacy/lang/fr/lemmatizer.py
ines: Ah, damn, I copy-pasted that around so this is everywhere now! I'll just do a search and replace!

adrianeboyd: It's not a great name and I'm not even sure the format is really standardized.

adrianeboyd: I can probably change it to a `MorphAnalysis` after this is merged, but for now it should be a `Dict`.

PR #5767, opened 2020-07-15
Remove corpus-specific tag maps and morph rules

adrianeboyd:
## Description

Remove corpus-specific tag maps from the language data for languages without custom tokenizers that provide tags. For languages with custom word segmenters that also provide tags (Japanese and Korean), the tag maps for the custom tokenizers are kept as the default.

The default tag maps for languages without custom tokenizers that provide tags are now the default tag map from `lang/tag_map.py`, UPOS -> UPOS.

Remove corpus-specific morph rules. Add options similar to tag maps to provide them in the `train` and `debug-data` CLIs.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
This requires #5764 for one test.

ines:
Screenshot 2020-07-15 at 16 01 59

😍

PR #5768, opened 2020-07-15
Improve tag map initialization and updating

adrianeboyd:
## Description

Generalize tag map initialization and updating so that a provided tag map can be loaded correctly in the CLI.

* normalize provided tag map as necessary
* use the same method for initializing and overwriting the tag map

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/cli/debug_data.py
+++ b/spacy/cli/debug_data.py
@@ -70,8 +70,8 @@ def debug_data(
     else:
         lang_cls = get_lang_class(lang)
         nlp = lang_cls()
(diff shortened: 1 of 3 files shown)


adrianeboyd:
I think just updating the tag map dict as before can cause subtle bugs related to the morph rules due to the use of the sorted tag names and tag ids. I don't think it actually causes problems in current v2 models because the default tag maps provide all the tags used in the morph rules, but these problems were revealed after removing the default tag maps in v3.

PR #5770, opened 2020-07-16
Hyphen infix

svlandeg:
## Description
Currently, splitting on a hyphen as infix happens only when the hyphen is inbetween `ALPHA` characters. This PR extends that to also include cases where the previous word ends with a number `0-9`.

This solves a case like `6,000-year` which makes it 3 tokens (instead of 1).

Funnily enough, checking the test suite, we previously were testing that "11-septembre" would also be one token in French. I don't see the rationale for this? With the current edit, this would be 3 tokens and I think that's fine?

### Types of change
enhancement?

## Checklist

### DIFF
--- a/spacy/lang/en/__init__.py
+++ b/spacy/lang/en/__init__.py
@@ -5,6 +5,7 @@ from .lex_attrs import LEX_ATTRS
 from .morph_rules import MORPH_RULES
 from .syntax_iterators import SYNTAX_ITERATORS
 
+from .punctuation import TOKENIZER_INFIXES
 from ..tokenizer_exceptions import BASE_EXCEPTIONS
 from ...language import Language
 from ...attrs import LANG
@@ -24,6 +25,7 @@ class EnglishDefaults(Language.Defaults):
     stop_words = STOP_WORDS
     morph_rules = MORPH_RULES
     syntax_iterators = SYNTAX_ITERATORS
+    infixes = TOKENIZER_INFIXES
     single_orth_variants = [
         {"tags": ["NFP"], "variants": ["…", "..."]},
         {"tags": [":"], "variants": ["-", "—", "–", "--", "---", "——"]},
--- a/spacy/lang/en/punctuation.py
+++ b/spacy/lang/en/punctuation.py
@@ -1,0 +1,19 @@
+from ..char_classes import LIST_ELLIPSES, LIST_ICONS, HYPHENS
+from ..char_classes import CONCAT_QUOTES, ALPHA_LOWER, ALPHA_UPPER, ALPHA, PUNCT
+
+_infixes = (
+    LIST_ELLIPSES
+    + LIST_ICONS
+    + [
+        r"(?<=[0-9])[+\-\*^](?=[0-9-])",
+        r"(?<=[{al}{q}])\.(?=[{au}{q}])".format(
+            al=ALPHA_LOWER, au=ALPHA_UPPER, q=CONCAT_QUOTES
+        ),
+        r"(?<=[{a}]),(?=[{a}])".format(a=ALPHA),
+        r"(?<=[{a}0-9])(?:{h})(?=[{a}])".format(a=ALPHA, h=HYPHENS),
+        r"(?<=[{a}0-9])[:<>=/](?=[{a}])".format(a=ALPHA),
+    ]
+)
+
+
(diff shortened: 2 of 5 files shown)


adrianeboyd:
The performance on the training corpora is more important than the individual test cases here, so I'd want to check carefully before making this change.

svlandeg:
I checked on French (UD GSD) which remained the same performance. For English (UD EWT) token accuracy also remains the same, though there are slightly less undertokenizations (from 0.0036 to 0.0034)

adrianeboyd:
This is the base infix setting that could potentially used or adapted by any language, though, and hyphens are one of the most problematic cases. It's obviously not frequent enough to make a huge difference, but I'd want to check with all of the training corpora just in case.

Or we could add it as an English setting rather than a global setting. It does seem to be the right setting for PTB tokenization. (It's not consistent in OntoNotes, but splitting on the hyphen for this pattern is much more common.)

adrianeboyd:
This lowers the token accuracy for UD_French-Sequoia the tinest smidge. It's totally fine as an English-specific setting, though.

svlandeg:
Hm, but Sequoia is much smaller than GSD. Also, correct me if I'm wrong, but the only case that seems affected in Sequoia is the token `Arg3-Pro4` which appears twice, and is not French-specific.

There's a difficult grey area in biomedical texts whether these kind of bonds should be one entity or two. I think we agreed that we'd prefer to oversegment in such grey areas, as we can always learn to merge them back together. (some related discussion [here](https://github.com/explosion/spaCy/issues/3242#issuecomment-464447606))

I did run a comparison performance on all languages in UD (taking the biggest treebank for each), there's virtually no change in overal F-score except for Tamil (`ta`), which drops precision from 98% to 97% and F from 99% to 98% because an increase in overtokenization from 0.0054 to 0.0096. So we could also make Tamil the exception instead.

Anyway. This feels like a lot of nitpicking/effort for virtually no effect on overall performance. I do think that general-purpose texts would benefit in general from this small edit, but it might be offset by some exceptions in domain-specific texts like biomedical. So in the end I guess it doesn't matter much. But the unit tests specifcally testing that `11-septembre` is one token, feels just wrong.

adrianeboyd:
I'm just hesitant to make this change at the top level without more testing. I don't object to this as an English-specific change or to the French test case as being weird.

The overall quality of the UD GSD corpora is poor (I don't trust their tokenization decisions, basically) and we're using UD_French-Sequoia internally, not UD_French-GSD.

The tokenizers have been customized for the training corpora the provided models (as much as I could with a reasonable about of effort, anyway; there's still room for improvement). Even just for one language within UD, there's no way to provide a single good tokenizer because each corpus makes different decisions. For example, the UD German corpora don't agree on whether to split compounds on hyphens.

For all of the tokenizer customization (except French and Romanian, which have tricky hyphens), you're working in the range between like 99.5% and 100% and very small differences in tokenization can have much bigger effects than you'd expect further in the pipeline.

svlandeg:
Ok. I've moved the edit to English, it was probably long overdue that English got its own `punctuation` file anyway, instead of assuming the default equals English. I still think the edit would make sense for most languages though, but I'll leave it up to you to test further and check that out if you want.

(ugh, entirely wrong commit message, damnit)

adrianeboyd:
It would be nice for the English configuration to look more like other languages than having the defaults kind of be English, yes.

It looks like I was unnecessarily worried here. It doesn't look like this causes any problems for the training corpora (for many languages the default infixes are completely overridden anyway):

| | current | this PR |
| --- | --- | --- |
| da-core-news | 99.85 | 99.85 |
| de-dep-news | 99.97 | 99.97 |
| el-dep-news | 99.98 | 99.98 |
| en-core-web | 99.92 | 99.95 |
| es-dep-news | 99.99 | 99.99 |
| fr-dep-news | 99.82 | 99.82 |
| it-dep-news | 99.93 | 99.93 |
| lt-dep-news | 100 | 100 |
| nb-core-news | 99.8 | 99.8 |
| nl-dep-news | 99.96 | 99.96 |
| pl-dep-sz | 100 | 100 |
| pt-dep-news | 99.93 | 99.93 |
| ro-dep-mixed | 99.84 | 99.84 |

honnibal:
Great, thanks for testing! I had the same worries but it looks like this is addressed.

PR #5771, opened 2020-07-16
Option for returning only greedy matches

svlandeg:
## Description
It used to be so that the `*` and `+` operators of the `Matcher` were greedy, resembling the behaviour of regular expressions. This was changed in favour of returning all possible matches instead. I think that is still the best default behaviour, but many use-cases seem to require some sort of greedy matching. 

This PR implements that functionality as part of the Matcher API instead of requiring each user to implement a similar post-processing filter. There are currently three options when calling `matcher.add`:
- no additional argument keeps the current functionality as-is
- `greediness="FIRST"`: this is the same behaviour as regular expressions: the left-most, most greedy result (functionality as before)
- `greediness="LONGEST"`: return the longest match within the full string

In the string `BBAAAAAB`, the expression `BA*B` would result in `BB` by a regexp or with `FIRST`, but in `BAAAAAB` with `LONGEST`. 

Note that the filtering is done for each set of results with the same key, so in theory some overlapping results with different keys are still possible. Not sure what is better. We could also filter across all results, so that the results can be directly used in `doc.ents`?

### Old API matcher cleanup
I've also removed `*patterns` from the API, as this has been deprecated for some time. Needed to change some unit tests accordingly. Or do we still want to keep this?

### Types of change
enhancement

## Checklist


spacy/matcher/matcher.pyx
```diff
@@ -101,16 +103,15 @@ cdef class Matcher:
         key (str): The match ID.
         patterns (list): The patterns to add for the given key.
         on_match (callable): Optional callback executed on match.
-        *_patterns (list): For backwards compatibility: list of patterns to add
-            as variable arguments. Will be ignored if a list of patterns is
-            provided as the second argument.
+        greediness (str): "FIRST" or "LONGEST".
```
svlandeg, line 106: or maybe some kind of `Enum` orso?

adrianeboyd: I tried to use `Enum` for the Chinese tokenizer settings and it seemed like a total hassle so I went back to a flimsy tuple, but it's kind of unsatisfying and maybe I just didn't get the details quite right?

svlandeg: Yea, I loved Enum's in my Java days, but can't really get the hang of them in Python...

honnibal: Yeah the Python enum feels super pointless, I guess if we're typing stuff really carefully it could be helpful but we're not, so...

We're not using it elsewhere so I think it's best not to here, either.

spacy/matcher/matcher.pyx
```diff
@@ -225,11 +227,37 @@ cdef class Matcher:
...
+            span_filter = self._filter.get(key)
+            if span_filter is not None:
+                pairs = pairs_by_id.get(key, [])
+                pairs.append((start,end,end-start))
+                pairs_by_id[key] = pairs
+            else:
+                final_matches.append((key, start, end))
+        for key, pairs in pairs_by_id.items():
+            span_filter = self._filter.get(key)
+            matched = [False for i in range(length)]
```
honnibal, line 243: It's hard to know whether to worry about efficiency here, because different problem sets have different characteristics. I think it's reasonable to make some small adjustments to get ahead of potential problems.

We have some very nested looping here that is likely to be a problem. I suggest we just make the `matched` check a C array that we allocate once outside of the `pairs_by_id` loop, something like:

```
cdef Pool tmp_pool = Pool()
matched = <char*>tmp_pool.alloc(length, sizeof(char))
```

We would use `memset` to reset the array to 0 in the `pairs_by_id.items` loop, and then replace the innermost loop where you're just setting to `True` with `memset` as well. We could also define an empty array and use it as a target for `memcmp` to replace the `all` loop. This would give us something like this:

```
cdef Pool tmp_pool = Pool()
matched = <char*>tmp_pool.alloc(length, sizeof(char))
empty = <char*>tmp_pool.alloc(length, sizeof(char))

for key, pairs in pairs_by_id.items():
    memset(matched, 0, length * sizeof(matched[0]))
    for (start, end, span_len) in sorted_pairs:
        assert 0 >= start < end # Defend against segfaults
        assert (end-start) == span_len
        # If no tokens in the span have matched
        if memcp(&matched[start], &empty[start], span_len * sizeof(matched[0])) == 0:
            final_matches.append((key, start, end)) 
            # Mark tokens that have matched.
            memset(&matched[start], 1, span_len * sizeof(matched[0]))
```

It's always a bit uncomfortable making this type of optimization before running the code and finding that it's actually a problem. But unfortunately we just have to stay a bit vigilant, especially in the matcher where the dynamics of different problems can be really different. It's easy to have some code that's not a problem on anything we test, and then on some problems it runs for 100 years. I think the code here is O(n^2) but perhaps even cubic (I don't know what the guarantees are about number of matches vs match length), which gets glacially slow if the bottom of the loop is setting into a Python list.

For future reference, if the sorting also becomes a problem, we can replace those with a C++ `vector[pair[int, int]]` I think. I'd rather not though as that gets really ugly, hopefully the sorting that's there is okay, although with the lambda it might still get slow on large lists, not sure. An intermediate construct that's still Python but would be a bit faster is to construct the list so that it sorts in the desired order by default, i.e. you apply the negatives etc and change the ordering, and then you sort that transformed list.

honnibal:
Great idea, and really like the API! I think we should make the little efficiency adjustment mentioned above, but otherwise, really like this.

spacy/matcher/matcher.pyx
honnibal: Hm, why does it loop like that actually? Can we do:

```python
for (start, end, span_len, key) in sorted_pairs:
```

adrianeboyd: And maybe it's possible to generalize this a bit outside of this particular component? I wrote a reimplementation of `util.filter_spans` for `SpanC` for the tokenizer here, but maybe this all can be part of a general `SpanC` util function somewhere? https://github.com/explosion/spaCy/blob/b81a89f0a94ce5a191720ba0eccff43667da6ba9/spacy/tokenizer.pyx#L290-L304

honnibal: Good point, although generalizing this type of code is often really hard. If we can't immediately see how to make that work, I think it's okay to just sprinkle in a bit of C to address the problem.

svlandeg:
Thanks for the very helpful reviews!

I ran a benchmark on a string with 4945 matches, filtered to 2182 cases with `LONGEST` and to 2473 cases with `FIRST`. Timing (across thousands of repeats) is avg 0.009 without filtering, and 0.013-0.014 with filtering. I know these numbers are small, but the results are consistent.

If I implement the C arrays as laid out by @honnibal, the timing of the filtering with `FIRST` drops to 0.010, which I think is reasonable given the baseline. For `LONGEST`, it takes consistently longer, ca. 0.011-0.012. I'm hypothesizing this is because the original order in which the `Matcher` returns all its results, is much closer to the required ordering of `FIRST`.

As suggested, I also tried removing the lambda function by storing the tuples in such a way that a default sorting could be used, but this didn't seem to help, and made the code much less readable IMO.

> Hm, why does it loop like that actually? Can we do:
> 
> ```python
> for (start, end, span_len, key) in sorted_pairs:
> ```
I'm not sure I understand this suggestion, but my thinking was to sort only within the set of results for the same key, as I think it's more efficient to have multiple sorts on smaller sets, rather than one larger sort on the full set?

spacy/matcher/matcher.pyx
```diff
@@ -69,7 +73,7 @@ cdef class Matcher:
         """
         return self._normalize_key(key) in self._patterns
 
-    def add(self, key, patterns, *_patterns, on_match=None):
+    def add(self, key, patterns, *, on_match=None, greediness: str=None):
```
ines, line 76: How would you feel about renaming this argument to just `greedy`? 

I feel like `greediness` could easily invite misspellings or confusion because of the `y` &rarr; `i` transformation (especially for non-native speakers). "greedy" as in "greedy matching" on the other hand is terminology people are familiar with.

If we're worried that `greedy` as a keyword argument sounds too much like a boolean flag, we could make it `greedy_{something}`? Although I don't think `greedy="FIRST"` is that bad.

svlandeg: I don't mind too much either way. I initially did want to avoid just `greedy` because it indeed sounds like a boolean flag. But then on the other hand, the error message would be descriptive anough to help steer users in the correct direction:
> ValueError: [E947] Matcher.add received invalid 'greediness' argument: expected a value from ['FIRST', 'LONGEST'] but got: 'True'

honnibal:
Thanks for benchmarking! Sounds quite ready.

> I ran a benchmark on a string with 4945 matches, filtered to 2182 cases with LONGEST and to 2473 cases with FIRST

It can be really tricky with non-linear runtime though, because it's not really easy to make sure we're benchmarking with the worst cases. I think if you had something like `{"pattern": [{"IS_ALPHA": True}], "quantifier": "*"}` on a long document, you'd end up with a very large number of matches, many of which would be very long. And it's the length of the matches that we're looping over in the inner loop.

> I'm not sure I understand this suggestion, but my thinking was to sort only within the set of results for the same key, as I think it's more efficient to have multiple sorts on smaller sets, rather than one larger sort on the full set?

Okay yeah I think I misunderstood the loop there. The structure you have is fine.

PR #5772, opened 2020-07-17
Apply special cases to tokens split by infixes

### LINKED ISSUE #5598: Tokenizer special cases do not work around infix punctuation
I would expect the two sentences below to be tokenized the same way. However, in the second, the special cases for "won't" and "can't" do not work.

    >>> import en_core_web_sm
    >>> nlp = en_core_web_sm.load()
    >>> [token.text for token in nlp("I can't / won't tolerate that.")]
    ['I', 'ca', "n't", '/', 'wo', "n't", 'tolerate', 'that', '.']
    >>> [token.text for token in nlp("I can't/won't tolerate that.")] 
    ['I', "can't", '/', "won't", 'tolerate', 'that', '.']

## Your Environment

* **spaCy version:** 2.3.0
* **Python version:** 3.7.4

adrianeboyd:
## Description

* Combine methods for special case and cache checking to simplify code
* Check for special cases after splitting on infixes

Fixes #5598.

## Checklist

### DIFF
--- a/spacy/tests/lang/en/test_prefix_suffix_infix.py
+++ b/spacy/tests/lang/en/test_prefix_suffix_infix.py
@@ -37,6 +37,17 @@ def test_en_tokenizer_splits_uneven_wrap(en_tokenizer, text):
     assert len(tokens) == 5
 
 
+@pytest.mark.parametrize("text", ["can't/won't"])
+def test_en_tokenizer_infix_special_cases(en_tokenizer, text):
+    tokens = en_tokenizer(text)
+    assert len(tokens) == 5
+
+
+@pytest.mark.parametrize("text", ["(can't/won't?)"])
+def test_en_tokenizer_infix_special_cases_uneven_wrap(en_tokenizer, text):
+    tokens = en_tokenizer(text)
+    assert len(tokens) == 8
+
 @pytest.mark.parametrize("text,length", [("U.S.", 1), ("us.", 2), ("(U.S.", 2)])
 def test_en_tokenizer_splits_prefix_interact(en_tokenizer, text, length):
     tokens = en_tokenizer(text)
--- a/spacy/tokenizer.pxd
+++ b/spacy/tokenizer.pxd
@@ -34,9 +34,9 @@ cdef class Tokenizer:
                                        vector[SpanC] &filtered)
     cdef int _retokenize_special_spans(self, Doc doc, TokenC* tokens,
                                        object span_data)
-    cdef int _try_cache(self, hash_t key, Doc tokens) except -1
-    cdef int _try_specials(self, hash_t key, Doc tokens,
-                           int* has_special) except -1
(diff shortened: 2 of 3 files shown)


honnibal:
Looks good, and it'll be nice to have this inconsistency resolved. The infixes weren't part of the original tokenizer design so I was never 100% sure how they should interact with the special-cases.

Have you evaluated the change to check what tokenizes differently in a sample of text? I don't think the runtime will be affected but it's good to check that quickly as well.

I'm also a little worried about how breaking this might be for people's tokenization rules. It's awkward to add an option, but if this makes a big difference I suppose that's something we could consider too.

adrianeboyd:
I benchmarked the speed in advance (no measurable difference), but you're right that it needs a more thorough evaluation for all the training corpora. (I think it's going to be rare enough that it's hard to find cases where it makes a difference.)

The other option is to add an option to extend the matcher-based special case handling. Right now the matcher-based special cases are only those that contain affix punctuation (for speed), but you could add an option to use the matcher-based handling for all the special cases, which would address this. It would be slower overall, but would potentially be more consistent. I'm not sure whether I'd want fast-without-infixes and slow-with-infixes to be the setting, though. The way the matcher special cases work, though, you couldn't get slow-without-infixes, though, since it matches over all the tokens, so you could get fast-with-infixes, fast-without-infixes, or slow-with-infixes, if that makes sense.

adrianeboyd:
Wow, I was really wrong about this:

| | current | this PR |
| --- | --- | --- |
| da-core-news | 99.85 | 99.84 |
| de-dep-news | 99.97 | 99.85 |
| el-dep-news | 99.98 | 99.97 |
| en-core-web | 99.92 | 99.43 |
| es-dep-news | 99.99 | 99.99 |
| fr-dep-news | 99.82 | 99.16 |
| it-dep-news | 99.93 | 98.10 |
| lt-dep-news | 100 | 100 |
| nb-core-news | 99.8 | 99.78 |
| nl-dep-news | 99.96 | 99.96 |
| pl-dep-sz | 100 | 99.91 |
| pt-dep-news | 99.93 | 99.86 |
| ro-dep-mixed | 99.84 | 99.84 |

I guess we could have a slow-with-infixes option for people who really wanted the more consistent behavior in a particular instance, but I don't think we should make this the default.

I ran it on some English wikipedia text and in a lot of cases it's better, like contractions next to emdash or `/` or ellipsis, but there are some like German names with `-im-` in the middle where it analyzes it like `I'm`. (Possibly the `im` `wed` `id` cases should be removed as default exceptions, though. I'd be annoyed at this default personally.)

honnibal:
Ouch :(

Okay I think this will be too breaking, especially to users with special cases. Should we just close this?

adrianeboyd:
I would keep the simplified code bits and consider adding an option for the thorough matcher special case handling, since I think it could be useful in some cases to have this be consistent rather than like the current tokenizer.

PR #5787, opened 2020-07-20
Initial draft of Morphologizer API docs

adrianeboyd:
## Description

Initial draft of Morphologizer API docs

### Types of change

Docs.

## Checklist

### DIFF
--- a/website/docs/api/morphologizer.md
+++ b/website/docs/api/morphologizer.md
@@ -5,10 +5,14 @@ source: spacy/pipeline/morphologizer.pyx
 new: 3
 ---
 
-A trainable pipeline component to predict morphological features. This class is
-a subclass of `Pipe` and follows the same API. The component is also available
-via the string name `"morphologizer"`. After initialization, it is typically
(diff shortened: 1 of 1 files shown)


honnibal:
Looks good!

I'd like to explain the "factories" ideas elsewhere in the docs, when we do we can change the term "string name" to "factory name" in these places too. It will be good to be careful about this distinction, because you'll be able to have multiple components with different names that are created by the same factory, e.g. multiply text classifiers or entity recognizers.

We'll also want a linguistic features section that gives more of an overview of using morphology: what it is, how the API works, etc. But I think there's no problem with merging this now.

adrianeboyd:
Yeah, this was just intended to be the API docs. I know the pipe creation stuff will change, but I wanted to fill in the main API parts. I'm waiting for the config dust to settle before looking at the usage docs.

ines:
@adrianeboyd Yeah, let's do the usage docs last – I'm currently thinking about how to best organise them and if we need a separate page explaining the config and function registries.

We can already collect general usage examples but a lot of the "How to" stuff will depend on how we end up introducing the new conceptual model.

PR #5791, opened 2020-07-21
Add Morphology and MorphAnalysis API docs

adrianeboyd:
## Description

Add initial draft of `Morphology` and `MorphAnalysis` API docs.

### Types of change

Docs.

## Checklist


website/meta/sidebars.json
```diff
@@ -70,6 +70,7 @@
                     { "text": "Token", "url": "/api/token" },
                     { "text": "Span", "url": "/api/span" },
                     { "text": "Lexeme", "url": "/api/lexeme" },
+                    { "text": "MorphAnalysis", "url": "/api/morphanalysis" },
```
ines, line 73: Maybe we should move this down below Morphology in the "Other" section? Even though the `MorphAnalysis` is technically a container, it's fairly abstract and probably not something most users will interact with directly.

adrianeboyd: I thought it seemed most similar to `Lexeme`, but I can see what you mean. You do get `MorphAnalysis` as `token.morph`, though, so it's not that obscure.

adrianeboyd: (I also think the sidebars are getting long enough that alphabetical might be better. I have trouble finding stuff sometimes.)

PR #5794, opened 2020-07-21
Improve warnings around normalization tables

adrianeboyd:
## Description

Provide more customized normalization table warnings when training a new model. Only suggest installing `spacy-lookups-data` if it's not already installed and it includes a table for this language (currently checked in a hard-coded list).

Possibly the first warning about the empty normalization table could be removed, it's kind of a lot of warnings in the end.

### Types of change

Enhancement.

## Checklist


honnibal:
I really would like these to change to debug logging statements instead, it definitely feels like a lot of warnings. Merging this for now though because we probably want to think about using logging more deeply.

PR #5798, opened 2020-07-22
Update language data config and lookups integration

ines:
## Description

- Remove tags and lemmas from tokenizer exceptions and tidy up language-specific data.
- Make `lookups_extra` and `omit_extra_lookups` mechanisms redundant.
- Simplify language data again and move defaults back to `LanguageDefaults`, rather than constructing all objects via config.
- `Vocab` and `Lemmatizer` now own their own relevant lookups.
- Orth variants now live in lookups data.

See the `feature/spacy-v3` branch on `spacy-lookups-data` for the updated data: https://github.com/explosion/spacy-lookups-data/tree/feature/spacy-v3

### Types of change

enhancement

## Checklist


spacy/gold/augment.py
```diff
@@ -25,8 +25,9 @@ def make_orth_variants(nlp, raw_text, orig_token_dict, orth_variant_level=0.0):
         lower = True
         if raw is not None:
             raw = raw.lower()
-    ndsv = nlp.Defaults.single_orth_variants
-    ndpv = nlp.Defaults.paired_orth_variants
+    orth_variants = nlp.vocab.lookups.get_table("orth_variants", {})
+    ndsv = orth_variants.get("single", [])
+    ndpv = orth_variants.get("pairsed", [])
```
adrianeboyd, line 30: typo: `paired`

spacy/default_config.cfg
```diff
@@ -2,13 +2,15 @@
 lang = null
 stop_words = []
 lex_attr_getters = {}
+vocab_data = {}
```
adrianeboyd, line 5: I'm wondering whether it would make sense to separate `nlp.vocab` section here?

adrianeboyd:
It's nice to get rid of the hacky extra lookups!

> * We should probably add a lazy-loading mechanism to the extra lookups because it otherwise makes things very slow?

I kind of feel like if you've included them in your config you want them to be loaded and you can wait 3 seconds, but maybe there's a good way to do it lazily.

> * Should languages define the extra lexeme data etc. as a default in their configs, or is this something the user/we can add if needed?

I don't think the extra tables should be included by default. As an alternative to any lazy loading, you could just have methods that load them explicitly?

> * How should serialization work? That ties in with the more deeper questions around components serializing their data, and how to express this via the config.

At the very least you want the lemmatizer to have its own serialization of its own lookup tables, right?

ines:
> I don't think the extra tables should be included by default. As an alternative to any lazy loading, you could just have methods that load them explicitly?

Okay, in that case, the way it work now is okay, I guess? If you include them in your config, the table are loaded when the vocab is created and you might have to wait a few seconds. If you don't (default), they're not loaded at all.

> At the very least you want the lemmatizer to have its own serialization of its own lookup tables, right?

Yep, then it can just call `lookups.to_disk`. It probably doesn't make much sense to implement this on the current lemmatizer at the moment if we're going to change it all anyways?

spacy/default_config.cfg
ines: I was thinking about this too, yes! For now, I tried to avoid adding more subsections and I wasn't sure what should go under `vocab` (as opposed to the top level `nlp` – like, is `stop_words` vocab? Or `lex_attr_getters`?)

spacy/default_config.cfg
```diff
@@ -2,13 +2,18 @@
...
+get_noun_chunks = null
 pipeline = []
 
 [nlp.tokenizer]
 @tokenizers = "spacy.Tokenizer.v1"
```
adrianeboyd, line 10: Why aren't all the tokenizer settings explicit here?

spacy/default_config.cfg
adrianeboyd: Yeah, there's not other than meta much at the top level if you do this, I guess. Maybe this could just be `data` then? What non-vocab data is there potentially? (Could this just be `data` to be more like the other sections?)

spacy/default_config.cfg
ines: I haven't integrated the exceptions/prefixes/suffixes/infixes yet – those are currently still read off the `Language.Defaults` in the tokenizer factory and not passed in from the function registry. So the tokenizer currently only has two settings, `token_match` and `url_match`.

adrianeboyd: But they will be handled the same in the end?

ines: Yeah, that's the idea. So languages would write:

```ini
[nlp.tokenizer]
@tokenizers = "spacy.Tokenizer.v1"
token_match = null
url_match = {"@language_data": "spacy.xx.url_match"}
exceptions = {"@language_data": "spacy.en.tokenizer_exceptions"}
# etc.
```

PR #5803, opened 2020-07-22
Mark Japanese documents as tagged.

### LINKED ISSUE #5802: JA model drops POS tags when using displacy.
## How to reproduce the behavior
The error is visible at 
https://explosion.ai/demos/displacy?text=若葉は薄い色合いから日に日に濃さを増している%E3%80%82&model=ja_core_news_sm&cpu=1&cph=1
```
import spacy
from spacy import displacy

text = "若葉は薄い色合いから日に日に濃さを増している。"
nlp = spacy.load("ja_core_news_lg")
doc = nlp(text)
displacy.render(doc, style="dep")
```

This seems to be caused by 
https://github.com/explosion/spaCy/blob/v2.3.2/spacy/displacy/__init__.py#L122
`    doc = Doc(orig_doc.vocab).from_bytes(orig_doc.to_bytes(exclude=["user_data"]))`

When I drop into `pdb` and examine the object `orig_doc` has the pos's tagged, but `doc` doesn't and it doesn't seem to be as simple as the fact that the Japanese model stores meta-data in the user_data dictionary.

## Your Environment

* **spaCy version:** 2.3.2
* **Platform:** macOS-10.15.5-x86_64-i386-64bit
* **Python version:** 3.8.2
* Environment Information:

solarmist:
Mark the document as tagged before returning it to the user from the JapaneseTokenizer.
Fixes #5802

### Types of change

Bug fix

## Checklist

### DIFF
--- a/.github/contributors/solarmist.md
+++ b/.github/contributors/solarmist.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Thanks for this PR!

PR #5809, opened 2020-07-24
Add Lexical Attribute and Tag-Map for Gujarati Language

yash1994:
## Description
Added Lexical attribute (like_num) and Part-of-speech tag map for the Gujarati language.

### Types of change
An enhancement.

## Checklist


adrianeboyd:
Hi, thanks for this PR! We're going to remove the tag maps from the core library soon (because they're specific to a particular corpus rather than to a particular language), so we'd be happy to accept this PR without the tag map and related settings.

yash1994:
Hi, i am currently working on morphological rule and tokenizer exceptions, we can close this PR and I will open a new one when I am done with morph rules and tokenizer exceptions (without tag map). Let me if know if this works for you. Also thanks for looking into this.

svlandeg:
@yash1994 : sounds good! Looking forward to your updated PR.

PR #5820, opened 2020-07-26
Update examples.py

oculusrepairo:
adding factual sentences to the list

## Checklist


adrianeboyd:
Thanks for this contribution! I'll the commas that are missing to separate the sentences, but otherwise this looks good.

PR #5833, opened 2020-07-28
Small adjustments to Scorer and docs

ines:
## Description

Types, formatting and docs. Also moved the default pipeline and language to arguments. Also, as @honnibal mentioned, we could consider breaking the `Scorer` up into functions and moving `Scorer.score` to `Language.score`.

### Types of change
enhancement

## Checklist


adrianeboyd:
In terms of the docs: all those score method arguments aren't keyword-only in their current form, are they?

PR #5834, opened 2020-07-28
Pretrained vectors for v3

honnibal:
In v2 the vectors worked by passing a string name into the `Tok2Vec` function that builds that part of the model. We then had a global variable behind the scenes that allows the `StaticVectors` function within Thinc to retrieve the vectors. Obviously, not good.

I went through a few iterations of fixing this, and I ended up with an implementation that relies on the `StaticVectors` layer seeing the `Doc` objects. This has some efficiency disadvantages, as it makes it harder to use the `uniqued` layer like we were to calculate the word representations per type per batch. However, I think it's worth it overall, especially since we no longer rely on that mechanism of setting the row on the lexemes.

I've refactored the `Tok2Vec` layers quite a lot to get this to work, including changing the config structure of them. The `spacy.Tok2Vec.v1` layer now specifies a separate `embed` and `encode` layer. The changes will also slightly affect what models compute. The difference in accuracy will be small but it might show up in some fine-grained testing.

The mechanism for actually specifying the vectors during training is basically the same as v2. We allow you to pass in a path or the name of a spaCy model. The only difference is you do it in the config instead of on the CLI.

## Checklist


spacy/cli/train.py
```diff
@@ -81,16 +81,20 @@ def train(
...
     if config["training"]["base_model"]:
-        base_nlp = util.load_model(config["training"]["base_model"])
         # TODO: do something to check base_nlp against regular nlp described in config?
-        nlp = base_nlp
+        # If everything matches it will look something like:
+        # base_nlp = util.load_model(config["training"]["base_model"])
+        # nlp = base_nlp
+        raise NotImplementedError("base_model not supported yet.")
+    if config["training"]["vectors"] is not None:
+        util.load_vectors_into_model(nlp, config["training"]["vectors"])
```
honnibal, line 95: @ines This part is worth noting, in case you want to recommend a different interaction with the config stuff. It's kind of ugly to do it so explicitly like this, but see the comment in `Language` below.

spacy/util.py
```diff
@@ -187,6 +189,20 @@ def get_module_path(module: ModuleType) -> Path:
     return Path(sys.modules[module.__module__].__file__).parent
 
 
+def load_vectors_into_model(
```
honnibal, line 192: @adrianeboyd Do you know of anything else we need to do when loading the vectors, other than these things?

We no longer have to set the `lexeme.rank` stuff.

PR #5837, opened 2020-07-29
Bugfix tagmap state

svlandeg:
## Description
A bug in the test suite (accumulating `tag_map` state on `en_tokenizer`) was obscuring a bug in one specific test (which should use `pos` instead of `tags` in `get_doc`). The test however did fail when ran individually.

This PR fixes both bugs + removes an instance of a hidden default in `self.cfg.get("set_morphology", True)`

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/pipeline/tagger.pyx
+++ b/spacy/pipeline/tagger.pyx
@@ -152,7 +152,7 @@ class Tagger(Pipe):
         cdef Doc doc
         cdef int idx = 0
         cdef Vocab vocab = self.vocab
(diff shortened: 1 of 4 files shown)


adrianeboyd:
I've removed/modified all of these spots in #5848 in removing the tag map from the vocab, so this may not be needed in the end?

svlandeg:
I saw you also fixed that broken unit test in that new PR, so yea that should be fine then.

honnibal:
Okay closing this in preference of #5848

PR #5842, opened 2020-07-30
Add AttributeRuler for token attribute exceptions

adrianeboyd:
## Description

Add the `AttributeRuler` to handle exceptions for token-level attributes. The `AttributeRuler` uses `Matcher` patterns to identify
target spans and applies the specified attributes to the token at the provided index in the matched span. A negative index can be used to index from the end of the matched span.

Helper functions can import existing tag maps and morph rules to the corresponding `Matcher` patterns.

Additional bugfixes / refactoring:
* Fix minor bug for `MORPH` attributes in the retokenizer to correctly normalize the values and to handle `MORPH` alongside `_` in an attrs dict.
* Add index bounds checks when indexing in a `Span`.
* Refactor retokenizer methods for normalizing and setting attributes on a single token for use here.

Docs to come after API is approved/stable.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
(Something non-deterministic is going on with the retokenizer, yay.)

spacy/pipeline/attributeruler.py
```diff
@@ -0,0 +1,252 @@
...
+    "attribute_ruler",
+    assigns=[],
+    default_config={},
+    scores=[],
+    default_score_weights={},
+)
+def make_attribute_ruler(
+    nlp: Language, name: str, pattern_dicts: Iterable[Dict] = tuple()
+):
+    return AttributeRuler(nlp.vocab, name, pattern_dicts=pattern_dicts)
```
ines, line 29:
```suggestion
    nlp: Language, name: str
):
    return AttributeRuler(nlp.vocab, name)
```

We probably shouldn't allow the `pattern_dicts` to be passed in via the config – otherwise, you'll end up with those values in your `config.cfg` (if you set them on initialization with `nlp.add_pipe`).

Instead, I think the recommended usage should be:

```python
ruler = nlp.create_pipe("attribute_ruler")
ruler.add(patterns)
```

spacy/pipeline/attributeruler.py
```diff
@@ -0,0 +1,252 @@
...
+
+MatcherPatternType = List[Dict[Union[int, str], Any]]
+
+
+@Language.factory(
+    "attribute_ruler",
+    assigns=[],
+    default_config={},
+    scores=[],
+    default_score_weights={},
```
ines, line 24:
```suggestion
```

Possible nitpick, but if the component isn't setting any of those, we might as well not set them here?

spacy/pipeline/attributeruler.py
adrianeboyd: This format may be a bad idea, but I think you need to be able to provide an asset containing patterns in the config somehow, similar to how you used to provide a tag map?

ines: Yeah, fair point. 

I think this ties into a more general question of how we can (and whether we should) find a way to enforce the following:

```python
# Bad
nlp.create_pipe("attribute_ruler", config={"patterns": [[{"TEXT": "foo"}]]})

# Good
nlp.create_pipe("attribute_ruler", config={"patterns": {"@assets": "my_patterns"}})
```

adrianeboyd: Is there a technical reason to want to disallow patterns to be specified in the config? Is there a disadvantage other than potentially having a config that's long and hard to read?

And looking at the APIs, whatever we choose here should also be used for `EntityRuler`.

adrianeboyd:
I'm running into a bit of a naming problem if I try to make this more similar to `EntityRuler`. The idea here in `add` is that you can have more than one `Matcher` pattern for a given rule, so these are `patterns`, but this isn't similar to what `patterns` means in `EntityRuler`, where they refer to an entry in the list of dicts that contain `label/pattern`. I have `pattern_dicts` instead here, but this also isn't great.

spacy/pipeline/attributeruler.py
ines: In the case of the patterns, it's mostly just that it clogs up the config and once it's in there, you can't really get it out. And we don't want people to get the idea that this is how their patterns should be stored – especially since they can get quite verbose. But I'd be willing to accept this as an unfortunate side-effect.

Where it gets trickier is when objects are JSON-serializable, because then you'd end up with their serialized version in your config, not the actual reference to the object. For example:

```python
from spacy.lookups import Lookups
from thinc.api import Config

config = Config({"config": {"lookups": Lookups()}})  
print(config.to_str())
```

spacy/pipeline/attributeruler.py
```diff
@@ -0,0 +1,268 @@
...
+
+        RETURNS (AttributeRuler): The AttributeRuler component.
+
+        DOCS: https://spacy.io/api/attributeruler#init
+        """
+        self.name = name
+        self.vocab = vocab
+        self.matcher = Matcher(self.vocab)
+        self.attrs = []
+        self.attrs_unnormed = [] # store for reference
```
honnibal, line 61: Can we at least make this "private", i.e. `_attrs_unnormed`? If it needs to be part of the API we can think more about the naming.

adrianeboyd: Sure. (I don't even really want to store it, but then it's hard to return legible patterns.)

honnibal:
Go ahead and merge when you're ready 👍

PR #5844, opened 2020-07-30
The Parser is now a Pipe (2)

svlandeg:
## Description
- Have `Parser` inherit from `Pipe`. Due to the `cdef` for the `Parser`, I could only get this to work with a `pipe.pxd`, but I think that's fine? Or am I missing a more easy solution?
- Obliterated the `syntax` folder - hooray!
   - the old `nn_parser` is moved to `transition_parser` in `spacy.pipeline`
   - the old `_parser_model` is moved to `spacy.models` as `parser_model`
   - everything else is now in `spacy.pipeline._parser_internals`
- Made some small code fixes in some of the `Pipe` classes
- Removed unnecessary imports

### Types of change
enhancement / refactor

## Checklist


honnibal:
Awesome! Yes, a `pipe.pxd` makes perfect sense.

PR #5829, opened 2020-07-28
Hindi: Adds tests for lexical attributes (norm and like_num)

rahul1990gupta:
This PR 
- adds some tests for lexical attribute function like_num 
- adds some tests for lexical attribute function norm 
- In norm function, it replaces `break` by `continue` which allows processing smaller words which it was not processing earlier. Added tests to ensure that. 
- It expands Hindi Number vocabulary. 
- It adds vocabulary to recognize ordinal numbers in Hindi as mentioned in the issue https://github.com/explosion/spaCy/issues/5827
 
### Types of change

- It fixes bugs in like_num and norm functions 
- It adds tests for both function 
- Expands like_num vocabulary. Not sure, if that qualifies as an enhancement.

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #5847, opened 2020-07-30
[Hindi] Breaking pronouns

rahul1990gupta:
This PR adds a rule for breaking up pronouns in Hindi 

This will help an ML algorithm help better understand the sentence structure. Consider these two sentences with very similar structure. 
```
राम ने करण को आम बेचे। 
मैंने आपको आम बेचे।  
```
For machine to be able to understand structure better, these two should be broken down as 
```
राम ने करण को आम बेचे। 
मैं ने आप को आम बेचे।  
```

This PR should be merged after https://github.com/explosion/spaCy/pull/5829
### Types of change

Enhancement
## Checklist

### DIFF
--- a/spacy/lang/hi/__init__.py
+++ b/spacy/lang/hi/__init__.py
@@ -1,17 +1,20 @@
 # coding: utf8
 from __future__ import unicode_literals
 
+from .tokenizer_exceptions import TOKENIZER_EXCEPTIONS
 from .stop_words import STOP_WORDS
 from .lex_attrs import LEX_ATTRS
 
 from ...language import Language
 from ...attrs import LANG
+from ...util import update_exc
 
 
 class HindiDefaults(Language.Defaults):
     lex_attr_getters = dict(Language.Defaults.lex_attr_getters)
     lex_attr_getters.update(LEX_ATTRS)
     lex_attr_getters[LANG] = lambda text: "hi"
+    tokenizer_exceptions = update_exc(TOKENIZER_EXCEPTIONS)
     stop_words = STOP_WORDS
 
 
--- a/spacy/lang/hi/tokenizer_exceptions.py
+++ b/spacy/lang/hi/tokenizer_exceptions.py
@@ -1,0 +1,21 @@
+# coding: utf8
+from __future__ import unicode_literals
+
+from ...symbols import ORTH, LEMMA, TAG, NORM, PRON_LEMMA
+
+# Hindi Pronoun: https://en.wiktionary.org/wiki/Category:Hindi_pronouns
+# https://hi.wikipedia.org/wiki/कारक 
+
+_exc = {
+    "मैंने": [
+        {ORTH: "मैं", LEMMA: PRON_LEMMA, NORM: "मैं"},
+        {ORTH: "ने", LEMMA: "ने", NORM: "ने"},
+    ],
+    "आपको":[
+        {ORTH: "आप", LEMMA: PRON_LEMMA, NORM: "आप"},
+        {ORTH: "को", LEMMA: "को", NORM: "को"},
+    ]
+} 
+
+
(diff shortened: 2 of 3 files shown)


adrianeboyd:
Hi, I'm slightly unsure because I don't know any Hindi, but I think these exceptions are accidentally changing the underlying text. In these pronoun combinations, `left != right1 + right2`, correct?

It's fine if you want to preprocess your texts outside spacy to make these kinds of changes, but spacy itself should never modify the provided text. So, in all tokenizer exceptions, the `ORTH` values on the right should concatenate to match the form on the left exactly.

In spacy v2 there unfortunately aren't any checks that prevent you from accidentally modify the text with tokenizer exceptions (there will be in v3!), but we want to avoid adding any exceptions like this.

rahul1990gupta:
Thanks @adrianeboyd  for the review. This makes total sense. I see the same pattern being followed here as well. 
https://github.com/explosion/spaCy/blob/master/spacy/lang/en/tokenizer_exceptions.py#L94-L97

What do you think about breaking prounouns like this for `orth` and `lemma` ?
मेरा = मे + रा [ orth breakup]
मेरा = मैं + का [lemma breakup]

honnibal:
@rahul1990gupta Is there a transliteration system we could use to talk about this with the latin alphabet? I'm having trouble telling whether `मेरा = मे + रा ` changes the characters, because I'm not sure whether there are ligatures :(. Sorry we don't have anyone more experienced with Hindi, please bear with us.

rahul1990gupta:
My bad. I will write in English as well as Hindi from now on. 

> मेरा = मे + रा [ orth breakup]
> मेरा = मैं + का [lemma breakup]

With transliteration it will look like this 
Mera = me + ra [for orth breakup]
Mera = main + ka [for lemma breakup]
Where `me` and `ra` are **NOT** valid Hindi dictionary words, while `main` and `ka` are Hindi dictionary words.

BTW, for Hindi transliteration google input is the best one I know. 
https://www.google.com/inputtools/try/

adrianeboyd:
The most important part is that the right-side orths add up to the left-side side with string concatentation. We also don't want to set `LEMMA` here anymore. This kind of tokenizer exception should set `NORM` instead.

If you're familiar with traditional tokenization for English `can't` as `ca` + `n't`, this should be similar. We don't want `ORTH` as `can` + `n't` or `can` + `not` because that would change the underlying text, but you can have `ORTH` `ca` + `n't` with the `NORM` `can` + `not`. The default models use `NORM` as a feature so they'll know that `ca` is similar to `can`.

honnibal:
> My bad. I will write in English as well as Hindi from now on.

Not at all, thanks for accommodating.

By the way, is there a widely used lossless Romanization system for Hindi? I think it would be really nice to have lossless romanization as an optional lexical attribute. By lossless I mean that the Romanization transformation should be reversible, and it should always result back at the original string.

rahul1990gupta:
@honnibal 
I tested romanization from [indic_transliteration](https://github.com/sanskrit-coders/indic_transliteration) here and it looks good. It uses international standards(e.g. Harvad-Kyoto) which is great. 
https://gist.github.com/rahul1990gupta/6f84dd20b3b95c4a9318b16e3426727e

It's under MIT license (which we can use). I can pull out the bits essential for romanization of Hindi texts and create a PR, if that would be useful. 

btw, Does spacy have romanization for other non-roman languages too ?

id18522096554 (approved):

svlandeg:
Unfortunately this PR has gone quite a bit out of date with the recent release of spaCy 3.0. The tokenizer exceptions also work a little bit differently now, so I'm actually unsure whether we still need this change? This might be resolved in 3.x.

If that's not the case though - feel free to open a new PR against the current `master`!

PR #5848, opened 2020-07-30
Add Lemmatizer and simplify related components

adrianeboyd:
## Description

* Add `Lemmatizer` pipe with `lookup` and `rule` modes using the `Lookups` tables.
* Reduce `Tagger` to a simple tagger that sets `Token.tag` (no pos or lemma)
* Reduce `Morphology` to only keep track of morph tags (no tag map, lemmatizer, or morph rules)
* Remove lemmatizer from `Vocab`
* Adjust many many tests

Differences:

* No default lookup lemmas
* No special treatment of TAG in `from_array` and similar required
* Easier to modify labels in a `Tagger`
* No extra strings added from morphology / tag map

In progress:

- [x] `Lemmatizer` config / serialization
- [x] Incorporating language-specific lemmatizers and `is_base_form`
- [x] Removing the previous lemmatizer (temporarily renamed to `OldLemmatizer`)
- [ ] `Tagger`/`Morphologizer` treatment of space tokens
- [x] Test lemmatizers with `spacy-lookups-data` (as much as possible, one French case fails due to removing lemmas from tokenizer exceptions)

### Types of change

Enhancement

## Checklist


honnibal:
Fast work! This looks really close I think. And definitely a massive improvement.

adrianeboyd:
This is also still clearly the slowest version of the current rule-based lemmatizer. It needs a cache, etc.

spacy/lang/en/__init__.py
```diff
@@ -46,4 +25,24 @@ class English(Language):
...
+def make_lemmatizer(
+    nlp: Language,
+    model: Optional[Model],
+    name: str,
+    mode: str,
+    lookups: Optional[Lookups],
+):
+    tables = ["lemma_lookup", "lemma_rules", "lemma_exc", "lemma_index"]
+    if lookups is None:
+        lookups = load_lookups(lang=nlp.lang, tables=tables)
```
ines, line 44: Now that we have the lemmatizer set up as a component, we could consider making `load_lookups` strict in the lemmatizers if `mode` is set to `"rule"` or `"lookup"` and no custom lookups are provided?

If the user doesn't have `spacy-lookups-data` installed, they'd see a useful error or at least a warning telling them to install it or provide their own rules because otherwise, there won't be any lemmas.

adrianeboyd: I think it's currently strict by default? I definitely saw errors about missing tables here if `spacy-lookups-data` wasn't installed.

ines: Ahh sorry, I think I remembered the default wrong. In that case, we could probably do something like this?

```python
if lookups is None:
    strict = mode in ("rule", "lookup")
    lookups = load_lookups(lang=nlp.lang, tables=tables, strict=strict)
```

And then have `load_lookups` actually raise if `spacy_lookups_data` isn't installed.

adrianeboyd: I tried to redo it like this and be more specific about the tables for each mode, but it still feels a little weird overall. I think it might be better to just not load any tables for modes that `make_lemmatizer` doesn't know about? I also changed the mode setting to make it immutable, which means you can't switch modes on the fly, but it makes the initialization a lot cleaner.

ines: I think it's fine to ignore the tables but maybe we could raise if you add a lemmatizer to the pipeline with mode `rule` or `lookup` for a language that we don't have any data for? Then the user gets instant feedback that lemmatization isn't going to produce any results out-of-the-box and that they need to provide their own function that returns lookup tables.

> I also changed the mode setting to make it immutable, which means you can't switch modes on the fly, but it makes the initialization a lot cleaner.

Yeah, that makes sense!

spacy/lang/nl/__init__.py
```diff
@@ -46,4 +29,24 @@ class Dutch(Language):
...
+)
+def make_lemmatizer(
+    nlp: Language,
+    model: Optional[Model],
+    name: str,
+    mode: str,
+    lookups: Optional[Lookups],
+):
+    config = dict(LOOKUPS_TABLES_CONFIG)
+    config["rules"] = ["lemma_lookup", "lemma_rules", "lemma_exc", "lemma_index"]
```
ines, line 47: I think this needs a `"required_tables"` key?

Also I wonder if we could streamline this part a bit so we don't have to keep passing around the `LOOKUPS_TABLES_CONFIG` and copy it, read from nested dicts etc.

There are basically 4 lists here, right? Optional and required tables, for rules and lookups? Maybe we could just make them arguments on `load_lemmatizer_lookups` then?

adrianeboyd: Ah, that didn't get updated correctly, thanks.

Yes, I think it would cleaner to have the required and optional tables as the arguments to `load_lemmatizer_lookups` instead of the whole config.

The reason for passing it around the config (I agree it's clunky) is that you need to know what the tables are when you try to create something like a lookup lemmatizer for `Dutch` without having to add anything to `DutchLemmatizer`. Maybe the defaults should be stored as class variables in `Lemmatizer` instead? I was trying to make it so that you didn't have to override `__init__` or `__call__` when extending the `Lemmatizer`, so for your Dutch rule-based lemmatizer you just had to implement `rule_lemmatize`.

Let me see what I can do...

spacy/pipeline/lemmatizer.py
```diff
@@ -0,0 +1,277 @@
...
+class Lemmatizer(Pipe):
+    """
+    The Lemmatizer supports simple part-of-speech-sensitive suffix rules and
+    lookup tables.
+
+    DOCS: https://spacy.io/api/lemmatizer
+    """
+
+    @classmethod
+    def get_lookups_config(cls, mode: str) -> Dict:
```
honnibal, line 50: Is it correct to be more specific on the `Dict` here?

```suggestion
    def get_lookups_config(cls, mode: str) -> Dict[str, List[str]]:
```

spacy/pipeline/lemmatizer.py
```diff
@@ -0,0 +1,277 @@
...
+        name (str): The component name. Defaults to "lemmatizer".
+        mode (str): The lemmatizer mode: "lookup", "rule". Defaults to "lookup".
+        lookups (Lookups): The lookups object containing the (optional) tables
+            such as "lemma_rules", "lemma_index", "lemma_exc" and
+            "lemma_lookup".
+        """
+        self.vocab = vocab
+        self.model = model
+        self._mode = mode
+        self.lookups = lookups if lookups is not None else Lookups()
```
honnibal, line 111: Just to double-check, will we be okay to modify a `Lookups` object that's passed in here? For instance, do we ever have one in a global scope, e.g. in tests?

spacy/pipeline/lemmatizer.py
```diff
@@ -0,0 +1,277 @@
...
+            if string.endswith(old):
+                form = string[: len(string) - len(old)] + new
+                if not form:
+                    pass
+                elif form in index or not form.isalpha():
+                    forms.append(form)
+                else:
+                    oov_forms.append(form)
+        # Remove duplicates but preserve the ordering of applied "rules"
+        forms = list(dict.fromkeys(forms))
```
honnibal, line 205: Wow, nice trick

adrianeboyd: I didn't touch this part!

honnibal:
Looking good! It's a big PR, really fast work.

Is the speed issue resolved? And is there anything I should look at more closely?

spacy/pipeline/lemmatizer.py
adrianeboyd: I was trying to leave it more open-ended. You could implement a lemmatizer with its own set of `get_lookups_config` and `load_lookups` that don't follow this exactly. (I have to say I'm not totally sure how it works out in terms of the typing, though.)

spacy/pipeline/lemmatizer.py
honnibal: Huh, so you didn't: https://github.com/explosion/spaCy/commit/955b95cb8ba5185525e736089f7c81743f32efa0

At least I was right that I didn't remember writing it...

spacy/pipeline/lemmatizer.py
adrianeboyd: I think so, but I don't really know for sure. I don't really understand the scope of registry stuff, but I moved all the lookups assets in the tests within the methods.

adrianeboyd:
I added a very simple cache for the rule-based lemmatizer, no cache anywhere else. The cache potentially needs a max size or something like that.

I wouldn't want to implement a cache in a way that made it hard for people to implement their own rule-based lemmatizers and use the cache, and very very preferably that they aren't forced to implement anything in cython in general. What do you think the possibilities are?

spacy/pipeline/lemmatizer.py
adrianeboyd: I did run into problems with global assets in the tests because I had two things with the same name that resolved to one of them and broke the other test.

honnibal:
Let's start using it without much caching and see what the performance looks like in practice. So I think it's fine to keep it like this for now.

adrianeboyd:
Just on my laptop, the speed difference for `en-core-web` `train` between just a tagger + attribute ruler (to do the POS mapping) and that plus a rule-based lemmatizer is pretty small: 7.9K words/s vs. 7.6 words/s. The difference on `dev` was slightly higher (more like 1K words/s), but I'm not sure exactly why. Maybe caching, but this is not a super clean kind of timing test.

honnibal:
Okay, sounds good. Go ahead and merge when you're ready.

PR #5849, opened 2020-07-31
Allow Doc.char_span to snap to token boundaries

adrianeboyd:
## Description

Add a `mode` option to allow `Doc.char_span` to snap to token boundaries. The `mode` options:

* `strict`: character offsets must match token boundaries (default, same as before)
* `inside`: all tokens completely within the character span
* `outside`: all tokens at least partially covered by the character span

Add a new helper function `token_by_char` that returns the token corresponding to a character position in the text. Update `token_by_start` and `token_by_end` to use `token_by_char` for more efficient searching.

### Types of change

Enhancement.

## Checklist


honnibal:
This is a good idea, although I think we should think about better naming for the modes.

What about something like `strict`, `shrink` and `expand`?

honnibal:
`contract` is definitely a better antonym for `expand`, but I worried it might be read as the noun, and then someone might wonder "What's the difference between `strict` and `contract`?"

adrianeboyd:
`shrink` is just too weird, I wouldn't be able to guess what it means. `reduce`?

honnibal:
Maybe `strict`, `smaller`, `larger`? Or possibly `exact`, `expand`, `contract`?

adrianeboyd:
Does this get better if the setting is renamed to something like `snap_mode`? `snap_mode` = `exact`, `expand`, `contract`?

honnibal:
I think tweaking the keyword name a little is good. I'm okay with any of the following:

* `snap_mode`: `exact`, `expand`, `contract`
* `if_misaligned`: `raise`, `expand`, `contract`?
* `tokens`: `exact`, `expand`, `contract`?
* `alignment`: `exact`, `expand`, `contract`?
* `borders`: `exact`, `expand`, `contract`?

So, feel free to pick one of the above (or make another suggestion).

adrianeboyd:
Do you think `alignment_mode` is too long? `alignment_mode`: `exact`, `expand`, `contract`?

honnibal:
I think it's okay to have a longer name for this. Let's do `alignment_mode`.

itssimon:
@adrianeboyd is there a reason why this wasn't implemented for `Span.char_span()`? Would it make sense to keep these to functions consistent?

PR #5857, opened 2020-08-02
Allow adding pipeline components from source model

ines:
## Description

This PR adds support for adding pipeline components from other loadable models.

In code:
```python
source_nlp = spacy.load("en_core_web_sm")
nlp = spacy_blank("en")
nlp.add_pipe("parser", source=source_nlp)
```

In config:
```ini
[components.my_parser]
source = "en_core_web_sm"
component = "parser"
```

* `source`: loadable spaCy model
* `component`: optional name of component in the source pipeline, defaults to instance name in the target pipeline

This also solves the "training from a base model" problem: instead of providing a base model, the config can provide the components to source from the base model. It can also provide a list of `frozen_components` that are not updated. In this example, the final trained pipeline will contain 3 components that are updated/not updated in different ways:

```ini
[nlp]
lang = "en"
pipeline = ["tagger", "parser", "ner"]

[training]
frozen_components = ["ner"]

[components.parser]  # new component trained from scratch
factory = "parser"

[components.tagger]  # sourced component, updated with more examples
source = "en_core_web_sm"

[components.ner]  # frozen component, not updated
source = "en_core_web_sm"
```

### Notes

- `add_pipe` takes a loaded source object, which allows the user to load the model however they want and share the same loaded model across multiple calls
- In `Language.from_config` we cache `source` models of the same name so each source is only ever loaded once. We also disable loading the vocab, because we don't need that. (This means that we still load all pipeline components, but the overhead from this is hopefully minimal – the alternative would be implemending a mechanism for explicitly _enabling_ specific components only, so we can only load the `parser`, even if we don't know any of the other components in the pipeline.)
- The behaviour of `add_pipe` with a `source` set is slightly inconsistent: the first argument here doesn't specify the factory name but the component instance name in the source `nlp` object. It does feel pretty natural and logical, though, so I think it's okay? The name here corresponds to the `component` in the config, so you can copy a component called `some_parser` powered by the factory `parser` and add it to your pipeline as `my_parser`, without any conflicts.

### Todos

- [ ] What to do about basic compatibility checks, e.g. vectors? We can't know what needs/uses them and whether the incompatibility matters or not... Also, what else should we check? Language? That can be an issue, but doesn't have to be.
- [x] Improve error handling

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1497,7 +1497,7 @@ class Language:
                         # We only need the components here, no other data
                         disable = ["vocab", "tokenizer"]
                         source_nlps[model] = util.load_model(model, disable=disable)
-                    source_name = pipe_cfg.get("name", pipe_name)
+                    source_name = pipe_cfg.get("component", pipe_name)
                     nlp.add_pipe(source_name, source=source_nlps[model], name=pipe_name)
         nlp.config = filled if auto_fill else config
         nlp.resolved = resolved
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -621,19 +621,13 @@ class Language:
         return resolved[factory_name]
 
     def create_pipe_from_source(
-        self,
-        source_name: str,
-        source: "Language",
-        *,
-        name: str,
-        reset_weights: bool = False,
+        self, source_name: str, source: "Language", *, name: str,
     ) -> Tuple[Callable[[Doc], Doc], str]:
         """Create a pipeline component by copying it from an existing model.
 
         source_name (str): Name of the component in the source pipeline.
         source (Language): The source nlp object to copy from.
         name (str): Optional alternative name to use in current pipeline.
-        reset_weights (bool): Whether to reset the component's weights.
         RETURNS (Tuple[Callable, str]): The component and its factory name.
         """
         # TODO: handle errors and mismatches (vectors etc.)
@@ -648,8 +642,6 @@ class Language:
                 )
             )
         pipe = source.get_pipe(source_name)
-        if reset_weights and hasattr(pipe, "begin_training"):
-            pipe.begin_training()
         pipe_config = util.copy_config(source.config["components"][source_name])
         self._pipe_configs[name] = pipe_config
         return pipe, pipe_config["factory"]
@@ -705,9 +697,8 @@ class Language:
         if source is not None:
             # We're loading the component from a model. After loading the
             # component, we know its real factory name
-            reset_weights = config.get("reset_weights", False)
             pipe_component, factory_name = self.create_pipe_from_source(
-                factory_name, source, name=name, reset_weights=reset_weights
+                factory_name, source, name=name
             )
         else:
             if not self.has_factory(factory_name):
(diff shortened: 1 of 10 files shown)


honnibal:
I'm a bit worried about the `name` argument. If we're adding from a factory we might have:

```
nlp.add_pipe("ner", name="hotels_ner")
nlp.add_pipe("ner", name="foods_ner")
```

But if we're adding from source and we did:

```
nlp.add_pipe("ner", source="hotels_and_food", name="hotels_ner")
```

We would end up with a component named "ner". So the interpretation of the `name` argument is getting switched around. I think in the `add_pipe` case we should make the first positional argument the one that looks up in the source model, and `name` is an optional override of what the component will be named within the pipeline after we add it.

The problem then becomes how to call these arguments so that it makes sense with what we're writing in the config. Perhaps in the config we should write `component` rather than `name`?

honnibal:
Looks good! I started to type up something about the last point in your notes, but then I saw you'd already explained it. I think with that and the error handling, it's looking good.

honnibal:
Hmm it's pretty difficult to catch vector mismatches. It's very difficult to know whether the components we're sourcing rely on the vectors, it's not explicit in the components anymore. And this is a very frequent mistake people will make.

spacy/language.py
```diff
@@ -1429,25 +1473,37 @@ def from_config(
...
+                        config=pipe_cfg,
+                        overrides=pipe_overrides,
+                        validate=validate,
+                    )
+                else:
+                    model = pipe_cfg["source"]
+                    if model not in source_nlps:
+                        # We only need the components here, no other data
+                        disable = ["vocab", "tokenizer"]
+                        source_nlps[model] = util.load_model(model, disable=disable)
```
honnibal, line 1504: Did we remove the `vocab` keyword argument to `load_model`?

I think we'll have problems if `source_nlps[model].vocab` isn't the same instance as `self.vocab`. We've been trying to relax the requirement that the `Vocab` instance is the same across the pipeline, but we haven't really tested it, and I can imagine problems occurring.

spacy/language.py
```diff
@@ -1123,6 +1164,7 @@ def evaluate(
             arguments for specific components.
         scorer_cfg (dict): An optional dictionary with extra keyword arguments
             for the scorer.
+        exclude (Iterable[str]): Names of components that shouldn't be evaluated.
```
honnibal, line 1167: Hm, but if we're not running the component, we're not just not evaluating it. The rest of the pipeline also can't use its predictions, which could cause problems.

I think we should probably drop this option, and make people use `{disable, select}_pipes` for evaluate.

spacy/language.py
ines: I think this falls more into the category of validation? Because if there are underlying problems that make the vocabs incompatible, it's not going to be solved by initializing the source models with the destination vocab... But I don't have a good answer for how to do this beyond (maybe) warning if the language is different.

spacy/language.py
ines: Sure, that works, too. I just wanted to make it consistent. Because you might have a component that you want to train and not evaluate, for whatever reason. But maybe that's confusing.

spacy/language.py
honnibal: Hmm I do think it's an implementation issue rather than validation.

For instance, we might have a `morphologizer` component that expects to write to `self.vocab.morphology` during `begin_training`. It then expects to save out that state and load it back in `from_disk`. But if that `morphologizer` component has been sourced, its vocab instance won't be the one in the `nlp` object, so that information won't come back correctly.

PR #5861, opened 2020-08-03
Add Span index boundary checks

adrianeboyd:
## Description

Add Span index boundary checks

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/tokens/span.pyx
+++ b/spacy/tokens/span.pyx
@@ -181,9 +181,17 @@ cdef class Span:
             return Span(self.doc, start + self.start, end + self.start)
         else:
             if i < 0:
-                return self.doc[self.end + i]
+                token = self.doc[self.end + i]
+                if self.start <= token.i < self.end:
+                    return token
+                else:
+                    raise IndexError(Errors.E201)
             else:
-                return self.doc[self.start + i]
+                token = self.doc[self.start + i]
+                if self.start <= token.i < self.end:
+                    return token
+                else:
+                    raise IndexError(Errors.E201)
 
     def __iter__(self):
         """Iterate over `Token` objects.
--- a/spacy/tokens/span.pyx
+++ b/spacy/tokens/span.pyx
@@ -182,17 +182,12 @@ cdef class Span:
         else:
             if i < 0:
                 token_i = self.end + i
-                if self.start <= token_i < self.end:
-                    return self.doc[token_i]
-                else:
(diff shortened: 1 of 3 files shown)


spacy/tokens/span.pyx
```diff
@@ -181,9 +181,17 @@ cdef class Span:
             return Span(self.doc, start + self.start, end + self.start)
         else:
             if i < 0:
-                return self.doc[self.end + i]
+                token = self.doc[self.end + i]
+                if self.start <= token.i < self.end:
+                    return token
```
honnibal, line 186:
```suggestion
                    return self.doc[self.end + i]
```

spacy/tokens/span.pyx
```diff
@@ -181,9 +181,17 @@ cdef class Span:
             return Span(self.doc, start + self.start, end + self.start)
         else:
             if i < 0:
-                return self.doc[self.end + i]
+                token = self.doc[self.end + i]
```
honnibal, line 184:
```suggestion

```

spacy/tokens/span.pyx
```diff
@@ -181,9 +181,17 @@ cdef class Span:
...
+                if self.start <= token.i < self.end:
+                    return token
+                else:
+                    raise IndexError(Errors.E201)
             else:
-                return self.doc[self.start + i]
+                token = self.doc[self.start + i]
```
honnibal, line 190:
```suggestion
```

spacy/tokens/span.pyx
```diff
@@ -181,9 +181,17 @@ cdef class Span:
...
+                if self.start <= token.i < self.end:
+                    return token
```
honnibal, line 192:
```suggestion
                 return self.doc[self.start + i] 
```

spacy/tokens/span.pyx
honnibal: It looks a bit odd to get the token first and check second, a reasonable reader would wonder whether this was a bug as they wouldn't be assured the indexing will be legal.

adrianeboyd: Hmm, I'm not sure I understand what you mean with this suggested change. I was trying to return the same `Doc`-based `IndexError` as before (`E040`) for indices outside the doc and just address the `Span` issue, but if this is what I think you might have meant (?), I agree that it would be nicer to return the same `IndexError` for all cases here.

PR #5865, opened 2020-08-03
Create corpus iterator and batcher from registry during training

honnibal:
Refactor the `Corpus` class and training CLI so that the data reader and batcher are created from the registry. This allows the user to specify custom data loading, including streaming data from Python.

There's currently only one place where we consume the whole training stream, and that's in the `begin_training` function. It would be good to have an option to disable that if the user knows it's not necessary.

We currently don't support having a separate reader for the evaluation data. We should have a separate block, maybe `[evaluation]`, although it's only used during training which seemed confusing (that's why I haven't done it).

Another change here is that I've removed the assumption that the two CLI args are actually paths. If the user has a Python function, they might be arbitrary strings. So maybe we should just have a `[data]` block and remove the positional args? Then you'd have `--data.train.loc` and `--data.dev.loc`, which would be config overrides?

## Checklist


spacy/gold/corpus.py
```diff
@@ -54,6 +64,40 @@ def walk_corpus(path: Union[str, Path]) -> List[Path]:
...
         return locs
 
+    def __call__(self, nlp: "Language", loc: Union[Path, str],) -> Iterator[Example]:
+        """Yield examples from the data.
+
+        nlp (Language): The current nlp object.
+        loc (Path): The file or directory to read from.
+        YIELDS (Example): The examples.
+
+        DOCS: https://spacy.io/api/corpus#train_dataset
```
ines, line 74:
```suggestion
        DOCS: https://spacy.io/api/corpus#call
```

spacy/gold/corpus.py
```diff
@@ -54,6 +64,40 @@ def walk_corpus(path: Union[str, Path]) -> List[Path]:
...
+            examples = self.make_examples(nlp, ref_docs, self.max_length)
+        yield from examples
+
+    def count(self, nlp: "Language", loc) -> int:
+        """Returns count of words in train examples.
+
+        nlp (Language): The current nlp. object.
+        RETURNS (int): The word count.
+
+        DOCS: https://spacy.io/api/corpus#count_train
```
ines, line 90:
```suggestion
        DOCS: https://spacy.io/api/corpus#count
```

spacy/gold/batchers.py
```diff
@@ -0,0 +1,168 @@
...
+from .example import Example
+from ..util import registry, minibatch
+
+
+Sizing = Union[Iterable[int], int]
+ItemT = TypeVar("ItemT")
+BatcherT = Callable[[Iterable[ItemT]], Iterable[List[ItemT]]]
+
+
+@registry.batchers("spacy.batch_by_padded.v1")
```
ines, line 15: This is all very nice! 💯

spacy/gold/batchers.py
```diff
@@ -0,0 +1,168 @@
...
+from typing import Optional, Any
+from functools import partial
+import itertools
+
+from .example import Example
+from ..util import registry, minibatch
+
+
+Sizing = Union[Iterable[int], int]
+ItemT = TypeVar("ItemT")
```
ines, line 11: So are we officially using `TypeVar`s in spaCy now? 😛

honnibal: It just felt right here!

spacy/gold/corpus.py
honnibal: We could also consider removing this method, as we'd rather not ask user-provided readers to supply it.

spacy/schemas.py
```diff
@@ -178,39 +183,32 @@ class ModelMetaSchema(BaseModel):
...
+    raw_text: Optional[StrictStr] = Field(default=None, title="Raw text")
+    eval_batch_size: Union[Sequence[int], int] = Field(..., title="The batch size during evaluation")
     optimizer: Optimizer = Field(..., title="The optimizer to use")
     # fmt: on
 
     class Config:
         extra = "forbid"
         arbitrary_types_allowed = True
 
+#eval_batch_size: StrictInt = Field(..., title="Evaluation batch size")
```
ines, line 211: Should this be removed?

```suggestion
```

spacy/gold/corpus.py
ines: Ah, I thought we needed this for the word counts. But it looks like we're currently only using it in one test?

I'm definitely in favour of removing stuff we don't need and keeping the API simple.

PR #5866, opened 2020-08-03
Add config callbacks for modifying nlp object before and after init

ines:
### Todos

- [x] raise if callbacks don't return nlp object

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/default_config.cfg
+++ b/spacy/default_config.cfg
@@ -12,8 +12,9 @@ use_pytorch_for_gpu_memory = false
 lang = null
 pipeline = []
 load_vocab_data = true
-before_init = null
-after_init = null
+before_creation = null
+after_creation = null
+after_pipeline_creation = null
 
 [nlp.tokenizer]
 @tokenizers = "spacy.Tokenizer.v1"
--- a/spacy/default_config.cfg
+++ b/spacy/default_config.cfg
@@ -2,6 +2,8 @@
 lang = null
 pipeline = []
 load_vocab_data = true
+before_init = null
+after_init = null
 
 [nlp.tokenizer]
 @tokenizers = "spacy.Tokenizer.v1"
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1464,11 +1464,12 @@ class Language:
         config["components"] = orig_pipeline
         create_tokenizer = resolved["nlp"]["tokenizer"]
         create_lemmatizer = resolved["nlp"]["lemmatizer"]
-        before_init = resolved["nlp"]["before_init"]
-        after_init = resolved["nlp"]["after_init"]
+        before_creation = resolved["nlp"]["before_creation"]
+        after_creation = resolved["nlp"]["after_creation"]
+        after_pipeline_creation = resolved["nlp"]["after_pipeline_creation"]
         lang_cls = cls
-        if before_init is not None:
-            lang_cls = before_init(cls)
+        if before_creation is not None:
+            lang_cls = before_creation(cls)
             if (
                 not isinstance(lang_cls, type)
                 or not issubclass(lang_cls, cls)
@@ -1480,6 +1481,10 @@ class Language:
             create_tokenizer=create_tokenizer,
             create_lemmatizer=create_lemmatizer,
         )
+        if after_creation is not None:
+            nlp = after_creation(nlp)
+            if not isinstance(nlp, cls):
+                raise ValueError(Errors.E942.format(name="creation", value=type(nlp)))
         # Note that we don't load vectors here, instead they get loaded explicitly
         # inside stuff like the spacy train function. If we loaded them here,
         # then we would load them twice at runtime: once when we make from config,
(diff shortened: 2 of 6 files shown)


spacy/default_config.cfg
```diff
@@ -2,6 +2,8 @@
 lang = null
 pipeline = []
 load_vocab_data = true
+before_init = null
```
honnibal, line 5: The big question is whether we also want to run the hooks at runtime, or is this a training-only step?

I think they do also need to run at runtime, right?

ines: I think they should also run a runtime, otherwise it would be a bit confusing and means people need separate workflows. If you're customising the tokenizer with a registered function, you should also be able to provide your before/after callbacks like this.

(Also, I still have this vision of somehow providing a workflow for users to automatically include their Python code provided via `--code` into their packaged model. That would solve the whole "train with custom code and ship your model" question...)

honnibal: I think that should be feasible, but as an argument of `spacy package` not `spacy train`, right? In package we can either just dump it into the `__init__`, or import it within the init.

ines: Yeah, exactly, that was the idea. `spacy train` doesn't package, so including the code anywhere wouldn't make sense here. But you can pass it in when you package and we can import it in the `__init__.py` (if it's all one module and self-contained, which makes sense here anyways).

spacy/schemas.py
```diff
@@ -218,6 +218,8 @@ class ConfigSchemaNlp(BaseModel):
     tokenizer: Callable = Field(..., title="The tokenizer to use")
     lemmatizer: Callable = Field(..., title="The lemmatizer to use")
     load_vocab_data: StrictBool = Field(..., title="Whether to load additional vocab data from spacy-lookups-data")
+    before_init: Optional[Callable] = Field(..., title="Optional callback to modify Language class before initialization")
+    after_init: Optional[Callable] = Field(..., title="Optional callback to modify nlp object after initialization")
```
honnibal, line 222: How would you feel about allowing these to default to `None`? The main disadvantage I can see is that we don't have much else that defaults to `None` here, so it'll be harder to document consistently.


```suggestion
    before_init: Optional[Callable] = Field(None, title="Optional callback to modify Language class before initialization")
    after_init: Optional[Callable] = Field(None, title="Optional callback to modify nlp object after initialization")
```

spacy/language.py
```diff
@@ -1508,6 +1512,8 @@ def from_config(
                     nlp.add_pipe(source_name, source=source_nlps[model], name=pipe_name)
         nlp.config = filled if auto_fill else config
         nlp.resolved = resolved
+        if after_init is not None:
+            after_init(nlp)
```
honnibal, line 1516: Will things break if the callback wants to return a different object? I don't think we rely on object identity for the `nlp` object anywhere outside here, so I think this will work?

```suggestion
            nlp = after_init(nlp)
```

If it works I would prefer this, as it may allow some situations that would otherwise be difficult to implement.

spacy/language.py
```diff
@@ -1463,6 +1463,10 @@ def from_config(
         config["components"] = orig_pipeline
         create_tokenizer = resolved["nlp"]["tokenizer"]
         create_lemmatizer = resolved["nlp"]["lemmatizer"]
+        before_init = resolved["nlp"]["before_init"]
+        after_init = resolved["nlp"]["after_init"]
+        if before_init is not None:
+            before_init(cls)
```
honnibal, line 1469:
```suggestion
            nlp = before_init(cls)
```

(See below)

spacy/tests/test_language.py
```diff
@@ -153,6 +154,40 @@ def test_language_pipe_stream(nlp2, n_process, texts):
...
+    ran_before = False
+    ran_after = False
+
+    @registry.callbacks(f"{name}_before")
+    def make_before_init():
+        def before_init(lang_cls):
+            nonlocal ran_before
+            ran_before = True
+            assert lang_cls is English
+            lang_cls.Defaults.foo = "bar"
```
honnibal, line 168:
```suggestion
            lang_cls.Defaults.foo = "bar"
            return lang_cls
```

honnibal:
Perhaps one more callback, `before_components`?

This would be called after the `nlp` object is created and the config is loaded in, but before the components are added. This would be the last chance to hijack the config.

spacy/schemas.py
ines: Yeah, this would be how we'd do it if we decide to make some config fields optional. The values here currently already default to `None`, but if we set it to `None` here, it will be auto-filled and the config validation will never complain about it not being set... currently that's inconsistent but if we want more settings to behave like this, we should definitely consider it for `before_init` and `after_init` as well.

spacy/language.py
ines: Ah, good point. I think I might have to adjust the type signature for the callback as well then.

spacy/tests/test_language.py
```diff
@@ -153,6 +154,40 @@ def test_language_pipe_stream(nlp2, n_process, texts):
...
+        return before_init
+
+    @registry.callbacks(f"{name}_after")
+    def make_after_init():
+        def after_init(nlp):
+            nonlocal ran_after
+            ran_after = True
+            assert isinstance(nlp, English)
+            assert nlp.Defaults.foo == "bar"
+            nlp.meta["foo"] = "bar"
```
ines, line 179:
```suggestion
            nlp.meta["foo"] = "bar"
            return nlp
```

spacy/schemas.py
honnibal: Okay cool

PR #5868, opened 2020-08-03
Recalculate alignment if tokenization differs

adrianeboyd:
## Description

Recalculate alignment if tokenization differs from tokenization at the point when the saved alignment was created.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/gold/example.pyx
+++ b/spacy/gold/example.pyx
@@ -1,6 +1,7 @@
 import warnings
 
 import numpy
+from typing import Optional, List
 
 from ..tokens.doc cimport Doc
 from ..tokens.span cimport Span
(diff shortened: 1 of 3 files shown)


adrianeboyd:
(No attachment to any of the naming here, it is not great.)

spacy/gold/example.pyx
```diff
@@ -26,6 +27,17 @@ cpdef Doc annotations2doc(vocab, tok_annot, doc_annot):
     return output
 
 
+class SavedAlignment:
+    def __init__(self,
+        alignment: Optional[Alignment] = None,
+        predicted_tokens: Optional[List[str]] = None,
+        reference_tokens: Optional[List[str]] = None,
+    ):
```
svlandeg, line 35: I personally wouldn't have the default values in this constructor, and only create this object if all 3 parameters are available (and have the `self._saved_alignment` field be `None` otherwise)

PR #5872, opened 2020-08-04
Default empty KB in EL component

svlandeg:
## Description
The new registry asset `spacy.EmptyKB.v1` creates an empty KB with a configurable entity length. This is used with a default length in case the user doesn't specify anything in their config for the `entity_linking` component. The KB constructor does not take a `Vocab` argument anymore - this can be set (`kb.initialize`) when the EL pipe is created.

This now generates a new EL component with custom length - without going through the hassle of calling the KnowledgeBase constructor etc:
```
nlp.add_pipe("entity_linker", config={"kb": {"entity_vector_length": 35}})
```

If OK, I'll document this separately on the nightly docs branch.

### Types of change
enhancement

## Checklist


honnibal:
Looks fine to me, go ahead and merge when you're ready.

PR #5873, opened 2020-08-04
Add warning when Matcher subpattern is discarded

### LINKED ISSUE #5870: Unexpected behavior when using EntityRuler patterns with a misnamed/miscased/empty object
I was able to recreate the "bug" described here: https://support.prodi.gy/t/regex-operator-in-the-patterns-file/1008/9 using the spaCy code below.

Basically, when a developer mistakenly uses the lowercase "regex" as a key (instead of the correct "REGEX" key), the patterns matcher will match all tokens in a document. Obviously, the main answer here is -- "don't do that". But I guess I'd expect spaCy/Prodigy to throw an error (or flag none of the tokens, rather than all of them).

```
import spacy

nlp = spacy.load("en_core_web_sm", disable=['ner'])
ruler = EntityRuler(nlp)
ruler.add_patterns([{"label":"AC","pattern":[{"text":{"regex":"^abc$"}}]}]) # subtle BUG: "regex" should be "REGEX" (uppercase)
nlp.add_pipe(ruler)

doc = nlp("My favorite word is abc.") # only "abc" should match

for ent in doc.ents:
    print(ent, ent.label_) # Note that *every* token matches
```

Output:
```
My AC
favorite AC
word AC
is AC
abc AC
. AC
```

Note that the offending line behaves the same way in the following circumstances:
* `ruler.add_patterns([{"label":"AC","pattern":[{"text":{"regex":"^abc$"}}]}])`
* `ruler.add_patterns([{"label":"AC","pattern":[{"text":{"random_nonsense":"^abc$"}}]}])`
* `ruler.add_patterns([{"label":"AC","pattern":[{"text":{}}]}])`

## Your Environment

* Operating System: windows 10
* Python Version Used: Python 3.7.8
* spaCy Version Used: 2.3.0
* Environment Information:

adrianeboyd:
## Description

* Add a warning when a subpattern is not processed and discarded
* Normalize subpattern attribute/operator keys to upper case like top-level attributes

Fixes #5870.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -133,6 +133,8 @@ class Warnings(object):
             "normalization table, please ignore this warning.")
     W034 = ("Please install the package spacy-lookups-data in order to include "
             "the default lexeme normalization table for the language '{lang}'.")
+    W035 = ('Discarding subpattern "{pattern}" due to an unrecognized '
+            "attribute or operator.")
 
 
 @add_codes
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -811,9 +811,11 @@ def _get_extra_predicates(spec, extra_predicates):
                 attr = "ORTH"
             attr = IDS.get(attr.upper())
         if isinstance(value, dict):
+            processed = False
+            value_with_upper_keys = {k.upper(): v for k, v in value.items()}
             for type_, cls in predicate_types.items():
-                if type_ in value:
-                    predicate = cls(len(extra_predicates), attr, value[type_], type_)
+                if type_ in value_with_upper_keys:
+                    predicate = cls(len(extra_predicates), attr, value_with_upper_keys[type_], type_)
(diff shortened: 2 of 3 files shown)


adrianeboyd:
I'm not sure we actually want to do this, though.

adrianeboyd:
Also I'm saying "we don't want to change the behavior" and then I'm changing the behavior by accepting `"regex"`, so maybe that part is a bad idea. It's a bit weird to accept `text` and not `regex`, though.

honnibal:
Accepting `regex` is definitely good, if you'd have asked me before I would've guessed `regex` would work.

Go ahead and merge this if you're happy with it? I do worry about all the warnings we're adding, I'd like to move many/most of them to logging. But we should to that for a bunch of them at once, so for consistency a warning is currently correct.

adrianeboyd:
I also thought the strict validator would accept `text` and it doesn't.

PR #5875, opened 2020-08-04
Bugfix in nlp.replace_pipe

svlandeg:
## Description
Current bug: if a pipeline would contain more than one component (let's say 3), and you `nlp.replace_pipe` the last one, you would get an `E959` error about "Can't insert component before index 2", because that last (old) component would already be removed so the pipeline is temporarily smaller. Fixed it by using  insertion with `after` for these edge cases.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -771,8 +771,10 @@ class Language:
         self.remove_pipe(name)
         if not len(self.pipeline):  # we have no components to insert before/after
             self.add_pipe(factory_name, name=name)
-        else:
+        elif pipe_index < len(self.pipeline):
             self.add_pipe(factory_name, name=name, before=pipe_index)
+        else:
+            self.add_pipe(factory_name, name=name, after=pipe_index-1)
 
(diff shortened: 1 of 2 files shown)


spacy/language.py
```diff
@@ -771,8 +771,10 @@ def replace_pipe(
         self.remove_pipe(name)
         if not len(self.pipeline):  # we have no components to insert before/after
             self.add_pipe(factory_name, name=name)
```
ines, line 773: Since `after=pipe_index - 1` is equivalent to `last=True` (default behaviour), maybe we can combine those two conditions above? Then we can save that extra conditional.

svlandeg:
@ines : I hope it's OK I went ahead and merged after addressing your comments, because I want to get another PR out that needs this fix :-)

ines:
@svlandeg Yeah, no problem, it's a very straightforward fix so feel free to merge stuff like that yourself 😃

PR #5876, opened 2020-08-05
Define candidate generator in EL config

svlandeg:
## Description
With v.3 coming nicely together, there's now a straightforward way to allow swapping in a different candidate generator for the Entity Linker. This PR builds on the ideas of @kabirkhan in PR #4988 but uses the new config system in v.3 (currently on the `develop` branch). 

Instead of overwriting as in the original PR, we now specifically allow defining a candidate generator of type `Callable[[KnowledgeBase, "Span"], Iterable[Candidate]]`. If none such is provided, the default `spacy.CandidateGenerator.v1` is used, which simply does a lookup in the KB's index:
```
return kb.get_alias_candidates(span.text)
```

If, for instance, we'd want to search the index with the lowercase text, we could swap out the candidate generator like this:
```
    def get_lowercased_candidates(kb, span):
        return kb.get_alias_candidates(span.text.lower())

    @registry.assets.register("spacy.LowercaseCandidateGenerator.v1")
    def create_candidates() -> Callable[[KnowledgeBase, "Span"], Iterable[Candidate]]:
        return get_lowercased_candidates

    # replace the pipe with a new one with with a different candidate generator
    nlp.replace_pipe(
        "entity_linker",
        "entity_linker",
        config={
            "kb": {"@assets": "myKB.v1"},
            "get_candidates": {"@assets": "spacy.LowercaseCandidateGenerator.v1"},
        },
    )
```

@kabirkhan: I'd be happy if you could review, as I hope that this will enable a more straightforward integration of your ANN linker!

### Types of change
enhancement (+ small bug fix in `nlp.replace_pipe` which didn't pass on some arguments)

## Checklist
- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information. (I'll do this separately)


spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -160,15 +170,71 @@ def test_candidate_generation(nlp):
...
+
+def test_el_pipe_configuration():
+    """Test correct candidate generation as part of the EL pipe"""
+    nlp = English()
+    nlp.add_pipe("sentencizer")
+    pattern = {"label": "PERSON", "pattern": [{"LOWER": "douglas"}]}
+    ruler = nlp.add_pipe("entity_ruler")
+    ruler.add_patterns([pattern])
+
+    @registry.assets.register("myAdamKB.v1")
```
honnibal, line 193: Thanks for putting this inside the test. It's something we have to watch for, as the tests may get imported at runtime (one of the disadvantages of shipping the tests inside the library).

spacy/pipeline/entity_linker.py
```diff
@@ -117,10 +121,12 @@ def __init__(
         }
         if not isinstance(kb, KnowledgeBase):
             raise ValueError(Errors.E990.format(type=type(self.kb)))
-        kb.initialize(vocab)
+        if kb.vocab is None:
```
honnibal, line 124: Is it normal that the KB's vocab is `None`? And will we have problems if `self.vocab is not kb.vocab`?

One way we can have problems from two vocab instances is that only one will be serialized, so if the component had relied on setting things in the vocab and adding strings, that state would be lost.

svlandeg: This was recently changed in PR #5872. Basically to allow for a default empty KB (`"spacy.EmptyKB.v1"`), I had to take out the `Vocab` field in the constructor. It feels like a circle: the EL needs the KB, but the KB needs the Vocab of the EL. So that's why I thought the EL constructor should be setting it. But then what should happen if you already have a predefined KB with entries that required a Vocab, before you even made the EL pipe?

svlandeg: But you're right that deserialization relies on the KB being initialized with self.vocab - so that it plainly wrong and requires fixing.

honnibal: Hm, well the EL factory gets the `nlp` object, so it should be able to use `nlp.vocab`?

Sorry for not seeing this in #5872, sometimes it just takes a few iterations to see all the consequences unfold.

svlandeg: Don't I know it ;-)

svlandeg: What do you think of the current solution, @honnibal? Basically instead of taking a `KnowledgeBase` as argument in the `EntityLinker` constructor, we take a function that creates a KB from a `Vocab`. That ensures that the EL pipe is always in charge of the creating & serializing the vocab.

spacy/pipeline/entity_linker.py
```diff
@@ -442,8 +435,7 @@ def load_model(p):
                 raise ValueError(Errors.E149)
 
         def load_kb(p):
```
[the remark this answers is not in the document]
svlandeg: Great catch, thanks! In fact, this `load_kb` bit needs to assume that `self.kb` has already been generated when the `nlp` pipeline recreated itself from its config. So yes, it definitely shouldn't call that constructor directly. 
See latest commit - I also added a unit test for using a `KnowledgeBase` subclass in the EL pipe.

svlandeg:
I think we probably want to rename the `KnowledgeBase` methods `dump` and `load_bulk`, that both take a file? These will be the methods that you'll need to reimplement if you want to use a custom KB...

Also we probably want an additional abstraction - some sort of KB interface (like `Pipe`) so that a subclass doesn't necessarily have to care about all the Cython stuff we have there for efficiency. I guess if someone wants a completely different implementation of a KB, that should work too, as long as it adheres to a simple API for access & IO.

Thoughts?

kabirkhan:
I agree this is probably a separate PR but yeah I think renaming the `KnowledgeBase` methods `dump` => `to_disk` and `load_bulk` => `from_disk` would make the most sense for consistency. 

I think the KB interface idea is maybe a nice to have but not required. I didn't have too much trouble with subclassing the KnowledgeBase in pure Python but if you do end up going this route I'd be happy to review and test on my side as well.

spacy/pipeline/entity_linker.py
honnibal: I think passing a `KnowledgeBase` instance into the `__init__` is good, but if you want the factory could take a callable that you provide the vocab to, that does make sense.

However...The complexity/intricacy possible starts to add up?

One possibility would be to move the `KnowledgeBase` to be an attribute of the `Vocab`. That would be a bit more consistent with stuff like the `Morphology` class, and it answers the question of what you'd do if you want multiple entity linkers to access the same KB. It does mean you could only have one KB per pipeline though. With this approach, you would want some sort of configuration on the NLP object to set up the knowledge-base.

honnibal:
Definitely yes to the renaming, but let's try to avoid having the abstract base classes as much as possible?

I try to avoid doing anything with inheritance within the library if I can, and try to leave inheritance as a "mechanism of last resort" that users can use to customise things. It's always difficult for the user subclasses to fit in alongside the existing class hierarchy, and it becomes really brittle as things change (because you're always moving methods up and down the hierarchy...).

We have to use the inheritance for `Pipe` and `Language`, but we don't want to have our own inheritance hierarchy for `Vocab`, `Morphology`, `Vectors`, `KnowledgeBase` if we can avoid it. Having composition (like this candidate generation PR) is way better.

spacy/pipeline/entity_linker.py
svlandeg: Hm, I'm really not sure about that last option :|

svlandeg:
Ok, as the consensus was to do only the renaming, I added that to this PR, updated some code that conflicted from a previously merged PR, and this should be good to go I think?

ines:
(Oh damn, looking at this PR made me realise that I just removed the API entries for `kb_loader` and `get_candidates` because I matched them up with the current implementation and thought they were referring to _outdated_ API rather than future API. Sorry about that! We can add it back once the PR is merged, I think it should only be 2 rows in the tables on the API page.)

svlandeg:
Hehe sorry Ines, I wanted to write the docs for this before I left on holiday, and it ended up getting merged before this PR :p Don't worry about it, I'll make sure to double-check the EL documentation once the API is stable!

PR #5879, opened 2020-08-05
Be less choosy about reporting textcat scores

honnibal:
I'm not 100% sure this is correct, but I think the micro-F scoring should be applicable regardless of whether there are multiple classes and regardless of whether the labels are mutually exclusive. I think the same is true for some of the other measures?

If so, we should just dump all those scores into the scores dict.

### DIFF
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -318,7 +318,7 @@ class Scorer:
         labels: Iterable[str] = tuple(),
         multi_label: bool = True,
         positive_label: Optional[str] = None,
-        threshold=Optional[float] = None,
+        threshold: Optional[float] = None,
         **cfg,
     ) -> Dict[str, Any]:
         """Returns PRF and ROC AUC scores for a doc-level attribute with a
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -318,6 +318,7 @@ class Scorer:
         labels: Iterable[str] = tuple(),
         multi_label: bool = True,
         positive_label: Optional[str] = None,
+        threshold=Optional[float] = None,
         **cfg,
     ) -> Dict[str, Any]:
         """Returns PRF and ROC AUC scores for a doc-level attribute with a
@@ -333,6 +334,9 @@ class Scorer:
             Defaults to True.
         positive_label (str): The positive label for a binary task with
             exclusive classes. Defaults to None.
+        threshold (float): Cutoff to consider a prediction "positive". Defaults
+            to 0.5 for multi-label, and 0.0 (i.e. whatever's highest scoring)
+            otherwise.
         RETURNS (Dict[str, Any]): A dictionary containing the scores, with
             inapplicable scores as None:
             for all:
@@ -346,6 +350,8 @@ class Scorer:
 
         DOCS: https://spacy.io/api/scorer#score_cats
         """
+        if threshold is None:
+            threshold = 0.5 if multi_label else 0.0
         f_per_type = {label: PRFScore() for label in labels}
         auc_per_type = {label: ROCAUCScore() for label in labels}
         labels = set(labels)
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I think this is okay except for plain P/R/F, which only make sense if there is a `positive_label`, so that should be a condition on adding it. It's only calculated if there is a `positive_label`, so it'd be `0` otherwise anyway, but I would avoid the clutter/confusion:

https://github.com/explosion/spaCy/blob/586d695775e478fa53296f6332d5feb069412731/spacy/scorer.py#L367-L371

honnibal:
I expected the plain p/r/f to just be the micro-averaged F score?

adrianeboyd:
I always have to work through a bunch of toy examples to remember all the details. The current plain PRF is definitely not a micro-anything. Let's see:

* there's no need for a separate PRF score for the positive label, it can just be extracted from `f_per_type`
* you could add overall accuracy for all configurations
* you could add micro PRF (would be the same as accuracy for multiclass, would possibly be hard to interpret for multilabel)

honnibal:
Yeah, I agree that it's confusing.

To me, it's actually "accuracy" that's unclear, rather than micro F. Micro F makes a lot of sense in the multi-label case: you can vary the number of predictions, so if you make only one correct prediction you'll have precision of 1 and terrible recall. But I don't really know what "accuracy" means in that situation, I guess it maps to precision?

I'll try refactoring this a bit, one sec.

honnibal:
Okay, I think the micro and macro F scores are right. I'm not sure about the AUC though, I've never really used that measure.

adrianeboyd:
Ah, right, the previous `f_per_type` was only set up for the multiclass option. Things I would change:

* I think that `0.5` should be a configurable parameter like `multi_label_threshold` with a default of 0.5 instead.

* I would call the micro scores `{attr}_micro_p/r/f/` to make it clearer. I think I would be confused by `macro_f` vs. `f`, it would make me unsure about what the contrast was.

spacy/scorer.py
```diff
@@ -338,89 +338,91 @@ def score_cats(
...
+            f"{attr}_macro_f": macro_f,
             f"{attr}_macro_auc": None,
             f"{attr}_f_per_type": {k: v.to_dict() for k, v in f_per_type.items()},
             f"{attr}_auc_per_type": {k: v.score for k, v in auc_per_type.items()},
         }
         if len(labels) == 2 and not multi_label and positive_label:
-            results[f"{attr}_p"] = score.precision
-            results[f"{attr}_r"] = score.recall
-            results[f"{attr}_f"] = score.fscore
             results[f"{attr}_score"] = results[f"{attr}_f"]
```
adrianeboyd, line 420: This should be:

```
results[f"{attr}_score"] = results[f"{attr}_f_per_type"][positive_label]['f']
```

honnibal:
Okay done. I was thinking of adding the threshold as well, and actually it can apply to the non-multi-label case too, so we don't need such an awkward parameter name.

spacy/scorer.py
```diff
@@ -333,94 +334,104 @@ def score_cats(
...
             exclusive classes. Defaults to None.
+        threshold (float): Cutoff to consider a prediction "positive". Defaults
+            to 0.5 for multi-label, and 0.0 (i.e. whatever's highest scoring)
+            otherwise.
         RETURNS (Dict[str, Any]): A dictionary containing the scores, with
             inapplicable scores as None:
             for all:
                 attr_score (one of attr_f / attr_macro_f / attr_macro_auc),
                 attr_score_desc (text description of the overall score),
+                attr_f,
```
adrianeboyd, line 345: `micro_f`

spacy/scorer.py
```diff
@@ -333,94 +334,104 @@ def score_cats(
             Defaults to True.
         positive_label (str): The positive label for a binary task with
             exclusive classes. Defaults to None.
+        threshold (float): Cutoff to consider a prediction "positive". Defaults
+            to 0.5 for multi-label, and 0.0 (i.e. whatever's highest scoring)
+            otherwise.
         RETURNS (Dict[str, Any]): A dictionary containing the scores, with
             inapplicable scores as None:
             for all:
                 attr_score (one of attr_f / attr_macro_f / attr_macro_auc),
```
adrianeboyd, line 343: `micro_f` here, too

PR #5882, opened 2020-08-05
Use "raise ... from" in custom errors for better tracebacks

ines:
## Description

Some background on `raise` with `from`: https://stackoverflow.com/a/24752607/6400719

This is mostly relevant for cases where we re-raise exceptions during exception handling. I think our general guidelines should be:

* `raise ... from None`: the exception we're handling isn't relevant to the custom error. Examples of this include: internal `pydantic` validation error we're using to provide custom config validation messages, third-party import error we're catching to provide a more descriptive message etc.
* `raise ... from original_exception`: related third-party exception that means whatever we're doing has to fail as well. The traceback will then read "The above exception was the direct cause of the following exception:".

(Not directly related, but I noticed this while going through all instances of `try`/`except`: there are a few cases where we can replace the `try`/`except` block with a simple conditional, which is always preferred over catching stuff like `IndexError`s or – worse, because it could mean anything – `ValueError`s. So we should fix those as well and replace them with conditional that checks that actually test for what _would_ cause the error.)

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -211,7 +211,7 @@ def create_evaluation_callback(
         except KeyError as e:
             keys = list(scores.keys())
             err = Errors.E983.format(dict="score_weights", key=str(e), keys=keys)
-            raise KeyError(err)
+            raise KeyError(err) from None
         return weighted_score, scores
 
     return evaluate
@@ -369,7 +369,7 @@ def setup_printer(
                 Errors.E983.format(
                     dict="scores (losses)", key=str(e), keys=list(info["losses"].keys())
                 )
-            )
+            ) from None
 
         try:
             scores = [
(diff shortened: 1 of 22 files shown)


honnibal:
> (Not directly related, but I noticed this while going through all instances of try/except: there are a few cases where we can replace the try/except block with a simple conditional, which is always preferred over catching stuff like IndexErrors or – worse, because it could mean anything – ValueErrors. So we should fix those as well and replace them with conditional that checks that actually test for what would cause the error.)

Yeah this is something I'd like us to stay vigilant on. It's quite common in Python code to use try/except in ways I really don't like, for instance I think it's important when you're using a try/except block to keep the code under the try as little as possible, usually exactly we want exactly one statement.

PR #5890, opened 2020-08-06
EL docs update

svlandeg:
## Description
- `DocBin` format
- update of EL docs
- architectures of EL & textcat
- various small fixes

### Types of change
docs

## Checklist

### DIFF
--- a/website/docs/api/entitylinker.md
+++ b/website/docs/api/entitylinker.md
@@ -9,6 +9,12 @@ api_string_name: entity_linker
 api_trainable: true
 ---
 
+An Entity Linker component disambiguates textual mentions (tagged as named
+entities) to unique identifiers, grounding the named entities into the "real
+world". It requires a Knowledge base, a function to generate plausible
+candidates from that Knowledge base given a certain textual mention, and a ML
+model to pick the right candidate, given the local context of the mention.
+
 ## Config and implementation {#config}
 
 The default config is defined by the pipeline component factory and describes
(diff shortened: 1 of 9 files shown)


website/docs/api/entitylinker.md
```diff
@@ -9,6 +9,12 @@ api_string_name: entity_linker
 api_trainable: true
 ---
 
+An Entity Linker component disambiguates textual mentions (tagged as named
```
ines, line 12: We should probably decide on either "The entity linker" or "The `EntityLinker`"

website/docs/api/entitylinker.md
```diff
@@ -9,6 +9,12 @@ api_string_name: entity_linker
...
+entities) to unique identifiers, grounding the named entities into the "real
+world". It requires a Knowledge base, a function to generate plausible
```
ines, line 14: I think same here – either knowledge base or `KnowledgeBase`

website/docs/api/entitylinker.md
svlandeg: You mean that those two variants are OK, but not the mixture "Entity Linker", right? (I agree, just want to double check you don't want to reduce it down to only one of the two options across the docs)

ines: Yeah, exactly, sorry if I phrased this in a confusing way!

svlandeg: Ok, I think the edit I made is fine then?

PR #5904, opened 2020-08-10
Fix typos in comments

graue70:


honnibal:
Thanks! I think I need to be more vigilant on US vs UK English, too...Officially we're supposed to have standardised on US English, but it creeps in in comments etc.

PR #5906, opened 2020-08-10
Tidy up pipes

ines:
## Description

- [x] add dummy `Pipe.update` method
- [x] add default `Pipe.__init__` method
- [x] replace unnecessary `try`/`except` with instance checks etc.
- [x] tidy up

### Types of change
enhancement

## Checklist


spacy/cli/train.py
```diff
@@ -295,7 +295,11 @@ def train_while_improving(
                 nlp.rehearse(raw_batch, sgd=optimizer, losses=losses, exclude=exclude)
         # TODO: refactor this so we don't have to run it separately in here
         for name, proc in nlp.pipeline:
-            if name not in exclude and hasattr(proc, "model"):
+            if (
+                name not in exclude
+                and hasattr(proc, "model")
+                and isinstance(proc.model, Model)
```
honnibal, line 301: We shouldn't do this `isinstance` check, if someone has a value set to the `model` attribute of their component that doesn't duck-type correctly, that's their bug.

We could check something like `not in (False, True, None)`. Or we could ask pipes to implement a `is_trainable` property.

spacy/language.py
```diff
@@ -958,7 +958,11 @@ def update(
             proc.update(examples, sgd=None, losses=losses, **component_cfg[name])
         if sgd not in (None, False):
             for name, proc in self.pipeline:
-                if name not in exclude and hasattr(proc, "model"):
+                if (
+                    name not in exclude
+                    and hasattr(proc, "model")
+                    and isinstance(proc.model, Model)
```
honnibal, line 964:
```suggestion
                and proc.model not in (True, False, None)
```

spacy/language.py
```diff
@@ -1649,7 +1653,7 @@ def _fix_pretrained_vectors_name(nlp: Language) -> None:
     else:
         raise ValueError(Errors.E092)
     for name, proc in nlp.pipeline:
-        if not hasattr(proc, "cfg"):
+        if not hasattr(proc, "cfg") or not isinstance(proc.cfg, dict):
```
honnibal, line 1656: This `isinstance` check is harder to replace, but I definitely think we should.

honnibal: Wait do we still need this function?

spacy/pipeline/pipe.pyx
```diff
@@ -99,7 +101,7 @@ cdef class Pipe:
         """
         if losses is None:
             losses = {}
-        if not hasattr(self, "model"):
+        if not hasattr(self, "model") or not isinstance(self.model, Model):
```
honnibal, line 104:
```suggestion
        if not hasattr(self, "model") or self.model in (None, True, False):
```

spacy/language.py
ines: I'm not sure? 

In general, we definitely need to make sure `Pipe.cfg` is a dict, though, because we make assumptions about it elsewhere.

spacy/gold/example.pyx
```diff
@@ -26,6 +26,40 @@ cpdef Doc annotations2doc(vocab, tok_annot, doc_annot):
...
+    if not isinstance(examples, IterableInstance):
+        err = Errors.E978.format(name=method, types=type(examples))
+        raise TypeError(err)
+    wrong = set([type(eg) for eg in examples if not isinstance(eg, Example)])
+    if wrong:
+        err = Errors.E978.format(name=method, types=wrong)
+        raise TypeError(err)
+
+
+def iter_get_examples(get_examples, name, method="begin_training"):
```
honnibal, line 45: Hmm I don't really love this function. I'd rather have the `begin_training` methods doing something like:

```
check_get_examples_is_callable(get_examples, name)
for example in get_examples():
    check_is_example_object(example, name)
```

ines: We could call it `validate_get_examples`?

The problem is that if we split this up, we'd have to refactor a bunch of code in different places that assumes it receives an iterable of examples, like `TransitionSystem.get_actions` for instance. Or we'd need _another_ wrapper function for this case that creates a generator of examples, similar to what we have now, so we can check them as they come in.

honnibal: Ah I forgot that we pass the callable around.

Maybe it's better not to make this check in the components? We can do it in the `Language` class which will provide the good user-experience.

ines: Not sure, but I thought the idea was to have it in the components as well, because in theory, they could be called spearately? It's also what we currently do for the other methods that take `Example` objects.

(It would be a lot easier without but it'd also mean that we could potentially end up with uncaught errors. So I thougth we wanted it to be clean and consistent.)

honnibal: I think it's okay to be a little less strict in how we check inputs within the components, as if people are calling the components directly we can have a bit more trust that they're passing in the right types.

PR #5909, opened 2020-08-11
Use init parameter of Lemmatizer

graue70:
## Description

It was probably overlooked. Should there be a unit test covering this? In that case, could somebody else add it? I'm not familiar with the testing practices of this project.

### Types of change

Bug fix

## Checklist


honnibal:
Thanks, yes this was overlooked. There should be a test but develop is moving quickly, and we should have one test to cover all the components. So I'll just merge this.

PR #5910, opened 2020-08-12
Add Armenian sentence-final verchaket,  Greek question mark and Arabic question mark to default punct

bittlingmayer:
## Description

See https://github.com/explosion/spaCy/issues/5911

I've simply added `:`, `;` and `؟` to the defaults.  They're unique to their respective scripts, so they don't clash with any others, and it's useful to have them in `xx`.

### Types of change
This a bug fix.  Sentencizer is available but completely broken for Armenian.

## Checklist
- [x] I ran the tests, and all new and existing tests passed.  (However, these features are not tested, which is why they were broken for so long.)


honnibal:
Thanks!

PR #5912, opened 2020-08-12
Update token.md

tilusnet:
`is_oov` description didn't make sense.

## Checklist


honnibal:
@tilusnet Actually it's more that the feature itself is kind of problematic and confusing. We're considering removing this for v3.

Given the current behaviour, I think the current documentation is more correct than your fix.

tilusnet:
Thanks @honnibal 
Yes, in the meantime I noticed from the data that `is_oov` does not always mean out-of-vocabulary as such.

Cool, I am happy if you are aware of this misalignment. Glad you are thinking about changing it.

PR #5920, opened 2020-08-14
Replace verbose warnings with logging, tidy up, small fixes

ines:
## Description

This PR includes various small fixes to warnings and logging, and some other small things I came across along the way.

- [x] add custom logger and set logging level via `--verbose` flag in `train` command (can also be adopted for other commands if we want to include more debugging logs in the library)
- [x] expose logger as `spacy.logger`
- [x] replace verbose lexeme norm warning with a logging statement
- [x] update test suite to either check for or ignore known warnings – IMO our own test suite should always come out green and never actually raise our own warnings
- [x] remove old `env_opt` mechanism – this is now redundant and all handled by the config
- [x] correctly pass the optimizer defined in the config to `nlp.begin_training` and `nlp.resume_training` (previously, this created the default optimizer, which seems inconsistent?)
- [x] simpliy default optimizer
- [x] small fixes to data reading and tok2vec listener

### TODO

- [x] check e4d0990 – is this correct?

### Types of change
bug fix, enhancement

## Checklist

### DIFF
--- a/spacy/__init__.py
+++ b/spacy/__init__.py
@@ -14,7 +14,7 @@ from . import pipeline  # noqa: F401
 from .cli.info import info  # noqa: F401
 from .glossary import explain  # noqa: F401
 from .about import __version__  # noqa: F401
-from .util import registry  # noqa: F401
+from .util import registry, logger  # noqa: F401
 
 from .errors import Errors
 from .language import Language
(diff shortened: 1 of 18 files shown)


honnibal:
>  check e4d0990 – is this correct?

Yes, that's correct. I hadn't thought of that edge-case. We'll need an analogous fix in `spacy-transformers`.

PR #5928, opened 2020-08-17
Fix is_new_osx() in setup.py to identify macOS 11 and later correctly

ryandesign:
## Description
This PR fixes the `is_new_osx()` function in setup.py to identify macOS 11 and later correctly. It also fixes the function's doc string which misstated what the function does.

### Types of change
Bug fix

## Checklist

### DIFF
--- a/.github/contributors/ryandesign.md
+++ b/.github/contributors/ryandesign.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
+documentation, or any other material posted or submitted by you to the project.
+
+2. With respect to any worldwide copyrights, or copyright applications and
+registrations, in your contribution:
+
+    * you hereby assign to us joint ownership, and to the extent that such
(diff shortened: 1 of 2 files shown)


svlandeg:
Thanks! There's a related PR on `develop` from some time back: PR #5376 

IIRC, at the time I submitted this to `develop` and not `master`, as there was some difference between Python 2 and 3, and support for Python 2 is dropped on the `develop` branch (which is work in progress for spaCy 3.0).

You're right that the docstring was wrong. And if we can make this function shorter, I'd be all for it!

ryandesign:
I should have mentioned that I tested that the new function worked correctly on python 2.7.18 and 3.8.5. I can test other versions if you wish.

ryandesign:
> There's a related PR on `develop` from some time back: PR #5376

I see... I didn't realize you had a develop branch. I just noticed a problem in your code, checked the default branch (master) to see if the problem still existed, which it did, so I submitted a fix. Hiding your latest code in a branch that is not the default branch makes it harder for people who aren't already familiar with your project to help you.

I also didn't realize that `distutils.util.get_platform()` returns the minimum platform for which the python binary was built (`MACOSX_DEPLOYMENT_TARGET`). (My python binary was built on my OS version so its deployment target is the same as my current OS version.) That distinction is probably important. For example, if you are building on 10.7 but with a deployment target of 10.6, you don't want to involve libc++ since libc++ didn't exist on 10.6. While I was rewriting the function it did occur to me that it should probably be looking at the value of `MACOSX_DEPLOYMENT_TARGET` being used to build spaCy, but it was my intention merely to fix the bug in the function and not change its behavior. I also don't know how to accomplish that as my experience with python is very limited.

Forcing the use of libc++ seems to be the only decision that's based on the result of `is_new_osx()`. Forcing the use of libc++ can be problematic. It's only ok if your use of C++ data structures is confined to your own software (i.e. you do not exchange C++ data structures with any library that you use or any other software that uses your software). So hopefully that is the case for spaCy.

Do you know why spaCy forces the use of libc++? It was done in 378ca4b46df9f88dfa359e994acab763b665d2f7 and the commit message says it fixes a problem but doesn't say what the problem was.

svlandeg:
> Hiding your latest code in a branch that is not the default branch makes it harder for people who aren't already familiar with your project to help you.

We're not hiding anything - we're working on a new release of spaCy v.3 on the `develop` branch, while simultaneously still having made a few new release on v2.3.x, on the `master` branch.
I'm aware that you may not know this information - which is why I pointed it out in my reply.

ryandesign:
> I want to let you know that I received this comment poorly.

I'm sorry, I should have worded it differently. I was frustrated that I had tried to provide a patch to your latest code but had been thwarted by the fact that your latest code was not where I expected to find it. I was upset I had wasted your time in reviewing a change that could not be applied to your latest code. I understand that there are surely valid reasons to develop code in another branch.

honnibal:
@ryandesign Thanks, I understand the frustration and that it was coming from a good place 👍

honnibal:
We should refactor the setup.py more generally and test on the different OSX versions we support. So I'm going to close this PR and revisit the issue later.

PR #5930, opened 2020-08-18
output_file required, spacy-transformers prefered instead of required

svlandeg:
### Description 
`init config` is going to be a very useful & crucial command for users to start exploring v.3. I think it can be improved a bit with respect to user experience though:
- If you don't specify any parameters, like I typically do when testing a command for the first time, it writes to stdout AND disables the logging. I propose to avoid disabling the logging by accident, by requiring the `output_file`. It can still be "-", just not by default.
- If you run this without any parameters and without having `spacy-transformers` installed, it errors with a long stacktrace and a message saying that "Using a transformer-based pipeline requires spacy-transformers to be
installed." (which you never really asked for). Instead, I would avoid creating a transformer-based config if the user doesn't have the correct library installed. But we'd still generate an informative message that this would be prefered: "Recommend to install 'spacy-transformers' to create a more efficient transformer-based pipeline."

### Types of change
UX enhancement

## Checklist

### DIFF
--- a/spacy/cli/init_config.py
+++ b/spacy/cli/init_config.py
@@ -39,7 +39,7 @@ class RecommendationSchema(BaseModel):
 @init_cli.command("config")
 def init_config_cli(
     # fmt: off
-    output_file: Path = Arg("-", help="File to save config.cfg to (or - for stdout)", allow_dash=True),
+    output_file: Path = Arg(..., help="File to save config.cfg to (or - for stdout, disabling logging)", allow_dash=True),
     lang: Optional[str] = Opt("en", "--lang", "-l", help="Two-letter code of the language to use"),
     pipeline: Optional[str] = Opt("tagger,parser,ner", "--pipeline", "-p", help="Comma-separated names of trainable pipeline components to include in the model (without 'tok2vec' or 'transformer')"),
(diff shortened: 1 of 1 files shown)


honnibal:
I think `-` instead of the output file is generally good, so I'm happy to have it here. What do you think @ines ?

Regarding the transformers, I think this is just about catching the error. I'm nervous about changing the recommendation based on what you had installed...The catch is that you need to install quite a lot of stuff to get `spacy-transformers` set up, and I think we need the library to do the config, otherwise we can't validate I think?

svlandeg:
> The catch is that you need to install quite a lot of stuff to get `spacy-transformers` set up

Right - so that's why I would avoid generating a transformers config if the user doesn't have the library installed yet? It just feels like that won't lead to the quickest path of trying out these new configs with v3, if the first thing that happens is that you need to install a new library? (worse, in fact, currently the `init config` command would just crash/error and not give you a config at all)

ines:
I guess we could make the `--cpu` flag a `--gpu` flag, so people need to set it explicitly if they have a GPU available and want to use GPU-friendly pipelines (i.e. transformer-based pipelines). Then it's maybe less surprising if it tells you to install an additional package?

I see what you mean about stdout as the default and I'm not sure what the best solution is  🤔 On the one hand, I always like it if a command just outputs stuff and doesn't need me to make it write to a file, which I then have to open etc. But if we write to sdtdout, there's not really a good way to output additional info because people need to be able to pipe the output forward to a file.

(Btw, related idea I just had to give people an `init config` command to get started with: we could make the quickstart widget in the docs also produce the `init config` commnand with arguments that match whatever they selected in the menu above.)

PR #5933, opened 2020-08-18
docs: custom reader and batcher

svlandeg:
## Description

- Training docs: example with custom reader and batcher (creating variants on the fly, filtering examples per batch)
- Several typo fixes across the docs
- Some formatting
- `python:setup` bugfix for Windows

### Types of change
docs

## Checklist


svlandeg:
Also, as a general comment. I changed `system.seed` to `system:seed` in two places, assuming this would always just be an integer value. I think we need to be really careful about these cases in the docs (and they're difficult to ctrl-F). I often get confused with the `:` and `.` notation myself :|
So probably good to keep an eye out for these cases and double-check them!

ines:
>  I think we need to be really careful about these cases in the docs (and they're difficult to ctrl-F). I often get confused with the : and . notation myself :|

Yeah, I know what you mean... maybe it would make sense to normalise it and use the same syntax for both sections and other values. I can see if I can make this work.

Initially, I thought the distinction here could be useful, but now that I think about it, whether something is just a regular value or a section can easily change, and this would affect all variables.

svlandeg:
If we can make it work with the same syntax, I think that would be even better. Like "size" can sometimes be a number, or a generator, etc.

PR #5936, opened 2020-08-19
Fix `Doc.from_docs` for empty string Doc

tamuhey:
## Description

`Doc.from_docs` will fail if the argument `docs` contains empty string docs, e.g. `len(doc.text) == 0` because `doc[-1]` is referenced in the method.

The new method doesn't add white space to emtpy docs, even if `ensure_space = True`. I'd like your opinion about it.

### Types of change

bug fix

## Checklist


adrianeboyd:
Thanks for testing / catching this!

I agree about not adding whitespace after empty docs. The idea is that the merged doc contains the exact same tokens as the original docs, so adding a whitespace token after the empty doc would go against that.

PR #5937, opened 2020-08-19
Fix `Doc.from_docs` for `doc.ents`

tamuhey:
## Description

`Doc.from_docs` doesn't merge `doc.ents` properly.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -884,7 +884,7 @@ cdef class Doc:
 
         if attrs is None:
             attrs = [LEMMA, NORM]
-            if all(doc.is_nered for doc in docs):
+            if any(doc.is_nered for doc in docs):
                 attrs.extend([ENT_IOB, ENT_KB_ID, ENT_TYPE])
             # TODO: separate for is_morphed?
             if all(doc.is_tagged for doc in docs):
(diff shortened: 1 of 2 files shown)


spacy/tokens/doc.pyx
```diff
@@ -884,7 +884,7 @@ cdef class Doc:
 
         if attrs is None:
             attrs = [LEMMA, NORM]
-            if all(doc.is_nered for doc in docs):
+            if any(doc.is_nered for doc in docs):
                 attrs.extend([ENT_IOB, ENT_KB_ID, ENT_TYPE])
```
svlandeg, line 888: Agreed that `any` makes more sense! What about the cases for `is_tagged` and `is_parsed` just a few lines below this one - shouldn't they be fixed in a similar fashion too?

svlandeg (approved):
Hi @tamuhey, thanks for the PR(s)! `Doc.from_docs` is new in v.3, and apparently requires more testing, so it's great to add more tests to the suite. 

We definitely appreciate you hunting bugs even before the release :-)

adrianeboyd:
Since there were a few related problems with these attributes, I tried to address this more generally in #5962. If that PR is merged with the relevant changes, then this one can be closed, I think.

tamuhey:
@adrianeboyd Cool! Why don't you add the test of this PR to #5962 ?

adrianeboyd:
Good suggestion, thanks!

adrianeboyd:
Okay, I think this is fixed by #6044.

PR #5952, opened 2020-08-22
Hebrew like num

idoshr:
Hebrew like num func


svlandeg (approved):
Looks good to me, thanks for your contribution!

PR #5954, opened 2020-08-23
Fix handling of optional [pretraining] block

ines:
## Description

* The `[pretraining]` block now doesn't default to `null` anymore, which would cause Thinc to (correctly) raise an error because it's a top-level key that's not a section.
* Make `[pretraining]` default to `{}` (empty block) and adjust checks accordingly.
* Read `training.seed` in `debug-model`, not `pretraining.seed` (possibly copy-paste mistake?)

### Open questions / todos

- [x] How should we manage the presence / absence of the `[pretraining]` block? We currently reset it in `init config` for transformer-based models, so some `nlp.config`s may have it, and some don't. What should be the recommended path if you want to pretrain using a config and your config doesn't have a `[pretraining]` block?

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/cli/debug_model.py
+++ b/spacy/cli/debug_model.py
@@ -51,7 +51,7 @@ def debug_model_cli(
     with show_validation_error(config_path):
         config = util.load_config(config_path, overrides=config_overrides)
         nlp, config = util.load_model_from_config(config_path)
-    seed = config["pretraining"]["seed"]
+    seed = config["training"]["seed"]
     if seed is not None:
         msg.info(f"Fixing random seed: {seed}")
         fix_random_seed(seed)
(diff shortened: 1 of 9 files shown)


honnibal:
Since pretraining is a feature that's only useful a minority of the time and isn't on the main first-use path, I would rather priortise not having it get in the way. Even if users have to paste in a block of text to pretrain, I think that would be okay, better than having this whole `[pretraining]` block in every config.

ines:
@honnibal Okay, so maybe we should even remove it from the `default_config.cfg` then, at least until we have a better solution? The nice thing is that auto-filling a config with a pretraining block will work once you have at least one setting in there (need to double-check but should be true). So the path to a filled pretraining config is pretty straightforward.

honnibal:
Yes, removing it from the default config sounds good.

PR #5959, opened 2020-08-24
Add 3rd party licenses file

adrianeboyd:
## Description

Add missing 3rd-party-licenses from #5694.

## Checklist

### DIFF
--- a/licenses/3rd_party_licenses.txt
+++ b/licenses/3rd_party_licenses.txt
@@ -1,0 +1,38 @@
+Third Party Licenses for spaCy
+==============================
+
+NumPy
+-----
+
+* Files: setup.py
+
+Copyright (c) 2005-2020, NumPy Developers.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+    * Redistributions of source code must retain the above copyright
+       notice, this list of conditions and the following disclaimer.
+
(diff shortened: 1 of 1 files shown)


ines:
👍

This is only going to be a single file, right? If so, can we put this at the root of the repo and avoid the extra `licenses` directory? (Or was there another reason we did this that I don't remember?)

adrianeboyd:
A lot of other packages do something similar and I thought it looked less clutter-y this way (I couldn't think of a shorter filename, but maybe you can?), but it can be anywhere as long the manifest/setup is adjusted accordingly.

ines:
Ah okay! I'm just always wary of adding more directories to the root, especially if they only contain very few files.

What about a file `OTHER_LICENSES` (?) or something like that? Is this too bad? @honnibal  What do you think?

honnibal:
I think a `licenses` directory makes sense? I don't see the problem with having another top-level directory, it looks obviously uninteresting so I don't think it's distracting.

ines:
But are we, realistically, ever going to have more than one file in there? It just seems pointless to me to have a directory for one file?

adrianeboyd:
I think some projects have one file per project / license instead, which may be why other projects had a directory instead. (The next step is to move the scorer license bits here.)

honnibal:
I do think it's cleaner to have the subdirectory, and it'll make it easier going forward. Plus it sorts in a better way.

PR #5962, opened 2020-08-24
Update Doc flags for separate pipes

adrianeboyd:
## Description

- Add flag `Doc.is_lemmatized`
- Separate `is_tagged` into `is_tagged` / `is_morphed` / `is_lemmatized`
- Add `Doc._get_array_attrs()` to have a single method that returns the
relevant attributes for `to_array` for a given doc: `Span.as_doc()`, `Doc.from_docs()`, `Doc.to_bytes()`
- Add `NORM` and `ENT_ID` to default attributes for `DocBin`

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -397,8 +397,8 @@ class Errors:
     E154 = ("One of the attributes or values is not supported for token "
             "patterns. Please use the option validate=True with Matcher, "
             "PhraseMatcher, or EntityRuler for more details.")
-    E155 = ("The pipeline needs to include a tagger in order to use "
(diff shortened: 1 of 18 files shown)


adrianeboyd:
Ugh, I don't really like having to manage all these flags. The naming is hard (`is_posed`?) and I keep ending up with some attribute missing when documents are converted because a flag wasn't set.

Are there performance reasons for not having them all be derived like `is_nered` and `is_sentenced`? Nothing jumps out at me but maybe I'm overlooking something?

adrianeboyd:
Replaced by #6044.

PR #5963, opened 2020-08-24
Add AttributeRuler.score

adrianeboyd:
## Description

Add scoring for TAG / POS / MORPH / LEMMA if these are present in the assigned token attributes.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/pipeline/attributeruler.py
+++ b/spacy/pipeline/attributeruler.py
@@ -4,9 +4,11 @@ from pathlib import Path
 
 from .pipe import Pipe
 from ..errors import Errors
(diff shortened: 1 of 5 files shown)


adrianeboyd:
You could also potentially add kind of dumb default score weights so the scores get added to a default config, but I'm not sure whether there's really a good way to do this. Having something like:

```
{"tag_acc": 0.25, "pos_acc": 0.25, ...}
```

is probably just going to produce weird default weights more than anything else.

honnibal:
Yeah I think it's best not to include the default score weights.

Issue #5961, opened 2020-08-24
cannot analyze ` ̄ ̄` with japanese models


TatsuyaShirakawa:
## How to reproduce the behaviour

When I tried the following very small script

```python
import spacy
nlp = spacy.load('ja_core_news_sm')
nlp(' ̄ ̄')
```
I got the following error

```shell
>>> nlp(' ̄ ̄')
nlp(' ̄ ̄')
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/spacy/lang/ja/__init__.py", line 106, in get_dtokens_and_spaces
    word_start = text[text_pos:].index(word)
ValueError: substring not found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/spacy/language.py", line 441, in __call__
    doc = self.make_doc(text)
  File "/usr/local/lib/python3.7/site-packages/spacy/lang/ja/__init__.py", line 281, in make_doc
    return self.tokenizer(text)
  File "/usr/local/lib/python3.7/site-packages/spacy/lang/ja/__init__.py", line 145, in __call__
    dtokens, spaces = get_dtokens_and_spaces(dtokens, text)
  File "/usr/local/lib/python3.7/site-packages/spacy/lang/ja/__init__.py", line 108, in get_dtokens_and_spaces
    raise ValueError(Errors.E194.format(text=text, words=words))
ValueError: [E194] Unable to aligned mismatched text ' ̄ ̄' and words '[' ', '̄', ' ̄']'.
```

The minimal Dockerfile is here

```
FROM python:3.7

RUN pip install spacy
RUN python -m spacy download ja_core_news_sm
```

## Your Environment

* Operating System: Linux 04a7a76544e5 4.19.76-linuxkit #1 SMP Thu Oct 17 19:31:58 UTC 2019 x86_64 GNU/Linux
* Python Version Used: 3.7.7
* spaCy Version Used: 2.3.2
* Environment Information: Minimal Dockerfile is as bellow
```
FROM python:3.7

RUN pip install spacy
RUN python -m spacy download ja_core_news_sm
```

svlandeg:
Thanks for the report! This is definitely a bug.

[...]

PR #5969, opened 2020-08-25
fix ja leading spaces

hiroshi-matsuda-rit:
Bug fix for `spacy.lang.ja`: exception raised when dtoken.surface starts with " " #5961 

## Description
Change the condition of space after by seeing the surface of the next token.

### Types of change
 a bug fix

## Checklist


svlandeg (commented):
Thanks for such a quick fix! I suggest to have the original case from that issue as a unit test? (and preferably some additional cases as well).

hiroshi-matsuda-rit:
@svlandeg I added a text to NAUGHTY_STRINGS. Is it okay?

svlandeg:
I think so! @adrianeboyd : does this look OK to you?

adrianeboyd:
I think it looks fine!

hiroshi-matsuda-rit:
CI tests done.
How about develop branch?
May I send the same PR to develop?

svlandeg:
There's no need for that, we sync such fixes periodically!

PR #5977, opened 2020-08-26
Remove the 'fix pretrained vector names' hack

honnibal:
Seems to not be necessary?

## Checklist


spacy/language.py
```diff
@@ -1646,25 +1641,6 @@ class FactoryMeta:
...
-    # data
-    if "vectors" in nlp.meta and "name" in nlp.meta["vectors"]:
-        nlp.vocab.vectors.name = nlp.meta["vectors"]["name"]
-    elif not nlp.vocab.vectors.size:
-        nlp.vocab.vectors.name = None
-    elif "name" in nlp.meta and "lang" in nlp.meta:
-        vectors_name = f"{nlp.meta['lang']}_{nlp.meta['name']}.vectors"
-        nlp.vocab.vectors.name = vectors_name
-    else:
-        raise ValueError(Errors.E092)
```
ines, line 1660: We should also remove `E092` from the `errors.py`, since it was only used here.

PR #5980, opened 2020-08-26
update to Tok2Vec and Transformer listeners

svlandeg:
~~Tests will fail until https://github.com/explosion/thinc/pull/386 is merged/released.~~ 

## Description
- `Tok2Vec` layer sets the `doc.tensor` through `spacy.tensor_setter.v1` (can be overwritten)
- `Tok2VecListener` doesn't rely on a specific attribute being set anymore
- wrote unit tests for the `Tok2VecListener`
- rename `spacy-transformers.Tok2VecListener.v1` to `spacy-transformers.TransformerListener.v1` (cf [PR for spacy-trf 1.0.0a7](https://github.com/explosion/spacy-transformers/pull/212))
- define `annotation_setters` catalogue in spaCy instead of spacy-trf
- bump to 3.0.0a13

I'm working on some related docs in a different branch, so if this gets merged, I'll make the corresponding changes there!

### Types of change
enhancement

## Checklist


spacy/tests/pipeline/test_tok2vec.py
```diff
@@ -0,0 +1,196 @@
...
+    tok2vec = build_Tok2Vec_model(
+        embed_arch(**embed_config),
+        encode_arch(**encode_config)
+    )
+    tok2vec.initialize(docs)
+    vectors, backprop = tok2vec.begin_update(docs)
+    assert len(vectors) == len(docs)
+    assert vectors[0].shape == (len(docs[0]), width)
+    backprop(vectors)
+
```
svlandeg, line 85: Not clear from the GH diff, but everything above this comment is just from the old file `spacy/tests/test_tok2vec.py` (which I moved). Everything below, are new tests.

spacy/util.py
```diff
@@ -93,6 +93,7 @@ class registry(thinc.registry):
     # environment. spaCy models packaged with `spacy package` will "advertise"
     # themselves via entry points.
     models = catalogue.create("spacy", "models", entry_points=True)
+    annotation_setters = catalogue.create("spacy", "annotation_setters", entry_points=True)
```
svlandeg, line 96: Hm, now I started wondering whether this is the right move, as the signature of spaCy's `Tok2Vec` `annotation_setter` is different to the one used in `spacy-transformers`. But it would also be weird to make two separate registries for them, no?

honnibal:
Ugh, I think we have to change course. The problem in PR is pretty bad: https://github.com/explosion/thinc/pull/386

If we have to choose between making the listeners ugly by having them listen to a specific doc attribute, and making _every other layer_ ugly by making them check for listeners, then it's definitely better to leave the problem within the listeners.

Is there a way to not set the data on the doc AND solve the "empty predict" problem? Could we give the listeners some way to come up with a fake output if their input isn't set?

honnibal:
Just remembered this hack: https://github.com/explosion/spaCy/blob/develop/spacy/pipeline/transition_parser.pyx#L432

svlandeg:
> If we have to choose between making the listeners ugly by having them listen to a specific doc attribute, and making every other layer ugly by making them check for listeners, then it's definitely better to leave the problem within the listeners.

I just want to clarify - the problem is not really caused by this PR. If you store the data in an attribute, like currently on `develop`, we still run into issues because `doc.tensor` would be empty. We just never noticed that before because we weren't calling `model.initialize` before with listeners in the model's layers.

svlandeg:
But yea I think we can solve this (with or without attribute) if we don't verify the inputs when `is_train` is `False`, and provide a better result "empty result" instead of `[array([], dtype=float32)]` which is what currently happens. Because now `with_array._get_array` handles that semi-empty input badly, and things go wrong from there. I'll have a look.

honnibal:
Okay yeah understood.

I think this does point against unifying the listeners though, because they each need to know how to make their empty result. How about something like this for the `Tok2VecListener`?

```
def forward(model: Tok2VecListener, inputs, is_train: bool):
    """Supply the outputs from the upstream Tok2Vec component."""

    if is_train:
        model.verify_inputs(inputs)
        return model._outputs, model._backprop
    else:
        if model._outputs is None:
            outputs = [model.ops.alloc2f(len(doc), width) for doc in inputs]
        else:
            outputs = model._outputs
        model._outputs = None
        return outputs, lambda dX: []
```

spacy/pipeline/tok2vec.py
```diff
@@ -295,4 +314,7 @@ def forward(model: Tok2VecListener, inputs, is_train: bool):
         model.verify_inputs(inputs)
         return model._outputs, model._backprop
     else:
-        return [doc.tensor for doc in inputs], lambda dX: []
+        outputs = model._outputs
+        if outputs is None:
+            outputs = [model.ops.alloc2f(len(doc), model.get_dim("nO")) for doc in inputs]
```
honnibal, line 319: Should we unset the `_outputs` and `_backprop` here? It's kind of hard to know...Maybe it's good to allow multiple runs of the model? There's not really a great answer. Maybe we should rely on the batch ID rather than checking whether `_outputs` is None?

svlandeg: Ah, yes, good idea to rely on `model._batch_id`, that also makes the code more clear. I don't really see why we should reset anything though? In theory the listener should be allowed to ask the results for the same batch ID twice?

honnibal: Yep, makes sense.

honnibal:
Ugh I think the problem's worse than I thought...I think we won't be able to get away from relying on `doc.tensor` and `doc._.trf_data` either. The problem is that different components can batch up the data differently, so during prediction we can easily be predicting on a different batch. So we need a per-doc view of the data...which is exactly what `doc.tensor` and `doc._.trf_data` provide. If we're going to break apart the batch, there's no reason to pass the results via the backchannel instead of communicating via the doc.

PR #5989, opened 2020-08-28
Update suffixes example

jgutix:
The current example will throw `TypeError: can only concatenate list (not "tuple") to list`

### Types of change

Documentation

## Checklist


honnibal:
Thanks!

PR #5990, opened 2020-08-28
update error type

svlandeg:
## Description
[This edit](https://github.com/explosion/spaCy/commit/d1780db6a4b94669d9400af403bb6d5997aa5bd5) caused a test failure, fixing here. As the error type is now different for the different options, I split it into two different tests for clarity?

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tests/test_cli.py
+++ b/spacy/tests/test_cli.py
@@ -1,4 +1,6 @@
 import pytest
+from click import NoSuchOption
+
 from spacy.gold import docs_to_json, biluo_tags_from_offsets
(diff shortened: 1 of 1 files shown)


ines:
Omg sorry, forgot to adjust this!

(Basically, I added the `NoSuchOption` error because for commands that accept CLI overrides, we just catch all extra arguments and then analyse them. But if you typo a regular argument name, you'd then get the "can't override section" error, which is super confusing. If there's no `.` in the override, we should just raise the default CLI error. It'd mean you also get that when you write `--nlp something`, but... that's probably quite rare.)

PR #5993, opened 2020-08-28
Add support for loaded but disabled components

ines:
## Description

This PR adds support for loading pipeline components but disabling them so they're not run as part of the pipeline. This lets us ship models with components that may or may not run – like a pretrained sentence segmenter vs. sentence segmentation by the parset etc.

- Store the full pipeline and pipe names as `Language.components` and `Language.component_names`, respectively
- Store the disabled components in `Language.disabled`
- Make `Language.pipeline` a property that only uses components that are not disabled
- Add `Language.enable_pipe` and `Language.disable_pipes` to enable/disable a component
- Simplify `Language.select_pipes`: we don't have to keep a copy of the pipeline anymore and can just enable/disable
- Add `exclude` to `spacy.load` (and related loading utilities) to explicitly **exclude** and not load a component. The `disable` keyword argument now specifies components to load but disable
- Ensure the settings are reflected correctly in the config and allow config to specify default disabled pipes in as `disabled` in the `[nlp]` block
- Add `SimpleFrozenList` so we can raise custom errors if user tries to write to computed properties like `nlp.pipeline` directly (e.g. `nlp.pipeline.append()`). We still use a list here, even though a tuple would be a more logical fit, because we want to avoid breaking backwards compatibility. Otherwise, even a simple `assert nlp.pipe_names == ["ner", "textcat"]` would raise errors in a user's code because the value would be a tuple.

### TODO

- [x] improve error handling
  * Do we want to raise if a user calls `enable_pipe` on an already enabled component or `disable_pipe` on an already disabled component?
  * Add checks for potential problems when calling `DisabledPipes.restore` if pipe to re-enable doesn't exist anymore (e.g. if user deleted it in the block)
- [x] decide on user-facing API: do we want to expose the "full" pipeline and pipe names? How should we do it so it doesn't get messy?
- [x] document everything

### Types of change
enhancement

## Checklist


honnibal:
> decide on user-facing API: do we want to expose the "full" pipeline and pipe names? How should we do it so it doesn't get messy?

Maybe we could call that `nlp.loaded_components`? It gets all components that have been loaded, not just the "pipeline" (which excludes disabled components).

Or we could just call that, `nlp.components`. Is that messy though?

> Do we want to raise if a user calls enable_pipe on an already enabled component or disable_pipe on an already disabled component?

I think we should let those no-op, so that you don't have to make an extra check to ensure you're enabling a component.

ines:
> Or we could just call that, nlp.components. Is that messy though?

🤔🤔🤔

It's not too bad, actually. It would correspond to `[components]` in the config...

spacy/language.py
```diff
@@ -249,6 +252,16 @@ def config(self) -> Config:
     def config(self, value: Config) -> None:
         self._config = value
 
+    @property
+    def disabled(self) -> List[str]:
```
honnibal, line 256: The question is whether we should return a set here. I think it's a tough trade-off. How often will people just want to make an `in` check?

I think I lean towards returning a `FrozenSet`. Otherwise, a tuple is definitely better than a list, because a list allows `nlp.disabled.append`, which will always be wrong.

ines: I previously used a set here, but it made the user-facing property inconsistent with all the other similar properties and it was already a pain to document. Under the hood, `self._disabled` is a set.

spacy/language.py
```diff
@@ -159,7 +159,8 @@ def __init__(
         self.vocab: Vocab = vocab
         if self.lang is None:
             self.lang = self.vocab.lang
-        self.pipeline = []
+        self.components = []
```
honnibal, line 162: Arguably we should make the data `self._components`, which would allow us to make `components` a property that returns a tuple. This would "protect" components from modification.

I'm not really decided about whether this is good? It's a bit...non-free-spirited, but I think it might be better overall.

ines: I guess making it a property would be more consistent – although, `self.pipeline` also worked this way before.

Not sure we should return a tuple, though, as this would make it inconsistent with all the other properties that all return lists.

spacy/language.py
honnibal: I don't think it's so inconsistent. Most of the rest of the things are lists because they are ordered, but why would people need to iterate over the list of names of disabled components in order? What you'll want to do with the disabled list is check against it. So I think a frozen set is a bit better.

spacy/language.py
```diff
@@ -257,9 +270,28 @@ def factory_names(self) -> List[str]:
...
+    def component_names(self) -> List[str]:
+        """Get the names of the available pipeline components. Includes all
+        active and inactive pipeline components.
+
+        RETURNS (List[str]): List of component name strings, in order.
+        """
+        return [pipe_name for pipe_name, _ in self.components]
+
+    @property
+    def pipeline(self) -> List[Tuple[str, Callable[[Doc], Doc]]]:
```
honnibal, line 283: I hadn't thought about the fact that we're returning lists for most of these properties. I think we should change them to tuples.

It's especially important for the pipeline, because users may well have code like `nlp.pipeline.append`. That will silently do nothing if we have this as a list.

ines: Okay, maybe we could have our own frozen list subclass, though (similar to `SimpleFrozenDict`) that lets us raise custom errors? The default error for a tuple is always kinda cryptic (especially if the user doesn't understand what's going on), so it'd be nice to raise more specific errors here.

honnibal: Sounds great, definitely up for that.

honnibal: (Make it a tuple subclass though?)

spacy/language.py
ines: I don't think it's too bad to have the disabled components be returned in the same order they appear in the original components? Everything else would be kinda confusing.

spacy/language.py
honnibal: I think we'd have to use some `__getattr__ trickery to do the raising though. We can't just have an `append` method, or it'll mess up people's type checks.

honnibal: Naming ideas:

* `ComputedTuple`
* `PropertyTuple`
* `ViewTuple`
* `AttributeView`
* `ImmutableView`

PR #5994, opened 2020-08-28
Allow component decorators to re-run with same function

ines:
## Description

This is especially relevant for environments where spaCy is live-reloaded (notebooks, Streamlit). The `@Language.factory` and `@Language.component` decorators should be allowed to re-run with the same function. This is a bit trickier for `@Language.component` where the decorator itself creates the factory function (and once that's done, it's harder to check whether the original stateless callable is the one we're using in the created factory).

(We can probably solve the `is_same_func` check more elegantly?)

## Checklist

### DIFF
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -689,7 +689,7 @@ def is_same_func(func1: Callable, func2: Callable) -> bool:
     same_name = func1.__qualname__ == func2.__qualname__
     same_file = inspect.getfile(func1) == inspect.getfile(func2)
     same_code = inspect.getsourcelines(func1) == inspect.getsourcelines(func2)
-    return all([same_name, same_file, same_code])
+    return same_name and same_file and same_code
 
 
 def get_cuda_stream(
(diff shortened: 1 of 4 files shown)


spacy/util.py
```diff
@@ -673,6 +673,25 @@ def get_object_name(obj: Any) -> str:
...
+    func1 (Callable): The first function.
+    func2 (Callable): The second function.
+    RETURNS (bool): Whether it's the same function (most likely).
+    """
+    if not callable(func1) or not callable(func2):
+        return False
+    same_name = func1.__qualname__ == func2.__qualname__
+    same_file = inspect.getfile(func1) == inspect.getfile(func2)
+    same_code = inspect.getsourcelines(func1) == inspect.getsourcelines(func2)
+    return all([same_name, same_file, same_code])
```
honnibal, line 692: Check looks solid! I think maybe this way of writing the logic is a bit nicer though.

```suggestion
    return same_name and same_file and same_code
```

PR #5998, opened 2020-08-29
Update usage docs for lemmatization and morphology

adrianeboyd:
## Description

Update usage docs for lemmatization and morphology

### Types of change

Docs

## Checklist


ines:
Yay, thanks, this is all really good 👍

I'll go ahead and merge. I might make one or two small modifications to the executable examples (maybe change some of the `assert`s to `print` so the user sees some output when they run it), but I'll just do that directly on `develop` so we don't keep triggering the CI on here.

PR #6003, opened 2020-08-31
Add as_spans to Matcher/PhraseMatcher and deprecate .pipe

ines:
## Description

Pretty basic, but it works? `Matcher._call__` and `PhraseMatcher.__call__` now both support an `as_spans` flag that returns a list of span objects, using the `match_id` as the span label.

This PR also deprecates `Matcher.pipe` and `PhraseMatcher.pipe`, which have always been pretty useless and misleading.

#### Todo

- [x] add docs

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -176,18 +176,10 @@ cdef class Matcher:
         return (self._callbacks[key], self._patterns[key])
 
     def pipe(self, docs, batch_size=1000, return_matches=False, as_tuples=False):
-        """Match a stream of documents, yielding them in turn.
-
-        docs (Iterable[Union[Doc, Span]]): A stream of documents or spans.
-        batch_size (int): Number of documents to accumulate into a working set.
-        return_matches (bool): Yield the match lists along with the docs, making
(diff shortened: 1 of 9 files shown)


spacy/matcher/matcher.pyx
```diff
@@ -203,13 +203,16 @@ cdef class Matcher:
                 else:
                     yield doc
 
-    def __call__(self, object doclike):
+    def __call__(self, object doclike, as_spans=False):
```
honnibal, line 206:
```suggestion
    def __call__(self, object doclike, *, as_spans=False):
```

spacy/matcher/phrasematcher.pyx
```diff
@@ -216,13 +217,16 @@ cdef class PhraseMatcher:
                 result = internal_node
             map_set(self.mem, <MapStruct*>result, self.vocab.strings[key], NULL)
 
-    def __call__(self, doc):
+    def __call__(self, doc, as_spans=False):
```
honnibal, line 220:
```suggestion
    def __call__(self, doc, *, as_spans=False):
```

PR #6012, opened 2020-09-02
Refactor specials and cache checks in tokenizer

adrianeboyd:
## Description

Simplify specials and cache checks (just some minimal refactoring taken from #5772).

### Types of change

?

## Checklist


honnibal:
Looks reasonable to me. Earlier iterations of the tokenizer actually had these together.

You're confident that this computes the same stuff, right? It seems to, but I don't really trust my ability to be sure from reading it. If you've run it on some data and it outputs the same stuff, then go ahead and merge.

adrianeboyd:
I'd tested it before, but I tested some longer texts and all the training corpora again and can confirm that there were no differences.

PR #6018, opened 2020-09-02
Officially support DependencyMatcher

adrianeboyd:
## Description

Add official support for the `DependencyMatcher`. Redesign the pattern specification. Fix and extend operator implementations. Update API docs and add usage docs.

### Patterns

Refactor pattern structure to:

```
{
  "LEFT_ID": str,
  "REL_OP": str,
  "RIGHT_ID": str,
  "RIGHT_ATTRS": dict,
}
```

The first node contains only `RIGHT_ID` and `RIGHT_ATTRS` and all subsequent nodes contain all four keys.

### New operators

Because of the way patterns are constructed from left to right, it's helpful to have `follows` operators along with `precedes` operators. Add operators for simple precedes / follows alongside immediate precedes / follows.

* `.*`: precedes
* `;`: immediately follows
* `;*`: follows

### Operator fixes

* `<` and `<<` do not include the node itself
* Fix reversed order for all operators involving linear precedence (`.`, all sibling operators)
* Linear precedence operators do not match nodes outside the same parse

### Additional fixes

* Use v3 Matcher API
* Support `get` and `remove`
* Support pickling

### Types of change

Enhancement

## Checklist


adrianeboyd:
This is basically ready for review, but I'll mark it as a WIP since it probably needs a bit of discussion.

I've implemented and documented everything based on the left/right pattern syntax described above, but there's probably some room for improvement here. I find the left/right names helpful because it helps me keep the operator direction in mind as in `left OP right` around all the pairs of operators that can go in either direction.

With the left/right syntax, having the first/anchor node look like `{RIGHT_ID: "", RIGHT_ATTRS: {}}` is a little weird. The first node could have its own keys as something like `ID/ATTRS` instead, but I'm not really sure that's better.

I've added a (draft! ugly!) diagram to the usage page of how the patterns are built up from left to right that I think is a relatively helpful way to picture the whole process. I think it's helpful to describe/picture this horizontally for two reasons:

* the operators appear between the nodes in the `left OP right` order
* it doesn't look like a vertical dependency graph, since having it look like it's an actual subtree that would match in the tree would be confusing when the operators can all appear in either direction

I'd appreciate it if someone else took a careful look over the tests, especially the operator tests, to make sure that the expected match counts make sense for all the operators, since it's easy to get confused after looking at it for too long.

The ugly diagram can be improved once the pattern syntax and example are more finalized. (It's ugly enough that I barely wanted to commit it here, but whatever, it gets the point across.)

adrianeboyd:
Also, if anyone has a better suggestion for the follows operator that would be fine. I wanted it to look a bit like `.` without conflicting with other common operators (also in other query languages, not just here). It's not great?

PR #6020, opened 2020-09-03
Updates spaCy Universe for amrlib

bjascob:
Update to the SpaCy Universe website for the addition of "amrlib"

## Description
Simple update to .json file

### Types of change
Documentation change

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -1968,6 +1968,28 @@
             "tags": ["inflection", "lemmatizer"]
         },
         {
+            "id": "amrlib",
+            "slogan": "A python library that makes AMR parsing, generation and visualization simple.",
+            "description": "amrlib is a python module and SpaCy add-in for Abstract Meaning Representation (AMR).  The system can parse sentences to AMR graphs or generate text from existing graphs.  It includes a GUI for visualization and experimentation.",
+            "github": "bjascob/amrlib",
+            "pip": "amrlib",
+            "code_example": [
+                "import spacy",
+                "import amrlib",
+                "amrlib.setup_spacy_extension()",
+                "nlp = spacy.load('en')",
+                "doc = nlp('This is a test of the SpaCy extension. The test has multiple sentences.')",
+                "graphs = doc._.to_amr()",
+                "for graph in graphs:",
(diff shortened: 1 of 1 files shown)


website/meta/universe.json
```diff
@@ -1967,6 +1967,28 @@
...
+            "id": "amrlib",
+            "slogan": "A python library that makes AMR parsing, generation and visualization simple.",
+            "description": "amrlib is a python module and SpaCy add-in for Abstract Meaning Representation (AMR).  The system can parse sentences to AMR graphs or generate text from existing graphs.  It includes a GUI for visualization and experimentation.",
+            "github": "bjascob/amrlib",
+            "pip": "amrlib",
+            "code_example": [
+                "import spacy",
+                "import amrlib",
+                "amrlib.setup_spacy_extension()",
+                "nlp = spacy.load('en')",
```
svlandeg, line 1980: Just FYI - we've stopped using this style in our docs some time ago, and prefer naming the entire package instead, e.g. 
```
nlp = spacy.load('en_core_web_md')
```
There are two main reasons: the symlink doesn't always work well on different platforms, and we don't want to think about the model packages as if there is just one English model. Instead we prefer to make it specific. For now the style with "en" is still supported, but this will change in the future.

website/meta/universe.json
```diff
@@ -1967,6 +1967,28 @@
             "category": ["pipeline"],
             "tags": ["inflection", "lemmatizer"]
         },
+        {
+            "id": "amrlib",
+            "slogan": "A python library that makes AMR parsing, generation and visualization simple.",
+            "description": "amrlib is a python module and SpaCy add-in for Abstract Meaning Representation (AMR).  The system can parse sentences to AMR graphs or generate text from existing graphs.  It includes a GUI for visualization and experimentation.",
```
svlandeg, line 1973: Another FYI - we prefer the spelling "spaCy", i.e. without capitalizing the first letter ;-)

svlandeg (commented):
Looks like a nice addition to the spaCy universe!

svlandeg:
https://spacy.io/universe/project/amrlib :-)

PR #6023, opened 2020-09-03
"model" terminology consistency in docs, messages, CLI

ines:
## Description

This PR makes the terminology used in docs, messages and CLI more consistent by using "(trained) pipeline (package)" in place of "model". Also fixed various small inconsistencies and explanations in the docs along the way. 

The `init model` command has been renamed to `init vocab`.

### Background

It sucks "model" is commonly used to mean so many different things, so we're trying to make this less confusing. Current policy:

* **Architecture:** A function returning a Model instance
* **Weights:** Weights and other state of the model that you'll serialize with to_bytes
* **Model:** A particular model instance, that is in some state
* **(Trained) Pipeline (Package)**: a serialized, optionally trained pipeline like `en_core_web_sm`

### Open questions

- So far, I've mostly touched the user-facing docs and messages and no internals. We still have other places where we use "model" in correctly (based on our new policy). Should we update those as well? For CLI options, we could introduce aliases to make things less breaking if we want to.
  - `model` argument names (less important for internals, more important for CLI arguments and flags)
  - `load_model_` and similar util methods
  - `model-best` / "best model" terminology for training

### Types of change
docs, ux

## Checklist

### DIFF
--- a/spacy/cli/init_model.py
+++ b/spacy/cli/init_model.py
@@ -28,7 +28,7 @@ except ImportError:
 DEFAULT_OOV_PROB = -20
 
 
-@init_cli.command("model")
+@init_cli.command("vectors")
 @app.command(
     "init-model",
     context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
@@ -37,8 +37,8 @@ DEFAULT_OOV_PROB = -20
 def init_model_cli(
     # fmt: off
     ctx: typer.Context,  # This is only used to read additional arguments
-    lang: str = Arg(..., help="Model language"),
-    output_dir: Path = Arg(..., help="Model output directory"),
+    lang: str = Arg(..., help="Pipeline language"),
+    output_dir: Path = Arg(..., help="Pipeline output directory"),
     freqs_loc: Optional[Path] = Arg(None, help="Location of words frequencies file", exists=True),
     clusters_loc: Optional[Path] = Opt(None, "--clusters-loc", "-c", help="Optional location of brown clusters data", exists=True),
     jsonl_loc: Optional[Path] = Opt(None, "--jsonl-loc", "-j", help="Location of JSONL-formatted attributes file", exists=True),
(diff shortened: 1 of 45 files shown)


website/docs/api/language.md
```diff
@@ -827,13 +828,13 @@ subclass of the built-in `dict`. It supports the additional methods `to_disk`
 
 ## Language.to_disk {#to_disk tag="method" new="2"}
 
-Save the current state to a directory. If a model is loaded, this will **include
-the model**.
+Save the current state to a directory. If a trained pipeline is loaded, this
+will **include all model data**.
```
svlandeg, line 832: I think it's not clear what "model" is here. All data for each model in each component of the pipeline?

svlandeg: Maybe something more like "If a trained pipeline is loaded, this will **include all data** for each component, including the trained weights of the internal models" ?

website/docs/usage/linguistic-features.md
```diff
@@ -1218,11 +1220,11 @@ print(doc.text, [token.text for token in doc])
 
 <Infobox title="Important note on tokenization and models" variant="warning">
 
-Keep in mind that your model's result may be less accurate if the tokenization
+Keep in mind that your models' results may be less accurate if the tokenization
 during training differs from the tokenization at runtime. So if you modify a
-pretrained model's tokenization afterwards, it may produce very different
-predictions. You should therefore train your model with the **same tokenizer**
-it will be using at runtime. See the docs on
+trained pipeline' tokenization afterwards, it may produce very different
```
svlandeg, line 1225: 's ?

website/docs/usage/v3.md
```diff
@@ -390,10 +389,10 @@ Note that spaCy v3.0 now requires **Python 3.6+**.
 | `GoldCorpus`                                             | [`Corpus`](/api/corpus)                                                                                      |
 | `KnowledgeBase.load_bulk`, `KnowledgeBase.dump`          | [`KnowledgeBase.from_disk`](/api/kb#from_disk), [`KnowledgeBase.to_disk`](/api/kb#to_disk)                   |
 | `Matcher.pipe`, `PhraseMatcher.pipe`                     | not needed                                                                                                   |
-| `spacy init-model`                                       | [`spacy init model`](/api/cli#init-model)                                                                    |
+| `spacy init-model`                                       | [`spacy init vocab`](/api/cli#init.vocab)                                                                    |
```
svlandeg, line 392: `init-vocab` instead of `init.vocab`

ines: Ah, then the anchor is wrong? The anchor should definitely be `#init-vocab`!

website/docs/api/language.md
ines: Good point! I just edited it and also added something about how `Language.to_disk` delegates to the component method, if available.

website/docs/usage/v3.md
svlandeg: The anchor is `#init-vocab`, right? But you're linking to `init.vocab` here?

ines: Arghhh I misread the diff 🤦‍♀️ Sorry!

spacy/cli/init_model.py
```diff
@@ -242,7 +243,8 @@ def add_vectors(
         if vectors_data is not None:
             nlp.vocab.vectors = Vectors(data=vectors_data, keys=vector_keys)
     if name is None:
-        nlp.vocab.vectors.name = f"{nlp.meta['lang']}_model.vectors"
+        # TODO: Is this correct? Does this matter?
+        nlp.vocab.vectors.name = f"{nlp.meta['lang']}_{nlp.meta['name']}.vectors"
```
ines, line 247: @honnibal Does this matter / do we need this?

honnibal: I think it's good to set this, yeah. Otherwise it might be hard to track down vectors later.

honnibal:
> model argument names (less important for internals, more important for CLI arguments and flags)
> load_model_ and similar util methods
> model-best / "best model" terminology for training

If you squint a little, I don't feel like these are inconsistent or confusing. For instance, `model-best` is talking about the best set of model weights for the pipeline -- each epoch if the models have improved you save out the pipeline to a directory, so when you load back the `model-best` directory you get the epoch with the best models. And if you load back the `model-final` directory you're loading back the pipeline with the models that were at the end of training, the final models. 

There are some places where referring to a string like `en_core_web_sm` as the "model" will be awkward, especially if the function accepts strings like `blank:en`. Naming the alternative will be tricky though, because an argument `pipeline` on the CLI will generally be a comma-separated list or something. So I think in many of these places, `model` won't be a worse name than any alternative, and if it's the name we've been using...well, that's a definite advantage for it, so I wouldn't necessarily change it.

Probably the best candidate for changing that I can see is `init-model` (or is it `init model`? I think maybe `init-model` is redundant now?). I guess this could be `init pipeline`. Feels weird though. `init nlp`? I can't think of anything better than "model" though so I guess I favor keeping it as it is? 

The util functions, we could rename or not, I think that's a relatively low stakes decision, if they're internals.

So I think we should probably just merge this as is? I don't think there are further fixes we need first.

spacy/cli/init_model.py
ines: Yeah, but what about the specific name and how it's generated?

PR #6024, opened 2020-09-03
Registry renaming

ines:
## Description

- `@assets` -> `@misc`
- add `spacy.` prefix to batchers
- add `spacy-transformers.` prefix to span getters (related: explosion/spacy-transformers#214)

### Types of change
enhancement

## Checklist

### DIFF
--- a/website/docs/api/top-level.md
+++ b/website/docs/api/top-level.md
@@ -317,11 +317,11 @@ factories.
 | `loggers`         | Registry for functions that log [training results](/usage/training).                                                                                                                                                                               |
 | `lookups`         | Registry for large lookup tables available via `vocab.lookups`.                                                                                                                                                                                    |
 | `losses`          | Registry for functions that create [losses](https://thinc.ai/docs/api-loss).                                                                                                                                                                       |
+| `misc`            | Registry for miscellaneous functions that return data assets, knowledge bases or anything else you may need.                                                                                                                                       |
 | `optimizers`      | Registry for functions that create [optimizers](https://thinc.ai/docs/api-optimizers).                                                                                                                                                             |
 | `readers`         | Registry for training and evaluation data readers like [`Corpus`](/api/corpus).                                                                                                                                                                    |
 | `schedules`       | Registry for functions that create [schedules](https://thinc.ai/docs/api-schedules).                                                                                                                                                               |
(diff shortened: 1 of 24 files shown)


website/docs/api/top-level.md
```diff
@@ -322,6 +321,7 @@ factories.
 | `readers`         | Registry for training and evaluation data readers like [`Corpus`](/api/corpus).                                                                                                                                                                    |
 | `schedules`       | Registry for functions that create [schedules](https://thinc.ai/docs/api-schedules).                                                                                                                                                               |
 | `tokenizers`      | Registry for tokenizer factories. Registered functions should return a callback that receives the `nlp` object and returns a [`Tokenizer`](/api/tokenizer) or a custom callable.                                                                   |
+| `misc`            | Registry for miscellaneous functions that return data assets, knowledge bases or anything else you may need.                                                                                                                                       |
```
svlandeg, line 324: I alphabetized this list as it was getting too long to look up a registry quickly. Shall we keep it so and move this between `losses` and `optimizers`?

website/docs/api/transformer.md
```diff
@@ -453,7 +453,7 @@ using the `@spacy.registry.span_getters` decorator.
 > #### Example
 >
 > ```python
-> @spacy.registry.span_getters("sent_spans.v1")
+> @spacy.registry.span_getters("custom_sent_spans")
 > def configure_get_sent_spans() -> Callable:
 >     def get_sent_spans(docs: Iterable[Doc]) -> List[List[Span]]:
 >         return [list(doc.sents) for doc in docs]
```
svlandeg, line 459: Come to think about it, this example is a little confusing? The code is the actual implementation for `sent_spans`  in `spacy-transformers`, so there would be no need to register a similar "custom" function in the "spacy" registry? I think we should either use an actual custom example that is not built-in, or show the code as it is in spacy-transformers.

website/docs/api/top-level.md
ines: Ahhh forgot this was alphabetized 😄 So yes, let's put it in the correct place then.

website/docs/api/transformer.md
ines: Hmmm yeah, I see your point... the main reason I renamed this is because if we keep the naming consistent, it would have to be named `spacy-transformers.sent_spans.v1`, and if a user just copy-pasted this example, they'd immediately see an error because (obviously) that function already exists 🤔

svlandeg: Meh, let's just keep it as is then :-)

PR #6034, opened 2020-09-07
Pipe API

svlandeg:
## Description
Some changes to the `Pipe` API:
* Removed the `set_output` hack that would directly set the `nO` dimension
* Instead, enforce that `Pipe` components always receive a few `Example` objects to do shape inference with. This is easily achieved by calling `nlp.begin_training`. 
* Within the `begin_training` method of a component,  `self.model.initialize` should be called with relevant arguments.
   * I wasn't sure about two cases: `multitask.py` (is this currently used?) and `transition_parser.pyx` (which does resizing which I think works fine as is)
* If a component wants to be able to resize its output layer, its model should have an entry `resize_output` in `self.model.attrs`. In this case, `component.is_resizable()` will be `True`.
* ~~I think we should avoid adding labels or inferring other arbitrary properties in the `begin_training` methods, as we ran into trouble with that before, and the provided examples should really only be a small sample. They are not supposed to contain all valid labels, for instance. So instead of adding them automagically, the component now throws an error if the labels are not properly defined beforehand.~~
* Throw an error when a script calls e.g. `component.add_label` when the component's output dimension would become incompatible (e.g. when `component.begin_training` had already been called previously) with the additional label AND the component can not be resized 

I tried to be careful, but the shape inference is sometimes a bit fiddly, and the original `set_output` hack was overwriting most of it. It seems to work now though. I hope this doesn't break anything (if so - we need more tests!)

Note: I wouldn't typically work with `assert...` statements in the main code flow, but I think it's useful to have them in the `begin_training`methods so the code crashes there instead of somewhere in a recursive function in Thinc. If these asserts fail, they point to a programming error, so this really shouldn't happen.

I'll update the docs after this PR gets merged.

### Types of change
enhancement
 
## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1155,21 +1155,21 @@ class Language:
 
         DOCS: https://spacy.io/api/language#begin_training
         """
-        # TODO: throw warning when get_gold_tuples is provided instead of get_examples
         if get_examples is None:
-            get_examples = lambda: []
-        else:  # Populate vocab
-            if not hasattr(get_examples, "__call__"):
-                err = Errors.E930.format(name="Language", obj=type(get_examples))
+            doc = Doc(self.vocab, words=["This is an example sentence."])
+            get_examples = lambda: [Example.from_dict(doc, {})]
+        # Populate vocab
+        if not hasattr(get_examples, "__call__"):
+            err = Errors.E930.format(name="Language", obj=type(get_examples))
+            raise ValueError(err)
+        for example in get_examples():
+            if not isinstance(example, Example):
+                err = Errors.E978.format(
+                    name="Language.begin_training", types=type(example)
+                )
                 raise ValueError(err)
-            for example in get_examples():
-                if not isinstance(example, Example):
-                    err = Errors.E978.format(
-                        name="Language.begin_training", types=type(example)
-                    )
-                    raise ValueError(err)
-                for word in [t.text for t in example.reference]:
-                    _ = self.vocab[word]  # noqa: F841
+            for word in [t.text for t in example.reference]:
+                _ = self.vocab[word]  # noqa: F841
         if device >= 0:  # TODO: do we need this here?
             require_gpu(device)
             if self.vocab.vectors.data.shape[1] >= 1:
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -656,7 +656,7 @@ class Language:
         return resolved[factory_name]
 
     def create_pipe_from_source(
-        self, source_name: str, source: "Language", *, name: str,
+        self, source_name: str, source: "Language", *, name: str
     ) -> Tuple[Callable[[Doc], Doc], str]:
         """Create a pipeline component by copying it from an existing model.
 
@@ -1156,7 +1156,10 @@ class Language:
         DOCS: https://nightly.spacy.io/api/language#begin_training
         """
         if get_examples is None:
-            doc = Doc(self.vocab, words=["This is an example sentence."])
+            util.logger.debug(
+                "No 'get_examples' callback provided to 'Language.begin_training', creating dummy examples"
+            )
+            doc = Doc(self.vocab, words=["x", "y", "z"])
             get_examples = lambda: [Example.from_dict(doc, {})]
         # Populate vocab
         if not hasattr(get_examples, "__call__"):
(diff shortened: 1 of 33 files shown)


spacy/pipeline/transition_parser.pyx
```diff
@@ -406,16 +406,14 @@ cdef class Parser(Pipe):
...
-            raise ValueError(err)
+        self._ensure_examples(get_examples)
         self.cfg.update(kwargs)
         lexeme_norms = self.vocab.lookups.get_table("lexeme_norm", {})
         if len(lexeme_norms) == 0 and self.vocab.lang in util.LEXEME_NORM_LANGS:
             langs = ", ".join(util.LEXEME_NORM_LANGS)
             util.logger.debug(Warnings.W033.format(model="parser or NER", langs=langs))
         actions = self.moves.get_actions(
-            examples=get_examples(),
+        #    examples=get_examples(),
```
svlandeg, line 416: I wasn't sure about this, but I think we don't need those examples. Moreover, we shouldn't rely on these examples to cover all possible cases/labels/actions?

spacy/pipeline/entity_linker.py
```diff
@@ -153,10 +154,17 @@ def begin_training(
...
+        self._require_kb()
         nO = self.kb.entity_vector_length
-        self.set_output(nO)
-        self.model.initialize()
+        doc_sample = []
+        vector_sample = []
+        for example in islice(get_examples(), 10):
+            doc_sample.append(example.x)
+            vector_sample.append(self.model.ops.alloc1f(nO))
+        assert doc_sample
```
ines, line 165: I agree with the reasoning for regular asserts here. Maybe we could have a default error message we can show alongside it that says something like "This is likely a bug in spaCy" and then use that? For instance:

```suggestion
        assert doc_sample, Errors.EXXX
```

spacy/language.py
```diff
@@ -1155,21 +1155,21 @@ def begin_training(
 
         DOCS: https://nightly.spacy.io/api/language#begin_training
         """
-        # TODO: throw warning when get_gold_tuples is provided instead of get_examples
         if get_examples is None:
-            get_examples = lambda: []
-        else:  # Populate vocab
-            if not hasattr(get_examples, "__call__"):
-                err = Errors.E930.format(name="Language", obj=type(get_examples))
+            doc = Doc(self.vocab, words=["This is an example sentence."])
```
ines, line 1159: It kinda doesn't matter but the use of `words` with a single sentence string here may confuse people reading the code base... Maybe we should always default to nonsense here? Like `["x", "y", "z"]` or something. Also to make it clear that it's not language-specific and doesn't actually matter.

Another thing to consider: if `get_examples is None`, we could use the new logger to log debugging info?
```python
util.logger.debug("No get_examples callback provided to Language.begin_training, creating dummy examples")
```

(This is a regular Python logger, so the output will only be shown if the user explicitly enables `DEBUG` logging. Added this recently to replace some of the hacky logging/intrusive warnings. We're not using it much at the moment but we can going forward.)

svlandeg: Ah, good points, thanks! (the `words` stuff is definitely not OK)

svlandeg:
There's something flaky going on with the `simple_ner` model when it's not trained, causing the test error to only occassionally pop up. Investigating.

honnibal:
> I think we should avoid adding labels or inferring other arbitrary properties in the begin_training methods, as we ran into trouble with that before, and the provided examples should really only be a small sample. They are not supposed to contain all valid labels, for instance. So instead of adding them automagically, the component now throws an error if the labels are not properly defined beforehand.

Ah I think this is not correct. The `component.begin_training` method is the right place to do stuff like gather labels from the training data, and we require that the training corpus (or a representative sample that has all the labels) be passed in during `component.begin_training`. It's then up to the components to make sure they're not passing too much data into their `model.initialize` methods.

If we don't do it this way, we then have to invent some other way to gather the labels, and we'll be back at the same sort of situation. At some point we have to let the components see the data so they can do what they need to do. The `component.begin_training` method is the place in the lifecycle where that happens.

honnibal:
> Note: I wouldn't typically work with assert... statements in the main code flow, but I think it's useful to have them in the begin_trainingmethods so the code crashes there instead of somewhere in a recursive function in Thinc. If these asserts fail, they point to a programming error, so this really shouldn't happen.

I kind of like asserts tbh, I find them helpful when I read the code. Of course we shouldn't have asserts failing, and a lot of my asserts did, which is how I think we got scarred a bit ;). But asserts themselves aren't bad.

honnibal:
> There's something flaky going on with the simple_ner model when it's not trained, causing the test error to only occassionally pop up. Investigating.

We should kill this `simple_ner` code. It's an experiment that didn't work well, there's no reason to maintain it in the main library.

honnibal:
Anyway. PR looks good to me!

ines (approved):

PR #6037, opened 2020-09-08
Add quick hack to allow CLI extensions

honnibal:
I'm not sure whether this will be the long-term plan, but I need a quick way to extend the CLI from an external package, in order to keep developing the `spacy-ray` integration.

What I've done is added a dummy registry called "cli", that just serves as a way to import modules that are trying to hook into the CLI. Within those modules, they'll be importing spaCy's CLI `app` object to extend it.

@ines is it okay to merge this until a better solution is developed, probably within typer?

## Checklist

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -13,7 +13,7 @@ from thinc.config import Config, ConfigValidationError
 from configparser import InterpolationError
 
 from ..schemas import ProjectConfigSchema, validate
-from ..util import import_file, run_command, make_tempdir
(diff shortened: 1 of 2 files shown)


ines (approved):
Yes, this looks good! I was initially thinking we could make the entry points just expose the functions and then add them to our CLI programmatically (so users don't have to import the decorator). But the disadvantage is that users then couldn't pass settings down from the decorator either, which may be inconvenient. So I think this solution is the best.

PR #6040, opened 2020-09-08
Improve v3 pretrain command

honnibal:
We intentionally let the `spacy pretrain` command fall behind our conventions a bit in v3, while we let the rest of the config system stabilise. This PR does some work to get it up to date, but there are still further improvements to do. We don't necessarily have to do them all now.

* [x] Make `spacy pretrain` load from a registry like we do in `spacy train`, instead of a text path. I've added a small `JsonlReader` class to `spacy.gold` for this purpose. I'm not sure whether we like this, especially the naming. We could also demand that we read from `.spacy`? It'd work fine I think.
* [x] Remove the ability to read data from stdin. You can use the registry to stream from a function if you want, and it stops us from having to special-case.
* [x] Use the `batchers` registry, instead of passing those as separate arguments.
* [ ] Refactor the training loop to be more consistent with `spacy train`
* [ ] Use the loggers registry, instead of the weird logging implementation that's there now.
* [ ] Use the registry to define the objective 
* [ ] Use the registry to build the pretraining model
* [ ] Support the `before_init` etc lifecycle hooks

I've also changed the way the `spacy pretrain` command accesses the sublayer to pretrain. Previously we defined this as a path within the config. However, you might not have your config exposing exactly the layer you want to pretrain. It also feels a bit unnatural.

What I've done instead is asked for the component and a reference. So a common thing would be something like `component = "ner"` and `layer = "tok2vec"`. This would fetch `nlp.get_pipe("ner").model.get_ref("tok2vec")`. You need to make sure the component's model has a named reference that matches your layer string. 

## Checklist


ines:
> We could also demand that we read from .spacy? It'd work fine I think.

That's an interesting idea... I guess reading from `Doc` objects isn't so bad because it offers a direct answer for how to provide custom tokenization? And it's consistent with the training data format so in theory, you could use data created in the same way for both. (Not sure if this makes it more confusing? And if we end up accepting `.spacy` for pretraining as well, should we warn if the `Doc` objects have annotations set, just to make it clear that we're not actually using them?)

>  Support the before_init etc lifecycle hooks

You mean `before_creation` etc. in the `[nlp]` block? That should already be taken care of by `load_model_from_config`, via `Language.from_config`.

honnibal:
> And if we end up accepting .spacy for pretraining as well, should we warn if the Doc objects have annotations set, just to make it clear that we're not actually using them?)

I don't think so, no. You could define pretraining objectives that do make use of some annotations, for instance you could be using sentence boundaries, use POS tags to choose better word substitutions to corrupt the text, etc. I guess I've talked myself into `Doc` objects being a good idea?

ines:
> You could define pretraining objectives that do make use of some annotations, for instance you could be using sentence boundaries, use POS tags to choose better word substitutions to corrupt the text, etc. I guess I've talked myself into Doc objects being a good idea?

Ah, that is pretty compelling then 🤔

honnibal:
I guess there's not really a decision to be made here. People can use a `.spacy` file with the `Corpus.v1` reader, or they can use this jsonl reader. or something else. It's up to them.

spacy/gold/corpus.py
```diff
@@ -151,3 +159,53 @@ def read_docbin(
...
+        limit: int = 0,
+        min_length: int = 0,
+        max_length: int = 0,
+    ) -> None:
+        self.path = util.ensure_path(path)
+        self.min_length = min_length
+        self.max_length = max_length
+        self.limit = limit
+
+    def __call__(self, nlp: "Language") -> Iterator[Doc]:
```
svlandeg, line 194: Can't this return `Iterator[Example]` instead? So we don't need the `ensure_docs` stuff? And then the `registry.readers` would contain readers with the same return type... (I guess I'm scared this type of "doc or example" will trickle back into the main codebase as it did before)

spacy/gold/corpus.py
```diff
@@ -21,6 +22,36 @@ def create_docbin_reader(
 ) -> Callable[["Language"], Iterable[Example]]:
     return Corpus(path, gold_preproc=gold_preproc, max_length=max_length, limit=limit)
 
+@util.registry.readers("spacy.JsonlReader.v1")
+def create_jsonl_reader(
+    path: Path, min_length: int=0, max_length: int = 0, limit: int = 0
```
svlandeg, line 27: I'm personally not a fan of using `0` for `min_length`'s default here, why not set it to some high int value so we don't need to treat it specially in the code? Or use `None` instead?

honnibal: If we use `None` we need to make it `Optional`, and then the type analysis is harder. 0 is an invalid maximum, so it makes sense to use it as the "no maximum" value. In code that needs to be C-typed, there's no alternative, it has to be 0 and cannot be `None`.

Using a "very large" value is much uglier imo, there should be a way to specify "no limit".

spacy/gold/corpus.py
honnibal: This can, yeah. I do like the idea of letting users iterate out `Doc` objects, constructing an `Example` seems sort of bureaucratic? I don't really mind though.

For now I'll just change this to produce `Example`, and I'll leave the `ensure_docs` in.

PR #6044, opened 2020-09-09
Refactor Docs.is_ flags

adrianeboyd:
## Description

* Add derived `Doc.has_annotation` method

  * `Doc.has_annotation(attr)` returns `True` for partial annotation

  * `Doc.has_annotation(attr, require_complete=True)` returns `True` for
    complete annotation

* Add deprecation warnings to `is_tagged`, `is_parsed`, `is_sentenced` and `is_nered`

* Add `Doc._get_array_attrs()`, which returns a full list of `Doc` attrs for use with `Doc.to_array`, `Doc.to_bytes` and `Doc.from_docs`. The list is the `DocBin` attributes list plus `SPACY` and `LENGTH`.

Notes on `Doc.has_annotation`:

* `HEAD` is converted to `DEP` because heads don't have an unset state

* Accept `IS_SENT_START` as a synonym of `SENT_START`

Additional changes:

* Add `NORM`, `ENT_ID` and `SENT_START` to default attributes for `DocBin`

* In `Doc.from_array()` the presence of `DEP` causes `HEAD` to override `SENT_START`

* In `Doc.from_array()` using `attrs` other than `Doc._get_array_attrs()` (i.e., a user's custom list rather than our default internal list) with both `HEAD` and `SENT_START` shows a warning that `HEAD` will override `SENT_START`

* `set_children_from_heads` does not require dependency labels to set sentence boundaries and sets `sent_start` for all non-sentence starts to `-1`

### Types of change

Enhancement

## Checklist


adrianeboyd:
Questions and concerns:

* The ever-present naming problem: what's a better name for `require_complete`?

* Why does `Doc.to_bytes` need `LENGTH`? (It breaks without it, but I'm not sure why.)

honnibal:
> The ever-present naming problem: what's a better name for require_complete?

I think `complete` is good?

> Why does Doc.to_bytes need LENGTH? (It breaks without it, but I'm not sure why.)

It saves the doc as `doc.text`, the lengths of the words, and the spaces boolean. It's not clear this is actually optimal, but I was trying to make a guess at a speed vs space vs robustness trade-off.

> Are we really sure we want to require DEP for parses? Hopefully using DEP to check both DEP and HEAD is mostly a non-issue for typical users, but it sure caused a lot of hassle in the tests that only used HEAD in a lot of cases.

Yeah I think this is fine. Changing it would introduce more problems and I don't think the alternatives are better.

honnibal:
It's very nice to consolidate these under the `has_annotation` method though. Those flags definitely sucked.

PR #6046, opened 2020-09-10
Clean up spacy.tokens

adrianeboyd:
## Description

* Update `set_children_from_heads`
  * Don't check `dep` when setting lr_* or sentence starts
  * Set all non-sentence starts to `False`

* Use `set_children_from_heads` in `Token.head` setter
  * Reduce similar/duplicate code (admittedly adds a bit of overhead)
  * Update sentence starts consistently

* Remove unused `Doc.set_parse`

* Minor changes:
  * Declare cython variables (to avoid cython warnings)
  * Clean up imports

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -868,7 +868,7 @@ cdef class Doc:
         self.is_tagged = bool(self.is_tagged or TAG in attrs or POS in attrs)
         # If document is parsed, set children
         if self.is_parsed:
-            set_children_from_heads(self.c, length)
+            set_children_from_heads(self.c, 0, length)
         return self
 
     @staticmethod
@@ -1305,12 +1305,13 @@ cdef int token_by_char(const TokenC* tokens, int length, int char_idx) except -2
             return mid
     return -1
 
-cdef int set_children_from_heads(TokenC* tokens, int length) except -1:
+cdef int set_children_from_heads(TokenC* tokens, int start, int end) except -1:
+    # note: end is exclusive
     cdef TokenC* head
     cdef TokenC* child
     cdef int i
     # Set number of left/right children to 0. We'll increment it in the loops.
-    for i in range(length):
+    for i in range(start, end):
         tokens[i].l_kids = 0
         tokens[i].r_kids = 0
         tokens[i].l_edge = i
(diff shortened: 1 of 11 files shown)


spacy/tokens/doc.pyx
```diff
@@ -1381,7 +1366,7 @@ cdef int _set_lr_kids_and_edges(TokenC* tokens, int length, int loop_count) exce
     # Get sentence start positions according to current state
     sent_starts = set()
     for i in range(length):
-        if tokens[i].head == 0 and tokens[i].dep != 0:
+        if tokens[i].head == 0:
```
honnibal, line 1369: Is this right?

spacy/tokens/doc.pyx
```diff
@@ -1362,7 +1347,7 @@ cdef int _set_lr_kids_and_edges(TokenC* tokens, int length, int loop_count) exce
     for i in range(length):
         child = &tokens[i]
         head = &tokens[i + child.head]
-        if child < head and loop_count == 0:
+        if loop_count == 0 and child < head:
```
honnibal, line 1350: Good idea reversing these conditions.

spacy/tokens/doc.pyx
adrianeboyd: This is actually a duplicate change from #6044. I had tried to replace the code in `Token.head` with `set_children_from_heads` before, but always ran into test failures and weird behavior. It had to do with gradually constructing or modifying a doc in a loop, which was done in some of the tests, like this:

```
for i in range(length):
  doc[i].dep_ = deps[i]
  doc[i].head = doc[heads[i]]
```

Depending on whether the dep or head is set first, you got different behavior in the lr_edges/kids/sent_start because of this. I really think `set_children_from_heads` should not care what the deps are. It should only care about the shape of the unlabelled graph. The additional code in `Token.head` also didn't adjust `is_sent_start` correctly, just the lr_edges/kids. I think maintaining only `set_children_from_heads` is going to be an overall improvement, even though it will be slightly slower because it goes through the whole doc.

spacy/tokens/doc.pyx
adrianeboyd: The `loop_count` code is kind of dumb, but I tried splitting into two functions instead, but it didn't seem like an improvement.

spacy/tokens/doc.pyx
honnibal: The problem is you can only really set the heads per token, so it's pretty unfortunate if looping over the document and changing the heads is `O(n**2)`.

I'll reluctantly agree to merge this because I agree that the code duplication is really bad. But is there a way we can break out the loop body of `set_children_from_heads` so that we can call it with only one token?

I'd rather you didn't spend a lot of time on this just now, so if it's tricky just leave a todo and we can come back to it.

adrianeboyd: I think it would be pretty easy to limit the range of the tokens adjusted by `Token.head`. I think you could go from the leftmost left edge of either root to the rightmost right edge of either root? Does that sound right?

honnibal: Yes that does make sense. Do you want to try that? I think it sounds like a good approach.

honnibal:
Looks good!

honnibal (approved):

PR #6048, opened 2020-09-10
Fix project clone compatibility

ines:
## Description

- [x] Add fallbacks for old versions of Git (not sure if this is right but it now works for me?)
- [x] Don't expect the whole subtree, e.g. cloning `x/y/z` will result in `z` in the current working directory
- [x] Adjust some error messages

### Types of change
ux, cli

## Checklist


ines:
I'll just go ahead and merge this and maybe adjust some error handling around it. It's not going to make things worse, I guess.

PR #6051, opened 2020-09-11
Support "list" pipeline argument

svlandeg:
## Description
The `pipeline` argument for `init config` should be a comma-seperated list like `-p "tagger, textcat"`. However, I think it is likely for people to try to write `-p "[textcat]"` (as in the config). So we may as well support this, because it gives an ugly error downstream otherwise. With this edit, we even support people writing something like `-p "["tagger", "textcat"]"`

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -336,10 +336,12 @@ def git_sparse_checkout(repo: str, subpath: str, dest: Path, *, branch: str = "m
         # Now pass those missings into another bit of git internals
         missings = " ".join([x[1:] for x in ret.stdout.split() if x.startswith("?")])
         if not missings:
-           err = f"Could not find any relevant files for '{subpath}'. " \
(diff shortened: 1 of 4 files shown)


ines:
This is a good idea and I think we should maybe have a more general-purpose helper for it? I could think of a few other places to use it (everything that takes comma-separated values).

svlandeg:
Yea, makes sense, in `cli/_util.py` ?

ines:
Yes, sounds good 👍

svlandeg:
> 
> **Edit:** Also, now that I think about it, maybe we do need more sophisticated logic here if we want to support the list syntax properly? For example, we need to strip out the quotes and handle both single and double quotes etc.?

I extended the function and wrote two unit tests, it should cover quite a lot of cases now :-)

ines:
Okay cool, this looks solid 😅💪

PR #6053, opened 2020-09-11
prevent overwriting score_weights

svlandeg:
## Description
The `score_weights` in the training's config was getting overwritten bij the `Language` class. Suppose your config would have 
```
[training.score_weights]
cats_macro_p = 0.0 
cats_macro_r = 0.0
cats_macro_f = 1.0
```
then this would get overwritten (due to the defaults) by `cats_score = 1.0`, and the console logger would only show the `cat_score` instead of the scores from the config.

I think adding this conditional is sufficient?

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -243,7 +243,8 @@ class Language:
         self._config["nlp"]["pipeline"] = list(self.component_names)
         self._config["nlp"]["disabled"] = list(self.disabled)
         self._config["components"] = pipeline
-        self._config["training"]["score_weights"] = combine_score_weights(score_weights)
+        if not self._config["training"].get("score_weights"):
+            self._config["training"]["score_weights"] = combine_score_weights(score_weights)
         if not srsly.is_json_serializable(self._config):
             raise ValueError(Errors.E961.format(config=self._config))
         return self._config


spacy/language.py
```diff
@@ -243,7 +243,8 @@ def config(self) -> Config:
         self._config["nlp"]["pipeline"] = list(self.component_names)
         self._config["nlp"]["disabled"] = list(self.disabled)
         self._config["components"] = pipeline
-        self._config["training"]["score_weights"] = combine_score_weights(score_weights)
+        if not self._config["training"].get("score_weights"):
```
ines, line 246: Alternatively, I guess we could use `setdefault` here? Then we don't need the extra `get`?

svlandeg: Erm, yea I guess we could? It feels a bit funny to use `setdefault` with an elaborate, inferred list of things that is different everytime this function is run though? Not sure it'll be as readable.

svlandeg: And come to think about it, it can't be more efficient? At some point, the internal dict needs to check whether it needs to use the default value anyway...

ines: Yeah we shouldn't worry about efficiency, it won't be significant here... sorry, I think my suggestion was kinda dumb, it really doesn't matter and it's not going to be more readable.

(Also, every time I use `setdefault` I have to double-check the syntax 😂 So that's already a bad sign.)

PR #6057, opened 2020-09-12
Fix tagger training when some tags are missing

honnibal:
Needed to specify that the missing value is `""`, or the loss function raises an error (it expects `None` by default).

## Checklist

### DIFF
--- a/spacy/pipeline/senter.pyx
+++ b/spacy/pipeline/senter.pyx
@@ -3,6 +3,7 @@ from itertools import islice
 
 import srsly
 from thinc.api import Model, SequenceCategoricalCrossentropy, Config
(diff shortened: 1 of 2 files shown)


honnibal:
I've copy/pasted the `Tagger.update` method into the `Senter` subclass, as I made a change to it that was specific to the tagger and was surprised when the subclass broke.

It can be pretty tough to have these inheritance relationships that you don't expect, and the senter class already overwrites almost everything from the tagger. I think it's better to avoid complicating things to try to reduce the code duplication. If we don't want that code to be duplicated, we should put it into a helper function and call the function from the two classes.

honnibal:
Hm I think this isn't right.

PR #6058, opened 2020-09-12
Fix conversion of older CoNLL parsing files

honnibal:
There are a billion "CoNLL" formats, depending on the tool producing
them. The Stanford v3.3 converter has a few quirks that the CoNLL-X
conversion wasn't handling:

* Sentences may have extra spacing in between the newlines
* The coarse-grained POS is the same as the fine-grained POS, so we need
  a tag map to get the coarse-grained POS.

Needing the tag map is particularly unfortunate, it feels like something
that should be patched on the source data? Adding the extra option may
be confusing to people, especially since it *overwrites* the corpus tag.

## Checklist

### DIFF
--- a/spacy/cli/convert.py
+++ b/spacy/cli/convert.py
@@ -51,6 +51,7 @@ def convert_cli(
     ner_map: Optional[Path] = Opt(None, "--ner-map", "-nm", help="NER tag mapping (as JSON-encoded dict of entity types)", exists=True),
     lang: Optional[str] = Opt(None, "--lang", "-l", help="Language (if tokenizer required)"),
     concatenate: bool = Opt(None, "--concatenate", "-C", help="Concatenate output to a single file"),
+    tag_map: Optional[Path] = Opt(None, "--tag-map", "-tm", help="Tag mapping (as JSON-encoded dict of tag types)", exists=True),
     # fmt: on
 ):
     """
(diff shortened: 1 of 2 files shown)


adrianeboyd:
The space handling seems harmless, if slightly annoying.

Where's an example of where you need this kind of tag map? It needs to be called something else (xpos map?) so it's not confused with the other tag map, I think.

In general I would be happier if this were a pure CoNLL-U converter (for files that validate with their validator) and that there were separate converters to deal with various CoNLL-X weirdness.

honnibal:
The Stanford v3.3 converter produces data like this:

```
1       No      _       RB      RB      _       7       discourse       _       _
2       ,       _       ,       ,       _       7       punct   _       _
3       it      _       PRP     PRP     _       7       nsubj   _       _
4       was     _       VBD     VBD     _       7       cop     _       _
5       n't     _       RB      RB      _       7       neg     _       _
6       Black   _       NNP     NNP     _       7       nn      _       _
7       Monday  _       NNP     NNP     _       0       root    _       _
8       .       _       .       .       _       7       punct   _       _
```

I'm looking at this now and I think I just misunderstood the data format. The extra POS column isn't a coarse-grained POS at all, if I remember right it's for gold vs predicted POS tags. So we definitely don't want this tag_map option.

PR #6064, opened 2020-09-14
 Fix sparse checkout and error handling 

ines:
## Description

- Make sparse checkout optional via `--sparse` flag (while still falling back to non-sparse for older Git versions, but with warning).
- Improve handling of subprocess error and how it's messages are exposed to the user.

### Types of change
cli, ux

## Checklist

### DIFF
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -680,7 +680,10 @@ def run_command(
             Errors.E970.format(str_command=" ".join(command), tool=command[0])
         ) from None
     except subprocess.CalledProcessError as e:
-        # We don't want a duplicate traceback here
+        # We don't want a duplicate traceback here so we're making sure the
+        # CalledProcessError isn't re-raised. We also print both the string
+        # message and the stderr, in case the error only has one of them.
+        print(e.stderr)
         print(e)
         sys.exit(1)
     if ret.returncode != 0:
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -300,7 +300,9 @@ def ensure_pathy(path):
     return Pathy(path)
 
 
-def git_sparse_checkout(repo: str, subpath: str, dest: Path, *, branch: str = "master"):
+def git_checkout(
+    repo: str, subpath: str, dest: Path, *, branch: str = "master", sparse: bool = False
+):
     git_version = get_git_version()
     if dest.exists():
         msg.fail("Destination of checkout must not exist", exits=1)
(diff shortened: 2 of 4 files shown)


spacy/util.py
```diff
@@ -680,7 +680,10 @@ def run_command(
             Errors.E970.format(str_command=" ".join(command), tool=command[0])
         ) from None
     except subprocess.CalledProcessError as e:
-        # We don't want a duplicate traceback here
+        # We don't want a duplicate traceback here so we're making sure the
+        # CalledProcessError isn't re-raised. We also print both the string
+        # message and the stderr, in case the error only has one of them.
+        print(e.stderr)
```
honnibal, line 686: I think this still might not be quite right =/

We'll end up with the `stdout` and `stderr` separately, they won't be interleaved. So in some situations the log will be really difficult to read and we'll be staring at that cursed docs page again in a few months 😰 

I think we maybe want:

```
 ret = subprocess.run(
            command,
            env=os.environ.copy(),
            input=stdin,
            encoding="utf8",
            check=False,
            stdout=subprocess.PIPE if capture else None,
            stderr=subprocess.STDOUT if capture else None,
        )
    if ret.returncode != 0:
        # TODO: Somehow print a traceback for the Python call?
        print(ret.stdout.read())
        sys.exit(ret.returncode)
```

honnibal: Ideas here:

* Preserving the exit code of the subprocess can be pretty important, especially in cloud environments, as the return code may indicate the machine being lost. We use this in our own job system, to figure out whether to retry.
* I don't think we can get the error code if we don't do `check=False`. If we let `subprocess` raise a `CalledProcessException`, we're going to exit via Python's nornal tracebacking and so we'll get exit code 1.
* I don't know how to "raise error with exit code". I couldn't figure that out, StackOverflow just led me to `sys.exit`.

In summary there's three pieces of information we want to preserve:

1. The Python traceback. We definitely need this, otherwise you don't know in what context the subprocess was called.
2. The subprocess's  stdout and stderr. Sometimes it's okay to look at only one, or to look at them separately, but in some programs they only make sense interleaved in the original order.
3. The exit code.

Of the three, obviously item 3 is the least important. And we'll need to have an option for whether to actually `sys.exit`, as sometimes we want to try/except instead.

ines: Okay, we can try this but I think this should be a separate ticket because it can easily take a while and a lot of trial and error. I'll go ahead and merge this PR to at least solve the more immediate issue and improve the error handling.

PR #6067, opened 2020-09-15
Handle config and meta correctly in spacy.blank

ines:
## Description

`spacy.blank` would silently swallow its keyword arguments and it didn't allow config overrides (because it instantiates the language class directly and we don't have config overrides on that – if you want to initialize from a config, use `Language.from_config`). Solution:

- make `spacy.blank` delegate to `.from_config`
- allow explicit keyword arguments: `config` (config overrides), `meta` (data for `nlp.meta`) and `vocab`
- allow `meta` keyword argument on `Language.from_config` for consistency

### Types of change
bug fix

## Checklist


spacy/__init__.py
```diff
@@ -46,12 +47,22 @@ def load(
     return util.load_model(name, disable=disable, exclude=exclude, config=config)
 
 
-def blank(name: str, **overrides) -> Language:
+def blank(
+    name: str,
+    *,
+    vocab: Union[Vocab, bool] = True,
```
honnibal, line 53: I do think `True` is the right default value to use here, for consistency with other places where we have used `None` or `False` to mean "not provided and also don't create it".

But what should we do if someone passes `vocab=None` or `vocab=False`? I guess we need to raise? Allowing `vocab=None` is one of those tiny steps towards the 'do what I mean' approach that gets us into trouble later.

ines: Sure, I'm not a fan of `True` but it is currently the default everywhere and how we do it in all other places.

spacy/language.py
```diff
@@ -1525,7 +1527,7 @@ def from_config(
         # inside stuff like the spacy train function. If we loaded them here,
         # then we would load them twice at runtime: once when we make from config,
         # and then again when we load from disk.
```
honnibal, line 1529:
```suggestion
        # and then again when we load from disk.
        if vocab in (False, None, 0, 1):
            raise ValueError(Errors.E1010(vocab=vocab))
```

With an error something like:

 E1010 = "Invalid value for vocab: {vocab}. Valid values are a 'Vocab' instance or 'True' to create one."

ines: This should happen in `__init__`, though, because that's where it is used

honnibal (approved):

PR #6068, opened 2020-09-15
fix W&B logger

svlandeg:
## Description
Fix W&B logger when called multiple times from same script, cf docs [here](https://docs.wandb.com/library/init#how-do-i-launch-multiple-runs-from-one-script).

### Types of change
bug fix

## Checklist


honnibal:
Ah is this why my experiments were hanging? Nice, thanks! I was about to grumble about this on Slack :p

honnibal (approved):

PR #6069, opened 2020-09-15
Add vectors option to CharacterEmbed

adrianeboyd:
## Description

Add vectors option as `also_use_static_vectors` to `CharacterEmbed`.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/ml/models/tok2vec.py
+++ b/spacy/ml/models/tok2vec.py
@@ -164,7 +164,7 @@ def MultiHashEmbed(
 
 
 @registry.architectures.register("spacy.CharacterEmbed.v1")
-def CharacterEmbed(width: int, rows: int, nM: int, nC: int):
+def CharacterEmbed(width: int, rows: int, nM: int, nC: int, also_use_static_vectors: bool):
     """Construct an embedded representation based on character embeddings, using
     a feed-forward network. A fixed number of UTF-8 byte characters are used for
     each word, taken from the beginning and end of the word equally. Padding is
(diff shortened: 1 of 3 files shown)


spacy/pipeline/morphologizer.pyx
```diff
@@ -25,6 +25,7 @@ default_model_config = """
 
 [model.tok2vec]
 @architectures = "spacy.Tok2Vec.v1"
+pretrained_vectors = null
```
honnibal, line 28: I don't think we need this line. The `pretrained_vectors` argument is passed to the `spacy.HashEmbedCNN.v1` function because that function takes all the arguments to the embedding and encoding layers together, because it builds them both. The `spacy.Tok2Vec.v1` layer takes in the embedding and encoding layers as arguments, so we don't have this config option.

spacy/ml/models/tok2vec.py
```diff
@@ -188,18 +188,35 @@ def CharacterEmbed(width: int, rows: int, nM: int, nC: int):
...
+    if also_use_static_vectors:
+        model = chain(
+            concatenate(
+                chain(_character_embed.CharacterEmbed(nM=nM, nC=nC), list2ragged()),
+                chain(
+                    FeatureExtractor([NORM]),
+                    list2ragged(),
+                    with_array(HashEmbed(nO=width, nV=rows, column=0, seed=5)),
+                ),
+                StaticVectors(width, dropout=0.0),
```
honnibal, line 203: This looks right to me. Is it working?

adrianeboyd: I haven't debugged all the way down to making sure it's really using the vectors, but it seems to work, yes.

I guess my concern was about having the option `also_use_static_vectors` directly in the config when this wasn't what it looked like in some of the other models (with `pretrained_vectors` instead), but if you think just including `also_use_static_vectors` is okay, then I guess it isn't that complicated.

One thing that seems kind of hard to do is:

```
nlp = spacy.load("model_with_vectors")
nlp.add_pipe("morphologizer", config={???})
```

Without any config, it takes the default config, which has `also_use_static_vectors` set to `false`. How do I even specify that here?

adrianeboyd: I would kind of expect this to use the vectors if they're available in the vocab, but that may be hard / undesirable? I can create a config with `init config` and then do a lot of editing to add vectors in various places, but it seems like it ought to be simpler.

honnibal: I think that sort of smartness has gotten us in trouble before, so I'd rather have it as an explicit setting instead of guessed. Usually for training you'll be using the config, and it can't be set after training. So I don't think the API issue around adding the pipe is so bad.

(If the question is "I start with a trained pipeline that has vectors, and I want to train a morphologizer for it. What do I do?", the answer would be to make a config and specify the other components as "sourced" from your previous pipeline. This is a slightly different emphasis from previously. Now we think of it as defining a new pipeline, and you're telling it where to get the components. Previously we thought of it as changing the other pipeline, which made things very tricky and difficult.)

PR #6072, opened 2020-09-15
Fix unit test with ExceptionInfo

svlandeg:
## Description
The new `test_language_init_invalid_vocab` fails for me locally (in Pycharm) as `str(e)` would be `<ExceptionInfo ValueError tblen=2>` instead of the expected string `"ValueError: [E918] Received invalid value for vocab..."`.
Rewriting it to explicitely check `e.value`, which is the original `ValueError`.

I have no clue why this fails locally and not on the CI, but I think this fix should be universal.

### Types of change
bug fix

## Checklist


ines:
Ah, sorry! Python is weird 🤪

PR #6074, opened 2020-09-16
Remove unneeded methods in senter and morphologizer

adrianeboyd:
## Description

Now that the tagger doesn't manage the tag map, the child classes senter and morphologizer don't need to override the serialization methods.

### Types of change

Enhancement?

## Checklist


adrianeboyd:
(Not as much to simplify as I thought.)

PR #6083, opened 2020-09-17
Fix/span.sent

tamuhey:
## Description

`span.sent` should raise error if `not doc.is_sentencized`, but it doesn't and returns `None`.
This is because `span.sent` runs `doc.sents` to check if the doc is sentencized.
Since `doc.sents` is a generator it doesn't raise any error immediately.

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -326,7 +326,9 @@ def test_span_boundaries(doc):
         _ = span[5]
 
 
-def test_sent(doc):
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Thanks for the PR! Let's see if the CI problems (not related to your patch) are straightened out.

And looking at the details, I'm not sure this does quite the right thing around user hooks for `Doc.sents`, but that's separate from this particular issue and I can have another look after this is merged.

PR #6084, opened 2020-09-17
add pretraining option to init config

svlandeg:
## Description
Right now, if you want a pretraining config, you need to run both `init config` and `init fill-config --pretraining` (the latter was undocumented btw).
But in general,  I expected `init config` to return a fully operational config, so it feels like we should add that `pretraining` option to that command as well? 

If the generated config gets a `transformer` instead of a `tok2vec`, an informative warning is given, so I don't think that should be a problem (and the result is the same as running the two commands separately otherwise)

### Types of change
enhancement

## Checklist


website/docs/api/cli.md
```diff
@@ -124,15 +124,16 @@ customize those settings in your config file later.
...
+| `--lang`, `-l`         | Optional code of the [language](/usage/models#languages) to use. Defaults to `"en"`. ~~str (option)~~                                                                                                                                                                                                                              |
+| `--pipeline`, `-p`     | Comma-separated list of trainable [pipeline components](/usage/processing-pipelines#built-in) to include. Defaults to `"tagger,parser,ner"`. ~~str (option)~~                                                                                                                                                                      |
+| `--pretraining`, `-pt` | Include config for pretraining (with 'spacy pretrain'). Default False. ~~bool~~                                                                                                                                                                                                                                                    |
```
ines, line 132:
```suggestion
| `--pretraining`, `-pt` | Include config for pretraining (with [`spacy pretrain`](/api/cli#pretrain)). Defaults to `False`. ~~bool (flag)~~                                                                                                                                                                                                                                                    |
```

website/docs/api/cli.md
```diff
@@ -160,13 +161,14 @@ validation error with more details.
...
-| `--help`, `-h` | Show help message and available arguments. ~~bool (flag)~~                                                                          |
-| **CREATES**    | Complete and auto-filled config file for training.                                                                                  |
+| Name                  | Description                                                                                                                         |
+| --------------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
+| `base_path`           | Path to base config to fill, e.g. generated by the [quickstart widget](/usage/training#quickstart). ~~Path (positional)~~           |
+| `output_file`         | Path to output `.cfg` file. If not set, the config is written to stdout so you can pipe it forward to a file. ~~Path (positional)~~ |
+| `--pretraining`, `-p` | Include config for pretraining (with 'spacy pretrain'). Default False. ~~bool~~                                                     |
```
ines, line 168:
```suggestion
| `--pretraining`, `-pt` | Include config for pretraining (with [`spacy pretrain`](/api/cli#pretrain)). Defaults to `False`. ~~bool (flag)~~                                                                                                                                                                                                                                                    |
```

svlandeg: it's actually `-p` for the `fill-config` and `-pt` for `init config` because the latter already had a `-p` for the pipeline. Do we want to use `-p` for both?

PR #6089, opened 2020-09-17
Modify setting missing and blocked entity tokens

adrianeboyd:
## Description

In order to make it easier to construct `Doc` objects as training data, modify how missing and blocked entity tokens are set to prioritize setting `O` for training purposes over setting blocked and missing entity tokens.

* `Doc.ents` setter sets tokens outside entity spans to `O` regardless of the current state of each token

* `Doc.set_ents(entities, blocked=, missing=, outside=, default="outside")` can be used to set complicated entity information

### Types of change

Enhancement?

## Checklist


adrianeboyd:
I know we hadn't agreed on this, but I wanted to see what the most minimal changes I could manage would look like here, since it really needs to be easier to set `O`.

honnibal:
I like the `Doc.block_ents`. But instead of:

> For Doc.ents, setting a span with a missing label sets the ent_iob to missing instead of blocked

What if this instead set (soft) `O`? Effectively this would mean that passing spans without a label did nothing significant, it would be the same as filtering them out.

We could then add a method `Doc.remove_ents` that would take an optional argument for whether to mark the token as (soft) `O` or simply missing.

honnibal:
Actually what if we simply had:

* Doc.ents setter sets tokens outside entity spans to O regardless of the current state of each token
* For Doc.ents, setting a span with a missing label sets the ent_iob to 2 (i.e. 'soft' O)
* We add a `Doc.set_ents` method, with options like:

```python
def set_ents(
    self,
    entities,
    *,
    blocked: List[Span]=[],
    missing: List[Union[Span, int, Token]]=[],
    outside: List[Union[Span, int, Token]]=[],
    default="outside"
):
    """Set named entity spans on the document, and mark whether non-entity
    spans should be treated as:
    * blocked (definitely not an entity),
    * missing (entity status unknown), or
    * outside (tagged 'O').

    Marking tokens as 'blocked' will prevent spaCy's built-in NER component from
    predicting them as entities, although other NER components may not behave the same way.
    """
```

adrianeboyd:
> Actually what if we simply had:
> 
>     * Doc.ents setter sets tokens outside entity spans to O regardless of the current state of each token
> 
>     * For Doc.ents, setting a span with a missing label sets the ent_iob to 2 (i.e. 'soft' O)

Then you may as well just ignore spans with missing labels, or potentially forbid missing labels so that there's no confusion with the new behavior.
 
>     * We add a `Doc.set_ents` method, with options like:
> 
> 
> ```python
> def set_ents(
>     self,
>     entities,
>     *,
>     blocked: List[Span]=[],
>     missing: List[Union[Span, int, Token]]=[],
>     outside: List[Union[Span, int, Token]]=[],
>     default="outside"
> ):
>     """Set named entity spans on the document, and mark whether non-entity
>     spans should be treated as:
>     * blocked (definitely not an entity),
>     * missing (entity status unknown), or
>     * outside (tagged 'O').
> 
>     Marking tokens as 'blocked' will prevent spaCy's built-in NER component from
>     predicting them as entities, although other NER components may not behave the same way.
>     """
> ```

I wish it were slightly easier to create labelled spans, but even so, I think this would be much cleaner if we only allowed spans, but didn't require them to be sorted like in `doc.ents`. (Actually, maybe we could remove that restriction, too?) Creating unlabelled spans is easy at least, so I'd rather not allow `int` and `Token` alongside the existing tuple mess. You could potentially allow tuples for the non-entity types, too.

```python
def set_ents(
    self,
    entities: List[Union[Span, Tuple]],
    *,
    blocked: List[Span]=[],
    missing: List[Span]=[],
    outside: List[Span]=[],
    default="outside"
):
```

And maybe `"default"` should have an option not to modify anything else so that you can add spans to an existing set more easily? But this may be overloading things a bit.

honnibal:
Yeah I was reluctant to add that `Union`, let's just not. `List[Span]` only, you're right.

The `Tuple` format is the worst, I really don't want to have more API accepting it. Let's just make people make spans. The problem with the tuples is there's 8 combinations of what it could be: is the label first or last, and are the indices characters or tokens, and does the end index exclusive or inclusive?

> And maybe "default" should have an option not to modify anything else so that you can add spans to an existing set more easily? But this may be overloading things a bit.

I think it'd be good to have that option in the method too, I was thinking that. What would you call it? I could see either `default=None` or `default='unchanged'`

adrianeboyd:
The main thing I'm not sure about:

I've forbidden empty labels when setting `doc.ents` so that there's no confusion in terms of blocked vs. missing entities. This however requires special treatment for the output of `spans_from_biluo_tags`, which currently returns missing entities as spans with empty labels, see: https://github.com/explosion/spaCy/pull/6089/files#diff-227463ef7a7a823172cc31d73702676eR307

spacy/tokens/doc.pyx
```diff
@@ -705,6 +716,108 @@ cdef class Doc:
...
+        if blocked is None:
+            blocked = tuple()
+        if missing is None:
+            missing = tuple()
+        if outside is None:
+            outside = tuple()
+
+        # Find all tokens covered by spans and check that none are overlapping
+        seen_tokens = set()
+        for span in entities:
```
honnibal, line 745: Can't we just do:

```
for span in (entities + blocked + missing + outside)
```

here, and avoid the repetition?

adrianeboyd: Yeah, that would better. (I was originally going to try to figure out which two overlapped for the error message, but decided it was too slow.)

honnibal:
> I've forbidden empty labels when setting doc.ents

I think we should just ignore them, rather than raising.

honnibal (approved):

PR #6091, opened 2020-09-18
Introducing the gpu_allocator

svlandeg:
Requires Thinc 8.0.0a34: https://github.com/explosion/thinc/pull/392

## Description
* Rename `use_pytorch_for_gpu_memory` (bool) to `gpu_allocator` (str)
* Refer to it from the `train` section, it can just be `None` by default, which doesn't do anything
* Updating the docs + some fixes

### Types of change
enhancement

## Checklist


spacy/cli/_util.py
```diff
@@ -439,3 +440,12 @@ def string_to_list(value: str, intify: bool = False) -> Union[List[str], List[in
...
     return result
+
+
+def set_gpu_allocator(allocator):
+    if allocator == "pytorch":
+        use_pytorch_for_gpu_memory()
+    elif allocator == "tensorflow":
+        use_tensorflow_for_gpu_memory()
+    else:
+        msg.warn(f"Ignored unknown 'gpu_allocator' argument: '{allocator}'")
```
svlandeg, line 451: should this be here, or in the main `util` class? Perhaps users will want to call this from within their own code?

honnibal: Should it be in Thinc instead?

spacy/default_config.cfg
```diff
@@ -6,7 +6,7 @@ init_tok2vec = null
 
 [system]
 seed = 0
-use_pytorch_for_gpu_memory = false
+gpu_allocator = ""
```
ines, line 9: Do we want the default to be `""` here? I know it's technically less consistent but `null` would just feel more natural?

PR #6092, opened 2020-09-18
Load vocab lookups tables at beginning of training

adrianeboyd:
## Description

Similar to how vectors are handled, move the vocab lookups to be loaded at the start of training rather than when the vocab is initialized, since the vocab doesn't have access to the full config when it's created.

The option moves from `nlp.load_vocab_data` to `training.lookups`.

Typically these tables will come from `spacy-lookups-data`, but any `Lookups` object can be provided.

The loading from `spacy-lookups-data` is now strict, so configs for each language should specify the exact tables required. This also makes it easier to control whether the larger clusters and probs tables are included.

To load `lexeme_norm` from `spacy-lookups-data`:

```
[training.lookups]
@misc = "spacy.LookupsDataLoader.v1"
lang = ${nlp.lang}
tables = ["lexeme_norm"]
```

### Types of change

?

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -31,6 +31,7 @@ from .schemas import ConfigSchema
 from .git_info import GIT_VERSION
 from . import util
 from . import about
+from .lookups import load_lookups
 
 
 # This is the base config will all settings (training etc.)
@@ -84,6 +85,12 @@ def create_tokenizer() -> Callable[["Language"], Tokenizer]:
         )
 
     return tokenizer_factory
+
+
+@registry.misc("spacy.LoadLookupsData.v1")
+def load_lookups_data(lang, tables):
+    lookups = load_lookups(lang=lang, tables=tables)
+    return lookups
 
 
 class Language:
@@ -152,7 +159,6 @@ class Language:
                 self.lang,
                 self.Defaults,
                 vectors_name=vectors_name,
-                load_data=self._config["nlp"]["load_vocab_data"],
             )
         else:
             if (self.lang and vocab.lang) and (self.lang != vocab.lang):
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -87,8 +87,9 @@ def create_tokenizer() -> Callable[["Language"], Tokenizer]:
     return tokenizer_factory
 
 
-@registry.misc("spacy.LoadLookupsData.v1")
+@registry.misc("spacy.LookupsDataLoader.v1")
 def load_lookups_data(lang, tables):
+    util.logger.debug(f"Loading lookups from spacy-lookups-data: {tables}")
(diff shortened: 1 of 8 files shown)


spacy/vocab.pyx
```diff
@@ -424,6 +417,14 @@ cdef class Vocab:
             orth = self.strings.add(orth)
         return orth in self.vectors
 
+    def load_lookups(self, lookups):
```
ines, line 420: I guess technically, this should be called `add_lookups`? Because it doesn't really load anxthing and takes an already loaded `Lookups` object?

spacy/language.py
```diff
@@ -86,6 +87,12 @@ def tokenizer_factory(nlp: "Language") -> Tokenizer:
     return tokenizer_factory
 
 
+@registry.misc("spacy.LoadLookupsData.v1")
+def load_lookups_data(lang, tables):
+    lookups = load_lookups(lang=lang, tables=tables)
```
ines, line 92: We could consider adding debug logging here (via `util.logger`) that mentions the tables that are loaded and that they're loaded from `spacy-lookups-data`. If the user has verbose logging enabled during training, they'll see this info (and it can potentially explain why loading takes a bit longer etc.)

ines (commented):
Yes, IMO this is a good compromise 👍

Some naming suggestions:

* `spacy.LoadLookupsData.v1` &rarr; `spacy.LookupsDataLoader.v1` (?)
* `load_vocab_data_into_model` &rarr; `add_lookups_to_vocab` (?)

I don't feel super strongly about any of them, though, and naming is hard.

spacy/vocab.pyx
adrianeboyd: I guess `set_lookups` then. Or we could make `lookups` a property. The `setter` is still weird because you layer the partial `NORM` calls and you can't really undo it, just add on top.

spacy/language.py
adrianeboyd: Sounds good, I'll add that.

adrianeboyd:
> Yes, IMO this is a good compromise +1
> 
> Some naming suggestions:
> 
>     * `spacy.LoadLookupsData.v1` → `spacy.LookupsDataLoader.v1` (?)

That's fine.

>     * `load_vocab_data_into_model` → `add_lookups_to_vocab` (?)

The idea was to make it flexible so you could add more kwargs in the future with additional stuff beyond `lookups`, but I'm not sure what that might be.

ines:
> The idea was to make it flexible so you could add more kwargs in the future with additional stuff beyond lookups, but I'm not sure what that might be.

Ah okay, fair point. It's mostly internals anyway at this point.

I like the property solution for the lookups btw, this is even better 👍

PR #6096, opened 2020-09-21
Allow config overrides from environment variables

ines:
## Description

- Env variables take precedence over CLI overrides (and values defined in the config).
- Allows overriding config from the `project` CLI.
- Will hopefully make it easier for us to set up automated CI tests for the project templates because we can force-override the number of epochs/steps and actually train.

The `SPACY_CONFIG_OVERRIDES` env variable can be set to a string and currently support the same format as the CLI, e.g.: `SPACY_CONFIG_OVERRIDES="--training.batch_size 128 --paths.train=/some_path"`. This way, we can re-use a lot of the parsing logic we already have (and you'll be able to easily copy-paste your overrides around without having to modify them).

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -7,6 +7,7 @@ import srsly
 import hashlib
 import typer
 from click import NoSuchOption
+from click.parser import split_arg_string
 from typer.main import get_command
 from contextlib import contextmanager
 from thinc.config import Config, ConfigValidationError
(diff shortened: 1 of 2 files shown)


honnibal:
> Currently, the env variable syntax uses a prefix SPACY_CONFIG_ and two underscores __ as a separator (e.g. to represent the .). Happy to change this if we find a better solution. Example usage: SPACY_CONFIG_TRAINING__BATCH_SIZE=128.

I think we should consider just making it a single string of overrides. Otherwise people will have to do this translation step of munging their config names into the environment variables, and we'll have to undo that logic when we process. So I think a single `SPACY_CONFIG_OVERRIDES` is better.

honnibal:
Ah yeah I had wondered whether the `--` style was good. It's a good idea and keeps the implementation simple.

PR #6098, opened 2020-09-21
Extend Doc.__init__ with additional annotation

adrianeboyd:
## Description

Mostly copying from `spacy.tests.util.get_doc`, add additional kwargs to `Doc.__init__` to initialize the most common doc/token values.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
This doesn't do `sent_starts`. It wouldn't be too hard to add if desired.

spacy/tokens/doc.pyx
```diff
@@ -158,17 +158,38 @@ cdef class Doc:
             raise ValueError(Errors.E046.format(name=name))
         return Underscore.doc_extensions.pop(name)
 
-    def __init__(self, Vocab vocab, words=None, spaces=None, user_data=None):
+    def __init__(
+        self,
+        Vocab vocab,
+        words=None,
+        spaces=None,
+        user_data=None,
```
ines, line 166: I wonder if we could make the `user_data` keyword-only 🤔 I doubt it's used much in user code and we haven't really been using it either. So I think it'd be safe?

ines (approved):

spacy/tokens/doc.pyx
adrianeboyd: I didn't want to change it, but it wasn't even in the docs.

ines: If it was undocumented, then let's just go or it and make it keyword-only!

honnibal:
I think the `sent_starts` is probably a good addition, yeah. Otherwise, looks good!

PR #6102, opened 2020-09-21
Update morph_rules.py

Nuccy90:
Added "dig" and "dej" ("you" in accusative form)

## Description
I noticed that 2nd person accusative forms of "you" were missing and I added them. Of course I don't know if there was a good historical reason for omitting them, it seemed like an easy fix.

### Types of change
Bug fix (possibly).

## Checklist

### DIFF
--- a/.github/contributors/Nuccy90.md
+++ b/.github/contributors/Nuccy90.md
@@ -1,3 +1,51 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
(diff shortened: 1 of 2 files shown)


.github/contributors/Nuccy90.md
```diff
@@ -0,0 +1,58 @@
+    consent of, pay or render an accounting to the other for any use or
```
svlandeg, line 1: It looks like this contributor argument is not entirely complete?

Nuccy90:
Ooops! Fixed it now.

adrianeboyd:
Thanks for the PR! (As a note, we're phasing out the tag maps and morph rules in the library for spacy v3, so this will just be in the upcoming patch releases for v2.3.)

PR #6103, opened 2020-09-21
Tidy up tests and docs

ines:
## Description

- Replace `get_doc` helper with `Doc.__init__`
- Tidy up test suite
- Allow string label in `ents` tuples
- Make sure our test suite doesn't raise our own warnings
- Update docs

### Types of change
tests, docs

## Checklist


spacy/tokens/doc.pyx
```diff
@@ -1596,4 +1596,6 @@ def get_entity_info(ent_info):
         ent_type, ent_kb_id, start, end = ent_info
     else:
         ent_id, ent_kb_id, ent_type, start, end = ent_info
+    if isinstance(ent_type, str):
+        ent_type = vocab.strings.add(ent_type)
```
ines, line 1600: @honnibal Is this okay? It feels like there should be a more elegant way but I'm not sure what that would be.

honnibal: You should probably use `hash_string` here and take care of the side-effect of adding to the string store in the calling function.

spacy/tests/README.md
```diff
@@ -41,17 +40,16 @@ py.test spacy/tests/tokenizer/test_exceptions.py::test_tokenizer_handles_emoji #
...
+- Very **extensive tests** that take a long time to run should be marked with `@pytest.mark.slow`. If your slow test is testing important behaviour, consider adding an additional simpler version.
+- If tests require **loading the models**, they should be added to the [`spacy-models`](https://github.com/explosion/spacy-models) tests.
+- Before requiring the models, always make sure there is no other way to test the particular behaviour. In a lot of cases, it's sufficient to simply create a `Doc` object manually. See the section on [helpers and utility functions](#helpers-and-utilities) for more info on this.
+- **Avoid unnecessary imports.** There should never be a need to explicitly import spaCy at the top of a file, and many components are available as [fixtures](#fixtures). You should also avoid wildcard imports (`from module import *`).
+- If you're importing from spaCy, **always use absolute imports**. For example: `from spacy.language import Language`.
+- Don't forget the **unicode declarations** at the top of each file. This way, unicode strings won't have to be prefixed with `u`.
```
svlandeg, line 51: we can remove this with python 3 :-)

spacy/tests/README.md
```diff
@@ -107,59 +104,32 @@ If all tests in a file require a specific configuration, or use the same complex
 
 Our new test setup comes with a few handy utility functions that can be imported from [`util.py`](util.py).
 
+### Constructing a `Doc` object manually with
```
svlandeg, line 107:
```suggestion
### Constructing a `Doc` object manually
````

PR #6106, opened 2020-09-22
extend quickstart components

svlandeg:
## Description
- add `textcat` component: `TextCatEnsemble` if optimizing for accuracy, `TextCatBOW` otherwise.
- add `entity_linker` component so it can share the `Transformer` / `Tok2Vec` layer with the NEL
- didn't add the Morphologizer & senter because I'm not sure it makes sense for them to share the `tok2vec`: the Morphologizer uses a `CharacterEmbed` by default and the `senter` a `tok2vec` layer with a much smaller width. We could add them here with those defaults as such, but it wouldn't make a difference (they can get added through their default factories anyway)

### Types of change
enhancement
 
## Checklist

### DIFF
--- a/spacy/cli/templates/quickstart_training.jinja
+++ b/spacy/cli/templates/quickstart_training.jinja
@@ -93,6 +93,22 @@ grad_factor = 1.0
 @layers = "reduce_mean.v1"
 {% endif -%}
 
+{% if "entity_linker" in components -%}
+[components.entity_linker]
+factory = "entity_linker"
+get_candidates = {"@misc":"spacy.CandidateGenerator.v1"}
+incl_context = true
+incl_prior = true
+
+[components.entity_linker.model]
+@architectures = "spacy.EntityLinker.v1"
+nO = null
+
+[components.entity_linker.model.tok2vec]
+@architectures = "spacy-transformers.TransformerListener.v1"
+grad_factor = 1.0
+{% endif -%}
+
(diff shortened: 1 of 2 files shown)


ines:
Yay, looks good! 🎉

> We could add them here with those defaults as such, but it wouldn't make a difference (they can get added through their default factories anyway)

Yeah, I agree, we should only have settings/blocks in the template if we actually want to customise the settings for different scenarios. For all other cases the default way the components are added is fine.

Btw, we still need to adjust the `[training.score_weights]` block!

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -93,6 +93,46 @@ grad_factor = 1.0
...
+incl_context = true
+incl_prior = true
+
+[components.entity_linker.model]
+@architectures = "spacy.EntityLinker.v1"
+nO = null
+
+[components.entity_linker.model.tok2vec]
+@architectures = "spacy-transformers.TransformerListener.v1"
+grad_factor = 1.0
```
honnibal, line 109: This is probably missing the pooling:

```suggestion
grad_factor = 1.0

[components.parser.model.tok2vec.pooling]
@layers = "reduce_mean.v1"
{%- endif %}
```

(Reminder: The pooling accounts for multi-alignment between the word-pieces and the spaCy tokens. It says "take the average of the wordpiece tokens".)

PR #6110, opened 2020-09-22
Validate section refs in debug config

ines:
## Description

Add logic to `debug config` to check that certain sections we know contain references to other config sections are valid and exist. Unfortunately, we have to kinda hard-code this and can't do it as part of the regular config validation process, since it needs to run on the resolved config, in case users have a `[corpora]` function that returns a dict etc.

But I made sure to set up the function and logic so it's easily extensible in the future.

### Types of change
enhancement, ux

## Checklist


svlandeg (approved):
Looks good to me!

PR #6123, opened 2020-09-23
avoid logging performance of frozen components

svlandeg:
## Description
By default, the logger would try to log all pipe names, even if they were marked as "frozen". That would result in a downstream error because the training process wouldn't have the loss of those non-trained components. Fixing that here.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -152,7 +152,7 @@ def train(
         exclude=frozen_components,
     )
     msg.info(f"Training. Initial learn rate: {optimizer.learn_rate}")
-    with nlp.select_pipes(disable=[*frozen_components]):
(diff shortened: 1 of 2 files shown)


spacy/cli/train.py
```diff
@@ -152,7 +152,8 @@ def train(
         exclude=frozen_components,
     )
     msg.info(f"Training. Initial learn rate: {optimizer.learn_rate}")
-    print_row, finalize_logger = train_logger(nlp)
+    with nlp.select_pipes(disable=[*frozen_components]):
```
ines, line 155: Is there a specific reason we can't just do `disable=frozen_components`?

PR #6124, opened 2020-09-23
Update Indonesian Example Phrases  

rasyidf:
The current example of the Indonesian language contains sensitive topics (Violent, Terrorism) and needs to be changed with more proper sentences.
 
## Description
No code changed, only strings. 

### Types of change
bugfix or enhancement

## Checklist


adrianeboyd:
Thanks for the PR! We definitely agree that it's better to have more neutral examples.

PR #6125, opened 2020-09-23
Add option to disable Matcher errors

adrianeboyd:
## Description

So that the `AttributeRuler` component runs and doesn't produce cryptic errors when some pipeline components are disabled:

* Add option to disable Matcher errors when a doc doesn't contain a particular type of annotation

Minor additional change:

* Update `AttributeRuler.load_from_morph_rules` to allow direct `MORPH`
values

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -195,7 +195,7 @@ cdef class Matcher:
                 else:
                     yield doc
 
-    def __call__(self, object doclike, *, as_spans=False):
+    def __call__(self, object doclike, *, as_spans=False, suppress_errors=False):
         """Find all token sequences matching the supplied pattern.
 
         doclike (Doc or Span): The document to match over.
@@ -215,16 +215,17 @@ cdef class Matcher:
         else:
             raise ValueError(Errors.E195.format(good="Doc or Span", got=type(doclike).__name__))
         cdef Pool tmp_pool = Pool()
-        if TAG in self._seen_attrs and not doc.has_annotation("TAG"):
-            raise ValueError(Errors.E155.format(pipe="tagger", attr="TAG"))
-        if POS in self._seen_attrs and not doc.has_annotation("POS"):
-            raise ValueError(Errors.E155.format(pipe="morphologizer", attr="POS"))
-        if MORPH in self._seen_attrs and not doc.has_annotation("MORPH"):
-            raise ValueError(Errors.E155.format(pipe="morphologizer", attr="MORPH"))
-        if LEMMA in self._seen_attrs and not doc.has_annotation("LEMMA"):
-            raise ValueError(Errors.E155.format(pipe="lemmatizer", attr="LEMMA"))
-        if DEP in self._seen_attrs and not doc.has_annotation("DEP"):
-            raise ValueError(Errors.E156.format())
+        if not suppress_errors:
+            if TAG in self._seen_attrs and not doc.has_annotation("TAG"):
+                raise ValueError(Errors.E155.format(pipe="tagger", attr="TAG"))
+            if POS in self._seen_attrs and not doc.has_annotation("POS"):
+                raise ValueError(Errors.E155.format(pipe="morphologizer", attr="POS"))
+            if MORPH in self._seen_attrs and not doc.has_annotation("MORPH"):
+                raise ValueError(Errors.E155.format(pipe="morphologizer", attr="MORPH"))
+            if LEMMA in self._seen_attrs and not doc.has_annotation("LEMMA"):
+                raise ValueError(Errors.E155.format(pipe="lemmatizer", attr="LEMMA"))
+            if DEP in self._seen_attrs and not doc.has_annotation("DEP"):
+                raise ValueError(Errors.E156.format())
         matches = find_matches(&self.patterns[0], self.patterns.size(), doclike, length,
                                 extensions=self._extensions, predicates=self._extra_predicates)
         final_matches = []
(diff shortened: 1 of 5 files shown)


adrianeboyd:
I could totally imagine a different setting (a class setting?) or something here. I'm not particularly picky about the details, I just want the functionality because otherwise there are going to be headaches with the default pipelines and the attribute ruler.

honnibal:
I think it makes sense to have a pipeline component that can match on more patterns than are set at one particular time. Like, you should be able to use the same patterns, and see how much more precise the matcher becomes when you have POS tags available vs when you don't. So I support the idea behind this.

`suppress_errors` sounds sort of ominous (and vague) though. How about `allow_missing`?

adrianeboyd:
`missing` is even more vague? I suppose there's no harm in a long option like `allow_missing_annotation`? But that's not even really accurate, since the docs only have to be partially annotated. Hmm...

adrianeboyd:
`skip_annotation_checks`?

honnibal:
I think we should just go with `allow_missing`. I think it's clear and memorable enough, it's not like either documents, patterns or anything else could be "missing". The only thing that can be missing are the values for the attributes. The problem I had with `suppress_errors` is that the user is wondering what type of errors? All errors?

spacy/matcher/matcher.pyx
```diff
@@ -195,7 +195,7 @@ cdef class Matcher:
                 else:
                     yield doc
 
-    def __call__(self, object doclike, *, as_spans=False):
+    def __call__(self, object doclike, *, as_spans=False, suppress_errors=False):
```
honnibal, line 198:
```suggestion
    def __call__(self, object doclike, *, as_spans=False, allow_missing=False):
```

spacy/matcher/matcher.pyx
```diff
@@ -215,16 +215,17 @@ cdef class Matcher:
...
-            raise ValueError(Errors.E155.format(pipe="tagger", attr="TAG"))
-        if POS in self._seen_attrs and not doc.has_annotation("POS"):
-            raise ValueError(Errors.E155.format(pipe="morphologizer", attr="POS"))
-        if MORPH in self._seen_attrs and not doc.has_annotation("MORPH"):
-            raise ValueError(Errors.E155.format(pipe="morphologizer", attr="MORPH"))
-        if LEMMA in self._seen_attrs and not doc.has_annotation("LEMMA"):
-            raise ValueError(Errors.E155.format(pipe="lemmatizer", attr="LEMMA"))
-        if DEP in self._seen_attrs and not doc.has_annotation("DEP"):
-            raise ValueError(Errors.E156.format())
+        if not suppress_errors:
```
honnibal, line 218:
```suggestion
        if not allow_missing:
```

spacy/pipeline/attributeruler.py
```diff
@@ -79,7 +79,7 @@ def __call__(self, doc: Doc) -> Doc:
 
         DOCS: https://nightly.spacy.io/api/attributeruler#call
         """
-        matches = sorted(self.matcher(doc))
+        matches = sorted(self.matcher(doc, suppress_errors=True))
```
honnibal, line 82:
```suggestion
        matches = sorted(self.matcher(doc, allow_missing=True))
```

spacy/tests/matcher/test_matcher_api.py
```diff
@@ -316,6 +316,9 @@ def test_attr_pipeline_checks(en_vocab):
         matcher(doc2)
     with pytest.raises(ValueError):
         matcher(doc3)
+    # errors can be suppressed if desired
+    matcher(doc2, suppress_errors=True)
```
honnibal, line 320:
```suggestion
    matcher(doc2, allow_missing=True)
```

spacy/tests/matcher/test_matcher_api.py
```diff
@@ -316,6 +316,9 @@ def test_attr_pipeline_checks(en_vocab):
...
+    matcher(doc3, suppress_errors=True)
```
honnibal, line 321:
```suggestion
    matcher(doc3, allow_missing=True)
```

spacy/matcher/matcher.pyx
```diff
@@ -215,16 +215,17 @@ cdef class Matcher:
...
+            if TAG in self._seen_attrs and not doc.has_annotation("TAG"):
+                raise ValueError(Errors.E155.format(pipe="tagger", attr="TAG"))
+            if POS in self._seen_attrs and not doc.has_annotation("POS"):
+                raise ValueError(Errors.E155.format(pipe="morphologizer", attr="POS"))
+            if MORPH in self._seen_attrs and not doc.has_annotation("MORPH"):
+                raise ValueError(Errors.E155.format(pipe="morphologizer", attr="MORPH"))
+            if LEMMA in self._seen_attrs and not doc.has_annotation("LEMMA"):
+                raise ValueError(Errors.E155.format(pipe="lemmatizer", attr="LEMMA"))
+            if DEP in self._seen_attrs and not doc.has_annotation("DEP"):
+                raise ValueError(Errors.E156.format())
```
honnibal, line 228: We should probably make this a loop:

```
for attr in [TAG, POS, MORPH, LEMMA, DEP]:
    if attr in self._seen_attrs and not doc.has_annotation(attr):
        raise ValueError(Errors.E155.format(pipe="morphologizer", attr=self.strings[attr]))
```

adrianeboyd: Yeah. (This is from when the is_flags weren't easily derivable from the features and they were the same for MORPH/POS.)

PR #6128, opened 2020-09-23
replace 'nr_feature_tokens'

svlandeg:
## Description
Replace `nr_feature_tokens` (int) with `state_type` (str: "parser" or "ner") and `extra_state_tokens` (bool)

### Types of change
refactor

## Checklist


spacy/ml/models/parser.py
```diff
@@ -12,7 +12,8 @@
 @registry.architectures.register("spacy.TransitionBasedParser.v1")
 def build_tb_parser_model(
     tok2vec: Model[List[Doc], List[Floats2d]],
-    nr_feature_tokens: Literal[3, 6, 8, 13],
+    state_type: str,
```
ines, line 15: Now that we have the `Literal` type set up, we could consider making this `Literal["parser", "ner"]` 🤔

PR #6134, opened 2020-09-24
Add [training.before_to_disk] callback

ines:
## Description

Optional function called on the `nlp` object before it's saved to disk during/after training and can return a modified version. Can be used to implement all kinds of things, including cleaning up the config.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -480,6 +480,9 @@ class Errors:
     E201 = ("Span index out of range.")
 
     # TODO: fix numbering after merging develop into master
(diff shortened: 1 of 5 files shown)


spacy/errors.py
```diff
@@ -480,6 +480,9 @@ class Errors:
     E201 = ("Span index out of range.")
 
     # TODO: fix numbering after merging develop into master
+    E914 = ("Executing {name} callback failed. Expected the function to "
+            "returnthe nlp object but got: {value}. Maybe you forgot to return "
```
svlandeg, line 484: typo "returnthe"

PR #6135, opened 2020-09-24
Proofread documentation

walterhenry:
Proofread some API docs

## Description
Fixed some typos

### Types of change
Docs

## Checklist


website/docs/api/doc.md
```diff
@@ -445,7 +445,7 @@ Mark a span for merging. The `attrs` will be applied to the resulting token (if
 they're context-dependent token attributes like `LEMMA` or `DEP`) or to the
 underlying lexeme (if they're context-independent lexical attributes like
 `LOWER` or `IS_STOP`). Writable custom extension attributes can be provided as a
-dictionary mapping attribute names to values as the `"_"` key.
+dictionary mapping attribute name to values as the `"_"` key.
```
svlandeg, line 448: I think "names" was correct: the dictionary maps attribute names to values. This sentence parses quite badly though, so perhaps we need to rewrite it into something like 
"... can be provided using the `"_"` key and specifying a dictionary that maps attribute name to values"

website/docs/api/pipeline-functions.md
```diff
@@ -26,7 +26,7 @@ Merge noun chunks into a single token. Also available via the string name
 
 <Infobox variant="warning">
 
-Since noun chunks require part-of-speech tags and the dependency parse, make
+Since noun chunks require part-of-speech tags and the dependency parser, make
```
svlandeg, line 29: I think the original was fine, because it refers to requiring the actual parse (which is produced by the parser), just like the first bit of the sentence says it requires the tags (output of the tagger).

svlandeg: (perhaps "a" is more clear than "the", though?) --> it requires a dependency parse ?

svlandeg:
Nice work, that must have involved days of reading! 😁

website/docs/api/pipeline-functions.md
walterhenry: Yeah, I wasn't quite sure on that one. I had never seen "parse" alone before.

ines:
Okay, resolved the conflict (which was caused by a small change I made today) 🎉

website/docs/api/doc.md
```diff
@@ -444,8 +444,7 @@ invalidated, although they may accidentally continue to work.
 Mark a span for merging. The `attrs` will be applied to the resulting token (if
 they're context-dependent token attributes like `LEMMA` or `DEP`) or to the
 underlying lexeme (if they're context-independent lexical attributes like
-`LOWER` or `IS_STOP`). Writable custom extension attributes can be provided as a
-dictionary mapping attribute name to values as the `"_"` key.
+`LOWER` or `IS_STOP`). Writable custom extension attributes can be provided using the `"_"` key and specifying a dictionary that maps attribute name to values.
```
ines, line 482:
```suggestion
`LOWER` or `IS_STOP`). Writable custom extension attributes can be provided using the `"_"` key and specifying a dictionary that maps attribute names to values.
```

svlandeg: ugh, my bad, even though I commented on the "name", I wrote it wrongly in the suggestion myself :|

ines: Ha, no worries!

PR #6137, opened 2020-09-24
Fix skipped documents in entity scorer

honnibal:
In our OntoNotes evaluation, we're working off documents that have some sentences that don't have NER annotations, because we've merged the annotations from the parse and name sections of the corpus onto one doc, so that we can train them together. The missing annotations are tricky to deal with during scoring.

Previously we were skipping documents that had missing data for the NER. This is fine when comparing spaCy models to other spaCy models, as they're all running over the same test set, but it means our test was slightly different from what other people were running, which makes it a bit harder to directly compare against other people's OntoNotes numbers. Our comparison is already a bit indirect anyway, because we do the task "whole document" without the sentence boundaries. But still, it's good to get this fixed, especially now that our alignment API is easier to work with.

After the fix, our NER scores for the CNN go down by about 0.5 on the OntoNotes 5 evaluation.

## Checklist

### DIFF
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -28,6 +28,7 @@ class PRFScore:
         self.tp += other.tp
         self.fp += other.fp
         self.fn += other.fn
+        return self
 
     def __add__(self, other):
         return PRFScore(
@@ -642,15 +643,15 @@ def get_ner_prf(examples: Iterable[Example]) -> Dict[str, PRFScore]:
                 scores[pred_ent.label_] = PRFScore()
             indices = align_x2y[pred_ent.start : pred_ent.end].dataXd.ravel()
             if len(indices):
-                g_span = eg.y[indices[0] : indices[-1]]
+                g_span = eg.y[indices[0] : indices[-1] + 1]
                 # Check we aren't missing annotation on this span. If so,
                 # our prediction is neither right nor wrong, we just
                 # ignore it.
                 if all(token.ent_iob != 0 for token in g_span):
-                    key = (pred_ent.label_, indices[0], indices[-1])
+                    key = (pred_ent.label_, indices[0], indices[-1] + 1)
                     if key in golds:
                         scores[pred_ent.label_].tp += 1
-                        golds.pop(span)
+                        golds.remove(key)
                     else:
                         scores[pred_ent.label_].fp += 1
         for label, start, end in golds:
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -295,20 +295,18 @@ class Scorer:
             # Find all predidate labels, for all and per type
             gold_spans = set()
             pred_spans = set()
-            # Special case for ents:
-            # If we have missing values in the gold, we can't easily tell
-            # whether our NER predictions are true.
-            # It seems bad but it's what we've always done.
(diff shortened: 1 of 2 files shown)


spacy/scorer.py
```diff
@@ -295,20 +295,23 @@ def score_spans(
...
-                pred_per_type[span.label_].add((span.label_, span.start, span.end - 1))
+            align_x2y = example.alignment.x2y
+            for pred_span in getter(pred_doc, attr):
+                indices = align_x2y[pred_span.start : pred_span.end].dataXd.ravel()
+                if len(indices):
+                    g_span = gold_doc[indices[0] : indices[-1]]
+                    # Check we aren't missing annotation on this span. If so,
+                    # our prediction is neither right nor wrong, we just
+                    # ignore it.
+                    if any(token.ent_iob == 0 for token in g_span):
```
adrianeboyd, line 311: Is this backwards?

honnibal: It's what I intended to write (although maybe my thinking is wrong!)

My idea is that we fetch the gold span, and if we don't know any of its annotations, we should skip this `pred_span` and not evaluate it.

This means we'll always by necessity be checking the accuracy of all the gold spans. If we don't recover any of them, we'll lose accuracy for that. But we don't lose precision for predicting spans that map onto tokens that have missing annotations.

honnibal: Oh wait, yes it is! Thanks.

honnibal, line 311:
```suggestion
                    if all(token.ent_iob != 0 for token in g_span):
```

spacy/scorer.py
```diff
@@ -295,20 +295,23 @@ def score_spans(
...
-                pred_per_type[span.label_].add((span.label_, span.start, span.end - 1))
+            align_x2y = example.alignment.x2y
+            for pred_span in getter(pred_doc, attr):
+                indices = align_x2y[pred_span.start : pred_span.end].dataXd.ravel()
+                if len(indices):
+                    g_span = gold_doc[indices[0] : indices[-1]]
+                    # Check we aren't missing annotation on this span. If so,
+                    # our prediction is neither right nor wrong, we just
+                    # ignore it.
+                    if all(token.ent_iob != 0 for token in g_span):
```
adrianeboyd, line 311: This is ents-specific scoring in a generic method, so it's breaking the `sents` scoring. I'm not sure whether you just want to create a separate `score_ents` method or work it in here. If neither `ents` or `sents` work with this as-is it's kind of bad, so maybe there's a way to redesign it so it works more easily. I do really want to be able to point to a generic scoring method for this kind of scoring rather than have too many special cases that make it hard to understand what the scorer is actually counting.

honnibal: Ah I forgot that this is a generic `score_spans`. Yeah I think we have different instincts in these scoring functions. I think making the code too generic here is very limiting. I suppose I'll just make my own entity scorer.

eyalho:
I think there is a small bug over the `__add__` method:
`def __add__(self, other):
        return PRFScore(
            tp=self.tp + other.tp, fp=self.fp + other.fp, fn=self.fn + other.fn
        )`
should be changed to 
`
  def __add__(self, other):
        return PRFScore() + other
`

because of 
`def __init__(self) -> None:`

PR #6154, opened 2020-09-27
Update ChineseTokenizer

adrianeboyd:
## Description

* Allow `pkuseg_model` to be set to `None` on initialization
* Force convert pkuseg_model to use pickle protocol 4 by reencoding with `pickle5` on serialization
* Update pkuseg serialization test

### Types of change

Bug fix, mostly

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 SHELL := /bin/bash
 
 ifndef SPACY_EXTRAS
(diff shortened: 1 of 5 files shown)


adrianeboyd:
This still has the problem that the tokenizer settings are saved in `tokenizer` and present in the original config, so it's not quite right.

PR #6160, opened 2020-09-28
Refactor training/config CLI and data initialization

ines:
## Description

- [x] move training utilities and functions to `training.loop`, `training.pretrain` and `training.initialize` and separate them from the CLI
- [x] add `[initialize]` block for paths and everything else that's only needed during initialization (not loaded at runtime)
- [x] resolve individual config sections with schema instead of whole config
- [x] initialize pipeline in `train` and re-use initialized `nlp` object if available (still needs to be extended with proper caching)
- [x] `Language.initialize`, `Pipe.initialize`
- [x] add `Tokenizer.initialize`
- [x] decide how to pass config from the `[initialize]` sub-blocks to the components and how to validate it
- [x] remove `sgd` argument from `Pipe.initialize`
- [x] pass `nlp` forward instead of just `nlp.pipeline`

### Todos

- [ ] better mechanism to decide whether to re-initialize or not
- [ ] docs

## Checklist


spacy/cli/_util.py
```diff
@@ -458,3 +446,23 @@ def string_to_list(value: str, intify: bool = False) -> Union[List[str], List[in
...
+
+class CliLogger:
+    """Helper mocking up the most commonly used logger methods. Can be passed
+    into functions like train() to make them output pretty-printed messages
+    on the CLI and regular logging if used from within Python.
+    """
+
+    debug = msg.text
+    info = msg.info
+    warn = msg.info
```
svlandeg, line 459:
```suggestion
    warn = msg.warn
````

File: spacy/cli/init_pipeline.py [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+    # fmt: on
+):
+    util.logger.setLevel(logging.DEBUG if verbose else logging.ERROR)
+    overrides = parse_config_overrides(ctx.args)
+    import_code(code_path)
+    setup_gpu(use_gpu)
+    with show_validation_error(config_path):
+        config = util.load_config(config_path, overrides=overrides)
+    with show_validation_error(hint_fill=False):
+        nlp = init_nlp(config, use_gpu=use_gpu, logger=CliLogger, on_succcess=msg.good)
```
svlandeg, line 35: typo "on_succcess"

spacy/training/initialize.py
```diff
@@ -0,0 +1,205 @@
...
+    nlp: Language, pretrain_config: Dict[str, Any], vocab_config: Dict[str, Any]
+) -> bool:
+    # Load pretrained tok2vec weights - cf. CLI command 'pretrain'
+    P = pretrain_config
+    V = vocab_config
+    weights_data = None
+    init_tok2vec = ensure_path(V["init_tok2vec"])
+    if init_tok2vec is not None:
+        if P["objective"].get("type") == "vectors" and not V["vectors"]:
+            err = 'need initialize.vectors if pretraining.objective.type is "vectors"'
```
svlandeg, line 140: `initialize.vocab.vectors` instead of `initialize.vectors`

spacy/training/initialize.py
```diff
@@ -0,0 +1,205 @@
...
+            nlp.resume_training(sgd=optimizer)
+    with nlp.select_pipes(disable=[*frozen_components, *resume_components]):
+        nlp.begin_training(lambda: train_corpus(nlp), sgd=optimizer)
+        on_success(f"Initialized pipeline components")
+    # Verify the config after calling 'begin_training' to ensure labels
+    # are properly initialized
+    verify_config(nlp)
+    if "pretraining" in config and config["pretraining"]:
+        P = registry.resolve(config["pretraining"], schema=ConfigSchemaPretrain)
+        loaded = add_tok2vec_weights(nlp, P, I)
```
svlandeg, line 61: The definition of  `add_tok2vec_weights` seems to expect `vocab_config`, which should be `V` instead of `I` in this call, I think?

spacy/training/initialize.py
svlandeg: (also needs to be updated in the `errors` dict below this line)

spacy/training/initialize.py
```diff
@@ -0,0 +1,205 @@
...
+    weights_data = None
+    init_tok2vec = ensure_path(V["init_tok2vec"])
+    if init_tok2vec is not None:
+        if P["objective"].get("type") == "vectors" and not V["vectors"]:
+            err = 'need initialize.vectors if pretraining.objective.type is "vectors"'
+            errors = [{"loc": ["initialize", "vectors"], "msg": err}]
+            raise ConfigValidationError(config=nlp.config, errors=errors)
+        if not init_tok2vec.exists():
+            err = f"can't find pretrained tok2vec: {init_tok2vec}"
+            errors = [{"loc": ["initialize", "vectors", "init_tok2vec"], "msg": err}]
```
svlandeg, line 145:
```suggestion
            errors = [{"loc": ["initialize", "vocab", "init_tok2vec"], "msg": err}]
```

spacy/pipeline/tagger.pyx
```diff
@@ -90,6 +90,16 @@ class Tagger(Pipe):
         """
         return tuple(self.cfg["labels"])
 
+    @property
+    def label_data(self):
+        """Data about the labels currently added to the component.
+
+        RETURNS (Dict): The labels data.
+
+        DOCS: https://nightly.spacy.io/api/tagger#labels
```
svlandeg, line 99: The links to the DOCS refers to `labels` instead of `label_data` (also for some other components)

website/docs/api/dependencyparser.md
```diff
@@ -151,11 +151,17 @@ validating the network,
...
+This method was previously called `begin_training`.
+
+</Infobox>
+
 > #### Example
 >
 > ```python
 > parser = nlp.add_pipe("parser")
-> optimizer = parser.begin_training(lambda: [], pipeline=nlp.pipeline)
+> optimizer = parser.initialize(lambda: [], pipeline=nlp.pipeline)
```
svlandeg, line 164:
```suggestion
> optimizer = parser.initialize(lambda: [], nlp=nlp)
```

spacy/pipeline/dep_parser.pyx
```diff
@@ -132,7 +132,7 @@ cdef class DependencyParser(Parser):
             labeller.model.set_dim("nO", len(self.labels))
             if labeller.model.has_ref("output_layer"):
                 labeller.model.get_ref("output_layer").set_dim("nO", len(self.labels))
-            labeller.begin_training(get_examples, pipeline=pipeline, sgd=sgd)
+            labeller.initialize(get_examples, pipeline=pipeline)
```
svlandeg, line 135: I guess `init_multitask_objectives` should get `nlp` as input so it can call `labeller.initialize(get_examples, nlp=nlp)`

website/docs/api/entitylinker.md
```diff
@@ -150,11 +150,17 @@ validating the network,
...
+This method was previously called `begin_training`.
+
+</Infobox>
+
 > #### Example
 >
 > ```python
 > entity_linker = nlp.add_pipe("entity_linker", last=True)
-> optimizer = entity_linker.begin_training(lambda: [], pipeline=nlp.pipeline)
+> optimizer = entity_linker.initialize(lambda: [], pipeline=nlp.pipeline)
```
svlandeg, line 163:
```suggestion
> optimizer = entity_linker.initialize(lambda: [], nlp=nlp)
```

spacy/pipeline/dep_parser.pyx
svlandeg: (same comment for ner.pyx)

website/docs/api/entitylinker.md
svlandeg: + similar changes for the other components in the API docs

spacy/cli/debug_model.py
```diff
@@ -77,7 +77,7 @@ def debug_model_cli(
             exits=1,
         )
     model = pipe.model
-    debug_model(C, nlp, model, print_settings=print_settings)
+    debug_model(T, nlp, model, print_settings=print_settings)
```
svlandeg, line 80: I think this edit needs to be reverted: `debug_model` assumes it gets the full config, and it needs it, because the training corpus can be defined outside the `training` block (looked up with `dot_to_object`)

spacy/training/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+    # Load pretrained tok2vec weights - cf. CLI command 'pretrain'
+    P = pretrain_config
+    I = init_config
+    weights_data = None
+    init_tok2vec = ensure_path(I["init_tok2vec"])
+    if init_tok2vec is not None:
+        if P["objective"].get("type") == "vectors" and not I["vectors"]:
+            err = 'need initialize.vocab.vectors if pretraining.objective.type is "vectors"'
+            errors = [{"loc": ["initialize", "vocab"], "msg": err}]
+            raise ConfigValidationError(config=nlp.config, errors=errors)
```
svlandeg, line 135: I think this has been refactored since my last comment, but it still seems out-of-sync?
"vectors" is now directly under "initialize" so the reference to "vocab" can be removed, both in `err` as well as in `errors`.

spacy/training/initialize.py
```diff
@@ -0,0 +1,296 @@
...
+        if not init_tok2vec.exists():
+            err = f"can't find pretrained tok2vec: {init_tok2vec}"
+            errors = [{"loc": ["initialize", "vocab", "init_tok2vec"], "msg": err}]
+            raise ConfigValidationError(config=nlp.config, errors=errors)
```
svlandeg, line 139: same comment: remove "vocab" from the path in `errors`

spacy/training/loop.py
```diff
@@ -0,0 +1,306 @@
...
+    """Train a pipeline.
+
+    nlp (Language): The initialized nlp object with the full config.
+    output_path (Path): Optional output path to save trained model to.
+    use_gpu (int): Whether to train on GPU. Make sure to call require_gpu
+        before calling this function.
+    logger (Callable[[Any], Any]): Optional logger exposing the methods info,
+        error, debug and  warn. Defaults to regular spaCy logger but can be
+        swapped for CLI logger.
+    RETURNS (Path / None): The path to the final exported model.
```
svlandeg, line 35: docstring is out-of-date (`logger` vs `silent`)

spacy/cli/debug_model.py
ines: Ahh, we should use the new `resolve_dot_names` helper here, which will walk through the config and resolve only the relevant referenced sections. See `loop.train` for an example.

(With the new Thinc, we're now able to resolve individual sections more effectively so we only want to be creating what we actually need.)

website/docs/api/transformer.md
```diff
@@ -158,7 +158,7 @@ applied to the `Doc` in order. Both [`__call__`](/api/transformer#call) and
 | `batch_size`   | The number of documents to buffer. Defaults to `128`. ~~int~~ |
 | **YIELDS**     | The processed documents in order. ~~Doc~~                     |
 
-## Transformer.begin_training {#begin_training tag="method"}
+## Transformer.initialize {#initialize tag="method"}
```
svlandeg, line 161: We need to make a corresponding update to https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/pipeline_component.py

spacy/cli/debug_model.py
svlandeg: Sure, but that doesn't change the fact that `debug_model` needs access to `C`, and not just the `T` section, right?

ines: Yeah, the `resolve_dot_names` would just receive the full (unresolved) config and would then take care of that itself, depending on which sections are needed and where the corpus is defined.

spacy/training/initialize.py
ines: Yes, this is a change I just made recently. The `[initialize.vocab]` section was kinda redundant and we might as well define those "general" settings under the top-level `[initialize]`

spacy/cli/debug_model.py
ines: Also just noticed there's something we need to fix in the `resolve_dot_names`, so I'm doing that now

website/docs/api/transformer.md
ines: ✅ https://github.com/explosion/spacy-transformers/commit/f61302482e9d580c442e9622f76ca373cac2cb43

PR #6162, opened 2020-09-29
Fix equivalance of unset and empty MorphAnalysis

adrianeboyd:
## Description

Make `token.morph` `MorphAnalysis` equivalent for `token.c.morph == 0` and `token.morph_ == Morphology.EMPTY_MORPH`.

Additional minor change:

* Revert change to `Morphology.EMPTY_MORPH` as a symbol, just use the string with a note.

## Checklist

### DIFF
--- a/spacy/morphology.pyx
+++ b/spacy/morphology.pyx
@@ -29,8 +29,8 @@ cdef class Morphology:
     FEATURE_SEP = "|"
     FIELD_SEP = "="
     VALUE_SEP = ","
-    # not an empty string so that the PreshMap key is not 0
-    EMPTY_MORPH = symbols.NAMES[symbols._]
+    EMPTY_MORPH = "_" # should be defined as a symbol so it's not missing from
+                      # an empty stringstore
 
     def __init__(self, StringStore strings):
         self.mem = Pool()
--- a/spacy/tests/doc/test_morphanalysis.py
+++ b/spacy/tests/doc/test_morphanalysis.py
@@ -94,3 +94,25 @@ def test_morph_property(tokenizer):
     tokenizer.vocab.strings.add("Feat=Val")
     doc[0].morph = tokenizer.vocab.strings.add("Feat=Val")
     assert doc[0].morph_ == "Feat=Val"
+
+
+def test_unset_equals_empty(en_tokenizer):
+    # one unset, one set to empty
+    doc1 = en_tokenizer("a b")
+    doc1[1].morph_ = ""
+
+    # both set to empty
+    doc2 = en_tokenizer("a b")
+    doc2[1].morph_ = ""
+    doc2[1].morph_ = ""
+
+    assert [t.morph for t in doc1] == [t.morph for t in doc2]
+    assert [t.morph_ for t in doc1] == [t.morph_ for t in doc2]
+
+    # both set to empty in different versions
+    doc2 = en_tokenizer("a b")
+    doc2[1].morph_ = "_"
+    doc2[1].morph_ = {}
+
+    assert [t.morph for t in doc1] == [t.morph for t in doc2]
+    assert [t.morph_ for t in doc1] == [t.morph_ for t in doc2]
--- a/spacy/tokens/morphanalysis.pyx
+++ b/spacy/tokens/morphanalysis.pyx
@@ -24,6 +24,8 @@ cdef class MorphAnalysis:
         """Create a morphological analysis from a given ID."""
         cdef MorphAnalysis morph = MorphAnalysis.__new__(MorphAnalysis, vocab)
         morph.vocab = vocab
+        if key == 0:
+            key = vocab.strings[Morphology.EMPTY_MORPH]
         morph.key = key
         analysis = <const MorphAnalysisC*>vocab.morphology.tags.get(key)
         if analysis is not NULL:


honnibal:
Hm, is there no way to make it work with just 0 and `""`? The problem is that the string encoding of `_` will be non-zero, which seems like a recipe for problems down the road.

adrianeboyd:
The original reason for it is that the hash is the key for `PreshMap` underneath and `0` is reserved.

honnibal:
I thought we'd fixed that in `PreshMap` though. 0 should work, should I take a look and see if there's a fix I need to make in `PreshMap`?

adrianeboyd:
Hmm, it does look `0` as a value is supported now. And `0` as a key should be less of an issue, I guess?

adrianeboyd:
Hmm, I remember that using `0` didn't work at some point, but it seems to be okay now. The value is pretty easy to swap out, but what ends up being more of a problem then is that you can't distinguish unset from empty value for `token.c.morph` underneath. Like `""` isn't a valid `TAG`, but it is a valid `MORPH` and then `Doc.has_annotation` doesn't work like I'd want it to.

honnibal:
Okay that makes sense. So we have `token.c.morph_ == ""` indicating that the morphology is unset, and `token.c.morph_ == "_"` indicating that it's set but the tagset is empty?

adrianeboyd:
It's a little frustrating that it's hard to tell what's going on in python given that you can't really inspect `token.c.morph`, but I think it's better to have `MorphAnalysis` be identical for both unset and empty because otherwise `token.morph` can look different when `token.morph_` doesn't, and I don't think I want different values for `token.morph_` for unset and empty.

I still need to figure out how deal with missing annotation in the scorer. I think you might have use `Token.get_struct_attr` or `doc.to_array` for `MORPH` instead of `getattr` to deal with it.

honnibal:
> I don't think I want different values for token.morph_ for unset and empty.

We're treating `token.c.morph` as a string hash, which means there's really no way we can have `token.c.morph` be different values while `token.morph_` returns the same value. So `token.morph_` will have to be `_` for the empty one and `""` for unset.

adrianeboyd:
We're not treating `token.c.morph` as a string hash, though. It's officially something to look up in the morphology table to get a `MorphAnalysis`. The value happens to correspond to the string hash for all values except empty, where both 0 and hash-of-underscore are resolved to the same `MorphAnalysis` as hash-of-underscore.

The `MorphAnalysis` could potentially indicate whether it's empty or unset (but not use this value in `__hash__`), but maybe the string doesn't? It's all kind unsatisfying, though...

adrianeboyd:
I think that having `"_"` be empty `token.morph_` is annoying, even more if you have to deal with both `""` and `"_"`.  If you decide to process the string yourself rather than going through `MorphAnalysis` it forces all kinds of annoying if/else logic that just `""` doesn't.

Now I think I'd expose `token.c.morph` as an attribute of `MorphAnalysis` that doesn't factor into `str`, `repr`, `__hash__`, or `__eq__`. Or does that sound too confusing?

honnibal:
> The value happens to correspond to the string hash for all values except empty, where both 0 and hash-of-underscore are resolved to the same MorphAnalysis as hash-of-underscore.

I really can't live with that exception, it's just too inconsistent with how everything else works. If something is going through the stringstore, it has to reverse uniquely.

If you don't want as much exception logic, perhaps you could make the morphology string always have a leading or trailing character? For instance if there's always a trailing `"_"` the empty set becomes `"_"`. You'll have to return the string with the trailing character from `token.morph_` though, because I want `strings[token.morph] == token.morph_`.

adrianeboyd:
What would you have `token.morph_` return instead?

honnibal:
I'd remove it, and the user would interact with the morphanalysis via the `token.morph` object. So you might write something like `str(token.morph)` or `token.morph.to_str()`.

adrianeboyd:
Having `token.morph_` return `str(token.morph)` was meant to be a convenience, but I see what you mean. (I also continually want to type things like `token.text_`. The meaning of underscore requires a bit too much insider knowledge.)

adrianeboyd:
And then empty and unset `MorphAnalysis` are equivalent or not?

Hmm, and the other reason to have `token.morph_` is to be able to set it from FEATS format instead of assigning to `token.morph`, which would require you to construct the `MorphAnalysis` object, which is a pretty big annoyance.

adrianeboyd:
I don't want `token.morph = string` if `token.morph` isn't a string.

honnibal:
The underscore isn't a self-describing convention, but that's why we have to be extremely disciplined about what it means, so that it always behaves the same across the whole library. It works if it's learned exactly once. `token.text` actually is an exception, and I think it's worth it -- but you're right that even that much of an exception chips away a little at the convention.

honnibal:
> Hmm, and the other reason to have token.morph_ is to be able to set it from FEATS format instead of assigning to token.morp

We should learn from the `doc.ents.__set__` problem here, and have `token.set_morph()`.

adrianeboyd:
So right now you can do

```python
token.morph = nlp.vocab.strings[feats_string]
```

which is probably bad. But constructing the `MorphAnalysis` objects is also a total pain.

adrianeboyd:
This is so different from `token.head` or anything else, though.

honnibal:
There is no `token.head.__set__` though.

If we have to resolve consistency, I would want to be deprecating `__set__` methods instead of adding new ones

PR #6167, opened 2020-09-29
small UX fix for DocBin

svlandeg:
## Description
Custom error message when a user serializes a `DocBin` with `store_user_data=False` but tries to read it back in with `store_user_data=True`

### Types of change
UX

## Checklist


honnibal:
I think we should just remove the option from the reader? If it's there you get it, if it's not you don't.

svlandeg:
I think you're right we shouldn't throw an error in the original case I detailed (writer `False`, reader `True`), it's just not there and `doc.user_data` can remain at its default values. I commited the code as such (sorry for messing up the commit message)

There's one doubtful case still, which is when the writer is `True` and the reader is `False`. Currently, the user data wouldn't be read and would remain the default values, as `from_bytes` (L206) has a safety check:  
```
        if self.store_user_data and "user_data" in msg:
            self.user_data = list(msg["user_data"])
```

However I think we can't remove this "safety check" and just load `if "user_data" in msg`. Because other methods like `add` and `merge` also only write to `self.user_data` if `self.store_user_data` is `True`. So I think we'd need to change those too to make sure the internal lists remain consistent and of equal length. And that would mean accumulating all the user data internally, and only at `to_disk` or `to_bytes` decide whether we actually want to write it. That may cause some overhead we don't want though.

svlandeg:
Ok, rewrote this so that `store_user_data` is only used at the point when writing the docs to file/bytes. This means that `get_docs` will return the data, if it was available, no matter the `store_user_data` setting.

PR #6170, opened 2020-10-01
Add FeatureExtractor from Thinc

svlandeg:
## Description 
As this layer is spaCy-specific, it makes more sense to have it here instead of in Thinc.

### Types of change
enhancement

## Checklist

### DIFF
--- a/website/docs/api/architectures.md
+++ b/website/docs/api/architectures.md
@@ -298,7 +298,7 @@ on [static vectors](/usage/embeddings-transformers#static-vectors) for details.
 > ```ini
 > [model]
 > @architectures = "spacy.FeatureExtractor.v1"
-> columns = [NORM, PREFIX, SUFFIX, SHAPE, ORTH]
+> columns = ["NORM", "PREFIX", "SUFFIX", "SHAPE", "ORTH"]
 > ```
 
 Extract arrays of input features from [`Doc`](/api/doc) objects. Expects a list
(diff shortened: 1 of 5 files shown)


website/docs/api/architectures.md
```diff
@@ -291,6 +291,24 @@ on [static vectors](/usage/embeddings-transformers#static-vectors) for details.
...
 | **CREATES** | The model using the architecture. ~~Model[List[Doc], Ragged]~~                                                                                                                                                          |
 
+### spacy.FeatureExtractor.v1 {#FeatureExtractor}
+
+> #### Example config
+>
+> ```ini
+> [model]
+> @architectures = "spacy.FeatureExtractor.v1"
+> columns = [NORM, PREFIX, SUFFIX, SHAPE, ORTH]
```
ines, line 301:
```suggestion
> columns = ["NORM", "PREFIX", "SUFFIX", "SHAPE", "ORTH"]
```

website/docs/api/architectures.md
```diff
@@ -144,9 +144,9 @@ argument that connects to the shared `tok2vec` component in the pipeline.
 Construct an embedding layer that separately embeds a number of lexical
 attributes using hash embedding, concatenates the results, and passes it through
 a feed-forward subnetwork to build mixed representations. The features used are
-the `NORM`, `PREFIX`, `SUFFIX` and `SHAPE`, which can have varying definitions
-depending on the `Vocab` of the `Doc` object passed in. Vectors from pretrained
-static vectors can also be incorporated into the concatenated representation.
+the `NORM`, `PREFIX`, `SUFFIX` and `SHAPE`, and they are extracted with a
+[FeatureExtractor](#FeatureExtractor) layer. Vectors from pretrained static
```
ines, line 148:
```suggestion
[FeatureExtractor](/api/architectures#FeatureExtractor) layer. Vectors from pretrained static
```

This is a nitpick but that little icon that's shown next to the architectures is picked based on the URL so we sometimes have to use a full link, even if it's on the same page so it looks nicer.

PR #6175, opened 2020-10-01
Refactor Token morph setting

adrianeboyd:
## Description

* Remove `Token.morph_`
* Add `Token.set_morph()`
  * `0` resets `token.c.morph` to unset
  * Any other values are passed to `Morphology.add`

### Types of change

?

## Checklist


honnibal:
Looks good. Should we somehow support passing features as a sequence, instead of a single string?

adrianeboyd:
It supports `str` and a `dict` of features. I'd add a `token.morph` setter than can set from `MorphAnalysis`, too.

PR #6181, opened 2020-10-02
add EntityRuler patterns to config

svlandeg:
## Description
Allow specifying a `patterns_path` argument for the `entity_ruler` that points to a `jsonl` file with the patterns:

```
[paths]
patterns = null

[components.entity_ruler]
factory = "entity_ruler"
ent_id_sep = "||"
overwrite_ents = false
phrase_matcher_attr = null
validate = false
patterns_path = ${paths.patterns}
```
and then run  something like 
```
"python -m spacy train ... --paths.patterns patterns.jsonl"
```

### Types of change
enhancement

## Checklist


ines:
Ah, this should go in `[initialize.components.entity_ruler]` – otherwise, we'll end up with a config that depends on a random local path at runtime (which is also the reason we didn't have an argument like this before).

The loading should probably work the way it does for the orth variants: https://github.com/explosion/spaCy/pull/6178

svlandeg:
Ugh, too many conflicts and messy commits, will start over.

PR #6184, opened 2020-10-02
Improve control of training progress and logging

honnibal:
* Call logger on each step
* Move progress bar under logging
* Expose `progress_bar` configuration option on `spacy.ConsoleLogger.v1` function
* Instead of `silent` boolean flag, allow functions to take the `wasabi.Printer` instance
* Rename `model-final` directory to `model-last`
* Refactor the `model-last` try/except logic slightly.

Overall this should give code more control of the output without changing the API. I think easily being able to disable or control the progress bar is especially good, in case loggers want to provide their own.


spacy/training/initialize.py
```diff
@@ -95,7 +95,7 @@ def load_vectors_into_model(
     """Load word vectors from an installed model or path into a model instance."""
     try:
         vectors_nlp = load_model(name)
-    except ConfigValidationError as e:
+    except ConfigValidationError:
```
ines, line 98: Oh, the real bug here was that `ConfigValidationError.from_error` should take `e` as the first argument

spacy/training/loggers.py
```diff
@@ -1,18 +1,22 @@
...
 from ..errors import Errors
-from wasabi import msg
 
 
 @registry.loggers("spacy.ConsoleLogger.v1")
-def console_logger():
+def console_logger(progress_bar: bool=False):
     def setup_printer(
         nlp: "Language",
+        printer: Optional[wasabi.Printer]=wasabi.Printer()
```
ines, line 14: Just a stylistic preference but I would probably do `from wasabi import Printer` etc.

honnibal: I just figured it'd be nice if people are looking at the signature line

honnibal: It's kind of tricky to use a type from some other library like this, so I leaned towards explicitness

ines: Now that I'm looking at the implementation again, I'm not sure I like having the `wasabi` printer detail leak into the API like this. I previously had it like this in an earlier version and then opted for making the _behaviour_ more explicit with `silent`, because that's what we're actually trying to solve here: we need a way to tell the function whether it's running in terminal/"pretty" context or not.

For a user implementing a custom logger, having the `Printer` object passed in like this from the CLI has absolutely no significance. It also doesn't help with defining custom behaviours depending on the CLI vs. non-CLI context.

ines (changes requested):
Now that I'm looking at the implementation again, I'm not sure I like having the `wasabi` printer detail leak into the API like this. I previously had it like this in an earlier version and then opted for making the _behaviour_ more explicit with `silent`, because that's what we're actually trying to solve here: we need a way to tell the function whether it's running in terminal/"pretty" context or not.

For a user implementing a custom logger, having the `Printer` object passed in like this from the CLI has absolutely no significance. It also doesn't help with defining custom behaviours depending on the CLI vs. non-CLI context.

honnibal:
> Now that I'm looking at the implementation again, I'm not sure I like having the wasabi printer detail leak into the API like this. I previously had it like this in an earlier version and then opted for making the behaviour more explicit with silent, because that's what we're actually trying to solve here: we need a way to tell the function whether it's running in terminal/"pretty" context or not.

I think it's important that users have the ability to pass an object in. It's not going to matter often, but the people who want to like, write remote logs to a database or something, should have the opportunity to do that. And they can't do it if it's a simple binary flag, unless they go and monkey-patch the object.

Wasabi actually offers a very nice opportunity here, because we can pass the instance forward, so that the user's custom object can be transmitted from one function to another. 

> For a user implementing a custom logger, having the Printer object passed in like this from the CLI has absolutely no significance. It also doesn't help with defining custom behaviours depending on the CLI vs. non-CLI context.

I'm really just thinking of the `spacy.training.loop.train` function. I want to be careful that this function is appropriately flexible for usage within the library, and we can't anticipate what people might need. For instance, a user might want to call this function in a number of processes, and reconcile the log lines.

ines:
But the problem is that this comes straight from the `train` CLI and if the function is executed in that context, it will always receive the `Printer`, even though it's completely useless to people. Any logger will have to work this way, because the signature of the function it returns needs to match.

But it does not help you with writing to a remote database or pass in any object – this is all going to happen via your custom registered logger function anyways (e.g. you'd define arguments for it in your config). That's also the part where our API needs to be modular.

If we want `spacy.training.loop.train` to be flexible within the library, it shouldn't take this arbitrary object. It should take a flag that the CLI will toggle. Otherwise, how would your custom logger know whether it should pretty-print if you don't want to use `wasabi`? Currently, the only way to know is to have to check `printer.no_print`.

> The API is quite simple and the source is easy to read, so creating a duck-typable instance will be very easy for the users who want this.

No, I disagree. I don't think a user should be implementing a custom class with a method `good`.

honnibal:
> But the problem is that this comes straight from the train CLI

But someone might call `spacy.training.loop.train` _not_ through the CLI. Then they can control the printing with the object.

> But it does not help you with writing to a remote database or pass in any object – this is all going to happen via your custom registered logger function anyways (e.g. you'd define arguments for it in your config).

I'm passing the `printer` instance forward into the logger callback --- that's what I mean by it being nice that it's an object.

And yeah these things can be done in the config, but if you're calling the function from Python, it's nice to be able to redirect or intercept the log lines without modifying the `nlp` object.

> No, I disagree. I don't think a user should be implementing a custom class with a method good.

Why not? Anyone who notices or cares about this just wants to be able to control it. It's definitely a "consenting adults" sort of situation where if your code breaks, fine, just fix it.

honnibal:
After further discussion, we'll pass in `stdout` and `stderr` handles to avoid leaking the wasabi details.

spacy/cli/train.py
```diff
@@ -50,14 +56,4 @@ def train_cli(
         nlp = init_nlp(config, use_gpu=use_gpu)
     msg.good("Initialized pipeline")
     msg.divider("Training pipeline")
-    train(nlp, output_path, use_gpu=use_gpu, silent=False)
-
-
-def verify_cli_args(config_path: Path, output_path: Optional[Path] = None) -> None:
```
ines, line 56: Are we going to revert all this stuff again now? If so, we should probably do this consistently across all the other functions that we changed based on your initial requests

website/docs/usage/training.md
```diff
@@ -729,13 +729,14 @@ def custom_logger(log_path):
                 file_.write(f"loss_{pipe}\\t")
             file_.write("\\n")
```
ines, line 731: We also need to add the `stdout` and `stderr` arguments to the `setup_logger` function in the example above

spacy/cli/train.py
honnibal: As we've removed almost all of the CLI args from that function, it felt a bit better to just inline it again. I wouldn't go through and make the other changes, I don't think it's important that that's consistent. I can revert it if you'd prefer they were all the same though.

website/docs/usage/training.md
honnibal: They're optional, so the example there is legal. If you want to options explicit then sure, I'll add them.

honnibal: Good point. I also updated the example slightly to use the `finalize` callback.

ines (approved):

PR #6186, opened 2020-10-02
Change type of texts argument in pipe to iterable

Stannislav:
The `texts` argument in `Language.pipe` should be `iterable`, not `iterator`.

A `list` is an `iterable`, but not an `iterator`.

## Description
- Iterable: defines `__iter__()`, returns an iterator
- Iterator: defines `__next__()`, defines `__iter()` that returns `self`.

Not all iterables are iterators. Therefore such iterables should not be passed to `Language.pipe` via the
`texts` parameter according to the docstring. One example of such iterable that is not an iterator is a list.
Therefore not even lists should be passed to `Language.pipe`, which seems wrong.

### Types of change
Wrong argument type in docstring.

## Checklist

### DIFF
--- a/.github/contributors/Stannislav.md
+++ b/.github/contributors/Stannislav.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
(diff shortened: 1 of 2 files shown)


svlandeg:
Good catch, thanks!

It's great to see others also enthousiastic about the (proper ;-)) use of typing in the code base. In the upcoming v3, we'll extend this considerably and have proper types integrated in the core lib. I just double checked and on the `develop` branch where we're preparing v3, we actually do have the type correct for this specific method:
```
def pipe(
        self,
        texts: Iterable[str],
        *,
        as_tuples: bool = False,
        batch_size: int = 1000,
        disable: Iterable[str] = SimpleFrozenList(),
        component_cfg: Optional[Dict[str, Dict[str, Any]]] = None,
        n_process: int = 1,
    ):
```
So yes - we agree with you it should be `iterable` and not `iterator`!

We'll definitely include your edit in the next bugfix release for v2. Thanks again!

PR #6187, opened 2020-10-02
compat warning for pipe.begin_training

svlandeg:
## Description
We currently provide bwd-compatibility in case users call the old `nlp.begin_training()` instead of the new `nlp.initialize()`. But there would also be a problem if a user has a custom component that extends `Pipe` and implements `begin_training`: because of the "dummy" implementation in `Pipe.initialize`, the user's `begin_training` would silently be ignored as `Pipe.initialize` would be called, causing potentially weird problems downstream when it looks like the component is initialized, but in practice it's not. 

I'm not particularly fond of the solution proposed in this PR, and I wouldn't necessarily keep it in the codebase very long, but I think we need something especially right after the launch of v3, to warn users that they should rename this function name in their custom component. Another solution would be to just raise the warning, but not actually call (the old) `proc.begin_training`.

(the reason why I'm not happy with this solution is because we need to put this `if` first and we can't have it second, as the condition `if hasattr(proc, "initialize")` is always `True` when inheriting from `Pipe`.)

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -85,9 +85,9 @@ class Warnings:
             "attribute or operator.")
 
     # TODO: fix numbering after merging develop into master
-    W089 = ("The 'begin_training' method has been renamed to 'initialize', "
-            "for calls to 'nlp' as well as for the individual pipeline "
-            "components.")
+    W088 = ("This component implements a 'begin_training' method, "
+            "which should probably be renamed to 'initialize'.")
+    W089 = ("The nlp.begin_training method has been renamed to nlp.initialize.")
     W090 = ("Could not locate any {format} files in path '{path}'.")
     W091 = ("Could not clean/remove the temp directory at {dir}: {msg}.")
     W092 = ("Ignoring annotations for sentence starts, as dependency heads are set.")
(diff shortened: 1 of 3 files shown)


ines:
> But there would also be a problem if a user has a custom component that extends Pipe and implements begin_training

I wonder how common this actually is, because this type of workflow isn't particularly convenient in v2.x and `Pipe` isn't even officially documented for v2. Also, any component like this will require some significant refactoring anyways to work with v3 (which will most likely lead the user to the `initialize` change). So I think we might be okay without the check in `Language`?

That said, some thing we could do are:

* Add a `Pipe.begin_training` method that raises
* Or, if we really want to guard against users implementing `begin_training` methods in subclasses: add an [`__init_subclass__`](https://www.python.org/dev/peps/pep-0487/) method to `Pipe` that raises or warns if the subclass has a `begin_training` attr

honnibal:
> Or, if we really want to guard against users implementing begin_training methods in subclasses: add an __init_subclass__ method to Pipe that raises or warns if the subclass has a begin_training attr

This was the solution I thought of, I think it's the right way: raise if the subclass has a `begin_training` but not an `initialize()`.

svlandeg:
I agree this is a nicer solution :-)

PR #6188, opened 2020-10-02
Few minor fixes

svlandeg:
## Description
* add types of `Tok2Vec` embedding layers
* remove redundant `str()` call
* consistently document as `--code` instead of sometimes using `--code-path`
* print the error message (as warning) when a `Pipe` is called and something goes wrong. Now that v3 makes it much easier to implement your own models, this can also go wrong, and their may be various reasons for a `KeyError`. So the actual message is useful for debugging purposes.

### Types of change
small fixes / enhancements

## Checklist


spacy/language.py
```diff
@@ -972,7 +972,8 @@ def __call__(
                 raise ValueError(Errors.E003.format(component=type(proc), name=name))
             try:
                 doc = proc(doc, **component_cfg.get(name, {}))
-            except KeyError:
+            except KeyError as e:
```
ines, line 975: I think we should just just catch all exceptions here and then raise our `E109` from them. So basically, `raise ValueError(...) from e`. 

This will make the traceback more verbose but it will actually show where the problem happened (with traceback) and will show up as something like "This error was the direct cause of the following...".

I'll just go ahead and update this and then merge your PR 🙂

PR #6195, opened 2020-10-04
upgrade to a fresher blis package

stonebig:
blis-0.7.1 is out since august 16th.

## Checklist


stonebig:
was a test, it would really be helpful that for Python>=3.8, you switch to blis-0.7.1

honnibal:
Hi @stonebig ,

We'll be updating to v0.7.1 for the v3 release, which is coming soon. But for this version of Blis, I've switched over to the manylinux2014 container standard so that I can compile the kernels properly for AMD. This will cause problems for people on older pip versions, so I don't want to switch spaCy over until we do a major version, as it could break users' CI if suddenly a new Blis version comes out and their installation fails.

PR #6197, opened 2020-10-04
docs on implementing new pipe component

svlandeg:
## Description
Docs section "Create new trainable components"

I wasn't sure at times how much detail to provide, but once the example project is finished, the full reference implementation will be available.

### Types of change
docs

## Checklist


ines:
Looks good, thanks!  🎉 I'll merge and maybe make some small edits on `devlop` as I proofread in detail.

I think the level of detail is definitely good here – ultimately, the project template will provide the full context and the docs should show the relevant parts for the different steps. So I think this is a good template to follow for future examples like this as well.

PR #6202, opened 2020-10-05
Include spaCy version check in project CLI

ines:
## Description

- [x] Include spaCy minor version and commit hash in `project.lock`
- [x] Re-run a command if it was last run with a different minor version of spaCy
- [x] Add mechanism to use git commit hash instead (stricter)
- [x] Include spaCy minor version in hashes used for remote storage
- [x] Add `spacy_version` to `project.yml` to allow projects to specify a compatible spaCy version range, and raise if it doesn't match

### Todo

- [x] decide how the stricter git commit check should be exposed – do we want a CLI flag on `project run`? Is this something we mostly intend to use internally for testing? Should it be an env variable?

### Types of change
enhancement

## Checklist


spacy/cli/project/run.py
```diff
@@ -171,12 +173,18 @@ def validate_subcommand(
         )
 
 
-def check_rerun(project_dir: Path, command: Dict[str, Any]) -> bool:
+def check_rerun(
+    project_dir: Path,
+    command: Dict[str, Any],
+    check_spacy_version: bool = True,
```
honnibal, line 179:
```suggestion
    *,
    check_spacy_version: bool = True,
```

honnibal:
> decide how the stricter git commit check should be exposed – do we want a CLI flag on project run? Is this something we mostly intend to use internally for testing? Should it be an env variable?

Maybe this is a dumb idea, but could this be a case for that `spacy.about.is_release` variable we've had sitting around? By default if `is_release` is `False` it uses the git version, but there's an environment variable to overrule?

ines:
> Maybe this is a dumb idea, but could this be a case for that spacy.about.is_release variable we've had sitting around?

Can we be sure that it's always `False` in development, though? Like, on `develop` it's currently `True`.

If it's mostly just for us internally for development, then I do think an env variable makes sense.

honnibal:
> Can we be sure that it's always False in development, though? Like, on develop it's currently True.

Yeah we'd actually have to start caring about it. I think we should either do something with it (like this) or get rid of it. Getting rid makes sense too.

ines:
> Yeah we'd actually have to start caring about it. I think we should either do something with it (like this) or get rid of it. Getting rid makes sense too.

I'd be fine with getting rid of it tbh.

PR #6204, opened 2020-10-05
Remove output dirs before training

ines:
## Description

- [x] make sure `model-best` and `model-last` are removed to prevent stale data (e.g. from components that don't exist anymore)
- [x] show warning if removing fails

### Types of change
fix

## Checklist

### DIFF
--- a/spacy/training/loop.py
+++ b/spacy/training/loop.py
@@ -3,17 +3,22 @@ from typing import Optional, TYPE_CHECKING
 from pathlib import Path
 from timeit import default_timer as timer
 from thinc.api import Optimizer, Config, constant, fix_random_seed, set_gpu_allocator
+from wasabi import Printer
 import random
-import wasabi
 import sys
+import shutil
 
 from .example import Example
 from ..schemas import ConfigSchemaTraining
-from ..errors import Errors
-from ..util import resolve_dot_names, registry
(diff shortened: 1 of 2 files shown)


spacy/training/loop.py
```diff
@@ -305,3 +311,20 @@ def before_to_disk(nlp: Language) -> Language:
...
+    by nlp.to_disk, which could preserve existing subdirectories (e.g.
+    components that don't exist anymore).
+    """
+    if path is not None and path.exists():
+        for subdir in [path / DIR_MODEL_BEST, path / DIR_MODEL_LAST]:
+            if subdir.exists():
+                try:
+                    shutil.rmtree(str(subdir))
+                    logger.debug(f"Removed existing output directory: {subdir}")
+                except Exception as e:
```
honnibal, line 328: Shouldn't we let this raise?

spacy/training/loop.py
```diff
@@ -305,3 +311,20 @@ def before_to_disk(nlp: Language) -> Language:
...
+                    err = Warnings.W087.format(path=subdir, err=e)
+                    stdout.write(msg.warn(err) + "\n")
```
honnibal, line 330: This would go to `stderr` rather than stdout I think. But I do think we should just let the error happen.

PR #6205, opened 2020-10-05
Improve MultiHashEmbed arguments

honnibal:
The `MultiHashEmbed` layer hard-coded which attributes it used, which was a bit inflexible. The new argument scheme allows attribute names to be passed in, and allows a separate number of rows to be specified for each embedding table.


spacy/ml/models/tok2vec.py
```diff
@@ -92,59 +96,90 @@ def build_Tok2Vec_model(
...
+        attrs=[NORM, PREFIX, SUFFIX, SHAPE] if also_embed_subwords else [NORM],
+        include_static_vectors=also_use_static_vectors,
+    )
+
+
+@registry.architectures.register("spacy.MultiHashEmbed.v2")
+def MultiHashEmbed(
+    width: int,
+    rows: int,
+    attrs: Union[List[Union[str, int]], Dict[Union[str, int], float]],
```
ines, line 116: I definitely see the point here but I do think as a type, this is too weird. We really wouldn't want arguments (especially not those of registered functions) to take very different types and behave differently based on what type you pass in.

Even though it's possible, it kinda goes against the idea of the config. In this case, the different data types will also end up looking _very_ different in the config (list vs. block), which can easily cause confusion.

So I'd rather have another argument here that can be `None` or a dict/list of factors (kinda like how the `words`/`spaces` work), with a default of `None` (default factors). That would make it easy to validate (simple length check) and wouldn't take up more space in the config if set.

honnibal: You'd normally be writing it inline rather than a block, like `attrs = {"NORM": 1.0, "PREFIX": 0.5, "SUFFIX": 0.5, "SHAPE: 0.5}` is what we'll usually have.

A parallel list is fine though, it was definitely my second choice. I'll change it.

honnibal: Alternatively we could always make the row sizing explicit, e.g.

```
rows = [5000, 2500, 2500, 2500]
```

Maybe that's better?

honnibal: Yeah the separate list is definitely better.

PR #6206, opened 2020-10-05
EntityRuler from config

svlandeg:
Various fixes to allow for a non-trainable, non-frozen component in the CLI training script, such as the `EntityRuler`.

## Description
- Have a `finish_update` method on `Pipe` that delegates to `self.model` by default
- Have an `is_trainable` method on `Pipe` that checks whether the component has a `self.model` of type `Model`
- Implement `initialize` for `EntityRuler` so it can read its patterns from file:
```
[initialize.components.entity_ruler]

[initialize.components.entity_ruler.patterns]
@readers = "srsly.read_jsonl.v1"
path = ${paths.patterns}
```

### Types of change
enhancement

## Checklist


spacy/pipeline/entityruler.py
```diff
@@ -183,6 +185,27 @@ def labels(self) -> Tuple[str, ...]:
                 all_labels.add(l)
         return tuple(all_labels)
 
+    def initialize(
+        self,
+        get_examples: Callable[[], Iterable[Example]],
+        *,
+        nlp: Optional[Language] = None,
+        patterns_path: Optional[Path] = None
```
ines, line 193: We should make this `patterns` and have it take the actual patterns. Reading from the path should happen in the config, e.g. via `srsly.read_json.v1` (just like we currently do it for the attribute ruler etc.)

svlandeg: Right, sorry, I lost track a bit of the latest refactors...

svlandeg: Ok fixed (I hope), will also update original post to avoid later confusion

svlandeg:
What annoys me still are the direct calls to `proc.pipe()`. Before, we would only check whether `pipe()` was an available method. However, non-trainable components inheriting from `Pipe` need to implement dummy versions of `predict` and `set_annotations`, and then their `pipe()` method will be a dummy in practice too.

This means that the check `hasattr(proc, "pipe")` needs to be much more verbose and also check `pipe.is_trainable`. I found this three times in the code base (one being the reason why the unit tests failed originally). Not sure what a clear way to solve this would be.

PR #6213, opened 2020-10-06
TrainablePipe

svlandeg:
Add `TrainablePipe` that inherits from `Pipe`.

## Description
- [x] Keep only model-related code in `TrainablePipe`
- [x] Add types to the code in `Pipe` and `TrainablePipe`
- [x] Fix types in `nlp._pipe` (this works on documents, not on examples)
- [x] Remove unnecessary functions from `attribute_ruler`, `entity_ruler` and `sentencizer`
- [x] Removed serialization methods and `vocab`/`cfg` fields from `Pipe` 
- [x] Moved `Pipe._ensure_examples` to `Example.validate_get_examples` to be more like `Example.validate_examples`
- [x] Focus on `TrainablePipe` in the docs, keep all links to refer to `/api/pipe`
- [x] Don't serialize the `vocab` within components
- [x] Track added strings (e.g. labels) for serialization so the components can be self-contained

## Checklist


ines:
> What to do with the serialization methods?

IMO `Pipe` shouldn't have dummy serialization methods because we shouldn't assume that all pipes need to serialize state and that they have a `vocab` or `cfg`. For the `TrainablePipe`, that's different of course.

>  Update docs. Do we document Pipe or only TrainablePipe?

I think we should focus on `TrainablePipe` for now, as you did in your edits. I'll also update the illustrations accordingly (I think there's one or two that currently reference `Pipe`). 

I'd also be happy to keep the page URL as `/api/pipe` for now.

svlandeg:
> I'd also be happy to keep the page URL as `/api/pipe` for now.

Yea, that's what I figured to keep the current URLs alive too...

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -0,0 +1,313 @@
...
+    def to_disk(self, path, *, exclude=tuple()):
+        """Serialize the pipe to disk.
+
+        path (str / Path): Path to a directory.
+        exclude (Iterable[str]): String names of serialization fields to exclude.
+
+        DOCS: https://nightly.spacy.io/api/pipe#to_disk
+        """
+        serialize = {}
+        serialize["cfg"] = lambda p: srsly.write_json(p, self.cfg)
```
ines, line 287: Should we try to be more accommodating here and include `hasattr` checks? Otherwise, we'd have to be very clear that custom `__init__` methods of a `TrainablePipe` subclass always _have to_ to define `self.vocab` and `self.cfg` (which is currently more of a convention but not required – and I'm not sure it should be?).

svlandeg: Good idea, can't hurt to add an extra check.

spacy/language.py
```diff
@@ -1300,7 +1300,7 @@ def evaluate(
             kwargs.setdefault("batch_size", batch_size)
             # non-trainable components may have a pipe() implementation that refers to dummy
             # predict and set_annotations methods
-            if not hasattr(pipe, "pipe") or not hasattr(pipe, "is_trainable") or not pipe.is_trainable():
+            if not hasattr(pipe, "pipe"):
```
honnibal, line 1303: Should we move this conditional into the `_pipe` function, and move that function to `spacy.util`?

spacy/pipeline/pipe.pyx
```diff
@@ -269,81 +78,24 @@ cdef class Pipe:
...
-        if hasattr(self, "vocab"):
-            serialize["vocab"] = self.vocab.to_bytes
-        return util.to_bytes(serialize, exclude)
-
-    def from_bytes(self, bytes_data, *, exclude=tuple()):
-        """Load the pipe from a bytestring.
-
-        exclude (Iterable[str]): String names of serialization fields to exclude.
-        RETURNS (Pipe): The loaded object.
+    def is_trainable(self) -> bool:
```
honnibal, line 81: Why did we make this a method rather than a property?

ines: Presumably because `is_resizable` is also a method (not sure why, though?)

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -0,0 +1,319 @@
...
+
+from ..training import validate_examples
+from ..errors import Errors
+from .pipe import Pipe, deserialize_config
+from .. import util
+from ..vocab import Vocab
+from ..language import Language
+from ..training import Example
+
+cdef class TrainablePipe(Pipe):
```
honnibal, line 16: I can't remember whether this needed to be a `cdef class`. Is it because we need its subclasses to be `cdef class`, and they can't inherit from Python types? I think that's it, but I just wanted to check.

ines: Yeah, I tested this during the refactor I think (and maybe @svlandeg did as well?) and I'm pretty sure we couldn't have the other `cdef` classes inherit from it otherwise.

honnibal:
I think the `TrainablePipe` class allows us to get in some extra consistency that previously we couldn't consider, because `Pipe` had to cater for simple components as well as trainable ones. I suggest:

* Make the vocab required for `TrainablePipe` instances, not optional.
* Change `TrainablePipe.vocab` to be declared as type `vocab`, rather than type `object`.
* Don't serialize the vocab

Currently we never serialize the vocab in the components anyway, because we always put it in the `exclude`. But this just makes a user trap...Getting this wrong would cause strange things to happen anyway. Typing the vocab instance will help subclasses use the Cython API of the vocab, without having to set it to a typed instance. It also allows compile-time checks.

spacy/pipeline/pipe.pyx
honnibal: I think they should both be properties.

svlandeg: Yep, agreed!

ines:
> Make the vocab required for TrainablePipe instances, not optional.

Okay, with the typing, this makes sense. We just need to remember to update the docs for trainable components then.

honnibal:
Another suggestion:

* Add a variable `TrainablePipe._added_strings`
* Add a method `TrainablePipe.add_string`
* Always serialize `TrainablePipe._added_strings` as `strings.json`
* During deserialization, ensure the strings are added to the `self.vocab.strings`

When components add labels etc, they should track their strings separately. This way if they're sourced into a new model, they are self-contained, and we don't get string hash errors.

svlandeg:
>     * Change `TrainablePipe.vocab` to be declared as type `vocab`, rather than type `object`.

I don't see where vocab is declared as `object`?

svlandeg:
I think this approach with `strings.json` makes a lot of sense. Just hope the implemenation is OK. I'll add a few more unit tests.

spacy/kb.pyx
```diff
@@ -97,6 +93,7 @@ cdef class KnowledgeBase:
         self.vocab = vocab
         self.vocab.strings.add("")
         self._create_empty_vectors(dummy_hash=self.vocab.strings[""])
+        self._added_strings = [""]
```
honnibal, line 96: This should definitely be a set, not a list. We also don't need the empty string, that's built-in.
```suggestion
        self._added_strings = set()
```

spacy/kb.pxd
```diff
@@ -28,8 +28,9 @@ cdef class Candidate:
 
 cdef class KnowledgeBase:
     cdef Pool mem
-    cpdef readonly Vocab vocab
+    cpdef public Vocab vocab
     cdef int64_t entity_vector_length
+    cdef public list _added_strings
```
honnibal, line 33:
```suggestion
    cdef public set _added_strings
```

spacy/kb.pxd
```diff
@@ -28,8 +28,9 @@ cdef class Candidate:
 
 cdef class KnowledgeBase:
     cdef Pool mem
-    cpdef readonly Vocab vocab
+    cpdef public Vocab vocab
```
honnibal, line 31:
```suggestion
    cdef readonly Vocab vocab
```

We'll normally want to make this readonly, as people could get segfaults if they mess with it.

spacy/kb.pyx
```diff
@@ -118,12 +115,16 @@ cdef class KnowledgeBase:
     def get_alias_strings(self):
         return [self.vocab.strings[x] for x in self._alias_index]
 
+    def add_string(self, string: str):
+        self._added_strings.append(string)
```
honnibal, line 119:
```suggestion
        self._added_strings.add(string)
```

spacy/pipeline/attributeruler.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         self.attrs = []
         self._attrs_unnormed = []  # store for reference
         self.indices = []
+        self._added_strings = []
```
honnibal, line 60:
```suggestion
        self._added_strings = set()
```

spacy/pipeline/attributeruler.py
```diff
@@ -202,11 +188,16 @@ def add(
...
         self.matcher.add(self.vocab.strings.add(key), patterns)
+        self.add_string(key)
         self._attrs_unnormed.append(attrs)
         attrs = normalize_token_attrs(self.vocab, attrs)
         self.attrs.append(attrs)
         self.indices.append(index)
 
+    def add_string(self, string: str):
+        if string not in self._added_strings:
+            self._added_strings.append(string)
```
honnibal, line 199:
```suggestion
            self._added_strings.add(string)
```

spacy/pipeline/morphologizer.pyx
```diff
@@ -95,6 +95,7 @@ class Morphologizer(Tagger):
         # add mappings for empty morph
         self.cfg["labels_morph"][Morphology.EMPTY_MORPH] = Morphology.EMPTY_MORPH
         self.cfg["labels_pos"][Morphology.EMPTY_MORPH] = POS_IDS[""]
+        self._added_strings = []
```
honnibal, line 98:
```suggestion
        self._added_strings = set()
```

spacy/pipeline/pipe.pyx
```diff
@@ -269,81 +78,25 @@ cdef class Pipe:
...
+    @property
+    def is_trainable(self) -> bool:
+        return False
 
-        path (str / Path): Path to a directory.
-        exclude (Iterable[str]): String names of serialization fields to exclude.
-        RETURNS (Pipe): The loaded object.
+    @property
+    def labels(self) -> Optional[Tuple[str]]:
+        return []
```
honnibal, line 87:
```suggestion
        return tuple()
```

spacy/pipeline/senter.pyx
```diff
@@ -62,6 +61,7 @@ class SentenceRecognizer(Tagger):
         self.name = name
         self._rehearsal_model = None
         self.cfg = {}
+        self._added_strings = []
```
honnibal, line 64:
```suggestion
        self._added_strings = set()
```

spacy/pipeline/tagger.pyx
```diff
@@ -77,6 +78,7 @@ class Tagger(Pipe):
         self._rehearsal_model = None
         cfg = {"labels": labels or []}
         self.cfg = dict(sorted(cfg.items()))
+        self._added_strings = []
```
honnibal, line 81:
```suggestion
        self._added_strings = set()
```

spacy/pipeline/textcat.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         self._rehearsal_model = None
         cfg = {"labels": [], "threshold": threshold, "positive_label": None}
         self.cfg = dict(cfg)
+        self._added_strings = []
```
honnibal, line 113:
```suggestion
        self._added_strings = set()
```

spacy/pipeline/tok2vec.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, vocab: Vocab, model: Model, name: str = "tok2vec") -> None:
         self.name = name
         self.listeners = []
         self.cfg = {}
+        self._added_strings = []
```
honnibal, line 67:
```suggestion
        self._added_strings = set()
```

spacy/pipeline/trainable_pipe.pxd
```diff
@@ -0,0 +1,8 @@
+from .pipe cimport Pipe
+from ..vocab cimport Vocab
+
+cdef class TrainablePipe(Pipe):
+    cdef public Vocab vocab
```
honnibal, line 5:
```suggestion
    cdef readonly Vocab vocab
```

Again it's tough to decide between public and readonly, maybe it's better to be flexible and allow public. But segfaults suck, and we don't want to worry that this is a problem, so let's just do readonly.

spacy/pipeline/trainable_pipe.pxd
```diff
@@ -0,0 +1,8 @@
...
+    cdef public object model
+    cdef public object cfg
+    cdef public list _added_strings
```
honnibal, line 8:
```suggestion
    cdef public set _added_strings
```

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -0,0 +1,326 @@
...
+        name (str): The component instance name.
+        **cfg: Additonal settings and config parameters.
+
+        DOCS: https://nightly.spacy.io/api/pipe#init
+        """
+        self.vocab = vocab
+        self.model = model
+        self.name = name
+        self.cfg = dict(cfg)
+        self._added_strings = []
```
honnibal, line 38:
```suggestion
        self._added_strings = set()
```

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -0,0 +1,326 @@
...
+
+        label (str): The label to add.
+        RETURNS (int): 0 if label is already present, otherwise 1.
+
+        DOCS: https://nightly.spacy.io/api/pipe#add_label
+        """
+        raise NotImplementedError(Errors.E931.format(parent="Pipe", method="add_label", name=self.name))
+
+    def add_string(self, string: str):
+        if string not in self._added_strings:
```
honnibal, line 202: Using a set ensures that adding is expected to be constant time, we don't have to make the membership check (which is `O(n)` for a list).

```suggestion
        self._added_strings.add(string)
```

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -0,0 +1,326 @@
...
+            self._added_strings.append(string)
+
+    @property
+    def is_trainable(self) -> bool:
+        return True
+
+    @property
+    def is_resizable(self) -> bool:
+        return hasattr(self, "model") and "resize_output" in self.model.attrs
```
honnibal, line 211: Some `TrainablePipe` instances will have `self.model` set to `None`, which fails if we just do `hasattr`. This should work:

```suggestion
        return getattr(self, "model", None) and "resize_output" in self.model.attrs
```

honnibal (changes requested):
Looking good! Just need to change the type of the variable, and those Cython details.

spacy/pipeline/trainable_pipe.pxd
svlandeg: It looks like we need to keep this one `public` - the factories (parsing `nlp` from config) seem to not work otherwise.

honnibal (approved):

honnibal:
Great, looks good!

Go ahead and merge it if it's ready, and then could you make the PR for `spacy-transformers`?

PR #6216, opened 2020-10-07
kb_loader in NEL's initialize method

svlandeg:
[EDIT] : removed the `init_config` thing again after discussion (see below), but kept the `kb_loader` as an argument of `entity_linker.initialize`

## Description
The `nlp` methods `rename_pipe` and `remove_pipe` currently already take care of syncing `nlp._config["initialize"]["components"]`, but I think we also need a setting in `add_pipe` and `create_pipe`. Proposal to define an additional argument `init_config` (the naming is very unfortunate... ideas?) 

This allows us to have the `kb_loader` argument on `entity_linker.initialize` instead of in the constructor. This makes reading a KB from a file path cleaner as it only happens during initialization and not at run-time.
```
config = {"incl_prior": False}
init_config = {"kb_loader": {"@misc": "spacy.KBFromFile.v1", "kb_path": "my_kb"}}
entity_linker = nlp.add_pipe("entity_linker", config=config, init_config=init_config, last=True)
```

Have not yet updated the docs - will do if merged.

### Types of change
enhancement
## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -600,7 +600,6 @@ class Language:
         *,
         config: Optional[Dict[str, Any]] = SimpleFrozenDict(),
         raw_config: Optional[Config] = None,
-        init_config: Optional[Dict[str, Any]] = SimpleFrozenDict(),
         validate: bool = True,
     ) -> Callable[[Doc], Doc]:
         """Create a pipeline component. Mostly used internally. To create and
@@ -612,9 +611,6 @@ class Language:
         config (Optional[Dict[str, Any]]): Config parameters to use for this
             component. Will be merged with default config, if available.
         raw_config (Optional[Config]): Internals: the non-interpolated config.
-        init_config (Optional[Dict[str, Any]]): Config parameters to use to
-            initialize this component. Will be used to update the internal
-            'initialize' config.
         validate (bool): Whether to validate the component config against the
             arguments and types expected by the factory.
         RETURNS (Callable[[Doc], Doc]): The pipeline component.
@@ -625,13 +621,9 @@ class Language:
         if not isinstance(config, dict):
             err = Errors.E962.format(style="config", name=name, cfg_type=type(config))
             raise ValueError(err)
-        if not isinstance(init_config, dict):
-            err = Errors.E962.format(style="init_config", name=name, cfg_type=type(init_config))
             raise ValueError(err)
         if not srsly.is_json_serializable(config):
             raise ValueError(Errors.E961.format(config=config))
-        if not srsly.is_json_serializable(init_config):
-            raise ValueError(Errors.E961.format(config=init_config))
         if not self.has_factory(factory_name):
             err = Errors.E002.format(
                 name=factory_name,
@@ -643,8 +635,6 @@ class Language:
             raise ValueError(err)
         pipe_meta = self.get_factory_meta(factory_name)
         config = config or {}
-        if init_config:
-            self._config["initialize"]["components"][name] = init_config
         # This is unideal, but the alternative would mean you always need to
         # specify the full config settings, which is not really viable.
         if pipe_meta.default_config:
@@ -719,7 +709,6 @@ class Language:
         source: Optional["Language"] = None,
         config: Optional[Dict[str, Any]] = SimpleFrozenDict(),
         raw_config: Optional[Config] = None,
-        init_config: Optional[Dict[str, Any]] = SimpleFrozenDict(),
         validate: bool = True,
     ) -> Callable[[Doc], Doc]:
         """Add a component to the processing pipeline. Valid components are
@@ -742,9 +731,6 @@ class Language:
         config (Optional[Dict[str, Any]]): Config parameters to use for this
             component. Will be merged with default config, if available.
         raw_config (Optional[Config]): Internals: the non-interpolated config.
-        init_config (Optional[Dict[str, Any]]): Config parameters to use to
-            initialize this component. Will be used to update the internal
-            'initialize' config.
         validate (bool): Whether to validate the component config against the
             arguments and types expected by the factory.
         RETURNS (Callable[[Doc], Doc]): The pipeline component.
@@ -778,7 +764,6 @@ class Language:
                 name=name,
                 config=config,
                 raw_config=raw_config,
-                init_config=init_config,
                 validate=validate,
             )
         pipe_index = self._get_pipe_index(before, after, first, last)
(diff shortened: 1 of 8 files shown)


ines:
I've been thinking about this as well but I wasn't sure if it's worth the extra complexity. Unlike the other arguments of `Language.add_pipe`, the initialization config is not relevant for the component creation, so it doesn't necessarily have to be specified when you create the component. If you're adding the component explicitly in Python, you can just call `component.initialize()` and pass it the loaded KB. (Or maybe I'm misunderstanding the specific goal here for the entity linker?)

I do agree that what's currently missing is a way for components to define defaults for initialization, if they want to? It's not necessarily needed for arguments like `label` (of the tagger/parser/NER) because that just defaults to `None`. But some components may always need data to be passed on on initialization, so the config you export might as well include this by default. But this should probably work like the `default_config` on the `@Language.factory` decorator.

svlandeg:
> If you're adding the component explicitly in Python, you can just call `component.initialize()` and pass it the loaded KB. 

That was my first idea as well, but that gets a little weird, because you also need to provide examples at that point. So you'd call entity_linker.initialize, but you still also need to call nlp.initialize (e.g. for initializing the NER), so you'd end up "initializing" the NEL twice, resulting in some redundancy and unnecessary looping over the data etc. I guess you could also initialize nlp without the entity_linker first, but that starts getting more messy I think?

[apologies for messing up your original comment. I hit edit instead of quote...]

ines:
> That was my first idea as well, but that gets a little weird, because you also need to provide examples at that point. So you'd call entity_linker.initialize, but you still also need to call nlp.initialize (e.g. for initializing the NER), so you'd end up "initializing" the NEL twice

Oh, that's a good point but I think this points to a deeper design/workflow question: a component should only ever be initialized once. If we're initializing from the config file, that's easy, because we can load everything at once and initialize the model from examples and the KB from a file.

But if you're setting up the pipeline manually and you really just care about loading in a KB from a file, you can do this:

```
entity_linker = nlp.add_pipe("entity_linker", config={"incl_prior": False})
entity_linker.kb.from_disk("/path")
```

The question is, should an action like this be represented in the `[initialize]` config so there's a record of the fact that the KB was loaded from disk? I'm not sure. I do think the initialization is something you consider before training and if we recorded too much of this under the hood without being explicit about it, you'd end up with pretty weird configs.

I even think that writing to `nlp.config["initialize"]` is something you shouldn't really do, unless you really want to modify the initialization config that's saved out with the model etc. (Obviously excluding our test suite, where we do this a lot to set up the right state.)

svlandeg:
But you agree on having the `kb_loader` in `entity_linker.initialize` instead of in the constructor, right? So if you do this with a config, you'd get that `initialize` block recorded which I think is nice, because you know where the KB came from. It would look like this:

```
[initialize.components.entity_linker.kb_loader]
@misc = "spacy.KBFromFile.v1"
kb_path = ${paths.kb}
```

I assumed we want this in the `initialize` block, because otherwise the `entity_linker` will, after serialization, still always refer to the "old" KB from `kb_path` instead of the one it serialized as part of its component folder (I think this was a bug before, TBH).

 So if we agree on that, the main question to me is how to "translate" the "config approach" to the pure python script approach. It feels weird to then not call `initialize` with this `kb_loader`, but replicate the functionality with `entity_linker.kb.from_disk` instead. I guess theoretically we could have a `set_kb` method like we had before though. But the `nlp.config` will look different as opposed to the full config approach.

To answer your question
> The question is, should an action like this be represented in the [initialize] config so there's a record of the fact that the KB was loaded from disk?

Following my reasoning above, I personally think that yes, if that's how it would be if we'd be working with a "config approach", I'd want it recorded. 

On the other issue:
> a component should only ever be initialized once

How should/could we prevent this happening more than once?

ines:
Yes, I think the `[initialize.components.entity_linker.kb_loader]` is good and that's how we should do it. This is also going to solve the awkwardnesses we previously had around initializing a pipeline for training with a KB.

> Following my reasoning above, I personally think that yes, if that's how it would be if we'd be working with a "config approach", I'd want it recorded.

That's kinda problematic, though and we don't do that anywhere else within the actual components. If you write to a component manually, load in data, update its settings etc. it's not updating the component config. The config only records the settings the component was created with.

I also think we should treat the initialization that way: realistically, it makes most sense for setting up pipelines before training because here, the config has to represent _everything_ and you can't just call arbitrary component methods in your code. If you're setting up the components in code, well, then you can also write to them directly.

Other components work in a similar way here: the `EntityRuler` and `AttributeRuler` can be initialized with patterns for a config-only workflow, but also have methods to add patterns manually. Or most trainable components have an `add_label` method that's called during initialization, but can also be called manually.

> How should/could we prevent this happening more than once?

Maybe there are unintended side-effects but I've been wondering if we should have a flag on the `TrainablePipe` class, like `is_initialized` that records whether the `initialized` method has been called before. And then we could log or warn if it's called more than once. Sometimes it might be what you want, to explicitly reset the weights. But in other cases it may indicate a bug.

svlandeg:
> If you're setting up the components in code, well, then you can also write to them directly.

I see your point. I guess my original idea was that we're creating a component with a `config` for its constructor and we're providing that config explicitely to `add_pipe`, so we'd want a similar mechanism for the initialization config. 


> Other components work in a similar way here: the `EntityRuler` and `AttributeRuler` can be initialized with patterns for a config-only workflow, but also have methods to add patterns manually. Or most trainable components have an `add_label` method that's called during initialization, but can also be called manually.

I actually think this may lead to some confusion in the long run as well, there are currently quite a few different ways in which to define labels, or implicitely deduce them. But I guess time will tell.

Ok having talked this over, it probably makes sense to not introduce this `init_config` right now, as it's an optional argument that we could always add later if we really want it. I'll rework the PR accordingly.

> Maybe there are unintended side-effects but I've been wondering if we should have a flag on the `TrainablePipe` class, like `is_initialized` that records whether the `initialized` method has been called before. And then we could log or warn if it's called more than once. Sometimes it might be what you want, to explicitly reset the weights. But in other cases it may indicate a bug.

A warning probably can't hurt. I can add that to this PR too.

ines:
> Ok having talked this over, it probably makes sense to not introduce this `init_config` right now, as it's an optional argument that we could always add later if we really want it. I'll rework the PR accordingly.

Yeah, I think we should definitely revisit that "default init settings" point later (probably after the nightly release).

> I actually think this may lead to some confusion in the long run as well, there are currently quite a few different ways in which to define labels, or implicitely deduce them.

I hope that the move to the `config.cfg` for training means that things like adding labels to trainable pipeline components by calling a method will become much less relevant. Outside of testing, you typically don't want to do that, ever. So the default out-of-the-box solution for labels that will "just work" is to read them from the data. And the slightly more advanced option once users are getting more serious about training and want to speed up the process is to use `init labels`, generate a file and load the labels in on initialization.

> A warning probably can't hurt. I can add that to this PR too.

Okay cool. Maybe something like: 

> ```
> Calling `.initialize` on already initialized component '{name}'. This can happen if you're calling `nlp.initialize` or `component.initialize` more than once and may indicate a bug in your code. Initializing a component typically means that its model weights and other data are reset, so you should only initialize it once.
> ```

Also, related: #6222

PR #6224, opened 2020-10-08
Fix morph in Doc.to_json

ines:
## Description

The method is mostly internals at the moment (I think because we weren't sure whether we wanted it like this?) but I've been using it in the Streamlit integration. This fixes the integration of `Token.morph`.

### Types of change
bug fix

## Checklist


spacy/tokens/doc.pyx
```diff
@@ -1421,7 +1419,7 @@ cdef class Doc:
             if include_annotation["POS"]:
                 token_data["pos"] = token.pos_
             if include_annotation["MORPH"]:
-                token_data["morph"] = token.morph_
+                token_data["morph"] = token.morph
```
adrianeboyd, line 1422: I think an explicit `str(token.morph)` would be better here.

adrianeboyd:
Ah, we should check for `morph_` anywhere, since it doesn't exist anymore.

PR #6230, opened 2020-10-09
Include Macedonian language

borijang:
## Description

Added language defaults, lemmatizer, lexical attributes, morph rules, stop words and tokenizer exceptions for Macedonian. Modified the global char_classes.py to include the \u0400-\u04FF Cyrillic block.

### Types of change

Enhancement

## Checklist


spacy/lang/char_classes.py
```diff
@@ -116,7 +116,7 @@
 _latin_extendedE = _latin_l_extendedE
 
 # phonetic letters - Greek, Latin, Cyrillic
-_latin_l_phonetic = r"\u0250-\u02AF\u1D00-\u1D25\u1D6B-\u1D77\u1D79-\u1D9A"
+_latin_l_phonetic = r"\u0250-\u02AF\u1D00-\u1D25\u1D6B-\u1D77\u1D79-\u1D9A\u0400-\u04FF"
```
adrianeboyd, line 119: This isn't the right place for this character range. This is just for phonetic alphabets like IPA characters.

spacy/lang/char_classes.py
```diff
@@ -213,8 +213,12 @@
 _ukrainian_upper = r"А-ЩЮЯІЇЄҐ"
 _ukrainian = r"а-щюяіїєґА-ЩЮЯІЇЄҐ"
 
-_upper = LATIN_UPPER + _russian_upper + _tatar_upper + _greek_upper + _ukrainian_upper
-_lower = LATIN_LOWER + _russian_lower + _tatar_lower + _greek_lower + _ukrainian_lower
+_macedonian_lower = r"ѓѕјњќѐѝа-ик-ш"
+_macedonian_upper = r"ЃЅЈЊЌЀЍА-ИК-Ш"
+_macedonian = r"ѓѕјњќѐѝа-ик-шЃЅЈЊЌЀЍА-ИК-Ш"
```
adrianeboyd, line 218: How much does this overlap with the Russian and Ukrainian character ranges? Do you think it makes sense to consolidate Russian / Tatar / Ukrainian / Macedonian here?

adrianeboyd (changes requested):
This is looking good! A few tests in `spacy/tests/lang/mk` would also be a good idea to make sure all the settings are working as expected. (Don't forget to add an empty `__init__.py` to the new test directory.)

spacy/lang/char_classes.py
borijang: Alright, I will revert this line.

spacy/lang/char_classes.py
borijang: 24 characters are overlapping with both Russian and Ukranian (\u0430 - \u0448 excluding \u0439). If we don't want character redundancy, I will remove them from the Macedonian set. That way, there will be 8 new characters left to be added here.

adrianeboyd (approved):
All the updates and additions look good, thanks!

PR #6246, opened 2020-10-13
Add (noun chunks) syntax iterators for Danish

ophelielacroix:
Added noun chunks function, syntax iterators and test for Danish.
Note: the noun chunks function is an adaptation of the Spanish one. 

### Types of change

new feature

## Checklist


spacy/lang/da/syntax_iterators.py
```diff
@@ -0,0 +1,77 @@
...
+    np_label = doc.vocab.strings.add("NP")
+    np_left_deps = [doc.vocab.strings.add(label) for label in left_labels]
+    np_right_deps = [doc.vocab.strings.add(label) for label in right_labels]
+    stop_deps = [doc.vocab.strings.add(label) for label in stop_labels]
+
+    for token in doclike:
+        if token.pos in [PROPN, NOUN, PRON]:
+            left, right = get_bounds(doc, token)
+            yield (left.i, right.i + 1, np_label)
+            token = right
```
adrianeboyd, line 74: This final line doesn't affect the next token in the `for` loop like you're intending. There is more than one way to do this, but other languages typically keep track of the end of the previous noun chunk so they don't accidentally produce overlapping noun chunks. See `prev_end` in Swedish for an example: https://github.com/explosion/spaCy/blob/bc027dc35c7b2e4482561ca8a818a9b8904b4984/spacy/lang/sv/syntax_iterators.py#L31-L48

spacy/tests/lang/da/test_noun_chunks.py
```diff
@@ -0,0 +1,16 @@
...
+def test_noun_chunks_is_parsed(da_tokenizer):
+    """Test that noun_chunks raises Value Error for 'da' language if Doc is not parsed.
+    To check this test, we're constructing a Doc
+    with a new Vocab here and forcing is_parsed to 'False'
+    to make sure the noun chunks don't run.
+    """
+    doc = da_tokenizer("Det er en sætning")
+    doc.is_parsed = False
+    with pytest.raises(ValueError):
+        list(doc.noun_chunks)
```
adrianeboyd, line 16: Could you add a few tests with example sentences with different kinds of noun chunks to make sure everything is working as intended? Swedish is a good example again here: https://github.com/explosion/spaCy/blob/bc027dc35c7b2e4482561ca8a818a9b8904b4984/spacy/tests/lang/sv/test_noun_chunks.py

Since the tests don't have access to models like `da_core_news_sm`, you have to construct the docs by hand that are used for the tests. The `get_doc` function uses relative heads rather than absolute heads, which you can calculate with `heads = [t.head.i - t.i for t in doc]` from an existing doc.

adrianeboyd (changes requested):
Thanks for this PR, this will be a nice addition! Unfortunately I think the Spanish noun chunker might not have been the best example to follow. At first glance I think it may potentially produce overlapping noun chunks, which typically isn't wanted for chunking and it causes problems for the `merge_noun_chunks` pipeline component.

I've used Swedish below as a pointer, but the noun chunker and tests for many other languages would be just as good as a guideline.

spacy/lang/da/syntax_iterators.py
ophelielacroix: Hi,
I chose to mimic the Spanish iterator because I find it better at finding all the noun chunks, but of course it makes sense to have non-overlapping chunks. 
I actually already had a loop for removing the nested chunks so I integrated it for this solution. Let me know if this solution works well for you.

spacy/tests/lang/da/test_noun_chunks.py
ophelielacroix: I added some sentences in the same type of test as the swedish one.

svlandeg:
Hi @ophelielacroix, it seems like the tests failed. Are you still working on this? Did the tests succeed for you locally?

ophelielacroix:
> Hi @ophelielacroix, it seems like the tests failed. Are you still working on this? Did the tests succeed for you locally?

Thanks for pointing it out, I hadn't seen that it failed. I fixed it. It passes the tests now.

adrianeboyd:
Sorry, there's still a problem with the filtering if it's producing overlapping noun chunks like `"Jacob Jensen"` and `"Jensen"`. This is very close, though, so let me sync this with `master` and add a `prev_end` check instead like I mentioned above. Then it can start yielding noun chunks before the whole document is processed.

adrianeboyd (approved):

PR #6248, opened 2020-10-13
Added Multext-East V5 tagset for Croatian language

danielvasic:
## Description

Added tagmap defined by [Multext-East V5 tagset](https://www.sketchengine.eu/multext-east-croatian-part-of-speech-tagset/) for Croatian language.

## Checklist


adrianeboyd:
We haven't been 100% consistent about PRs like this in the past, but we're trying to move away from maintaining corpus-specific tag maps as part of the language data in the main library, since this can lead to maintaining huge maps that cover multiple corpora or multiple versions of the same corpus.

Instead of loading the tag maps from spacy itself, you can use the `--tag-map-path` option for `spacy train` to load a custom tag map in your new model. Using JSON instead of python is a minor hassle, but see the very bottom section here for a short script that does the conversion: https://spacy.io/usage/v2-3

We'll accept this contribution for v2.3.3, but it will only be included in future v2 releases. In spacy v3, there will no longer be tag maps in the main library and the functionality will be moved from the tagger to a more general-purpose component called `AttributeRuler`: https://nightly.spacy.io/usage/v3#migrating-training-mappings-exceptions

PR #6249, opened 2020-10-13
Test runtime reproducibility

svlandeg:
## Description
Add tests to ensure reproducibilty of trained ML models when making predictions at runtime.
Added similar tests to Thinc in https://github.com/explosion/thinc/pull/407.

### Types of change
tests

## Checklist

### DIFF
--- a/spacy/tests/pipeline/test_models.py
+++ b/spacy/tests/pipeline/test_models.py
@@ -74,7 +74,6 @@ def test_layers_batching_all(model, in_data, out_data):
             util_batch_unbatch_docs_ragged(model, in_data, out_data)
 
 
-
 def util_batch_unbatch_docs_list(
     model: Model[List[Doc], List[Array2d]], in_data: List[Doc], out_data: List[Array2d]
 ):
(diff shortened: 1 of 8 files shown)


spacy/tests/pipeline/test_models.py
```diff
@@ -0,0 +1,106 @@
...
+from spacy.vocab import Vocab
+from thinc.api import get_current_ops, Model, data_validation
+from thinc.types import Array2d, Ragged
+
+from spacy.lang.en import English
+from spacy.ml import FeatureExtractor, StaticVectors
+from spacy.ml._character_embed import CharacterEmbed
+from spacy.tokens import Doc
+
+OPS = get_current_ops()
```
honnibal, line 15: Doing `get_current_ops()` in the global scope is probably bad, because you're either going to get the default (NumpyOps) or maybe somewhere else is setting CupyOps as the default and then it comes down to an import race.

So, I would either move this into a function, or just make a `NumpyOps` instance. Either one is fine.

ines: This should probably just be a `pytest.fixture`?

PR #6227, opened 2020-10-09
Update develop from master

adrianeboyd:
## Description

Update 3.0.0a36 from master

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #6255, opened 2020-10-15
Sync develop and master

ines:
## Checklist


adrianeboyd:
I think the sync should have been up-to-date as of #6227. I intentionally skipped the tag map and morph rules PRs and handled a bunch of python2/v3 updates / merging for the other PRs up to this point.

adrianeboyd:
(I think I rebased too much to make it look cleaner and that was a mistake for this workflow?)

ines:
Yeah, I noticed there wasn't really anything to update or any real conflicts... but for some reason, Git(Hub) still thought a clean merge wasn't possible? 🤷‍♀️ So I guess we just have to keep doing this dance and merge one into the other to make it happy.

adrianeboyd:
But there are a bunch of commits listed that we don't want? I am confused...

ines:
Yeah but otherwise, that history would be gone... so I think the idea is: we update the branch with all commits, resolve the conflicts they create, remove what we don't want and then we have a clean merge. And then as far as Git(Hub) is concerned, it's as if we had updated `master` and then made changes to it on `develop` to end up with the latest state.

Issue #6258, opened 2020-10-15
TokenPatternSchema minItems field is not respected by the validator


jmargeta:
## How to reproduce the behaviour
Based on the TokenPatternSchema schema 
```python
class TokenPatternSchema(BaseModel):
    pattern: List[TokenPattern] = Field(..., minItems=1)
…
```

This should fail to validate
```python
TokenPatternSchema(pattern=[])
```

## Your Environment
- **spaCy version:** 3.0.0rc1
- **Platform:** Linux-5.3.0-7648-generic-x86_64-with-glibc2.27
- **Python version:** 3.8.6

PR #6259, opened 2020-10-15
 Fix TokenPatternSchema validation

jmargeta:
This fixes the non-emptiness check of the field  "pattern" in TokenPatternSchema

## Description
This fixes issue #6258 with a simple replacement of minItems by min_items as described in Pydantic's docs:
https://pydantic-docs.helpmanual.io/usage/schema/

### Types of change
This is a minor fix in one of TokenPatternSchema field constraint

## Checklist


ines:
Thanks a lot! 👍

PR #5143, opened 2020-03-13
Train textcat with config

svlandeg:
[Update 18 March: requires [Thinc 8.0.0a3](https://github.com/explosion/thinc/pull/327)]

## Description
Adopted `textcat` for the new 3.0 design:
* Found & fixed a few bugs in the new implementation. Requires `thinc` > 8.0.0a2 with the new `softmax_activation`
* Implemented the old `build_text_classifier` and use that as current default
* added `textcat_bow_defaults.cfg` and `textcat_cnn_defaults.cfg` to reflect the default implementations on `master` for those specific architectures
* added textcat unit tests for a variety of possible configs (which proved to be extremely useful during development)

Changes to `train_textcat.py:`
* build the NLP model from a config file + provide a default to make the transition to config-based as easy as possible
* support for multi-class and multi-label datasets
* enable specifying a different dataset that gets loaded with `ml_datasets.loaders()` by using `-d`
* enable specifying a threshold for multi-label classification. Labels with an occurrence count below that threshold are discarded from training.

With respect to Tok2Vec:
* revert shenanigans with `_set_dims` for Tok2Vec as initialization can now be done with `tok2vec.initialize(X=docs)`

## Evaluation:
To ensure there are no `textcat` regressions moving from Thinc 7 / spaCy 2 to Thinc 8 / spaCy 3, I performed several tests. Showing in the table the best F-score across 50 epochs, using 2000 training articles, e.g. `train_textcat.py train_textcat_config.cfg -n 50 -d dbpedia`.

Datasets:
 * IMDB: **binary classification**
 * DBpedia: **multi-class, single-label classification**

| Dataset | Architecture | spaCy 2 (current `master`) | spaCy 3 (current PR) |
| ------- | ------------ | -------------------------- | -------------------- |
| IMDB    | BOW          | 78.5 %                       | 81.4 %                 |
| IMDB    | CNN          | 79.1 %                       | 79.4 %                 |
| IMDB    | DEFAULT      | 76.4 %                       | 77.1 %                 |
| DBpedia | BOW          | 90.5 %                       | 90.7 %                 |
| DBpedia | CNN          | 84.8 %                       | 86.4 %                 |
| DBpedia | DEFAULT      | 89.7 %                       | 90.0 %                 |   

This clearly shows there are no regressions - on the contrary, the new implementations seem to work slightly better, even though we attempted to make a 1-1 mapping. I assume Thinc 8 has some issues fixed from the previous version.

[...]

PR #6263, opened 2020-10-16
TextCat updates and fixes

svlandeg:
Various fixes & updates, mainly for TextCat & the train CLI. Bump to `3.0.0rc2` because of the new `spacy.TextCatEnsemble.v2`.

### Textcat
* Implemented `spacy.TextCatEnsemble.v2` which takes a `tok2vec` and a `linear_model` instead of building them internally 
   * kept `spacy.TextCatEnsemble.v1` in the code (for now)
   * updated docs with v2 but kept the old table in an `Accordion` element
* fixed the references to `cats_p`, `cats_r`, `cats_f` --> should be micro/macro
 
### Scorer
* Changed the way `ROCAUCScore.score` works if there's just one class. Before, it gave a `-Inf` which would cause the serialization of the `meta.json` to crash. I also don't think this is a proper value: it's not like the performance is infinitely bad, it's just not possible to measure the performance as there's only 1 gold label. So `None` would be more appropriate. I added a function `is_binary()` that can be called before calculating the `score`. This also helps in calculating the macro AUC score: you can take only the scores for those classes where you've already seen two labels.
* Updated the docs around the `score_cats` function which had gone out-of-date since the last refactor

### CLI
* where we `mkdir()` for the output folder, added `parents=True` to allow creating subdirs of a "training" folder. I found this to be useful in a project when running several training commands.
* added a custom error when `config["train"]["train_corpus"]` or `config["train"]["dev_corpus"]` is not a string, because we need these to be dot notations and it would crash a bit more obscurely otherwise. Though we probably want to fix the dot-notation validation more properly going forward, I think this is a good quick hack to avoid confusion if the config is wrong.

## Benchmarks

I'll add a spaCy project for this, but I've also benchmarked the textcat architectures, using similar settings as after the previous refactor https://github.com/explosion/spaCy/pull/5143. The results, reporting micro F:

| Dataset | Architecture | v2.2.4  | PR #5143 (3.0.0.dev4) | This PR (3.0.0rc2)
| ------- | ------------ | -------------------------- | -------------------- |  --- |
| IMDB    | BOW          | 78.5 %                       | 81.4 %                 | 84.2%
| IMDB    | CNN          | 79.1 %                       | 79.4 %                 | 81.4%
| IMDB    | DEFAULT      | 76.4 %                       | 77.1 %                 | 81.6%
| DBpedia | BOW          | 90.5 %                       | 90.7 %                 | 89.2%
| DBpedia | CNN          | 84.8 %                       | 86.4 %                 | 87.8%
| DBpedia | DEFAULT      | 89.7 %                       | 90.0 %                 |  90.7%

The evaluation is not exactly the same, because the training script changed quite a bit, but this looks good overall, I think :-) 

### Types of change
bug fix, enhancement, docs

## Checklist


honnibal:
All looks good! The figures are quite interesting, I wonder why v2.2 was so much worse. Could be lots of reasons; I wouldn't spend time digging into it.

svlandeg:
I think they're not exactly the same dev articles because of the way the data was previously loaded in the specific textcat train script, and how it's done now with the more generic train CLI. I agree it's not worth digging further into, but it's good to see there are no regressions.

PR #6268, opened 2020-10-17
Turkish tokenization improvements

DuyguA:
## Description
Improvements on the Turkish tokenization. Modifications considering
- ordinal numbers
- apostrophe
- refined rules for Turkish inflection paradigm, including abbreviation inflections and number inflections

### Types of change
- Improvement of the tokenization algorithm
- Tokenizer tests for special cases and general cases

## Checklist


spacy/lang/tr/tokenizer_exceptions.py
```diff
@@ -1,119 +1,184 @@
...
+_roman_ord = r"({rn})\.".format(rn=_roman_num)
+_time_exp = r"\d+(:\d+)*"
+
+_inflections = r"'[{al}]+".format(al=ALPHA_LOWER)
+_abbrev_inflected = r"[{a}]+\.'[{al}]+".format(a=ALPHA, al=ALPHA_LOWER)
+
+_nums = r"(({d})|({dn})|({te})|({on})|({n})|({ro})|({rn}))({inf})?".format(d=_date, dn=_dash_num, te=_time_exp, on=_ord_num, n=_num, ro=_roman_ord, rn=_roman_num, inf=_inflections)
 
 TOKENIZER_EXCEPTIONS = _exc
+TOKEN_MATCH = re.compile(r"^({u})|({abbr})|({n})$".format(u=URL_PATTERN, n=_nums, abbr=_abbrev_inflected)).match
```
adrianeboyd, line 184: If you want to support splitting trailing commas or parentheses from URLs, it would be better to have the URL pattern as `URL_MATCH` (you can just use the default `URL_MATCH` setting) and only have the other tokens as part of `TOKEN_MATCH`.

DuyguA: Thanks for pointing out, I made the corresponding change and removed the `URL_PATTERN` from the custom match.

adrianeboyd:
Thanks, this looks good!

As a note, as of v3, the tokenizer exceptions will only be able to set `ORTH` and `NORM`, so the lemmas will be removed in the future.

DuyguA:
Sure, looking forward to v3 :1st_place_medal:

PR #6279, opened 2020-10-20
Replace pytokenizations with internal alignment

adrianeboyd:
## Description

The internal alignment algorithm requires the concatenation of the two tokens lists to be identical with the following exceptions:

* Differing capitalization is allowed
* Differing whitespace within tokens containing non-whitespace is allowed
* Differing leading and trailing whitespace tokens are allowed as long there is a potential alignment on token boundaries
* Differing internal whitespace-only tokens are not permitted

## Checklist

### DIFF
--- a/spacy/training/align.py
+++ b/spacy/training/align.py
@@ -38,10 +38,11 @@ def get_alignments(A: List[str], B: List[str]) -> (List[List[int]], List[List[in
     tokens_a (List[str]): The first tokenization.
     tokens_b (List[str]): The second tokenization.
     RETURNS: (tuple): A 5-tuple consisting of the following information:
-      * a2b (List[int]): Mapping of indices in `tokens_a` to indices in `tokens_b`.
-        For instance, if `a2b[4] == [6]`, that means that `tokens_a[4]` aligns
-        to `tokens_b[6]`.
-      * b2a (List[int]): The same as `a2b`, but mapping the other direction.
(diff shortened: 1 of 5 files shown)


spacy/training/align.py
```diff
@@ -32,3 +30,78 @@ def _make_ragged(indices):
...
+    tokens_a = tokens_a[:trailing_ws_start_a] + trailing_ws_a
+    tokens_b = tokens_b[:trailing_ws_start_b] + trailing_ws_b
+    # Check that the strings align, allowing only differences in capitalization
+    if "".join(tokens_a).lower() != "".join(tokens_b).lower():
+        raise ValueError(Errors.E949)
+    # Store the set of aligned token indices per token
+    a2b = [set() for _ in range(len(tokens_a))]
+    b2a = [set() for _ in range(len(tokens_b))]
+    # Create iterators for token indices per character position
+    token_by_char_a = chain(*[[i] * len(x) for i, x in enumerate(tokens_a)])
```
honnibal, line 76: Is this fast enough? It looks terrible.

adrianeboyd: It does look terrible. I'll refactor it a bit.

adrianeboyd:
The earlier version was ~2-3x slower than pytokenizations and the refactored version is ~1.1x slower? It depends a bit on how misaligned the texts are and how long the tokens are.

adrianeboyd:
Let me look at the whitespace handling again...

PR #6286, opened 2020-10-22
Handle missing reference values in scorer

adrianeboyd:
## Description

Handle missing values in reference doc during scoring where it is possible to detect an unset state for the attribute. If no reference docs contain annotation, `None` is returned instead of a score. `spacy evaluate` displays `-` for missing scores and the missing scores are saved as `None`/`null`.

Attributes without unset states:

* `token.head`: relies on `token.dep` to recognize unset values
* `doc.cats`: unable to handle missing annotation

Additional changes:

* add optional `has_annotation` check to `score_spans` to replace `doc.sents` hack
* update `score_token_attr_per_feat` to handle missing and empty morph representations
* fix bug in `Doc.has_annotation` for normalization of `IS_SENT_START` vs. `SENT_START`

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/cli/evaluate.py
+++ b/spacy/cli/evaluate.py
@@ -98,10 +98,13 @@ def evaluate(
         if key in scores:
             if key == "cats_score":
                 metric = metric + " (" + scores.get("cats_score_desc", "unk") + ")"
-            if key == "speed":
-                results[metric] = f"{scores[key]:.0f}"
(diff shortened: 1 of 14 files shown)


honnibal:
This seems like a decent solution. I'm still interested in refactoring the scorer somewhat, but this is definitely better than the hack I had in there, and the return `None` solution is pretty good.

PR #6293, opened 2020-10-22
Replace pytokenizations with internal alignment

adrianeboyd:
## Description

Replace pytokenizations with internal alignment algorithm that is restricted to only allow differences in whitespace and capitalization.

* Rename `spacy.training.align` to `spacy.training.alignment` to contain the `Alignment` dataclass
* Implement `get_alignments` in `spacy.training.align`

### Types of change

Enhancement

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,6 @@ requires = [
     "murmurhash>=0.28.0,<1.1.0",
     "thinc>=8.0.0rc0,<8.1.0",
     "blis>=0.4.0,<0.8.0",
-    "pytokenizations",
     "pathy"
 ]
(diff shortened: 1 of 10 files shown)


adrianeboyd:
Now with an alignment algorithm that looks normal and handles differing whitespace much more generally. The resulting alignments are identical to `pytokenizations`.

The speed compared to `pytokenizations` for internal training corpora is ~0.8x-1.5x.

honnibal:
Thanks for the refactor. It looks good like this, just needs the small conflict to be fixed and then you can merge. We can always tweak the speed later if it needs it.

PR #6297, opened 2020-10-23
Setup / install / quickstart updates

adrianeboyd:
## Description

* Add `cuda110` to setup.cfg and quickstart dropdown
* Switch to `pip` for pip-only packages in conda quickstart instructions
* Update zh pkuseg install message with version range and conda
* Remove `zh` from `extras_require` because the default doesn't require additional packages

### Types of change

?

## Checklist


adrianeboyd:
Is there a good/informative way to kind of disable the `conda` option in the nightly quickstart? I think even with the updates the warning isn't prominent enough and it's kind of confusing that you can follow the instructions and end up with the wrong versions of everything.

Issue #6307, opened 2020-10-26
Remove questionable stopwords from default lists


alexcombessie:
Hi everyone,

I got myself into an interesting rabbit hole recently: stopwords. 

My context is that I need to bundle default stop words in my own application. The application being industry-agnostic, I am looking for generic stopwords which are not domain-specific. In addition to this, these stopwords should work well with several tokenizers: a simple whitespace tokenizer and the slightly better [regex tokenizer](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html#sklearn.feature_extraction.text.CountVectorizer) from scikit-learn.

 As I am a big fan of your work 💥  so my initial idea was to use spaCy for this, with some post-processing on my side to account for the different tokenizers. Hence, I started doing a small inspection of the stopwords bundled in spaCy 3.0-rc1. I only looked at French and English which I speak fluently.

```python
import spacy
print(spacy.blank("en").Defaults.stop_words)
print(spacy.blank("fr").Defaults.stop_words)
```

I have to admit I was surprised by inclusion of stopwords such as: `serious`, `well`, `never`, `very`, `please` in English. Excluding these stopwords would be a loss of valuable information for any human analysis or Machine Learning task, e.g. sentiment analysis.

I was even more surprised when I looked at French stop words. My compatriots will quickly understand the problem 🇫🇷  `zut`, `clic`, `clac`, `remarquable`, `sacrebleu`, `sapristi`, `bien`, `nul`, `pff`, `olé`, `paf`, `sein`, `plouf`, ... It looks like the words come from a Tintin cartoon.

Joke aside I realize this is not a simple topic, and that there is a balance between completeness, cross-domain applicability, and need for customization. For instance, scikit-learn is quite upfront in their [documentation](https://scikit-learn.org/stable/modules/feature_extraction.html#stop-words) about the issues with the English stopwords they are bundling. They linked an interesting academic paper on this topic: https://www.aclweb.org/anthology/W18-2502/

Perhaps a good idea to start with would be to remove the questionable stop words I spotted in English and French? Longer term, I would suggest some crowdsourcing with native speakers for each language.

Cheers,

Alex

adrianeboyd:
[...]

PR #6310, opened 2020-10-27
Remove questionable French stopwords

alexcombessie:
Linked to #6307

### DIFF
--- a/.github/contributors/alexcombessie.md
+++ b/.github/contributors/alexcombessie.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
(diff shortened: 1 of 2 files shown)


svlandeg (approved):
I'm not sure how exactly we want to move forward with these stop lists in the code base, but I agree with the general gist of this PR, which removes semantically meaningful French words from the stop list.

svlandeg:
Perhaps we could just version them in v3 and put them in a registry so that edits don't break people's code, but just create newer versions? And store the old versions somewhere, like we'll do with older versions of model implementations?

alexcombessie:
@svlandeg it's a good question. Personally, I would favor no versioning and a direct update of the default list. I am not sure it would break anyone's code. All I did was remove meaningful stopwords from the list, so assuming users are automatically filtering, it will just give them more tokens, i.e. more information.

adrianeboyd:
Let us discuss this and get back to you. There are also some more general issues related to versioning, lexical attributes, and serialization.

svlandeg:
I think "breaking code" was phrased badly on my part, but it would change the results of an NLP/ML model run on top of these stop lists. So there is a factor of reproducibility that we need to take into account.

> @svlandeg it's a good question. Personally, I would favor no versioning and a direct update of the default list. I am not sure it would break anyone's code. All I did was remove meaningful stopwords from the list, so assuming users are automatically filtering, it will just give them more tokens, i.e. more information.

honnibal:
Thanks! It's hard for us to judge stop-words, especially in languages we don't speak well, but the change looks reasonable :).

PR #6317, opened 2020-10-28
Identify final Matcher pattern node by quantifier

### LINKED ISSUE #6148: Matcher.remove crashes if {"OP": "!"} token is in any pattern of any rule
```
nlp = en_core_web_sm.load()
matcher = Matcher(nlp.vocab)

offending_pattern = [{'OP': '!'}, ]
matcher.add('test-rule', [offending_pattern ])
matcher.add('random-rule', [[{"LOWER": "hi"}]])
assert 'random-rule' in matcher
matcher.remove('random-rule')  # crashes here: "Process finished with exit code -1073741819 (0xC0000005)"
assert 'random-rule' not in matcher

```

When `{"OP": "!"}` is inside any pattern, removing __any__ rules will result in program crashing `"Process finished with exit code -1073741819 (0xC0000005)"`

## Your Environment

* Python Version Used: Python 3.6
* spaCy Version Used: Spacy v2.3.2
* Environment Information: nil

adrianeboyd:
## Description

Modify the internal pattern representation in `Matcher` patterns to identify the final ID state using a unique quantifier rather than a combination of other attributes.

It was insufficient to identify the final ID node based on an uninitialized `quantifier` (coincidentally being the same as the `ZERO`) with `nr_attr` as 0. (In addition, it was potentially bug-prone that `nr_attr` was set to 0 even though attrs were allocated.)

In the case of `{"OP": "!"}` (a valid, if pointless, pattern), `nr_attr` is 0 and the quantifier is `ZERO`, so the previous methods for incrementing to the ID node at the end of the pattern weren't able to distinguish the final ID node from the `{"OP": "!"}` pattern.

Fixes #6148.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pxd
+++ b/spacy/matcher/matcher.pxd
@@ -26,6 +26,7 @@ cdef enum quantifier_t:
     ZERO_PLUS
     ONE
     ONE_PLUS
+    FINAL_ID
 
 
 cdef struct AttrValueC:
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -558,7 +558,7 @@ cdef char get_is_match(PatternStateC state,
 
 
 cdef char get_is_final(PatternStateC state) nogil:
-    if state.pattern[1].nr_attr == 0 and state.pattern[1].attrs != NULL:
+    if state.pattern[1].quantifier == FINAL_ID:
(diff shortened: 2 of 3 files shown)


honnibal:
Nice work on this! Definitely correct.

PR #6345, opened 2020-11-04
Update stop_words.py in Portuguese (a,o,e)

Cristianasp:
Added three aditional stopwords: "a" and "o" that means "the", and "e" that means "and".

## Description
Added 3 stopwords to portuguese stopwords set.
"a" means "the" for female subjects
"o" means "the" for male objects
"e" means "and"

### Types of change
an enhancement

## Checklist

- [ X ] I have submitted the spaCy Contributor Agreement.
- [  ] I ran the tests, and all new and existing tests passed.
- [ X ] My changes don't require a change to the documentation, or if they do, I've added all required information.

### DIFF
--- a/.github/contributors/cristianasp.md
+++ b/.github/contributors/cristianasp.md
@@ -1,0 +1,107 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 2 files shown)


adrianeboyd:
(As a note, the test failure is not related to your PR.)

honnibal:
Thanks! It's hard for us to judge stop-words, especially in languages we don't speak, but the change looks reasonable :).

PR #6354, opened 2020-11-06
CI updates for python 3.5

adrianeboyd:
## Description

This is just a minor change separated from #6338 so we don't have python 3.5 test failures for now.

* Use `--prefer-binary` (`requirements.txt`) and `--binary-only :all:` (sdist)
* Delete all installed packages before testing the sdist install

### Types of change

Chore

## Checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -80,8 +80,8 @@ jobs:
       architecture: 'x64'
 
   - script: |
-      python -m pip install -U pip setuptools
-      pip install -r requirements.txt
+      python -m pip install -U setuptools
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Okay, that didn't work. I guess I didn't fix the `preshed` problems like I thought. Let's switch to `--prefer-binary` instead.

adrianeboyd:
I can't replicate any of the python 3.5 failures locally in linux (the sdist install not working with the `maps.cpp` error, `--prefer-binary` not working), so who knows what combination of `pip` / `setuptools` / (no) `wheel` / whatever is actually causing problems. But if this works, I think we can use it for now.

Issue #5051, opened 2020-02-22
Memory sharing in Python 3.8 for SpaCy models


EdwardJB:
This is more of a question than a feature request. But could SpaCy use [the new shared memory interface in Python 3.8](https://docs.python.org/3/library/multiprocessing.shared_memory.html) to share a single copy of its language model across multiple processes? Is it possible? If so, are there plans to support it?

grutt:
I'd also like to see if this is possible

tiwansh:
This would save so much memory.

forgetso:
I think something of this sort may already be happening. When I load sense2vec at a global level and then access the object from within multiprocess python instances, each instance is apparently using 6.2GB of memory. But the overall system resource sits at 9.6GB, swap use is negligible.

honnibal:
Most of the memory in spaCy models lives in numpy (or cupy) arrays, so it's possible this is happening for us automagically? The main data that isn't shared is the `Vocab`, but this should only be 100 or 200mb. The vocab is also written to frequently, so I don't think it's a good candidate for sharing across the processes.

If not, the main chunk of memory that you'll want to share is the vectors. The non-vectors memory in the models is really small. So all you need to do is reallocate `nlp.vocab.vectors.data` to use a numpy array backed by shared memory.

honnibal:
If someone's looking for project ideas, I think this could be a good blog post project? You could:

* Explain this new feature of Python 3.8 (which a lot of developers will find interesting)
* Implement a function which walks over the spaCy `Language` object and replaces all the arrays with ones backed by shared memory.
* Show the results in a benchmark

forgetso:
Turns out I was wrong about the memory being shared across multiprocesses...[](url)
![s2v_ram_usage](https://user-images.githubusercontent.com/7630866/84577942-7c58bf00-adb8-11ea-8220-f7604831df87.png)

I am using sense2vec pretty extensively. If I max out my RAM (very likely) then I may tackle this.

forgetso:
Looking at the data in the reddit model, the vectors file is 4.2Gb and is presumably causing the majority of RAM usage. Wouldn't `Vectors.data` need to be replaced in spacy/vectors.pyx with a shared memory? So modify load_vectors to point to a shared memory instead of a numpy array, e.g.

`       # change load vectors to load a shared memory if it exists
        def load_vectors(path):
            xp = Model.ops.xp
            if self.shared_memory_name is not None and self.shared_memory_shape is not None:
[...]

PR #6378, opened 2020-11-11
Allow binarized vectors to save memory

forgetso:
## Description
It is possible to binarize existing word embeddings so that they use 37.5 times less memory and perform a top-k query 30 times faster (See [Near-lossless Binarization of Word Embeddings](https://arxiv.org/pdf/1803.09065.pdf)). spaCy should be able to init models and perform most_similar calculations using binary vectors. This pull request enables spaCy to do this by checking the data type of vectors during init-model and by using an appropriate similarity function in most_similar.

### Types of change
Enhancement

## Checklist

### DIFF
--- a/.github/contributors/forgetso.md
+++ b/.github/contributors/forgetso.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
+documentation, or any other material posted or submitted by you to the project.
+
+2. With respect to any worldwide copyrights, or copyright applications and
+registrations, in your contribution:
+
+    * you hereby assign to us joint ownership, and to the extent that such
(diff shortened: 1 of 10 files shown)


honnibal:
This is a neat idea but unfortunately I don't think I want to accept this change. I'm worried about the implementation complexity that will result from this, as there are lots of other places in the code that will try to use the vectors and be surprised to get back non-float results. 

There's already many places where I think this implementation will fail, for instance in the the `doc.vector`, `token.vector` and `lexeme.vector` methods, and also in the `StaticVectors` layer which uses the vectors in the models.

forgetso:
This is a good point. The Sokal Michener similarity function (or similar) and dtypes would need to exist in:

- spacy/lexeme.pyx
- spacy/tests/util.py
- spacy/tests/vocab_vectors/test_similarity.py
- spacy/tokens/doc.pyx
- spacy/tokens/span.pyx
- spacy/tokens/token.pyx
- spacy/vectors.pyx
- spacy/vocab.pyx

I have modified the above to include a dynamic similarity function and also allowed `dtype` to be specified when `reset_vectors` is called in `vocab`. `dtype` is then passed through to `Vectors` and is set to `"f"` if it is `None`. `vectors_dtype` is set as a property in `vocab` also. The tests are passing and I added new tests into `test_similarity.py` for binarized vectors.

I'm not entirely comfortable with the implementation however as it involves repeating the same similarity functions in each of `doc`, `span`, `token`, and `vocab`. Is there a common place where these should be stored?

The only other mentions of `"f"` or `"float32"` seem to be in pipes.pyx which I _think_ is training. I don't think there is any need to change dtypes here as binarization should only be applied to pre-trained vectors.

`StaticVectors` seems to just load the vectors as is from spacy.

`VECTORS[key] = nlp.vocab.vectors.data`

It might work if you loaded binarized vectors into spacy and then passed them to thinc for ML purposes. However, it's probably not expected behaviour to be working with these kind of vectors for ML in the first place.

honnibal:
It's just that it's really the sort of change that I've found comes back to haunt us :(. Suddenly this data type can be two different things, and there's no way to abstract it without ruining the efficiency that motivates it. And then developers will easily forget that it can be two data types, because they're only using one. 

I think the problem is pretty fundamental so I don't think you should spend more time on the patch. It's a shame because I do really appreciate the effort to open the PR, and I can see how the paper is appealing. It's just that there are some techniques that would be worth it for a specific code-base and model, but can't really work in a more general framework like spaCy that is designed to accommodate many use-cases.

forgetso:
That's ok. I needed it for my own stuff anyway. I'm only using pre-trained models - not doing any training myself - to tag millions of records using my PC. As I have limited RAM and GPU RAM, it makes sense for me to work with binarized vectors as my GPU was not able to handle the Common Crawl model whilst also running n most_similar on thousands of vectors.

I also patched this into my fork of spaCy that stores the vectors data in shared memory. It turns out that this requires two separate classes of vectors (int and float) because of how Cython works. There are more details [here](https://github.com/explosion/spaCy/issues/5051) should you ever decide to go down this route.

PR #6391, opened 2020-11-16
Update source install instructions

adrianeboyd:
## Description

* Use `pip install` instead of `python setup.py install`
* For developers recommend:
  * `python setup.py build_ext --inplace -j N`
  * `python setup.py develop`

### Types of change

Docs.

## Checklist

### DIFF
--- a/website/docs/usage/index.md
+++ b/website/docs/usage/index.md
@@ -174,7 +174,13 @@ $ source .env/bin/activate                      # activate virtual env
 $ export PYTHONPATH=`pwd`                       # set Python path to spaCy dir
 $ pip install -r requirements.txt               # install all requirements
 $ python setup.py build_ext --inplace           # compile spaCy
-$ python setup.py install                       # install spaCy
+$ pip install .                                 # install spaCy
+```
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Hmm, I'm still not quite sure what to recommend here. I haven't used the `PYTHONPATH` option in the past and that probably conflicts or potentially does weird things in combination with `pip install` since `pip install` is going to do an isolated build with potentially different dependencies (numpy).

At this point, you could just say:

```
$ python -m pip install -U pip setuptools       # update pip
$ git clone https://github.com/explosion/spaCy  # clone spaCy
$ cd spaCy                                      # navigate into dir
$ python -m venv .env                           # create environment in .env
$ source .env/bin/activate
$ pip install .
```

This installs from source with an isolated build with `pip` using the `pyproject.toml` numpy pins.

And then the other options could be suggestions for developers rather than instructions for a plain source install.

I think for the editable `pip` install you have to recommend:

```
pip install -r requirements.txt
pip install --no-build-isolation --editable .
```

PR #6407, opened 2020-11-18
new package for Spacy universe

PilarHidalgo:
I make a mini-program, that given a text (input_data) and a dictionary created especially to identify SPAN's, related terms or words, this little program can extract terms in three formats: tuple (e.g ('label','term/list of terms')), just the term/terms or just the label of the SPAN.  Further details are in the comments within the code. Some examples of dictionaries and usage were developed into the repository. Even the package stills in progress is absolutely useful. 
Thanks in advance,
Cheers, Pilar.

### DIFF
--- a/.github/CONTRIBUTOR_AGREEMENT.md
+++ b/.github/CONTRIBUTOR_AGREEMENT.md
@@ -87,20 +87,20 @@ U.S. Federal law. Any choice of law rules will not apply.
 7. Please place an “x” on one of the applicable statement below. Please do NOT
 mark both statements:
 
-    * [ ] I am signing on behalf of myself as an individual and no other person
+    * [x] I am signing on behalf of myself as an individual and no other person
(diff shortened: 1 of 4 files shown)


svlandeg:
Hi! Thanks for submitting this PR, it's always great to see use-cases of spaCy :-) 

This looks like a very useful function for your specific project, but I'm not sure it's generic enough to be applicable in other projects. I don't think it's a good fit for the spaCy universe because it's not really a standalone package or extension with new functionality. So I'm afraid we won't be accepting the PR for that reason. 

FYI, just for future reference: please avoid editing the template files (eg `.github/CONTRIBUTOR_AGREEMENT.md` and `.github/PULL_REQUEST_TEMPLATE.md`) - these should remain as is. The file `.github/contributors/PilarHidalgo` looks good though - you can definitely keep that and include it in your next PR ;-)

Issue #6410, opened 2020-11-19
[Rule-Based Matching] Clarify that while attribute names are case-insensitive, attribute values are case-sensitive


jabortell:
On the [Rule-based Matching page](https://spacy.io/usage/rule-based-matching), there's a Q&A box titled "Does it matter if the attribute names are uppercase or lowercase?", and the answer is "No, it shouldn't." While using the [Matcher Explorer](https://explosion.ai/demos/matcher), I discovered after some confusion that while the attribute names (POS, TAG, DEP, etc) can be any case, the attribute values (VERB, nsubj) are case-sensitive. I think it'd be helpful to clarify that this is the expected behavior, or to make the attribute values case-insensitive as well, although that would of course leave the realm of a documentation issue entirely.

## Which page or section is this issue related to?
https://spacy.io/usage/rule-based-matching

svlandeg:
I'm a little unsure. I understand your confusion, but I wonder whether adding a sentence about "values" might actually confuse other readers? How would you phrase it? The problem is that that section is really specifically about the attribute names, so I'm not even sure about the terminology "values" - we don't really use it up until that point.

Perhaps it would help to add an example with the 4 possibilities - name & value lower-cased or upper-cased, and show the resulting matches? Perhaps some sentence with a wild card where matching on a lower-cased value gives a different result than the upper-cased value? But they each remain the same when changing the attribute from lower to upper?

jabortell:
I see that I could've been more specific about what I meant  the attribute "values", as some of them can be bools, regexes, etc, for which of course it wouldn't make sense to talk about case sensitivity. And to be clear, I wasn't talking about string-matching attributes (ORTH, etc), for which the user is using for its case-sensitivity, but about the attributes (at least POS and DEP) which take values ("VERB" and "nsubj") that are looked up in a dictionary. I think it makes sense to mention *somewhere* that the user can go wrong if the case doesn't match.

It makes sense in my brain that similar information should be close together, so likewise, it'd make sense to me that documentation about the case-sensitivity of values of POS/DEP/etc would go with the documentation about the case-sensitivity of the attribute names.

That's my thinking on it, but I'm not a contributor or anything like that, so feel free to disregard and move on :)

svlandeg:
> It makes sense in my brain that similar information should be close together, so likewise, it'd make sense to me that documentation about the case-sensitivity of values of POS/DEP/etc would go with the documentation about the case-sensitivity of the attribute names.

[...]

PR #6421, opened 2020-11-20
Update rule-based-matching.md

jabortell:
Update rule-based-matching.md

## Description

* Clarified case-sensitivity of dictionary-referencing attributes (POS/TAG/DEP/etc).
* Clarified "Type" column header to "Value Type"
* Addresses issue "[Rule-Based Matching] Clarify that while attribute names are case-insensitive, attribute values are case-sensitive" #6410

### Types of change
Changes documentation of the [Rule-Based Matching page](https://spacy.io/usage/rule-based-matching).

## Checklist
- [ ] I ran the tests, and all new and existing tests passed. -- N/A
- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information. -- N/A: changes are to the documentation

### DIFF
--- a/website/docs/usage/rule-based-matching.md
+++ b/website/docs/usage/rule-based-matching.md
@@ -157,20 +157,20 @@ The available token pattern keys correspond to a number of
 [`Token` attributes](/api/token#attributes). The supported attributes for
 rule-based matching are:
 
-| Attribute                              | Type    |  Description                                                                                           |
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Looks good to me! Thanks for taking care of the proper formatting as well :-)

Perhaps I would phrase it as "these attributes" instead of "this type of attribute", that just seems more clear/direct to me.

jabortell:
I agree, that is more clear. I've updated the wording in that last commit.

Issue #6401, opened 2020-11-17
[Spacy Nightly] init fill-config on textcat transformer produces pooling error


thomashacker:
## How to reproduce the behaviour

Hey, i am trying to get a training config for a transformer-based textcat pipeline.

The fix [#6371](https://github.com/explosion/spaCy/pull/6371) seems not to be working for me :( 
Using `init fill-config` on the transformer textcat config (CLI) still produced the pooling error:

✘ Can't construct config: calling registry function
(transformer_listener_tok2vec_v1) failed
spacy-transformers.TransformerListener.v1   transformer_listener_tok2vec_v1() missing 1 required positional argument: 'pooling'

---------------------------------------------------------------------------------------

I tried fixing this by manually adding the pooling attribute but then this error occurs:

✘ Can't construct config: calling registry function
(build_text_classifier) failed
spacy.TextCatEnsemble.v2   Cannot get dimension 'nO' for model 'transformer-listener>>trfs2arrays': value unset

{'model': {'@architectures': 'spacy.TextCatEnsemble.v2', 'nO': None, 'linear_model': {'@architectures': 'spacy.TextCatBOW.v1', 'exclusive_classes': False, 'ngram_size': 1, 'no_output_layer': False}, 'tok2vec': {'@architectures': 'spacy-transformers.TransformerListener.v1', 'grad_factor': 1.0, 'pooling': {'@layers': 'reduce_mean.v1'}}}}

---------------------------------------------------------------------------------------

## Info about spaCy
- **spaCy version:** 3.0.0rc2
- **Platform:** Linux-5.4.0-52-generic-x86_64-with-glibc2.27
- **Python version:** 3.8.0
- **Pipelines:** en_core_web_trf (3.0.0a0),  en_core_web_lg (3.0.0a0),

svlandeg:
Do you have the 3.0.0rc2 installed with `pip`, or did you compile from source? 

The fixes in PR #6371 haven't made it into a release yet - they'll be available from 3.0.0rc3 onwards. If you don't want to wait, you can either install the `develop` branch from source, or make the corresponding edits as you can see them in https://github.com/explosion/spaCy/pull/6371/files. 

Let us know if that works for you?

thomashacker:
I installed via pip. Tried fixing by making the edits and `init fill-config` works now!
But now the `train` produces this error:

Data validation error in 'para-attn'
X: <class 'thinc.types.Ragged'> Y: <class 'numpy.ndarray'>

[...]

PR #6431, opened 2020-11-23
Fix Transformer width in TextCatEnsemble

svlandeg:
Related to Issue #6401 and pre-requisite to merge https://github.com/explosion/spacy-transformers/pull/231.

## Description
We typically set the `width` of our tok2vec layer as the `nO` dimension of the `Tok2VecListener` by referring to `${components.tok2vec.model.encode.width}` in the config. This then ensures that other components can retrieve the width of the token vectors.

Unfortunately this doesn't work when using a `TransformerListener`, because the `transformer` component doesn't have an explicit `width` setting in the config. Instead, it internally [deduces](https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/layers/transformer_model.py#L69) its output dimension after loading in the model. I think we should keep that mechanism to avoid unnecessary burden on the user when they want to change the Transformer model. But it means we need to do some hacking around this.

`spacy.TextCatEnsemble.v2` needs the tok2vec width to set various dimensions on other layers in the model. When this architecture is called from the config, the `tok2vec` sublayer (a `TransformerListener`) doesn't know its `nO` yet, because that's only defined when [registering the listener](https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/pipeline_component.py#L127). And even then, we can't access it easily because the `TransformerListener` is [wrapped](https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/architectures.py#L34) in a chain layer. 

### Proposed solution:
- Register the reference to the actual `TransformerListener` explicitely - cf [PR 231 @ spacy-transformers](https://github.com/explosion/spacy-transformers/pull/231/files#diff-93cfc74a5aa23579f5438ff6fffc7aed548a18cffb4987a6d7a4adf6f21b3922)
- Create a convenience function `get_tok2vec_width` that searches for a relevant `tok2vec` layer, and perhaps a `listener` sublayer, to obtain the Tok2Vec width in a model - cf [this PR](https://github.com/explosion/spaCy/pull/6431/files#diff-2e7967d7187f4a8bdabd45ae09925fd4954d4fd25b7acf748569a627622757bc)
- Extend the `init` method of the `build_text_classifier_v2` function to set the appropriate Tok2Vec widths on the appropriate layers, and then proceed with the normal `chain` initialization function  - cf [this PR](https://github.com/explosion/spaCy/pull/6431/files#diff-ea413219be1be0d7bb0180f491a345720d83db031aa7e15b4696bf6a33e78a4bR102)
- An additional (unrelated) fix is required in the `residual` layer in Thinc to ensure everything can run: [PR 431 @ Thinc](https://github.com/explosion/thinc/pull/431/files#diff-02ab8f27233249022b0e683e303a759c8eee51184730c0a217f9b063d4434b80)
- When this PR and the one on Thinc are merged, this PR's new unit test should hopefully go green: [PR 231 @ spacy-transformers](https://github.com/explosion/spacy-transformers/pull/231/files#diff-5f9ee312df741049d954e022e19f31dc63ea38719144e7d8f45883eaf74d0ba6)

### Types of change
bug fix

## Checklist


honnibal:
Yeah this has been a problem, and I think the helper is a good solution. Let's merge this once the relevant Thinc and `spacy-transformers` RCs are up.

PR #6437, opened 2020-11-24
Fix tag map clobbering in spacy train

adrianeboyd:
## Description

Fix bug from #5768 where the tag map is clobbered if a custom tag map isn't provided.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
The CI doesn't actually test this functionality, so I'll merge this and fix the Travis problems in a separate PR.

PR #6441, opened 2020-11-24
Fix char embed for train CLI and tagger/parser

adrianeboyd:
## Description

* Fix char embed for train CLI and tagger/parser
* Mark `--use-chars` as experimental, CPU only

### Types of change

Bug fix.

## Checklist


adrianeboyd:
I think I'm submitting this more for reference than anything, since I'm not sure we want to do this. I'm not 100% sure it's right and without retraining a lot of models I wouldn't feel comfortable making this kind of change.

This seems like the minimum to get this working, but I'm not sure it's a good idea to give the impression that it works at all, since it's not supported on GPU. It mainly serves to remind me how bad the old config setup was.

I might instead suggest just removing the option from the docs and the train CLI?

honnibal:
I would favour removing the option, yeah. I'd rather have the last v2.x branch versions removing these unsupported odds and ends, because it'll be kind of an archival version and we should favour stability for it.

Issue #6318, opened 2020-10-28
Cannot use BiLSTM encoder with transition based NER parser. [spacy-nightly]


AndriyMulyar:
I get the following error when attempting to use the BiLSTM encoder with the NER transition based parser:
```bash
ℹ Using CPU

=========================== Initializing pipeline ===========================
✘ Can't construct config: calling registry function
(build_Tok2Vec_model) failed
spacy.Tok2Vec.v1   "Cannot get dimension 'nO' for model 'with_padded(with_padded(pytorch))'"

{'model': {'@architectures': 'spacy.Tok2Vec.v1', 'embed': {'@architectures': 'spacy.MultiHashEmbed.v1', 'width': 100, 'attrs': ['ORTH', 'SHAPE'], 'rows': [5000, 2500], 'include_static_vectors': 'True'}, 'encode': {'@architectures': 'spacy.TorchBiLSTMEncoder.v1', 'width': 100, 'depth': 2, 'dropout': 0.30000000000000004}}}
```
Could the pytorch BiLSTM wrapper encoder perhaps be missing a line like this (which appears in the CNN encoders):
https://github.com/explosion/spaCy/blob/dc816bba9d564ae572af28a17cbf0580ba11db5e/spacy/ml/models/tok2vec.py#L274

https://github.com/explosion/spaCy/blob/dc816bba9d564ae572af28a17cbf0580ba11db5e/spacy/ml/models/tok2vec.py#L302

Also the docstring is not accurate (seems to be copied from the CNNs)

My model components look like this:
```python
[components]

[components.tok2vec]
factory = "tok2vec"

[components.tok2vec.model]
@architectures = "spacy.Tok2Vec.v1"

[components.tok2vec.model.embed]
@architectures = "spacy.MultiHashEmbed.v1"
width = ${components.tok2vec.model.encode.width}
attrs = ["ORTH", "SHAPE"]
rows = [5000, 2500]
include_static_vectors = True

[components.tok2vec.model.encode]
@architectures = "spacy.TorchBiLSTMEncoder.v1"
width = 100
depth = 2
dropout = ${training.dropout}

[components.ner]
factory = "ner"

[components.ner.model]
@architectures = "spacy.TransitionBasedParser.v1"
state_type = "ner"
extra_state_tokens = false
hidden_width = 64
maxout_pieces = 2
use_upper = true
nO = null

[components.ner.model.tok2vec]
@architectures = "spacy.Tok2VecListener.v1"
[...]

PR #6442, opened 2020-11-24
Fix types of Tok2Vec encoding architectures

svlandeg:
This fix is required to resolve https://github.com/explosion/spaCy/issues/6318, together with https://github.com/explosion/thinc/pull/432.

## Description
I think the "encode" architectures were not returning the correct type, and they should incorporate `with_array` themselves instead of relying on `Tok2Vec` to do so. Technically this changes these models, so I've bumped the versions and kept the old ones in the same file (for now).

This was necessary to make `TorchBiLSTMEncoder.v1` function correctly, and I think it's better overall.

Also fixed the documentation of `TorchBiLSTMEncoder.v1` which was referring to wrong parameters.

### Types of change
bug fix + docs fix

## Checklist

### DIFF
--- a/spacy/cli/templates/quickstart_training.jinja
+++ b/spacy/cli/templates/quickstart_training.jinja
@@ -173,7 +173,7 @@ no_output_layer = false
 factory = "tok2vec"
 
 [components.tok2vec.model]
-@architectures = "spacy.Tok2Vec.v1"
+@architectures = "spacy.Tok2Vec.v2"
 
 [components.tok2vec.model.embed]
 @architectures = "spacy.MultiHashEmbed.v1"
@@ -188,7 +188,7 @@ rows = [5000, 2500]
 include_static_vectors = {{ "true" if optimize == "accuracy" else "false" }}
 
 [components.tok2vec.model.encode]
-@architectures = "spacy.MaxoutWindowEncoder.v1"
+@architectures = "spacy.MaxoutWindowEncoder.v2"
 width = {{ 96 if optimize == "efficiency" else 256 }}
 depth = {{ 4 if optimize == "efficiency" else 8 }}
 window_size = 1
--- a/spacy/ml/models/tok2vec.py
+++ b/spacy/ml/models/tok2vec.py
@@ -76,6 +76,7 @@ def build_hash_embed_cnn_tok2vec(
     )
 
 
+# TODO: archive
 @registry.architectures.register("spacy.Tok2Vec.v1")
 def build_Tok2Vec_model(
     embed: Model[List[Doc], List[Floats2d]],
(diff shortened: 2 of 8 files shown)


honnibal:
> they should incorporate with_array themselves instead of relying on Tok2Vec to do so. 

I think that isn't quite right, as I think it messes up the padding requirement. If you have a `List[Floats2d]` as input to the `MaxoutWindowEncoder`, you have to take care that you add padding to avoid it reading context from surrounding arrays. So it seems incorrect to add the `with_array` within the layer.

honnibal (changes requested):
I think moving the `with_array` layer inside probably isn't correct.

svlandeg:
Hm, ok, thanks, I'll have another look!

svlandeg:
@honnibal : Hm, I had another look, and I'm not sure I understand what's wrong in this PR :(

The original `v1` implementation of `MaxoutWindowEncoder` seems to be - to me - a `Model[Floats2D, Floats2D]` architecture, and not a `Model[List[Floats2d], List[Floats2d]]` architecture as the type suggests. This encoder does the cloning of the layers according to `depth` within its definition:
```
cnn = chain(expand_window, Maxout)
model = clone(residual(cnn), depth)
return model
```

It is fed as input to `Tok2Vec.v1` which does
```
tok2vec = chain(embed, with_array(encode, pad=receptive_field))
```

My proposal would be, for the `v2` implementation, to add `with_array` just at the end of `MaxoutWindowEncoder`, which means that it's still applied to the whole block and not just layer by layer, no?
```
cnn = chain(expand_window, Maxout)
model = clone(residual(cnn), depth)
return with_array(model, pad=receptive_field)
```
Then the `Tok2Vec.v2` will get an actual encoding architecture of type `Model[List[Floats2d], List[Floats2d]]` and can just do 
```
tok2vec = chain(embed, encode)
```

All this matters because the `BiLSTMEncoder` seems to be (currently) the only encoder that actually returns the correct type ` Model[List[Floats2d], List[Floats2d]]`:
```
return with_padded(PyTorchLSTM)
``` 
 and then if you use that encoder with `Tok2Vec.v1`, you get 
```
tok2vec = embed > with_array(with_padded(PyTorchLSTM))
```
while I think it should just be 
```
tok2vec = embed > with_padded(PyTorchLSTM)
```
?

If I'm entirely wrong - I still think we need to change the types of `spacy.MaxoutWindowEncoder.v1` & `spacy.MishWindowEncoder.v1` AND find another way to fix the combination of `spacy.TorchBiLSTMEncoder.v1`+ `spacy.Tok2Vec.v1` (cf https://github.com/explosion/spaCy/issues/6318)

honnibal:
Ah sorry, I see what you mean now. My mistake was that I thought the `MaxoutWindowEncoder` layer was doing one CNN, instead of the whole CNN block. It does make sense to give it the `with_array` responsibility.

PR #6451, opened 2020-11-25
pretrain architectures

svlandeg:
## Description
Adding the pretrain architectures to the registry instead of hard-coding them, so you can do this:

```
[pretraining.objective]
@architectures = "spacy.PretrainVectors.v1"
maxout_pieces = 3
hidden_size = 300
loss = "cosine"
```
or
```
[pretraining.objective]
@architectures = "spacy.PretrainCharacters.v1"
maxout_pieces = 3
hidden_size = 300
n_characters = 4
```
The loss function is stored in `model.attrs["loss"]`.

Not sure about the name & whether they should be in the `architectures` registry, because these two new functions are of type `Callable[["Vocab", "Model"], "Model"]` instead of just `Model`, as they need access to the `vocab` and `tok2vec` objects from the general pipeline.

### Types of change
enhancement

## Checklist


honnibal:
Well generally things in `@architectures` return `Model` instances, and they can take arbitrary arguments. So I think they look okay in that registry to me. Nice to have this task checked off!

PR #6455, opened 2020-11-26
Improve blis and numpy build dependencies

adrianeboyd:
## Description

* Add `blis` to `pyproject.toml` and `setup_requires`. (The correct requirements were in `thinc`, but not `spacy`.)
* Remove detailed `numpy` build constraints from `pyproject.toml`
* Test without `--only-binary :all:` to test this for more typical use
* Reduce number of builds in CI to speed things up

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,31 +41,31 @@ jobs:
       Python35Windows:
         imageName: 'vs2017-win2016'
         python.version: '3.5'
+# Test on one OS per python 3.6/3.7/3.8 to speed up CI
       Python36Linux:
         imageName: 'ubuntu-16.04'
         python.version: '3.6'
-      Python36Windows:
-        imageName: 'vs2017-win2016'
-        python.version: '3.6'
-      Python36Mac:
-        imageName: 'macos-10.14'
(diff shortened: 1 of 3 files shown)


adrianeboyd:
The difference between `preshed` and `murmurhash` is that `preshed` doesn't have `wheel` in `setup_requires`. I can't reproduce this problem locally with python 3.5 and it works in windows, so there must be some weird quirks here. But the solution is probably to go through all the dependencies and add `pyproject.toml` so this works better overall.

I also thought that these dependencies weren't including the `.cpp` files in the sdists anymore, but they are and I can't figure out the right settings to exclude them.

And as a few notes for anyone coming across this in the future:

* `python_version` isn't supported in `setup_requires` as far as I can tell
* `setup_requires` isn't consistent between `setup.py` and `setup.cfg`:
  * `setuptools` does install dependencies from `setup_requires` in `setup.py` (in `.eggs`)
  * `setuptools` doesn't install dependencies from `setup_requires` in `setup.cfg`, but `pip` does appear to check that they're satisfied (I think, anyway?)

PR #6474, opened 2020-12-01
multi-label textcat component

svlandeg:
## Description
* Make the `textcat` single-label
* Create a new `textcat_multilabel` component

### Types of change
enhancement

## Checklist
- [ ] TODO: test CLI functionality in more detail
- [ ] TODO: update projects that use `textcat` functionality


svlandeg:
It looks like this PR is uncovering a textcat reproducibility bug on CPU and non-Windows. I'll probably deal with this separately, because I don't think it's really related to the changes in this PR. Will investigate.

svlandeg:
I think this PR is done, but I'd love a detailed review. This will create breaking changes because the normal `textcat` from people's pipelines and configuration files will not be able to handle multi-label data anymore.

honnibal:
I should really have merged this sooner: I've come back to it a few times looking for something substantive to say, but I think the PR really does what we discussed, which yes has some trade-offs (including the backwards incompatibility), but I think that's a worthy trade-off.

We have the `validate_categories` call in the `TextCategorizer` to help with the backwards incompatibility, so it should be no problem.

PR #6503, opened 2020-12-04
Warn on empty POS for the rule-based lemmatizer

adrianeboyd:
## Description

Add a warning to the rule-based lemmatizer for any tokens without POS annotation.

### Types of change

Enhancement

## Checklist


adrianeboyd:
Potentially kind of a lot of warnings. I'm not sure whether a warning or logger.warn is better here.

ines:
Let's leave it as `logging.warn` for now so it doesn't clog up people's terminals. We can always make the first one more prominent (as in, a proper warning) if it's not explicit enough.

PR #6515, opened 2020-12-07
JapaneseTokenizer.pipe added

KoichiYasuoka:
## Description
For [spacymoji](https://spacy.io/universe/project/spacymoji)  with `Japanese()`.

### Types of change
To use JapaneseTokenizer.pipe as same as Tokenizer.pipe

## Checklist

### DIFF
--- a/spacy/lang/ja/__init__.py
+++ b/spacy/lang/ja/__init__.py
@@ -223,10 +223,6 @@ class JapaneseTokenizer(DummyTokenizer):
     def _set_config(self, config={}):
         self.split_mode = config.get("split_mode", None)
 
-    def pipe(self, texts, batch_size=1000, n_threads=-1):
-        for text in texts:
-            yield self(text)
-
     def to_bytes(self, **kwargs):
         serializers = OrderedDict(
             (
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Hi, thanks for this PR! Looking at the details, I think it would make more sense to add it to `DummyTokenizer` so it fixes this issue for several languages at once (Chinese, Japanese, Korean, Thai), but then you also have to handle an unimplemented `__call__` method in `DummyTokenizer`. I think you could add:

```
class DummyTokenizer(object):
    def __call__(self, text):
        raise NotImplementedError

    def pipe(self, texts, **kwargs):
        for text in texts:
            yield self(text)

    ...
```

If you wanted you could warn about `n_threads` as in `Tokenizer.pipe`, but I don't think this is a big deal. None of the tokenizers do any batching, and `nlp.pipe()` just calls `nlp.make_doc()`, so you don't notice the missing `tokenizer.pipe()` in most typical use cases.

adrianeboyd:
Thanks, this looks good!

KoichiYasuoka:
Thank you for suggestion, @adrianeboyd , and `Thai()` seems OK now with `spacymoji`.

adrianeboyd:
Don't worry about the python 3.5 test, that's unrelated (and should be fixed by #6522).

PR #6520, opened 2020-12-08
Make training.loop return nlp object and path

ines:
## Description

Make the training loop return the current `nlp` object (which includes the accuracy numbers via `nlp.meta`), plus the path the model was saved to.

This makes it easier to implement abstractions on top of the training loop, like Prodigy's train curve, which needs to collect the accuracies of multiple runs.

### Types of change
enhancement

## Checklist


honnibal:
Makes sense

PR #6531, opened 2020-12-09
Include custom code via spacy package command

ines:
## Description

- `spacy package` now takes an optional `--code` argument that can point to one or more files
- files will be copied into the package and imported in the `__init__.py` (all that's needed for registered functions)
- updated documentation accordingly

### Types of change
enhancement

## Checklist


honnibal:
I was sort of skeptical of this feature a while ago, but seeing it come together, it's really neat!

PR #6566, opened 2020-12-14
Update CITATION to point to Zenodo

rafguns:
## Description

This updates the CITATION file to point to the Zenodo DOI, per [this Twitter exchange](https://twitter.com/_inesmontani/status/1336776683202859008). It is pretty straightforward but there are still several aspects I was unsure of:
- The default citation of Zenodo points to a specific version. I assume you don't want that because it would require updating the CITATION for each new release. The DOI link will always point to the latest release.
- The default citation of Zenodo can look kind of weird, e.g. with a title like "explosion/spaCy: v2.3.4: Fix beam parser API (Version v2.3.4)". Since this CITATION is not tied to a specific version, it's not obvious what title to use.
- Who should be listed as author?

For the last two issues, I took inspiration from what was previously in the CITATION. Let me know what you would like to change, if anything.

### Types of change
enhancement

## Checklist


CITATION
```diff
@@ -1,6 +1,8 @@
-@unpublished{spacy2,
-    AUTHOR = {Honnibal, Matthew and Montani, Ines},
-    TITLE  = {{spaCy 2}: Natural language understanding with {B}loom embeddings, convolutional neural networks and incremental parsing},
-    YEAR   = {2017},
-    Note   = {To appear}
+@software{spacy,
+  author = {Honnibal, Matthew and Montani, Ines},
+  title = {{spaCy: Natural language understanding with Bloom embeddings, convolutional neural networks and incremental parsing}},
```
ines, line 3: If we're updating the citation anyways, we might as well update the title? I think this would be most consistent:

```suggestion
  title = {{spaCy: Industrial-strength Natural Language Processing in Python}},
```

CITATION
```diff
@@ -1,6 +1,8 @@
-@unpublished{spacy2,
-    AUTHOR = {Honnibal, Matthew and Montani, Ines},
-    TITLE  = {{spaCy 2}: Natural language understanding with {B}loom embeddings, convolutional neural networks and incremental parsing},
-    YEAR   = {2017},
-    Note   = {To appear}
+@software{spacy,
+  author = {Honnibal, Matthew and Montani, Ines},
```
ines, line 2:
```suggestion
  author = {Honnibal, Matthew and Montani, Ines and Van Landeghem, Sofie and Boyd, Adriane},
```

rafguns:
Thanks for updating and merging! Yesterday was my birthday, wich is why I was slow to respond😊🎂

PR #6572, opened 2020-12-15
Update PoS tags in 'tag_map.py'

PeterGilles:
Update of the PoS tag set. Now using the Stuttgart Tag Set (STTS).

## Description
The simplified PoS tag set for Luxembourgish has been modified to comply with the Stuttgart Tag Set (STTS). Like in the German tag_map, the more fine-grained PoS tags are mapped to Universal Commons. The file tag_map.py is now basically identical to the one for German.

### Types of change
enhancement

## Checklist

- [ X] My changes don't require a change to the documentation, or if they do, I've added all required information.


adrianeboyd:
Hi, since others may be relying on the existing tag map, I don't want to accept a PR that removes existing tags. We haven't been 100% consistent about this, especially for new languages, but in general, I'd rather not accept PRs that modify tag maps in v2 anymore since we're going to remove them all from the main library in v3. (Tag maps are really corpus-specific, not language-specific To be fair, so are the tokenizers, but the differences are usually much less drastic.)

You can supply a custom tag map in JSON to `spacy train` with the `--tag-map-path` option. The last section here shows how to export a tag map to JSON: https://spacy.io/usage/v2-3

PeterGilles:
I understand the rationale behind your answer. The original tag_map was developed by myself as a first draft to implement this language for spaCy. They are really rudimentary and loosely modelled along with the Universal Commons tags. These new ones I am proposing, are much more consistent and of course more language-specific. Thus, they would improve the model in the long run. 

Do I understand you correctly that tag maps will be removed in general from spaCy in the next version? If yes, an update would be unnecessary. If not, I would consider the new tag map as an improvement.

Issue #4313, opened 2019-09-20
(re-opening bug) thinc.extra.search.Beam.advance has assertion error with custom entity labels


sshegheva:
I couldn't figure out how to re-open an issue, so I am creating a new one with a link to the previous https://github.com/explosion/spaCy/issues/3047.

The behavior is still present. 

Basically, here is the line where the assertion is triggered.

https://github.com/explosion/thinc/blob/master/thinc/extra/search.pyx#L149

I can't quite figure out from the code what is the variable `size` trying to represent here. Perhaps, I can do some workaround, but I need to understand the root cause.    

Also, if there is a different recipe for getting the confidence on the extracted entities, I am happy to consider that (I haven't found an alternative)

svlandeg:
Hi @sshegheva , thanks for the report! I can't immediately reproduce the bug. Could you provide a minimal working (crashing) code snippet (with toy data), and could you also provide your system information by pasting the output of `python -m spacy info --markdown` ?

sshegheva:
Spacy's markdown info:


## Info about spaCy

* **spaCy version:** 2.1.8
* **Platform:** Darwin-18.6.0-x86_64-i386-64bit
* **Python version:** 3.6.7

Here is a minimal code snippet that reproduces the issue:

```
from spacy_lookup import Entity
import spacy
nlp = spacy.load("en_core_web_md")
entity = Entity(keywords_list=["gradient", "neural network"], label="ML")
text = "you have to be aware of a vanishing gradient when training a neural network"
nlp.add_pipe(entity, last=True)

docs = list(nlp.pipe([text], disable=["ner"]))
beams, _ = nlp.entity.beam_parse(docs,
                                      beam_width=3,
                                      beam_density=0.001)
entity_scores = defaultdict(float)
for doc, beam in zip(docs, beams):
    for score, ents in nlp.entity.moves.get_beam_parses(beam):
        for start, end, label in ents:
            ent = doc[start:end]
            if ent.text:  # do not write an empty entity
                entity_scores[(ent.text.lower(), label)] += score
```

And this is the assertion error:

```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-12-f8fac7fe1cd4> in <module>
      2 beams, _ = nlp.entity.beam_parse(docs,
      3                                       beam_width=3,
----> 4                                       beam_density=0.001)
      5 entity_scores = defaultdict(float)
      6 for doc, beam in zip(docs, beams):

[...]

PR #6575, opened 2020-12-15
Getting scores out of beam_ner

svlandeg:
Adding back the `get_beam_parses` method for the NER & added a `scored_ents` function. Not entirely sure this is how we want to do this?

TODO: parser (I'll do this in a separate PR)

## Description
Enabling the test for Issue https://github.com/explosion/spaCy/issues/4313 again - unfortunately currently broken again.

### Types of change
bug fix

## Checklist


spacy/tests/regression/test_issue4001-4500.py
```diff
@@ -288,35 +288,33 @@ def set_annotations(self, docs, scores):
...
+    }
+    ner = nlp.add_pipe("beam_ner", config=config)
     ner.add_label("SOME_LABEL")
-    ner.initialize(lambda: [])
+    nlp.initialize()
     # add a new label to the doc
     doc = nlp("What do you think about Apple ?")
     assert len(ner.labels) == 1
     assert "SOME_LABEL" in ner.labels
+    ner.add_label("MY_ORG")   # TODO: not sure if we want this to be necessary...
```
svlandeg, line 307: I think this shouldn't be necessary: when we add a previously trained NER to a pipeline with a matcher, you'd need to add the matcher's labels to the NER. Ideally, this test should work without this line.

honnibal: On the other hand the resize-while-running thing is pretty quirky. We have a separate discussion about that, so I'll leave this alone for here.

honnibal:
This looks right to me, go ahead and merge after that conflict is resolved 👍

spacy/tests/regression/test_issue4001-4500.py
svlandeg: Ok! I've created a quick ticket as a reminder: https://github.com/explosion/spaCy/issues/6681

PR #6592, opened 2020-12-18
Only generate dev examples once while training

adrianeboyd:
## Description

Instead of regenerating dev examples at each eval step, process the dev corpus more similarly to the train corpus and only generate the dev examples once at the start of training.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/training/loop.py
+++ b/spacy/training/loop.py
@@ -66,7 +66,7 @@ def train(
         nlp,
         optimizer,
         create_train_batches(train_corpus(nlp), batcher, T["max_epochs"]),
-        create_evaluation_callback(nlp, dev_corpus, score_weights),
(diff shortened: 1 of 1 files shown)


honnibal:
I guess this doesn't cause problems after the change to reset `example.x`?

honnibal (approved):

adrianeboyd:
Ah, hmm. I think it should be okay. Are there cases where we need to worry about preserving the state of `example.x` at all?

adrianeboyd:
The only concern is speed, and the difference probably isn't major, so we can just close this if you want.

honnibal:
I think I'd rather leave this alone: I don't think it's a major performance improvement, and it makes me nervous that resulting bugs would cause evaluation results to be off. We can revisit the question if we do find the performance gain to be worth it.

PR #6621, opened 2020-12-23
add new Romanian stopwords

lorenanda:
Added new words in stop_words.py - ro.

## Description
I added 15 new Romanian stop words (regional pronouns, adverbs of time and place), with and without special characters.

### Types of change
enhancement

## Checklist


honnibal:
Thanks! It's hard for us to judge stop-words, especially in foreign languages, but the change looks reasonable :).

PR #6629, opened 2020-12-24
spaCy v3 is not saving the best version in training loop

bratao:
## Description

During a training session using the latest spaCy v3 I realized that it was not saving the "best" version during training, it was saving only the latest. 
After a debug I found the error. During the training loop the check condition for is_best_checkpoint was wrong. I also realized that it was always saving the optimizer average, and also put a condition to check this. 

### Types of change
It is a bugfix to the training loop

## Checklist

### DIFF
--- a/spacy/training/loop.py
+++ b/spacy/training/loop.py
@@ -88,17 +88,17 @@ def train(
                 with nlp.select_pipes(disable=frozen_components):
                     update_meta(T, nlp, info)
                     if optimizer.averages:
-                        with self.nlp.use_params(optimizer.averages):
-                            self.nlp.to_disk(self.model_folder / DIR_MODEL_LAST)
+                        with nlp.use_params(optimizer.averages):
+                            nlp.to_disk(output_path / DIR_MODEL_LAST)
                     else:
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks!

The `optimizer.averages` check shouldn't be necessary: inside `model.use_params`, it should be a no-op if the parameters are `None`. This was to avoid the annoyance of conditioning on the context manager.

bratao:
@honnibal Got it. I remove the average check in the latest commit

spacy/training/loop.py
```diff
@@ -87,9 +87,12 @@ def train(
             if is_best_checkpoint is not None and output_path is not None:
                 with nlp.select_pipes(disable=frozen_components):
                     update_meta(T, nlp, info)
-                with nlp.use_params(optimizer.averages):
-                    nlp = before_to_disk(nlp)
```
adrianeboyd, line 91: The `before_to_disk` callback should be preserved in these changes.

bratao: Done. I did not called before_to_disk in the save below because this one will always be called

PR #6643, opened 2020-12-28
Update tokenizer_exceptions.py

bittlingmayer:
`EE.UU.` (meaing "USA") is already a tokenizer exception for Spanish.  Now we're adding `EE. UU.` with a space.

## Description
This variant is also common, for example in Wikipedia and as the output of Google Translate.

### Types of change
Enhancement

## Checklist

### DIFF
--- a/spacy/lang/es/tokenizer_exceptions.py
+++ b/spacy/lang/es/tokenizer_exceptions.py
@@ -22,7 +22,6 @@ for exc_data in [
     {ORTH: "Uds.", LEMMA: PRON_LEMMA, NORM: "ustedes"},
     {ORTH: "Vds.", LEMMA: PRON_LEMMA, NORM: "ustedes"},
     {ORTH: "vol.", NORM: "volúmen"},
-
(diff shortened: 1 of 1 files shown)


adrianeboyd:
This is a reasonable suggestion, but unfortunately this exception won't work with spacy v2 because spaces are not supported in exceptions. The tokenizer splits the text on whitespace and applies its patterns and rules to each space-separated substring separately.

Exceptions including spaces will be supported in spacy v3, so this PR could be made for the `develop` branch instead of the `master` branch. You may be able to edit the PR at the top to switch the base branch, but it's possible there will be some conflicts and it'd be easier to redo the PR from a new branch. (Also only `ORTH` and `NORM` are supported in the tokenizer exceptions for v3, too. `LEMMA` and other features are no longer permitted.)

bittlingmayer:
I've created a new PR as you suggested.  There were also some bugs in that branch.

https://github.com/explosion/spaCy/pull/6645

PR #6645, opened 2020-12-29
Update tokenizer_exceptions.py

bittlingmayer:
`EE.UU.` (meaning "USA") is already a tokenizer exception for Spanish.

Now we're adding `EE. UU.` with a space, and other variants.

## Description
This variant with the space is also common, for example in Wikipedia and as the output of Google Translate.

Spaces are not allowed in exceptions, so we just add the final subtoken.

I'm also fixing the missing commas, which is a nasty risk with Python.  (The strings get concatenated...)

### Types of change
Enhancement
Bug fix

## Checklist

### DIFF
--- a/spacy/lang/es/tokenizer_exceptions.py
+++ b/spacy/lang/es/tokenizer_exceptions.py
@@ -52,8 +52,8 @@ for orth in [
     "Dra.",
     "EE.UU.",
     "Ee.Uu.",
-    "UU.", # For "EE. UU."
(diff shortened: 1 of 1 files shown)


adrianeboyd:
It's good I had a look at this again. To improve the speed, we restricted the cases where the new exception handler is used to those containing special affix characters, but whitespace needs to be added to this list, too. (Underneath it's a `PhraseMatcher` that matches the special cases in the text.) So this PR is fine, but it's not going to work until I fix this bug.

PR #6651, opened 2020-12-30
Fixing CLI info command

svlandeg:
## Description
Fixes to `spacy info` command:
- Avoid issues when trying to parse a URL as a `Path`
- Fix code when path is not a package but a directory
- Avoid printing labels by setting a default "exclude" parameter. This makes the output pretty much unreadable otherwise if there's a `morphologizer`. If you do still want the labels, you can run with
```
spacy info es_core_news_lg --exclude ""
```
(this is not very intuitive but I think it's fine, as typically we'll probably be the ones to ask this to users if we really want the label information to solve a specific ticket)

### Types of change
bug fixes & enhancement

## Checklist

### DIFF
--- a/spacy/cli/info.py
+++ b/spacy/cli/info.py
@@ -19,7 +19,7 @@ def info_cli(
     # fmt: on
 ):
     """
-    Print info about spaCy installation. If a pipeline is speficied as an argument,
+    Print info about spaCy installation. If a pipeline is specified as an argument,
     print its meta information. Flag --markdown prints details in Markdown for easy
     copy-pasting to GitHub issues.
 
--- a/spacy/cli/info.py
+++ b/spacy/cli/info.py
@@ -1,10 +1,10 @@
-from typing import Optional, Dict, Any, Union
+from typing import Optional, Dict, Any, Union, List
 import platform
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


ines:
Yes, I think that's a good solution 👍

spacy/cli/info.py
```diff
@@ -15,20 +15,22 @@ def info_cli(
...
     DOCS: https://nightly.spacy.io/api/cli#info
     """
-    info(model, markdown=markdown, silent=silent)
+    exclude = string_to_list(exclude)
+    info(model, markdown=markdown, silent=silent, exclude=exclude)
 
 
 def info(
-    model: Optional[str] = None, *, markdown: bool = False, silent: bool = True
+    model: Optional[str] = None, *, markdown: bool = False, silent: bool = True, exclude: List[str]
```
KeshavG-lb, line 33: Why are we not setting default value here . Why we making `exclude` required field?

svlandeg: This function is mainly internals - the actual CLI `info` command invokes `def info_cli` which does in fact have a default, `"labels"`. For this internal command, you could argue that you really don't need any default values. It's better to have them all in one place - i.e. where you collect the user input.

PR #6653, opened 2020-12-30
Prevent 0-length mem alloc

svlandeg:
## Description
Adding sanity checks to avoid 0-length memory allocations, which would not result in NULLs, as we learnt [previously](https://github.com/explosion/spaCy/pull/4423).

Zero allocation occassionally happened in `create_morph_tag`. It didn't seem to do much damage, but best to avoid anyway.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/vocab.pyx
+++ b/spacy/vocab.pyx
@@ -164,7 +164,7 @@ cdef class Vocab:
         if len(string) < 3 or self.length < 10000:
             mem = self.mem
         cdef bint is_oov = mem is not self.mem
-        lex = <LexemeC*>mem.alloc(sizeof(LexemeC), 1)
+        lex = <LexemeC*>mem.alloc(1, sizeof(LexemeC))
         lex.orth = self.strings.add(string)
         lex.length = len(string)
         if self.vectors is not None:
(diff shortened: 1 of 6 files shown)


spacy/vocab.pyx
```diff
@@ -164,7 +164,7 @@ cdef class Vocab:
         if len(string) < 3 or self.length < 10000:
             mem = self.mem
         cdef bint is_oov = mem is not self.mem
-        lex = <LexemeC*>mem.alloc(sizeof(LexemeC), 1)
+        lex = <LexemeC*>mem.alloc(1, sizeof(LexemeC))
```
svlandeg, line 167: I'm a little baffled by this one, because this has been in the code for years. But the fix is right, yes? no? I'm so confused. I guess it just worked magically?

honnibal: I can never remember which way calloc's arguments work =/. It should be equivalent regardless, right? The code in `Pool.alloc` seems to only use the two multiplied together.

svlandeg: Ah! Yes ofcourse. That explains why it was never an issue :-)

PR #6668, opened 2021-01-04
Use special matcher for exceptions with spaces

adrianeboyd:
## Description

Use the special cases phrase matcher for exceptions that include space characters so that exceptions including spaces are supported.

### Types of change

Bug fix.

## Checklist


honnibal:
Good idea.

PR #6689, opened 2021-01-07
Sync missing and misaligned values in Tagger loss

adrianeboyd:
## Description

Use `None` for both missing and misaligned annotation in `Tagger.get_loss`, reverting to the default missing value in the loss
function.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/pipeline/tagger.pyx
+++ b/spacy/pipeline/tagger.pyx
@@ -256,8 +256,14 @@ class Tagger(TrainablePipe):
         DOCS: https://nightly.spacy.io/api/tagger#get_loss
         """
         validate_examples(examples, "Tagger.get_loss")
-        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False, missing_value="")
-        truths = [eg.get_aligned("TAG", as_string=True) for eg in examples]
+        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False)
+        # Convert empty tag "" to missing value None so that both misaligned
+        # tokens and tokens with missing annotation have the default missing
+        # value None.
+        truths = []
+        for eg in examples:
+            eg_truths = [tag if tag is not "" else None for tag in eg.get_aligned("TAG", as_string=True)]
+            truths.append(eg_truths)
         d_scores, loss = loss_func(scores, truths)
         if self.model.ops.xp.isnan(loss):
             raise ValueError(Errors.E910.format(name=self.name))
--- a/spacy/tests/pipeline/test_tagger.py
+++ b/spacy/tests/pipeline/test_tagger.py
@@ -37,7 +37,16 @@ TRAIN_DATA = [
 ]
 
 PARTIAL_DATA = [
+    # partial annotation
     ("I like green eggs", {"tags": ["", "V", "J", ""]}),
+    # misaligned partial annotation
+    (
+        "He hates green eggs",
+        {
+            "words": ["He", "hate", "s", "green", "eggs"],
+            "tags": ["", "V", "S", "J", ""],
+        },
+    ),
 ]
 
 


spacy/pipeline/tagger.pyx
```diff
@@ -256,8 +256,14 @@ class Tagger(TrainablePipe):
...
-        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False, missing_value="")
-        truths = [eg.get_aligned("TAG", as_string=True) for eg in examples]
+        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False)
+        # Convert empty tag "" to missing value None so that both misaligned
+        # tokens and tokens with missing annotation have the default missing
+        # value None.
+        truths = []
+        for eg in examples:
+            eg_truths = [tag if tag is not "" else None for tag in eg.get_aligned("TAG", as_string=True)]
+            truths.append(eg_truths)
```
svlandeg, line 266: I would specifically put `missing_value=None` when constructing `SequenceCategoricalCrossentropy`, for clarity but also to avoid future mismatches if the default would ever change in Thinc

svlandeg (commented):
Good catch on the misaligned use-case. I wonder whether we can't make this consistent internally? Can't we just use `""` for misaligned values (genuinely asking, there's probably a good reason why we didn't)?

adrianeboyd:
My initial thought is that we should keep it separate in `get_aligned` because there's the possibility that an empty label is a valid label, which should be treated differently from misaligned annotation.

The morphologizer is a bit more of a headache here and I need to look at it, too, because you can have both empty and unset morph values (empty is `"_"` internally), and I don't think it currently handles the unset values correctly, just the misaligned ones. (You have to go to a lot of extra effort to unset morph annotation in the first place, so it's rare that both POS+morph are unset in normal use because morph is nearly always set.)

spacy/pipeline/tagger.pyx
adrianeboyd: I can't imagine what value it would get changed to in the future and this seems like overkill to me? (We're not going to change the default ourselves without noticing that it breaks stuff?)

svlandeg: Specifically defining a parameter that you're counting on to have a certain value, doesn't feel like overkill to me...

adrianeboyd: But the value is coming from `Example` not `Tagger`, so it feels like it's just more places where this value is specified and things could go wrong?

svlandeg: It's the same place though: your list comprehension specifically states `else None`, and it's this `None` value that you're counting on to be interpreted as `missing_value`.

svlandeg: And yes there are too many additional dependencies on different types of "missing" values across the code base - hence my initial remark on making this more consistent internally ;-)

adrianeboyd: I think we've pretty consistently used `None` as a missing value in places like this, but now we have to interpret missing values from `Doc` instead of `GoldParse`, which doesn't have the same nice clean `None` defaults. The `Scorer` has the default missing values as anything in `[None, 0, ""]`, but it's still tricky around the features that can't be unset (head, morph (in the python API), cats) or where there are some weird edge cases (sentence boundaries).

adrianeboyd:
Valid empty labels aren't the issue (and may be something we want to disallow by fiat).

For the morphologizer, where you have to handle annotation from two attributes that may not be aligned in the exact same way, it is helpful for have the distinction between misaligned and unset in the output of `Example.get_aligned`. Then misaligned annotation on either attribute can be treated as a missing label, but if only one of the two happens to be unset, you still want to use the annotation from the other as the label. (I think, anyway?)

spacy/pipeline/tagger.pyx
svlandeg: Sure, but we're mixing up two conversations. Just from a software engineering perspective, this code decides to put `None` as missing value (instead of `""`) so I still say we should make it equally explicit to `SequenceCategoricalCrossentropy`, to ensure these two are always synced within these few lines of code. In my view it doesn't add a place where this could go wrong - it removes one. That's the reason why `SequenceCategoricalCrossentropy` has that parameter in the first place.

adrianeboyd: Then I think it would be better to define something like `Example.MISSING_VALUE` and use that across the board instead of adding more literal `None` values that you have to keep in sync.

honnibal: One thing to remember is that we have to score missing and misaligned values differently, so there are some bits of code where we want to know the difference.

For instance, in the NER evaluation our scoring was slightly wrong for a long time, because we were ignoring gold-standard entities where we had tokenization misalignments, because they looked the same as missing values.

honnibal: The change here looks right to me

svlandeg: Agreed that the change is right, my point was more about programming preferences. Marking as resolved in order not to hold up the PR!

honnibal: Ugh, right, I think I meant to write that on the top-level instead.

PR #6691, opened 2021-01-07
Train on partial parsing information

svlandeg:
## Description
It used to be possible to specify missing information for the `heads` annotations in the `GoldParse` object, e.g. `GoldParse(doc, heads=[None, 1, 1], deps=[None, "ROOT", "R"])`. With the (otherwise awesome) refactor to `Example` and storing all information on the `Doc` level, we lost that option as any token head should be an integer.

The proposal here is to store a missing `head` value of e.g. token `i=6` as `-1` (absolute head) or `-7` (relative head). I'm not sure I like it, but I don't see a better option. 

I also added a specific string `MISSING_LABEL` for a missing dep label.

When running `projectivize`, we ensure that the missing data remains unchanged.

### Types of change
enhancement

## Checklist


adrianeboyd:
In other places (`has_annotation`, the scorer) we had decided to rely on the presence of a `dep` tag to determine whether the token is parsed. I think we should do the same thing here rather than using an out-of-bounds head index. The missing dep label should be `0`/`""`.

svlandeg:
We could do that. Then I would consider changing the `token.head` property accordingly to return `None` when the `dep` tag is missing to be more internally consistent.

But how would that then interact with the `projectivize` function? I worry that we may end up training on data that looks like it's set, while it wasn't.

honnibal:
I think it's probably a good idea to have an explicit missing value, but let's make it consistent and use something like `MAX_INT`. (i.e. 2**31-1 which is 2,147,483,647)?

I'm nervous about changing the type of `token.head` to be `Optional[Token]` though. Is it actually better than the alternatives of either raising an error or just returning `self`?

adrianeboyd:
Whatever we do here for `head` should probably be done for `morph`, too, which has similar problems. You can't currently tell whether a morph is unset or empty in python. You can unset it in python with `token.set_morph`, but that's it.

I also think it would be too breaking to have `token.head` be `Optional[Token]`. I think you could have `get_` or `has_` plus `set_` methods that allow access to the underlying data as `Optional[Token]` and `Optional[MorphAnalysis]`, maybe?

svlandeg:
Ok, I've removed the more "controversial" parts of this PR and want to focus on two main points:
1. Being able to provide `None` values as dep/head annotations, as before. This feels like the most intuitive way of defining missing information from a user-point of view. See `PARTIAL_DATA` and `test_incomplete_data` for an example.
2. Ensuring the internal code base stores & retrieves missing parse information consistently.

I spend a bit of time going back and forth between the several options discussed here, and in the end I agree that having an explicit missing head value is a total pain. 

> In other places (`has_annotation`, the scorer) we had decided to rely on the presence of a `dep` tag to determine whether the token is parsed. The missing dep label should be `0`/`""`.

I think we can keep this convention. I've made it explicit by defining `cdef int MISSING_DEP = 0` in `token.pxd`. While writing some more unit tests, I discovered a few places where the code base is currently inconsistent though:
- The `sent_start` property in `set_children_from_heads` would be defined based on the token head being 0, but wouldn't check the dep annotation. This means that missing deps/heads would result in artificial sentence starts, which feels like a bad default. This is (still) fine for the first token in the sentence, ofcourse.
- `example.get_aligned_parse` would change token heads that were previously `0` and belonging to a missing dep. 
- We had unit tests that only defined heads and no deps. I've fixed those by adding `deps = ["dep"] * len(heads)`.

To facilitate all this, there are two new Cython functions `missing_dep` & `missing_head` and corresponding Python functions `has_head` and `has_dep`. I kind of like the similarity to the `has_dim` and `get_dim` stuff in Thinc. Though in this case we're returning the token itself when the head is missing. We could also raise an exception but I'm not sure that's better.

adrianeboyd:
There used to a be a dep check in `set_children_from_heads` but I took it out because if you're constructing a doc incrementally, the results were different depending on whether you set the deps or heads first, and this seemed wrong. This caused weird behavior even in the test suite.

Hmm, maybe the right thing is only to set sent starts if the span passed to `set_children_from_heads` is fully parsed? I'm not sure what sent starts in a partially parsed doc mean anyway?

svlandeg:
>  I'm not sure what sent starts in a partially parsed doc mean anyway?

Yea I was wondering about that too :|

adrianeboyd:
There used to be two different functions that adjusted children/heads, one in `set_children_from_heads` that was used after parsing and one in the `token.head` setter. When I replaced the `token.head` code with `set_children_from_heads`, some of the tests started failing even though the code should have been equivalent, but it turned it that it had to do with constructing docs token-by-token in a loop in the test suite and the `dep` check is what was causing the behavior to be different. Probably we've replaced a number of these cases with the `Doc` constructor, too, so they're no longer constructed incrementally.

honnibal:
Okay so the current code looks reasonable. I agree that having an explicit value for missing heads touches a lot of code for little tangible benefit.

Is this ready to merge now?

svlandeg:
yep!

PR #6704, opened 2021-01-11
Handle unset token.morph in Morphologizer

adrianeboyd:
## Description

Handle unset `token.morph` in `Morphologizer.initialize` and `Morphologizer.get_loss`. If both `token.morph` and `token.pos` are unset, treat the annotation as missing rather than empty.

### Types of change

Bug fix?

## Checklist


adrianeboyd:
This will need to be updated if we improve the head/morph missing API as discussed in #6691.

Issue #6708, opened 2021-01-11
Danish noun_chunk is always empty


KennethEnevoldsen:
## How to reproduce the behavior

As seen in the following minimal example the noun_chunks is always empty for Danish. This seems like a potential bug. The dataset used here contains >4000 samples.

```
import datasets  # huggingface's datasets

import spacy

# load sample danish dataset
ds = datasets.load_dataset("dane", split="train")

nlp = spacy.load('da_core_news_lg')
docs = nlp.pipe(ds["text"])

max_length_noun_chunks = max([len(list(doc.noun_chunks)) for doc in docs])
print(max_length_noun_chunks)
```

```
0
```

## Your Environment

* **spaCy version:** 2.3.2
* **Platform:** macOS-10.15.7-x86_64-i386-64bit
* **Python version:** 3.8.5
* **Models:** en

svlandeg:
Hi! The noun chunker for Danish has been [implemented](https://github.com/explosion/spaCy/pull/6246) very recently, and will be available in the next release. Currently, `doc.noun_chunks` returns an empty list when it's not (yet) implemented, which is kind of confusing.

adrianeboyd:
I think we could change this in v3 to do `raise NotImplementedError` instead? I remember mentioning it in https://github.com/explosion/spaCy/discussions/6337. Can anyone think of any particularly bad side effects of changing this?

svlandeg:
I agree that a `NotImplementedError` would make more sense!

KennethEnevoldsen:
Thanks for the quick response 🙏

Not implemted yet would indeed make a lot of sense. The only issue is that I probably wouldn't have gotten to know that the feature was indeed implemented.

KennethEnevoldsen:
For any other having this issue, installing the newest Spacy version from github solve the issue (as per previous comment).
```
pip uninstall spacy
pip install git+https://github.com/explosion/spaCy
```

PR #6711, opened 2021-01-11
raise NotImplementedError when noun_chunks iterator is not implemented

svlandeg:
## Description
When the `noun_chunks` iterator is not implemented for a given language, raise a `NotImplementedError` instead of returning an empty generator/list, which is confusing (see also https://github.com/explosion/spaCy/issues/6708 and https://github.com/explosion/spaCy/discussions/6337)

~~Also removing the `noun_chunks` function on the `Span`, which just delegated to the `span.doc.noun_chunks_iterator`. The docstring said it's documented at https://spacy.io/api/span#noun_chunks but it's not, so I doubt this was actually being used by anyone.~~
Also fixed the `noun_chunks` function on the `Span`, which didn't respect the proper boundaries (and wasn't documented)

### Types of change
UX enhancement

## Checklist


spacy/tokens/span.pyx
```diff
@@ -469,30 +469,6 @@ cdef class Span:
...
-        """
-        # Accumulate the result before beginning to iterate over it. This
-        # prevents the tokenisation from being changed out from under us
-        # during the iteration. The tricky thing here is that Span accepts
-        # its tokenisation changing, so it's okay once we have the Span
-        # objects. See Issue #375
-        spans = []
-        cdef attr_t label
-        if self.doc.noun_chunks_iterator is not None:
-            for start, end, label in self.doc.noun_chunks_iterator(self):
```
honnibal, line 492: Wait I think we do want this, although the implementation looks broken to me?

I think the idea is that it should get just the noun chunks within the span. But I can't see how this code would work, isn't the `start` and `end` here wrong?

The functionality is useful to harmonize the `Doc` and `Span` APIs, and to get all noun chunks within a sentence.

honnibal:
Agree with raising `NotImplementedError`, but I think we want to keep `span.noun_chunks`.

PR #6720, opened 2021-01-12
Expand initialize/training config validation

adrianeboyd:
## Description

Validate both `[initialize]` and `[training]` in `debug data` and `nlp.initialize()` with separate config validation error blocks that indicate which block of the config is being validated.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/training/initialize.py
+++ b/spacy/training/initialize.py
@@ -1,4 +1,5 @@
 from typing import Union, Dict, Optional, Any, List, IO, TYPE_CHECKING
+from pydantic import BaseModel
 from thinc.api import Config, fix_random_seed, set_gpu_allocator
 from thinc.api import ConfigValidationError
 from pathlib import Path
@@ -12,7 +13,7 @@ import tqdm
 from ..lookups import Lookups
 from ..vectors import Vectors
 from ..errors import Errors
-from ..schemas import ConfigSchemaTraining
+from ..schemas import ConfigSchemaInit, ConfigSchemaTraining
 from ..util import registry, load_model_from_config, resolve_dot_names, logger
 from ..util import load_model, ensure_path, OOV_RANK, DEFAULT_OOV_PROB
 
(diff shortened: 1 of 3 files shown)


adrianeboyd:
I'm not sure the details here are 100% what we want, but there seem to be a few places where validation is missing and error messages about subblocks are unclear about where the missing values belong.

spacy/training/initialize.py
```diff
@@ -23,6 +24,9 @@
 def init_nlp(config: Config, *, use_gpu: int = -1) -> "Language":
     raw_config = config
     config = raw_config.interpolate()
+    # Validate config before accessing values
+    _validate_config_block(config, "initialize", ConfigSchemaInit)
+    _validate_config_block(config, "training", ConfigSchemaTraining)
```
ines, line 29: I don't think we can really do this here because it means we're creating objects multiple times and the config is resolved twice. That's really unideal at runtime.

Instead, we should just wrap the calls to `registry.resolve` (maybe using a context manager,  similar to `cli._util.show_validation_error`). If we do this here and in `Language.initialize`, the result should be the same – the error may just be raised slightly later (which is an okay tradeo-ff IMO).

ines (changes requested):

adrianeboyd:
The concrete problem that I'm trying to solve is that parts of the training config are accessed before it's validated, leading to potentially confusing errors for users:

https://github.com/explosion/spaCy/blob/31a92b28aee804d915a757bd38e3693ca280c825/spacy/training/initialize.py#L26-L37

What would you want to do instead?

ines:
So you mean the `seed` and `gpu_allocator` values specifically? Those are the only ones that we have to access like this because they need to be set before anything else is created. Luckily, I'm pretty sure there'll be a limited number of these, so I think it's fine to just validate them separately and "manually", just like we do for the basic `[nlp]` settings at the very beginning of `Language.from_config`.

PR #6726, opened 2021-01-14
Add token_splitter component

adrianeboyd:
## Description

Add a `token_splitter` component for use with transformer pipelines. This component splits up long tokens like URLs into smaller tokens. This is particularly relevant for pretrained pipelines with `strided_spans`, since the user can't change the length of the span `window` and may not wish to preprocess the input texts.

The `token_splitter` splits tokens that are at least `min_length` chars long into smaller tokens of `split_length` chars size.

Notes:

* Since this is intended for use as the first component in a pipeline, the token splitter does not try to preserve any token annotation.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/pipeline/functions.py
+++ b/spacy/pipeline/functions.py
@@ -71,44 +71,33 @@ def merge_subtokens(doc: Doc, label: str = "subtok") -> Doc:
     return doc
 
 
-default_long_token_splitter_config = """
-[config]
-long_token_length = 25
-split_length = 10
-"""
-
-DEFAULT_LONG_TOKEN_SPLITTER_CONFIG = Config().from_str(
-    default_long_token_splitter_config
-)["config"]
-
-
 @Language.factory(
-    "long_token_splitter",
-    default_config=DEFAULT_LONG_TOKEN_SPLITTER_CONFIG,
+    "token_splitter",
+    default_config={"min_length": 25, "split_length": 10},
     retokenizes=True,
 )
 def make_long_token_splitter(
     nlp: Language,
     name: str,
     *,
-    long_token_length=0,
+    min_length=0,
     split_length=0,
 ):
-    return LongTokenSplitter(
-        long_token_length=long_token_length, split_length=split_length
+    return TokenSplitter(
+        min_length=min_length, split_length=split_length
     )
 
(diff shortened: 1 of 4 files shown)


adrianeboyd:
(I am not wedded to any of the naming choices here.)

spacy/pipeline/functions.py
```diff
@@ -65,3 +69,88 @@ def merge_subtokens(doc: Doc, label: str = "subtok") -> Doc:
...
+"""
+
+DEFAULT_LONG_TOKEN_SPLITTER_CONFIG = Config().from_str(
+    default_long_token_splitter_config
+)["config"]
+
+
+@Language.factory(
+    "long_token_splitter",
+    default_config=DEFAULT_LONG_TOKEN_SPLITTER_CONFIG,
```
ines, line 87:
```suggestion
    default_config={"long_token_length": 25, "split_length": 10},
```

Maybe we can just inline this as a dict here? We haven't been entirely consistent with this but I think if the default config doesn't contain nested sections and is short enough, we shouldn't have to do the whole `Config.from_str` roundtrip.

spacy/pipeline/functions.py
```diff
@@ -65,3 +69,88 @@ def merge_subtokens(doc: Doc, label: str = "subtok") -> Doc:
...
+split_length = 10
+"""
+
+DEFAULT_LONG_TOKEN_SPLITTER_CONFIG = Config().from_str(
+    default_long_token_splitter_config
+)["config"]
+
+
+@Language.factory(
+    "long_token_splitter",
```
ines, line 86: We could just call it "token splitter"? I mean, if you want to use it to split short tokens, why not 😅

```suggestion
    "token_splitter",
```

website/docs/usage/embeddings-transformers.md
```diff
@@ -481,6 +481,50 @@ custom learning rate for each component. Instead of a constant, you can also
 provide a schedule, allowing you to freeze the shared parameters at the start of
 training.
 
+### Managing transformer model max length limitations {#transformer-max-length}
```
ines, line 484: This is great! 💖

PR #6744, opened 2021-01-16
DependencyMatcher improvements (fix #6678)

### LINKED ISSUE #6678: DependencyMatcher has exponential time complexity
## TL;DR
Unfortunately the current implementation of DependencyMatcher can take very very long time to match large documents ^^

## How slow ?

This is best illustrated with an example. Consider this dependency tree: 

```
 [[
    {"RIGHT_ID": "is", "RIGHT_ATTRS": {"LEMMA": "be"}},
    {
        "LEFT_ID": "is",
        "REL_OP": ">",
        "RIGHT_ID": "subj",
        "RIGHT_ATTRS": {"DEP": "nsubj"},
    },
    {
        "LEFT_ID": "is",
        "REL_OP": ">",
        "RIGHT_ID": "adj",
        "RIGHT_ATTRS": {"POS": "ADJ",},
    },
]])
```
Now let's try to match from 1 to 100 repetitions of the sentence "The dress is beautiful. " and observe how match-time increases as we increase the size of the document. To compare them let's also benchmark the time taken to parse the document, the time taken by a simple matcher and the time taken by a simple matcher with an on_match callback parsing the dependency tree.

Here is the full benchmark script:

```
import spacy
from spacy.matcher import DependencyMatcher, Matcher
from time import time


text = "The dress is beautiful. "

dependency_matcher = DependencyMatcher(nlp.vocab)
dependency_matcher.add("test", [[
    {"RIGHT_ID": "is", "RIGHT_ATTRS": {"LEMMA": "be"}},
    {
        "LEFT_ID": "is",
        "REL_OP": ">",
        "RIGHT_ID": "subj",
        "RIGHT_ATTRS": {"DEP": "nsubj"},
    },
    {
        "LEFT_ID": "is",
        "REL_OP": ">",
        "RIGHT_ID": "adj",
        "RIGHT_ATTRS": {"POS": "ADJ",},
    },
]])

matcher = Matcher(nlp.vocab)
matcher.add("test", [[
    {"DEP": "nsubj"},
    {"LEMMA": "be"},
    {"POS": "ADJ"},
]])

def callback(matcher, doc, i, matches):
    _, start, _ = matches[i]
    # We are looking for a single token
    match = doc[start]
    subjs = []
    adjs = []
    for child in match.children:
        if child.dep_ == "nsubj":
            subjs.append(child)
        elif child.pos_ == "ADJ":
            adjs.append(child)

matcher_with_callback = Matcher(nlp.vocab)
matcher_with_callback.add("test", [[
    {"LEMMA": "be"},
]], on_match=callback)


def test(n):
    input_text = text*n

    # Benchmark pipeline
    start = time()
    doc = nlp(input_text)
    end = time()

    parse_time = end - start

    # Benchmark a simple matcher
    start = time()
    nb_matches_matcher = len(matcher(doc))
    end = time()

    matcher_time = end - start

    # Benchmark a matcher with on_match callback
    start = time()
    nb_matches_matcher_with_callack  = len(matcher_with_callback(doc))
    end = time()

    matcher_with_callback_time = end - start

    # Benchmark dependency matcher
    start = time()
    nb_matches_dependency_matcher = len(dependency_matcher(doc))
    end = time()

    dependency_matcher_time = end - start

    print(
        f"{n}, {parse_time}, {nb_matches_matcher}, {matcher_time}, "
        f"{nb_matches_matcher_with_callack}, {matcher_with_callback_time}, "
        f"{nb_matches_dependency_matcher}, {dependency_matcher_time}"
    )

for n in range(1,100):
    test(n)
```

Let's also enable cProfile to get some additional insights:

```
$ python -m cProfile -s cumtime bechmark.py
```

Raw output:
- timing: http://ix.io/2L0K
- cProfile: http://ix.io/2L0O

Here is a plot of the result:

image

And here is the same plot in log scale:

image

The dependency matcher (in yellow) takes almost 14 seconds to match 100 repetitions of the string "The dress is beautiful.", that's really a lot and makes it unfeasible to use DependencyMatcher to process large amount of data or use it for real-time applications.
What's worse is that processing time grows exponentially, which makes DependencyMatcher only usable for small documents.

## Why so slow ?
Here are the top entries of the cProfile report:

```
   126702727 function calls (101821149 primitive calls) in 331.038 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    907/1    0.004    0.000  331.045  331.045 {built-in method builtins.exec}
        1    0.000    0.000  331.045  331.045 test.py:1(<module>)
       99    0.025    0.000  330.145    3.335 test.py:52(test)
       99    0.021    0.000  327.809    3.311 dependencymatcher.pyx:242(__call__)
24835899/99  106.136    0.000  327.747    3.311 dependencymatcher.pyx:300(recurse)
 49338927    7.338    0.000  221.613    0.000 _asarray.py:14(asarray)
 49343983  214.279    0.000  214.279    0.000 {built-in method numpy.array}
       99    0.388    0.004    2.268    0.023 language.py:952(__call__)
      396    0.003    0.000    1.369    0.003 model.py:308(predict)
```
The benchmarks script took 331 seconds to run. About 327 seconds (99% of the time) was spent inside DependencyMatcher.recurse, 221 seconds of which were used just to build numpy arrays.

The reason why DependencyMatcher.recurse is so slow lies in this lines:

https://github.com/explosion/spaCy/blob/87562e470d0a38d1919c60941afbda5765f97ef7/spacy/matcher/dependencymatcher.pyx#L262-L264

In practice we are calling recurse with all possible combinations of matched nodes, no matter their location within the document.
For example, consider two repetitions of the above string:

"The dress is beautiful. The dress is beautiful."

Spacy will try correctly match the first "is" with the subject and adjective from the first sentence, however it will also try to match the first "is" with the subject and adjective from the second sentence, same for "is" in the second sentence and the tokens of the first sentence. This also apply to any other combinations. Multiply this for 100 repetitions of the sentence and you got 24835899 calls to recurse (which explains the 14 seconds runtime) :P 

## How can we fix this ?

A way to largely improve performance would be to abandon the recurse method and instead use an iterative method that:
- first, groups matches that belong to the same tree
- then, matches tokens from each group separately

I believe this should execute in linear time with respect to the document size.

In addition I would remove the conversion to numpy arrays if not necessary, as it turns out to be quite expensive.

Let me know what you think :) 

PS: thank you very much for developing and maintaining this awesome tool ^^

## Info about spaCy

- **spaCy version:** 3.0.0rc2
- **Platform:** Linux-5.10.3-arch1-1-x86_64-with-glibc2.2.5
- **Python version:** 3.8.7
- **Pipelines:** en_core_web_sm (3.0.0a0), en_core_web_trf (3.0.0a0)

werew:
## Description

This PR addresses #6678 and introduces big performance improvements to spaCy's `DependencyMatcher`.
It includes some changes to the tree-matching algorithm as well as some general cleaning/refactoring of the code with the goal of making it more maintainable and performant.

Changelog:
- Improved documentation
- Simplified and cleaned code
- Made code more conform to PEP8 styling standards
- Improvements to the matching algorithm:
  - Grouped tokens by common root before matching => fixed the exponential time complexity issue
  - Removed recursion => faster, lower memory usage
  - Memoized node+operator map => cleaner and faster as it avoids re-computation across different patters
  - Other minor improvements (e.g. use faster data structures)
- Added `root` property the `Token` class

I didn't add any site-doc for the new `root` property as I couldn't get it to build. Since this is a very small change, can some of the maintainers of this repo do it for me pls ? :) 

### Testing
All tests passing: `$  python setup.py build_ext --inplace &&  pytest spacy/tests/`

Performance benchmark using the same benchmark script as for #6678: `$ python bechmark.py`

Before - max matching time ~14 secs:
image

After fixes (without grouping tokens by common root) - max matching time ~0.6 secs:
image

After all fixes (logscale) - max matching time ~0.001 secs:
image

### Types of change

- Performance enhancement
- New feature (new `root` property for `Token`)

## Checklist


honnibal:
Wow, really appreciate this! Thanks for following through on it, the graphs are very convincing.

I'll look through the changes in more detail. I totally understand why it's necessary to refactor as part of the process of working through a piece of code like this, I do the same thing. We might want to reverse some of the small stuff like which things were moved to private though.

werew:
> Wow, really appreciate this! Thanks for following through on it, the graphs are very convincing.

Happy to contribute :)

> I'll look through the changes in more detail. I totally understand why it's necessary to refactor as part of the process of working through a piece of code like this, I do the same thing. We might want to reverse some of the small stuff like which things were moved to private though.

Thanks for looking into this PR. Regarding the things moved to private, I based that on the current website doc, hiding almost everything that is not mentioned there (furthermore those methods/attributes are inherently internal to the class so I thought it would be better to have them private). Ofc, feel free to change anything or ping me here for any question :)

adrianeboyd:
Yes, thanks for your work on this!

I'm a little concerned that a `token.root` property might be a little confusing for users, who might think it's something more like "root form" from the name. There's already `root` implemented for `Span`, so I think you could use this, without needing a new token property.

```python
root = doc[i].sent.root
```

werew:
> Yes, thanks for your work on this!
> 
> I'm a little concerned that a `token.root` property might be a little confusing for users, who might think it's something more like "root form" from the name. There's already `root` implemented for `Span`, so I think you could use this, without needing a new token property.
> 
> ```python
> root = doc[i].sent.root
> ```

I get your point and I think you are right, the name might be confusing. To be honest I am not too worry as long as the API doc is clear. We can also rename it to something else: `dep_root`, `tree_root`, `root_parent`, etc..

Regarding using `token.sent.root` the main reason why I opted for `token.root` is speed. In DependencyMatcher we do root lookups for every matched token (which are potentially a lot of tokens) so it is important that roots lookups are fast. 

`token.sent.root` will (under the hood) create two instances of `Span`, loop to the start of the sentence, loop to the end of the sentence, check user hooks at least three times and, finally, loop over all tokens in the span to find the root.

On the other hand `token.root` involves just a simple loop over the ancestors (which is also shorter than looping over every sentence-token).

spacy/tokens/token.pyx
```diff
@@ -633,6 +633,18 @@ cdef class Token:
             yield self.doc[head_ptr - (self.c - self.i)]
             i += 1
 
+    @property
+    def root(self):
```
honnibal, line 637: This property on the `Token` does something different from the `span.root`, so I wouldn't want it to be part of the public API.

We should just move the two lines here into the body of `is_ancestor`.

spacy/tokens/token.pyx
```diff
@@ -680,9 +692,8 @@ cdef class Token:
             # Find the widest l/r_edges of the roots of the two tokens involved
             # to limit the number of tokens for set_children_from_heads
             cdef Token self_root, new_head_root
-            self_ancestors = list(self.ancestors)
+            self_root = self.root
```
honnibal, line 695:
```suggestion
            self_root = ([self] + list(self.ancestors))[-1]
```

spacy/matcher/dependencymatcher.pyx
```diff
@@ -235,9 +264,24 @@ cdef class DependencyMatcher:
...
+        Processes the doc and groups all matches by their root and match id.
+        Returns a dict mapping each (root, match id) pair to the list of
+        tokens indices which are descendants of root and match the token
+        pattern identified by the given match id.
+
+        e.g. keys_to_position_maps[root_index][match_id] = [...]
+        """
+        keys_to_position_maps = defaultdict(lambda: defaultdict(list))
+        for match_id, start, _ in self._matcher(doc):
+            root = doc[start].root
```
honnibal, line 282:
```suggestion
            start = doc[start]
            root = ([start] + list(start.ancestors))[-1]
```

honnibal:
> I get your point and I think you are right, the name might be confusing. To be honest I am not too worry as long as the API doc is clear. We can also rename it to something else: dep_root, tree_root, root_parent, etc..

Just a quick word on how we weigh up the trade-offs around this sort of stuff:

It's very important that we keep the API of the core objects like `Doc`, `Token`, `Span`, `Language` etc really tight. We don't want to add extra public API to these objects, because every extra property is another item in the list the user has to look through to select the right thing. The names on the object need to make sense by themselves, not just with reference to the documentation page. Every extra name also makes future namings harder, because you have to think about the contrasts that are created. 

So, if we're just writing something to be used elsewhere in the library, we'll usually prefer to make a helper function instead of adding stuff onto the objects themselves. All the things on the objects themselves have to pull their weight in terms of being a useful set of things to users, not just internals.

adrianeboyd:
And the cython loops in `.sent.root` are faster than you'd think! I did a quick speed test because I was curious, and the `.sent.root` version is only ~10% slower. (I really wasn't sure what would be faster going in.)

honnibal:
If this is really a bottleneck we could just store the root indices for each token in an array? Something like:

```
roots = numpy.ndarray((len(doc),), dtype="i")
for i, token in enumerate(doc):
    ancestors = list(token)
    roots[i] = token.i if not ancestors else ancestors[-1].i
```

If for some reason we don't want to cache it, the Cython version should be fast:

cdef int get_root_index(const TokenC* tokens, int i, int length) nogil:
    cdef int n_loops = 0
    while i < length and tokens[i].head != 0:
        i += tokens[i].head
        n_loops += 1
        if n_loops >= length:
            # Guard against infinite loops from bad data.
            return -1
    return i
```

**Comment by werew** [comment]
@honnibal thanks for reviewing my changes :)

I've applied suggestions + removed the `.root` property. I also re-ran the benchmark and there are no noticeable differences.

> It's very important that we keep the API of the core objects like Doc, Token, Span, Language etc really tight...

Makes sense

> If this is really a bottleneck...

Not really, the difference is negligible. I think we can ship it like this.

> And the cython loops in .sent.root are faster than you'd think! I did a quick speed test because I was curious, and the .sent.root version is only ~10% slower (I really wasn't sure what would be faster going in.)

@adrianeboyd That's interesting, 10% is very low looking at the code I would have guessed much higher :)

File: spacy/matcher/dependencymatcher.pyx [hunk 4]
```diff
@@ -253,146 +298,137 @@ cdef class DependencyMatcher:
...
+            valid_right_matches = self._resolve_node_operator(doc, left_pos, relop)
+
+            # If the proposed right token is not within the valid ones, fail
+            if right_pos not in valid_right_matches:
+                return False
+
+        return True
+
+
+    @lru_cache(maxsize=None)
```
honnibal, line 366: I'm a bit nervous about this. I've had a couple of bad experiences with the code in functools, I think the module doesn't really consider all of the production considerations.

I have some specific worries here:

* `maxsize=None` is okay given that we ask for the cache to be cleared, and it's an internal method...But we can easily end up with a memory leak here if we mess it up later.
* Does the lru_cache decorator mess up pickling? Method decorators often have that problem.
* Every time we call into this, it will have to call `hash(doc)` in order to make the key to the cache. I think currently this just uses `id(doc)` so it's fast, but we've debated changing it and it shouldn't be necessary.

I suggest we remove the `lru_cache` decorator and just do the memoisation ourselves:

```python

def _resolve_node_operator(self, cache, doc, node, operator): 
    key = (node, operator)
    if key not in cache:
        cache[key] = [token.i for token in self._ops[operator](doc, node)]
    return cache[key]
```

This makes the cache a local variable of the scope that's calling into this function, rather than a global, so it shouldn't be possible to leak memory even if we screw up later.

werew: Updated

honnibal:
Great, that's everything covered! Thanks again for this.

I'd like to have better ways to ensure people get good value from their contributions. So uh, in case it helps you:

> Dear potential employer who might be reviewing this,
> Contributions of this quality are rare. This module is a difficult piece of functionality that several people had already looked at. You should hire this guy.

Happy to write something to this effect on Linkedin if you want (feel free to email me if I don't see a follow up on this PR)

werew:
> Great, that's everything covered! Thanks again for this.

Yay! Happy to know this has finally been integrated into spaCy :)  

> I'd like to have better ways to ensure people get good value from their contributions. So uh, in case it helps you:
> 
> > Dear potential employer who might be reviewing this,
> > Contributions of this quality are rare. This module is a difficult piece of functionality that several people had already looked at. You should hire this guy.
> 
> Happy to write something to this effect on Linkedin if you want (feel free to email me if I don't see a follow up on this PR)

Thank you, that's very kind of you. Of course I would never say no to a feedback on Linkedin, especially it is coming from you, so here is my profile: https://www.linkedin.com/in/luigiconiglio/ :P

PR #6745, opened 2021-01-16
Fix `spacy.util.minibatch` when the size iterator is finished

bryant1410:
Fix the `minibatch` function in `spacy.util` when the size `iterator` is finished.

## Description

This may seem unnecessary, and I'd be skeptical of the change. However, since Python 3.7, if a `StopIteration` exception is raised inside generators or coroutines, they are converted into runtime errors, thus not interpreted as the end of the iteration anymore but they're vanilla exceptions. See [this SO answer](https://stackoverflow.com/a/51701040/1165181) for more info.

So when an iterator is provided in `minibatch` (instead of an `int`), there are chances it's finished, and then this problem can happen. I added a default value to `next` so it isn't raised and the behavior is the expected one.

### Types of change

Bugfix.

## Checklist

### DIFF
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -513,7 +513,7 @@ def minibatch(items, size=8):
         size_ = size
     items = iter(items)
     while True:
-        batch_size = next(size_)
+        batch_size = next(size_, 0)  # StopIteration isn't handled in generators in Python >= 3.7.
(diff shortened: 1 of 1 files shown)


honnibal:
Fair point, and I think your solution is elegant. Thanks!

PR #6747, opened 2021-01-17
Add SpanCategorizer component

honnibal:
I wanted to write another new component to give the v3 config/model/component systems another test. It also fits in well with the new `SpanGroup` class.

The `SpanCategorizer` component provides a version of the idea discussed in https://github.com/explosion/spaCy/discussions/3961. The component takes a function that proposes `(start, end)` offsets in the documents as possible spans to be recognized. A model then assigns label probabilities to each span. Spans with probability above a given threshold are added to a span group of a specified key. The span group is also used to get training data.

~~I haven't tested this on an actual task yet, and I think the default model I've proposed probably won't work very well.~~

Edit: Now testing on an [Indonesian NER dataset](https://github.com/indobenchmark/indonlu/blob/master/dataset/nergrit_ner-grit/). It does work, ~~although the model I have is indeed not so good~~, and the accuracies are already looking okay.

| Component | NER P | NER R | NER F |
| ------------ | -----  | ------ | ------ |
| `SpanCategorizer` | 76.0 | 62.7 | 68.7 |
| `EntityRecognizer` | 71.1 | 66.8 | 68.9 |

 I'm sure an attention layer over the spans will do better than the current janky thing I'm doing. The `SpanCategorizer` is classifying all ngrams of length 1, 2, 3 or 4. The model also doesn't get to exploit the knowledge that the task has no overlapping or nested entities. I'm sure adding some Viterbi decoding would push its score up a bit higher. 

Another really nice thing about this component is that it predicts directly over the spans, so we'll finally have a component that can give meaningful entity confidences 🎉 . You'll be able to ask the model the probability that any arbitrary span of text is an entity -- you just have to propose the span in the suggester. 

### Types of change

Enhancement.

## Checklist


honnibal:
The design makes it a bit awkward to have a suggester that's built from the data. For instance, what if you want any span that was seen as an entity in the training data to be suggested? There's currently no good way to do that. It's also really hard to provide data files to the suggester.

The best workaround at the moment would be to move the suggestion into a different component, that sets the candidates into the `doc.spans` under a different key. The suggester function would then just read this key from the `doc.spans`.

Maybe we want to recommend that pattern instead of trying to build a more complex API for non-trivial suggester functions?

spacy/ml/models/spancat.py
```diff
@@ -0,0 +1,61 @@
...
+def build_mean_max_reducer(hidden_size: int):
+    """Reduce sequences by concatenating their mean and max pooled vectors,
+    and then combine the concatenated vectors with a hidden layer.
+    """
+    return chain(
+        # TODO: Add reduce_first and reduce_last when Thinc has them.
+        #concatenate(reduce_last(), reduce_first(), reduce_mean(), reduce_max()),
+        concatenate(reduce_mean(), reduce_max()),
+        Maxout(nO=hidden_size, normalize=True, dropout=0.0)
+    )
```
svlandeg, line 33: Neat!

spacy/ml/models/spancat.py
```diff
@@ -0,0 +1,61 @@
...
+#from thinc.api import reduce_first, reduce_last
+from thinc.api import zero_init, residual
+from thinc.types import Ragged, Floats2d
+from ...util import registry
+from ...tokens import Doc
+from ..extract_spans import extract_spans
+
+
+@registry.architectures.register("spacy.LinearLogistic.v1")
+def build_linear_logistic(nO=None, nI=None):
```
svlandeg, line 16: `-> Model[Floats2d, Floats2d]`

spacy/ml/models/spancat.py
```diff
@@ -0,0 +1,61 @@
...
+    followed by a logistic activation.
+    """
+    return chain(
+        Linear(nO=nO, nI=nI, init_W=zero_init),
+        Logistic()
+    )
+
+
+@registry.architectures.register("spacy.mean_max_reducer.v1")
+def build_mean_max_reducer(hidden_size: int):
```
svlandeg, line 24: `-> Model[Ragged, Floats2d]`

svlandeg (commented):
Few comments about the type annotations

spacy/ml/models/spancat.py
```diff
@@ -0,0 +1,59 @@
...
+    tok2vec (Model[List[Doc], List[Floats2d]]): The tok2vec model.
+    reducer (Model[Ragged, Floats2d]): The reducer model.
+    scorer (Model[Floats2d, Floats2d]): The scorer model.
+    """
+    model = chain(
+        with_getitem(0, chain(tok2vec, list2ragged())),
+        extract_spans(),
+        reducer,
+        scorer,
+    )
```
svlandeg, line 55: I don't understand this model definition :( You get a `Ragged` array containing the token vectors and lengths corresponding to the document lengths. Then what does `with_getitem` do with that `Ragged` object?

honnibal: Remember that we're passing in a tuple of `(docs, span_indices)`. The `with_getitem` wrapper lets us set aside the `span_indices`, and then make another tuple with them with the output. It applies the model to the first item of the tuple, so we can go from `Tuple[List[Doc], SpanIndices]` to `Tuple[TokVecs, SpanIndices]`.

It's basically doing this:

```
def predict(tok2vec, extract_spans, reducer, scorer, docs_spans):
    docs, span_indices = docs_spans
    tokvecs = tok2vec(docs)
    spanseqs = extract_spans((tokvecs, span_indices))
    spanvecs = reducer(spanseqs)
    span_scores = scorer(spanvecs)
```

spacy/ml/extract_spans.py
```diff
@@ -0,0 +1,60 @@
+from typing import Tuple
+from thinc.api import Model
+from thinc.types import Ragged, Ints1d
+
+
+def extract_spans() -> Model[Tuple[Ragged, Ragged], Tuple[Ragged, Ragged]]:
```
honnibal, line 6:
```suggestion
def extract_spans() -> Model[Tuple[Ragged, Ragged], Ragged]:
```

spacy/ml/extract_spans.py
```diff
@@ -0,0 +1,60 @@
...
+
+def init(model, X=None, Y=None):
+    pass
+
+
+def forward(
+    model: Model,
+    source_spans: Tuple[Ragged, Ragged],
+    is_train: bool
+) -> Ragged:
```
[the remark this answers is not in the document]
honnibal:
```suggestion
) -> Tuple[Ragged, Callable]:
```

spacy/ml/extract_spans.py
```diff
@@ -0,0 +1,60 @@
...
+        ops.scatter_add(dX.dataXd, indices, dY.dataXd)
+        return (dX, spans)
+
+    return Y, backprop_windows
+
+
+def _get_span_indices(ops, spans: Ragged, lengths: Ints1d) -> Ints1d:
+    """Construct a flat array that has the indices we want to extract from the
+    source data. For instance, if we want the spans (5, 9), (8, 10) the
+    indices will be [5, 6, 7, 8, 8, 9, 10].
```
honnibal, line 51:
```suggestion
    indices will be [5, 6, 7, 8, 8, 9].
```

spacy/ml/models/spancat.py
```diff
@@ -0,0 +1,59 @@
...
+# TODO: Uncomment these when next Thinc makes them available.
+# from thinc.api import reduce_first, reduce_last
+
+from ...util import registry
+from ...tokens import Doc
+from ..extract_spans import extract_spans
+
+
+@registry.architectures.register("spacy.LinearLogistic.v1")
+def build_linear_logistic(nO=None, nI=None):
```
honnibal, line 16:
```suggestion
def build_linear_logistic(nO=None, nI=None) -> Model[Floats2d, Floats2d]:
```

spacy/ml/models/spancat.py
```diff
@@ -0,0 +1,59 @@
...
+    """An output layer for multi-label classification. It uses a linear layer
+    followed by a logistic activation.
+    """
+    return chain(Linear(nO=nO, nI=nI, init_W=zero_init), Logistic())
+
+
+@registry.architectures.register("spacy.mean_max_reducer.v1")
+def build_mean_max_reducer(hidden_size: int):
```
honnibal, line 24:
```suggestion
def build_mean_max_reducer(hidden_size: int) -> Model[Ragged, Floats2d]:
```

website/docs/api/spancategorizer.md
```diff
@@ -0,0 +1,454 @@
...
+new: 3
+teaser: 'Pipeline component for labeling potentially overlapping spans of text'
+api_base_class: /api/pipe
+api_string_name: spancat
+api_trainable: true
+---
+
+A span categorizer consists of two parts: a [suggester function](#suggesters)
+that proposes candidate spans, which may or may not overlap, and a labeler model
+that predicts one or more labels for each span.
```
honnibal, line 14:
```suggestion
that predicts zero or more labels for each candidate.
```

ines:
> Maybe we want to recommend that pattern instead of trying to build a more complex API for non-trivial suggester functions?

I think that makes sense. We could have a suggester that reads from a given key in the `Doc.spans`. And then we could provide a few other built-in suggesters, like `suggest_noun_chunks` etc.

honnibal:
I've just hacked up the "span ruler" thing in my test project to try it out.

It's kind of awkward to do it this way =/. The `SpanRuler` won't run during `update()`, so the doc won't have the candidates set there. It also feels pretty unnatural in general.

Perhaps the `SpanCategorizer` can check for an `initialize()` method of its `suggester`, and call it if it's there? I suppose we could do the same with an `update` method.

spacy/pipeline/spancat.py
```diff
@@ -0,0 +1,400 @@
...
+        """
+        return tuple(self.cfg["labels"])
+
+    @property
+    def label_data(self) -> List[str]:
+        """RETURNS (List[str]): Information about the component's labels.
+
+        DOCS: https://nightly.spacy.io/api/spancategorizer#label_data
+        """
+        return self.labels
```
svlandeg, line 196:
```suggestion
        return list(self.labels)
```

spacy/tests/pipeline/test_spancat.py
```diff
@@ -0,0 +1,38 @@
...
+    ("I like London and Berlin.", {"entities": [(7, 13, "LOC"), (18, 24, "LOC")]}),
+]
+
+
+def make_get_examples(nlp, spans_key="spans"):
+    train_examples = []
+    for t in TRAIN_DATA:
+        eg = Example.from_dict(nlp.make_doc(t[0]), t[1])
+        eg.reference.spans[spans_key] = list(eg.reference.ents)
+        train_examples.append(eg)
```
svlandeg, line 17: We'll need a proper way to read these custom spans into `doc.spans` without passing through `doc.ents`.

And for documentation examples I wouldn't use the key `"spans"` ;-)

svlandeg (commented):
I think the implementation is pretty elegant, it's nice how concise & powerful the operations with `Ragged` are :-)

Just had a few more small comments

svlandeg:
> 
> 
> I've just hacked up the "span ruler" thing in my test project to try it out.
> 
> It's kind of awkward to do it this way =/. The `SpanRuler` won't run during `update()`, so the doc won't have the candidates set there. It also feels pretty unnatural in general.
> 
This feels similar to the trouble I had with the `EntityLinker`, where I went back and forth a bit as well.
Right now, the `EntityLinker`'s `update` function accesses `eg.reference.ents` directly to get the entities which it needs to predict a link for. I guess you'd need to do something similar?

honnibal:
> 
> 
> > I've just hacked up the "span ruler" thing in my test project to try it out.
> > It's kind of awkward to do it this way =/. The `SpanRuler` won't run during `update()`, so the doc won't have the candidates set there. It also feels pretty unnatural in general.
> 
> This feels similar to the trouble I had with the `EntityLinker`, where I went back and forth a bit as well.
> Right now, the `EntityLinker`'s `update` function accesses `eg.reference.ents` directly to get the entities which it needs to predict a link for. I guess you'd need to do something similar?

We need to predict for the incorrect candidates though =/. Hm.

spacy/pipeline/spancat.py
```diff
@@ -0,0 +1,402 @@
...
+            )
+            offset += indices.lengths[i]
+
+    def update(
+        self,
+        examples: Iterable[Example],
+        *,
+        drop: float = 0.0,
+        set_annotations: bool = False,
+        sgd: Optional[Optimizer] = None,
```
svlandeg, line 234:
```suggestion
        drop: float = 0.0,
        sgd: Optional[Optimizer] = None,
```

spacy/pipeline/spancat.py
```diff
@@ -0,0 +1,402 @@
...
+        losses: Optional[Dict[str, float]] = None,
+    ) -> Dict[str, float]:
+        """Learn from a batch of documents and gold-standard information,
+        updating the pipe's model. Delegates to predict and get_loss.
+
+        examples (Iterable[Example]): A batch of Example objects.
+        drop (float): The dropout rate.
+        set_annotations (bool): Whether or not to update the Example objects
+            with the predictions.
+        sgd (thinc.api.Optimizer): The optimizer.
```
svlandeg, line 244:
```suggestion
        drop (float): The dropout rate.
        sgd (thinc.api.Optimizer): The optimizer.
```

website/docs/api/spancategorizer.md
```diff
@@ -0,0 +1,454 @@
...
+> losses = spancat.update(examples, sgd=optimizer)
+> ```
+
+| Name              | Description                                                                                                                        |
+| ----------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| `examples`        | A batch of [`Example`](/api/example) objects to learn from. ~~Iterable[Example]~~                                                  |
+| _keyword-only_    |                                                                                                                                    |
+| `drop`            | The dropout rate. ~~float~~                                                                                                        |
+| `set_annotations` | Whether or not to update the `Example` objects with the predictions, delegating to [`set_annotations`](#set_annotations). ~~bool~~ |
+| `sgd`             | An optimizer. Will be created via [`create_optimizer`](#create_optimizer) if not set. ~~Optional[Optimizer]~~                      |
```
svlandeg, line 222:
```suggestion
| `drop`            | The dropout rate. ~~float~~                                                                                                        |
| `sgd`             | An optimizer. Will be created via [`create_optimizer`](#create_optimizer) if not set. ~~Optional[Optimizer]~~                      |
```

website/docs/api/spancategorizer.md
```diff
@@ -0,0 +1,454 @@
...
+| -------- | --------------------------------------------------------- |
+| `docs`   | The documents to modify. ~~Iterable[Doc]~~                |
+| `scores` | The scores to set, produced by `SpanCategorizer.predict`. |
+
+## SpanCategorizer.update {#update tag="method"}
+
+Learn from a batch of [`Example`](/api/example) objects containing the
+predictions and gold-standard annotations, and update the component's model.
+Delegates to [`predict`](/api/spancategorizer#predict) and
+[`get_loss`](/api/spancategorizer#get_loss).
```
svlandeg, line 206:
```suggestion
Delegates to [`predict`](/api/spancategorizer#predict),
[`get_loss`](/api/spancategorizer#get_loss) and 
[`set_annotations`](/api/spancategorizer#set_annotations).
```

svlandeg (commented):
Removing the `set_annotations` boolean flag

spacy/ml/extract_spans.py
```diff
@@ -0,0 +1,52 @@
+from typing import Tuple, Callable
+from thinc.api import Model
+from thinc.types import Ragged, Ints1d
+
+
+def extract_spans() -> Model[Tuple[Ragged, Ragged], Ragged]:
```
adrianeboyd, line 6: Register this layer?

svlandeg: For 'legacy' purposes later on, or do you want to have an explicit reference to an 'extract spans' layer in the config of `spacy.SpanCategorizer.v1` ? I'm not sure it would make much sense to overwrite this functionality?

adrianeboyd: No, I think our policy is that the active config should use the import, so this is just adding it so it's there later for legacy purposes.

svlandeg: okay. Actually, looking at this, now I'm thinking that `spacy.extract_spans.v1` and `spacy.mean_max_reducer.v1` should be `layers` instead of `architectures`? Not sure we have a crisp definition to distinguish between the two.

adrianeboyd: Oh, I hadn't looked at that in detail elsewhere. I thought this would be a `layer`, but you're probably right that the distinction isn't clear.

svlandeg: Yea I think we tend to use `architectures` if they are the top thing used by a `model`.

adrianeboyd: I think `layers` is the right choice here, just like for the other layers in `spacy/ml`?

spacy/pipeline/spancat.py
```diff
@@ -0,0 +1,402 @@
...
+        examples (Iterable[Example]): The examples to score.
+        RETURNS (Dict[str, Any]): The scores, produced by Scorer.score_cats.
+
+        DOCS: https://spacy.io/api/spancategorizer#score
+        """
+        validate_examples(examples, "SpanCategorizer.score")
+        self._validate_categories(examples)
+        kwargs = dict(kwargs)
+        kwargs.setdefault("attr", self.key)
+        kwargs.setdefault("attr_name", "spans")  # ensure the dict contains 'spans_f' etc
```
adrianeboyd, line 369: As a note after running the example project and look at the `spacy evaluate` results, I don't think using `"spans"` here is going to scale well. We can do things like add `spans_p/r/f` to what's shown in `spacy evaluate`, but you're going to run into problems as soon as you want to evaluate two `spancat` components, which I think might be a pretty common use case?

I'm not entirely sure what to do about the `spacy evaluate` view (possibly just parse the returned scores more flexibly rather than having a fixed list), but I think it'd be better not to use `spans` for everything, and we'd have to figure out how to handle the `default_score_weights`, too, I guess?

I think the current defaults are fine, but it's going to be confusing for users who change the key and have to adjust the score weights by hand. I'm not sure what the options are there...

svlandeg: Yea I pretty much had the same internal dialogue when implementing this. 

I kind of assumed that people wouldn't be training two `spancat` components at the same time. I'm trying to imagine the use-case. Obviously you can have a set of different labels in the same `doc.spans[key]` - what would be the use-case of annotating two distinct sets of spans (each with a distinct set of labels?) on the same texts?

> I think the current defaults are fine, but it's going to be confusing for users who change the key and have to adjust the score weights by hand.

This was my reasoning to add this `attr_name` setting to ensure that users don't have to change the score weights, and that the `default_score_weights` just work, for any value of `spans_key`.

adrianeboyd: I think if the default config works for one component that is basically good enough, and as far as I can tell, it would work for `"spans"`? Users with more complicated cases would have to manually adjust the score weights for now, but they'd also have to manually adjust the component name and spans key, too, so it doesn't seem too difficult?

I don't think you'd have two components with the same key, but I think it could be common to have two separate `spancat` components that have separate data + evaluations for two different tasks? If you do it this way, the second component clobbers the scores of the first one during the evaluation in the returned score dict, so you can't even see the scores of the first component.

I think we should not override `attr_name` here because otherwise there's not even a good workaround for evaluation two separate `spancat` components.

svlandeg: I'm not sure we're 100% on the same page, so just to elaborate. With the current implementation, if you change the `spans_key` to something else, like `"labeled_spans"`, the `kwargs.setdefault("attr_name", "spans")` actually ensures that works out-of-the box with the default score settings. The defaults are `"spans_r"` etc but before I made this edit, the score dictionary would have put `"labeled_spans_r"` and then the training wouldn't find the required accuracy setting, and it would show `0.0` by default for `"spans_r"`. So the `attr_name` thing is just to normalize the output column name in the final dictionary, to allow changing the key without further edits.

But yes, that approach wouldn't work for multiple `spancat` components, and yes, they should be able to manage the `kwargs` in those cases.

adrianeboyd: Ah, I see what you were thinking with the use of `kwargs`. But I think it wouldn't work during training, where you can't easily pass in those args, or in `spacy evaluate`. I think (at least initially until we have a better option?) having to adjust the score weights names in the config is a better option, so you can train and evaluate two components at the same time.

adrianeboyd: I think `spans_{self.key}` might be a better default, so that the fact that this is an evaluation related to `doc.spans` is always clear?

svlandeg: But then the column name during training will be huge... :/

adrianeboyd: Well, we can name the default key `morphologizer_loss` and all will be well.

My actual suggestion would be use a really short default key that's not `spans` (which is going to obviously be confusing, just like you said in the project), so something like `sc`. Then the default scores are `spans_sc_p/r/f` and I think that's manageable for the output? With multiple components you could have `spans_sc1_f` and `spans_sc1_f` and I think it would still be readable even with the truncated column headers?

I think then we could just provide `spans_{self.key}` as `attr` and not need to make any changes to the scoring method, given the custom `getattr` that manages the rest? I haven't tried that out yet, though.

adrianeboyd: Okay, the `default_score_weights` are a bit more of a hassle than I thought, since there's no way to parameterize them, and I think the `[training.score_weights]` overrides weren't all being added as expected. Here's what would be required for what I described above:

https://github.com/adrianeboyd/spaCy/commit/79a0403b02c73f4c30d7d40ab2e71234a69efe5f

And the output with `sc` and `sc2` looks like this:

```none
E    #       LOSS SPANCAT  LOSS SPANCAT2  SPANS_SC_F  SPANS_SC_P  SPANS_SC_R  SPANS_SC2_F  SPANS_SC2_P  SPANS_SC2_R  SCORE 
---  ------  ------------  -------------  ----------  ----------  ----------  -----------  -----------  -----------  ------
  0       0        184.50         184.50        2.31        1.50        4.98         2.73         1.90         4.83    0.03
  0     200       2658.05        2696.12        0.00        0.00        0.00         0.00         0.00         0.00    0.00
  0     400       1309.46        1350.80       30.41       83.67       18.58        33.45        74.09        21.60    0.32
```

(For the future: these two components that are being trained on the exact same data but the results aren't identical, so something may be non-deterministic in the model somewhere.)

spacy/ml/extract_spans.py
svlandeg: Yea, I agree, changed them to `layers`

spacy/pipeline/spancat.py
svlandeg: This suggestion is fine by me, and yea I would like to have a different default key than "spans". "sc" is fine I guess. If we want to support two spancat components in the same pipeline, we'll need to do something like this. Do you want to commit?

website/docs/api/spancategorizer.md
```diff
@@ -0,0 +1,453 @@
+---
+title: SpanCategorizer
+tag: class
+source: spacy/pipeline/spancat.py
+new: 3
```
adrianeboyd, line 5:
```suggestion
new: 3.1
```

website/docs/api/spancategorizer.md
```diff
@@ -0,0 +1,453 @@
+---
+title: SpanCategorizer
+tag: class
```
adrianeboyd, line 3:
```suggestion
tag: class,experimental
```

spacy/pipeline/spancat.py
adrianeboyd: I do think it's best to make sure the first version of this can handle separate scoring for multiple components.

Ideally, it'd be nice to parameterize the default score weights, but I don't think that's realistic for v3.1, so the `overrides` change looks like the easiest alternative.

However, I want to make sure I understand the side effects of the `overrides` issue. It's possible it will make the train output really ugly if you're just updating one component in a long pipeline and you'd end up with a lot of extra values shown in the table that wouldn't be there otherwise, but I need to test it out.

adrianeboyd: It looks like the scores for the frozen components are included in the output either way, so I don't think allowing the extra score overrides makes much of a UX difference. And this change would only potentially add some extra scores/columns to the output, but I really don't think it has the potential to break any existing functionality.

website/docs/api/scorer.md
```diff
@@ -149,6 +149,7 @@ Returns PRF scores for labeled or unlabeled spans.
 | `has_annotation` | Defaults to `None`. If provided, `has_annotation(doc)` should return whether a `Doc` has annotation for this `attr`. Docs without annotation are skipped for scoring purposes. ~~str~~      |
 | `labeled`        | Defaults to `True`. If set to `False`, two spans will be considered equal if their start and end match, irrespective of their label. ~~bool~~                                               |
 | `allow_overlap`  | Defaults to `False`. Whether or not to allow overlapping spans. If set to `False`, the alignment will automatically resolve conflicts. ~~bool~~                                             |
+| `attr_name`      | The name of the scored attribute as used in the scoring dictionary. Defaults to `None`, in which case the value of `attr` is used.  ~~Optional[str]~~                                       |
```
svlandeg, line 152:
```suggestion
```

adrianeboyd: Whoops! (I think this feature would be fine at some point if we really need it, but I don't think it's necessary now/here?)

svlandeg: Yep, no worries, if we can do without for now that's fine by me.

adrianeboyd:
I think this looks good now, with the experimental label. Anything else outstanding?

spacy/pipeline/spancat.py
```diff
@@ -0,0 +1,407 @@
...
+        if ops is None:
+            ops = get_current_ops()
+        spans = []
+        lengths = []
+        for doc in docs:
+            starts = ops.xp.arange(len(doc), dtype="i")
+            starts = starts.reshape((-1, 1))
+            length = 0
+            for size in sizes:
+                if size < len(doc):
```
adrianeboyd, line 63:
```suggestion
                if size <= len(doc):
```

Otherwise docs of length size have no ngrams suggested and `spans` can end up empty.

spacy/pipeline/spancat.py
```diff
@@ -0,0 +1,407 @@
...
+                    starts_size = starts[:-size]
+                    spans.append(ops.xp.hstack((starts_size, starts_size + size)))
+                    length += spans[-1].shape[0]
+                if spans:
+                    assert spans[-1].ndim == 2, spans[-1].shape
+            lengths.append(length)
+        output = Ragged(ops.xp.vstack(spans), ops.asarray(lengths, dtype="i"))
```
adrianeboyd, line 70:
```suggestion
        if len(spans) > 0:
            output = Ragged(ops.xp.vstack(spans), ops.asarray(lengths, dtype="i"))
        else:
            output = Ragged(ops.xp.zeros((0,)), ops.xp.zeros((0,), dtype="i"))
```

`vstack` fails for empty lists. Is there a better empty `Ragged` option?

adrianeboyd: This wrong, let me do a full proposal with tests.

spacy/pipeline/spancat.py
adrianeboyd: This is right, but would also be better with tests.

PR #6766, opened 2021-01-19
warn when frozen components break listener pattern

svlandeg:
## Description
- Change in `Tok2Vec` to store not only the listener `Model` instances, but also the downstream component names to which those listeners belong.
- Warn when trying to train a component which has a frozen tok2vec layer, or when (re)training a tok2vec layer which has a downstream frozen component listening to it.
- Added a few notes in the documentation

### Types of change
UX enhancement

## Checklist

## TODO
- [x] Update `Transformer` accordingly if we agree on this PR

### DIFF
--- a/spacy/pipeline/tok2vec.py
+++ b/spacy/pipeline/tok2vec.py
@@ -85,10 +85,10 @@ class Tok2Vec(TrainablePipe):
         self.listener_map[component_name].append(listener)
 
     def find_listeners(self, component) -> None:
-        """Walk over a model of a processing component, looking for layers that are Tok2vecListener
-        subclasses that have an upstream_name that matches this component.
-        Listeners can also set their upstream_name attribute to the wildcard
-        string '*' to match any `Tok2Vec`.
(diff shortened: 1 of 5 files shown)


spacy/pipeline/tok2vec.py
```diff
@@ -62,28 +62,44 @@ def __init__(self, vocab: Vocab, model: Model, name: str = "tok2vec") -> None:
...
+        return [m for c in self.listening_components for m in self.listener_map[c]]
+
+    @property
+    def listening_components(self) -> List[str]:
+        """RETURNS (List[str]): The downstream components listening to this
+        component. Usually internals.
+        """
+        return list(self.listener_map.keys())
+
+    def add_listener(self, listener: "Tok2VecListener", component: str) -> None:
```
honnibal, line 82: I think people would pass in the actual component if we call this only `component` (especially since we have a `component` argument in another method). Perhaps `component_name`?

honnibal:
I think this'll definitely help! And yeah I agree that the name is the best thing to keep track of.

PR #6767, opened 2021-01-19
Set annotations in update

svlandeg:
## Description
Follow-up to PR https://github.com/explosion/spaCy/pull/6765 - ensuring `component.update()` calls `component.set_annotations()` and removing the boolean flag, also from documentation. 

### Types of change
(Potentially breaking) change -> bumping to `3.0.0rc4` to avoid confusion with the recently released `3.0.0rc3`

## Checklist

## TODO
- [ ] Update `spacy-transformers` accordingly

### DIFF
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -193,18 +193,16 @@ class EntityLinker(TrainablePipe):
         self,
         examples: Iterable[Example],
         *,
-        set_annotations: bool = False,
         drop: float = 0.0,
         sgd: Optional[Optimizer] = None,
         losses: Optional[Dict[str, float]] = None,
     ) -> Dict[str, float]:
         """Learn from a batch of documents and gold-standard information,
-        updating the pipe's model. Delegates to predict and get_loss.
+        updating the pipe's model. Delegates to predict, get_loss and
+        set_annotations.
 
         examples (Iterable[Example]): A batch of Example objects.
         drop (float): The dropout rate.
-        set_annotations (bool): Whether or not to update the Example objects
-            with the predictions.
         sgd (thinc.api.Optimizer): The optimizer.
         losses (Dict[str, float]): Optional record of the loss during training.
             Updated using the component name as the key.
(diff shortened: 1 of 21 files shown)


spacy/pipeline/entity_linker.py
```diff
@@ -220,11 +218,13 @@ def update(
...
         validate_examples(examples, "EntityLinker.update")
         sentence_docs = []
-        docs = [eg.predicted for eg in examples]
-        if set_annotations:
-            # This seems simpler than other ways to get that exact output -- but
-            # it does run the model twice :(
-            predictions = self.model.predict(docs)
+        docs = []
+        for eg in examples:
+            eg.predicted.ents = eg.reference.ents
```
honnibal, line 223: Now that we're using `set_annotations` during the `update` flow, is it possible to learn from the predicted entities? We might still want to use the gold-standard entities as well as candidates, but we can at least learn with the predicted outputs?

spacy/pipeline/entity_linker.py
```diff
@@ -220,11 +218,13 @@ def update(
...
+            docs.append(eg.predicted)
+        # This seems simpler than other ways to get that exact output -- but
+        # it does run the model twice :(
+        predictions = self.predict(docs)
```
honnibal, line 227: Hmm, I see why this ends up convenient based on the code in the `predict()` method, but I hope we can refactor that in a separate PR at some point. I might have a look today.

We want to get to a point where we're doing something like this:

```python

sentence_encodings, bp_context = self.model.begin_update(sentence_docs)
predictions = self._encodings2predictions(sentence_encodings, sentence_docs)
self.set_annotations(predictions)
```

This would also get us to where we're calling the model on the full batch of data, instead of per-sentence as we are currently.

svlandeg: Yea I know, I was looking at this code when making the PR and thinking this should really be cleaned up further :| (but wanted this PR to go out first)

spacy/pipeline/entity_linker.py
svlandeg: I'm not sure how it would work to learn from the predicted entities? We only have KB IDs for the gold-standard entities, and so we'd only be able to learn from those predicted ones that match gold-standard ones. 

I guess the only advantage might be that you'd focus the EL on those entities that the NER can actually predict, but at the same time a drop in recall for the NER will create a gap in knowledge for the EL?

honnibal: I think having incorrect entities can be important when figuring out thresholds, because you want to return `NIL` for them. Anyway, a question for another PR :).

PR #6771, opened 2021-01-20
support IS_SENT_START in PhraseMatcher

### LINKED ISSUE #6770: Phrasematcher; OverflowError: Python int too large to convert to C long

Greetings;

I am trying to use Spacy PhraseMatcher with token attr `'IS_SENT_START'`. I guess like other attributes like '`IS_PUNCT`' it should be possible to use the is_sent_start attr yet I got the following error.

```
`matcher = PhraseMatcher(nlp.vocab, attr = "IS_SENT_START")

  File "<ipython-input-74-c8e7d50d6aec>", line 1, in <module>
    matcher = PhraseMatcher(nlp.vocab, attr = "IS_SENT_START")

  File "phrasematcher.pyx", line 63, in spacy.matcher.phrasematcher.PhraseMatcher.__init__

OverflowError: Python int too large to convert to C long`
```

Here is the sample code;

```
import spacy
from spacy.matcher import PhraseMatcher
sub_list = ['however', 'although', 'on the other hand', 'nonetheless']
patterns = [nlp.make_doc(text) for text in sub_list]
matcher = PhraseMatcher(nlp.vocab, attr = 'IS_SENT_START')
```

Why so? I am trying to see if phrases in list are sent_start = `True or False`. No problem with Matcher yet it is not working with Phrasematcher. Can you help me with the solution or a work around? 

## Your Environment

* Operating System: macOS Catalina
* Python Version Used: 3.7.9
* spaCy Version Used: 2.0
* Environment Information: Anaconda-Spyder

svlandeg:
Fixes #6770

## Description
support `IS_SENT_START` in `PhraseMatcher`, just like we do in `Matcher`

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/matcher/phrasematcher.pyx
+++ b/spacy/matcher/phrasematcher.pyx
@@ -8,6 +8,7 @@ from preshed.maps cimport map_init, map_set, map_get, map_clear, map_iter
 
 import warnings
 
+from ..attrs import IDS
 from ..attrs cimport ORTH, POS, TAG, DEP, LEMMA
 from ..structs cimport TokenC
 from ..tokens.token cimport Token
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I don't think catching the `OverflowError` is the best way to handle this. We shouldn't be assigning an arbitrary value from `strings` to an `attr_id_t` value. We should check whether it's in `attr.IDS` instead, I think? It's possible the `Matcher` doesn't handle this well, either, but that weird values don't get past the validation in the first place.

svlandeg:
To be honest I think we can even do without the `try-except`. We already check the attribute in `TOKEN_PATTERN_SCHEMA`, we basically just hit a freak case with `IS_SENT_START` which is valid in that schema but is not in the enum.

honnibal:
Do we perhaps want the `spacy.strings.get_string_id` function?

adrianeboyd:
No, that would still potentially return a value you can't assign to `attr_id_t`.

adrianeboyd:
The `Matcher` has `attr = IDS.get(attr)`.

PR #6793, opened 2021-01-22
Spacy Cli info method causing backward compatibility issues 

KeshavG-lb:
## Description
Spacy Cli info method was causing backward compatibility issues with all other third 
party consumers.
Reference issue : https://github.com/explosion/spaCy/pull/6793#issue-560075424

### Types of change
Fixed backward compatibility by setting default value to exclude in info
method.

## Checklist


svlandeg (approved):
Thanks for the PR! Agreed that it's better to have a non-mutable parameter instead of an empty list directly.

PR #6794, opened 2021-01-22
Implement overwrite param for all custom lemmatizers

adrianeboyd:
## Description

Implement overwrite param for all custom lemmatizers

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/lang/bn/__init__.py
+++ b/spacy/lang/bn/__init__.py
@@ -23,11 +23,13 @@ class Bengali(Language):
 @Bengali.factory(
     "lemmatizer",
     assigns=["token.lemma"],
-    default_config={"model": None, "mode": "rule"},
+    default_config={"model": None, "mode": "rule", "overwrite": False},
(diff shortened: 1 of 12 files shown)


honnibal:
Does this break existing configs?

adrianeboyd:
Yes, it does break existing configs.

I'm not sure how I ended up with it missing in all these configs in the first place, but maybe I added `overwrite` to the main class after the first big update PR. I kind of feel like this should be harder to do wrong, like it could fill the derived class config from the parent config or warn you somehow when the child class doesn't provide all the same parameters, but it's probably too complicated, and of course there's no reason you couldn't have a child class with fewer parameters.

And another thing that needs to be added to all the language-specific lemmatizers is the warning when `pos` isn't set, which is only in the main rule-based lemmatizer method at this point. I don't see any general way to handle that, we just have to keep track of propagating changes like that ourselves.

And the `Language.factory` cases are a little different from most other places, since they're not versioned.

PR #6800, opened 2021-01-24
Avoid assuming encode nO dimension is preset in tok2vec

honnibal:
We've been making a number of fixes to the `tok2vec` layer definitions, including @sofie's fix to make other encoders work. Those problems masked this issue, which I hit when doing the spancat testing.

## Checklist


svlandeg:
Makes sense.

Looks like the wrong Sofie got pinged here though ;-)
Unfortunately I haven't been so lucky to be able to use my own name as github handle 😁

PR #6802, opened 2021-01-24
Add Kyrgyz language (ky) support

jumasheff:
This PR adds initial support for the [Kyrgyz language](https://en.wikipedia.org/wiki/Kyrgyz_language) (Cyrillic alphabet).

## Description
Adds tokenization exceptions, stopwords, like_num (for cardinal numbers), translated examples, and tokenizer tests for the Kyrgyz language (ky).

Kyrgyz letters are covered by the Russian and Tatar alphabet, that's why they were not included in `char_classes.py`.
I borrowed `punctuation.py` from Tatar, but had to add `r"(?<=[0-9])-(?=[{a}])".format(a=ALPHA),` to support proper tokenization of words like `39–мектеп` while dealing with multi-word phrases like `кажы-кужу` as single entities.

### Types of change
New language support.

## Checklist


svlandeg (approved):
Thanks so much for this contribution, this looks great! Awesome to be having Kyrgyz support soon :-)

I reviewed your code in detail and it all looked perfectly fine to me. Really appreciate the many tests, too.

I'll just leave this open without merging for a bit, in case any of my colleagues still want to review as well.

ines:
Agree, thanks for so many tests, this is awesome! 💯

PR #6804, opened 2021-01-25
Avoid Doc owning LexemeC data and fix parser set_annotations during update

honnibal:
Fix two training bugs that were introduced by #6765 and #6767 .

1. In #6765, we're calling into `Doc.copy()`. It turns out this method had a bug. It didn't account for the fact that the `Doc` object was allowed to own the `LexemeC` struct. If the original `Doc` becomes freed before the copy, we would orphan the `LexemeC` struct. I think the idea of having the `Doc` own the `LexemeC` sometimes is just bad. I introduced this as a memory efficiency for long-running processes, but it's a weird hack that cuts against the original design, and I think the codebase is better off without it. To minimise the changes I've only made a change in the `Vocab`, but I should refactor things based on this.
2. The parser needs to be updated carefully to do the #6767 change of calling `set_annotations` in its update method. The parser plays this trick of dividing up the batch into sub-sequences, so that we can make the batch more even in length. This means we don't have a single predicted sequence for the states. I've changed it to use the oracle sequence, instead of making new predictions.

### DIFF
--- a/spacy/about.py
+++ b/spacy/about.py
@@ -1,6 +1,6 @@
 # fmt: off
 __title__ = "spacy-nightly"
-__version__ = "3.0.0rc4.dev19"
+__version__ = "3.0.0rc4.dev20"
 __download_url__ = "https://github.com/explosion/spacy-models/releases/download"
 __compatibility__ = "https://raw.githubusercontent.com/explosion/spacy-models/master/compatibility.json"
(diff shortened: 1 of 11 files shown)


honnibal:
The `set_annotations` calls during `component.update()` are too breaking. Reverting this.

PR #6809, opened 2021-01-25
Fix Doc.copy bugs

honnibal:
* Avoid letting the `Doc` object own `LexemeC` data. Instead all lexemes should be owned by the `Vocab`. Also avoids weird heuristics.
* Copy `doc.spans` attribute.


adrianeboyd:
Definitely a fan of getting rid of the weird heuristics here!

PR #6811, opened 2021-01-25
Update develop from master for v3.0.0rc5

adrianeboyd:
## Description

Update develop from master for v3.0.0rc5

### Types of change

?

## Checklist


spacy/tests/regression/test_issue6755.py
```diff
@@ -0,0 +1,9 @@
+# coding: utf8
+from __future__ import unicode_literals
```
svlandeg, line 2:
```suggestion
```

Issue #6630, opened 2020-12-25
catching errors with nlp.pipe


louisabraham:
In a very large corpus, some of them have special characters. When launching en_trf_bertbaseuncased_lg with nlp.pipe, I get exceptions like the one under but I have no way to catch them. This is related to #4847.

```
/usr/local/lib/python3.6/dist-packages/spacy_transformers/pipeline/wordpiecer.py in pipe(self, stream, batch_size)
     74             docs = list(docs)
     75             outputs = self.predict(docs)
---> 76             self.set_annotations(docs, outputs)
     77             for doc in docs:
     78                 yield doc

/usr/local/lib/python3.6/dist-packages/spacy_transformers/pipeline/wordpiecer.py in set_annotations(self, docs, outputs, tensors)
    177                 self.model.max_len = max_len
    178                 raise ValueError(
--> 179                     f"Error calculating word pieces for sentences. Total number "
    180                     f"of wordpieces in the doc was {nr_word}, but adding up the "
    181                     f"wordpieces for its sentences we get {words_per_sent}. This "

ValueError: Error calculating word pieces for sentences. Total number of wordpieces in the doc was 1437, but adding up the wordpieces for its sentences we get 1435. This means there's a bug in the extension attributes or the tokenizer.add_special_tokens() logic, often when a spaCy sentence aligns against 0 wordpieces.
```

svlandeg:
Right, so there are two issues here:

- The `ValueError` that is being thrown by `en_trf_bertbaseuncased_lg ` feels like a bug, though I couldn't say whether it's coming from `spacy-transformers` itself or from the underlying transformer library. For [spaCy v3](https://nightly.spacy.io/usage/v3), we've revamped the `spacy-transformers` library quite significantly. In v2, this was very much still experimental and I think fixing (potential) bugs in that old version will have lower priority now that we're on the verge of releasing 3.0. Ofcourse if this is still an issue in v3, we'd definitely need to look into this further. A reproducible code snippet with sample data would help us to investigate further.
[...]

Issue #4847, opened 2019-12-29
Error handling with nlp.pipe


AlJohri:
I'm using `nlp.pipe` (with multiple processes) in the latest spaCy (master branch) and having trouble doing proper error handling in a generator.

I have a custom pipeline component which occasionally errors on some documents with `ValueError: [E103] Trying to set conflicting doc.ents: '(1387, 1388, 'PERSON')' and '(1387, 1388, 'PERSON')'. A token can only be part of one entity, so make sure the entities you're setting don't overlap.`.

#### Issue

1) I can't easily figure out which document ID caused the issue since my custom pipeline components doesn't have access to the `context`. The `context` is provided to `nlp.pipe` via the `as_tuples=True` argument.

    ```python
    gen = tqdm(
        nlp.pipe(zip(df.text, df.id), as_tuples=True, batch_size=10, n_process=16),
        total=len(df),
    )
    ```

2) I can't figure out how to ignore the exception and continue processing since it doesn't seem like there's an easy way to catch the error and continue the generator, especially in a multiprocessing setting.

#### Potential Solution

Can we add an `ignore_exceptions=True, error_handler=callable` which allows ignoring exceptions and running some error handler function that has access to the `handler(error, doc, context)`?

#### Info about spaCy

* **spaCy version:** 2.2.3
* **Platform:** Darwin-18.7.0-x86_64-i386-64bit
* **Python version:** 3.7.5

adrianeboyd:
General error handling is a larger issue that should be addressed, but to avoid the immediate `ValueError [E103]` above in your custom component, you can use `util.filter_spans()` to filter out overlapping spans before setting entities: https://spacy.io/api/top-level#util.filter_spans

AlJohri:
Thanks- I think for my particular use case I might need to keep the entities separate. Once I get my hands on the offending document, I'll know more.

I also have highly varying sizes of documents in my list. In terms of number of characters:

```
count     40389.000000
mean      15318.084256
std       16388.481913
min           1.000000
25%        4220.000000
50%        7209.000000
75%       22718.000000
max      242799.000000
Name: length, dtype: float64
```

I keep lowering the batch size on each invocation but I keep running into:

```
MemoryError: Unable to allocate 542. MiB for an array with shape (142106, 1000) and data type float32
```

I am now trying a batch size of 1 even though I'm running on a fairly beefy c5.4xlarge machine (16 cpu, 32gb ram).

[...]

PR #6817, opened 2021-01-25
Error handling in nlp.pipe

svlandeg:
This seems a frequently asked feature - e.g. https://github.com/explosion/spaCy/issues/6630 and https://github.com/explosion/spaCy/issues/4847, and I've been wanting it myself too ;-)

## Description
With the `Pipe` / `TrainablePipe` API revamped for spaCy 3, I thought it would make sense to also add an error handler to `nlp.pipe` and the `pipe` methods of trainable pipeline components.

The default error handler changes nothing: it just reraises the same exception. But the user has an option of specifying its own error handling function that can ignore the batch, inspect the batch, warn, ... 

## Open question
Which arguments should the error handler function take? Currently it takes the component's name, the offending batch of documents, and the original error that was thrown.

### Types of change
enhancement, UX

## Checklist

## TODO
- [ ] In multi-processing context

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -8,7 +8,7 @@ from contextlib import contextmanager
 from copy import deepcopy
 from pathlib import Path
 import warnings
-from thinc.api import Model, get_current_ops, Config, Optimizer
+from thinc.api import get_current_ops, Config, Optimizer
 import srsly
 import multiprocessing as mp
 from itertools import chain, cycle
@@ -20,7 +20,7 @@ from .pipe_analysis import validate_attrs, analyze_pipes, print_pipe_analysis
 from .training import Example, validate_examples
 from .training.initialize import init_vocab, init_tok2vec
 from .scorer import Scorer
-from .util import registry, SimpleFrozenList, _pipe
+from .util import registry, SimpleFrozenList, _pipe, raise_error
 from .util import SimpleFrozenDict, combine_score_weights, CONFIG_SECTION_ORDER
 from .lang.tokenizer_exceptions import URL_MATCH, BASE_EXCEPTIONS
 from .lang.punctuation import TOKENIZER_PREFIXES, TOKENIZER_SUFFIXES
@@ -1282,6 +1282,7 @@ class Language:
         scorer: Optional[Scorer] = None,
         component_cfg: Optional[Dict[str, Dict[str, Any]]] = None,
         scorer_cfg: Optional[Dict[str, Any]] = None,
+        error_handler: Callable[[str, List[Doc], Exception], Any] = raise_error,
     ) -> Dict[str, Union[float, dict]]:
         """Evaluate a model's pipeline components.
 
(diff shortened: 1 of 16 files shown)


honnibal:
This is a really cool idea! I'll avoid merging it just yet so we can think about the open questions, but I'm definitely in favour of this.

honnibal:
> Which arguments should the error handler function take? Currently it takes the component's name, the offending batch of documents, and the original error that was thrown.

May as well pass the component itself, right?

honnibal:
Do we have to have this in the components?

I'd rather have it just on the `Language.pipe`. It seems like the sort of argument people will always be forgetting to add to their components' `pipe()` methods, causing errors at unexpected times.

I don't know how to make the error handling work in `Language` with the generator trickery though. But I do get worried at this change to the `pipe` signature.

honnibal:
How come we have so many of these `pipe` implementations, anyway? Aren't there a bunch that could be using `TrainablePipe.pipe`?

svlandeg:
> Edit: There's really no way to make it work in `Language.pipe` without passing it into the generators themselves, is there. 

I don't directly see how. Alternatively, we could change the API of the components `pipe` methods entirely and make them take in a batch instead of a generator, and do the batching (only) in `Language`, but that's equally breaking and will cause a bit of headache as well.

honnibal:
The `component.pipe()` thing has more disadvantages than I realised when I did it, but it also has advantages, and I think we should accept that it's just How Things Are.

How about this, to avoid changing the arguments of `component.pipe()`:

* Add new methods `Pipe.set_error_handler` and `Pipe.get_error_handler`
* Add a new method `set_error_handler` to `Language`. This will go over the pipeline and set the error handler on components, _if_ they have a `set_error_handler` method. Alternatively, you can set different error handlers to specific components.
* Inside the `.pipe` and `.__call__` methods, we try to get the error handler, or use the default one if we don't have one. Again we don't assume the API is there, we have to do `hasattr` checks etc (gross, I know...).

I think carrying the `error_handler` as state instead of an argument is a bit nicer, because it lets you set different error handlers on different components. It also doesn't really depend on the data you want to pipe through, and it's something I can imagine setting in the config, which wouldn't work if it's an argument of `Language.pipe`.

svlandeg:
I've been thinking about this some more and agree that keeping this as a state instead of an additional argument probably makes more sense. I changed it such that the `Language` object (also) keeps track of a default error handler, so that it can do something useful also for components that don't do it themselves (like simple functions).

I'm still a little unsure about the current implementation.  `TrainablePipe` now takes care of error handling in its `pipe` method (and I removed the redundant overriding duplicates) but then what about the `__call__` methods? It feels like it would be weird not to have the same error handling when you do `proc(doc)`, especially considering the error handler is now saved on the `proc` object. But if you want to make this consistent, you end up with large `try-except` blocks in other `__call__` methods like the one in `AttributeRuler`, and I really really detest those. Then again we're not catching and ignoring, we're just giving the user the option to do whatever, which is nice I think. It just looks really awful.

honnibal:
> But if you want to make this consistent, you end up with large try-except blocks in other __call__ methods like the one in AttributeRuler, and I really really detest those. Then again we're not catching and ignoring, we're just giving the user the option to do whatever, which is nice I think. It just looks really awful.

Ugh, yeah. We can at least refactor these. Even just moving the body to a `_call` method is a more than cosmetic improvement, because if you subclass, you can change the `_call` and get the error handling from the parent.

I find the `.predict()` and `set_annotations()` division appealing. If we find `predict()` misleading for a "ruler", we could call it `match()`?

svlandeg:
> (I sense the growth of another base type here, something like `MatcherPipe`, but let's not get into that.)

I shall restrain myself for now ;-)

PR #6818, opened 2021-01-26
WIP: Various small training changes

ines:
## Description

Mostly stuff that came up in Prodigy nightly.

### Types of change
bug fix, enhancement

## Checklist


spacy/scorer.py
```diff
@@ -461,7 +461,7 @@ def score_cats(
                 if gold_score is not None and gold_score > 0:
                     f_per_type[gold_label].fn += 1
             elif pred_cats:
-                pred_label, pred_score = max(pred_cats, key=lambda it: it[1])
+                pred_label, pred_score = max(pred_cats.items(), key=lambda it: it[1])
```
ines, line 464: Is this right? I was getting strange errors because `pred_cats` was a dict, and this fixed it. Is this a condition that normally doesn't happen so it never came up before?

honnibal: That code is surely wrong, your fix looks right.

PR #6822, opened 2021-01-26
semantically more correct explanation of is_oov in docs

jganseman:
## Description
semantically more correct explanation of is_oov in token and lexeme docs
(since a True value actually means it is NOT in the vocabulary.)

### Types of change
Change in the documentation

## Checklist


ines:
Thanks, appreciate the attention to detail! 🙏

PR #6828, opened 2021-01-27
Sync develop and master

ines:
### Types of change
chore

## Checklist


adrianeboyd:
This is having problems because I already did most of this, but it looks like Matt accidentally squashed it. You probably want to revert the squashed commit and then start again from that PR branch (the merge commit itself should contain all the v3 modifications up to that point): https://github.com/explosion/spaCy/pull/6811

ines:
Ah, damn, I didn't see that. I mostly did this update because I merged the `ky` changes and wanted to get them ported over.

But actually, looking at the tests, it seems like the only problem was an update to the `ky` test that I missed? Maybe the squashed commit already took care of the matcher-related stuff (because thinking about it, I was surprised that I didn't get any merge conflicts for it – this might explain why).

PR #6852, opened 2021-01-29
Add option to replace listeners for sourced components

ines:
## Description

Related discussion: #6649

Sourcing components from an existing model and only training some of them is a very common workflow. For example, a user may want to source the tagger and parser from the `en_core_web_sm` pipeline, freeze those two components and only train an NER component on top of it. By default, all component models "listen" to the same token-to-vector component (e.g. `Tok2Vec` or `Transformer`), so even though the tagger/parser are frozen and not updated, the tok2vec component will be updated together with the NER, which can lead to significantly degraded performance for the frozen component. This is logical, but pretty unintuitive.

A solution for this is to replace the tok2vec listeners with standalone tok2vec layers so that components can "own" their tok2vec weights instead of sharing them. This may result in slightly larger package sizes and copied weights, but allows components to be more modular.

This PR adds support for this process via a new `Language.replace_listeners` method that takes care of updating the component model in place and adjusting the config accordingly. In the config, a sourced component can now optionally define a list of `replace_listeners`, the paths to the listener layers to replace (either none or all of them).

**Example 1:** Freeze the tagger, replace its listener with a copy of the tok2vec layer and train the NER and tok2vec components

```ini
[training]
frozen_components = ["tagger"]

[components]

[components.tok2vec]
source = "en_core_web_sm"

[components.tagger]
source = "en_core_web_sm"
replace_listeners = ["model.tok2vec"]

[components.ner]
source = "en_core_web_sm"
```

**Example 2:** Freeze the tok2vec and tagger component, replace the NER's tok2vec listener with a copy of the tok2vec layer and train only the NER component

```ini
[training]
frozen_components = ["tok2vec", "tagger"]

[components]

[components.tok2vec]
source = "en_core_web_sm"

[components.tagger]
source = "en_core_web_sm"

[components.ner]
source = "en_core_web_sm"
replace_listeners = ["model.tok2vec"]
```

### Todo

- [x] more testing – it's a bit tricky to test this in the test suite, but we could potentially initialize a pipeline with weights, save it to a temp directory and then define a config to source the components from, with different configurations
- [x] update documentation (once we're happy with the implementation)

### Types of change
enhancement

## Checklist


spacy/language.py
```diff
@@ -671,6 +671,81 @@ def create_pipe(
...
+                )
+                raise ValueError(err)
+            pipe = self.get_pipe(pipe_name)
+            # Go over the listener layers and replace them
+            for listener in pipe_listeners:
+                util.replace_model_node(pipe.model, listener, tok2vec.model.copy())
+            # Update the config accordingly by coping the tok2vec model to all
+            # sections defined in the listener paths
+            for listener_path in listeners:
+                # Check if the path actually exists in the config
```
honnibal, line 739: I think it'd be good to reorder things slightly so that this check happens before we do the model replacing?

spacy/util.py
```diff
@@ -738,6 +738,24 @@ def get_package_path(name: str) -> Path:
     return Path(pkg.__file__).parent
 
 
+def replace_model_node(model: Model, target: Model, replacement: Model) -> None:
```
honnibal, line 741: Arguably we should have this in Thinc. If we do put a version of this in Thinc I guess we can remove this utility later.

spacy/util.py
```diff
@@ -1279,6 +1297,25 @@ def dot_to_object(config: Config, section: str):
...
+    """Update a config at a given position from a dot notation.
+
+    config (Config): The config.
+    section (str): The dot notation of the section in the config.
+    value (Any): The value to set in the config.
+    """
+    component = config
+    parts = section.split(".")
+    for i, item in enumerate(parts):
+        try:
```
honnibal, line 1310: Is there no alternative to this try/except?

spacy/language.py
ines: Sure, we can also swap this and exit early if the paths don't match.

spacy/util.py
ines: 🤷‍♀️ It's also how we do it in the very similar `dot_to_object` helper. I guess the alternative would be to do an instance check for dict and check if the key exists. Not sure if there are any other cases we'd have to cover.

spacy/training/initialize.py
```diff
@@ -68,10 +69,11 @@ def init_nlp(config: Config, *, use_gpu: int = -1) -> "Language":
...
     # Detect components with listeners that are not frozen consistently
     for name, proc in nlp.pipeline:
-        if getattr(proc, "listening_components", None):
+        if getattr(proc, "listening_components", None):  # e.g. tok2vec/transformer
             for listener in proc.listening_components:
                 if listener in frozen_components and name not in frozen_components:
                     logger.warning(Warnings.W087.format(name=name, listener=listener))
+                # We always check this regardless, in case user freezes tok2vec
                 if listener not in frozen_components and name in frozen_components:
                     logger.warning(Warnings.W086.format(name=name, listener=listener))
```
svlandeg, line 78: Perhaps W087 and WO86 can hint to this new functionality as well?

svlandeg (commented):
This is awesome! 

It won't be a very easy/straightforward utility to use, with the listener components already being a bit complex/unintuitive. But hopefully we'll be able to guide people through this when their use-case requires it. 

Having to define the `replace_listeners` list makes things more difficult though, I wish there was a way to avoid it (and still be generic enough). Couldn't we add this information to the internal `listener_map` of the `Tok2Vec` / `Transformer` component? Because when you're calling `find_listeners`, you're actually collecting all the relevant paths. So basically the `Tok2Vec` would know which sublayers of `Tagger` are listening to it, and then in the config you could just say "make the `tagger` component fully independent of the `tok2vec` component"?

ines:
@svlandeg 

> Having to define the replace_listeners list makes things more difficult though, I wish there was a way to avoid it (and still be generic enough). Couldn't we add this information to the internal listener_map of the Tok2Vec / Transformer component? 

That's an interesting idea, but I'm not sure it would work? 🤔 (It's all pretty abstract, though, so if you have an idea, feel free to submit a quick PR, might be easier to discuss if there's code 😅)

The main reason I ended up adding this list is that we can't know where in the config the listeners are. In theory, they could be anywhere and passed in at any point as an argument. Once an object (model) is created, we can find all listeners, but we're not able to reconstruct where they were defined in the config 😞

spacy/pipeline/tok2vec.py
```diff
@@ -82,6 +82,15 @@ def add_listener(self, listener: "Tok2VecListener", component_name: str) -> None
         self.listener_map.setdefault(component_name, [])
         self.listener_map[component_name].append(listener)
 
+    def remove_listener(self, listener: "Tok2VecListener", component_name: str) -> None:
```
ines, line 85: ⚠️ Important note: We need to remember to add this to the `Transformer` component as well.

svlandeg: I'll just go ahead and create that PR now

svlandeg:
> The main reason I ended up adding this list is that we can't know where in the config the listeners are. In theory, they could be anywhere and passed in at any point as an argument. Once an object (model) is created, we can find all listeners, but we're not able to reconstruct where they were defined in the config 😞

Yea I think you're right 😭

spacy/language.py
```diff
@@ -1671,8 +1672,92 @@ def from_config(
...
+            not hasattr(tok2vec, "model")
+            or not hasattr(tok2vec, "listener_map")
+            or "model" not in tok2vec_cfg
+        ):
+            raise ValueError(Errors.E888.format(name=tok2vec_name, pipe=type(tok2vec)))
+        pipe_listeners = tok2vec.listener_map.get(pipe_name, [])
+        pipe_cfg = self._pipe_configs[pipe_name]
+        if listeners:
+            util.logger.debug(f"Replacing listeners of component '{pipe_name}'")
+            if len(listeners) != len(pipe_listeners):
```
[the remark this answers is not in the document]
ines: Yeah, in that case, you'd specify both paths and end up with two copies, which is unideal... but also such a specific and advanced use case that you should probably consider not sourcing components all and setting up your config from scratch (or not replacing listeners on that particular component and on all other components instead, or something like that) 😅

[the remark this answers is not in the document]
svlandeg: I was mainly contemplating whether these objects should be sets instead of lists, but the more complex use-case might work with lists, and not sets, so yea, erm, I think all is good? :-)

honnibal:
> Agreed. There is a spaCy-specific issue here though. The TransformerListener.v1 specifically sets a ref "listener" on its model to be able to get the actual TransformerListener layer which is chained into the model. This is used in the get_tok2vec_width utility function that is needed for shape inference when we combine a transformer with textcat.

I'm not really convinced that problem needs solving. Sure, we'll have a ref there named `listener` that returns a non-listener result, due to this quirk of the model's history. But it does return the thing the listener pointed to, so...

spacy/util.py
honnibal: I'm not really convinced that problem needs solving. Sure, we'll have a ref there named listener that returns a non-listener result, due to this quirk of the model's history. But it does return the thing the listener pointed to, so...Trying to clean it up would potentially cause just as many, but different, problems.

svlandeg: It's not a huge problem and it won't lead to any bugs *right now*, but I can just feel us coming back to this in 6 months going "oh right, that listener ref is not actually a listener and now this new thing here broke down". I'm OK with leaving this as is though, because my proposed solution definitely isn't clean either.

PR #6857, opened 2021-01-29
Remove --no-cache-dir when downloading models

tupui:
## Description
When `--no-cache-dir` is present, it prevents caching to properly function.

If the user still wants to do this, there is the possibility to pass options with `user_pip_args`.
But you should not enforce options like these. In my case this is preventing some docker build (using buildkit caching) to have proper caching of models.

### Types of change
This PR removes the enforced `pip` options and leave this to the user.

## Checklist

### DIFF
--- a/.github/contributors/tupui.md
+++ b/.github/contributors/tupui.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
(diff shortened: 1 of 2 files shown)


ines:
Thanks for the PR, this is definitely reasonable! 👍

I think we initially added this because people kept running into this problem: pypa/pip#2984 (and we may have not yet had support for passing through arbitrary pip args). But if we document the `--no-cache-dir` suggestion, users can just set this manually if they run into memory problems.

(On a related note, we might also consider distributing models as binary wheels, at least as an additional option. My PR #6860 adds the ability to build wheels via `spacy package`.)

tupui:
Thanks @ines for handling this quickly, really appreciate 😀. So another reason for me to use latest 😉.

The wheels are for sure a good thing! Looking forward to try this.

PR #6859, opened 2021-01-29
Fix linking resumed components

svlandeg:
Related to https://github.com/explosion/spaCy/discussions/6789

## Description
When linking a `TransformerListener` to a downstream component, the `nO` dimension of the listener is set according to the dimension of the `Transformer`. This linking is triggered by `nlp.initialize`, but we have to ensure that we also link resumed/frozen components. I think, in hindsight, having the linking implicitely in `nlp.initialize` is therefor probably a bad idea.

As a minimal fix, this PR adds a flag whether or not the linking should happen by `nlp.initialize`, allowing the CLI to turn it off and instead perform the linking on *all* components.

At least this fixes the bug from https://github.com/explosion/spaCy/discussions/6789, but we may want to refactor/reconsider this further at some point.

Not sure whether to add this flag to the docs, as it's mostly internals and kind of hard to explain clearly...

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1528,7 +1528,7 @@ class Language:
             for proc in procs:
                 proc.terminate()
 
-    def link_components(self) -> None:
+    def _link_components(self) -> None:
         """Register 'listeners' within pipeline components, to allow them to
         effectively share weights.
         """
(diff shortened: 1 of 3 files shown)


honnibal:
Sorry I've dragged my feet on this. It felt inelegant so I was reluctant to merge it, but I also didn't put in the time to offer another suggestion. We should merge this though because it's a working workaround, and doesn't do anything we can't undo.

PR #6878, opened 2021-02-01
Add capture argument to project_run

svlandeg:
## Description
Add `capture` flag to `project_run` instead of having it default at `False`. From the unit tests of `projects` it would be good to be able to set this to `True` so we can debug cases where the CLI fails.

Bumping the version to 3.0.1. so we can potentially release it to test the projects CLI further.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/about.py
+++ b/spacy/about.py
@@ -1,6 +1,6 @@
 # fmt: off
 __title__ = "spacy"
-__version__ = "3.0.0"
+__version__ = "3.0.1"
 __download_url__ = "https://github.com/explosion/spacy-models/releases/download"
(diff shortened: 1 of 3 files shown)


spacy/about.py
```diff
@@ -1,6 +1,6 @@
 # fmt: off
 __title__ = "spacy"
-__version__ = "3.0.0"
+__version__ = "3.0.1"
```
honnibal, line 3:
```suggestion
__version__ = "3.0.1.dev0"
```

Let's just build a dev version so we can publish it to pypi without stress?

honnibal:
I think the flag is a good idea 👍

PR #6917, opened 2021-02-03
Add open and closed exclamation and question marks to is_left_punct and is_right_punct

0x2b3bfa0:
## Description
Add open and closed exclamation and question marks to is_left_punct and is_right_punct

[Spanish](https://en.wikipedia.org/wiki/Inverted_question_and_exclamation_marks) and some of its co-official languages use open and closed question and exclamation marks. I think that it would be really useful to include them on the `is_left_punct` and `is_right_punct` functions, given that they're similar to any other opening/closing symbol pair.

### Types of change
This change is both an enhancement and a bug fix, if we consider the previous behavior as a bug.

## Checklist
- [X] I have submitted the spaCy Contributor Agreement 👉  #6916 👈 
- [X] I ran the tests, and all new and existing tests ***either*** passed ***or were skipped***.

### DIFF
--- a/spacy/lang/lex_attrs.py
+++ b/spacy/lang/lex_attrs.py
@@ -64,13 +64,13 @@ def is_quote(text: str) -> bool:
 
 def is_left_punct(text: str) -> bool:
     # fmt: off
-    left_punct = ("(", "[", "{", "<", '"', "'", "«", "‘", "‚", "‛", "“", "„", "‟", "‹", "❮", "``")
+    left_punct = ("(", "[", "{", "<", '"', "'", "«", "‘", "‚", "‛", "“", "„", "‟", "‹", "❮", "``", "¿", "¡")
     # fmt: on
     return text in left_punct
 
 
 def is_right_punct(text: str) -> bool:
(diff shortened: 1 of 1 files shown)


svlandeg:
Thanks for the PR!

To be honest I'm not entirely sure about this. I see your point about the Spanish pairs, but then wouldn't it make sense to extend this list just for Spanish (and related languages) instead? Because in other languages `?` or `!` are not really part of a pair. Then again the current implementation has `"‚"` as "left punct" as well. So I guess it depends how people are using this piece of information...

0x2b3bfa0:
I would consider the left/right puntctuation lists as a collection of all the symbols that have a symmetric counterpart, but other users may think differently. My main concern with reducing this change to a few languages is that it probably wouldn't be easily available in language-agnostic use cases like mine.

What would be the impact of this change from the API stability standpoint?

honnibal:
I had to go back and do some refreshing on what the left and right punctuation marks are supposed to indicate, but it does seem clear that it should only be paired punctuation, so @svlandeg is right that we don't want to change the top-level functions. We should only implement this for the languages that have these paired punctuation symbols.

It's true that this can be unfortunate for language-agnostic processing, but you can always create your own language subclass that resolves these features the way you want them.

Issue #6839, opened 2021-01-27
Run PhraseMatcher on Spans (like Matcher)


peter-exos:
## How to reproduce the behavior

The input to Matcher can be of type Doc or Span (called `doclike`), 
but PhraseMatcher does not accept inputs of type Span.

Running 
```python
from spacy.matcher import PhraseMatcher, Matcher
import en_core_web_sm

nlp = en_core_web_sm.load()
matcher, phrase_matcher = Matcher(nlp.vocab), PhraseMatcher(nlp.vocab)

doc = nlp('This is a test sentence. And here is another one.')
for sent in doc.sents:
    matcher(sent)
    phrase_matcher(sent)
```
breaks when `phrase_matcher` is called:
```
Traceback (most recent call last):
  File ".../matcher.py", line 10, in <module>
    phrase_matcher(sent)
  File "spacy/matcher/phrasematcher.pyx", line 249, in spacy.matcher.phrasematcher.PhraseMatcher.__call__
TypeError: Cannot convert spacy.tokens.span.Span to spacy.tokens.doc.Doc
```

This behavior is in line with documentation, so technically it's not a bug. But it would be great if both matchers were more consistent in their interface and both accepted Doc and Span as inputs
(another inconsistency between the two matchers is that `PhraseMatcher.add` has no `greedy` keyword argument, see https://github.com/explosion/spaCy/discussions/6741). 

With both matchers accepting both Doc and Span, the EntityRuler could also accept both input types.
One obvious use case for this is to run the EntityRuler on individual sentences within a Doc without converting the sentence Spans to Docs or writing patterns that make sure there is no intervening sentence boundary. 

## Info about spaCy

* **spaCy version:** 2.3.5 **and** 3.0.0rc3 
* **Platform:** macOS-10.14.6-x86_64-i386-64bit
* **Python version:** 3.8.6

PR #6918, opened 2021-02-03
Run PhraseMatcher on Spans 

peter-exos:
## Description
Addresses https://github.com/explosion/spaCy/issues/6839

Since you may be busy with the spacy 3 release (thank you and congrats!), I went ahead and tried to implement the requested feature myself. It was easier than I thought: 
I extended PhraseMatcher to run on Spans and Docs (like Matcher) by adding start and end index arguments to `PhraseMatcher.find_matches` and using them to constrain the loops. 
Created a regression test and a functional test. All tests passed.

I adjusted the relevant doc strings, but the website docs may need to be updated as well. I have not done that.

This is my first contribution to this repo. 
Please let me know if I am missing anything, or if I can improve something in this PR.

### Types of change
enhancement, matcher/feat

## Checklist

### DIFF
--- a/spacy/tests/matcher/test_phrase_matcher.py
+++ b/spacy/tests/matcher/test_phrase_matcher.py
@@ -337,3 +337,19 @@ def test_span_in_phrasematcher(en_vocab):
     matches_span = matcher(span)
     assert len(matches_doc) == 1
     assert len(matches_span) == 1
+
+
+def test_span_v_doc_in_phrasematcher(en_vocab):
+    """Ensure that PhraseMatcher only returns matches in input Span and not in entire Doc"""
+    doc = Doc(en_vocab,
+              words=["I", "like", "Spans", "and", "Docs", "in", "my", "input", ",", 
+                    "Spans", "and", "Docs", "in", "my", "matchers", ","
+                    "and", "Spans", "and", "Docs", "everywhere" "."])
+    span = doc[9:15]  # second clause
+    pattern = Doc(en_vocab, words=["Spans", "and", "Docs"])
+    matcher = PhraseMatcher(en_vocab)
+    matcher.add("SPACY", [pattern])
+    matches_doc = matcher(doc)
+    matches_span = matcher(span)
+    assert len(matches_doc) == 3
+    assert len(matches_span) == 1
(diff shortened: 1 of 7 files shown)


spacy/tests/matcher/test_phrase_matcher.py
```diff
@@ -323,3 +323,17 @@ def test_phrase_matcher_deprecated(en_vocab):
...
+    doc = Doc(en_vocab,
+              words=["I", "like", "Spans", "and", "Docs", "in", "my", "input", ",", "and", "nothing", "else", "."])
+    span = doc[:8]
+    pattern = Doc(en_vocab, words=["Spans", "and", "Docs"])
+    matcher = PhraseMatcher(en_vocab)
+    matcher.add("SPACY", [pattern])
+    matches_doc = matcher(doc)
+    matches_span = matcher(span)
+    assert len(matches_doc) == 1
+    assert len(matches_span) == 1
```
svlandeg, line 339: Thanks for adding a few tests! I think these are great. Perhaps you could also add one that has multiple matches in the original `doc`, say 3 matches, and then you try and match against a `span` that would only include the middle one? Just to ensure that we're not matching (by accident) outside of the span's boundaries.

svlandeg (approved):
Thanks for the PR! It's clear that you've done an effort to ensure consistency with the existing codebase, using the same `doclike` variable name, `E195` etc, that's really nice :-)

The solution itself is indeed quite elegant, I like it. I have one additional request for a unit test, but otherwise it looks good to me. Perhaps my colleagues want to add anything so let's leave this open for review just a bit longer.

svlandeg:
FYI - if you do want to update the docs already as well, you can find the relevant file under `website/docs/api/phrasematcher.md`. 
I think you'd mainly want to edit the `## PhraseMatcher.\_\_call\_\_` section.

Typically when we edit `.md` files, we apply the "prettier" tool for automatic reformatting (but it's not a big deal if you don't for small edits).

It's also no problem if you don't update the docs yet, then we'll just add it later, so no worries either way.

peter-exos:
Thank you for your great feedback!

I just added another test (`test_span_v_doc_in_phrasematcher`), where the pattern appears 3 times in the Doc, but only 1 time in the Span. Is this what you were thinking about? 
This should check that PhraseMatcher does not match the entire Doc when given a Span, but only the matches in the Span.
If you have any more suggestions how to make the tests more precise, please let me know.

I also updated the website. It's a small change, but please have a look.  
I did not install the website, so I could not see the actual rendering.

spacy/tests/matcher/test_phrase_matcher.py
peter-exos: Added another test (`test_span_v_doc_in_phrasematcher`).  
See comment below.

svlandeg:
This looks great to me, thanks!

spacy/tests/matcher/test_phrase_matcher.py
svlandeg: Looks good!

adrianeboyd:
Very nice PR, thanks! I just made one small edit to the tokenizer.

PR #6965, opened 2021-02-08
Rephrase error related to sample data initialization

adrianeboyd:
## Description

Now that the initialize step is fully implemented, the source of E923 is typically missing or improperly converted/formatted data rather than a bug in spaCy, so rephrase the error and message and remove the prompt to open an issue.

### Types of change

?

## Checklist


svlandeg:
Good point, I've been meaning to change this as well :D

PR #6974, opened 2021-02-08
textcat scoring fix and multi_label docs

svlandeg:
As the new `textcat_multilabel` component has been released as part of v3, it should be documented in some more detail as well.

## Description
* v3 migration guide
* warning note on textcat architectures
* warning note on "normal" textcat to point to changed functionality in comparison to v2
* added the component to some lists of built-in components
* added the link to the API sidebar. Unfortunately it splits over 2 lines though (cf image). @ines: wat do you think?

image

### Types of change
docs update

## Checklist


website/meta/sidebars.json
```diff
@@ -91,6 +91,7 @@
                     { "text": "EntityRuler", "url": "/api/entityruler" },
                     { "text": "Lemmatizer", "url": "/api/lemmatizer" },
                     { "text": "Morphologizer", "url": "/api/morphologizer" },
+                    { "text": "Multi-label TextCategorizer", "url": "/api/multilabel_textcategorizer" },
```
ines, line 94: I don't think we should have a separate page for this (also, all the other pages here use the class name, so this one would have to be `MultiLabel_TextCategorizer`). Since it's a subclass of the `TextCategorizer` with _very_ minor modifications, I think it's fine to document this together with the `TextCategorizer`, basically as a sepcialised variant.

ines:
FYI, I'll also look into adding this to the quickstart!

website/meta/sidebars.json
svlandeg: Ah, ok, I should merge the information from `/api/multilabel_textcategorizer` back into the normal textcat one then.

svlandeg:
Added two small fixes to this PR (didn't feel like merge conflicts putting this in different PRs):
- avoid storing `threshold` in `self.cfg`
- allow the scorer to set custom values for `threshold` and `positive_label`

svlandeg:
(setting to draft, I want to double check something)

adrianeboyd:
Why does it need to be reverted? I think then the threshold should be in `[initialize]` instead like the labels.

svlandeg:
I only reverted to have the discussion internally/separately and to split it off from this PR, so this PR isn't further delayed (as you wanted to work towards a next release).

(I shouldn't have pushed it onto this PR to begin with)

PR #6983, opened 2021-02-09
Several callable objects do not have __qualname__

KoichiYasuoka:
Check of `__qualname__` attribute existance.

## Description
When making some components with class objects, they do not have `__qualname__` before they are included in `__pycache__`.

### Types of change
Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/pipeline/test_pipe_factories.py
+++ b/spacy/tests/pipeline/test_pipe_factories.py
@@ -451,13 +451,27 @@ def test_pipe_factories_from_source_config():
     assert config["arg"] == "world"
 
 
-def test_pipe_factories_decorator_idempotent():
+class PipeFactoriesIdempotent:
+    def __init__(self, nlp, name):
+        ...
+
+    def __call__(self, doc):
+        ...
+
+
+@pytest.mark.parametrize(
+    "i,func,func2",
+    [
+        (0, lambda nlp, name: lambda doc: doc, lambda doc: doc),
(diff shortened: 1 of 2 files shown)


ines:
Thanks! Do you have a simple example that caused `is_same_func` to not return the correct value? It would be cool to include a test for this 🙂

KoichiYasuoka:
Ah well... sorry for my program so dirty:

```py
import numpy
from spacy.symbols import POS,NOUN,X
SPACY_V3=hasattr(Language,"component")

class DummyTagger(object):
  name="DummyTagger"
  def __init__(self,lang,vocab):
    self.lang=lang
    self.vocab=vocab
  def __call__(self,doc):
    pos=[X if self.lang=="xx" else NOUN]*len(doc)
    doc.from_array([POS],numpy.array(pos,dtype="uint64"))
    if not SPACY_V3:
      doc.is_tagged=True
    return doc

def load(lang):
  exec("import spacy.lang."+lang+" as p")
  exec("q=p."+locals()["p"].__all__[0])
  nlp=locals()["q"]()
  if SPACY_V3:
    Language.component("DummyTagger",func=DummyTagger(lang,nlp.vocab))
    nlp.add_pipe("DummyTagger")
  else:
    nlp.add_pipe(DummyTagger(lang,nlp.vocab))
  return nlp

nlp1=load("xx")
nlp2=load("en")
print(list(nlp1("It isn't.")))
print(list(nlp2("It isn't.")))
```

Issue #6950, opened 2021-02-05
Cannot pickle en_core_web_sm object in spacy==3.0


epeters3:
## How to reproduce the behaviour

```python
import spacy
import pickle

en = spacy.load("en_core_web_sm")
# Error only happens *after* the `English` object is called
en("hello world")

with open("temp.pkl", "wb") as f:
    pickle.dump(en, f)

# AttributeError: Can't pickle local object 'Tok2Vec.predict.<locals>.<lambda>'
```

## Your Environment

- **spaCy version:** 3.0.1
- **Platform:** Linux-4.19.112+-x86_64-with-Ubuntu-18.04-bionic
- **Python version:** 3.6.9
- **Pipelines:** en_core_web_sm (3.0.0)

Here is a [colab notebook](https://colab.research.google.com/drive/1flGS_X5D4uDsCUEa12U0IauE1N9_UNty) that reproduces the behavior.

epeters3:
Thanks for fixing that, you guys are doing excellent work!

PR #7025, opened 2021-02-11
Fix issue #6950: allow pickling Tok2Vec with listeners

ines:
## Description

The `lambda` in `Tok2Vec.predict` would result in the `nlp` object not being picklable if the pipeline is initialized and has tok2vec listeners.

Open question: Could we simplify this and just use `None` here, like we do in `spacy-transformers`? https://github.com/explosion/spacy-transformers/blob/392b442eb39e2cbf1f6d6152690a7547f48877ff/spacy_transformers/pipeline_component.py#L220-L221

### Types of change
bug fix, serialization

## Checklist

### DIFF
--- a/spacy/pipeline/tok2vec.py
+++ b/spacy/pipeline/tok2vec.py
@@ -121,7 +121,7 @@ class Tok2Vec(TrainablePipe):
         tokvecs = self.model.predict(docs)
         batch_id = Tok2VecListener.get_batch_id(docs)
         for listener in self.listeners:
-            listener.receive(batch_id, tokvecs, lambda dX: [])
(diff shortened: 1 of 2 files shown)


honnibal:
> Open question: Could we simplify this and just use None here, like we do in spacy-transformers?

We could change the `Tok2VecListener` to support `None`, but that just pushes the conditional somewhere else, so it's unclear it's actually better. We'd need to check for `None` either in the `receive` function or in the `forward`. I think the way you have it is good, although it's troubling to have it inconsistent with `spacy-transformers`. Here's the listener code
https://github.com/explosion/spaCy/blob/master/spacy/pipeline/tok2vec.py#L285

honnibal (approved):

PR #7039, opened 2021-02-11
Fix listener architecture with empty doc in batch

### LINKED ISSUE #7029: POS will be all "ADP" when the inputs have empty string when using .pipe()
When I use `spacy_model.pipe()` to process documents in parallel, the POS tagging will give wrong result if there is any empty string int the documents.
For example:
(1) Normally:
```
docs=["Hello world!","How are you?"]
spacy_docs=spacy_model.pipe(docs)
for spacy_doc in spacy_docs:
    for sentence in spacy_doc.sents:
        print([(token.text,token.pos_) for token in sentence])
```
Output:
```
[('Hello', 'INTJ'), ('world', 'NOUN'), ('!', 'PUNCT')]
[('How', 'ADV'), ('are', 'VERB'), ('you', 'PRON'), ('?', 'PUNCT')]
```
(2) Input has empty string:
```
docs=["Hello world!","How are you?",""]
spacy_docs=spacy_model.pipe(docs)
for spacy_doc in spacy_docs:
    for sentence in spacy_doc.sents:
        print([(token.text,token.pos_) for token in sentence])
```
Output:
```
[('Hello', 'ADP'), ('world', 'ADP'), ('!', 'ADP')]
[('How', 'ADP'), ('are', 'ADP'), ('you', 'ADP'), ('?', 'ADP')]
```
In my opinion, the empty string should not affect the output of other documents. I think there is bug.

### Info about spaCy
- **spaCy version:** 3.0.1
- **Platform:** Darwin-18.7.0-x86_64-i386-64bit
- **Python version:** 3.6.10
- **Pipelines:** en_core_web_sm (3.0.0)

* Operating System: MacOS and Linux
* Python Version Used: python 3.6 and python 3.7
* spaCy Version Used: 3.0.1
* Environment Information:

svlandeg:
Fixes #7029 

## Description
The listener returns some random data with a valid shape for shape inference, when `doc.tensor` isn't set. However using `all` for this check, means that it will hit also if we have an empty document in the batch - effectively wiping the previously calculated `tok2vec` results for the non-empty documents in the same batch. We could use `any` instead as a minimal fix, but really I think we might need to revise this code bit altogether - it's really bug-prone :(

### Types of change
bug fix

## Checklist


svlandeg:
We probably need to adjust https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/layers/listener.py#L63 as well...

spacy/pipeline/tok2vec.py
```diff
@@ -291,7 +291,7 @@ def forward(model: Tok2VecListener, inputs, is_train: bool):
         # of data.
         # When the components batch differently, we don't receive a matching
         # prediction from the upstream, so we can't predict.
-        if not all(doc.tensor.size for doc in inputs):
+        if not any(doc.tensor.size for doc in inputs):
```
honnibal, line 294: Thanks for finding this!

I can't decide whether it's more or less brittle to do it as a loop?

```
outputs = []
width = model.get_dim("nO")
for doc in docs:
    if doc.tensor.size == 0:
        outputs.append(model.ops.alloc2f(len(doc), width))
    else:
        outputs.append(doc.tensor)
```

Maybe the loop is easier to understand at least?

svlandeg: I'm definitely fine with the loop, but I still think we should consider warning here. In most cases, this code path results in trouble. I know we need it for initialization & shape inference, but we could have a clever way (hopefully) to avoid the warnings in that specific case, no?

honnibal: I see what you're saying but I don't want to add more guessing. Let's get the loop in and get this fix deployed, and then plot our next mitigation. 😬

PR #7051, opened 2021-02-12
added spacy-dbpedia-spotlight

MartinoMensio:
## Description
This pull request adds a project to the SpaCy Universe.

[spacy-dbpedia-spotlight](https://github.com/MartinoMensio/spacy-dbpedia-spotlight) allows to get annotations from DBpedia Spotlight within SpaCy. The `doc.ents` are populated with the linked entities.

### Types of change
It is a change to the documentation, since it is only an addition to Universe.

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -1,6 +1,34 @@
 {
     "resources": [
-    	{
+        {
+            "id": "spacy-dbpedia-spotlight",
+            "title": "DBpedia Spotlight for SpaCy",
+            "slogan": "Use DBpedia Spotlight to link entities inside SpaCy",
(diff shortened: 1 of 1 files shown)


ines:
Thanks, this is super cool! 🤩 Also really love the out-of-the-box integration via entry points.

Btw, I was just thinking, one example you could add to the usage docs: If a user is training a pipeline and wants to include the component in it, they can just add the following to their `config.cfg`:

```ini
[nlp]
lang = "en"
pipeline = ["dbpedia"]

[components]

[components.dbpedia]
factory = "dbpedia_spotlight"
overwrite_ents = false
debug = false
```

MartinoMensio:
Thank you very much @ines !
Your feedback is very useful :) and the idea with SpanGroup is brilliant. I will add it to the implementation soon!
I just started playing with version 3 of spaCy and it's amazing 🤓

PR #7073, opened 2021-02-15
Add time and level to default logging formatter

adrianeboyd:
## Description

Add `asctime` and `levelname` to default logging formatter so that it's easier for users to distinguish warnings warnings from logger warnings.

(No particular effort to make this pretty. I mainly added the timestamp to make it look extremely logger-like. Alternate/prettier formatting would be fine.)

### Types of change

Enhancement.

## Checklist


ines:
Just tested it and I think it looks alright 🤷‍♀️

Issue #6990, opened 2021-02-09
Error when GPU initialization and model loading are in different notebook cells


bitmman:
## How to reproduce the behaviour

Installed the latest version following the instruction
```
pip install -U pip setuptools wheel
pip install -U spacy[cuda110]
python -m spacy download en_core_web_trf
```
The installation was successful, and gpu is available:
```
import spacy
from thinc.api import set_gpu_allocator, require_gpu
import spacy_transformers
print(spacy.prefer_gpu(),require_gpu(0))
True True
```

However, Error: "a bytes-like object is required, not 'cupy.core.core.ndarray'" happened, when using "en_core_web_trf".
```
nlp = spacy.load("en_core_web_trf")
sent = 'Hello World'
doc = nlp(sent)
```

```
TypeError                                 Traceback (most recent call last)
<ipython-input-12-66e94dc9d1fd> in <module>
      1 sent = 'Hello World'
----> 2 doc = nlp(sent)


~/anaconda3/envs/acl/lib/python3.8/site-packages/spacy/language.py in __call__(self, text, disable, component_cfg)
    992                 raise ValueError(Errors.E109.format(name=name)) from e
    993             except Exception as e:
--> 994                 error_handler(name, proc, [doc], e)
    995             if doc is None:
    996                 raise ValueError(Errors.E005.format(name=name))

~/anaconda3/envs/acl/lib/python3.8/site-packages/spacy/util.py in raise_error(proc_name, proc, docs, e)
   1493 
   1494 def raise_error(proc_name, proc, docs, e):
-> 1495     raise e
   1496 
   1497 

~/anaconda3/envs/acl/lib/python3.8/site-packages/spacy/language.py in __call__(self, text, disable, component_cfg)
    987                 error_handler = proc.get_error_handler()
    988             try:
--> 989                 doc = proc(doc, **component_cfg.get(name, {}))
    990             except KeyError as e:
    991                 # This typically happens if a component is not initialized

~/anaconda3/envs/acl/lib/python3.8/site-packages/spacy/pipeline/trainable_pipe.pyx in spacy.pipeline.trainable_pipe.TrainablePipe.__call__()

~/anaconda3/envs/acl/lib/python3.8/site-packages/spacy/util.py in raise_error(proc_name, proc, docs, e)
   1493 
   1494 def raise_error(proc_name, proc, docs, e):
-> 1495     raise e
   1496 
   1497 

[...]

PR #7075, opened 2021-02-15
Add warning about GPU selection in Jupyter notebooks

adrianeboyd:
## Description

Due to the bug reported in #6990, warn users with `cupy` available who are working in a Jupyter notebook that they need to run `require_gpu` or `prefer_gpu` in the same cell as `spacy.load` for the model to be loaded on the correct device.

`thinc` uses `contextvars` to store the GPU selection and there is a bug in IPython (https://github.com/ipython/ipython/issues/11565) that does not preserve `contextvars` settings across cells.

### Types of change

Enhancement?

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1615,11 +1615,15 @@ class Language:
                 or lang_cls is not cls
             ):
                 raise ValueError(Errors.E943.format(value=type(lang_cls)))
-        # Warn about require_gpu if a jupyter notebook + cupy is detected
+        # Warn about require_gpu if a jupyter notebook + cupy + mismatched
+        # contextvars vs. thread ops are detected
         if is_in_jupyter():
             from thinc.backends.cupy_ops import CupyOps
             if CupyOps.xp is not None:
-                warnings.warn(Warnings.W111)
+                from thinc.backends import contextvars_eq_thread_ops
+                if not contextvars_eq_thread_ops():
+                    warnings.warn(Warnings.W111)
+
         # Note that we don't load vectors here, instead they get loaded explicitly
         # inside stuff like the spacy train function. If we loaded them here,
         # then we would load them twice at runtime: once when we make from config,
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -20,8 +20,9 @@ from .pipe_analysis import validate_attrs, analyze_pipes, print_pipe_analysis
 from .training import Example, validate_examples
 from .training.initialize import init_vocab, init_tok2vec
 from .scorer import Scorer
-from .util import registry, SimpleFrozenList, _pipe, raise_error, is_in_jupyter
+from .util import registry, SimpleFrozenList, _pipe, raise_error
 from .util import SimpleFrozenDict, combine_score_weights, CONFIG_SECTION_ORDER
+from .util import warn_if_jupyter_cupy
 from .lang.tokenizer_exceptions import URL_MATCH, BASE_EXCEPTIONS
 from .lang.punctuation import TOKENIZER_PREFIXES, TOKENIZER_SUFFIXES
 from .lang.punctuation import TOKENIZER_INFIXES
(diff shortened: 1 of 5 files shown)


website/docs/usage/v3.md
```diff
@@ -1168,3 +1168,15 @@ This means that spaCy knows how to initialize `my_component`, even if your
...
+
+In Jupyter notebooks, run [`prefer_gpu`](/api/top-level#spacy.prefer_gpu),
+[`require_gpu`](/api/top-level#spacy.require_gpu) or
+[`require_cpu`](/api/top-level#spacy.require_cpu) in the same cell as
+[`spacy.load`](/api/top-level#spacy.load) to ensure that the model is loaded on the correct device.
+
+Due to a bug related to `contextvars` (see the [bug
+report](https://github.com/ipython/ipython/issues/11565)), the GPU settings may
+not be preserved correctly across cells, resulting in models being loaded on
+on the wrong device or only partially on GPU.
```
svlandeg, line 1182:
```suggestion
not be preserved correctly across cells, resulting in models being loaded on
the wrong device or only partially on GPU.
```

spacy/language.py
```diff
@@ -1615,6 +1615,11 @@ def from_config(
                 or lang_cls is not cls
             ):
                 raise ValueError(Errors.E943.format(value=type(lang_cls)))
+        # Warn about require_gpu if a jupyter notebook + cupy is detected
+        if is_in_jupyter():
+            from thinc.backends.cupy_ops import CupyOps
+            if CupyOps.xp is not None:
+                warnings.warn(Warnings.W111)
```
honnibal, line 1622: I don't really love the idea of just displaying this warning always, no matter what the user does. We should try to only warn if the user gets it wrong.

How about this:

* We set a global variable reflecting that the GPU has been enabled in some context. This global isn't going to work with threads or contexts, it's not the "source of truth". But it does tell us that something was done.
* When we load, we check whether the following is true: a) we're in a Jupyter notebook, b) The "use gpu" global is set, c) The "use gpu" context var isn't set. If all of those are true we should know we have a problem and can warn, right?

adrianeboyd: Should we move all this logic into thinc, or do you just want to handle common spacy misusage here?

adrianeboyd:
This will fail without the thinc updates from https://github.com/explosion/thinc/pull/469.

spacy/language.py
honnibal: We should move a lot of the logic into Thinc, but I don't see how to detect the different cells between `require_gpu` and `spacy.load` only in Thinc.

adrianeboyd: See: https://github.com/explosion/thinc/pull/469

adrianeboyd: No, I don't think we can detect what you've written, just that the basic thread stuff doesn't match the basic contextvars stuff. Obviously there are environments where this check wouldn't work, but it seems okay for the `is_in_jupyter` + `cupy` context?

website/docs/api/top-level.md
```diff
@@ -158,6 +165,13 @@ if no GPU is available. If data has already been allocated on CPU, it will not
 be moved. Ideally, this function should be called right after importing spaCy
 and _before_ loading any pipelines.
 
+<Infobox variant="warning" title="Jupyter notebook usage">
```
ines, line 168: Maybe we should also link to the `ipython` issue here as well (e.g. as "see this IPython issue for background on this")? If a user comes across this later and the issue happens to already be resolved, it'll be more clear why the warning is there and that it's not a general limitation of spaCy.

spacy/language.py
```diff
@@ -1615,6 +1615,15 @@ def from_config(
                 or lang_cls is not cls
             ):
                 raise ValueError(Errors.E943.format(value=type(lang_cls)))
+        # Warn about require_gpu if a jupyter notebook + cupy + mismatched
+        # contextvars vs. thread ops are detected
+        if is_in_jupyter():
```
ines, line 1620: Not your fault but this kis kinda ugly so maybe we should move it into a helper function? 🤔

website/docs/api/top-level.md
adrianeboyd: Maybe link to the new v3 usage section that does have the link?

ines: Yeah, or that 👍

spacy/language.py
ines: Yeah I think the helper function is worth it, especially considering the `Language.from_config` is already not the easiest method to read/follow. 

It'll hopefully all be temporary anyways.

adrianeboyd: The bug report is old enough that I wouldn't have much short-term hope about an actual solution.

ines (approved):

PR #7111, opened 2021-02-18
var batch is useless

palandlom:
It seems that nlp.update(examples) should be nlp.update(batch)

## Checklist


svlandeg:
Good catch, thanks!

PR #7122, opened 2021-02-19
Fix displacy output in evaluate CLI

adrianeboyd:
## Description

Now that `nlp.evaluate()` does not modify the examples, rerun the pipeline on the (limited) texts in order to provide the predicted annotation in the displacy output option.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
This is kind of unfortunate, but I'm not sure I'd spend a lot of time trying to optimize the performance here. (You could add an option to `nlp.evaluate` to run in-place, I guess? But I'm not sure that's a good idea.)

honnibal:
Yes I can't see any alternative, and this will only trigger if they did ask for the displacy output.

PR #7185, opened 2021-02-24
Update sentencizer documentation example with sentencizer pipe name

clippered:
Update sentencizer documentation example with sentencizer pipe name

## Description
One of the examples is using `entity_ruler` as a pipe name, so just changed it to `sentencizer`

### Types of change
Documentation

## Checklist


svlandeg:
Thanks! Good catch :-)

PR #7197, opened 2021-02-24
Support doc.spans in Example.from_dict

svlandeg:
## Description
For quick tests, it's useful to be able to define `doc.spans` via `Example.from_dict` (e.g. for the upcoming coref work)

cf also comment here: https://github.com/explosion/spaCy/pull/6747#discussion_r559596379. The workaround provided in that PR wouldn't work if the spans are overlapping. See also new unit tests.

### Types of change
enhancement

## Checklist


honnibal:
Ah tupley entity format, we cannot leave you.

Looks good though!

PR #7206, opened 2021-02-25
Add Polish tokenizer exceptions

bittlingmayer:
Polish sentence segmentation will work better with these exceptions.

Ideally there would not be so much repeated code across languages.  Rather these would be JSON files, which would avoid issues with trailing commas etc.

## Description
I copy-pasted the code from /lang/pt (Portuguese).  I took the data from https://en.wiktionary.org/wiki/Category:Polish_abbreviations and added `"ds."` which occurred in our client data.

### Types of change
Enhancement or bux fix?

## Checklist


adrianeboyd:
Unfortunately we don't want to accept these changes for the default Polish tokenizer settings.

The tokenizer settings for languages with pretrained pipelines are customized for our training corpora, in this case [UD_Polish-SZ v2.3](https://github.com/UniversalDependencies/UD_Polish-PDB/tree/r2.3), which always splits the trailing periods from abbreviations into a separate token.

If you need custom settings for your task, see:

* Customizing the tokenizer in a loaded pipeline: https://spacy.io/usage/linguistic-features#native-tokenizer-additions
* Customizing `spacy.Tokenizer.v1` settings in a config: https://spacy.io/usage/linguistic-features#custom-tokenizer-training
* Using a custom tokenizer in a config: https://spacy.io/usage/training#custom-tokenizer

bittlingmayer:
Nothing task-specific or custom here.

The problem is that for many many languages, the rules-based *sentence segmentation* is not good.

(In this case, a sentence is getting split on the `.` in [`ds.`](https://pl.wiktionary.org/wiki/ds.).)

That in turn causes anything else in the pipeline (parsing, classification... - really any task) to fail wildly.
  
All the mid-size languages I have looked at have this problem to some degree, even with the most common abbreviations, which are typically for:
- year
- born
- died
- circa
- the name of the language
- titles of address

Unfortunately right now we have ended up with a kind of "2-class" system - FIGS, Portuguese and a few others get these exceptions, but languages 10-100 do not.

One universal solution would be to at least never split sentences on 1- or 2-letter lowercase words, because the probability that they're an abbreviation is quite high (and the consequences of splitting a sentence in half are typically much worse than the consequences of sending two sentences downstream).

What do you recommend?

From the links you sent, it's not obvious to me where I would add this list.  It would also be unfortunate if many users like we are building and maintaining these lists separately in our private client code, when there is nothing specific to our tasks about them.

adrianeboyd:
It is task-specific, though.

We want `spacy.blank("pl")` to provide tokenization that works well for the training corpora that we are using for `pl_core_news_sm`, in particular for the parser. The spacy pretrained `core` pipelines are a little different than from many other libraries: the parser performs the sentence segmentation. It works well for Polish with a sentence segmentation F-score of 0.98-0.99.

You want `spacy.blank("pl")` to work well with the `sentencizer` (I think?), but that's not the task we're using to make decisions about the default tokenizer settings.

It's certainly possible that the `sentencizer` or your rule-based sentence segmentation algorithm does not work well with the default spacy tokenization for some languages, but everything is super-configurable, so you can modify the tokenization or the sentence segmentation algorithm to work well for you task.

Can you post a bit more about what you're trying to do as a new discussion thread?

bittlingmayer:
Sure, I created a discussion here: https://github.com/explosion/spaCy/discussions/7218

PR #7208, opened 2021-02-25
Support large/infinite training corpora

adrianeboyd:
## Description

Support a training corpus that should not be loaded into memory in `spacy train`:

Training loop:

* Use `max_epochs = -1` to have a train corpus that is streamed rather than loaded in memory and shuffled
* Refactor `create_train_batches` so that the `max_epochs` option that controls whether the corpus is:
  * Loaded in memory once and shuffled by the train loop in subsequent epochs
  * Streamed from the corpus again for each epoch and only batched

### Types of change

Enhancement.

## Checklist


adrianeboyd:
Concerns:

* Would it be better to manage this with a training setting like `nlp.training.infinite_train_corpus` with an alternate batching function in the training loop rather potentially allowing memory leaks by handling it through `max_epochs`?
* What is the best way to coordinate adding `spacy.Corpus.v1` to `spacy-legacy` when this kind of change is made?

svlandeg:
> What is the best way to coordinate adding `spacy.Corpus.v1` to `spacy-legacy` when this kind of change is made?

I think we always need to create the "parallel PR" on `spacy-legacy` that defines the old function in the [setup.cfg](https://github.com/explosion/spacy-legacy/blob/master/setup.cfg) and bumps the version.

The spaCy PR like this one should then bump the lower pin of `spacy-legacy` and probably remain in draft until the new `spacy-legacy` is released?

What I'm not so sure about, is where/how we'll be documenting the old ones?

svlandeg:
I agree with your concern that this solution feels kind of wobbly and too implicit. Having a specific setting for infinite training seems fine to me? It's not like this comes as a surprise when you're training on your data - you know whether it's finite or infinite, right?

adrianeboyd:
Yeah, I'm thinking now the distinction is more load-the-whole-corpus-into-memory vs. not and that would determine whether you shuffle in the training loop or through some other mechanism in the reader. You might still want to have multiple epochs for a large corpus that you don't want to load in memory. Maybe two separate `create_train_batches` functions would better here.

So, let's see:

```ini
[training]
load_training_corpus_in_memory = True
```

And something like:

* `create_train_batches_in memory`: load examples into list, shuffle for each epoch
* `create_train_batches_from_stream`: no shuffling, only batching

I think you'd want the input arg to be `train_corpus` rather than `train_corpus(nlp)` so that you can have multiple epochs for a large finite corpus.

If you want custom behavior, you can always choose the `stream` option and control all the details from the custom reader.

adrianeboyd:
The naming is still ... eh.

adrianeboyd:
Attached is a minimal project that shows how this works: [inf.tar.gz](https://github.com/explosion/spaCy/files/6068752/inf.tar.gz)

adrianeboyd:
`training.shuffle_train_corpus` is a confusing/bad name here. Back to something more like  `load_train_corpus_in_memory` with documentation that it shuffles in the training loop so the reader doesn't have to?

Joozty:
Are you planning to add also support for testing corpus? Currently, it's loaded as a whole to memory when evaulating. :slightly_smiling_face:

adrianeboyd:
No, I think that if the dev corpus is that large, it's not a good fit for the training loop because the evaluation step will take too long. The evaluation method would also have to be completely restructured for it to work with a stream.

What's the kind of use case where it's a problem to load the dev corpus in memory?

adrianeboyd:
As a note, we're planning to go back to using `max_epochs = -1` for streaming a train corpus rather than having an extra setting.

Joozty:
@adrianeboyd For example I want to run multiple trainings with different settings in parallel to see what is performing the best. Something similar to [this](https://github.com/explosion/projects/blob/v3/integrations/wandb/scripts/customize_configs.py#L8). 

The training dataset will be streamed thanks to your PR but if I run multiple trainings and they start evaluating at the same time I can run into the memory issues and eventually run out of memory which results in killing all training instances (or most of them).

Obviously more RAM would solve the problem but like I cannot afford it if I want to run let's say 10 trainings and each requires at least 15gb memory. Sequential training is not possible for me. I would run it for the rest of my life :smile:

adrianeboyd:
I think the simplest solution is to use a smaller but still representative dev corpus. More complicated solutions could involve a custom training loop that limits the number of eval steps that can run concurrently.

spacy/training/initialize.py
```diff
@@ -68,7 +69,11 @@ def init_nlp(config: Config, *, use_gpu: int = -1) -> "Language":
     # Make sure that listeners are defined before initializing further
     nlp._link_components()
     with nlp.select_pipes(disable=[*frozen_components, *resume_components]):
-        nlp.initialize(lambda: train_corpus(nlp), sgd=optimizer)
+        if T["max_epochs"] == -1:
+            logger.debug("Due to streamed train corpus, using only first 100 examples for initialization. If necessary, provide all labels in [initialize]. More info: https://spacy.io/api/cli#init_labels")
+            nlp.initialize(lambda: islice(train_corpus(nlp), 100), sgd=optimizer)
```
adrianeboyd, line 74: Doing this means that:

* you can use a config without labels for textcat if all docs are labelled with all cats
* you can run `init labels` with an infinite training corpus, it will output the file in the right format even if the 100 examples don't contain all labels.

adrianeboyd:
Also, since this is no longer changing the config, we could potentially merge it before v3.1? The current docs make it sound like infinite generators are supported (without further details, which would need to be added) when they aren't.

PR #7209, opened 2021-02-25
Extend score_spans for overlapping & non-labeled spans

svlandeg:
## Description
Extending the functionality of `Scorer.score_spans` so it'll be applicable also for non-NE evaluations like coref mentions, that could be overlapping and do not necessarily have labels.

### Types of change
enhancement

## Checklist


spacy/scorer.py
```diff
@@ -305,6 +305,8 @@ def score_spans(
         *,
         getter: Callable[[Doc, str], Iterable[Span]] = getattr,
         has_annotation: Optional[Callable[[Doc], bool]] = None,
+        consider_label: bool = True,
```
adrianeboyd, line 308: I think `include_label` would be a clearer name.

spacy/scorer.py
```diff
@@ -343,32 +350,39 @@ def score_spans(
             pred_spans = set()
             for span in getter(gold_doc, attr):
                 gold_span = (span.label_, span.start, span.end - 1)
+                if not consider_label:
+                    gold_span = (span.start, span.end - 1)
```
adrianeboyd, line 354: I think an `if consider_label / else` would be clearer here (also in a few other spots).

adrianeboyd (commented):
Just a few minor things. (It's nice to see that it wasn't too hard to extend, or at least I hope it wasn't!)

As a side note, I was wondering in general whether we should switch the whole method (and also the NER PRF method) to use character offsets without referring to any token alignments. I think the scorer originally used token alignments because we didn't necessarily have a character representation of the text in an unspaced `GoldParse`, but I don't think that's an issue now.

spacy/scorer.py
svlandeg: You think? Feels equally clear to me, and the current version is a bit more concise... But it doesn't matter much either way.

spacy/scorer.py
svlandeg: You know I'm notoriously bad at naming things, so happy to take your advice ;-)

svlandeg:
> As a side note, I was wondering in general whether we should switch the whole method (and also the NER PRF method) to use character offsets without referring to any token alignments. I think the scorer originally used token alignments because we didn't necessarily have a character representation of the text in an unspaced `GoldParse`, but I don't think that's an issue now.

That's a good point. Also I think `doc.spans` could hold any span, they shouldn't necessarily align to tokens, right? 

Let's perhaps do it in a follow-up PR though?

adrianeboyd:
No, I meant the "side note" part! It just came up a few times recently and I mentioned it since we were both looking at the details.

I don't think you can have spans that don't line up with tokens, though?

spacy/scorer.py
adrianeboyd: No, it doesn't really matter, but it surprised me when I was reading through it to have it set twice. (I also go back and forth about this kind of thing.)

svlandeg:
Oh right ofcourse, they're just going to be `None` :-)

spacy/scorer.py
svlandeg: @adrianeboyd @honnibal `include_label`, or `include_labels` ?

adrianeboyd: I think since it's `span.label` that `include_label` is better. Maybe Matt has a better suggestion than `include`, but it's also exactly what you wrote in the description, so?

honnibal:
> That's a good point. Also I think doc.spans could hold any span, they shouldn't necessarily align to tokens, right?

`doc.spans` has to hold `Span` objects, which do have to be token-aligned.

svlandeg:
Yea I don't know where my brain was when I typed that ;-) 

Anyway Matt I think if you agree with the naming, this is good to merge?

spacy/scorer.py
honnibal: How about `labelled`?

honnibal:
@svlandeg About the naming I think I'd suggest just `labelled`. I think it's too hard to guess the name if it's `verb_label`

adrianeboyd:
Ooh, a double-l conundrum! I think our official position would be `labeled`. (As a result, I helpfully don't like either!)

svlandeg:
> I think it's too hard to guess the name if it's `verb_label`

I see your point, but an existing parameter is called `has_annotation`, so perhaps it would be consistent to call this one `has_label`? The other parameter I added is currently called `allow_overlap` ... :|

PR #7215, opened 2021-02-26
Allow sourcing disabled components

adrianeboyd:
## Description

Check `component_names` instead of `pipe_names` to allow sourcing disabled components.

### Types of change

Bug fix.

## Checklist


svlandeg:
Just out of curiosity - when/how did this come up in your example NER project?

adrianeboyd:
I'm trying to write NER demos where you update just NER or replace just NER in a pipeline line `en_core_web_sm`, so I want to copy all of `en_core_web_sm` to a new config with everything except `ner` frozen. All components are replaced with `source` and `senter` fails to load in the new config with:

```ini
[components.senter]
source = "en_core_web_sm"
```

PR #7225, opened 2021-02-27
Fix spans weak ref in doc copy

svlandeg:
## Description
`Doc.copy()` would corrupt the doc reference stored in `doc.spans`, because the copied `spans` would keep pointing to the old doc, which would get garbage-collected at some point, eventually leading to dead references.

Proposed solution: allow `Spangroups.copy()` to optionally define a new `Doc` object, so that `Doc.copy()` sets the correct new `Doc`.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tests/doc/test_doc_api.py
+++ b/spacy/tests/doc/test_doc_api.py
@@ -1,3 +1,5 @@
+import weakref
+
 import pytest
 import numpy
 import logging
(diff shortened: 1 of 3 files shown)


svlandeg:
I guess an alternative solution would be to have a `doc.spans` setter which makes sure that you can't set an incompatible `SpanGroups` object on a given `Doc` - because I can imagine this bug popping up in other places as well.

honnibal:
Ah good catch, thanks.

PR #7247, opened 2021-03-02
sync pins with thinc

svlandeg:
## Description
sync pins with thinc

### Types of change
chore

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,4 +28,4 @@ pytest>=5.2.0
 pytest-timeout>=1.3.0,<2.0.0
 mock>=2.0.0,<3.0.0
 flake8>=3.5.0,<3.6.0
-hypothesis>=3.27.0,<6.0.0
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I think we should test thinc with the most recent hypothesis and if possible, raise the upper pin for hypothesis to `<7.0.0` instead. 

It has caused minor complications to have it limited to old-ish versions for conda-forge where new python+arch combinations sometimes only have fairly recent versions. We did have to update some deprecated functions for v5 but it was pretty minor overall. The deprecation warnings looked really awful, but the actual updates were just a few lines in the end. And the changelog says that if it was running without deprecation warnings in v5, which it should have been, you should be able to update to v6 without issues.

svlandeg:
Ok cool I'll adjust the Thinc PR and let's see what the CI says!

adrianeboyd:
Locally it was fine, so I bet it will be okay.

PR #7251, opened 2021-03-02
Fix copy & paste error in API docs

graue70:


ines:
Thanks a lot! (And definitely a classic copy-paste mistake 😅)

PR #7255, opened 2021-03-02
Omit unused tok2vec/transformer components

adrianeboyd:
## Description

Omit unused tok2vec/transformer components in quickstart template.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
This was bugging me. The template has become kind of unusably unwieldy at this point. I wouldn't be sure what to do about the website quickstart, but I think we may want to switch to doing this in python longer-term.

Issue #6420, opened 2020-11-20
Lookaround operators on Matcher patterns


kinghuang:
## Feature description

The [Matcher](https://spacy.io/api/matcher) supports `!`, `?`, `+`, and `*` [operators and quantifiers](https://spacy.io/usage/rule-based-matching#quantifiers). I have text where it would be useful to have something like the regex lookaround patterns, where a pattern should or should not be matched, but is not included as part of the matched range.

For example, consider the following text.

```
Haul from AB CD site to XY site.
```

I want to create patterns for `AB CD site` and `XY site` and label them as source and destination spans. The `from` and `to` tokens are needed to distinguish between `AB CD site` and `XY site`, but should not be part of the match.

```python
from spacy.lang.en import English
from spacy.matcher import Matcher

nlp = English()

m = Matcher(nlp.vocab)
m.add("from_loc", None, [{"ORTH": "from"}, {"OP": "+"}, {"ORTH": "site"}])
m.add("to_loc",   None, [{"ORTH": "to"},   {"OP": "+"}, {"ORTH": "site"}])

doc = nlp.make_doc("Haul from AB CD site to XY site.")
matches = m(doc)

for match_id, start, end in matches:
  print(doc[start:end])
```

```
from AB CD site
to XY site
from AB CD site to XY site
```

The first match span the tokens for `from AB CD site`. I want just `AB CD site` back as the match. Same for the second match.

## Proposal

The Matcher should support the following new ops, roughly based on the regex counterparts.

| Op | Name | Description |
|---|---|---|
| `?=` | Positive lookaround | The token pattern matches, but is not part of the match result. |
| `?!` | Negative lookaround | The token pattern does not match, and is not part of the match result. |

Zero or more lookaround can be used as the start and end of the pattern. A lookaround operator cannot be surrounded on both sides by non-lookaround operators in a pattern.

While there is a distinction between lookahead and lookbehind in regex, these operators are just positive/negative matchers that are not included in the result.

```python
m = Matcher(nlp.vocab)
m.add("from_loc", None, [{"ORTH": "from", "OP": "?="}, {"OP": "+"}, {"ORTH": "site"}])
m.add("to_loc",   None, [{"ORTH": "to", "OP": "?="},   {"OP": "+"}, {"ORTH": "site"}])

doc = nlp.make_doc("Haul from AB CD site to XY site.")
matches = m(doc)

for match_id, start, end in matches:
  print(doc[start:end])
```

```
AB CD site
XY site
[...]

Issue #4642, opened 2019-11-13
spaCy Token Matcher does not support group capture


courtarro:
## Feature description
When using the spaCy Token Matcher, the output is the overall sequence that is matched. While this is useful in many cases, one powerful feature is missing: group capture. At least [two](https://nlp.stanford.edu/nlp/javadoc/javanlp/edu/stanford/nlp/ling/tokensregex/SequenceMatchRules.html) [matchers](https://nlp.stanford.edu/nlp/javadoc/javanlp/edu/stanford/nlp/ling/tokensregex/TokenSequencePattern.html) in Stanford NLP support this explicitly. The [JAPE mechanism in GATE](https://gate.ac.uk/sale/tao/splitch8.html) supports this by including assignment features within the pattern.

Looking over [the code for the current Matcher](https://github.com/explosion/spaCy/blob/master/spacy/matcher/matcher.pyx), I do not see a mechanism for passing out the tokens that satisfy each part of the pattern. My experiments with the Dependency Matcher show that it returns the nodes that satisfy parts of the pattern, so it effectively enables group capture that way. However, its performance is such that it's not a practical replacement for the regular Matcher.

I wrote [a detailed question about this on Stack Overflow](https://stackoverflow.com/questions/58347132/how-do-i-select-the-parts-of-a-spacy-pattern-match-rather-than-the-entire-mat) a while back and got no answer, and looking at the code, it seems that the answer is that it's not currently possible.

adrianeboyd:
No, I don't think it's currently possible. See: #3275.

svlandeg:
Labeling this as "help needed" as we'd certainly accept a PR if someone would like to help contribute to this feature!

polm:
I believe that this is covered by https://github.com/explosion/spaCy/pull/7321, which has been merged and should be released in 3.1, so I'm going to go ahead and close this.

Issue #3275, opened 2019-02-14
Sub-pattern labeling for pattern matcher


cyclecycle:
My problem is that I would like to match patterns based on linguistic annotations, and then label certain tokens within the match, depending on which token_specs those matched tokens correspond to. I therefore need to know which tokens in the match span correspond to which token_specs in the provided patterns. For patterns which don't use operators, this is easy, as the match will be the same length as the pattern, so a corresponding one-to-one list of labels would suffice. For token_specs with operators, it is more complicated, as the number of corresponding tokens in the match is variable.

N.B. Which tokens get which labels is dependent upon their position in a given pattern, so the labeling is tightly coupled to the pattern matching in this context.

I see two solutions:

    - Alter the code in matcher.pyx so that for each token added to each open match, also store the current token_spec.
    - Reproduce the pattern matching logic in a callback in order to map tokens in the match to token_specs in the pattern.

Am I missing an easier solution? I prefer the former option, so I could fork and attempt to make a satisfactory change. I have no idea whether such a feature would merit its cost for the broader community.

Thanks,

Nick

ines:
Hi! Do you have an example of a pattern and token spec / subpattern? Just so I can make sure I understand the question and use case correctly.

It does sound like something that should be possible with the current API and without hacking around too much. Maybe you could use two matchers for this? For instance, if pattern A is matched, you could call `.as_doc()` on the matched span, and then match it again with your other token spec matcher?

cyclecycle:
Hi ines, thanks a lot for your response.

For example, let's say I have the text:

`The action of x on y.`

And I want to extract a relationship in the form of:

predicate: action of
argument1: x
argument2: y

One approach is to use the pattern:

`pattern = [{'LOWER': 'action'}, {'LOWER': 'of'}, {}, {'LOWER': 'on'}, {}]`

Now I want to assign the appropriate labels (pred, arg1, arg2, none) to the tokens in the returned match span. In this example, I could simply use a list of labels `['pred', '-', 'arg1, 'pred', 'arg2']` and zip that with the match span. to get the desired output. However, in patterns with operators, you don't know what the shape of the match span is exactly ahead of time, so this approach doesn't work. You'd ideally want to know which tokens were matched by a certain token spec. 

I hope that is clearer. I'm not sure that the title is a good description of the solution I'm seeking.

honnibal:
I think I get what you mean --- sort of like the `?:` grouping in a regex, right?

[...]

PR #7321, opened 2021-03-06
Support match alignments

broaddeep:
## Description

**Support for match alignments.**

Many users wanted rule-base matcher to support subgroup labeling(#3275) or Group capture(#4642), Look-around operator like regex(#6420). 
However, it wasn't an easy task as seen in the issue(#3275).

To address this issue, I propose the concept of **match alignments.** 
It represents the part of a token pattern that contributed to the match.

For example, suppose we have pattern 
`[{"ORTH": "a", "OP": "+"}, {"ORTH": "b"}]` 
and the text is given as
`a a a b`

The matched span will have four tokens(in the longest greedy setup). 
We can easily verify that the first three matched tokens(`a a a`) was matched by the first token pattern (`{"ORTH": "a", "OP": "+"}`),
and the last token(`b`) was matched by the second token pattern (`{"ORTH": "b"}`)

We can rewrite this in `List[int]`, `[0, 0, 0, 1]`.

Using this information, it can be applied to have the same effect as group capture or look around operator, subgroup labeling.
Any better alternative is welcome. The API needs to be strongly managed by maintainers, so it is not necessary to use the expression `match_alignments.`

## Implementation details

- Each time the state changes, it keeps track of the index of the token pattern at that time and the length of the span.

## API 
```
import spacy

pattern = [
    {'ENT_TYPE': 'PERSON', 'OP': '+'}, 
    {'LEMMA': 'love'}, 
    {'ENT_TYPE': 'PERSON', 'OP': '+'}
]

matcher = Matcher(nlp.vocab)
matcher.add("test", [pattern], greedy='LONGEST')

doc = nlp("John Doe loves Jane Doe. John loves Jane.")

matches = matcher(doc, match_alignments=True)

for m in matches:
    print(m)
# (1618900948208871284, 0, 5, [0, 0, 1, 2, 2])
# (1618900948208871284, 6, 9, [0, 1, 2])
```

- It does not require breaking changes. (All test passed)
- Test case added.

### Types of change

New feature

## Checklist


honnibal:
Thanks, this is a very interesting solution, and the implementation looks good at first glance.

The matcher code has been a source of many bugs over the years, and it's often not easy to debug. So I want to be careful before merging this.

Another concern is whether any changes can change the time-complexity or runtime unfavourably. It looks like your change is good in this respect. There's a copy of the alignment vector on each extend action I believe, it might be better if that could be avoided, especially if the alignments aren't being used? On the other hand it might make no difference at all to the practical runtime.

honnibal:
I think we can get this merged for v3.1.

adrianeboyd:
I agree that this looks like a useful addition. It looks really lightweight, but speed looks like it might be more of a concern than I thought at first glance. Here's a very rough comparison:

| | # Keywords | # Docs | # Matches | Time |
|---|---|---|---|---|
|`master`|10|1000|0|0.72|
| |100|1000|41|4.03|
| |1000|1000|305|36.88|
| |10|10000|0|8.78|
| |100|10000|587|48.31|
| |1000|10000|8006|438.94|
|this PR|10|1000|0|0.89|
| |100|1000|41|5.49|
| |1000|1000|305|50.78|
| |10|10000|0|10.55|
| |100|10000|587|62.91|
| |1000|10000|8006|595.40|

As to naming: `with_alignments`?

honnibal:
Thanks for the benchmark! Yes that's surprising.

Can we only copy the vector if we're using the alignments?

broaddeep:
@adrianeboyd That naming looks better. `with_alignments=True`
@honnibal Yes, it is possible to execute conditionally depending on `with_alignments=True or False`, but the key is how much code duplication and complexity can be reduced. I will try several things to see if there is a better alternative.

broaddeep:
@honnibal I made the vector copying happen according to the `with_alignments` option. 
- The code is a bit more complicated, but the rules are clear. 
  - The `states` - `align_states` vector, `matches` - `align_matches` always have a 1:1 match relationship. 
  - Enabling this option will copy the data into the `align_*` vectors before any changes to the `states/matches` vector occur, otherwise the logic is the same as the existing flow.
  - Therefore, I believe that disabling this option will not make a significant difference in execution speed.
- Added test case, validation error message.

adrianeboyd:
Again, this is just a rough timing test, but the changes look good!

| | # Keywords | # Docs | # Matches | Time |
|---|--:|--:|--:|--:|
|`master`|10|1000|0|0.70|
| |100|1000|41|4.03|
| |1000|1000|305|38.00|
|this PR without alignments|10|1000|0|0.81|
| |100|1000|41|4.08|
| |1000|1000|305|37.11|
|this PR with alignments|10|1000|0|0.76|
| |100|1000|41|4.55|
| |1000|1000|305|42.93|

I don't think we need the additional check/error for `bool` for the kwarg. The reason for the `doclike` type error is that cython doesn't support a `Union[Doc,Span]` kind of type, but the boolean args can just be typed `bint` (cython is quirky) and then you don't need the check or the conversion. I think it's okay for whatever the user puts in here to be accepted as `bool(val)` since we don't do this kind of type checking elsewhere for similar args, either.

And it would be nice to revert all the whitespace edits. If it would be simpler, I can handle the remaining edits if you'd like.

broaddeep:
@adrianeboyd Thank you for the good feedback.
- dropped type checking for bool type
- added bint type for function args
- revert all the whitespace edits.

adrianeboyd:
Thanks, everything looks good! I'll make a few minor edits and add it to the API docs.

adrianeboyd (approved):

honnibal (approved):
Looks great, thanks!

website/docs/api/matcher.md
```diff
@@ -120,13 +120,14 @@ Find all token sequences matching the supplied patterns on the `Doc` or `Span`.
...
+| `as_spans` <Tag variant="new">3</Tag>          | Instead of tuples, return a list of [`Span`](/api/span) objects of the matches, with the `match_id` assigned as the span label. Defaults to `False`. ~~bool~~                                                                                                                                            |
+| `allow_missing` <Tag variant="new">3</Tag>     | Whether to skip checks for missing annotation for attributes included in patterns. Defaults to `False`. ~~bool~~                                                                                                                                                                                         |
+| `with_alignments` <Tag variant="new">3.1</Tag> | Return match alignment information as part of the match tuple as `List[int]` with the same length as the matched span. Each entry denotes the corresponding index of the token pattern. If `as_spans` is set to True, this setting is ignored. Defaults to `False`. ~~bool~~                             |
```
svlandeg, line 129: Sorry for the nitpicking: 

```suggestion
| `with_alignments` <Tag variant="new">3.1</Tag> | Return match alignment information as part of the match tuple as `List[int]` with the same length as the matched span. Each entry denotes the corresponding index of the token pattern. If `as_spans` is set to `True`, this setting is ignored. Defaults to `False`. ~~bool~~                             |
```

kwhumphreys:
@adrianeboyd @honnibal Is there any way to access alignments in callbacks?
It looks like callbacks are called before alignments are built: https://github.com/explosion/spaCy/blob/master/spacy/matcher/matcher.pyx#L287
but it would be useful for callbacks to see the same matches that the Matcher returns.

kwhumphreys:
proposed change at https://github.com/explosion/spaCy/pull/9001

PR #7325, opened 2021-03-07
fix pretraining initialization

svlandeg:
## Description
When running `pretrain`, the `nlp` object should be initialized with the relevant train corpus to ensure the initialization of components happens correctly. Otherwise, the tagger for instance may complain about not having any labels.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1222,10 +1222,6 @@ class Language:
         init_vocab(
             self, data=I["vocab_data"], lookups=I["lookups"], vectors=I["vectors"]
         )
-        pretrain_cfg = config.get("pretraining")
-        if pretrain_cfg:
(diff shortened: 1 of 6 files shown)


ines:
This is tricky, though, because it'd mean that the pretraining will _also_ need to have some arbitrary training data available that's not actually used (in addition to the `pretrain` corpus that _is_ used for pretraining). So unless I'm missing something, this is potentially very confusing 🤔 Should we even need to create any components that are not the pretrained component specified in the config?

svlandeg:
Maybe there's a better way around it. But right now we're fetching the `tok2vec` sublayer of the component we want to pretrain, and to be able to get that we need to initialize that component, right?

svlandeg:
Rewrote entirely and reopened the PR as https://github.com/explosion/spaCy/pull/7342

PR #7336, opened 2021-03-08
Fix lowercase augmentation

adrianeboyd:
## Description

* Fix aborted/skipped augmentation for `spacy.orth_variants.v1` if lowercasing was enabled for an example
* Simplify `spacy.orth_variants.v1` for `Example` vs. `GoldParse`
* Preserve reference tokenization in `spacy.lower_case.v1`

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/training/test_augmenters.py
+++ b/spacy/tests/training/test_augmenters.py
@@ -38,18 +38,58 @@ def doc(nlp):
 
 
 @pytest.mark.filterwarnings("ignore::UserWarning")
-def test_make_orth_variants(nlp, doc):
+def test_make_orth_variants(nlp):
     single = [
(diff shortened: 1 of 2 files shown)


adrianeboyd:
`spacy.lower_case.v1` was broken and is now fixed. Its behavior for examples where it worked is still the same, so I think the version can stay the same.

For `spacy.orth_variants.v1` the intended behavior has not changed, but the actual behavior has changed with these bug fixes (actual lowercasing + corrected whitespace in variant text), so I'm not sure whether the version should be increased.

honnibal:
@adrianeboyd I think it's not worth increasing the version here, as the previous behaviour wasn't really functional and I doubt anyone was relying on it.

Issue #7289, opened 2021-03-05
error when loading config for pretraining


moyid:
I'm working on pretraining a tok2vec model in spacy v3, and I set up my config file with `python -m spacy init fill-config configs/config.cfg configs/config_pretrain.cfg --pretraining` 

When I try to load the model from config, I get this error: 
script run:
```
config_path = Path('./configs/config_pretrain.cfg')
config = util.load_config(config_path, interpolate = False)
nlp = init_nlp(config, use_gpu=0)
```
error:

```
Config validation error
pretraining -> optimizer      instance of Optimizer expected
pretraining -> objective      Promise(registry='architectures', name='spacy.PretrainCharacters.v1', args=[], kwargs={'maxout_pieces': 3, 'hidden_size': 300, 'n_characters': 4}) is not callable
pretraining -> batcher        extra fields not permitted    
pretraining -> component      extra fields not permitted    
pretraining -> corpus         extra fields not permitted    
pretraining -> dropout        extra fields not permitted    
pretraining -> layer          extra fields not permitted    
pretraining -> max_epochs     extra fields not permitted    
pretraining -> n_save_every   extra fields not permitted    
pretraining -> objective      extra fields not permitted    
pretraining -> optimizer      extra fields not permitted    
```

polm:
I guess this is related to #7285. Looks like something is weird with your config, can you run this on it and check the output?

    python -m spacy debug data ./configs/config_pretrain.cfg

It would also help if you could post your training config.

moyid:
`python -m spacy debug data ./configs/config_pretrain.cfg` output:

```
============================ Data file validation ============================
✔ Corpus is loadable
Traceback (most recent call last):
  File "/opt/conda/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/conda/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/jupyter/spaCy/spacy/__main__.py", line 4, in <module>
    setup_cli()
  File "/home/jupyter/spaCy/spacy/cli/_util.py", line 69, in setup_cli
    command(prog_name=COMMAND)
  File "/home/jupyter/NEL_pipeline/env/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/jupyter/NEL_pipeline/env/lib/python3.7/site-packages/click/core.py", line 782, in main
[...]

Issue #6952, opened 2021-02-05
init_tok2vec does not seem to have an impact on model's training


cverluise:
Hello,

first of all, a huge thanks for all your work and the new features supported by the v3. I'm sure it'll make a 💥  for the community in the coming months!

## How to reproduce the behaviour

I'm trying to train a `ner` component with pre-trained vectors.  Here is what I do (config files below):

```shell script
spacy pretrain en_t2vner_pretraining.cfg en_gbpatentxx --paths.raw_text raw_gbpatentxx.jsonl
```

As expected, this produces the `model*.bin`(and runs for a _very_ long time)

Next, I compare the output with and without the pre-trained vectors.

```shell script
# without pre-trained vectors
spacy train en_t2vner.cfg --paths.train train_gpatent01.spacy --paths.dev test_gpatent01.spacy 

# with pre-trained vectors
spacy train en_t2vner.cfg --paths.train train_gpatent01.spacy --paths.dev test_gpatent01.spacy --paths.init_tok2vec en_gbpatentxx/model999.bin

# Nb: dry run
``` 

I get the exact same sequence of loss and everything, suggesting that the `init_tok2vec` does not change anything. Am I doing something wrong or is there a bug?

Thanks in advance,

Cyril

<details><summary>en_t2vner.cfg</summary>

```cfg
[paths]
train = null
dev = null
vectors = null
init_tok2vec = null

[system]
gpu_allocator = null
seed = 0

[nlp]
lang = "en"
pipeline = ["tok2vec","ner"]
batch_size = 1000
disabled = []
before_creation = null
after_creation = null
after_pipeline_creation = null
tokenizer = {"@tokenizers":"spacy.Tokenizer.v1"}

[components]

[components.ner]
factory = "ner"
moves = null
update_with_oracle_cut_size = 100

[components.ner.model]
@architectures = "spacy.TransitionBasedParser.v2"
state_type = "ner"
extra_state_tokens = false
hidden_width = 64
maxout_pieces = 2
use_upper = true
nO = null

[components.ner.model.tok2vec]
@architectures = "spacy.Tok2VecListener.v1"
width = ${components.tok2vec.model.encode.width}
upstream = "*"

[components.tok2vec]
factory = "tok2vec"

[components.tok2vec.model]
@architectures = "spacy.Tok2Vec.v2"

[components.tok2vec.model.embed]
@architectures = "spacy.MultiHashEmbed.v1"
width = ${components.tok2vec.model.encode.width}
[...]

PR #7342, opened 2021-03-08
Fixing pretrain

svlandeg:
- Fixes https://github.com/explosion/spaCy/issues/6952
- Fixes https://github.com/explosion/spaCy/issues/7289
- Fixes https://github.com/explosion/spaCy/discussions/7285

## Description
The `tok2vec` pretraining functionality got pretty broken, I think during the final refactors of the release candidates for spaCy 3. This PR fixes quite a few bugs and introduces better tests:
- in `nlp.initialize`, `init_tok2vec` was called before the components got initialized themselves, meaning that the latter would overwrite the pretrained weights.
- `nlp.from_config` did not play well with the schema because of `pretraining: Union[ConfigSchemaPretrain, ConfigSchemaPretrainEmpty] = {}` in `ConfigSchema`. I'm not sure whether the current solution of temporarily popping the pretraining section is OK, other solutions are welcome but this seems to work at least without breaking anything else...
- Avoid initializing the full `nlp` component when analyzing the requirements for the `tok2vec` pretraining, by disabling all components. This ensures that for instance `init_vocab` still runs correctly. The model of the `tok2vec` layer shouldn't need any other information like labels, so we can `walk` through it finding the correct sublayer without issue. If we do run into issues, typically the pretraining component/layer were specified wrongly (`E874`)
- Removing the outdated check on `P["objective"].get("type")` and replaced with a new check in `PretrainVectors` (`E875`)
- Created a `get_tok2vec_ref` function that can be used both by `pretrain` and `train`
- Added a logger info line when pretrained weights were succesfully loaded
- Tests for actually running the pretrain step with various configs
- Test for running the pretrain step + using the output in a subsequent train step
- More tests for `init_config` and `load_model_from_config`

### Types of change
multiple bug fixes

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -487,7 +487,10 @@ class Errors:
     E202 = ("Unsupported alignment mode '{mode}'. Supported modes: {modes}.")
 
     # New errors added in v3.x
-
+    E874 = ("Could not initialize the tok2vec model from component "
+            "'{component}' and layer '{layer}'.")
+    E875 = ("To use the PretrainVectors objective, make sure that static vectors are loaded. "
+            "In the config, these are defined by the initialize.vectors setting.")
     E879 = ("Unexpected type for 'spans' data. Provide a dictionary mapping keys to "
             "a list of spans, with each span represented by a tuple (start_char, end_char). "
             "The tuple can be optionally extended with a label and a KB ID.")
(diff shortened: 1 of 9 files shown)


honnibal:
Great work! Agree we should get this out quickly.

PR #7353, opened 2021-03-09
Sort EntityRuler labels 

### LINKED ISSUE #7352: EntityRuler labels order not deterministic
If you run the below script multiple times you will realize that the order of labels is random.
```python
# myscript.py
import json
import spacy

nlp = spacy.blank("en")
ruler = nlp.add_pipe("entity_ruler")
patterns = [
    {"label": "label_1", "pattern": "pattern_1"},
    {"label": "label_2", "pattern": "pattern_2"},
]
ruler.add_patterns(patterns)

nlp.to_disk("output_folder")

with open("output_folder/meta.json", "r") as f:
    print(json.load(f)["labels"]["entity_ruler"])

```
Locally, I run
```bash
for i in {1..10}
do
  python myscript.py
done
```
and get
```bash
['label_2', 'label_1']
['label_1', 'label_2']
['label_1', 'label_2']
['label_2', 'label_1']
['label_1', 'label_2']
['label_1', 'label_2']
['label_1', 'label_2']
['label_2', 'label_1']
['label_1', 'label_2']
['label_1', 'label_2']
```
Note that this is problematic because the MD5 hash of the `meta.json` is not deterministic (tools like DVC rely on it).

## Your Environment

* Operating System: MacOS
* Python Version Used: 3.8.8
* spaCy Version Used: very recent master commit 3f3e8110dc6ec3c1449b120c29ffc7d5475ef622. The same issue should therefore be present in multiple Spacy versions.
* Environment Information: nothing special I assume

jankrepl:
Closes #7352

## Description

The idea is to sort the labels. This way the order will always be the same.

### Types of change

This is a bug fix.
## Checklist

### DIFF
--- a/.github/contributors/jankrepl.md
+++ b/.github/contributors/jankrepl.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I think the labels are mainly just for reference in `meta.json` and can't think of any reason that sorting them would cause issues, especially if the order was undefined before. They're not included in this format for serialization, so the side effects of this are really minimal.

PR #7407, opened 2021-03-11
Add note that conversion supports BILOU tags

polm:
The converters support all BILOU tags, but that isn't mentioned in the
docs anywhere, so this adds a mention of that. Came up in this
discussion:

https://github.com/explosion/spaCy/discussions/7351#discussioncomment-467288

One thing is that internally the tags are referred to as "BILUO", though
"BILOU" seems to be more common.

## Description

Updates mention of IOB/IOB2 tags to also mention BILOU tags.

### Types of change

Doc fix.

## Checklist


adrianeboyd:
This is going to conflict with #7394. I'll just make the same change there?

adrianeboyd:
And we use BILUO internally everywhere, so I think it's okay if the docs are the same. I actually don't know what's more common, but I find both in use?

PR #7417, opened 2021-03-11
Optionally append lang for packaged model name

svlandeg:
## Description
* Only prepend the language code if it's not there already - otherwise calling `package` with for instance `name="nl_core_my_pipeline"` would result in an `"nl_nl_core_my_pipeline"` package.
* Nitpicking: add empty newline at the end of the Python files that are written.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -113,7 +113,7 @@ def package(
         print("\n".join(errors))
         sys.exit(1)
     model_name = meta["name"]
-    if not model_name.startswith(meta['lang']):
+    if not model_name.startswith(meta['lang'] + "_"):
         model_name = f"{meta['lang']}_{model_name}"
     model_name_v = model_name + "-" + meta["version"]
     main_path = output_dir / model_name_v
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -294,7 +294,8 @@ def setup_package():
 
 if __name__ == '__main__':
     setup_package()
-""".strip()
+
+"""
 
 
 TEMPLATE_MANIFEST = """
@@ -314,4 +315,5 @@ __version__ = get_model_meta(Path(__file__).parent)['version']
 
 def load(**overrides):
     return load_model_from_init_py(__file__, **overrides)
-""".strip()
+    
+"""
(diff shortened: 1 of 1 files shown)


spacy/cli/package.py
```diff
@@ -112,7 +112,9 @@ def package(
         msg.fail("Invalid pipeline meta.json")
         print("\n".join(errors))
         sys.exit(1)
-    model_name = meta["lang"] + "_" + meta["name"]
+    model_name = meta["name"]
+    if not model_name.startswith(meta['lang']):
```
adrianeboyd, line 116: Maybe `model_name.startswith(meta['lang'] + "_")` instead?

Overall I'm a little worried this change is going to cause too much unexpected behavior.

svlandeg: You think? I think people who were using it correctly (not giving the language code with the name) won't see a difference, right? You think there's people actually aiming for a name like "nl_nl_core_my_pipeline"? :|

adrianeboyd: No, I think you might want `nl_nlcorpusname_something`.

svlandeg: Yea, potentially. I actually had the check first like you suggested so I'll put it back like that.

svlandeg, line 116:
```suggestion
    if not model_name.startswith(meta['lang'] + "_"):
```

svlandeg: Do you still think it's potential breaking/unexpected then?

I actually found the "nl_nl_" bit unexpected myself :p

adrianeboyd: I guess I'm not worried that it's unexpected, just that it's a change that could break things for existing workflows.

svlandeg: There's always a chance, true :| 
We could put it behind an additional flag, a boolean whether to prepend lang, that is by default `True`? 
Or whatever, I'm not invested in this feature either 😁

adrianeboyd: That seems like overkill, and I think the number of potentially affected users is really really small.

svlandeg:
We should hold off on merging until we're ready to publish 3.1

PR #7441, opened 2021-03-15
Clean up Morphology imports and definitions

adrianeboyd:
## Description

Clean up `Morphology` imports and definitions

### Types of change

?

## Checklist


svlandeg:
As discussed - we should probably hold off on merging until 3.1 to avoid binary incompatibility?

adrianeboyd:
Hmm, I renamed the v3.0 tag because that seemed easy but now I see that was a mistake. I should have added a new one or it makes the old PRs confusing.

honnibal (approved):

PR #7471, opened 2021-03-17
Don't warn about components not in the pipeline

polm:
See here:

https://github.com/explosion/spaCy/discussions/7463

Still need to check if there are any side effects of listeners being
present but not in the pipeline, but this commit will silence the
warnings.

### Types of change

Warning fix.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1686,15 +1686,21 @@ class Language:
                 )
         # Detect components with listeners that are not frozen consistently
         for name, proc in nlp.pipeline:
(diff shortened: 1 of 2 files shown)


svlandeg:
Looks good to me!

To prevent this all together, I wonder whether we should do something in `nlp.from_config`?  Around L1688 we're already checking the listeners, perhaps at that point we'd need to call `proc.remove_listener` if the listener is not in the pipeline?

svlandeg:
I think it's fine to keep the changes in `initialize` as is - an additional check can't hurt.

PR #7480, opened 2021-03-18
Update processing-pipelines.md to mention method for doc metadata

langdonholmes:
## Description

Under "things to try," inform users they can save metadata when using nlp.pipe(foobar, as_tuples=True)

Link to a new example on the attributes page detailing the following:
> ```
> data = [
>   ("Some text to process", {"meta": "foo"}),
>   ("And more text...", {"meta": "bar"})
> ]
> 
> for doc, context in nlp.pipe(data, as_tuples=True):
>     # Let's assume you have a "meta" extension registered on the Doc
>     doc._.meta = context["meta"]
> ```

from (one of) [Ines' comments on StackOverflow](https://stackoverflow.com/questions/57058798/make-spacy-nlp-pipe-process-tuples-of-text-and-additional-information-to-add-as)

### Types of change
Update the docs.

## Checklist


adrianeboyd:
Hi, thanks for this contribution! I think it's a great idea to have an example for `as_tuples` in the docs, but it's a bit out-of-place in the proposed locations.

Could you add it instead as a short paragraph + executable example at the end of the "Processing text" section? The main focus of the example would be on using `as_tuples` rather than the custom extension itself, since those are introduced in a separate section, but you can demonstrate how `as_tuples` is useful by assigning the context info to the custom extension.

langdonholmes:
Excellent suggestions. I think I have implemented them and updated the pull request, but I am a bit new to Git so I'm not 100% sure I did that correctly. I am happy to make any additional changes or a new pull request as needed.

adrianeboyd:
Thanks for the updates! I reformatted and rephrased a bit and I'll leave this open for a little while for further feedback from others...

svlandeg:
Looks good to me! I'll go ahead and merge this and push the update to the docs :-)

PR #7488, opened 2021-03-18
replace "is not" with !=

Findus23:
## Description

When running spacy with python3.9 I get the following warning
```text
/srv/venv/venvname/lib/python3.9/site-packages/spacy/training/pretrain.py:145: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  tok2vec.upstream_name if tok2vec.upstream_name is not "*" else "tok2vec"
```

### Types of change
replace "is not" with `!=` as we are comparing with a string

## Checklist


svlandeg (approved):
Good catch, thanks!

PR #7490, opened 2021-03-19
Improve checks for sourced components

adrianeboyd:
## Description

Improve checks for sourced components

* Error:
  * The source object is not a `Language` object
* Logger warning:
  * Vectors aren't identical

I'm not sure whether we want to try to add checks related to listeners.

### Types of change

Enhancement

## Checklist


spacy/language.py
```diff
@@ -682,9 +682,17 @@ def create_pipe_from_source(
...
+        # Check source type
+        if not isinstance(source, Language):
             raise ValueError(Errors.E945.format(name=source_name, source=type(source)))
+        # Check that current lang is a subclass of source lang
+        if not issubclass(type(self), type(source)):
+            warnings.warn(Warnings.W112.format(name=source_name, curr_lang=self.lang, source_lang=source.lang))
+        # Check vectors, with faster checks first
+        if self.vocab.vectors.shape != source.vocab.vectors.shape or \
+                self.vocab.vectors.key2row != source.vocab.vectors.key2row or \
+                self.vocab.vectors.to_bytes() != source.vocab.vectors.to_bytes():
```
ines, line 694: I guess this is the only way but I'm worried this might take too long or that there could be unintended side-effects or false positives? Then again, this would only ever happen once before training, so 🤷‍♀️

adrianeboyd: It's very likely that the other checks will fail first before the slow `to_bytes` check, this only runs once, and it only results in a warning.

It could definitely warn unnecessarily, like you'd added one row to your custom vectors and it's otherwise the same or you're adding a rule-based component from some pipeline without vectors, but I don't think those kinds of checks are feasible. We can't even really inspect whether a component uses the vectors.

adrianeboyd: The vectors one (and maybe also the other one) could be a debug message rather than a python warning?

PR #7528, opened 2021-03-22
Preserve user data for DependencyMatcher on spans

adrianeboyd:
## Description

Preserve user data for DependencyMatcher on spans

### Types of change

Bug fix.

## Checklist


adrianeboyd:
Hmm, while this patch isn't wrong per se, I don't think `span.as_doc(copy_user_data=True)` actually works for a lot of custom extension values because it doesn't adjust the stored offsets. It will work for custom extensions that are purely getters, but not others.

adrianeboyd:
Requires fix from #7574.

PR #7557, opened 2021-03-25
Fix __add__ method of PRFScore

graue70:
## Description

The `__add__` method of `PRFScore` can't be run because of incorrect constructor usage. It isn't used anywhere in the code. Because it would be weird to have `__iadd__` without `__add__`, I decided to fix it instead of removing it.

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -20,10 +20,16 @@ MISSING_VALUES = frozenset([None, 0, ""])
 class PRFScore:
     """A precision / recall / F score."""
 
-    def __init__(self) -> None:
-        self.tp = 0
-        self.fp = 0
-        self.fn = 0
+    def __init__(
+        self,
+        *,
+        tp: Optional[int] = None,
+        fp: Optional[int] = None,
+        fn: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


spacy/scorer.py
```diff
@@ -20,10 +20,16 @@
...
+    def __init__(
+        self,
+        *,
+        tp: Optional[int] = None,
+        fp: Optional[int] = None,
+        fn: Optional[int] = None,
+    ) -> None:
+        self.tp = 0 if tp is None else tp
+        self.fp = 0 if fp is None else fp
+        self.fn = 0 if fn is None else fn
```
svlandeg, line 32: Good catch, and thanks for the fix!

Perhaps we could simplify further?

```suggestion
    def __init__(
        self,
        *,
        tp: Optional[int] = 0,
        fp: Optional[int] = 0,
        fn: Optional[int] = 0,
    ) -> None:
        self.tp = tp
        self.fp = fp
        self.fn = fn
```

svlandeg:
(also really appreciate adding the unit tests!)

PR #7602, opened 2021-03-29
Update debug data further for v3

adrianeboyd:
## Description

* Remove new/existing label distinction (new labels are not immediately distinguishable because the pipeline is already initialized)
* Warn on missing labels in training data for all components except parser
* Separate textcat and textcat_multilabel sections
* Add section for morphologizer

### Types of change

?

## Checklist

### DIFF
--- a/spacy/cli/debug_data.py
+++ b/spacy/cli/debug_data.py
@@ -218,9 +218,9 @@ def debug_data(
         missing_labels = model_labels - labels
         if missing_labels:
             msg.warn(
-                "Some model labels are not present in train data. The model "
-                "performance may be degraded for these labels after training: "
-                f"{_format_labels(missing_labels)}."
+                "Some model labels are not present in the train data. The "
+                "model performance may be degraded for these labels after "
+                f"training: {_format_labels(missing_labels)}."
             )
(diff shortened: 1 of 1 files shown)


spacy/cli/debug_data.py
```diff
@@ -194,32 +196,32 @@ def debug_data(
...
+        ]
+        labels_with_counts = _format_labels(labels_with_counts, counts=True)
+        msg.text(f"Labels in train data: {_format_labels(labels)}", show=verbose)
+        missing_labels = model_labels - labels
+        if missing_labels:
+            msg.warn(
+                "Some model labels are not present in train data. The model "
+                "performance may be degraded for these labels after training: "
+                f"{_format_labels(missing_labels)}."
+            )
```
svlandeg, line 224: Good idea to warn about these!

(do we want to put "in the train data" instead of "in train data"?)

svlandeg (approved):
Looks good to me!

spacy/cli/debug_data.py
adrianeboyd: I think I did that because some other existing message said "in train data", but as a full sentence that would be fine here.

PR #7603, opened 2021-03-29
Make vocab update in get_docs deterministic

Stannislav:
The attribute `DocBin.strings` is a set. In `DocBin.get_docs`
a given vocab is updated by iterating over this set.
Iteration over a python set produces an arbitrary ordering,
therefore vocab is updated non-deterministically.

When training (fine-tuning) a spacy model, the base model's
vocabulary will be updated with the new vocabulary in the
training data in exactly the way described above. After
serialization, the file `model/vocab/strings.json` will
be sorted in an arbitrary way. This prevents reproducible
model training.

## Description

Here's a small code snippet that will reproduce the non-determinism.
```python
# spacy_strings.py
import sys

import spacy
from spacy.training import Corpus
from spacy.util import fix_random_seed


DATASET_PATH = "<some path>/train.spacy"


def run(prefix):
    # This shouldn't affect anything, just for the peace of mind
    fix_random_seed(7)

    nlp = spacy.blank("en")
    corpus = Corpus(DATASET_PATH)

    # This will update nlp's vocab as a side effect
    for _ in corpus(nlp):
        pass

    # Save the vocab to disk
    nlp.vocab.strings.to_disk(f"{prefix}_strings.json")


if __name__ == "__main__":
    run(sys.argv[1])

```
Run this this script twice and compare the results:
```bash
python spacy_strings.py 1
python spacy_strings.py 2
diff -q 1_strings.json 2_strings.json
# Files 1_strings.json and 2_strings.json differ
```
### Types of change
This is a bug fix.

## Checklist

### DIFF
--- a/spacy/strings.pyx
+++ b/spacy/strings.pyx
@@ -223,7 +223,7 @@ cdef class StringStore:
             it doesn't exist. Paths may be either strings or Path-like objects.
         """
         path = util.ensure_path(path)
-        strings = list(self)
+        strings = sorted(self)
         srsly.write_json(path, strings)
 
     def from_disk(self, path):
@@ -247,7 +247,7 @@ cdef class StringStore:
 
         RETURNS (bytes): The serialized form of the `StringStore` object.
         """
-        return srsly.json_dumps(list(self))
+        return srsly.json_dumps(sorted(self))
 
     def from_bytes(self, bytes_data, **kwargs):
         """Load state from a binary string.
--- a/spacy/tests/serialize/test_serialize_vocab_strings.py
+++ b/spacy/tests/serialize/test_serialize_vocab_strings.py
@@ -49,9 +49,9 @@ def test_serialize_vocab_roundtrip_disk(strings1, strings2):
         vocab1_d = Vocab().from_disk(file_path1)
         vocab2_d = Vocab().from_disk(file_path2)
         # check strings rather than lexemes, which are only reloaded on demand
-        assert strings1 == [s for s in vocab1_d.strings]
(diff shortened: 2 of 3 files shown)


svlandeg:
> After serialization, the file `model/vocab/strings.json` will
be sorted in an arbitrary way. This prevents reproducible
model training.

The file might not be serialized exactly the same way, but how does that influence the model training itself? In other words - are you actually seeing a difference in ML performance?

Stannislav:
> Thanks for the PR! I see what you mean.
> 
> > After serialization, the file `model/vocab/strings.json` will
> > be sorted in an arbitrary way. This prevents reproducible
> > model training.
> 
> The file might not be serialized exactly the same way, but how does that influence the model training itself? In other words - are you actually seeing a difference in ML performance?

Ah, sorry, I didn't explain it properly.

I meant that the model checkpoint will have a different hash. So it's not obvious that the model is actually the same. This also breaks any hash-based tracking like DVC.

svlandeg:
Ah, gotcha, thanks for clarifying!

I think the proposed edit looks fine. I guess it might cause a small delay when serializing, but it hopefully shouldn't be too bad? I do see the added advantage of ensuring model hash equality...

I'll leave this open a bit to allow my colleagues to chime in ;-)

polm:
It looks good to me. I think having consistent output has a number of advantages.

Since vocabularies can be large I was concerned about extra memory use, but since the list of strings is a set I'm not sure that can be improved easily.

honnibal:
I think we should make the `strings` variable a different data structure, probably a dict? I'm worried that the sorting can easily make a performance difference.

PR #7621, opened 2021-03-31
Terminology: deprecated vs obsolete

BramVanroy:
Typically, deprecated is used for functionality that is bound to become unavailable but that can still be used. Obsolete is used for features that have been removed. In E941, I think what is meant is "obsolete" since loading a model by a shortcut simply does not work anymore (and throws an error). This is different from downloading a model with a shortcut, which is deprecated but still works.

In light of this, perhaps all other error codes should be checked as well.

### Types of change
Change in error code

## Checklist

### DIFF
--- a/spacy/cli/__init__.py
+++ b/spacy/cli/__init__.py
@@ -29,9 +29,9 @@ from .project.document import project_document  # noqa: F401
 
 @app.command("link", no_args_is_help=True, deprecated=True, hidden=True)
 def link(*args, **kwargs):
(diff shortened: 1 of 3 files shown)


svlandeg:
Hi Bram,

Thanks for the PR. You're right that we haven't been using the terminology entirely consistently.
The changes you propose are fine and can be merged. I've also gone through the other warning/error codes, and I think those are all OK, because of many of the functionality is in fact deprecated but not yet removed/obsolete.

I've made a few additional small changes to some documentation, but then I think this can be merged.

PR #7293, opened 2021-03-05
Set up GPU CI testing

adrianeboyd:
## Description

* Set up GPU CI testing
* Update tests for enabled GPU
* Fix minor GPU-related bugs

### Types of change

?

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [ ] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #7626, opened 2021-03-31
Fix vectors data on GPU

svlandeg:
## Description
 `vectors.data` can be of type `numpy.ndarray` or `cupy.ndarray`, but [recent GPU testing](https://github.com/explosion/spaCy/pull/7293) highlighted some trouble when using `cupy`. This PR makes the code more general by relying on current ops, and also avoids using the implicit conversion to scalar by `numpy` (which `cupy` doesn't do).

Tested locally, this seemed to fix the trouble from PR #7293 🤞 

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -361,8 +361,10 @@ cdef class Vectors:
         # Account for numerical error we want to return in range -1, 1
         scores = xp.clip(scores, a_min=-1, a_max=1, out=scores)
         row2key = {row: key for key, row in self.key2row.items()}
+
+        numpy_rows = get_current_ops().to_numpy(best_rows)
         keys = xp.asarray(
-            [[row2key[int(row)] for row in best_rows[i] if int(row) in row2key]
+            [[row2key[row] for row in numpy_rows[i] if row in row2key]
                     for i in range(len(queries)) ], dtype="uint64")
         return (keys, best_rows, scores)
 
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -55,7 +55,7 @@ cdef class Vectors:
         """Create a new vector store.
 
         shape (tuple): Size of the table, as (# entries, # columns)
-        data (numpy.ndarray): The vector data.
+        data (numpy.ndarray or cupy.ndarray): The vector data.
         keys (iterable): A sequence of keys, aligned with the data.
         name (str): A name to identify the vectors table.
 
@@ -65,7 +65,8 @@ cdef class Vectors:
         if data is None:
             if shape is None:
                 shape = (0,0)
-            data = numpy.zeros(shape, dtype="f")
+            ops = get_current_ops()
+            data = ops.xp.zeros(shape, dtype="f")
         self.data = data
         self.key2row = {}
         if self.data is not None:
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I can't seem to add this as a suggestion because it's too far from the existing edits, but after running the GPU tests with all the suggested changes I still run into problems with `Vectors.add`, I think this PR also needs something like this:

```diff
         if vector is not None:
+            xp = get_array_module(self.data)
+            vector = xp.asarray(vector)
             self.data[row] = vector
```

svlandeg:
Hm, maybe? I tried to ensure that that wouldn't be necessary in the edits, but we can add it for robustness anyway.

adrianeboyd:
Maybe I did something wrong, but there are a number of tests that add numpy arrays as the new vector row, which leads to `ValueError: non-scalar numpy.ndarray cannot be used for fill`, like this:

https://github.com/explosion/spaCy/blob/8b3eec6e62122ece5ff5188a9fece06af4f46b92/spacy/tests/doc/test_doc_api.py#L235-L240

I don't think we want to require people using `Vector.add` to have to provide the new vector with the right ops?

svlandeg:
Oh shoot, I thought I tracked down most of the tests and changed them to use current ops.
Anyway I already made the change you suggested - does that fix it?

spacy/vocab.pyx
```diff
@@ -332,8 +332,8 @@ cdef class Vocab:
         remap = {}
         for i, key in enumerate(keys[nr_row:]):
             self.vectors.add(key, row=syn_rows[i][0])
-            word = self.strings[key]
-            synonym = self.strings[syn_keys[i][0]]
+            word = self.strings[int(key)]
```
honnibal, line 335: If this is designed to copy data from cupy to numpy, then this will be very slow, because we'll be syncing each element one-by-one. We should call `keys.get()` on the whole array (or slice that we need) before the loop.

svlandeg: I need to dive into the detail again, but I think the `int()` bit was required when running one of the tests from Adriane's PR On GPU, because `key` would be a cupy array with just one element that doesn't get automatically converted to a singe element like numpy does.

spacy/vectors.pyx
```diff
@@ -357,8 +361,10 @@ cdef class Vectors:
         # Account for numerical error we want to return in range -1, 1
         scores = xp.clip(scores, a_min=-1, a_max=1, out=scores)
         row2key = {row: key for key, row in self.key2row.items()}
+
+        numpy_rows = get_current_ops().to_numpy(best_rows)
```
adrianeboyd, line 365: `xp` might not be the same as `get_current_ops`, I think?

adrianeboyd: Nevermind, it shouldn't matter...

svlandeg: This could only go wrong if for some reason `vectors.data` is on GPU while the current ops is `NumpyOps` - I don't really see how that could happen? Still, we could call `CupyOps.to_numpy()` directly, which will always work.

adrianeboyd: You can totally have:

```python
spacy.require_gpu()
nlp = spacy.load("model")
spacy.require_cpu()
nlp.do_something()
```

The thinc models should keep working in this context. (The torch ones don't because of the default tensor type.)

It probably doesn't come up too often, if it's just the rows, then maybe `xp.asarray()` would be better? (I don't know how much data would theoretically be copied here?)

adrianeboyd: (Wait, that's nonsense.)

adrianeboyd: `to_numpy` is fine, it's in the base `Ops`.

svlandeg: Yes, but the base ops only works on numpy arrays. I'm confused now whether this issue is resolved or not :p

adrianeboyd: Argh, I just keep confusing myself, sorry.

PR #7629, opened 2021-03-31
Fix scoring normalization

svlandeg:
## Description
During training, the weights for scoring the components are normalized by the number of components, and set to 1.0 in total for each component. 
However, if there's a component that isn't being scored (e.g. all 0 and/or `None`), then that component probably shouldn't count towards the total. Otherwise you'd define something like
```
[training.score_weights]
tag_acc = 0.0
ents_f = 1.0
ents_p = 0.0
ents_r = 0.0
```
And the scoring would have your entity F halved, which is quite unexpected:
```
ℹ Pipeline: ['tok2vec', 'tagger', 'ner']
ℹ Initial learn rate: 0.001
E    #       LOSS TOK2VEC  LOSS TAGGER  LOSS NER  TAG_ACC  ENTS_F  ENTS_P  ENTS_R  SCORE
---  ------  ------------  -----------  --------  -------  ------  ------  ------  ------
  0       0          0.00        85.80     36.77    23.40    0.00    0.00    0.00    0.00
  0     200      17041.98      7663.22   1446.57    89.49   12.78   25.00    8.59    0.06

```
So instead, I propose to divide by the number of components that are actually being weighted, cf with this PR:
```
ℹ Pipeline: ['tok2vec', 'tagger', 'ner']
ℹ Initial learn rate: 0.001
E    #       LOSS TOK2VEC  LOSS TAGGER  LOSS NER  TAG_ACC  ENTS_F  ENTS_P  ENTS_R  SCORE
---  ------  ------------  -----------  --------  -------  ------  ------  ------  ------
  0       0          0.00        85.80     36.77    23.40    0.00    0.00    0.00    0.00
  0     200      17041.98      7663.22   1446.57    89.49   12.78   25.00    8.59    0.13

```

I also added a few unit tests with the `overrides` argument of `combine_score_weights`. The GH diff reads a bit weird, but I haven't actually changed any existing tests.

### Types of change
(bug) fix

## Checklist


honnibal:
Wouldn't we want to normalize by the total normalization weight, not the number of components?

svlandeg:
I agree that behaviour feels more intuitive - I changed the implementation & unit tests accordingly.

PR #7634, opened 2021-04-01
Remove torino from stop words for Italian

jacopofar:
I stumbled upon the fact that "torino" is in the list of stop words for Italian. Torino is the proper name of a city and the token has no other common meanings.
I think the inclusion is unintended since most likely people want it to not be ignored and may even want a NER tagger to detect it.

### Types of change
Bug fix

## Checklist

- [ ] I ran the tests, and all new and existing tests passed. (I didn't, as it's a small dataset change)


svlandeg (approved):
Makes sense, thanks for the PR! 

Perhaps we want to wait until 3.1 to merge this, because conceptually changing the stop words can change people models, though this particular case seems like a clear change we need to make.

PR #7650, opened 2021-04-04
Add Numerizer to SpaCy universe

jaidevd:
Numerizer is a spaCy extension that converts numbers written in natural language
into numeric strings. This commit adds Numerizer to `website/meta/universe.json`.

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -1,5 +1,27 @@
 {
     "resources": [
+	{
+	    "id": "numerizer",
+	    "title": "numerizer",
+	    "slogan": "Convert natural language numerics into ints and floats.",
+	    "description": "A SpaCy extension for Docs, Spans and Tokens that converts numerical words and quantitative named entities into numeric strings.",
+	    "github": "jaidevd/numerizer",
+	    "pip": "numerizer",
+	    "code_example": [
+		"from spacy import load",
+		"import numerizer",
+		"nlp = load('en_core_web_sm') # or any other model",
+		"doc = nlp('The Hogwarts Express is at platform nine and three quarters')",
+		"doc._.numerize()",
+		"# {nine and three quarters: '9.75'}"
+	    ],
+	    "author": "Jaidev Deshpande",
+	    "author_links": {
+		"github": "jaidevd",
+		"twitter": "jaidevd"
+	    },
+	    "category": ["standalone"]
+	},
         {
             "id": "spikex",
             "title": "SpikeX - SpaCy Pipes for Knowledge Extraction",


website/meta/universe.json
```diff
@@ -1,5 +1,27 @@
...
+	    "id": "numerizer",
+	    "title": "numerizer",
+	    "slogan": "Convert natural language numerics into ints and floats.",
+	    "description": "A SpaCy extension for Docs, Spans and Tokens that converts numerical words and quantitative named entities into numeric strings.",
+	    "github": "jaidevd/numerizer",
+	    "pip": "numerizer",
+	    "code_example": [
+		"from spacy import load",
+		"import numerizer",
+		"nlp = load('en_core_web_sm') # or any other model",
```
svlandeg, line 13: While this reads "or any other model", the code at https://github.com/jaidevd/numerizer/blob/master/numerizer/numerizer.py has 
```
try:
    import spacy
    SPACY_INSTALLED = True
except ImportError:
    SPACY_INSTALLED = False
```
which sort of implies a dependency on `en_core_web_sm`. I think that may be confusing for users?

```

File: website/meta/universe.json [hunk 2]
```diff
@@ -1,5 +1,27 @@
...
+		"doc = nlp('The Hogwarts Express is at platform nine and three quarters')",
+		"doc._.numerize()",
```
svlandeg, line 15: I haven't looked in the code in detail, but I'm not sure I understand the API here. The `numerize()` method is called on spaCy's `Underscore` method? Why not add a [custom pipeline](https://spacy.io/usage/processing-pipelines#custom-components) to `nlp` instead that sets a [custom attribute](https://spacy.io/usage/processing-pipelines#custom-components-attributes)?

svlandeg (commented):
Thanks for the PR! I think a lot of people will find this quite a useful extension :-)

Just out of curiosity I was looking a bit at the details of the implementation and had some small questions/remarks I was wondering about :-)

website/meta/universe.json
jaidevd: Ouch! It turns out that that line is unnecessary. I only need the `Doc`, `Token` and `Span` classes to exist - no need to have a pipeline there. I'll remove it and notify you. Thanks for pointing this out!

website/meta/universe.json
jaidevd: I found calling a `.set_extension` on the `Doc` - and specifying it's methods / getters / setters a lot easier than adding a custom pipeline component. Additionally, it seems like adding a custom pipeline component set's an extra step for the user (calling `.add_pipe` before using the pipeline.) I thought that this way, all one has to do is `import numerizer` and their extension is ready.

Is this a valid assessment? If not, I can look into the custom component.

svlandeg: Ultimately the choice is up to you, ofcourse, as you're designing the plugin. I just felt like many spaCy users would be looking to add a custom component from your library, as it seems like the "standard approach". I was actually confused myself for a second that that's not happening. But like I said, the choice of UI is yours, just wanted to share my 2 cents ;-)

svlandeg: I'll merge this as-is for now, in case you'd ever want to change the API on your side, you can always send a new PR with an update!

jaidevd:
Thanks a lot, @svlandeg! And thanks especially for (even briefly) going through the code! I'll update this soon.

PR #7674, opened 2021-04-06
Replace negative rows with 0 in StaticVectors

adrianeboyd:
## Description

Replace negative row indices with 0-vectors in `StaticVectors`.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/ml/staticvectors.py
+++ b/spacy/ml/staticvectors.py
@@ -43,7 +43,7 @@ def forward(
         [doc.vocab.vectors.find(keys=doc.to_array(key_attr)) for doc in docs]
     )
     try:
-        vectors_data = model.ops.gemm(model.ops.as_contig(V[rows]), W, trans2=True)
+        vectors_data = model.ops.gemm(model.ops.as_contig(V[rows]) * (rows>=0).T[:, None], W, trans2=True)
     except ValueError:
         raise RuntimeError(Errors.E896)
     output = Ragged(
--- a/spacy/ml/staticvectors.py
+++ b/spacy/ml/staticvectors.py
@@ -8,7 +8,7 @@ from ..tokens import Doc
 from ..errors import Errors
 
 
-@registry.layers("spacy.StaticVectors.v1")
+@registry.layers("spacy.StaticVectors.v2")
 def StaticVectors(
     nO: Optional[int] = None,
     nM: Optional[int] = None,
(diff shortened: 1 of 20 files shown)


adrianeboyd:
Since the current models were trained with this behavior and this changes their predictions, I guess this should wait until v3.1.

spacy/ml/staticvectors.py
```diff
@@ -43,7 +43,7 @@ def forward(
         [doc.vocab.vectors.find(keys=doc.to_array(key_attr)) for doc in docs]
     )
     try:
-        vectors_data = model.ops.gemm(model.ops.as_contig(V[rows]), W, trans2=True)
+        vectors_data = model.ops.gemm(model.ops.as_contig(V[rows]) * (rows>=0).T[:, None], W, trans2=True)
```
honnibal, line 46: I think then following will introduce less copying, and should allow us to merge this less obtrusively.

```suggestion
        vectors_data = model.ops.as_contig(V[rows])
        # If any rows have a negative index, set the values of those vectors to 0.
        # This is placed behind an attr flag, so that we can merge this change prior to
        # v3.1 if we want. We could set the attr flag in the init function, which would 
        # introduce the behavior on models that are trained, but not old models that
        # are loaded.
        if model.attrs.get("set_negative_indices_to_zero"):
            vectors_data[rows < 0] = 0
        vectors_data = model.ops.gemm(vectors_data, W, trans2=True, out=vectors_data)
```

The implementation takes the data and makes it contiguous, and then sets the values of the negative rows to 0, before doing the multiplication.

adrianeboyd: Hmm, I think the flag is excessive and just complicates the transition to v3.1. Looking at it again, maybe it would be better to just call the whole thing with different behavior `spacy.StaticVectors.v2`?

Because of how new vectors are added, negative indices should never be used to index into the table, so I don't think it's a sensible option to make it sound like negative indices should be indices. I could see adding an option for a positive index as `UNK` or something like that, but I don't like the sound of this option.

honnibal: Hmm yeah okay, the flag is a dumb idea. A v2 version would be better.

adrianeboyd: Okay, to clarify: for now, just call it v2 with the 0-ing behavior and then if we want other options like an `UNK` row we can add that in the future?

adrianeboyd: Ugh, this is messier than I thought since the architectures use the `StaticVectors` class directly.

adrianeboyd:
This is now waiting on https://github.com/explosion/spacy-legacy/pull/5, but that PR is :cry:.

All these things are versioned, but I don't think there's any way to refer to the versions in the architecture implementations, so I find this all extremely brittle and displeasing: Say, here:

https://github.com/explosion/spacy-legacy/pull/5/files#diff-8153494fc386aa9e391d530c783bf2cf54bbe7160125c7f0763842cf80a00e3dR128-R141

Where it's referring to `build_Tok2Vec_model` and `MaxoutWindowEncoder` from the standard library unversioned, so if those change in the future this all breaks.

adrianeboyd:
As a note, adding legacy functions becomes less brittle if every single layer and architecture is pulled from the registry with a specific version when it's moved the `spacy-legacy`, e.g.:

https://github.com/explosion/spacy-legacy/blob/f5855a8eb023fdfc71bb5ad82724da111970070a/spacy_legacy/architectures/tok2vec.py#L183-L186

Also note that layers need to registered under the `thinc_layers` entry point rather than `spacy_layers` (which is accessible through `spacy.util.registry.layers` but doesn't exist as a named entry point):

https://github.com/explosion/spacy-legacy/blob/f5855a8eb023fdfc71bb5ad82724da111970070a/setup.cfg#L45-L46

PR #7738, opened 2021-04-10
Default code for Setting Entity annotations on the website errors

armsp:
## Description
To the uninitiated, the example at https://spacy.io/usage/linguistic-features#setting-entities errors on Binder because the code tries to set the entity for "fb" twice.
The exact error as reported is -
```
ValueError                                Traceback (most recent call last)
<ipython-input-1-73c2a9e8d913> in <module>
     15
     16 # Option 2: Assign a complete list of ents to doc.ents
---> 17 doc.ents = list(doc.ents) + [fb_ent]
     18
     19 ents = [(e.text, e.start, e.end, e.label_) for e in doc.ents]

/srv/conda/envs/notebook/lib/python3.7/site-packages/spacy/tokens/doc.pyx in spacy.tokens.doc.Doc.ents.__set__()

/srv/conda/envs/notebook/lib/python3.7/site-packages/spacy/tokens/doc.pyx in spacy.tokens.doc.Doc.set_ents()

ValueError: [E1010] Unable to set entity information for token 0 which is included in more than one span in entities, blocked, missing or outside.
```

People short on time or coming here the first time or just beginners may assume that this is
a shortcoming of the library when they are supposed to actually try one of the two options.

So I have commented Option 2 by default, and added a comment adjacent to it stating that
comment option 1 before running option 2.

### Types of change
Very basic change to the documentation

## Checklist

The changes are so basic that they do not require any tests to be run.

Feel free to ignore this PR if you think it is not important.

### DIFF
--- a/.github/contributors/armsp.md
+++ b/.github/contributors/armsp.md
@@ -98,9 +98,9 @@ mark both statements:
 
 | Field                          | Entry                |
 |------------------------------- | -------------------- |
-| Name                           |  Shantam             |
+| Name                           |  Shantam Raj         |
 | Company name (if applicable)   |                      |
 | Title or role (if applicable)  |                      |
-| Date                           |   21/5/2018          |
+| Date                           |   10/4/2021          |
 | GitHub username                |     armsp            |
-| Website (optional)             |                      |
+| Website (optional)             |https://shantamraj.com|
--- a/website/docs/usage/linguistic-features.md
+++ b/website/docs/usage/linguistic-features.md
@@ -601,12 +601,13 @@ print('Before', ents)
 
 # Create a span for the new entity
 fb_ent = Span(doc, 0, 1, label="ORG")
+orig_ents = list(doc.ents)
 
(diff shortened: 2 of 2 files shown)


svlandeg:
Hi! Thanks for the report, you're right that this might confuse users if they just try to run the code as is. The idea is that one of the two options should be picked, but the code does crash if you accidentally run both.

Your solution works, but I personally thinks it looks a bit odd having a section "Option 2" with code that is commented out. What about if we do this instead:

```
fb_ent = Span(doc, 0, 1, label="ORG")
orig_ents = list(doc.ents)

# Option 1
...

# Option 2
doc.ents = orig_ents + [fb_ent]
```
i.e. we work with a temporary variable `orig_ents` that is defined BEFORE option 1 runs, and thus ensures that option 1 and 2 don't actually conflict if you run them both.

armsp:
@svlandeg I like this idea better too 😅
I will submit another commit to this PR with the necessary changes.

svlandeg:
Great, thanks again!

PR #7750, opened 2021-04-12
Add callback to copy vocab/tokenizer from model

adrianeboyd:
## Description

Add callback `spacy.copy_from_base_model.v1` to copy the tokenizer settings and/or vocab (including vectors) from a base model.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -499,6 +499,9 @@ class Errors:
     E202 = ("Unsupported alignment mode '{mode}'. Supported modes: {modes}.")
 
     # New errors added in v3.x
+    E872 = ("Unable to copy tokenizer from base model due to different "
+            'tokenizer settings: current tokenizer config "{curr_config}" '
+            'vs. base model "{base_config}"')
     E873 = ("Unable to merge a span from doc.spans with key '{key}' and text "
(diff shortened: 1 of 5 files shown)


adrianeboyd:
In terms of naming and exact functionality, this can be viewed mainly as a proposal / sketch, but I think it would be useful to offer something along these lines as a provided callback.

See #7582 and #7740 for related discussions.

It would also be included here, replacing the `vectors` override with a more complete copy of the relevant info out of the model that's being updated:

https://github.com/explosion/projects/blob/9d5fce5f95ddf5f35c3370b2074b25e995525f51/pipelines/ner_demo_update/scripts/create_config.py

The main reason to copy the whole vocab and not just vectors is lookups tables like lexeme norms and probabilities.

adrianeboyd:
I'm still not quite sure I like this. Would something like this be better?

```ini
[initialize.before_init]
@callbacks = "spacy.copy_from_base_model.v1"
tokenizer = "model_name" # Optional[str]
vocab = "model_name"
```

I don't like the `base_model` parameter name and although I can't think of a sensible case where you'd want to do this from two different models, maybe you might? (Or you can write your own callback, of course.)

PR #7767, opened 2021-04-13
Add training option to set annotations on update

adrianeboyd:
## Description

Add a `[training]` option called `set_annotations_on_update` to specify a list of components for which the predicted annotations should be set on `example.predicted` immediately after that component has been updated. The predicted annotations can be accessed by later components in the pipeline during the processing of the batch in the same `update` call.

### Types of change

Enhancement.

## Checklist


spacy/default_config.cfg
```diff
@@ -80,6 +80,8 @@ eval_frequency = 200
 score_weights = {}
 # Names of pipeline components that shouldn't be updated during training
 frozen_components = []
+# Names of pipeline components that should set annotations on update
+set_annotations_on_update = []
```
svlandeg, line 84: is "on update" clear for most users? Because the config hides some of the API calls, should we consider calling this something more intuitive like "during training"? Not that I want to make this any longer :|

honnibal: It's within the training block, so how about calling it `annotates`? It feels like that matches up with `frozen` better. If that's too terse, perhaps `set_annotations`, but that looks like it should be a boolean.

svlandeg: or "overrides_gold"?

svlandeg:
This PR still doesn't allow you to run a rule-based component as part of the training process, and use those predictions, right? I'm thinking e.g. entity ruler + NEL... 

The workaround is to preprocess the training data with those rule-based components, but I think it could be nice if the training loop could provide this functionality too?

adrianeboyd:
No, it does. See the test with the sentencizer.

spacy/default_config.cfg
adrianeboyd: `set_annotations_on_update` is intentional clunky and easy to rename.

I agree that `update` is unclear from the `[training]` block/script perspective. I know that `frozen_components` turns into `exclude`, but this feels weird. This needs to be the option for `nlp.update`, too, and it'd be nice to have the same for both?

Something with `predicted` would be better. Something along the lines of `annotate_predicted`?

spacy/language.py
```diff
@@ -1098,22 +1100,35 @@ def update(
...
+            if name not in exclude and hasattr(proc, "update"):
+                proc.update(examples, sgd=None, losses=losses, **component_cfg[name])
+            if sgd not in (None, False):
                 if (
                     name not in exclude
                     and hasattr(proc, "is_trainable")
                     and proc.is_trainable
                     and proc.model not in (True, False, None)
                 ):
                     proc.finish_update(sgd)
```
adrianeboyd, line 1119: @honnibal: Does running `finish_update` incrementally rather than for all components at the end at once have any side effects I might not be aware of?

spacy/default_config.cfg
svlandeg: `persist_predictions`? (I realise I'm making things worse 😁 )

adrianeboyd: `predictions` is bad, if we use it I want it to sound like `Example.predicted`.

adrianeboyd: Planning to rename to `annotating_components` in `[training]` and `annotates` for `Language.update`.

PR #7783, opened 2021-04-14
assemble CLI command

svlandeg:
## Description
With the recent shift to using config files for training, I think it makes sense to have an additional `assemble` command you can use just to create & initialize a pipeline when there are no components to actually train. You could mimick this by "training" just for one iteration, but that's a little silly.

cf https://github.com/explosion/spaCy/discussions/7688 for some background discussion.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/cli/assemble.py
+++ b/spacy/cli/assemble.py
@@ -1,0 +1,57 @@
+from typing import Optional
+from pathlib import Path
+from wasabi import msg
+import typer
+import logging
+
+from ._util import app, Arg, Opt, parse_config_overrides, show_validation_error, create_before_to_disk_callback
+from ._util import import_code
+from ..training.initialize import init_nlp
+from .. import util
+from ..util import registry
+from ..schemas import ConfigSchemaTraining
+
+
+@app.command(
+    "assemble",
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
+def assemble_cli(
+    # fmt: off
+    ctx: typer.Context,  # This is only used to read additional arguments
+    config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
+    output_path: Path = Arg(..., help="Output directory to store assembled pipeline in"),
+    code_path: Optional[Path] = Opt(None, "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    verbose: bool = Opt(False, "--verbose", "-V", "-VV", help="Display more information for debugging purposes"),
+    # fmt: on
+):
+    """
+    Assemble a spaCy pipeline from a config file. The config file includes
+    all settings for initializing the pipeline. To override settings in the
+    config, e.g. settings that point to local paths or that you want to
+    experiment with, you can override them as command line options. The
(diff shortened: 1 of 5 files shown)


spacy/cli/assemble.py
```diff
@@ -0,0 +1,57 @@
...
+        config = util.load_config(config_path, overrides=overrides, interpolate=False)
+    msg.divider("Initializing pipeline")
+    with show_validation_error(config_path, hint_fill=False):
+        nlp = init_nlp(config)
+    msg.good("Initialized pipeline")
+    msg.divider("Serializing to disk")
+    config = nlp.config.interpolate()
+    T = registry.resolve(config["training"], schema=ConfigSchemaTraining)
+    before_to_disk = create_before_to_disk_callback(T["before_to_disk"])
+    before_to_disk(nlp).to_disk(output_path)
```
svlandeg, line 57: Not sure whether or not we should call `before_to_disk` which is technically defined in the `training` block and not in the `nlp` block...

adrianeboyd: I would lean towards no? Is there anything that couldn't be managed with the `after_creation` callback?

Along the same lines, I'm not sure I'd want to use `init_nlp` here? There are lots of `[training]` settings you don't need like the corpus. Is it possible to do this entirely without the `[training]` block? That seems like it would be useful and closer to what a "collate" script is doing.

`assemble` is annoying if you have to set up toy corpora to run it even though you're not training.

svlandeg: I agree - conceptually this should work if the entire `training` block is missing from the config.

adrianeboyd: I think it should basically just do this one line:

https://github.com/explosion/spaCy/blob/ed561cf428494c2b7a6790cd4b91b5326102b59d/spacy/training/initialize.py#L40

There are going to be headaches with initializing rule-based components that load stuff, though, hmm...

adrianeboyd: I guess you can just require them to be saved/sourced and initialized elsewhere, which is more what `assemble` is for anyway. Or you can run `initialize` for non-frozen components with an empty `get_examples` or something, but users are going to have trouble with that.

svlandeg: I'm not sure I fully understand what you mean, Adriane. The main motivation for this additional command would be that the usage would be very similar to `spacy train`, including the initialization and argument overrides. Imagine having a config file with rule-based components with custom initialization, and then an NER model to train. Basically if you take out the NER from that config file, you'd still want the same easy process to initialize & assemble all the other components in the pipeline.

I do agree that this should work without specifying the `[training]` block. I just tested the current code as such, and that runs fine. The disadvantage is that now we're copying some code over from `init_nlp`, but I think this is acceptable...

svlandeg: There's also no need for toy corpora or such. `nlp.initialize` will create some dummy objects, and you'd assume that the non-trainable custom components won't rely on training data being fed in.

adrianeboyd: I think it looks good now!

Issue #7732, opened 2021-04-09
spacy ray command missing when run via entrypoint


kinghuang:
## How to reproduce the behaviour

I installed the `spacy-ray` module, and expected to find a `spacy ray` command. But, it seems the `ray` command isn't available when the `spacy` CLI is run through its entrypoint script versus calling the module via the `python` CLI.

**`spacy` entrypoint:**
```
root@271ba8192ed4:/project/ner# spacy ray --help
Usage: spacy [OPTIONS] COMMAND [ARGS]...
Try 'spacy --help' for help.

Error: No such command 'ray'.
```

**`spacy` module via `python`:**
```
root@271ba8192ed4:/project/ner# python -m spacy ray --help
Usage: python -m spacy ray [OPTIONS] COMMAND [ARGS]...

  CLI for parallel and distributed computing via Ray. See the Ray
  documentation for details: https://ray.io.

Options:
  --help  Show this message and exit.

Commands:
  train  Train a spaCy pipeline using Ray for parallel training.
```

## Your Environment

* Operating System: Debian 10 (python container)
* Python Version Used: 3.8
* spaCy Version Used: 3.0.5
* Environment Information:

svlandeg:
Thanks for the detailed report! I could replicate this, and PR #7799 should fix this.

PR #7799, opened 2021-04-15
ensure 'spacy ray' works

svlandeg:
Fixes https://github.com/explosion/spaCy/issues/7732

## Description
If `spacy-ray` is installed, it only works when calling `python -m spacy ray`. With this fix, it also works with just `spacy ray` - cf https://github.com/explosion/spaCy/issues/7732 for details.

### Types of change
fix/enhancement

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,7 +62,7 @@ install_requires =
 
 [options.entry_points]
 console_scripts =
-    spacy = spacy.cli:app
+    spacy = spacy.cli:setup_cli
 
 [options.extras_require]
 lookups =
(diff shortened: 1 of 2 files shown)


svlandeg:
This is a little annoying because I think we wanted to avoid an explicit import :|

adrianeboyd:
Yeah, I was hoping Ines might have a good idea because it would be nice if this worked more generally, but I didn't see any obvious solution, either.

svlandeg:
Put in a question here: https://github.com/tiangolo/typer/issues/273

svlandeg:
Updated this PR after advice from @kludex : rerouting the console script for `spacy` to `setup_cli` so this function gets executed, just as it would when spacy's [`__main__`](https://github.com/explosion/spaCy/blob/master/spacy/__main__.py) is called.

I had the reinstall to see the change, but then it works, and `spacy --help` shows the `ray` command if it's available :-)

Kludex (approved):

PR #7826, opened 2021-04-19
Add entry for Coreferee project to universe.json

richardpaulhudson:
Adding an entry for the Coreferee project to universe.json.

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -2224,6 +2224,39 @@
             }
         },
         {
+            "id": "coreferee",
+            "title": "Coreferee",
+            "slogan": "Coreference resolution for multiple languages",
+            "github": "msg-systems/coreferee",
+            "url": "https://github.com/msg-systems/coreferee",
+            "description": "Coreferee is a pipeline plugin that performs coreference resolution for English, German and Polish. It is designed so that it is easy to add support for new languages and optimised for limited training data. It uses a mixture of neural networks and programmed rules. Please note you will need to [install models](https://github.com/msg-systems/coreferee#getting-started) before running the code example.",
+            "pip": "coreferee",
+            "category": ["pipeline", "models", "standalone"],
+            "tags": ["coreference-resolution", "anaphora"],
+            "code_example": [
+                "import coreferee, spacy",
+                "nlp = spacy.load('en_core_web_trf')",
+                "nlp.add_pipe('coreferee')",
+                "doc = nlp('Although he was very busy with his work, Peter had had enough of it. He and his wife decided they needed a holiday. They travelled to Spain because they loved the country very much.')",
+                "doc._.coref_chains.print()",
+                "# Output:",
+                "#",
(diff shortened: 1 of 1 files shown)


richardpaulhudson:
Presenting Coreferee, a coreference resolution solution that works with English, German and Polish and is designed to make it easy to add further languages. The main aim of the project was to get coreference resolution working cross-linguistically, but some people may also find the English functionality useful as Neuralcoref no longer seems to support the latest spaCy models. Best wishes, Richard

honnibal:
This is amazing! We're working on a neural coreference model, patterned largely after the Kenton Lee et al. model that does joint mention detection and resolution.

The algorithm you have here looks a bit more like the Haghigi et al style where there's a cascade of filters. Is it based on any particular paper, or did you just experiment and come up with a model that made sense on your data?

richardpaulhudson:
Because the focus was on getting things working quickly cross-linguistically, I didn’t base the algorithm on any specific research. Instead, I hardcoded the grammatical rules that affect coreference and especially anaphora resolution for each language (rules.py and annotation.py) to reduce the number of mentions that needed to be considered for each anaphor. I then gathered all the information that was readily available from the spaCy components further up the pipeline and that I thought could possibly be relevant based on my knowledge of pragmatics and used them as inputs to a multilayer perceptron (tendencies.py).

I then experimented and discovered that (unsurprisingly given the limited training data) using an ensemble increased the accuracy and the stability of the results; that four seemed to be the optimal number of training epochs; and that using vector squeezers (see https://github.com/msg-systems/coreferee#the-neural-ensemble) led to better results than feeding the vectors in at full width. However, there are all sorts of other things, including the activation functions, which I had no time to investigate (and my research budget is unfortunately at an end now :-)).

What I imagine is more likely to be of general interest than the neural network algorithm is the contribution of the rules. In many cases, they reduce the number of mentions that have to be considered for an anaphor to a point where there is only one contender left — or at least only one serious contender. There is no quick way for me to try out the latest state-of-the-art algorithms, but Neuralcoref, certainly (although it amazing at finding complex coreferences between noun phrases that Coreferee does not even attempt to capture), given a text like ‘Peter talked to John. He was pleased to see him’, suggests a chain ‘Peter – He – him’. The fact that ‘he’ and ‘him’ cannot possibly corefer is easily established with a few lines of code that are almost cross-linguistically valid; not to use them seems to me to be to ignore low-hanging fruit. To put it another way, why ask a machine to learn what you can easily tell it? The more you can tell it, the more effectively it will be able to learn what is left.

The rules-based and machine-learning-based parts of Coreferee are minimally coupled. Perhaps it would make sense to consider using something based on the rules-based part within spaCy in conjunction with the algorithm you are already developing? (The project is Apache 2 because of issues with the word lists and models, but I could give you any of the Python modules under MIT.)

One open issue is that I have not yet been able to benchmark Coreferee. Do you have any ideas how this might be possible given that the range of entities Coreferee is attempting to find is probably not exactly the same as any of the benchmarks?

PR #7839, opened 2021-04-20
update EL training data format in docs

svlandeg:
## Description
As pointed out in https://github.com/explosion/spaCy/discussions/7817, the EL training data format from the docs was not complete and could be confusing, as the EL needs gold sentences & entities. 

### Types of change
doc fix

## Checklist

### DIFF
--- a/website/docs/api/data-formats.md
+++ b/website/docs/api/data-formats.md
@@ -455,7 +455,7 @@ example = Example.from_dict(doc, gold_dict)
 doc = nlp("Russ Cochran his reprints include EC Comics.")
 gold_dict = {"entities": [(0, 12, "PERSON")],
              "links": {(0, 12): {"Q7381115": 1.0, "Q2146908": 0.0}},
-             "sent_starts": [1, -1, 0, 0, 0, 0, 0, 0]})}
+             "sent_starts": [1, -1, 0, 0, 0, 0, 0, 0]}
 example = Example.from_dict(doc, gold_dict)
 ```
 
--- a/website/docs/api/data-formats.md
+++ b/website/docs/api/data-formats.md
@@ -451,9 +451,11 @@ doc = nlp("I'm pretty happy about that!")
 gold_dict = {"cats": {"POSITIVE": 1.0, "NEGATIVE": 0.0}}
 example = Example.from_dict(doc, gold_dict)
 
-# Training data for an Entity Linking component
+# Training data for an Entity Linking component (also requires entities & sentences)
 doc = nlp("Russ Cochran his reprints include EC Comics.")
-gold_dict = {"links": {(0, 12): {"Q7381115": 1.0, "Q2146908": 0.0}}}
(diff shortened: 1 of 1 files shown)


website/docs/api/data-formats.md
```diff
@@ -451,9 +451,11 @@ doc = nlp("I'm pretty happy about that!")
 gold_dict = {"cats": {"POSITIVE": 1.0, "NEGATIVE": 0.0}}
 example = Example.from_dict(doc, gold_dict)
 
-# Training data for an Entity Linking component
+# Training data for an Entity Linking component (also requires entities & sentences)
 doc = nlp("Russ Cochran his reprints include EC Comics.")
-gold_dict = {"links": {(0, 12): {"Q7381115": 1.0, "Q2146908": 0.0}}}
+gold_dict = {"entities": [(0, 12, "PERSON")],
+             "links": {(0, 12): {"Q7381115": 1.0, "Q2146908": 0.0}},
+             "sent_starts": [1, -1, 0, 0, 0, 0, 0, 0]})}
```
adrianeboyd, line 458: I think these `sent_starts` look a bit confusing. I'd prefer `[True, False, False, ... ]` to parallel `Token.is_sent_start`, but if you want the shorter ints, then all `-1` instead of `0`?

There's also a typo at the end of the dict: `"sent_starts": [...]}`.

svlandeg: > There's also a typo

Thanks - fixed!

> I think these `sent_starts` look a bit confusing

Hehe, I'll admit to being a bit sneaky here. I put the `sent_starts` in like this on purpose. You can't actually put `[1, 0, 0, 0, 0, 0, 0, 0]` or `[True, False, False, False, False, False, False, False]` because you'll get a `ValueError: [E030] Sentence boundaries unset` (cf `test_ussue7062`) because this default pattern means that `doc.has_annotation("SENT_START")` is `False`, which I still think is pretty odd and unexpected. I ran into this when I designed that tests and similar ones, before, if you remember, and I still feel like we should fix this somehow...

adrianeboyd: No, I understand why, I just think that mixed `-1` and `0` is more confusing than just `-1` for a simple example.

adrianeboyd: (Which is also why I think `[True, False, False, False, ...]` is clearer.)

adrianeboyd: Or am I wrong about how `False` is converted here? If so, then all `-1` is better, I think?

svlandeg: Yea `False` isn't converted to `-1`. I'll put all `-1`, which I think is pretty ugly, but I've made my point ;-)

adrianeboyd: Yeah, I think that should be fixed so that `False` is converted to `-1` in `_annot2array`, possibly also `None` to `0`. It's weird-ish, but better than this, and potentially the `sent_starts` docs should be updated to not just be `List[bool]`.

PR #7886, opened 2021-04-26
Fix percent unk display in debug data

polm:
This was showing (ratio %), so 10% would show as 0.10%. Fix by
multiplying ratio by 100.

Might want to add a warning if this is over a threshold.

### Types of change

Minor fix to display output of debug data.

## Checklist


honnibal:
We should probably convert to integer if we're doing percent? Previously the precision was 2 decimal places, so now whole percents would be the equivalent. Agree it's nicer.

polm:
Makes sense, updated it to only show integer percentages. (Using `{:.0f}` will round numbers, so `0.9` will be rendered as `1`.)

PR #7907, opened 2021-04-26
Update load_lookups return type and docstring

### LINKED ISSUE #7894: `load_lookups` return type misleading
`load_lookups` returns a `Lookup` object instead of a `Dict[str, Any]` as suggested in the function definition: https://github.com/explosion/spaCy/blob/df3444421aba611d4ad1238610ce189df158d85a/spacy/lookups.py#L15-L17

This makes typecheckers fail when calling methods of the object, e.g. `lookups.get_table('lexeme_prob')`.

Moreover, the current docstring is misleading as `Lookups` doesn't even implement a `__getitem__` method.

jklaise:
## Description
Fixes #7894.

### Types of change
Updated docstring and return type hint of a function.

## Checklist

### DIFF
--- a/.github/contributors/jklaise.md
+++ b/.github/contributors/jklaise.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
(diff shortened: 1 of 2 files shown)


svlandeg:
Thanks!

PR #7931, opened 2021-04-28
compare literal with ==

svlandeg:
## Description
This started throwing a warning on the command line when running `spacy train` or such :| 

### Types of change
fix/chore/ux

## Checklist

### DIFF
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -1535,9 +1535,9 @@ def to_ternary_int(val) -> int:
     """
     if isinstance(val, float):
         val = int(val)
-    if val is True or val is 1:
+    if val is True or val == 1:
         return 1
-    elif val is None or val is 0:
+    elif val is None or val == 0:
         return 0
     else:
         return -1


adrianeboyd:
No, this won't work because `False == 0` and we want `False -> -1` in the `else`. (Yay.) What kind of warning did you get?

I also tried making a mapping dict which doesn't work for the same reason, since:

```python
a = {False: 3}
assert a[0] == 3
```

I wondered at first why the tests pass, but the tests reference this values from this function instead of hard-coding anything. I guess that is a mistake.

svlandeg:
Hm. Sorry, I didn't look at the semantics details, but either way this needs to be rewritten. This is the output I got after pulling in `master` locally, so every user would get this, every time (joy!)

```
$ python -m spacy train ner_config.cfg 
...\spacy\spacy\util.py:1538: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if val is True or val is 1:
...\spacy\spacy\util.py:1540: SyntaxWarning: "is" with a literal. Did you mean "=="?
  elif val is None or val is 0:
ℹ Using CPU
ℹ To switch to GPU 0, use the option: --gpu-id 0

=========================== Initializing pipeline ===========================
[2021-04-28 13:38:30,802] [INFO] Set up nlp object from config
[2021-04-28 13:38:30,818] [INFO] Pipeline: ['ner']
[2021-04-28 13:38:30,818] [INFO] Created vocabulary
[2021-04-28 13:38:30,818] [INFO] Finished initializing nlp object
...
```

adrianeboyd:
Okay, I guess it will just have to be really ugly.

svlandeg:
Closed in favour of https://github.com/explosion/spaCy/pull/7944

PR #7961, opened 2021-04-30
Adds support for logging training loop metrics, models, datasets to M…

narayanacharya6:
This PR is to add support for logging metrics, models and datasets to [MLflow](https://www.mlflow.org/) during spaCy training loop.

## Description
Users who wish to use the new `spacy.MLflowLogger.v1` need to have mlflow installed in their environment. The changes were tested against `mlflow==1.15.0`. Steps to replicate usage include:

1. Install `mlflow`:
`pip install mlflow`

2. Start `mlflow` tracking server locally (or skip if you already have an `mlflo` server ready and running).
`mlflow server --backend-store-uri sqlite:///mlflow.db --host localhost --port 5001 --default-artifact-root ./artifacts`

3. Setup required environment variable (change`localhost:5001` based on your setup).
`export MLFLOW_TRACKING_URI=http://localhost:5001`

4. Update `training.logger` config (below is a bare minimum config, there are other optional parameters with reasonable defaults and these are documented):
```
[training.logger]
@loggers = "spacy.MLflowLogger.v1"
experiment_name = "test_re_ner_pipe"
```

### Types of change
The changes include optional enhancement for users and also updates documentation informing how to use the enhancement. The new logger (similar to the existing WandbLogger) also outputs to the console.

## Checklist

### DIFF
--- a/spacy/training/loggers.py
+++ b/spacy/training/loggers.py
@@ -2,6 +2,7 @@ from typing import TYPE_CHECKING, Dict, Any, Tuple, Callable, List, Optional, IO
 from wasabi import Printer
 import tqdm
 import sys
+import os
 
 from ..util import registry
 from .. import util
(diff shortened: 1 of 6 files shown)


spacy/training/loggers.py
```diff
@@ -175,3 +176,68 @@ def finalize() -> None:
...
+
+                if log_losses and losses:
+                    log_metrics({f"loss_{k}": v for k, v in losses.items()}, step=step)
+
+                if log_other_scores and isinstance(other_scores, dict):
+                    log_dict(other_scores, artifact_file='scores_{}.json'.format(step))
+
+                if model_artifact_path:
+                    from spacy import load
+                    model_path = info["output_path"]
```
narayanacharya6, line 230: To load a saved model here just log it MLflow does not seem ideal. I'd like to have access to the `nlp` object here directly. Need input if saving the model to MLflow can be executed better if done with hooking a custom callback for `training.before_to_disk`. Any thoughts?

narayanacharya6:
Hi @svlandeg, sorry to @ you, but I was hoping someone could review this PR, and provide some feedback :)

svlandeg:
Hi! Thanks for the PR, and apologies for the delay in our response. There's no need to ping anyone specific though, this will certainly be reviewed at some point. There's a bit of a backlog with the PRs right now, and some are more "easy" to merge than others. I hope you understand ;-)

spacy/training/loggers.py
svlandeg: I'm not sure I understand the issue - `setup_logger` receives the `nlp` object, right?

spacy/training/loggers.py
```diff
@@ -175,3 +176,68 @@ def finalize() -> None:
...
+                if log_scores:
+                    log_metric("score", score, step=step)
+
+                if log_losses and losses:
+                    log_metrics({f"loss_{k}": v for k, v in losses.items()}, step=step)
+
+                if log_other_scores and isinstance(other_scores, dict):
+                    log_dict(other_scores, artifact_file='scores_{}.json'.format(step))
+
+                if model_artifact_path:
```
svlandeg, line 228:
```suggestion
                if model_artifact_path and info.get("output_path"):
```
We need this additional check, because `spacy train` can be run without `output_dir`.

svlandeg:
Hi @narayanacharya6! While reviewing this PR, I noticed that `MLFlow` already has [spaCy-specific logging](https://github.com/mlflow/mlflow/blob/2cf363c43618dddacd2cef19a26c704b37561775/mlflow/spacy.py) functionality that you're calling into.

As such, we think it would make sense for this custom logger to be implemented by MLFlow instead. Then underlying changes to this logging functionality can easily be kept in-sync in the `mlflow` lib itself. So I would consider submitting this PR there.

Basically with [entry points](https://spacy.io/usage/saving-loading#entry-points), the functionality would then become available to anyone who has `mlflow` installed - it would work pretty transparently and it's probably a better solution overall. A simple example of using such an entry point can be found in our own `ml-datasets` lib, which doesn't import `spacy` but still provides custom `readers`: https://github.com/explosion/ml-datasets/blob/master/setup.cfg#L22

spacy/training/loggers.py
narayanacharya6: Ah, missed the obvious. I was distracted by the `log_step` receiving only the `info` dictionary. My bad.

narayanacharya6:
Thanks for reviewing the PR @svlandeg. Now that I think of it your suggestions make sense. This should reside in MLflow or a separate utility library and not within spaCy.  

Thanks for pointing me to the entry points functionality. It will be helpful for this as well as some other projects we are working on. Skimming through the documentation, it looks like this should work well for installing pipeline components (with the `spacy_factories` entry point). Will it also work for this case where we need to register a custom logger from another package and do we use the same `spacy_factories` entry point for all such custom stuff?

spacy/training/loggers.py
svlandeg: No worries! The nested functions can sometimes be deceiving ;-)

svlandeg:
> Will it also work for this case where we need to register a custom logger from another package and do we use the same `spacy_factories` entry point for all such custom stuff?

Ah, no. You can get an overview of the available registries in `spacy` here: https://spacy.io/api/top-level#registry. For instance, the example I linked in `ml-datasets` uses `spacy_readers` - but for this you'll want to use `spacy_loggers`.

If you'll make the PR on the `mlflow`, is it OK if I close this? I hope you can use my review comments to finalize the functionality, it looks good otherwise, and it'll be great if MLFlow can include this out-of-the box for spaCy users :-)

narayanacharya6:
Understood, and thanks for the review comments! Yes, please feel free to
close this PR :)

PR #7984, opened 2021-05-03
Preserve existing entity span annotation in NER

adrianeboyd:
## Description

Preserve existing entity annotation in NER component by skipping any existing entities when assigning predictions from the model. This preserves `kb_id` on the existing entity spans without requiring changes to the transition system.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
The additional pass through the doc is displeasing but hopefully fast. Changing the transition system to track `kb_id` along with `label` didn't look like a better option, although changing it to silently skip existing annotation might be possible instead? I am not sure about the details...

adrianeboyd:
Actually, nevermind, I'm not sure there's a particularly good reason to set from spans like this in the first place...

PR #8005, opened 2021-05-05
Fix pretraining objectives fragment

polm:
The fragment here is reused from a heading higher up, so you couldn't
link to this section.

## Checklist


svlandeg (commented):
Good catch :-) 

I think where it reads `[objective](#pretraining-details)` on this page, this can be changed to `[objective](#pretraining-objectives)` as well.

PR #8012, opened 2021-05-05
Ensemble textcat with listener

svlandeg:
## Description
Stumbled upon a very specific issue thanks to https://github.com/explosion/spaCy/discussions/7651. If the `TextCatEnsemble` architecture is used, in combination with a `Tok2VecListener`, AND there's another component further down the pipeline also using a `Tok2VecListener`, then, and only then, we'll get a crash in Thinc's `concatenate` layer:
```
  File "/home/user/.virtualenvs/spacy-experiments/lib/python3.8/site-packages/thinc/layers/concatenate.py", line 68, in backprop
    dX += bwd(dY)
TypeError: 'NoneType' object is not iterable
```
This happens because the ensemble textcat architecture is the only one that `concatenate`'s (and backpropagates) results of a `Tok2VecListener`. All other architectures simply use `chain`. And all this works if there is just one listener. But when there are multiple, only the last one runs the `backprop` call, and all the others run `accumulate_gradient`, which doesn't return any gradient (https://github.com/explosion/spaCy/blob/master/spacy/pipeline/tok2vec.py).

This fixes the issue by returning an array of zeros instead of `None` in `accumulate_gradient`.

### Types of change
regression test

## Checklist
- [x] I ran the tests locally, and all new and existing tests passed.

### DIFF
--- a/spacy/pipeline/tok2vec.py
+++ b/spacy/pipeline/tok2vec.py
@@ -173,6 +173,7 @@ class Tok2Vec(TrainablePipe):
             for i in range(len(one_d_tokvecs)):
                 d_tokvecs[i] += one_d_tokvecs[i]
                 losses[self.name] += float((one_d_tokvecs[i] ** 2).sum())
+            return [self.model.ops.alloc2f(*t2v.shape) for t2v in tokvecs]
 
         def backprop(one_d_tokvecs):
             """Callback to actually do the backprop. Passed to last listener."""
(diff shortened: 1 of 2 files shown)


SharbelWired:
Please note that I am getting the same behavior in spacy 3.7.2 .  I was getting the error "TypeError: 'NoneType' object is not iterable" when trying to train a new pipeline that has these components => 

`pipeline = ["transformer","spancat","ner","textcat_multilabel"]`

I was getting the error above until I moved the "textcat_multilabel" component to the end of the list.

PR #8022, opened 2021-05-06
Update negspacy example code for spaCy 3.0

jenojp:
## Description
Updating code example for v1.0 of negspacy. 

### Types of change
Website update (universe.json)

## Checklist


svlandeg:
Yay! Great to see you've updated `negspacy` and thanks for the update here as well :-)

svlandeg:
The code is now live and looks good. The only thing I noticed just now is that the import `from negspacy.negation import Negex` is actually redundant...

jenojp:
Oh you're right! Old habits... I'll work to update this throughout

Issue #7627, opened 2021-03-31
Improved behaviour (and documentation) of DocBin.to_disk() with large corpora


arc12:
The failure of DocBin to_disk() is seemingly well-known, e.g. discussion: #7061 
This is a bit of a WTF moment for newbies, especially when considering using the docbin output for training, as it isn't quickly obvious that you can pass a directory to train. Additionally, we're left guessing/experimenting about what "too large" looks like, so the chunking will be rather hit-or-miss.

Suggestion:
1. to_disk() either provides an option to automatically split output, or makes this the default, and takes an informed view of what the chunk size should be.
2. documentation of to_disk() explains this, and that giving a directory for paths.train works as it does.

arc12:
thanks @svlandeg for your considered and educational reply.
I appreciate the call to make a PR, but I don't have the knowledge (yet?) to make a sensible contribution.

svlandeg:
No worries. If you ever do feel like contributing, https://github.com/explosion/spaCy/blob/master/CONTRIBUTING.md#contributing-to-the-code-base has some pointers ;-)

PR #8069, opened 2021-05-11
Custom warning if the doc_bin is too large

svlandeg:
## Description
Custom error when `doc_bin` is too large, hinting to the possibility of using a directory of `.spacy` files instead of just one file (when using the default `spacy.Corpus.v1` during training)

cf https://github.com/explosion/spaCy/discussions/7061, fixes https://github.com/explosion/spaCy/issues/7627

### Types of change
UX

## Checklist


spacy/errors.py
```diff
@@ -501,6 +501,11 @@ class Errors:
     E202 = ("Unsupported alignment mode '{mode}'. Supported modes: {modes}.")
 
     # New errors added in v3.x
+    E871 = ("Could not serialize the 'doc_bin' because it is too large. "
+            "Consider splitting up your documents into several doc bins "
+            "and serializing each separately. "
+            "These files can be read by spacy.Corpus.v1 if you provide a "
+            "directory instead of a file name as argument.")
```
adrianeboyd, line 508:
```suggestion
    E871 = ("Could not serialize the DocBin because it is too large. Consider "
            "splitting up your documents into several doc bins and serializing "
            "each separately. spacy.Corpus.v1 will search recursively for all "
            "*.spacy files if you provide a directory instead of a filename as "
            "the 'path'.")
```

adrianeboyd:
Hmm, the merge clobbered my suggestion somehow?

svlandeg:
Safe to say I managed to mess this merge up entirely. I think I forgot to pull your committed suggestion before fixing the merge conflict with `master`. I think it's OK and good to merge now?

PR #8090, opened 2021-05-13
util.filter_spans possibility to remove longest spans

JAugusto97:
## Description
The util.filter_spans function only provides the option of removing the smallest spans. This modification makes it possible for the user to select if he wants to remove the smallest or the longest spans. 

### Types of change

Enchancement
## Checklist

### DIFF
--- a/.github/contributors/JAugusto97.md
+++ b/.github/contributors/JAugusto97.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Hi, this is a nice idea, but I'm hesitant to support too many options in the core library because I'm worried it will become difficult to maintain and document, see a related discussion here:

https://github.com/explosion/spaCy/discussions/7355#discussioncomment-487435

In this case, you would also probably want to change the sort key, since for overlapping spans, this will return the last span rather than the first span.

JAugusto97:
> Hi, this is a nice idea, but I'm hesitant to support too many options in the core library because I'm worried it will become difficult to maintain and document, see a related discussion here:
> 
> [#7355 (reply in thread)](https://github.com/explosion/spaCy/discussions/7355#discussioncomment-487435)

Thanks for your answer!

I would suggest adding this enhancement to filter_spans despite the given counterpoint because I believe this function follows a similar concept of a sorting function in python. I think no method is "more important" than the other. Only being capable of removing the smallest overlapping spans would be similar to not having the "reverse" option in the sorted function. The option depends on the application, as mine would require this modification.

> In this case, you would also probably want to change the sort key, since, for overlapping spans, this will return the last span rather than the first span.

Sorry, I'm not sure I understood what you meant. The last span is the smallest overlapping span after sorting, which we want if longest=True. I've tested this function, made some test cases and It seems like it is behaving as it should. A simple example with 3 overlapping spans and one non-overlapping span like:
spans = [(5, 9), (14, 29), (20, 29), (25, 29)]

returns [(5, 9), (25, 19)], as it should.

Sorry if I'm missing something and thanks again for the answer!

adrianeboyd:
Sorry, I didn't explain that well. If you have overlapping spans of the same length, this prefers the later span rather than the earlier span. The second sort key `-span.start` + `reverse=True` works to prefer the earliest span for the longest spans of the same length, but if you want the shortest spans, then it's `start` + `reverse=False`.

adrianeboyd:
Hi, thanks again for the PR, but we discussed this internally and decided that we want to keep this function as it is without any additional options. It should hopefully not be too difficult for users to implement custom filtering methods as needed for their tasks.

PR #8105, opened 2021-05-15
[Finnish tokenizer] Support for common contractions

aajanki:
## Description
This improves the Finnish language tokenizer by adding the support for common contractions of a conjunction and the negative verb. I also added a new tests cases.

### Types of change
Enhancement

## Checklist


adrianeboyd:
Sorry for the late reply! Thanks for this contribution (and especially for including tests), it looks fine to be included in the upcoming v3.1.

PR #8106, opened 2021-05-15
Support negative examples in partial NER annotations

honnibal:
Many spaCy components have some support for training from partial annotations, or annotations with missing information. In v2, there was partial support for training from entity annotations that specified that some sequence of tokens was _not_ an entity of a given label. However, there were many limitations to this functionality:

* The negative examples had to be specified in BILUO format, which meant each token could only be part of one negative entity. You couldn't specify two incorrect entities for a single span, or incorrect entities that overlapped.
* The exclusion mechanism looked at whether some _action_ was incorrect, which isn't really what we want. Even if an entity is incorrect, maybe we should still use the `B-` action on its first token, because it might just be the end-point of the entity that's incorrect.

In spaCy v3, we dropped support for this feature, as there was no good way to write these non-entity annotations into the `Doc` object to train from them.

This patch restores and improves the functionality:

* A new configuration key is added to the NER factory. This is provisionally called `negative_samples_key`, but we should change that to something better. This key is used to fetch a span-group from the reference parse. All spans in this group are treated as incorrect, and the NER will learn not to predict (exactly) those spans. The spans are allowed to overlap, so you can specify multiple incorrect labels now.

* The objective now handles the `B-` action correctly, and generally doesn't mark `B-` actions as incorrect, as the model can't assume no entity starts at that word. Perhaps we should have a config setting for this? This objective does give the greedy model little chance of learning from the partial annotations well. On the other hand, the greedy model is probably a bad choice for this anyway. The beam model really makes more sense here.

# TODOs

* [x] Test beam training
* [x] More tests
* [x] Fix naming
* [ ] Example project

## Checklist


spacy/pipeline/_parser_internals/ner.pyx
```diff
@@ -373,23 +399,32 @@ cdef class Begin:
...
+            cost += 1
+        if s.buffer_length() < 3:
+            # Handle negatives. In general we can't really do much to block
+            # B, because we don't know whether the whole entity is going to
+            # be correct or not. However, we can at least tell whether we're
+            # going to be opening an entity where there's only one possible
+            # L.
+            for span in gold.negs[:gold.nr_neg]:
+                if span.label == label and span.start == b0:
+                    cost += 1
```
svlandeg, line 426: should this one have a `break` statement like in `Last` and `Unit` to stop iterating over `gold.negs` once we found one relevant negative example?

svlandeg:
> This is provisionally called `negative_samples_key`, but we should change that to something better.

`incorrect_spans_key`?

spacy/pipeline/ner.pyx
```diff
@@ -40,6 +40,7 @@ DEFAULT_NER_MODEL = Config().from_str(default_model_config)["model"]
         "moves": None,
         "update_with_oracle_cut_size": 100,
         "model": DEFAULT_NER_MODEL,
+        "negative_samples_key": None
```
svlandeg, line 43: I'm not sure it's a good idea to specify this parameter here at the component factory. Imagine sourcing an `ner` component from a base model - you won't be able to change this parameter in the config. It feels like it should be more of a reader thing?

spacy/pipeline/_parser_internals/ner.pyx
honnibal: Hm, yes I think so

spacy/pipeline/_parser_internals/ner.pyx
```diff
@@ -156,8 +172,6 @@ cdef class BiluoPushDown(TransitionSystem):
         cdef attr_t label
         if name == '-' or name == '' or name is None:
             return Transition(clas=0, move=MISSING, label=0, score=0)
-        elif name == '!O':
-            return Transition(clas=0, move=ISNT, label=0, score=0)
         elif '-' in name:
             move_str, label_str = name.split('-', 1)
             # Hacky way to denote 'not this entity'
```
svlandeg, line 177: The lines below this need to be removed, as `"x"` won't be in `MOVE_NAMES` anymore.

But perhaps we need to throw a warning here, as old code might still be assuming something like "B-!GPE" is meaningful?

honnibal:
We've now completed a simple check on our English OntoNotes 5 benchmark. We're still working on a project that will be more suitable for external sharing, but the PR can be merged already.

What we did was run the model over the development data and collect the correct and incorrect predictions. We stored the incorrect predictions on the “negative spans” key of `doc.spans` and saved the correct spans in `doc.ents`. All other tokens were marked as `None`. However, we added one variation: if the document was entirely correct (no incorrect entities, no missing entities), we marked its non-entity tokens `O`. Basically we have a binary annotation marking the document as entirely correct or not entirely correct, and then binary annotations on the individual entity predictions.

In my initial analysis, I thought that the negative examples would only work during beam search training. This is incorrect. The problem I identified, that the model would have no reason to prefer `O` over e.g. `U-WORK_OF_ART`, occurs equally for beam and greedy training.

The solution to this problem is to have at least some fully labelled examples in the mix. This can be done by confirming the fully correct examples, or perhaps also by including some documents that have a high prior on having no entities.

spacy/pipeline/_parser_internals/ner.pyx
```diff
@@ -79,9 +79,17 @@ cdef GoldNERStateC create_gold_state(
...
+    ner_ents, ner_tags = example.get_aligned_ents_and_ner()
     for i, ner_tag in enumerate(ner_tags):
         gs.ner[i] = moves.lookup_transition(ner_tag)
+
+    # Prevent conflicting spans in the data. For NER, spans are equal if they have the same offsets and label.
+    neg_span_hashes = [hash((neg_ent.start_char, neg_ent.end_char, neg_ent.label)) for neg_ent in negs]
+    for pos_span in ner_ents:
+        pos_hash = hash((pos_span.start_char, pos_span.end_char, pos_span.label))
+        if pos_hash in neg_span_hashes:
+            raise ValueError(Errors.E868.format(span=(pos_span.start_char, pos_span.end_char, pos_span.label_)))
```
honnibal, line 91: I think we want:

```python

neg_span_hash = {(e.start_char, e.end_char, e.label) for e in negs}
for pos_span in ner_ents:
    if (pos_span.start_char, pos_span.end_char, pos_span.label) in neg_span_hash:
        raise ValueError(Errors.E868.format(span=(pos_span.start_char, pos_span.end_char, pos_span.label_)))
```

It's important to use sets instead of lists for set membership, as the list membership check will be `O(n)` in the length of the list. Since we'll be making `m` checks (for the number of entities), and both the number of entities and the number of negative entities can be proportional to the document length, the list membership check can be quadratic (`O(n^2)`).

PR #8132, opened 2021-05-18
Remove unsupported attrs from attrs.IDS

adrianeboyd:
## Description

The attributes `PROB`, `CLUSTER` and `SENT_END` are not supported by `Lexeme.get_struct_attr` so should not be included through `attrs.IDS` as supported attributes in `Doc.to_array` and other methods.

### Types of change

Bug fix?

## Checklist

### DIFF
--- a/spacy/attrs.pyx
+++ b/spacy/attrs.pyx
@@ -74,7 +74,6 @@ IDS = {
     "SUFFIX": SUFFIX,
 
     "LENGTH": LENGTH,
-    "CLUSTER": CLUSTER,
     "LEMMA": LEMMA,
     "POS": POS,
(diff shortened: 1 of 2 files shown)


adrianeboyd:
So, I think that `PROB` never worked with `Doc.to_array` because `Lexeme.get_struct_attr` only gets `attr_t` values. My original plan was to add `PROB` and `CLUSTER` back to `LexemeC` and then support them in `Doc.to_array` and set both in the lookups table and in the vocab with the setter as for `NORM`.

It seems kind of unnecessary to add `CLUSTER` back by itself, so instead it seemed to make sense to clean up `attrs.IDS` this way instead.

PR #8142, opened 2021-05-19
Switch default train corpus max_length to 0 in quickstart

adrianeboyd:
## Description

The behavior of `spacy.Corpus.v1` is unexpected enough for `max_length != 0` that `0` is a better default for users creating a new config with the quickstart.

If not, documents are skipped, sometimes the entire corpus is skipped, and sometimes documents are (quite unexpectedly for your average user) split into sentences.

### Types of change

?

## Checklist

### DIFF
--- a/spacy/cli/templates/quickstart_training.jinja
+++ b/spacy/cli/templates/quickstart_training.jinja
@@ -372,7 +372,7 @@ factory = "{{ pipe }}"
 [corpora.train]
 @readers = "spacy.Corpus.v1"
 path = ${paths.train}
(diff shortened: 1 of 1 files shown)


svlandeg:
Agreed with the change. Shall we put this in 3.1 because it may influence existing workflows (starting from `init config` etc)?

adrianeboyd:
Hmm, I guess that's reasonable. We've made other changes in the default config without waiting for v3.1, but I guess they were closer to bug fixes.

svlandeg:
That commit list now looks pretty scary. Are we sure this is going to go fine? :p

PR #8158, opened 2021-05-20
ensure tolerance is properly passed on

svlandeg:
## Description
Looks like `spacy.batch_by_words.v1` wasn't correctly passing onwards the `tolerance` parameter.

Should we bump its version to v2?

### Types of change
bug fix

## Checklist


honnibal:
Hmm. I can see the argument of bumping the version, but it feels kind of bad? I'd rather not

svlandeg:
Fine by me to just merge without bumping ;-)

PR #8212, opened 2021-05-27
Add all symbols in Unicode Currency Symbols block

polm:
In #8102 it came up that the rupee symbol was treated different from
dollar / euro / yen symbols. This adds many symbols not already
included.

https://en.wikipedia.org/wiki/Currency_Symbols_(Unicode_block)

There are probably other symbols that could be added.

### Types of change

Minor enhancement (?)

## Checklist

### DIFF
--- a/spacy/tests/training/test_training.py
+++ b/spacy/tests/training/test_training.py
@@ -336,15 +336,15 @@ def test_gold_biluo_additional_whitespace(en_vocab, en_tokenizer):
 
 
 def test_gold_biluo_4791(en_vocab, en_tokenizer):
-    doc = en_tokenizer("I'll return the ₹54 amount")
-    gold_words = ["I", "'ll", "return", "the", "₹", "54", "amount"]
+    doc = en_tokenizer("I'll return the A54 amount")
+    gold_words = ["I", "'ll", "return", "the", "A", "54", "amount"]
     gold_spaces = [False, True, True, True, False, True, False]
     entities = [(16, 19, "MONEY")]
     example = Example.from_dict(
         doc, {"words": gold_words, "spaces": gold_spaces, "entities": entities}
     )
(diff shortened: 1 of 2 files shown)


polm:
This is failing the tests for https://github.com/explosion/spaCy/issues/4791. It looks like that issue also involved the treatment of the rupee symbol, so I guess leaving these characters out of currency is intentional.

spacy/tests/training/test_training.py
```diff
@@ -344,7 +344,7 @@ def test_gold_biluo_4791(en_vocab, en_tokenizer):
         doc, {"words": gold_words, "spaces": gold_spaces, "entities": entities}
     )
     ner_tags = example.get_aligned_ner()
-    assert ner_tags == ["O", "O", "O", "O", "U-MONEY", "O"]
+    assert ner_tags == ["O", "O", "O", "O", "B-MONEY", "L-MONEY", "O"]
```
adrianeboyd, line 347: This test is about alignment, not tokenization, so this makes it equivalent to the following `$` case rather than testing the alignment. I think a better change would be to use `A` instead of `₹` or something else that results in one token with the default tokenizer. Or the other option is to use a custom tokenizer rather than `en_tokenizer` that treats `₹` as before. Or I guess this was probably more of an issue with `GoldParse`, and I guess a third option is to discard the test entirely.

polm: Ah, thanks for explaining that. It looks like changing the rupee symbol to A works, so I'll do that, but since this is GoldParse related maybe the easiest thing is to just delete it?

adrianeboyd: I think we can leave it for now. Overall the test suite could use some spring cleaning because it's starting to be too slow, but I think this is fine for now.

PR #8246, opened 2021-05-31
Don't add duplicate patterns all the time in EntityRuler (fix #8216)

### LINKED ISSUE #8216: Patterns get duplicated when calling EntityRuler.add_patterns multiple times
When calling `EntityRuler.add_patterns` multiple times, existing rules in the underlying 'matcher' get duplicated on every call. I believe it is done here (`entityruler.py line 322`):
```python
for label, patterns in self.token_patterns.items():
     self.matcher.add(label, patterns)
```
#### This code fragment demonstrates the behavior:

```python
import spacy

entity_ruler = EntityRuler(nlp, validate = True)

patterns = [
    {'label' : 'pattern 1', 'pattern' : [{'LOWER' : 'a'}], 'id' : '1'},
    {'label' : 'pattern 2', 'pattern' : [{'LOWER' : 'b'}], 'id' : '2'},    
]

print("---------After adding 2 patterns ---------")
entity_ruler.add_patterns(patterns)
for name, pattern in entity_ruler.matcher._patterns.items() :
    print(nlp.vocab.strings[name], pattern)

entity_ruler.add_patterns([])
print("---------After adding empty array ---------")
for name, pattern in entity_ruler.matcher._patterns.items() :
    print(nlp.vocab.strings[name], pattern)

entity_ruler.add_patterns([ {'label' : 'pattern 3', 'pattern' : [{'LOWER' : 'c'}], 'id' : '3'},])
print("---------After adding a new pattern ---------")
for name, pattern in entity_ruler.matcher._patterns.items() :
    print(nlp.vocab.strings[name], pattern)

```

#### Output:

```
---------After adding 2 patterns ---------
pattern 1||1 [[{'LOWER': 'a'}]]
pattern 2||2 [[{'LOWER': 'b'}]]
---------After adding empty array ---------
pattern 1||1 [[{'LOWER': 'a'}], [{'LOWER': 'a'}]]
pattern 2||2 [[{'LOWER': 'b'}], [{'LOWER': 'b'}]]
---------After adding a new pattern ---------
pattern 1||1 [[{'LOWER': 'a'}], [{'LOWER': 'a'}], [{'LOWER': 'a'}]]
pattern 2||2 [[{'LOWER': 'b'}], [{'LOWER': 'b'}], [{'LOWER': 'b'}]]
pattern 3||3 [[{'LOWER': 'c'}]]

```

## Info about spaCy

- **spaCy version:** 3.0.6
- **Platform:** Darwin-20.3.0-x86_64-i386-64bit
- **Python version:** 3.7.9
- **Pipelines:** en_core_web_lg (3.0.0), en_core_web_md (3.0.0), en_core_web_sm (3.0.0)

polm:
## Description

Before this change, the EntityRuler `add` would add new patterns to internal lists, and then use those lists to add patterns to the internal Matcher and PhraseMatcher. This resulted in the same pattern being added multiple times. 

This adds new patterns one at a time without reference to the internal lists, so the duplicating behavior is avoided.

### Types of change

Bugfix

## Checklist

### DIFF
--- a/spacy/tests/pipeline/test_entity_ruler.py
+++ b/spacy/tests/pipeline/test_entity_ruler.py
@@ -91,17 +91,6 @@ def test_entity_ruler_clear(nlp, patterns):
     assert len(doc.ents) == 0
 
 
-def test_entity_ruler_fix8216(nlp, patterns):
-    """Test that patterns don't get added excessively."""
-    ruler = nlp.add_pipe("entity_ruler", config={"validate": True})
-    ruler.add_patterns(patterns)
-    pattern_count = sum(len(mm) for mm in ruler.matcher._patterns.values())
-    assert pattern_count > 0
-    ruler.add_patterns([])
-    after_count = sum(len(mm) for mm in ruler.matcher._patterns.values())
-    assert after_count == pattern_count
-
-
 def test_entity_ruler_existing(nlp, patterns):
     ruler = nlp.add_pipe("entity_ruler")
     ruler.add_patterns(patterns)
--- a/spacy/tests/pipeline/test_entity_ruler.py
+++ b/spacy/tests/pipeline/test_entity_ruler.py
@@ -77,6 +77,18 @@ def test_entity_ruler_init_clear(nlp, patterns):
     ruler.initialize(lambda: [])
     assert len(ruler.labels) == 0
 
+def test_entity_ruler_clear(nlp, patterns):
+    """Test that initialization clears patterns."""
+    ruler = nlp.add_pipe("entity_ruler")
+    ruler.add_patterns(patterns)
+    assert len(ruler.labels) == 4
+    doc = nlp("hello world")
+    assert len(doc.ents) == 1
+    ruler.clear()
+    assert len(ruler.labels) == 0
+    doc = nlp("hello world")
+    assert len(doc.ents) == 0
+
 
 def test_entity_ruler_existing(nlp, patterns):
     ruler = nlp.add_pipe("entity_ruler")
--- a/spacy/tests/pipeline/test_entity_ruler.py
+++ b/spacy/tests/pipeline/test_entity_ruler.py
@@ -77,6 +77,7 @@ def test_entity_ruler_init_clear(nlp, patterns):
     ruler.initialize(lambda: [])
     assert len(ruler.labels) == 0
 
+
 def test_entity_ruler_clear(nlp, patterns):
     """Test that initialization clears patterns."""
     ruler = nlp.add_pipe("entity_ruler")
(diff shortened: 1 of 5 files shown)


polm:
Note: poking around I think I found more issues in EntityRuler, so this PR shouldn't be merged yet. I think the changes so far are fine but there should probably be more on top.

svlandeg:
Looks good! Feel free to mark this PR as draft while you're still working on it :-)

polm:
Going to wait for the tests to go green but I think this is good now. 

The extra change is that now `.clear()` resets all state. Before it cleared the internal list of patterns, but didn't reset the Matcher/PhraseMatcher, so results could be weird if `.clear()` was called externally. 

As part of making this I made a  small change to the PhraseMatcher - if the attr is passed in as `None` that will be converted to `ORTH`. Normally I would like to avoid that, but the PhraseMatcher is already converting "TEXT" to "ORTH" so it didn't change that much, and it allowed eliminating checks to avoid passing a `None` in several places.

One thing I had a look at is whether the EntityRuler internal lists of patterns could be eliminated entirely in favor of using the Matcher/PhraseMatcher data directly. It looks like the main thing in the lists but not in the Matchers is the Entity IDs, as distinct from labels. Since the lists aren't causing problems now I didn't spend more time looking at it but there might be a way to simplify the way the data is stored.

polm:
Hm, well, that failed a test. It looks like there's a test that assumes that calling `nlp.initialize` doesn't reset an EntityRuler - `EntityRuler.initialize` calls `clear()` internally, and so it doesn't reset patterns right now. Not sure if that's intended behavior or not. 

I think it makes sense for `initialize` to reset things, so I'll fix the test to match that assumption for now.

svlandeg:
> I think it makes sense for `initialize` to reset things, so I'll fix the test to match that assumption for now.

Agreed that that makes sense!

polm:
In that case this should be good to go.

spacy/tests/pipeline/test_entity_ruler.py
```diff
@@ -90,6 +91,17 @@ def test_entity_ruler_clear(nlp, patterns):
     assert len(doc.ents) == 0
 
 
+def test_entity_ruler_fix8216(nlp, patterns):
```
svlandeg, line 94: Test looks good to me, but as it's a regression test tied to a specific issue, I'd put it in the `tests/regression` folder :-)

polm: Ah, thanks for the tip - I didn't realize that existed.

polm: And done.

PR #8256, opened 2021-06-01
Add scorer option to components

adrianeboyd:
## Description

Add an optional `scorer` parameter to all pipeline components. If a scoring function is provided, it overrides the default scoring method for that component.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/pipeline/attributeruler.py
+++ b/spacy/pipeline/attributeruler.py
@@ -23,9 +23,11 @@ TagMapType = Dict[str, Dict[Union[int, str], Union[int, str]]]
 MorphRulesType = Dict[str, Dict[str, Dict[Union[int, str], Union[int, str]]]]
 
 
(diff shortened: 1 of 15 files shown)


adrianeboyd:
We would want to decide whether we want a new registry for this.

In some ways it might be easier for users packaging pipelines if this is a training feature rather than a pipeline feature, but it makes sense for users to be able to run the same evaluation in the future with the pipeline.

svlandeg:
We could also put the current (default) scoring mechanisms in a registry function and fill those in by default?

adrianeboyd:
Putting the scoring functions into a registry could be a good option, but would be a pretty major refactor. Converting this to a draft for now...

adrianeboyd:
Most likely superseded by #8766, which includes these changes before refactoring further.

PR #8335, opened 2021-06-10
Raise error if deps not provided with heads

polm:
Before this change, if heads were passed without deps they would be
silently ignored, which could be confusing. See #8334.

### Types of change

Minor bugfix

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -848,8 +848,6 @@ class Errors:
              "placeholder deps such as `deps=[\"dep\"]*len(heads)`.")
 
 
-
-
 # Deprecated model shortcuts, only used in errors and warnings
 OLD_MODEL_SHORTCUTS = {
     "en": "en_core_web_sm", "de": "de_core_news_sm", "es": "es_core_news_sm",
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -843,6 +843,10 @@ class Errors:
              "DependencyMatcher token patterns. The token pattern in "
              "RIGHT_ATTR should return matches that are each exactly one token "
              "long. Invalid pattern:\n{node}")
+    E1017 = ("Heads passed without deps. Normally this is a mistake, but if "
+             "intentional then pass filler deps such as `[\"dep\"] * len(heads)`.")
+
+
 
 
 # Deprecated model shortcuts, only used in errors and warnings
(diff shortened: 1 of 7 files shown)


adrianeboyd:
It would be better to use a non-empty string here, and we normally use `"dep"` for a placeholder non-empty dep.

The problem with the head attribute is that there's no value for a missing head, so in `Doc.has_annotation` and other places, whether the `dep` is set is used to decide whether both heads and deps are set. So even if this works for the `Doc` creation (and I'm not sure it does in the end?), in other places it still won't look like there's a dependency parse.

spacy/tokens/doc.pyx
```diff
@@ -275,6 +275,8 @@ cdef class Doc:
             deps = [dep if dep is not None else MISSING_DEP_ for dep in deps]
         if deps and not heads:
             heads = [0] * len(deps)
+        if heads and not deps:
+            deps = ['dep'] * len(heads)
```
svlandeg, line 279: Nitpicking: while `black` doesn't support `pyx` so isn't helpful here, we typically use double quotes for strings throughout the code base. Just for consistency :-)

adrianeboyd:
I think an error as Matt suggested would be fine as an alternative to this. You could give a suggestion about how to fill in default deps as `deps=["dep"]*len(heads)` or something like that in the error message.

spacy/tokens/doc.pyx
polm: Ah, I knew that was the style, I'll have to keep an eye out. My instinct is always to use double quotes for human-facing strings and single quotes for internal-use strings, though I don't know where I got that from.

polm:
Huh, turns out there's more than few tests that pass heads without deps.

spacy/errors.py
```diff
@@ -843,6 +843,10 @@ class Errors:
              "DependencyMatcher token patterns. The token pattern in "
              "RIGHT_ATTR should return matches that are each exactly one token "
              "long. Invalid pattern:\n{node}")
+    E1017 = ("Heads passed without deps. Normally this is a mistake, but if "
+             "intentional then pass filler deps such as `[\"dep\"] * len(heads)`.")
```
adrianeboyd, line 847: I think it would be better to give a little more explanation and not assume that an unlabeled dependency parse is a mistake, something like:

A Doc object requires both 'deps' and 'heads' for dependency parses. If no dependency labels are available, provide placeholder deps such as `deps=[\"dep\"]*len(heads)`.

polm: That sounds good, updated the message.

adrianeboyd:
Looks fine, I will just do some minor whitespace cleanup.

PR #8371, opened 2021-06-12
Preserve whitespace in span lemma (fix #8368)

### LINKED ISSUE #8368: Span.lemma joins tokens with spaces unconditionally, which is wrong for languages like Japanese
https://github.com/explosion/spaCy/blob/f4008bdb13e262c389e3d0c7017a634605f6e706/spacy/tokens/span.pyx#L757

```
import spacy
nlp = spacy.load("ja_core_news_sm")
doc = nlp("本須麗乃は本が好きだ。")
name = doc[:3]
print(name.text)
print(name.lemma_)
print(name.text != name.lemma_)
```

The lemma output gives us "本 須 麗乃" which introduces spaces between tokens.

polm:
## Description

Before this the `lemma` of a Span joined terms using a single space unconditionally, this uses the token whitespace attributes. 

This behavior is probably desirable for noun chunks. Consider "hot-dogs" as three tokens - it makes sense that the lemma would preserve the spacing rather than rendering it as "hot - dog". 

### Types of change

Minor behavior change / bug? fix. 

## Checklist

### DIFF
--- a/spacy/tokens/span.pyx
+++ b/spacy/tokens/span.pyx
@@ -736,7 +736,7 @@ cdef class Span:
     @property
     def lemma_(self):
         """RETURNS (str): The span's lemma."""
-        return " ".join([t.lemma_ for t in self]).strip()
+        return "".join([t.lemma_ + t.whitespace_ for t in self]).strip()
 
     property label_:
         """RETURNS (str): The span's label."""


adrianeboyd:
Ugh, the branches diverge too much to just switch the base branch. Could you redo this against `master` instead? I think it would be useful to have this bugfix in v3.0.x, too.

svlandeg:
Perhaps also add a regression test for #8368?

polm:
Closed in favor of #8391

PR #8388, opened 2021-06-14
Don't use the same vocab for source models

adrianeboyd:
## Description

The source models should not be loaded with the vocab from the current pipeline because this loads the vectors from the source model into the current vocab when they should be loaded separately in the pipeline config. If multiple components are sourced from separate pipelines, the current pipeline ends up with the vectors from the last sourced component and no warnings are shown about potentially incompatible vectors because at each point when a component was sourced, the current vectors looked compatible.

The strings are all copied in `Language.create_pipe_from_source`, so if the vectors are configured correctly in the current pipeline, the sourced component will work as expected. If there is a vector mismatch, a warning is shown. (It's not possible to inspect whether the vectors are actually used by the component, so a warning is the best option.)

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1679,8 +1679,11 @@ class Language:
                 else:
                     model = pipe_cfg["source"]
                     if model not in source_nlps:
-                        # We only need the components here and we need to init
-                        # model with the same vocab as the current nlp object
+                        # We only need the components here and we intentionally
+                        # do not load the model with the same vocab because
+                        # this would cause the vectors to be copied into the
+                        # current nlp object (all the strings will be added in
+                        # create_pipe_from_source)
                         source_nlps[model] = util.load_model(model)
                     source_name = pipe_cfg.get("component", pipe_name)
                     listeners_replaced = False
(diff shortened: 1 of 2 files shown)


spacy/language.py
```diff
@@ -1681,7 +1681,7 @@ def from_config(
                     if model not in source_nlps:
                         # We only need the components here and we need to init
                         # model with the same vocab as the current nlp object
-                        source_nlps[model] = util.load_model(model, vocab=nlp.vocab)
```
svlandeg, line 1684: I guess adjust the comment, too ;-)

svlandeg (commented):
Sounds reasonable to me!

spacy/language.py
adrianeboyd: Will do!

adrianeboyd:
It seems to work (you should just need the strings, you should see warnings if the vectors are different), but still I feel slightly nervous that this might break things in some way I'm not seeing. Some related discussion is here: https://github.com/explosion/spaCy/discussions/8278

The behavior around `vocab=nlp.vocab` also seems a bit unexpected in terms of how the vectors are simply clobbered, which breaks the first pipeline, so I'm not sure if there are additional things we should address or test there.

```python
import spacy
nlp = spacy.load("en_core_web_md")
nlp2 = spacy.load("en_core_web_lg", vocab=nlp.vocab)
assert nlp.vocab.vectors.shape == (684830, 300)
```

PR #8391, opened 2021-06-15
Change span lemmas to use original whitespace (fix #8368)

### LINKED ISSUE #8368: Span.lemma joins tokens with spaces unconditionally, which is wrong for languages like Japanese

polm:
This is a redo of #8371 based off master.

The test for this required some changes to existing tests. I don't think
the changes were significant but I'd like someone to check them.

### Types of change

Minor bugfix.

## Checklist

### DIFF
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -296,7 +296,6 @@ def test_span_attrs_writable(doc):
 
 
 def test_span_ents_property(doc):
-    """Test span.ents for the"""
     doc.ents = [
         (doc.vocab.strings["PRODUCT"], 0, 1),
         (doc.vocab.strings["PRODUCT"], 7, 8),
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -20,7 +20,17 @@ def doc(en_tokenizer):
             "O", "O", "O", "O", "O"]
     # fmt: on
     tokens = en_tokenizer(text)
-    return Doc(tokens.vocab, words=[t.text for t in tokens], heads=heads, deps=deps, ents=ents)
+    lemmas = [t.text for t in tokens]  # this is not correct, just a placeholder
+    spaces = [bool(t.whitespace_) for t in tokens]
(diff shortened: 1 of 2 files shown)


spacy/tests/doc/test_span.py
```diff
@@ -20,7 +20,17 @@ def doc(en_tokenizer):
             "O", "O", "O", "O", "O"]
     # fmt: on
     tokens = en_tokenizer(text)
-    return Doc(tokens.vocab, words=[t.text for t in tokens], heads=heads, deps=deps, ents=ents)
+    lemmas = [t.text for t in tokens]  # this is not correct, just a placeholder
```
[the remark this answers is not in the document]
adrianeboyd: To check the whitespace in `span.lemma_` in the tests later, otherwise it's all set to `""`.

spacy/tests/doc/test_span.py
```diff
@@ -286,7 +296,7 @@ def test_span_attrs_writable(doc):
 
 
 def test_span_ents_property(doc):
-    """Test span.ents for the """
+    """Test span.ents for the"""
```
adrianeboyd, line 299: What is this comment even supposed to say? "Test span.ents for the doc"?

polm: That's actually a good question. It looks like the comment has been like that since 2018, I just modified it due to running black on the code. 

I guess it can just be changed to `Test span.ents` or removed?

adrianeboyd: I know you didn't write it, it just jumped out at me when I was reading through.

PR #8397, opened 2021-06-15
Merge develop into master for v3.1

adrianeboyd:
## Description

Merge develop into master for v3.1

### Types of change

?

## Checklist


adrianeboyd:
The only failure is the new model download test, which is restricted to python 3.8. I would go ahead and merge this and then submit the compatibility check PR? For the test to pass, we'll need a published `en_core_web_sm`, maybe we can go ahead and publish v3.1.0.a2 or similar for testing purposes?

adrianeboyd:
We could also just test the compatibility table here instead of actually downloading a model? That is fairly easy with the validate methods.

PR #8405, opened 2021-06-16
adjust whitespace tokenizer to avoid sep in split()

svlandeg:
## Description
The docs show a custom whitespace tokenizer. Adjusting that example to not use a `sep` in the `text.split` call, as that could easily [lead to empty strings](https://docs.python.org/3.3/library/stdtypes.html?highlight=split#str.split), which are prohibited as tokens by spaCy. Ofcourse this is a silly example, but it could quickly become confusing to users when `E031` gets thrown.

### Types of change
small docs fix

## Checklist

### DIFF
--- a/spacy/tests/test_language.py
+++ b/spacy/tests/test_language.py
@@ -427,13 +427,27 @@ def test_language_whitespace_tokenizer():
             self.vocab = vocab
 
         def __call__(self, text):
-            words = text.split()
-            return Doc(self.vocab, words=words)
+            words = text.split(" ")
+            spaces = [True] * len(words)
+            # Avoid zero-length tokens
+            for i, word in enumerate(words):
+                if word == "":
+                    words[i] = " "
+                    spaces[i] = False
+            # Remove the final trailing space
+            if words[-1] == " ":
+                words = words[0:-1]
+                spaces = spaces[0:-1]
+            else:
+                spaces[-1] = False
+
+            return Doc(self.vocab, words=words, spaces=spaces)
 
     nlp = spacy.blank("en")
     nlp.tokenizer = WhitespaceTokenizer(nlp.vocab)
-    doc = nlp("What's happened to    me? he thought. It wasn't a dream. ")
-    assert doc
+    text = "   What's happened to    me? he thought. It wasn't a dream.    "
+    doc = nlp(text)
+    assert doc.text == text
 
 
 def test_language_custom_tokenizer():
--- a/spacy/tests/test_language.py
+++ b/spacy/tests/test_language.py
@@ -419,6 +419,23 @@ def test_language_from_config_before_after_init_invalid():
             English.from_config(config)
 
 
+def test_language_whitespace_tokenizer():
+    """Test the custom whitespace tokenizer from the docs."""
+
+    class WhitespaceTokenizer:
+        def __init__(self, vocab):
+            self.vocab = vocab
+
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Hmm, this doesn't really work because it doesn't preserve the original text. You'd have to add additional handling for extra whitespace. Maybe it would be better to skip empty tokens than to take this approach?

svlandeg:
I mean, it's never going to really work with such a simpel approach ;-)

adrianeboyd:
No, it can totally work. We should not have a demo/example that does not preserve the original text.

adrianeboyd:
```python
        words, spaces = spacy.util.get_words_and_spaces(text.split(), text)
        return Doc(self.vocab, words=words, spaces=spaces)
```

svlandeg:
But this would be sacrificing the simplicity of the original example, where the point is you can implement whatever you want, rather than rely on built-in functionality in spaCy... So I don't think that's a satisfying solution either.

adrianeboyd:
I don't mind deferring this decision to others, but for the record, I think it's extremely terrible to have this kind of example not preserve the original text. And you're obviously using spacy already, so it's not like this is pulling in an additional library just for tokenization.

svlandeg:
But spaCy allows this functionality, right? There's a difference between the tokenization we provide (always preserving the original text) and what is technically possible. There's even a unit test `test_language_custom_tokenizer` that specifically adds an underscore to each token.

adrianeboyd:
It's allowed, but I don't think we should present something as a `WhitespaceTokenizer` that doesn't really work, at least not without a very clear warning. (The previous behavior where you saw errors related to empty tokens made its limitations clearer, I think.)

svlandeg:
I've extended the example without using built-in functionality that a user would have to look up in the source to know what's happening. While the example now isn't as simple anymore, at least it's still pretty comprehensible (I hope), and at least it actually works on real texts, ànd preserves the original text.

adrianeboyd:
What if we went back to your original proposal, which does seem like a useful option, but just renamed it so it was clear that it changed the text?

```python
class NormalizingWhitespaceTokenizer:
    def __init__(self, vocab):
        self.vocab = vocab

    def __call__(self, text):
        words = text.strip().split()
        spaces = [True] * len(words)
        if len(spaces) > 0:
            spaces[-1] = False
        return Doc(self.vocab, words=words, spaces=spaces)
```

Or `PreprocessingWhitespaceTokenizer` or something. None of the names are great, but at least it's clear that it's doing more than just whitespace tokenization?

ines:
Copying over my comment from Slack:

> Hmm so I've played with a few other options (including iterating over the characters and choosing when to "make a token") but it all ends up equally as complex. I guess the answer here is "non-destructive tokenization is more complex than you think" 😅
>
> I agree that it's important to have an example that preserves the original text so maybe we should go with the slightly more complex example as it currently is because it at least shows some of the considerations you have to make if you do choose to implement a custom solution.

To add to this: I do think it's good to have an example of the "correct" way to do non-destructive basic whitespace tokenization and that shows why just `string.split(" ")` doesn't work (which may not be immediately intuitive if you haven't though about it much). So I'll go ahead and merge this.

We could still consider having another example that shows the "destructive" way – but I also wouldn't want to focus too much on it because it does break spaCy's general paradigm of `text == doc.text`.

PR #8424, opened 2021-06-17
Improve UX for W036 for entity ruler, etc.

adrianeboyd:
## Description

Improve UX for W036 around the entity ruler:

* Since an entity ruler may contain only matcher or only phrase matcher patterns, ignore `W036` when calling the internal matchers.
* Check for this warning if the entity ruler has no patterns in tests.
* Filter this warning in other tests where not relevant.

Additional:

* Don't run the entity ruler itself as part of the pipeline when adding phrase patterns.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/pipeline/entityruler.py
+++ b/spacy/pipeline/entityruler.py
@@ -3,6 +3,7 @@ from typing import Optional, Union, List, Dict, Tuple, Iterable, Any, Callable, 
 from collections import defaultdict
 from pathlib import Path
 import srsly
+import warnings
 
 from .pipe import Pipe
 from ..training import Example
@@ -141,7 +142,9 @@ class EntityRuler(Pipe):
 
     def match(self, doc: Doc):
         self._require_patterns()
-        matches = list(self.matcher(doc)) + list(self.phrase_matcher(doc))
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="\\[W036")
+            matches = list(self.matcher(doc)) + list(self.phrase_matcher(doc))
         matches = set(
             [(m_id, start, end) for m_id, start, end in matches if start != end]
         )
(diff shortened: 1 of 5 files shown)


spacy/pipeline/entityruler.py
```diff
@@ -276,7 +279,7 @@ def add_patterns(self, patterns: List[PatternType]) -> None:
                     current_index = i
                     break
             subsequent_pipes = [
-                pipe for pipe in self.nlp.pipe_names[current_index + 1 :]
+                pipe for pipe in self.nlp.pipe_names[current_index :]
```
adrianeboyd, line 282: @svlandeg: Is there a good reason to run the entity ruler itself as part of the pipeline when adding phrase patterns? It seems a little circular to me, but maybe there's a reason I'm missing?

svlandeg: Huh, no idea. This seems to be in the code base already for some time. I think your edit should be fine?

adrianeboyd: It seems okay in all my tests, so I'll leave it for now.

spacy/pipeline/entityruler.py
```diff
@@ -141,7 +142,9 @@ def __call__(self, doc: Doc) -> Doc:
 
     def match(self, doc: Doc):
         self._require_patterns()
-        matches = list(self.matcher(doc)) + list(self.phrase_matcher(doc))
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="\\[W036")
```
svlandeg, line 146: You could also call `errors.filter_warning` which does the escaping for you, but it's still only a one-liner so either way's fine

svlandeg (approved):

PR #8439, opened 2021-06-19
Add Spans to Evaluate CLI

NirantK:
## Description
Added Spans with default key as "sc" in cli/`evaluate`

### Types of change
Minor CLI improvement

## Checklist

### DIFF
--- a/spacy/cli/evaluate.py
+++ b/spacy/cli/evaluate.py
@@ -60,6 +60,7 @@ def evaluate(
     displacy_path: Optional[Path] = None,
     displacy_limit: int = 25,
     silent: bool = True,
+    span_key="sc",
 ) -> Scorer:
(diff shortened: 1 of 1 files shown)


spacy/cli/evaluate.py
```diff
@@ -60,6 +60,7 @@ def evaluate(
     displacy_path: Optional[Path] = None,
     displacy_limit: int = 25,
     silent: bool = True,
+    span_key="sc",
```
adrianeboyd, line 63: Let's call this `spans_key` to keep it consistent with the `spancat` config? (Hopefully a future version will parse all the `spans_` scores in some more general purpose way, but we don't want the v3.1 release to wait on this right now.)

NirantK: Ohh, yeah. The consistency helps. Thanks for mentioning this. Have made the change now

PR #8441, opened 2021-06-20
Add the right return type for Language.pipe and an overload for the as_tuples case

ldorigo:
## Description
Small fix to provide the correct return type for the Language.pipe function and to add an overload to properly handle passing tuples (when as_type is set to True).

The overload would be better using `Literal`s to enforce that `as_tuples` be true when passing an iterator of tuples as input, but  spaCy requires compatibility with python3.6, which doesn't support `Literal`, so I only overload based on the type of Iterable.

Sorry for the additional changed lines in the diff - I am using black as a formatter as specified in Contributing.md, so I'm not sure why it changed all those lines.

### Types of change

Typing improvement

## Checklist

- [x] I ran the tests, and all new and existing tests passed. (well, actually quite a few fail and have warnings both before and after - but my changes didn't add anything)


adrianeboyd:
Thanks for this contribution! The code duplication here makes me worried about longer-term maintenance, but this does look like the best option for providing the types right now.

We should update our internal requirements to a newer version of `flake8` that can handle `@overload` and I'll deal with a few of the formatting issues, but otherwise this looks good! (We don't have any hooks that automatically run black, so sometimes it's been a while since we did a larger cleanup of the code base. Nothing you did wrong!)

ldorigo:
Thanks for reviewing this. I have a few other typing-related PRs coming up.  Regarding overloads and code duplication, as far as I'm aware, that's the only way to do it in Python :-) 

If you are interested in improving spacy's typing, you might consider adding https://pypi.org/project/typing-extensions/ to your dependencies. It allows using more recent additions to the Typing module while retaining compatibility with Python3.6,  and would for instance allow to use `Literal` to make better overloads (in this case, we could express that the input can only be a tuple IF as_tuples=True)

adrianeboyd:
While we're definitely looking forward to using newer features like `Literal`, I have mixed feelings about including even more backport libraries, especially ones that also include experimental features. We've had trouble with similar libraries recently and we might prefer to wait until we drop python 3.6. On the other hand, we're not planning to drop 3.6 anytime soon, so it's something we can consider.

What would be nice is if we could go back in time and change this API, but ah, well...

adrianeboyd:
Actually, we do already have `typing-extensions` as a dependency in `thinc`, so it shouldn't be an issue to add it here, too.

adrianeboyd:
We can use this version for now for the upcoming v3.1.0 and decide about `typing-extensions` in the future.

ldorigo:
Sounds good. When I have a few minutes I can submit a new PR with the typing-extensions version and we can continue the discussion there.

ldorigo:
Sounds good. When I have a few minutes I can submit a new PR with the typing-extensions version and we can continue the discussion there.

PR #8442, opened 2021-06-20
Raise custom error in EntityLinker when KB is not set

### LINKED ISSUE #8280: TypeError in entity linker if kb is not set
I tried something similar to the example from the [docs](https://spacy.io/api/entitylinker#set_kb):
```python
def create_kb(vocab):
    kb = KnowledgeBase(vocab, entity_vector_length=128)
    kb.add_entity(...)
    kb.add_alias(...)
entity_linker = nlp.add_pipe("entity_linker")
entity_linker.set_kb(create_kb)
```
I actually forgot to add the line `return kb` in `create_kb()`. Running the code leads to the following, not very nice looking error:
```
File "[...]/venv/lib/python3.8/site-packages/spacy/pipeline/entity_linker.py", line 159, in validate_kb
    if len(self.kb) == 0:
TypeError: object of type 'NoneType' has no len()
```

I think there should be a simple check for `None` added to this line:
https://github.com/explosion/spaCy/blob/ff5cf3606cf354c7639ce40b81cacc93bfb03583/spacy/pipeline/entity_linker.py#L159

I would open a pull request, but I don't know whether you want a new error message or whether I should change the existing one.

xadrianzetx:
## Description
* Adds error code `E1018`  that mentions knowledge base is not set.
* This error can be raised when user calls `validate_kb()`
* resolves #8280 

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -856,6 +856,8 @@ class Errors:
     E1017 = ("A Doc object requires both 'deps' and 'heads' for dependency "
              "parses. If no dependency labels are available, provide "
              "placeholder deps such as `deps=[\"dep\"]*len(heads)`.")
+    E1018 = ("Knowledge base for component '{name}' is not set. "
+             "Use the methods `kb.add_entity` and `kb.add_alias` to add entries.")
 
 
 # Deprecated model shortcuts, only used in errors and warnings
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -857,7 +857,8 @@ class Errors:
              "parses. If no dependency labels are available, provide "
              "placeholder deps such as `deps=[\"dep\"]*len(heads)`.")
     E1018 = ("Knowledge base for component '{name}' is not set. "
-             "Use the methods `kb.add_entity` and `kb.add_alias` to add entries.")
+             "Make sure either `nel.initialize` or `nel.set_kb` "
+             "is called with a `kb_loader` function.")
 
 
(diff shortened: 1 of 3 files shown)


spacy/errors.py
```diff
@@ -856,6 +856,8 @@ class Errors:
     E1017 = ("A Doc object requires both 'deps' and 'heads' for dependency "
              "parses. If no dependency labels are available, provide "
              "placeholder deps such as `deps=[\"dep\"]*len(heads)`.")
+    E1018 = ("Knowledge base for component '{name}' is not set. "
+             "Use the methods `kb.add_entity` and `kb.add_alias` to add entries.")
```
svlandeg, line 860:
```suggestion
    E1018 = ("Knowledge base for component '{name}' is not set. "
             "Make sure either `nel.initialize` or `nel.set_kb` is called with a `kb_loader` function.")
```

svlandeg (changes requested):
This looks great, thanks for making this PR!

I only have a small suggestion to change the error message. If the `kb` is `None`, it won't help to call `add_entity` or `add_alias` functions - so I'd guide users to the `set_kb` and `initialize` methods instead.

spacy/errors.py
xadrianzetx: Yeah, this error message is much more helpful. Thanks!

svlandeg (approved):
Looks good to me!

svlandeg:
Thanks again for the PR :-)

PR #8505, opened 2021-06-26
Fix double slash in model release web page

bryant1410:
Fixes that the links are like:

https://github.com/explosion/spacy-models/releases//tag/en_core_web_sm-3.0.0

The link still works, but who knows if in the future GitHub would prevent this from working.


ines:
Thanks, good catch 👍

Issue #7457, opened 2021-03-16
The 'noun_chunks' syntax iterator is not implemented for language 'nl'. (and potentially other languages)


davidberenstein1957:
Thanks, the info about multiprocessing vs. not was a helpful clue. This is indeed a bug and the problem is that `noun_chunks` isn't being preserved when the doc is pickled. We will get this fixed in the next patch release of v3.0.

_Originally posted by @adrianeboyd in https://github.com/explosion/spaCy/issues/6891#issuecomment-777590412_

davidberenstein1957:
This issue seems to have only been fixed for the 'en' language models.

adrianeboyd:
Hi, there are two separate issues here:

* Instead of silently returning zero noun chunks, as of v3.0 you get a `NotImplementedError` if noun chunks aren't implemented for a particular language. There's no implementation of `noun_chunks` for Dutch at this point.

   We'd be happy to accept a PR that adds this. You can see how the noun chunks iterator is implemented for different languages by looking at `spacy/lang/*/syntax_iterators.py`. Since the Dutch models use UD trees, you could have a look at languages other than English or German for examples of how noun chunks are can be extracted from UD POS + dependency labels.

* The bug related to #6891 is fixed as of v3.0.2 (for all languages) and isn't related to this issue with Dutch.

adrianeboyd:
Let me transfer this to the new discussions board. This issue will be locked, but the new discussion thread will be open.

PR #8529, opened 2021-06-28
Adding noun_chunks to the DUTCH language model (nl)

julien-talkair:
## Description

Following work done on FR and SV (#4422) for UD tags, this adds the noun_chunks iterator to the NL language.

### Types of change

Enhancement: NL language now has noun_chunks
There is a link with #7457 (closed) where a PR like this one is mentioned.

## Checklist


julien-talkair:
(additional comment)
- `black` suggested changes in 13 files out of the scope of this PR. I did not include these changes in here, but might be an idea for someone to pick it up

spacy/lang/nl/syntax_iterators.py
```diff
@@ -0,0 +1,39 @@
...
+            head = word.head
+            while head.dep == conj and head.head.i < head.i:
+                head = head.head
+            # If the head is an NP, and we're coordinated to it, we're an NP
+            if head.dep in np_deps:
+                prev_end = word.right_edge.i
+                yield word.left_edge.i, word.right_edge.i + 1, np_label
+
+
+SYNTAX_ITERATORS = {"noun_chunks": noun_chunks}
```
svlandeg, line 39: I've been testing this a bit with `nl_core_news_sm` and it seems to work quite well. One consistent error that I seem to get, is with a possessive. Take a simple example like "Haar vriend lacht luid." - it looks like the parse & POS tags are correct, but the noun chunck only includes the possessive "Haar" and not the noun "vriend". Try similar examples - they seem to have the same issue...

spacy/tests/lang/nl/test_noun_chunks.py
```diff
@@ -0,0 +1,10 @@
...
+    doc = nl_tokenizer(txt)
+    with pytest.raises(ValueError):
+        list(doc.noun_chunks)
```
svlandeg, line 10: Always very happy to see unit tests included with a new contribution :-) 

In addition to the "failing" test, it would be good to also have a positive test. You could define a document with tokens and parsing information, and the expected noun chunks, as has been done for Danish here: https://github.com/explosion/spaCy/blob/master/spacy/tests/lang/da/test_noun_chunks.py

Would you be able to add a few example cases for Dutch? We don't yet have these for all languages yet, but going forward I think it makes sense to include some more tests :-)

svlandeg (changes requested):
Thanks for the contribution, this looks great! 

I had a few reviewer comments so I'll put this on "request changes" for now just to help with our internal overview & administration of open PRs. Let me know if you have issues addressing any of it though!

spacy/lang/nl/syntax_iterators.py
julien-talkair: Will do

spacy/tests/lang/nl/test_noun_chunks.py
julien-talkair: Will do

adrianeboyd: Also for this test, a short text (one sentence) is fine. There's no benefit to having a long text here.

svlandeg: Ah, good point. Actually, this text isn't even correct Dutch most of the time, so just one correct Dutch sentence will suffice ;-)

julien-talkair:
Hi,

Taking remarks in to account, I got back to basics and definition of noun chunks (NLTK), also got a reference noun-phrase chunker in dutch (Frog) to make a reference chunking of a sample text.

One difficulty is that the SpaCy POS tagging has issues, marking a VERB as NOUN. The DEP tagging also marked a possessive determiner "Haar" in "Haar vriend" with a DEP=nmod:poss, where I think it should be DEP=det:poss.

I added also one error code, as in fact the noun_chunks should raise an exception if there is no DEP annotation, or if there is POS annotation.

All in all, the current code is quite different from my original copy/paste of FR and SV. Which leads me to question whether the implementation for these languages is correct.

Regards

spacy/tests/lang/nl/test_noun_chunks.py
julien-talkair: See recent commit, I have a short text with 4 sentences, along with a reference chunking.
There is one test to check that the ValueError exception is raised, and another test to check the chunking itself.

julien-talkair:
Looks like it didn't like the addition of the binary sample file. 
Fixed it by adding sample data as pytest fixture.

spacy/lang/nl/syntax_iterators.py
julien-talkair: Rewrote the entire function.

julien-talkair:
@svlandeg @adrianeboyd 
Reminder to review.

svlandeg:
Hi!

We'll definitely have a look, no reminders or additional pings necessary ;-)

spacy/lang/nl/syntax_iterators.py
```diff
@@ -0,0 +1,75 @@
...
+from ...symbols import NOUN, PRON
+from ...errors import Errors
+from ...tokens import Doc, Span
+
+
+def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Span]:
+    """
+    Detect base noun phrases from a dependency parse. Works on Doc and Span.
+    Takes definition from the enemy: https://www.nltk.org/book/ch07.html
+    Consider : [Noun + determinant / adjective] and also [Pronoun]df
```
svlandeg, line 12:
```suggestion
    The definition is inspired by https://www.nltk.org/book/ch07.html
    Consider : [Noun + determinant / adjective] and also [Pronoun]
```

spacy/tests/lang/nl/test_noun_chunks.py
```diff
@@ -0,0 +1,210 @@
...
+
+    with pytest.raises(ValueError):
+        list(doc.noun_chunks)
+
+
+def test_chunking(nl_sample, nl_reference_chunking):
+    """
+    Test the noun chunks of a sample text. Uses a sample.
+    The sample text was made as a Doc object with nl_core_news_md loaded. The Doc object was serialized with
+    the to_bytes() method.
```
svlandeg, line 207:
```suggestion
    The sample text simulates a Doc object as would be produced by nl_core_news_md.
```

spacy/lang/nl/syntax_iterators.py
```diff
@@ -0,0 +1,75 @@
...
+
+from ...symbols import NOUN, PRON
+from ...errors import Errors
+from ...tokens import Doc, Span
+
+
+def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Span]:
+    """
+    Detect base noun phrases from a dependency parse. Works on Doc and Span.
+    The definition is inspired by https://www.nltk.org/book/ch07.html
```
svlandeg, line 11: Did you only use the definition, or also the reference implementation from NLTK?

spacy/lang/nl/syntax_iterators.py
```diff
@@ -0,0 +1,75 @@
...
+            nsubjs = filter(
+                lambda x: x.dep == doc.vocab.strings["nsubj"], word.children
+            )
+            try:
+                _ = next(nsubjs)
+                # No exception, so we found some nsubj, we skip this word
+                continue
+            except StopIteration:
+                # No such thing, consider it a normal NOUN
+                pass
```
svlandeg, line 58: I think this is equivalent but shorter & easier to follow?

```suggestion
            next_word = next(nsubjs, None)
            if next_word:
                # We found some nsubj, so we skip this word. Otherwise, consider it a normal NOUN
                continue
```

svlandeg:
This looks great! I had just a few more minor comments/questions but otherwise I think this is fine to merge. The tests looks good and I did some additional runs myself where the results seemed perfect!

spacy/lang/nl/syntax_iterators.py
julien-talkair: I used the definition for my reference, but I did not have a look at the NLTK implementation.

spacy/lang/nl/syntax_iterators.py
julien-talkair: :+1: Indeed.

svlandeg (approved):

svlandeg:
Merged - thanks again for this contribution!

PR #8559, opened 2021-06-30
Fix vectors check for sourced components

adrianeboyd:
## Description

Since vectors are not loaded when components are sourced, store a hash for the vectors of each sourced component and compare it to the loaded vectors after the vectors are loaded from the `[initialize]` block.

Suppress the warning related to mismatched vectors in `add_pipe`, since the vectors are not loaded yet.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -934,6 +934,7 @@ class Language:
         # because factory may be used for something else
         self._pipe_meta.pop(name)
         self._pipe_configs.pop(name)
+        self.meta.get("_sourced_vectors_hashes", {}).pop(name)
         # Make sure name is removed from the [initialize] config
         if name in self._config["initialize"]["components"]:
             self._config["initialize"]["components"].pop(name)
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1680,6 +1680,8 @@ class Language:
         # If components are loaded from a source (existing models), we cache
         # them here so they're only loaded once
         source_nlps = {}
+        source_nlp_vectors_hashes = {}
+        nlp.meta["_sourced_vectors_hashes"] = {}
         for pipe_name in config["nlp"]["pipeline"]:
             if pipe_name not in pipeline:
                 opts = ", ".join(pipeline.keys())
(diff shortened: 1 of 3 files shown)


adrianeboyd:
This is clunky and brittle and hard to test. I think that at least it should fail silently if it fails. I think that the solution to providing a vectors warning without reloading anything is going to be something along these lines, but there may be a nicer way to implement it?

spacy/language.py
```diff
@@ -1719,7 +1721,12 @@ def from_config(
                                     name, source_name, pipe_cfg["replace_listeners"]
                                 )
                                 listeners_replaced = True
-                    nlp.add_pipe(source_name, source=source_nlps[model], name=pipe_name)
+                    with warnings.catch_warnings():
+                        warnings.filterwarnings("ignore", message="\\[W113\\]")
+                        nlp.add_pipe(source_name, source=source_nlps[model], name=pipe_name)
+                    if model not in source_nlp_vectors_hashes:
+                        source_nlp_vectors_hashes[model] = hash(source_nlps[model].vocab.vectors.to_bytes())
```
honnibal, line 1728: Ah this will take forever :(. We'd at least want to use murmurhash or something?

If we absolutely need to make some sort of signature, can we use a subset? Like maybe, first column, last column, first row, last row, size?

I just really don't think we can have this sort of overhead for a verification check. I could maybe consider it for an option that's off by default, but definitely not one that's on by default.

honnibal: Hm actually this only runs on sourcing right, not on every load?

I'm still not really thrilled about it, but it's less bad than I first thought.

adrianeboyd: In the situation where the alternative is that models fail silently, I think a few seconds for this check is okay. It should only run once for each specified source model, so it most use cases only for 1-2 models total. In tests with `spacy assemble`, the additional time wasn't particularly noticeable given that it also takes a few seconds to load models with vectors.

adrianeboyd:
For future reference, notes about alternatives here:

* attribute of `Language`: isn't serialized
* stored in config: is serialized, but isn't a valid config
* in meta: looks hacky, but works for serialization

The troublesome case where serialization is relevant is the following:

```python
nlp = Language().from_config(config)
nlp.to_disk("/path")
nlp = spacy.load("/path")
nlp.initialize()
```

svlandeg:
Ah, great idea to add these tests. I think we should just merge this - there's currently not a better alternative.

PR #8573, opened 2021-07-01
Black / Flake8 as Pre-Commit hooks

julien-talkair:
## Description
Using the pre-commit framework, installs a git pre-commit hook so that `black` and `flake8` run automatically on modified files when user initiates as `git commit`

### Types of change

New feature for devs.
Documentation change in CONTRIBUTING.md

## Checklist

### DIFF
--- a/.github/contributors/julien-talkair.md
+++ b/.github/contributors/julien-talkair.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
+documentation, or any other material posted or submitted by you to the project.
+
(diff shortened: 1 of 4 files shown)


ines:
Thanks, really appreciate the repo maintenance help 🙌

We've been thinking about giving pre-commit hooks for formatting a try for a while!

One quick question: Do you happen to know if there's an easy way to configure the settings for `flake8` for the pre-commit hook? Idealy, we would want to only run the most important checks (syntax errors etc.) that we also currently run in the CI:

https://github.com/explosion/spaCy/blob/2fc67e2aeb764760e0e950ffd99c43afddf11825/azure-pipelines.yml#L32

In general, we do want to make sure that the code follows all the rules defined in the `setup.cfg` (e.g. unused variables or imports), but we wouldn't necessarily want to enforce them on each commit. Very clear errors are obviously good, though, because those will always be bugs.

julien-talkair:
@ines it is feasible to give arguments to `flake8` indeed.
I tried with voluntary errors, and `flake8` failed only on those in the list.

On the other hand, since the pre-hook starts with `black`, some of the `flake8` errors will result anyway in an error with `black`. 

Just for conversation, I see that `black` is not used at all on the CI, is that a deliberate choice?
It's a requisite for devs in your [CONTRIBUTING.md](CONTRIBUTING.md).

julien-talkair:
On the other hand, maybe you meant to use [setup.cfg](setup.cfg) as the source for arguments... I have to check that.

ines:
> On the other hand, maybe you meant to use setup.cfg as the source for arguments... I have to check that.

Yes, that's how we do it for our end-to-end linting and it's detected automatically from there. But it's not necessarily something we would want to enforce for everyone – although, it's obviously desired that new code passes all of these linting rules.

> Just for conversation, I see that black is not used at all on the CI, is that a deliberate choice?

Yeah, we were actually just discussing this internally the other day and dedided to not enforce this because it'd otherwise lead to too many problems with user-submitted PRs, especially for users who are newer to Python programming and less familar with using linters and auto-formatters.  

Instead, we opted for an `autoblack` GitHub action that runs once a week and submits a PR: https://github.com/explosion/spaCy/actions/workflows/autoblack.yml (Due to how GitHub tokens work, it's unfortunately not possible to run this automatically for PRs, because those are mostly submitted from forks, which the token doesn't grant permissions to.)

julien-talkair:
@ines
Now flake8 in the pre-commit hook uses `setup.cfg` as a configuration.
I tested manually by adding intentional E266 anti-patterns, it was detected when the  `E266` was removed from the `ignore` line, and `flake8` stayed silent when it was in the `ignore` line.

ines:
Cool, thanks so much for your work on this 🙏 I'll go ahead and merge this and I hope this addition will be helpful to others.

PR #8606, opened 2021-07-05
Added ancient Greek language support

jmyerston:
Creating alpha support for ancient Greek

### Types of change

New feature

## Checklist

- [ X] I ran the tests, and all new and existing tests passed.


spacy/lang/grc/CONTRIBUTOR_AGREEMENT.md
```diff
@@ -0,0 +1,106 @@
...
+
+7. Please place an “x” on one of the applicable statement below. Please do NOT
+mark both statements:
+
+    * [x] I am signing on behalf of myself as an individual and no other person
+      or entity, including my employer, has or will have rights with respect to my
+      contributions.
+
+    * [x] I am signing on behalf of my employer or a legal entity and I have the
+      actual authority to contractually bind that entity.
```
svlandeg, line 95: Could you select either one of the two statements - but not both?

Also, this contributor statement should be moved to `.github/contributors` and renamed to `jmyerston.md`

spacy/lang/grc/__init__.py
```diff
@@ -0,0 +1,21 @@
+from typing import Optional
+from thinc.api import Model
```
svlandeg, line 2: I think these two imports can be removed

spacy/lang/grc/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+from .tokenizer_exceptions import TOKENIZER_EXCEPTIONS
+from .stop_words import STOP_WORDS
+from .lex_attrs import LEX_ATTRS
+from ...language import Language
+
+
+
```
svlandeg, line 9: Could you format your files with [`black`](https://black.readthedocs.io/en/stable/)? It will automatically make sure the format adheres to the rest of the code base - e.g. using only 2 newlines here instead of 3.

svlandeg (commented):
Hi, thanks so much for this contribution, this is great! I added a few minor comments that could be addressed, but I think in general this looks good. 

I'll change this PR to "in draft" to signal to everyone in the team that this is still being worked on and not entirely ready to merge yet. Once you're done - feel free to hit the "ready for review" button again!

svlandeg:
Ah, one more comment. Could you include some very basic test, that checks that the language can be run? These should go in a new directory `spacy/tests/lang/grc`.

For instance have a look at https://github.com/explosion/spaCy/blob/master/spacy/tests/lang/cs/test_text.py which does a simple test with a tokenizer & `like_num` functionality. For this to work, you'll also need to add a `grc_tokenizer` in `spacy/tests/conftest.py`.

Let us know if you need any more help with any of this!

jmyerston:
@svlandeg Thanks for your help! I made and pushed all changes requested. The test passed and models using UD grc corpora can be built with this addition. (I will be working on the models in the future after adding punctuation and noun chunks).  I do not know, though, if I made the commits and push in the right way: I could not find the "ready for review" button. It is my first time contributing to a project like this. 

I will be submitting the lemma and normalization tables to spacy-lookup-data soon.

spacy/lang/grc/examples.py
```diff
@@ -0,0 +1,17 @@
...
+>>> docs = nlp.pipe(sentences)
+"""
+
+
+sentences = [
+    "ἐρᾷ μὲν ἁγνὸς οὐρανὸς τρῶσαι χθόνα, ἔρως δὲ γαῖαν λαμβάνει γάμου τυχεῖν·",
+    "εὐδαίμων Χαρίτων καὶ Μελάνιππος ἔφυ, θείας ἁγητῆρες ἐφαμερίοις φιλότατος."
+    "ὃ μὲν δὴ ἀπόστολος ἐς τὴν Μίλητον ἦν.",
+    "Θρασύβουλος δὲ σαφέως προπεπυσμένος πάντα λόγον καὶ εἰδὼς τὰ Ἀλυάττης μέλλοι ποιήσειν μηχανᾶται τοιάδε.",
+    "φιλόπαις δ' ἦν ἐκμανῶς καὶ Ἀλέξανδρος ὁ βασιλεύς."
```
adrianeboyd, line 14: Looking good overall! It looks like there are a few commas missing between items in this list.

jmyerston: Oh yeah, sorry,

spacy/lang/grc/lex_attrs.py
```diff
@@ -0,0 +1,309 @@
...
+    "δυοῖν",
+    "τρεῖς",
+    "τριῶν",
+    "τρισί",
+    "τρία",
+    "τέτταρες",
+    "τεττάρων",
+    "τέτταρσι",
+    "τέτταρα",
+    "τέτταρας" "πέντε",
```
adrianeboyd, line 25: Here reformatting with `black` shows where the commas are missing. (This is a very common typo in these kinds of lists, no worries!)

spacy/tests/lang/grc/test_text.py
```diff
@@ -0,0 +1,23 @@
+import pytest
```
adrianeboyd, line 1: Here, be sure to also add `__init__.py` in this directory or `pytest` will run into problems.

adrianeboyd: To clarify, `__init__.py` should be an empty file.

spacy/lang/grc/lex_attrs.py
jmyerston: I ran black on it, bu I did not see any.

svlandeg:
@jmyerston: the changes look great! Thanks for your quick follow-up :-)

spacy/lang/grc/lex_attrs.py
jmyerston: Now I see how black works. Thanks.

spacy/tests/lang/grc/test_text.py
jmyerston: Does the __init__.py go empty? The one for Spanish has no text in it...

jmyerston: figured out. thanks.

adrianeboyd:
Thanks for the quick responses and updates! The main remaining concern: the stop word list seems to be unusually long. Can you explain a bit about the source of the words in this list? We would probably like to provide a shorter list as the default because this file is included with every single spacy installation.

jmyerston:
This is the standard world list for ancient Greek compiled by Aurélien Berra. It is now also part of the R Quanteda's word lists. It has been used for topic analysis and authorship attribution with good results.  I think it is longer than other word lists because of the highly inflected nature of ancient Greek and the fact that grc has several dialects in which various texts were written. So, the list may contain the same words in different dialects. 

The list could be shortened by restricting it to one dialect like Attic, but it would not work for those designing applications that process the entire grc corpus.  It is up to you guys, but I would recommend leave it as it is.

jmyerston:
Ok.  I see that spacy stop word lists are about 3k. I will then redo the stops and bring it to a similar size.

svlandeg:
Thanks again for your work on this!

PR #8609, opened 2021-07-06
Fix a command typo in models.md

mathcass:
"dowmload" -> "download"

## Description

Just a PR to fix a typo in the documentation on the [models page](https://spacy.io/usage/models). 

### Types of change

Changes a typo in a suggested command on the models page. Just to be extra sure, I tried out the command. (It doesn't work 😄.)

```sh
$ python -m spacy dowmload en_core_web_sm
Usage: python -m spacy [OPTIONS] COMMAND [ARGS]...
Try 'python -m spacy --help' for help.

Error: No such command 'dowmload'.
```

## Checklist

- [ ] I have submitted the spaCy Contributor Agreement.
- [ ] I ran the tests, and all new and existing tests passed.
- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #8610, opened 2021-07-06
Update models.md

c-chaitanya:
## Checklist


polm:
Thanks for the PR, #8609 beat you to the fix though, so I'm going to close this one.

Also note that we don't need a contributor agreement for tiny fixes like this.

PR #8640, opened 2021-07-08
Update lexeme ranks for loaded vectors

adrianeboyd:
## Description

Update the ranks for any lexemes that have been added to the vocab before the vectors are added to the model.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
Maybe it would better to get rid of this link altogether, though, and modify `PretrainVectors` like `StaticVectors`?

honnibal:
I think modifying `PretrainVectors` to not use the `lex.rank` sounds good. Double check the speed but I don't think that will be an issue.

PR #8702, opened 2021-07-13
Added DaCy to universe

KennethEnevoldsen:
### Types of change
Documentation/universe update

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -897,7 +897,7 @@
             "pip": "dacy",
             "code_example": [
                 "import dacy",
-                "dacy.models() # get a list of dacy models",
+                "print(dacy.models()) # get a list of dacy models",
                 "nlp = dacy.load('medium')  # load your spacy pipeline",
                 "",
                 "# DaCy also includes functionality for adding other Danish models to the pipeline",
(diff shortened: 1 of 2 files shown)


website/meta/universe.json
```diff
@@ -888,6 +888,31 @@
...
+        {
+            "id": "dacy",
+            "title": "DaCy",
+            "slogan": "An efficient Pipeline for Danish NLP",
+            "description": "DaCy is a Danish preprocessing pipeline trained in SpaCy. It has achieved State-of-the-Art performance on Named entity recognition, part-of-speech tagging and dependency parsing for Danish. This repository contains material for using the DaCy, reproducing the results and guides on usage of the package. Furthermore, it also contains a series of behavioural test for biases and robustness of Danish NLP pipelines.",
+            "github": "centre-for-humanities-computing/DaCy",
+            "pip": "dacy",
+            "code_example": [
+                "import dacy",
+                "dacy.models() # get a list of dacy models",
```
adrianeboyd, line 900: We're happy to add this! I noticed that this line doesn't show anything in the demo script?

KennethEnevoldsen: Glad to hear! - I am not entirely sure I understand the last part though @adrianeboyd?

ines: I think what @adrianeboyd meant is that `dacy.models()` returns the list of models so just running the example script as is wouldn't output anything, which is potentially confusing. So in that case, maybe one of the following would be clearer:

```suggestion
                "print(dacy.models()) # get a list of dacy models",
```

```suggestion
                "model_names = dacy.models() # get a list of dacy models",
```

KennethEnevoldsen: Ahh thanks for the clarification - I have changed it to print now

PR #8745, opened 2021-07-17
Add development docs for Language and code conventions

ines:
## Description

I ended up putting this in `extra` so we can have a top-level README and because it's otherwise too difficult to keep the files together with the code (since `language.py` lives within the top-level package).

Rendered view: https://github.com/explosion/spaCy/tree/docs/dev-docs-language/extra/DEVELOPER_DOCS

### Types of change
docs

## Checklist


extra/DEVELOPER_DOCS/Language.md
```diff
@@ -0,0 +1,136 @@
...
+- Resolve the config and create all functions it refers to (e.g. `model`). This gives us the actual component function that we can insert into the pipeline.
+
+### Sourcing a pipeline component
+
+spaCy also allows "sourcing" a component, which will copy it over from an existing pipeline. In this case, `Language.add_pipe` will delegate to `Language.create_pipe_from_source`. In order to copy a component effectively and validate it, the source pipeline first needs to be loaded. This is done in `Language.from_config`, so a source pipeline only has to be loaded once if multiple components source from it. Sourcing a component will perform the following checks and modifications:
+
+- If the source pipeline contains vectors, check that they match the destination pipeline's vectors and warn if not (since different vectors that are used as features in components can lead to degraded performance). This is done by comparing the shape first and finally falls back to comparing the actual byte representation of the vectors (which is slower). We also store a hash of the vectors in the pipeline meta so we're able to check this more easily.
```
adrianeboyd, line 70: I think this is missing an explanation of the delayed vector loading so that they're not loaded twice. The only reason for storing the hash is for `source` from a config, not for `create_pipe_from_source` itself.

ines: Good point! Do you want to add that part? Since you implemented the source hashing, I think you might be able to explain this better than I can.

extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -0,0 +1,546 @@
...
+It's fine to include code comments that indicate future TODOs, using the `TODO:` prefix. Modern editors typically format this in a different color, so it's easy to spot. TODOs don't necessarily have to be things that are absolutely critical to fix fight now – those should already be addressed in your pull request once it's ready for review. But they can include notes about potential future improvements.
+
+```diff
++ # TODO: this is currently pretty slow
+dir_checksum = hashlib.md5()
+for sub_file in sorted(fp for fp in path.rglob("*") if fp.is_file()):
+    dir_checksum.update(sub_file.read_bytes())
+```
+
+If any of the TODOs you've added are important and should be fixed soon, you should add a task for this on the project board or an issue on the issue tracker to make sure we don't forget to address it.
```
svlandeg, line 136:
```suggestion
If any of the TODOs you've added are important and should be fixed soon, you should add a task for this on Explosion's internal Ora board or an issue on the public issue tracker to make sure we don't forget to address it.
```

extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -0,0 +1,546 @@
...
+
+### Don't use `print`
+
+The core library never `print`s anything. While we encourage using `print` statements for simple debugging (it's the most straightforward way of looking at what's happening), make sure to clean them up once you're ready to submit your pull request. If you want to output warnings or debugging information for users, use the respective dedicated mechanisms for this instead (see sections on warnings and logging for details).
+
+The only exceptions are the CLI functions, which pretty-print messages for the user, and methods that are explicitly intended for printing things, e.g. `Language.analyze_pipes` with `pretty=True` enabled. For this, we use our lightweight helper library [`wasabi`](https://github.com/ines/wasabi).
+
+## Naming
+
+Naming is hard and it's probably one of the main things we discuss in our regular PR meetings. We don't expect you to come up with the perfect names for everything you write – finding the right names is often an iterative and collaborative process. That said, we do try to follow some basic conventions.
```
svlandeg, line 310:
```suggestion
Naming is hard and often a topic of long internal discussions. We don't expect you to come up with the perfect names for everything you write – finding the right names is often an iterative and collaborative process. That said, we do try to follow some basic conventions.
```

extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -0,0 +1,546 @@
...
+### Warnings
+
+Instead of raising an error, some parts of the code base can raise warnings to notify the user of a potential problem. This is done using Python's `warnings.warn` and the messages defined in `Warnings` in the `errors.py`. Whether or not warnings are shown can be controlled by the user, including custom filters for disabling specific warnings using a regular expression matching our internal codes, e.g. `W123`.
+
+```diff
+- print("Warning: No examples provided for validation")
++ warnings.warn(Warnings.W123)
+```
+
+When adding warnings, make sure you're not calling `warnings.warn` repeatedly, e.g. in a loop, which will clog up the terminal output. Instead, you can collect the potential problems first and then raise a single warning. If the problem is critical, consider raising an error instead and to exit early.
```
svlandeg, line 417: This didn't parse for me, and I think "raising an error" implies exiting early anyway.

```suggestion
When adding warnings, make sure you're not calling `warnings.warn` repeatedly, e.g. in a loop, which will clog up the terminal output. Instead, you can collect the potential problems first and then raise a single warning. If the problem is critical, consider raising an error instead.
```

extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -0,0 +1,546 @@
...
+
+The test suite also provides [fixtures](https://github.com/explosion/spaCy/blob/master/spacy/tests/conftest.py) for different language tokenizers that can be used as function arguments of the same name and will be passed in automatically. Those should only be used for tests related to those specific languages. We also have [test utility functions](https://github.com/explosion/spaCy/blob/master/spacy/tests/util.py) for common operations, like creating a temporary file.
+
+### Constructing objects and state
+
+Test functions usually follow the same simple structure: they set up some state, perform the operation you want to test and `assert` conditions that you expect to be true, usually before and after the operation.
+
+Tests should focus on exactly what they're testing and avoid dependencies on other unrelated library functionality wherever possible. If all your tests needs is a `Doc` object with certain annotations set, you should always construct it manually:
```
svlandeg, line 455:
```suggestion
Tests should focus on exactly what they're testing and avoid dependencies on other unrelated library functionality wherever possible. If all your test needs is a `Doc` object with certain annotations set, you should always construct it manually:
```

extra/DEVELOPER_DOCS/Language.md
```diff
@@ -0,0 +1,137 @@
...
+- Fill the config to make sure all unspecified defaults from the function arguments are added and update the `raw_config` (uniterpolated with variables intact) with that information, so the component config we store in `nlp.config` is up to date. We do this by adding the `raw_config` _into_ the filled config – otherwise, the references to variables would be overwritten.
+- Resolve the config and create all functions it refers to (e.g. `model`). This gives us the actual component function that we can insert into the pipeline.
+
+### Sourcing a pipeline component
+
+spaCy also allows "sourcing" a component, which will copy it over from an existing pipeline. In this case, `Language.add_pipe` will delegate to `Language.create_pipe_from_source`. In order to copy a component effectively and validate it, the source pipeline first needs to be loaded. This is done in `Language.from_config`, so a source pipeline only has to be loaded once if multiple components source from it. Sourcing a component will perform the following checks and modifications:
```
svlandeg, line 68: Maybe link to https://spacy.io/usage/processing-pipelines#sourced-components ?

extra/DEVELOPER_DOCS/Language.md
```diff
@@ -0,0 +1,137 @@
...
+- Check if we have a language-specific factory for the given `nlp.lang` and if not, fall back to the global factory.
+- Construct the component config, consisting of whatever arguments were provided, plus the current `nlp` object and `name`, which are default expected arguments of all factories. We also add a reference to the `@factories` registry, so we can resolve the config via the registry, like any other config. With the added `nlp` and `name`, it should now include all expected arguments of the given function.
+- Fill the config to make sure all unspecified defaults from the function arguments are added and update the `raw_config` (uniterpolated with variables intact) with that information, so the component config we store in `nlp.config` is up to date. We do this by adding the `raw_config` _into_ the filled config – otherwise, the references to variables would be overwritten.
```
svlandeg, line 63:
```suggestion
- Fill the config to make sure all unspecified defaults from the function arguments are added and update the `raw_config` (uninterpolated with variables intact) with that information, so the component config we store in `nlp.config` is up to date. We do this by adding the `raw_config` _into_ the filled config – otherwise, the references to variables would be overwritten.
```

svlandeg:
This is great!

- I think it makes sense to have this under `extra` so it all stays together in one place.
- **Code conventions**: it's really nice to have this detailed overview. I learned most of this by inspecting the current code base and by getting reviewer comments on my PRs, but it makes so much sense to make a lot of this explicit. I think this will be very useful for new people to jump in :-)
- **Language**: While there is a LOT of useful information on this page, I feel like it's not very easy to navigate through. I'm imaging a new dev on the team wanting to debug some behaviour and looking for some relevant background. Will they need to read through all of this? Do we need more hyperlinks? A more extended index table at the top? Maybe even a table with the different sections and a one-sentence highlight what that section covers? Some schema's? Because there is just so much and all of it is so detailed and technical, it's almost undoable for me to read through in one go (and I already know most of the code, so I imagine it's even more difficult otherwise)

extra/DEVELOPER_DOCS/Code Conventions.md
ines: Maybe we could make this "raising an error instead (which will exit early)"? I just wanted to include this as a reminder that it's generally good to exit as early as possible if you're anticipating an error.

ines:
I definitely see what you mean about the Language page, but the idea here is really to provide a more general conceptual overview. So if someone is working on the component generation from the config, I do think the answer is "just read the whole section" 😅 It's difficult to break this down more, because it's potentially all relevant background information. We can definitely have a more detailed table of contents, though!

extra/DEVELOPER_DOCS/Code Conventions.md
svlandeg: Sure!

svlandeg:
Ok, I restructured it slightly so that there are 5 main subtitles, and I numbered them because I think it helps with processing all of this text + extended the index on top.

extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -0,0 +1,546 @@
...
++    raise ValueError(Errors.E456.format(value=value))
+```
+
+As a general rule of thumb, all error messages raised within the **core library** should be added to `Errors`. The only place where we write errors and messages as strings is `spacy.cli`, since these functions typically pretty-print and generate a lot of output that'd otherwise be very difficult to separate from the actual logic.
+
+### Re-raising exceptions
+
+If we anticipate possible errors in third-party code that we don't control, or our own code in a very different context, we typically try to provide custom and more specific error messages if possible. If we need to re-raise an exception within a `try`/`except` block, we can re-raise a custom exception.
+
+[Re-raising `from`](https://docs.python.org/3/tutorial/errors.html#exception-chaining) the original caught exception lets us chain the exceptions, so the user sees both the original error, as well as the custom message with a note "The above exception was the direct cause of the following exception".
```
honnibal, line 372: I think we should use `raise from` extremely carefully, as long tracebacks are much preferable to swallowing the error. So I'd probably not mention it in the docs.

extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -0,0 +1,546 @@
...
+
++ value1 = get_some_value()
++ value2 = get_some_other_value()
++ try:
++     score = external_library.compute_some_score(value1, value2)
++ except ValueError:
++     score = 0.0
+```
+
+### Avoid lambda functions
```
honnibal, line 260: I think this section is a bit strongly worded. There's still tonnes of times where lambdas are the right thing to do, e.g. when you're passing a key to a sorting function.

The issue is really when you're passing lambdas into somewhere where they might get stored. So people should be a bit careful of using them, but avoiding lambdas entirely could make for some ugly code.

extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -0,0 +1,546 @@
...
++    raise ValueError(Errors.E123)
+
+if not isinstance(value, int):
+-    raise ValueError(f"Unexpected value: {value}")
++    raise ValueError(Errors.E456.format(value=value))
+```
+
+As a general rule of thumb, all error messages raised within the **core library** should be added to `Errors`. The only place where we write errors and messages as strings is `spacy.cli`, since these functions typically pretty-print and generate a lot of output that'd otherwise be very difficult to separate from the actual logic.
+
+### Re-raising exceptions
```
honnibal, line 368: Nervous about mentioning this, as it can cause a lot of trouble. I think we use this much too often ourselves, so I'm reluctant to encourage people to author code with it.

ines: This section is about re-raising exceptions in general, though, and I think the way we're using it is fine? Re-raising from an existing exception is a really good pattern IMO and has zero impact on the traceback. It only changes the wording of how the re-raised exception is presented in Python ("...was the direct cause..."). I think this is very good and useful.

Re-raising from `None` is more problematic if it's used incorrectly, but I don't think we're really doing that. We previously had a problem where we were polluting people's traceback with chained errors that were not relevant to the exception we were re-raising. Maybe we just need a more explicit warning here that it should only be used for exceptions we can very clearly anticipate, and that it requires careful error messages.

extra/DEVELOPER_DOCS/Code Conventions.md
ines: Sure, we can phrase this so it sounds less strict. I do think there are very few cases where lambdas are the right choice. Sort keys that are short enough to be readable in one line are the main exception.

Btw, fun fact: The only places in spaCy where we currently use lambdas are serializers, `get_examples` callbacks and tests. Nothing else.

PR #8754, opened 2021-07-18
[minor] removed outdated spacy version for spacymoji

KennethEnevoldsen:
From the documentation of spacymoji (and the requirements.txt) it seems like it is not only for version 2.

## Checklist
- [ ] I ran the tests, and all new and existing tests passed. (The change shouldn't affect tests)


polm:
Thanks for the fix!

PR #8766, opened 2021-07-20
Refactor scoring methods to use registered functions

adrianeboyd:
## Description

Add scorer option to components

Add registered scorers for all components

* Add `scorers` registry
* Move all scoring methods outside of components as independent
  functions and register
* Use the registered scoring methods as defaults in configs and inits

Additional:

* The scoring methods no longer have access to the full component, so
  use settings from `cfg` as default scorer options to handle settings
  such as `labels`, `threshold`, and `positive_label`
* The `attribute_ruler` scoring method no longer has access to the
  patterns, so all scoring methods are called
* Bug fix: `spancat` scoring method is updated to set `allow_overlap` to
  score overlapping spans correctly

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/pipeline/pipe.pyx
+++ b/spacy/pipeline/pipe.pyx
@@ -86,6 +86,9 @@ cdef class Pipe:
             # use default settings from cfg (e.g., threshold)
             if hasattr(self, "cfg") and isinstance(self.cfg, dict):
                 scorer_kwargs.update(self.cfg)
+            # override self.cfg["labels"] with self.labels
+            if hasattr(self, "labels"):
+                scorer_kwargs["labels"] = self.labels
             # override with kwargs settings
             scorer_kwargs.update(kwargs)
             return self.scorer(examples, **scorer_kwargs)
--- a/spacy/pipeline/pipe.pyx
+++ b/spacy/pipeline/pipe.pyx
@@ -84,7 +84,7 @@ cdef class Pipe:
         if hasattr(self, "scorer") and self.scorer is not None:
             scorer_kwargs = {}
             # use default settings from cfg (e.g., threshold)
-            if hasattr(self, "cfg"):
+            if hasattr(self, "cfg") and isinstance(cfg, dict):
                 scorer_kwargs.update(self.cfg)
             # override with kwargs settings
             scorer_kwargs.update(kwargs)
--- a/spacy/pipeline/pipe.pyx
+++ b/spacy/pipeline/pipe.pyx
@@ -84,7 +84,7 @@ cdef class Pipe:
         if hasattr(self, "scorer") and self.scorer is not None:
             scorer_kwargs = {}
             # use default settings from cfg (e.g., threshold)
-            if hasattr(self, "cfg") and isinstance(cfg, dict):
+            if hasattr(self, "cfg") and isinstance(self.cfg, dict):
                 scorer_kwargs.update(self.cfg)
             # override with kwargs settings
             scorer_kwargs.update(kwargs)
--- a/spacy/pipeline/pipe.pyx
+++ b/spacy/pipeline/pipe.pyx
@@ -81,6 +81,14 @@ cdef class Pipe:
 
         DOCS: https://spacy.io/api/pipe#score
         """
+        if hasattr(self, "scorer") and self.scorer is not None:
+            scorer_kwargs = {}
+            # use default settings from cfg (e.g., threshold)
+            if hasattr(self, "cfg"):
+                scorer_kwargs.update(self.cfg)
+            # override with kwargs settings
+            scorer_kwargs.update(kwargs)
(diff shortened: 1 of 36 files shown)


adrianeboyd:
Docs to be updated after the rest is finalized.

The intention is that this is backwards compatible, so I don't think it needs to wait until v3.2.

spacy/pipeline/sentencizer.pyx
```diff
@@ -41,7 +45,13 @@ class Sentencizer(Pipe):
             '𑩃', '𑪛', '𑪜', '𑱁', '𑱂', '𖩮', '𖩯', '𖫵', '𖬷', '𖬸', '𖭄', '𛲟', '𝪈',
             '｡', '。']
 
-    def __init__(self, name="sentencizer", *, punct_chars=None):
+    def __init__(
+        self,
+        name="sentencizer",
+        *,
+        punct_chars=None,
+        scorer=senter_score
```
svlandeg, line 53:
```suggestion
        scorer=senter_score,
```

spacy/pipeline/senter.pyx
```diff
@@ -34,19 +36,35 @@ DEFAULT_SENTER_MODEL = Config().from_str(default_model_config)["model"]
...
     default_score_weights={"sents_f": 1.0, "sents_p": 0.0, "sents_r": 0.0},
 )
-def make_senter(nlp: Language, name: str, model: Model):
-    return SentenceRecognizer(nlp.vocab, model, name)
+def make_senter(nlp: Language, name: str, model: Model, scorer: Optional[Callable]):
+    return SentenceRecognizer(nlp.vocab, model, name, scorer=scorer)
+
+
+def senter_score(examples, **kwargs):
+    validate_examples(examples, "SentenceRecognizer.score")
```
svlandeg, line 47: The reference to `SentenceRecognizer` should probably be generalized to use the component's (class) name instead, as this function is also used by the Sentencizer. I guess the same is true for all other registered scoring components, as they can also be used in different components. We could also just call `validate_examples` with `method="score"` and avoid the whole naming altogether...

spacy/pipeline/pipe.pyx
```diff
@@ -81,6 +81,14 @@ cdef class Pipe:
 
         DOCS: https://spacy.io/api/pipe#score
         """
+        if hasattr(self, "scorer") and self.scorer is not None:
+            scorer_kwargs = {}
+            # use default settings from cfg (e.g., threshold)
+            if hasattr(self, "cfg") and isinstance(self.cfg, dict):
+                scorer_kwargs.update(self.cfg)
```
svlandeg, line 88: This make sense. 

One exception here is that `self.cfg["labels"]` isn't always the same as `self.labels`. Maybe for robustness, it would make sense to account for that field specifically?

Something like
```
if hasattr(self, "labels"):
    scorer_kwargs["labels"] = self.labels
```

spacy/pipeline/spancat.py
```diff
@@ -373,28 +397,6 @@ def initialize(
...
-        RETURNS (Dict[str, Any]): The scores, produced by Scorer.score_cats.
-
-        DOCS: https://spacy.io/api/spancategorizer#score
-        """
-        validate_examples(examples, "SpanCategorizer.score")
-        self._validate_categories(examples)
-        kwargs = dict(kwargs)
-        attr_prefix = "spans_"
-        kwargs.setdefault("attr", f"{attr_prefix}{self.key}")
-        kwargs.setdefault("labels", self.labels)
```
svlandeg, line 389: Odd that this was in the codebase, `score_spans` doesn't actually use a `labels` argument...

svlandeg (commented):
I really like this - it fits in perfectly with the modular approach of v3. I only had a few minor comments.

I do think I'd prefer having this in 3.2, as it's quite a substantial change & there might be subtle problems underneath with the `kwargs` being updated from `self.cfg` etc. I also get nervous about potential binary incompatibilities because there are changes in the `pxd` files. 

We could merge this to `develop` instead?

spacy/pipeline/spancat.py
adrianeboyd: Yeah, I think there were a number of copy-paste problems with the scoring, but nothing major other than `multi_label`.

spacy/pipeline/senter.pyx
adrianeboyd: I think `"score"` might be okay. Or maybe we could skip the `validate_examples` in all the individual scoring functions and just leave it `Language.evaluate` instead? I think we can live with not catching this error in advance in the individual `.score` methods? I haven't done any timing tests, but it's a lot of repetitive checking.

svlandeg: I've actually been wondering whether we truly need this, yea. I wouldn't mind removing it, it would certainly make things cleaner.

PR #8773, opened 2021-07-21
Improve type annotation for `Language.pipe`

### LINKED ISSUE #8772: Type annotation of `Language.pipe` missing main variant

Here's an example from the docs of how to use `Language.pipe`:

```python
# Setup code
import spacy

# From docs
texts = ["One document.", "...", "Lots of documents"]
for doc in nlp.pipe(texts, batch_size=50):
    assert doc.has_annotation("DEP")
```

Running `mypy` on this file produces the following errors:
```
test.py:7: error: Need type annotation for "doc"
test.py:7: error: Argument 1 to "pipe" of "Language" has incompatible type "List[str]"; expected "Iterable[Tuple[str, <nothing>]]"
test.py:8: error: "Tuple[Any, Any]" has no attribute "has_annotation"
Found 3 errors in 1 file (checked 1 source file)
```

Here's [the type signature of `Language.pipe`](https://github.com/explosion/spaCy/blob/d48c01a6f77a90db06e936ae903d1d0ea151ae31/spacy/language.py#L1437-L1459):

```python
    @overload
    def pipe(
        self,
        texts: Iterable[Tuple[str, _AnyContext]],
        *,
        as_tuples: bool = ...,
        batch_size: Optional[int] = ...,
        disable: Iterable[str] = ...,
        component_cfg: Optional[Dict[str, Dict[str, Any]]] = ...,
        n_process: int = ...,
    ) -> Iterator[Tuple[Doc, _AnyContext]]:
        ...


    def pipe(  # noqa: F811
        self,
        texts: Iterable[str],
        *,
        as_tuples: bool = False,
        batch_size: Optional[int] = None,
        disable: Iterable[str] = SimpleFrozenList(),
        component_cfg: Optional[Dict[str, Dict[str, Any]]] = None,
        n_process: int = 1,
    ) -> Iterator[Doc]:
```

The Python `typing` docs state:

> A series of @overload-decorated definitions must be followed by exactly one non-@overload-decorated definition (for the same function/method). The @overload-decorated definitions are for the benefit of the type checker only, since they will be overwritten by the non-@overload-decorated definition, while the latter is used at runtime but should be ignored by a type checker.

In other words, `mypy` is only looking at the type signature defined with `@overload`, and not the "main" type signature.

To fix this, the main type signature should be copied and added as an `@overload` variant for `Language.pipe`.

Ideally the type signature of the non-`@overload`-decorated definition should also be updated to indicate that it can be called in either form.

Even better, the `as_tuples` `@overload` variant should be modified to use `typing.Literal` (`spacy.compat.Literal`) to help type checkers infer which variant of `pipe` is being called.

## Your Environment

- **spaCy version:** 3.1.1
- **Platform:** macOS-11.4-x86_64-i386-64bit
- **Python version:** 3.9.6
- **Pipelines:** en_core_web_sm (3.1.0)

connorbrinton:
## Description

These changes add a missing overload variant to the type signature of
`Language.pipe`. Additionally, the type signature is enhanced to allow
type checkers to differentiate between the two overload variants based
on the `as_tuple` parameter.

Fixes #8772.

Since `mypy` isn't listed in `requirements.txt`, I wasn't able to come up with a regression test for this issue. I have tested manually and have verified that Mypy now accepts usage of `Language.pipe` with `as_tuples=False` with these changes. If there's interest in including `mypy` as a development dependency and writing a regression test I can add that to this PR 🙂 

### Types of change

Bug fix and minor enhancement

## Checklist

### DIFF
--- a/.github/contributors/connorbrinton.md
+++ b/.github/contributors/connorbrinton.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
+[ExplosionAI GmbH](https://explosion.ai/legal). The term
+**"you"** shall mean the person or entity identified below.
+
+If you agree to be bound by these terms, fill in the information requested
+below and include the filled-in version with your first pull request, under the
+folder [`.github/contributors/`](/.github/contributors/). The name of the file
+should be your GitHub username, with the extension `.md`. For example, the user
+example_user would create the file `.github/contributors/example_user.md`.
+
+Read this agreement carefully before signing. These terms and conditions
+constitute a binding legal agreement.
+
+## Contributor Agreement
+
+1. The term "contribution" or "contributed materials" means any source code,
+object code, patch, tool, sample, graphic, specification, manual,
+documentation, or any other material posted or submitted by you to the project.
+
+2. With respect to any worldwide copyrights, or copyright applications and
+registrations, in your contribution:
+
+    * you hereby assign to us joint ownership, and to the extent that such
+    assignment is or becomes invalid, ineffective or unenforceable, you hereby
+    grant to us a perpetual, irrevocable, non-exclusive, worldwide, no-charge,
+    royalty-free, unrestricted license to exercise all rights under those
+    copyrights. This includes, at our option, the right to sublicense these same
+    rights to third parties through multiple levels of sublicensees or other
+    licensing arrangements;
+
+    * you agree that each of us can do all things in relation to your
+    contribution as if each of us were the sole owners, and if one of us makes
+    a derivative work of your contribution, the one who makes the derivative
+    work (or has it made will be the sole owner of that derivative work;
+
+    * you agree that you will not assert any moral rights in your contribution
+    against us, our licensees or transferees;
+
+    * you agree that we may register a copyright in your contribution and
+    exercise all ownership rights associated with it; and
+
+    * you agree that neither of us has any duty to consult with, obtain the
+    consent of, pay or render an accounting to the other for any use or
+    distribution of your contribution.
+
+3. With respect to any patents you own, or that you can license without payment
+to any third party, you hereby grant to us a perpetual, irrevocable,
+non-exclusive, worldwide, no-charge, royalty-free license to:
+
+    * make, have made, use, sell, offer to sell, import, and otherwise transfer
+    your contribution in whole or in part, alone or in combination with or
+    included in any product, work or materials arising out of the project to
+    which your contribution was submitted, and
+
+    * at our option, to sublicense these same rights to third parties through
+    multiple levels of sublicensees or other licensing arrangements.
+
+4. Except as set out above, you keep all right, title, and interest in your
+contribution. The rights that you grant to us under these terms are effective
+on the date you first submitted a contribution to us, even if your submission
+took place before the date you sign these terms.
+
+5. You covenant, represent, warrant and agree that:
+
+    * Each contribution that you submit is and shall be an original work of
+    authorship and you can legally grant the rights set out in this SCA;
+
+    * to the best of your knowledge, each contribution will not violate any
+    third party's copyrights, trademarks, patents, or other intellectual
+    property rights; and
+
+    * each contribution shall be in compliance with U.S. export control laws and
+    other applicable export and import laws. You agree to notify us if you
+    become aware of any circumstance which would make any of the foregoing
+    representations inaccurate in any respect. We may publicly disclose your
+    participation in the project, including the fact that you have signed the SCA.
+
+6. This SCA is governed by the laws of the State of California and applicable
+U.S. Federal law. Any choice of law rules will not apply.
+
+7. Please place an “x” on one of the applicable statement below. Please do NOT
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Sure, it would probably make sense to add `mypy` and tests here, too, which looks like it will turn up a few more issues.

As a minimal check, we could add this to the CI, similar to what we do in `thinc`:

```none
  - script: python -m mypy spacy
    displayName: 'Run mypy'
```

I currently get an `INTERNAL ERROR` related to with `tok2vec`, so this may be a minor adventure, but in the end it's probably all something that needs to be addressed in some way.

connorbrinton:
> I currently get an `INTERNAL ERROR` related to `tok2vec`, so this may be a minor adventure, but in the end it's probably all something that needs to be addressed in some way.

I ran into that while manually testing this PR as well. The traceback indicates that the crash originates in `thinc`'s `mypy` plugin. I temporarily disabled that plugin to manually test this PR, but obviously it would be better to get the root issue fixed. I'll file a bug report on the `thinc` repository and see if I can figure out the cause 👍

> Sure, it would probably make sense to add `mypy` and tests here, too, which looks like it will turn up a few more issues.
> 
> As a minimal check, we could add this to the CI, similar to what we do in `thinc`:
> 
> ```
>   - script: python -m mypy spacy
>     displayName: 'Run mypy'
> ```

Agreed, I think that would be a good idea. Since there currently are a number of typing errors, a simple path to getting this check deployed might be:
* Fix the `thinc` `mypy` plugin crash
* Run `mypy` on the codebase
* Add `# type: ignore[<code>]` directives to every problematic line
* Add the `mypy` check to the CI pipeline
* Over time, eliminate the `# type: ignore[<code>]` directives

Adding all of the necessary `# type: ignore[<code>]` directives probably wouldn't be ideal from a `git blame` perspective, but I suppose it will depend on how many lines are affected.

I'll try to figure out the `thinc` `mypy` plugin bug and create a new PR adding a `mypy` check to the codebase within the next week 👍

svlandeg:
Hi @connorbrinton, thanks for looking into this! Let us know how you go. I guess in the meantime I could mark this PR as draft?

Alternatively, we could also merge and add the regression test (hopefully) later. What do you think, @adrianeboyd ?

connorbrinton:
Hi @svlandeg! I went down a bit of a rabbit hole investigating the Thinc Mypy plugin crash (😅), but just finished submitting an issue describing the issue and the cause/potential fix.

I'm happy to wait for a resolution on that issue before merging here with a regression test. Alternatively there are a number of usages of `pipe(Sequence[str])` within the SpaCy codebase (like [this one](https://github.com/explosion/spaCy/blob/175847f92cb89f2515f848c5eda8872af247954f/spacy/cli/evaluate.py#L119)) which I think could serve as effective regression tests once Mypy is enabled across SpaCy's codebase (and `# type: ignore` directives are removed). I'm totally happy with either way 🙂

connorbrinton:
Update here: The fix to the Thinc plugin has been merged so we're just waiting for the next release. In the meantime I've been working on getting Mypy into SpaCy's CI pipeline. The first step is to ignore all current typing failures. Here's a small script to automatically add `# type: ignore[code]` directives to a project:

```python
import re
import subprocess
from pathlib import Path
from typing import List

import typer


ERROR_RE = re.compile(r"(?P<filepath>.*?):(?P<line_number>\d+): error: .* \[(?P<error_code>[\w\-]+)\]")


@typer.run
def main(
    files: List[Path] = typer.Argument(
        ...,
        exists=True,
        file_okay=True,
        dir_okay=True,
        readable=True,
    )
) -> None:
    """Execute Mypy on the given files and add `# type: ignore[error-code]` directives"""
    # Execute Mypy on the given files
    completed_process = subprocess.run(
        (
            "python",
            "-m",
            "mypy",
            "--show-error-codes",
            *[str(file) for file in files],
        ),
        capture_output=True,
    )

    # Extract structured representation of each error
    for match_groups in ERROR_RE.findall(completed_process.stdout.decode()):
        filepath, line_number, error_code = match_groups

        # Apply # type: ignore[code] directive to the file
        with open(filepath, "r") as source:
            # Read all lines
            lines = source.readlines()

        # Look up the line to modify
        line = lines[int(line_number) - 1]

        # Add directive, working around newline at end of line
        line = f"{line[:-1]}  # type: ignore[{error_code}]{line[-1]}"

        # Store updated line
        lines[int(line_number) - 1] = line

        # Rewrite file
        with open(filepath, "w") as sink:
            sink.writelines(lines)
```

It's still pretty hacky (for example, doesn't handle lines that already have a `# type: ignore` directive) and needs to be run multiple times (since Mypy output changes once initial errors have been ignored), but hopefully is a step in the right direction 🙂

I'll keep working on it until it can automatically get hide all Mypy-discovered errors. That way the "regression test" for this PR can be removal of specific `# type: ignore` directives, rather than an additional (potentially slow) Mypy invocation

svlandeg:
@connorbrinton : Thinc 8.0.9 with your bugfix for the `mypy` plugin is now released and Thinc's version bumped on `master`, so hopefully you can continue with this PR? Feel free to leave it in draft while you're working on it though!

connorbrinton:
> Thinc 8.0.9 with your bugfix for the `mypy` plugin is now released and Thinc's version bumped on `master`, so hopefully you can continue with this PR?

@svlandeg Awesome! I went ahead and created a new PR since ignoring existing Mypy errors affects a large number of lines:
https://github.com/explosion/spaCy/pull/9167

Once that PR is merged I'll rebase this one and remove the `# type: ignore` directives hiding the errors this PR will fix :+1:

svlandeg:
Closing in favour of PR https://github.com/explosion/spaCy/pull/9167 where these edits have been added to 🎉

PR #8775, opened 2021-07-21
Add basic tests for morph set Matcher operators

polm:
## Description

The Matcher recently added `IS_SUBSET` / `IS_SUPERSET` operators for Morph attributes, but there are no examples of usage or tests. This adds minimal tests. There was also some inconsistency in whether the name contained an underscore, this resolves that by changing the docs and schema to use an undescore, in line with the implementation. 

Marking this WIP because I want to add usage examples to the docs. 

### Types of change

Mostly docfix?

## Checklist

- [x] I ran the tests, and all new tests passed.


adrianeboyd:
I made the same docs/schema changes in #8784 because I thought the conflicts might get annoying otherwise. I think better examples are still a good idea?

polm:
OK, I'll close this and open a new one when the new examples are ready then.

PR #8784, opened 2021-07-21
Support list values and INTERSECTS in Matcher

adrianeboyd:
## Description

* Support list values as token attributes for set operators, not just as
pattern values.

* Add `INTERSECTS` operator.

* Fix incorrect `ISSUBSET` and `ISSUPERSET` in schema and docs.

### Types of change

Enhancement, minor docs/bug fix.

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #8795, opened 2021-07-23
Update subset/superset docs

polm:
## Description

This adds docs for the the `IS_SUBSET` matcher. It's not much but there are a few examples. Related to #8784. 

Even the small number of examples here is more than for some of the related matchers, like `IN`, but I think it's good to have more detail because these matchers are more specialized / it can be less obvious what they're doing.

### Types of change

Minor docs enhancement.

## Checklist

### DIFF
--- a/website/docs/usage/rule-based-matching.md
+++ b/website/docs/usage/rule-based-matching.md
@@ -232,14 +232,24 @@ following rich comparison attributes are available:
 >
 > # Matches tokens of length >= 10
 > pattern2 = [{"LENGTH": {">=": 10}}]
+>
+> # Match based on morph attributes
+> pattern3 = [{"MORPH": {"IS_SUBSET": ["Number=Singular", "Gender=Neuter"]}}]
+> doc = nlp(...)
+> # this will match because it's an exact match
+> doc[0].set_morph("Number=Singular|Gender=Neutral")
+> # this will not match
+> doc[0].set_morph("Number=Plural|Gender=Neutral")
+> # this will also not match because it's a superset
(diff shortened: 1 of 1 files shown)


website/docs/usage/rule-based-matching.md
```diff
@@ -232,14 +232,24 @@ following rich comparison attributes are available:
...
+>
+> # Match based on morph attributes
+> pattern3 = [{"MORPH": {"IS_SUBSET": ["Number=Singular", "Gender=Neuter"]}}]
+> doc = nlp(...)
+> # this will match because it's an exact match
+> doc[0].set_morph("Number=Singular|Gender=Neutral")
+> # this will not match
+> doc[0].set_morph("Number=Plural|Gender=Neutral")
+> # this will also not match because it's a superset
+> doc[0].set_morph("Number=Singular|Gender=Neutral|Politeness=Informal")
```
adrianeboyd, line 244: I think it would be clearer (and also fix the inconsistency around `Neutral`/`Neuter`) to use the UD inventory for this example, and also to include a proper subset to demonstrate. I don't think the `doc` / `set_morph` bits fit into this particular sidebar, so I, uh, sort of rewrote the whole thing in the suggestion, but I am less concerned about that.

```suggestion
> pattern3 = [{"MORPH": {"IS_SUBSET": ["Number=Sing", "Gender=Neut"]}}]
> # "", "Number=Sing" and "Number=Sing|Gender=Neut" will match as subsets
> # "Number=Plur|Gender=Neut" will not match
> # "Number=Sing|Gender=Neut|Polite=Infm" will not match because it's a superset
```

polm: Thank you, that's great!

PR #8807, opened 2021-07-26
Fix check for RIGHT_ATTRS in dep matcher

polm:
If a non-anchor node does not have RIGHT_ATTRS, the dep matcher throws
an E100, which says that non-anchor nodes must have LEFT_ID, REL_OP, and
RIGHT_ID. It specifically does not say RIGHT_ATTRS is required.

An empty dict for RIGHT_ATTRS is also valid, and patterns with one will be
accepted. While not normal, sometimes a REL_OP is enough to specify a
non-anchor node - maybe you just want the head of another node
unconditionally, for example.

This change just sets RIGHT_ATTRS to {} if not present. Alternatively
changing E100 to state RIGHT_ATTRS is required could also be reasonable.

Became aware of this issue while looking at https://github.com/explosion/spaCy/discussions/8803. 

### Types of change

Bug fix?

## Checklist

### DIFF
--- a/spacy/matcher/dependencymatcher.pyx
+++ b/spacy/matcher/dependencymatcher.pyx
@@ -130,7 +130,7 @@ cdef class DependencyMatcher:
                 if missing:
                     missing_txt = ", ".join(list(missing))
                     raise ValueError(Errors.E100.format(
-                        key=key,
+                        required=required_keys,
                         missing=missing_txt
                     ))
                 if (
--- a/spacy/matcher/dependencymatcher.pyx
+++ b/spacy/matcher/dependencymatcher.pyx
@@ -122,14 +122,17 @@ cdef class DependencyMatcher:
                     raise ValueError(Errors.E099.format(key=key))
                 visited_nodes[relation["RIGHT_ID"]] = True
             else:
-                if not "RIGHT_ATTRS" in relation:
-                    relation["RIGHT_ATTRS"] = {}
-                if not(
-                    "RIGHT_ID" in relation
-                    and "REL_OP" in relation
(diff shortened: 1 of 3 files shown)


svlandeg:
I personally think that `E100` should just be updated and mention `RIGHT_ATTRS` specifically. This does seem to be the original intention, as there is an explicit unit test checking for this error to be thrown when `RIGHT_ATTRS` is missing. 

Alternatively, a slightly different error msg could be given for each missing attribute, mentioning the exact one that's missing and potentially also a default value that can be used (instead of setting this implicitely in code).

polm:
Sounds good to me. I changed the error message to list missing attributes.

It would be good to suggest default values, but I wasn't sure how to add that without making the error too verbose.

spacy/matcher/dependencymatcher.pyx
```diff
@@ -122,13 +122,17 @@ cdef class DependencyMatcher:
...
+                    ("RIGHT_ID", "RIGHT_ATTRS", "REL_OP", "LEFT_ID")
+                )
+                relation_keys = set(relation.keys())
+                missing = required_keys - relation_keys
+                if missing:
+                    missing_txt = ", ".join(list(missing))
+                    raise ValueError(Errors.E100.format(
+                        key=key,
+                        missing=missing_txt
+                    ))
```
svlandeg, line 135: I like how you've generalized this. Actually, now you can take it one step further and incorporate `required_keys` as a parameter for `E100` instead of having the error code hard-code the required nodes again :-)

polm: Great point, added it!

PR #8824, opened 2021-07-28
Pass excludes when serializing vocab

adrianeboyd:
## Description

Pass excludes when serializing vocab

Additional minor bug fix:

* Deserialize vocab in `EntityLinker.from_disk`

### Types of change

Bug fix.

## Checklist


svlandeg:
Looks good to me. It's probably a bit fiddly, but can we have a regression test for 8819?

PR #8832, opened 2021-07-28
[loggers] Wandb: Added entity option 

filipematos95:
enhancement -> In the loggers: Added entity option as a config option to be able to log into a wandb team directly.

### DIFF
--- a/spacy/training/loggers.py
+++ b/spacy/training/loggers.py
@@ -104,6 +104,7 @@ def console_logger(progress_bar: bool = False):
 @registry.loggers("spacy.WandbLogger.v2")
 def wandb_logger(
     project_name: str,
+    entity_name: str,
     remove_config_values: List[str] = [],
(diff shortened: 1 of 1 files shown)


ines:
Thanks! Can the `entity_name` be `None` as the default? What does the `wandb.init` call expect here?

The way it's currently set up, `entity_name` would become a required config value, which is unideal and probably not what we want. We might also consider making this `v3` of the function, since it'd otherwise require users to run `fill-config` to include the default for `entity_name` in their configs.

svlandeg:
(maintenance note: I'll mark this as WIP until the above comments are addressed/discussed further)

svlandeg:
I'll go ahead and close this for now, as this hasn't been followed up on, and we'll probably be refactoring the logging functionality in the near future.

svlandeg:
Final note: this has been included in https://github.com/explosion/spaCy/pull/9202

PR #8840, opened 2021-07-29
Add note about SPEED in output

polm:
In #8823 it was pointed out that the `SPEED` value wasn't documented
anywhere.

Maybe there's a better place for this? 

### Types of change

Doc fix

## Checklist

### DIFF
--- a/website/docs/usage/training.md
+++ b/website/docs/usage/training.md
@@ -652,7 +652,7 @@ excluded from the logs and the score won't be weighted.
 | **Recall** (R)             | Percentage of reference annotations recovered. Should increase.                                                         |
 | **F-Score** (F)            | Harmonic mean of precision and recall. Should increase.                                                                 |
 | **UAS** / **LAS**          | Unlabeled and labeled attachment score for the dependency parser, i.e. the percentage of correct arcs. Should increase. |
-| Speed | Prediction speed in words per second (WPS). Should stay stable.                                                               |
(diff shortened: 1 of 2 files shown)


ines:
I think this is a better fit for the "metrics" section in the usage docs: https://spacy.io/usage/training#score-types

I just noticed that we already mention it there, but potentially in a confusing way, because the table calls it "words per second" instead of "speed". So maybe we should just update that?

polm:
Ah, I hadn't seen that! If we update that it should be fine.

website/docs/usage/training.md
```diff
@@ -652,7 +652,7 @@ excluded from the logs and the score won't be weighted.
 | **Recall** (R)             | Percentage of reference annotations recovered. Should increase.                                                         |
 | **F-Score** (F)            | Harmonic mean of precision and recall. Should increase.                                                                 |
 | **UAS** / **LAS**          | Unlabeled and labeled attachment score for the dependency parser, i.e. the percentage of correct arcs. Should increase. |
-| **Words per second** (WPS) | Prediction speed in words per second. Should stay stable.                                                               |
+| Speed | Prediction speed in words per second (WPS). Should stay stable.                                                               |
```
ines, line 655:
```suggestion
| **Speed** | Prediction speed in words per second (WPS). Should stay stable.                                                               |
```

Issue #5917, opened 2020-08-13
Confidence Score for NER 


gsevrodrigues:
It is important for my current project to output the confidence score for each entity we get from NER.
Doing my research, I've found out some issues regarding that. Those were particulary usefull for me: https://github.com/explosion/spaCy/issues/3996, https://github.com/explosion/spaCy/issues/881, https://github.com/explosion/spaCy/issues/831
They presented a method by our great honnibal that calculates the confidences from the spaCy NER extractions.

HOWEVER, after updating my `en_core_web_md` model to version `2.3.0`, the confidences I got for **a lot** of my NER extractions are `1.0` or `0.0`  (before updating the model, I used to see all types of scores, ranging from` 0.666...` to `0.999... `). Something seems just off... I wonder if I can trust those values. I also wonder how, internally, spaCy decides if a NER entity is acceptable or not considering its confidence (or if it taken into consideration at all).
Here is an snippet of my project where I get the confidences:

```python

nlp = spacy.load('en_core_web_md')
with nlp.disable_pipes('ner'):
    doc = nlp(text)

beams = nlp.entity.beam_parse([ doc ], beam_width = 16, beam_density = 0.0001)

entity_scores = defaultdict(float)
    for beam in beams:
        for score, ents in nlp.entity.moves.get_beam_parses(beam):
            for start, end, label in ents:
                entity_scores[(start, label)] += score
```

Am I doing anything incredibly wrong here? 

PS: I wonder if we are going to have an easy access to a confidence score from the entities in the near future? That would be great.

## Environment
* Operating System:  Windows 10 (but I'm also executing the code inside a container that runs Ubuntu and got the same issue)
* Python Version Used:  3.7.7
* spaCy Version Used: 2.3.0

fcggamou:
I believe in order to get proper beam probabilities you need to train your model with a global objective; check the answer to issue #2601. #5915 may be useful as well.

This is such a common use case, I hope it'll be easily supported in 3.0.

themrmax:
The beam-parse approach hasn't worked for me, however I have been having some success with training a separate `textcat` pipe to classify correct vs incorrect entity text (requires labeled examples of incorrectly extracted entities), and applying this serially to the output of the NER model to generate my scores. The downside of course is the extra training and classification time, not sure whether there are any optimisations here, or whether it would be worth supporting this set-up as a core feature of sPacy?

lucinvitae:
[...]

PR #8855, opened 2021-07-31
Add scores to output in spancat

polm:
This exposes the scores as an attribute on the SpanGroup. Includes a
basic test.

## Description

Scores can now be accessed in spancat output for all spans.

```
doc = nlp(...)
spans = doc.spans["spancat"] # SpanGroup
print(spans.attrs["scores"]) # list of numbers, span length as SpanGroup
```

### Types of change

Minor enhancement.

(Docs update still needed)

## Checklist

### DIFF
--- a/spacy/pipeline/spancat.py
+++ b/spacy/pipeline/spancat.py
@@ -405,7 +405,7 @@ class SpanCategorizer(TrainablePipe):
     def _make_span_group(
         self, doc: Doc, indices: Ints2d, scores: Floats2d, labels: List[str]
     ) -> SpanGroup:
-        spans = SpanGroup(doc, name=self.key)
+        spans = SpanGroup(doc, name=self.key, attrs={"scores": []})
         max_positive = self.cfg["max_positive"]
         threshold = self.cfg["threshold"]
         for i in range(indices.shape[0]):
@@ -420,4 +420,5 @@ class SpanCategorizer(TrainablePipe):
                 positives = positives[:max_positive]
             for score, start, end, label in positives:
                 spans.append(Span(doc, start, end, label=label))
+                spans.attrs["scores"].append(score)
         return spans
--- a/spacy/pipeline/spancat.py
+++ b/spacy/pipeline/spancat.py
@@ -405,20 +405,20 @@ class SpanCategorizer(TrainablePipe):
     def _make_span_group(
         self, doc: Doc, indices: Ints2d, scores: Floats2d, labels: List[str]
     ) -> SpanGroup:
-        spans = SpanGroup(doc, name=self.key, attrs={"scores": []})
+        spans = SpanGroup(doc, name=self.key)
         max_positive = self.cfg["max_positive"]
         threshold = self.cfg["threshold"]
+
+        keeps = scores >= threshold
+        ranked = (scores * -1).argsort()
+        keeps[ranked[:, max_positive:]] = False
+        spans.attrs["scores"] = scores[keeps].flatten()
         for i in range(indices.shape[0]):
             start = int(indices[i, 0])
             end = int(indices[i, 1])
-            positives = []
-            for j, score in enumerate(scores[i]):
-                if score >= threshold:
-                    positives.append((score, start, end, labels[j]))
-            positives.sort(reverse=True)
-            if max_positive:
-                positives = positives[:max_positive]
-            for score, start, end, label in positives:
-                spans.append(Span(doc, start, end, label=label))
-                spans.attrs["scores"].append(score)
+
+            for j, keep in enumerate(keeps[i]):
+                if keep:
+                    spans.append(Span(doc, start, end, label=labels[j]))
+
         return spans
--- a/spacy/pipeline/spancat.py
+++ b/spacy/pipeline/spancat.py
@@ -413,6 +413,10 @@ class SpanCategorizer(TrainablePipe):
         ranked = (scores * -1).argsort()
         keeps[ranked[:, max_positive:]] = False
         spans.attrs["scores"] = scores[keeps].flatten()
+
+        indices = self.model.ops.to_numpy(indices)
+        keeps = self.model.ops.to_numpy(keeps)
+
         for i in range(indices.shape[0]):
             start = int(indices[i, 0])
             end = int(indices[i, 1])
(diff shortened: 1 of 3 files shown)


website/docs/api/spancategorizer.md
```diff
@@ -13,6 +13,9 @@ A span categorizer consists of two parts: a [suggester function](#suggesters)
 that proposes candidate spans, which may or may not overlap, and a labeler model
 that predicts zero or more labels for each candidate.
 
+Predicted spans will be saved in a [`SpanGroup`](/api/spangroup) on the doc.
+Individual span scores can be found in `spangroup.attrs["scores"]`.
```
polm, line 17: Was not really sure where to put the docs for this so I just put it in this opening section. Is there some convention for this I can follow? 

I looked at other components, but for example the textcat docs never explicity say "the categories are saved in `doc.cats`", though the code explicitly notes that's assigned to.

mbrunecky:
I really hope this minor change gets into release ASAP.
I am only starting with the SpanCategorizer, but when requesting individual span predictions, having individual prediction confidence is essential.

website/docs/api/spancategorizer.md
```diff
@@ -13,6 +13,22 @@ A span categorizer consists of two parts: a [suggester function](#suggesters)
...
+Annotations will be saved to `Doc.spans[SPAN_KEY]` as a
+[`SpanGroup`](/api/spangroup). The scores for the spans in the `SpanGroup`
+will be saved in `SpanGroup.attrs["scores"]`. 
+
+`SPAN_KEY` defaults to `"spans"`, but can be passed as a parameter.
+
+| Location              | Value                      |
+| --------------------- | -------------------------------- |
+| `Doc.spans[SPAN_KEY]`  | The annotated spans. ~~SpanGroup~~ |
+| `Doc.spans[SPAN_KEY].attrs["scores"]` | The score for each span in the `SpanGroup`. ~~Floats1d~~  |
```
adrianeboyd, line 30:
```suggestion
Annotations will be saved to `Doc.spans[spans_key]` as a
[`SpanGroup`](/api/spangroup). The scores for the spans in the `SpanGroup`
will be saved in `SpanGroup.attrs["scores"]`. 

`spans_key` defaults to `"sc"`, but can be passed as a parameter.

| Location              | Value                      |
| --------------------- | -------------------------------- |
| `Doc.spans[spans_key]`  | The annotated spans. ~~SpanGroup~~ |
| `Doc.spans[spans_key].attrs["scores"]` | The score for each span in the `SpanGroup`. ~~Floats1d~~  |
```

website/docs/api/spancategorizer.md
```diff
@@ -13,6 +13,22 @@ A span categorizer consists of two parts: a [suggester function](#suggesters)
 that proposes candidate spans, which may or may not overlap, and a labeler model
 that predicts zero or more labels for each candidate.
 
+Predicted spans will be saved in a [`SpanGroup`](/api/spangroup) on the doc.
+Individual span scores can be found in `spangroup.attrs["scores"]`.
+
+## Annotation Format {#annotations}
```
adrianeboyd, line 19: I think "Annotation Format" is a confusing name for this. In the factory it's called "assigns", which seems less confusing with training corpus annotation formats, but I'm still not sure what a good section name is here...

svlandeg: I agree - to me "annotation" refers to the input format of the training data, while this section describes the resulting output (predictions) of the component. Hm. Something like "Predicted fields"? "Output format"? :|

mbrunecky:
I applied the #8855 spancat.py changes (+code to get confidences) and on my sample (~7000 docs, 2 entities) I am getting very 'reasonable' confidence values. I an saying 'reasonable' because it will take more analyses to find how well they work. My only complaint is that the 'default' span suggester eats memory - my 64G is not enough to go beyond span sizes 1-10 (I need 16+).
**Thanks for GOOD JOB**!
_By the way, is there some SpanCategorizer discussion forum?_

website/docs/api/spancategorizer.md
polm: I'm not sure "annotation" is the wrong word here - when components set values during training, don't we call them "annotating components"? I think this is also in line with the v3 design that an Example is just a manually created version of the hoped-for output; the format is what the user will need to provide in training anyway. 

That said "Predicted fields" or "Assigned fields" also sound fine. "Output format" also words but I guess I avoided that because it's a side effect.

svlandeg: Ah, that's actually a good point, as the `Example` objects take their annotated format on the `Docs`, the input and output format is actually the same!

adrianeboyd: It still reads way to much like "corpus annotation format" for me, which I wouldn't associate with a `Doc` even though I understand that it's technically equivalent. I feel like "predicted", "field", and "format" are a bit vague or confusing.

Maybe "Assigned Attributes"? We also say "attributes" here: https://spacy.io/api/language#analyze_pipes

polm: I like "Assigned Attributes"! Should we go with that?

svlandeg: Fine by me!

polm: With that resolved I think this is good to go?

svlandeg:
Does this still require any further docs updates?

polm:
No, the scores are just added automatically so there's no config options or anything, so I don't think the docs need any other updates.

spacy/pipeline/spancat.py
```diff
@@ -408,16 +408,17 @@ def _make_span_group(
         spans = SpanGroup(doc, name=self.key)
         max_positive = self.cfg["max_positive"]
         threshold = self.cfg["threshold"]
+
+        keeps = scores >= threshold
+        ranked = (scores * -1).argsort()
+        keeps[ranked[:, max_positive:]] = False
+        spans.attrs["scores"] = scores[keeps].flatten()
```
honnibal, line 415: While we're here, let's do:

```
indices = to_numpy(indices)
keeps = to_numpy(keeps)
```

Before the loop. We can then remove the `int()` calls around the array access. It's worth noting that if we ever have to call `int()` on individual items to copy back to CPU, we're probably doing something wrong --- we shouldn't be syncing from GPU to CPU within a loop, or it'll be really slow.

andreys42:
this task is postponed?

polm:
@andreys42 This was merged last year and has been released for a while now, as covered [in the docs](https://spacy.io/api/spancategorizer#assigned-attributes). 

If you have a usage question about it feel free to open a new Discussion.

lisabecker-ml6:
So to link back to #5917, we still can't get confidence scores for entities without training our own custom model, right?

Issue #8822, opened 2021-07-28
Spacy project pull does not pull outputs when dependencies are missing


nsorros:
## How to reproduce the behaviour
This might as well be a feature but as the behaviour is a bit unexpected I am opening this as a bug. We are using spacy projects
to track the data and our workflows. I noticed that running `spacy project pull` does not fetch all the outputs created from another
member working on the project and it comes down to some of the dependencies missing due to this check https://github.com/explosion/spaCy/blob/8547514aa4d17bd5f457d2b58d6c3efc3748520d/spacy/cli/project/pull.py#L43

The dependencies missing are data produced from the previous steps which creates a chicken and egg problem. If we want to fetch the outputs we need to run the workflow. If we run the workflow we do not need to fetch the outputs.

Is the intention to only depend on code or assets?

## Your Environment
============================== Info about spaCy ==============================

spaCy version    3.0.6                         
Location         /home/nsorros/code/overton/build/virtualenv/lib/python3.8/site-packages/spacy
Platform         Linux-5.8.0-53-generic-x86_64-with-glibc2.29
Python version   3.8.5                         
Pipelines        en_core_web_md (3.0.0)

honnibal:
> Is the intention to only depend on code or assets?

No, this does seem like a bug. Thanks for the report!

nsorros:
> > Is the intention to only depend on code or assets?
> 
> No, this does seem like it could be a bug. There could also be a problem with your project file.
> 
> The intention behind the while loop is that it should find at least one command where all of the prerequisities are present and matching, and then it will pull that command's outputs. This should give other commands the opportunity to have their dependencies met, in the next iteration of the loop.
> 
> Have a look whether there's some reason why the dependencies in your project are unsatisfiable maybe?

Thanks for the pointer @honnibal. I tracked it down to a dependency which was the folder the data was meant to be downloaded. As such the hash calculated when the data were there was different from when the data were not. The intention of the dependency was to ensure the folder is present before downloading the data. Removing this dependency unblocked the syncing of all the outputs which were dependent on the data.

I was wondering whether it would be worth adding some `logger.debug` in some of the project `push` / `pull` and `remote_storage` code for easier debugging. I already did add some to resolve this so I am happy to open a PR if this is something you want to embed in the library.

ines:
[...]

PR #8860, opened 2021-08-01
Add logger debug for project push and pull

nsorros:
## Description
As discussed in #8822 it would be helpful to have some debug information for spacy project push and pull in case the sync does not go as planned.

### Types of change
Adds debugging information for spacy project pull and push.

## Checklist

### DIFF
--- a/.github/contributors/nsorros.md
+++ b/.github/contributors/nsorros.md
@@ -1,0 +1,106 @@
+# spaCy contributor agreement
+
+This spaCy Contributor Agreement (**"SCA"**) is based on the
+[Oracle Contributor Agreement](http://www.oracle.com/technetwork/oca-405177.pdf).
+The SCA applies to any contribution that you make to any product or project
+managed by us (the **"project"**), and sets out the intellectual property rights
+you grant to us in the contributed materials. The term **"us"** shall mean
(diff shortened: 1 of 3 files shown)


nsorros:
This is a first take to add some debug information for pull and push operation of `spacy project`. Does this look like its going the right direction? @ines @honnibal?

Some points that I was unsure of
* whether to use f string formatting 
* whether to display one variable in each debug or combine them in one debug statement for example `logger.debug(f"URL: {url} for {output_path} with command hash {cmd_hash}")`
* whether to display the dependency missing which might require a small rework of the if statements to not run the for loop twice i.e. run it once `any(not dep.exists() for dep in deps)` and once again to display which deb is missing

ines:
Thanks, this looks good!

> whether to use f string formatting

Yes, we generally use f-strings for logging statements and other string formatting. (The only place where we use format strings is within `spacy.errors`, so we can have customisable error messages shared across the code base. But we don't use those within the CLI, since it'd otherwise get too messy.)

> whether to display one variable in each debug or combine them in one debug statement

IMO, the way you did it is good and I think it's nice to have one single debugging statement that includes all the relevant information for that particular step.

svlandeg:
Looks good to me as well, so I'll go ahead and merge this. Thanks for the contribution!

PR #8862, opened 2021-08-02
Improve the stop words and the tokenizer exceptions in Bulgarian language.

syrull:
## Description
I noticed that the Bulgarian language lacks of some stop words and the exceptions could be improved.

### Types of change
I have improved the readability of the stop words while adding several more. I also improved the exceptions for the tokenizer, adding categories for different types and improving the readability of abbreviations. I have also added a lot of new abbreviations.

## Checklist


spacy/lang/bg/stop_words.py
```diff
@@ -1,265 +1,74 @@
-# Source: https://github.com/Alir3z4/stop-words
```
svlandeg, line 1: Can you provide a bit more detail why you're removing this URL?

It's difficult to judge from the GH diff, because the list was reformatted, but it looks like you've made some additions & corrections to the original stop list - which means that you didn't start from scratch, correct? If you didn't start from scratch, I think we should keep the note to this link, perhaps rephrase slightly: "Based upon : ..." instead of just "Source".

syrull: Yes you are right, I forgot about this URL I will add it to the References on top of it with the appropriate explanations. 

I've added corrections & additions to the original one that is correct. I get why it is difficult to see what are the new additions, I wanted to format it like the other languages (ex. EN) so the code would be more coherent.

adrianeboyd:
We only make changes to the language defaults in the next minor version, not in patch releases, so I'm going to switch this to `develop` in preparation for v3.2. If it doesn't switch cleanly (it might show some unrelated commits temporarily), we'll handle the details.

svlandeg:
Thanks again for this contribution! We'll merge it and it will be released in spaCy 3.2 :-)

PR #8882, opened 2021-08-04
Allow passing in array vars for speedup

### LINKED ISSUE #8845: Speed issue with span.as_doc() when used repeatedly on the same document
Given a use case when we want to split a big document into much smaller chunks (e.g. for NER) its almost impossible to take advantage of current `span.as_doc()` implementation since it repeatedly calls `doc.to_array` on every single span (assuming we are working with the same doc).

Here's some code to benchmark current implementation with some naive python code to fix it:
```python
import timeit
import spacy


def _spans_to_docs(spans) -> spacy.tokens.Doc:
    """Convert all spans into actual spacy documents.

    This is the same as using spacy.tokens.Span.as_doc except faster since we
    only perform doc.to_array once. All spans must belong to the same doc.

    https://github.com/explosion/spaCy/blob/82fa81d/spacy/tokens/span.pyx#L216
    """
    array_head = None
    array = None
    for span in spans:
        if array_head is None:
            array_head = span.doc._get_array_attrs()
            array = span.doc.to_array(array_head)
        _doc = spacy.tokens.Doc(
            span.doc.vocab,
            words=[t.text for t in span],
            spaces=[bool(t.whitespace_) for t in span],
        )
        yield _doc.from_array(array_head, array[span.start : span.end])


nlp = spacy.blank('en')

for doc_size in [1, 10, 100, 1000, 10000]:
    doc = nlp('a ' * doc_size)
    spans = [doc[token.i : token.i + 1] for token in doc]
    print('doc_size', doc_size)
    print(timeit.timeit(lambda: [i.as_doc() for i in spans], number=100))
    print(timeit.timeit(lambda: list(_spans_to_docs(spans)), number=100))
```

Output:
```
doc_size 1
0.007284133999999831
0.004935189999999423
doc_size 10
0.06548612899999995
0.03388849400000016
doc_size 100
1.0578330280000001
0.3229357369999999
doc_size 1000
11.929702706
3.2446097149999993
doc_size 10000
561.8109379389999
35.17712277599992
```

I think this can be way faster with solution written in cython, maybe quick hack could be to allow passing optional pre-constructed document-array into `as_doc(doc_array=array)` and then avoid calling `array = self.doc.to_array(array_head)` if we already passed it in? Or having separated function on `doc` like `doc.spans_as_docs(spans)` which maybe could be further optimized.

polm:
This fixes #8845. Not sure about the docstring changes here...

### Types of change

Minor enhancement

## Checklist


svlandeg:
Looks good to me, but we'll need to update the docs as well?

spacy/tokens/span.pyx
```diff
@@ -213,19 +213,23 @@ cdef class Span:
         return Underscore(Underscore.span_extensions, self,
                           start=self.c.start_char, end=self.c.end_char)
 
-    def as_doc(self, *, bint copy_user_data=False):
+    def as_doc(self, *, bint copy_user_data=False, array_head=None, array=None):
         """Create a `Doc` object with a copy of the `Span`'s data.
 
         copy_user_data (bool): Whether or not to copy the original doc's user data.
+        array_head: `Doc` array attrs, can be passed in to speed up computation.
+        array: `Doc` as array, can be passed in to speed up computation.
```
svlandeg, line 221:
```suggestion
        array_head (tuple): `Doc` array attrs, can be passed in to speed up computation.
        array (ndarray): `Doc` as array, can be passed in to speed up computation.
```

PR #8885, opened 2021-08-04
No output info message in train

kabirkhan:
Add info message when no output directory was provided in `spacy train` CLI

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -43,9 +43,12 @@ def train_cli(
     # Make sure all files and paths exists if they are needed
     if not config_path or (str(config_path) != "-" and not config_path.exists()):
         msg.fail("Config file not found", config_path, exits=1)
-    if output_path is not None and not output_path.exists():
-        output_path.mkdir(parents=True)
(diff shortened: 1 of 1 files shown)


spacy/cli/train.py
```diff
@@ -43,9 +43,12 @@ def train_cli(
...
         msg.fail("Config file not found", config_path, exits=1)
-    if output_path is not None and not output_path.exists():
-        output_path.mkdir(parents=True)
-        msg.good(f"Created output directory: {output_path}")
+    if output_path is not None:
+        if not output_path.exists():
+            output_path.mkdir(parents=True)
+            msg.good(f"Created output directory: {output_path}")
+    else:
+        msg.info("No output directory provided")
```
svlandeg, line 51: Looks good :-)

Two small nitpicks: we might as well also log the output dir if it exists so we cover all cases, and personally I find it easier to read if the `if` has the non-negated version (i.e. switch the outer `if` and `else` around)

kabirkhan: ok, was making minimal changes but agreed :)

PR #8900, opened 2021-08-07
Update Tigrinya ትግርኛ language support

fgaim:
Some enhancements to the Tigrinya language support

## Description
This PR expands the stop-words, fixes numerals & ordinals, and adds missing punctuation.

### Types of change
Enhancements to Tigrinya language support

## Checklist


adrianeboyd:
Thanks for the contribution! I have one small comment, but overall this looks good.

We only make changes to the language defaults in the next minor version, not in patch releases, so I'm going to switch this to `develop` in preparation for v3.2. If it doesn't switch cleanly (it might show some unrelated commits temporarily), we'll handle the details.

spacy/lang/ti/lex_attrs.py
```diff
@@ -11,67 +11,47 @@
...
+    "ዓሰርተ",
+    "ዓሰርተ ሓደ",
+    "ዓሰርተ ክልተ",
+    "ዓሰርተ ሰለስተ",
+    "ዓሰርተ ኣርባዕተ",
+    "ዓሰርተ ሓሙሽተ",
+    "ዓሰርተ ሽድሽተ",
+    "ዓሰርተ ሸውዓተ",
+    "ዓሰርተ ሽሞንተ",
+    "ዓሰርተ ትሽዓተ",
```
adrianeboyd, line 23: I think that these entries were a mistake in this list because the tokenizer isn't going to produce tokens with spaces in them, so these are never going to match. It looks like each of the individual tokens is already in the list, so I don't think we need the additional entries with space at all ... or am I misunderstanding the details here?

fgaim: That's a good point, @adrianeboyd. I wondered about that myself but again since they existed in the previous version, I left that in tact. And the tokens do exist separately in the list, so unless there is a criteria where we need to cover all numbers sequentially, we can remove of the double word entries. I will push a commit with the changes.

fgaim: Done.

BTW, similar issues exist with Amharic [am] as well. We can have those changes as part of this PR if you want.

adrianeboyd: If you don't mind, I think a separate PR would be cleaner. (Create a new branch first! It's better not to create PRs directly from `master` because if you do and then you update or modigy `master` in your repo, it will affect the PR here, too.)

svlandeg:
Thanks for the contribution! This will be available from spaCy 3.2 onwards :-)

yosiasz:
is this now available?

PR #8909, opened 2021-08-09
Add support for fasttext-bloom hash-only vectors

adrianeboyd:
## Description

Overview:

* Extend `Vectors` to have two modes: `default` and `ngram`
  * `default` is the default mode and equivalent to the current `Vectors`
  * `ngram` supports the hash-only ngram tables from `fasttext-bloom`
* Extend `spacy.StaticVectors.v2` to handle both modes with no changes for `default` vectors
* Extend `spacy init vectors` to support ngram tables

The `ngram` mode **only** supports vector tables produced by the [`fasttext-bloom` fork of fastText](https://github.com/adrianeboyd/fastText/tree/feature/bloom), which adds an option to represent all vectors using only the ngram buckets table and which uses the exact same ngram generation algorithm and hash function (`MurmurHash3_x64_128`). `fasttext-bloom` produces an additional `.hashvec` table, which can be loaded by `spacy init vectors --fasttext-bloom-vectors`.

Implementation details:

* `Vectors` now includes the `StringStore` as `Vectors.strings` so that the API can stay consistent for both `default` (which can look up from `str` or `int`) and `ngram` (which requires `str` to calculate the ngrams).

* In ngram mode `Vectors` uses a default `Vectors` object as a cache since the ngram vectors lookups are relatively expensive.

  * The default cache size is the same size as the provided ngram vector table.

  * Once the cache is full, no more entries are added. The user is responsible for managing the cache in cases where the initial documents are not representative of the texts.

  * The cache can be resized by setting `Vectors.ngram_cache_size` or cleared with `vectors._ngram_cache.clear()`.

* The API ends up a bit split between methods for `default` and for `ngram`, so functions that only make sense for `default` or `ngram` include warnings with custom messages suggesting alternatives where possible.

* `Vocab.vectors` becomes a property so that the string stores can be synced when assigning vectors to a vocab.

* `Vectors` serializes its own config settings as `vectors.cfg`.

* The `Vectors` serialization methods have added support for `exclude` so that the `Vocab` can exclude the `Vectors` strings while serializing.

Removed:

* The `minn` and `maxn` options and related code from `Vocab.get_vector`, which does not work in a meaningful way for default vector tables.

* The unused `GlobalRegistry` in `Vectors`.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/cli/init_pipeline.py
+++ b/spacy/cli/init_pipeline.py
@@ -23,6 +23,7 @@ def init_vectors_cli(
     name: Optional[str] = Opt(None, "--name", "-n", help="Optional name for the word vectors, e.g. en_core_web_lg.vectors"),
     verbose: bool = Opt(False, "--verbose", "-V", "-VV", help="Display more information for debugging purposes"),
     jsonl_loc: Optional[Path] = Opt(None, "--lexemes-jsonl", "-j", help="Location of JSONL-formatted attributes file", hidden=True),
+    fasttext_bloom_vectors: bool = Opt(False, "--fasttext-bloom-vectors", "-F", help="Vectors in fastText Bloom hash ngram format"),
     # fmt: on
 ):
     """Convert word vectors for use with spaCy. Will export an nlp object that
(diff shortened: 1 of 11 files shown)


adrianeboyd:
Demo projects: https://github.com/adrianeboyd/fasttext-bloom-demo-projects

adrianeboyd:
Additional comments:

* The `fasttext` fork/tool needs a better name.

* `ngram` is not a particularly good name for the spacy vectors mode because you can do it without char-ngrams in fasttext if you want. `hashonly` (as used in the `fasttext-bloom` options) is more accurate but clunkier, and `hash` is bad as a python built-in. I guess we could call it `hashembed`, but I'm worried that could get confusing.

* It's separate from this PR, but we'd need a size name for models trained with these vectors that is reminiscent of `md`.

* If we publish this, we should publish a blog post with the content from https://support.prodi.gy/t/can-you-explain-how-exactly-hashembed-works/564 with additional info about how this works in the fasttext fork.

PR #8910, opened 2021-08-09
updated unv json for new book

DuyguA:
## Description
Information about my new book added to spaCy universe

## Checklist


ines:
Thanks for adding this! We had this on our internal list for "things to add to the universe" but didn't get around to it yet, so this is great 💕

Issue #5280, opened 2020-04-09
Pretraining with option --n-save-every still saves all models


chopeen:
I am running the following command:
```
!spacy pretrain $FILE_RF_SENTENCES en_core_sci_lg $DIR_MODELS_RF_SENT \
    --use-vectors --n-save-every 5
```
In order to use less space, I specified the option `--n-save-every` to save a model every X batches.

However, all models are still saved, with additional `.temp.bin` files:
```
$ ls -al /kaggle/working/models/tok2vec_rf_sent_sci
[snip]
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model110.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model110.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model111.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model111.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model112.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model112.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model113.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model113.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model114.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model114.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model115.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model115.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model116.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model116.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model117.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model117.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model118.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model118.temp.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model119.bin
-rw-r--r-- 1 root root 3889626 Apr  9 08:49 model119.temp.bin
[snip]
```

## Info about spaCy

* **spaCy version:** 2.2.4
[...]

PR #8912, opened 2021-08-09
Add new parameter for saving every n epoch in pretraining

thomashacker:
## Description

Issue https://github.com/explosion/spaCy/issues/5280#

Added new config parameter `n_save_epoch` for pretraining, which saves the model every n iteration + first & last.

### Types of change

New feature

## Checklist


svlandeg:
I'm happy with the name for this new feature, but unhappy with the name of the existing `n_save_every` option, which causes [confusion](https://github.com/explosion/spaCy/issues/5280#issuecomment-611971233). Not sure we want to go back and change that now though, it would be breaking.

PR #8948, opened 2021-08-13
Auto-detect package dependencies in spacy package

ines:
## Description

Previously, when packaging a pipeline that depends on components or functions from packages that aren't installed with spaCy by default (e.g. `spacy-transformers` or a third-party plugin), the user would have to take care of adding those packages to the `"requirements"` in their `meta.json` or `nlp.meta["requirements"]` manually. This is pretty inconvenient and unintuitive, because it's an easy thing to forget and only really comes up once you install the packaged pipeline in a _different_ environment that doesn't have the dependencies installed. And it really broke the nice workflow of "just install the pipeline package and get everything you need".

This PR introduces a helper in `spacy package` that traverses the config, finds all function references, uses `importlib_metadata` to find the packages those functions are in and their versions and generates an appropriate version range to use in the requirements. Those will automatically become requirements of the generated Python package.

If the function is from one of our packages that spaCy installs, it's excluded. A requirement is also only added if it's not already in the `meta["requirements"]`. This allows users to customise the version ranges if they want to, or add more packages manually (e.g. if custom `--code` is provided that depends on certain libraries – spaCy can't easily know that without actually parsing the Python code).

There are still edge cases we may not be covering, but I tested it for the most common use cases (e.g. `spacy-transformers`) and checked that functions provided with `--code` (that don't have a module and are copied into the package) are skipped. If we're too nervous about this feature, we could introduce a CLI flag that skips the requirements discovery, in case users run into a problem.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/compat.py
+++ b/spacy/compat.py
@@ -27,6 +27,13 @@ try:  # Python 3.8+
 except ImportError:
     from typing_extensions import Literal  # noqa: F401
 
+# Important note: The importlib_metadata "backport" includes functionality
+# that's not part of the built-in importlib.metadata. We should treat this
+# import like the built-in and only use what's available there.
+try:  # Python 3.8+
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    from catalogue import _importlib_metadata as importlib_metadata  # noqa: F401
 
 from thinc.api import Optimizer  # noqa: F401
 
--- a/spacy/compat.py
+++ b/spacy/compat.py
@@ -28,12 +28,6 @@ except ImportError:
     from typing_extensions import Literal  # noqa: F401
 
 
-try:  # Python 3.8+
-    import importlib.metadata as importlib_metadata
-except ImportError:
-    from catalogue import _importlib_metadata as importlib_metadata  # noqa: F401
-
-
 from thinc.api import Optimizer  # noqa: F401
 
 pickle = pickle
--- a/spacy/compat.py
+++ b/spacy/compat.py
@@ -27,6 +27,13 @@ try:  # Python 3.8+
 except ImportError:
     from typing_extensions import Literal  # noqa: F401
 
+
+try:  # Python 3.8+
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    from catalogue import _importlib_metadata as importlib_metadata  # noqa: F401
+
+
 from thinc.api import Optimizer  # noqa: F401
 
 pickle = pickle
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -1553,6 +1553,10 @@ def to_ternary_int(val) -> int:
         return -1
 
 
+# The following implementation of packages_distributions() is adapted from
+# importlib_metadata, which is distributed under the Apache 2.0 License.
+# Copyright (c) 2017-2019 Jason R. Coombs, Barry Warsaw
+# See licenses/3rd_party_licenses.txt
 def packages_distributions() -> Dict[str, List[str]]:
     """Return a mapping of top-level packages to their distributions. We're
     inlining this helper from the importlib_metadata "backport" here, since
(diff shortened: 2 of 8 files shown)


spacy/compat.py
```diff
@@ -27,6 +27,13 @@
 except ImportError:
     from typing_extensions import Literal  # noqa: F401
 
+
+try:  # Python 3.8+
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    from catalogue import _importlib_metadata as importlib_metadata  # noqa: F401
+
+
```
adrianeboyd, line 36: I don't think the kind of compat is a good idea because `packages_distributions` is backport from a pretty recent version of `importlib_metadata` and I don't think it's necessarily in `importlib.metadata`, if at all. This definitely needs some cross-python tests.

ines: Oh, I was under the impression that `packages_distributions` was always part of `importlib.metadata` in Python 3.8 and above, but only just recently added to the backport.

I'll add a test that just checks that the functions runs and we don't hit any weird errors. (Of course, in our CI, I expect the list of third-party dependencies to be empty, and it'll be difficult to actually test the detection in our CI.)

ines: Okay, turns out there's actually no `packages_distributions` in `importlib.metadata`... so I guess the solution is to just always import this function from our fork?

ines:
Ugh, these failures are weird? I guess the alternative would be to inline `packages_distributions` as a util function in spaCy (because I think it only calls into the default `importlib.metadata` helpers that are included in both the built-in and backport)?

adrianeboyd:
It's because the `catalogue` setup does assume it's only used as a backport:

https://github.com/explosion/catalogue/blob/master/requirements.txt

Maybe it would just be easier to have it as a util function elsewhere. Or we could change the `catalogue` requirements, but that's probably not a great solution if it's not really necessary. It was just clean/easy to add this backport from the upstream `importlib_metadata`, but there's no reason to necessarily do it that way.

ines:
Okay, let's see if this passes with the inlined function.

I do think moving `importlib_metadata` to `spacy.compat` could be good, though, because it clearly signals that "this should always be the same package". What the packages expose may be different depending on your Python version (just like it's the case for various built-ins), but we should only ever be using functions that are available everywhere.

svlandeg (commented):
Looks good! I wonder whether we should move this to `develop` (3.2) as technically this change can break/change existing workflows?

spacy/util.py
```diff
@@ -1549,3 +1551,15 @@ def to_ternary_int(val) -> int:
...
+def packages_distributions() -> Dict[str, List[str]]:
+    """Return a mapping of top-level packages to their distributions. We're
+    inlining this helper from the importlib_metadata "backport" here, since
+    it's not available in the builtin importlib.metadata.
+    """
+    pkg_to_dist = defaultdict(list)
+    for dist in importlib_metadata.distributions():
+        for pkg in (dist.read_text("top_level.txt") or "").split():
+            pkg_to_dist[pkg].append(dist.metadata["Name"])
+    return dict(pkg_to_dist)
```
adrianeboyd, line 1565: We should be careful about attribution/licenses when including it this way. I think we should an attribution here and add the license to the 3rd party licenses.

ines:
@svlandeg 

> I wonder whether we should move this to `develop` (3.2) as technically this change can break/change existing workflows?

Hmm yes, I see your point. It'd be great to get this into the next patch release, though, as it currently causes a bunch of problems for people packaging their models and uploading them somewhere (e.g. the HF hub). In theory, the fix isn't _that_ breaking because it'll only ever add stuff that should have been there in the first place, and it'll respect existing `meta.json` settings... but of course, that can always lead to unexpected side-effects 🤔

PR #8954, opened 2021-08-13
Docs for new spacy-trf architectures

svlandeg:
This PR should not be merged until `spacy-transformers` 1.1.0 is out. Keeping in draft until then.

## Description
* docs for [spacy-transformers#268](https://github.com/explosion/spacy-transformers/pull/268) & [spacy-transformers#287](https://github.com/explosion/spacy-transformers/pull/287)
* bump `spacy-transformers` to 1.1.0

This PR also updates the quickstart with the v2 model. I guess this can go wrong if people upgrade `spacy` but not `spacy-transformers` - they'd end up with an architecture that won't be available. Not sure how to avoid that though, it also seems unfortunate to keep recommending an older architecture in the quickstart.

### Types of change
docs update

## Checklist


adrianeboyd:
Should `spacy-transformers.Tok2VecTransformer.v1` get added to the legacy page?

svlandeg:
@adrianeboyd : we discussed this on Slack and I thought the decision was not to move them to legacy as the backoff only handles `spacy.`? Anyway we probably do want to find a solution for this if we accumulate too many legacy architectures here.

adrianeboyd:
I mean the legacy page in the docs, not `spacy-legacy`.

svlandeg:
Yea, hm. I thought about that, but decided not to because the title reads "Archived implementations available through spacy-legacy" :/

adrianeboyd:
Hmm, I didn't realize that. Maybe then leave it documented on the transformer page? It seems bad that the v1 documentation is completely gone.

adrianeboyd:
There is a similar problem for the `ModelOutput` changes, how to document both versions of the `TransformerData` and `FullTransformerBatch` dataclasses.

svlandeg:
The v1 docs aren't completely gone though, there are two Accordion sections which state they are the same, minus the `transformer_config` argument, so I guess I was hoping that was sufficient :p

adrianeboyd:
Ah, I see. As long you as can actually find and see that using the site search (which honestly isn't working very well for me these days), then I think it's fine.

website/docs/api/architectures.md
```diff
@@ -366,12 +369,27 @@ transformer weights across your pipeline. For a layer that's configured for use
...
+| `tokenizer_config`   | Tokenizer settings passed to [`transformers.AutoTokenizer`](https://huggingface.co/transformers/model_doc/auto.html#transformers.AutoTokenizer). ~~Dict[str, Any]~~                                                                                   |
+| `transformer_config` | Settings to pass to the transformers forward pass. ~~Dict[str, Any]~~                                                                                                                                                                                 |
+| `mixed_precision`    | Replace whitelisted ops by half-precision counterparts. Speeds up training and prediction on GPUs with [Tensor Cores](https://developer.nvidia.com/tensor-cores) and reduces GPU memory use. ~~bool~~                                                 |
+| `grad_scaler_config` | Configuration to pass to `thinc.api.PyTorchGradScaler` during training when `mixed_precision` is enabled. ~~Dict[str, Any]~~                                                                                                                          |
```
danieldk, line 379: We would need to add `PyTorchGradScaler` to the thinc docs to be able to link this. Not too sure what to do with this, since we aren't sure if `PyTorchGradScaler` is the final design.

svlandeg: I would add a note that this is a `beta` feature, and maybe hold off on documenting in Thinc until it's finalized?

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -32,9 +32,10 @@ batch_size = {{ 128 if hardware == "gpu" else 1000 }}
 factory = "transformer"
 
 [components.transformer.model]
-@architectures = "spacy-transformers.TransformerModel.v1"
+@architectures = "spacy-transformers.TransformerModel.v2"
 name = "{{ transformer["name"] }}"
 tokenizer_config = {"use_fast": true}
+transformer_config = {"output_attentions": false}
```
svlandeg, line 38: @danieldk : should we bump to v3 in the quickstart too? And add the default arguments for `mixed_precision` and `grad_scaler_config`? I know the feature is experimental and the thinc side of things might change, but I suppose these two arguments will remain roughly the same?

website/docs/api/architectures.md
```diff
@@ -366,12 +369,31 @@ transformer weights across your pipeline. For a layer that's configured for use
...
+<Infobox title="Mixed precision support" variant="warning">
+Mixed-precision support is currently an experimental feature.
+</Infobox>
+
+<Accordion title="Previous versions of spacy-transformers.TransformerModel" spaced>
+
+* The `transformer_config` argument was added in
+`spacy-transformers.TransformerModel.v2`.
+* The `mixed_precision` and `grad_scaler_config` arguments were added in
+`spacy-transformers.TransformerModel.v3`
```
svlandeg, line 392:
```suggestion
`spacy-transformers.TransformerModel.v3`.
```

website/docs/api/architectures.md
```diff
@@ -421,13 +446,32 @@ Use a transformer as a [`Tok2Vec`](/api/tok2vec) layer directly. This does
...
+<Infobox title="Mixed precision support" variant="warning">
+Mixed-precision support is currently an experimental feature.
+</Infobox>
+
+<Accordion title="Previous versions of spacy-transformers.Tok2VecTransformer" spaced>
+
+* The `transformer_config` argument was added in
+`spacy-transformers.Tok2VecTransformer.v2`.
+* The `mixed_precision` and `grad_scaler_config` arguments were added in
+`spacy-transformers.Tok2VecTransformer.v3`
```
svlandeg, line 470:
```suggestion
`spacy-transformers.Tok2VecTransformer.v3`.
```

website/docs/api/transformer.md
```diff
@@ -83,9 +83,10 @@ https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/p
 > # Construction via add_pipe with custom config
 > config = {
 >     "model": {
->         "@architectures": "spacy-transformers.TransformerModel.v1",
+>         "@architectures": "spacy-transformers.TransformerModel.v2",
```
svlandeg, line 86: @danieldk : update this one too, I suppose

website/docs/usage/embeddings-transformers.md
```diff
@@ -351,9 +351,10 @@ factory = "transformer"
 max_batch_items = 4096
 
 [components.transformer.model]
-@architectures = "spacy-transformers.TransformerModel.v1"
+@architectures = "spacy-transformers.TransformerModel.v2"
```
svlandeg, line 354: And this one (and again at L370)

spacy/cli/templates/quickstart_training.jinja
danieldk: I have updated the `v2` references to `v2`. I am not sure about the quickstart. People are probably still using Pascal-generation GPUs like the 1080Ti for finetuning transformers, which support FP16 but do not have Tensor Cores, and do usually not benefit from mixed-precision training.

Maybe we should add the two options to the example, but suggest `mixed_precision = false` by default?

svlandeg: Yea I agree it's a good idea to put the default at `False`!

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -32,9 +32,10 @@ batch_size = {{ 128 if hardware == "gpu" else 1000 }}
 factory = "transformer"
 
 [components.transformer.model]
-@architectures = "spacy-transformers.TransformerModel.v1"
+@architectures = "spacy-transformers.TransformerModel.v3"
 name = "{{ transformer["name"] }}"
 tokenizer_config = {"use_fast": true}
+transformer_config = {"output_attentions": false}
```
adrianeboyd, line 38: I would not add this to the quickstart. It is empty by default (it will be filled in) and that is fine.

website/docs/api/transformer.md
```diff
@@ -83,9 +83,12 @@ https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/p
 > # Construction via add_pipe with custom config
 > config = {
 >     "model": {
->         "@architectures": "spacy-transformers.TransformerModel.v1",
+>         "@architectures": "spacy-transformers.TransformerModel.v3",
 >         "name": "bert-base-uncased",
->         "tokenizer_config": {"use_fast": True}
+>         "tokenizer_config": {"use_fast": True},
+>         "transformer_config": {"output_attentions": False},
```
adrianeboyd, line 89: For a custom config, I'd add `True` as a example.

website/docs/usage/embeddings-transformers.md
```diff
@@ -351,9 +351,12 @@ factory = "transformer"
 max_batch_items = 4096
 
 [components.transformer.model]
-@architectures = "spacy-transformers.TransformerModel.v1"
+@architectures = "spacy-transformers.TransformerModel.v3"
 name = "bert-base-cased"
 tokenizer_config = {"use_fast": true}
+transformer_config = {"output_attentions": false}
```
adrianeboyd, line 357:
```suggestion
transformer_config = {"output_attentions": true}
```

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -32,9 +32,12 @@ batch_size = {{ 128 if hardware == "gpu" else 1000 }}
...
+mixed_precision = false
+grad_scaler_config = {"init_scale": 32768}
```
adrianeboyd, line 40: What are the defaults here? We don't want to add/maintain things here that have good defaults. (And if `false` isn't the default, we need to do something about that.)

setup.cfg
```diff
@@ -70,7 +70,7 @@ console_scripts =
 lookups =
     spacy_lookups_data>=1.0.2,<1.1.0
 transformers =
-    spacy_transformers>=1.0.1,<1.1.0
+    spacy_transformers>=1.1.0,<1.2.0
```
adrianeboyd, line 73: I don't think we need to require this? People may have reasons to want to continue using v1.0 models and it's not strictly necessary?

website/docs/api/architectures.md
```diff
@@ -403,16 +425,19 @@ a single token vector given zero or more wordpiece vectors.
...
 
 > #### Example Config
 >
 > ```ini
 > [model]
-> @architectures = "spacy-transformers.Tok2VecTransformer.v1"
+> @architectures = "spacy-transformers.Tok2VecTransformer.v3"
 > name = "albert-base-v2"
 > tokenizer_config = {"use_fast": false}
+> transformer_config = {"output_attentions": false}
```
adrianeboyd, line 437:
```suggestion
> transformer_config = {}
```

website/docs/api/architectures.md
```diff
@@ -332,15 +332,18 @@ for details and system requirements.
...
 
 > #### Example Config
 >
 > ```ini
 > [model]
-> @architectures = "spacy-transformers.TransformerModel.v1"
+> @architectures = "spacy-transformers.TransformerModel.v3"
 > name = "roberta-base"
 > tokenizer_config = {"use_fast": true}
+> transformer_config = {"output_attentions": false}
```
adrianeboyd, line 344:
```suggestion
> transformer_config = {}
```

setup.cfg
svlandeg: Technically we don't need to require it no. I do think the IO was pretty broken before 1.1.0 so it would be better if people upgraded, but we don't have to enforce it, it'll happen naturally.

spacy/cli/templates/quickstart_training.jinja
svlandeg: Ah, sure. I don't have a strong opinion on that.

spacy/cli/templates/quickstart_training.jinja
svlandeg: The defaults are here: https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/architectures.py#L196, i.e. `false` and empty dict. You're right Adriane it's better to keep the quickstart shorter.

website/docs/api/transformer.md
svlandeg: Good point!

website/docs/usage/embeddings-transformers.md
svlandeg: Wait here we need to ensure that the config mimicks the Python code above it. So either add the same non-default values to the Python code, or update the config to use the defaults. Either way, it's probably best to be explicit in both versions, so they can be matched visually easily by the reader of the docs.

adrianeboyd: Ah, I think I got the sequence mixed up with the python vs. config blocks. Whatever you do, it doesn't make sense to say `{"output_attentions": false}` everywhere. Just leave the default as it is and use `{}` or nothing instead.

adrianeboyd, line 357:
```suggestion
```

website/docs/usage/embeddings-transformers.md
```diff
@@ -351,9 +351,12 @@ factory = "transformer"
...
+mixed_precision = true
+grad_scaler_config = {"init_scale": 32768}
```
adrianeboyd, line 359: Sorry, a better suggestion would be:

```suggestion
```

PR #8972, opened 2021-08-16
Update custom tokenizer APIs and pickling

adrianeboyd:
## Description

* Fix incorrect pickling of Japanese and Korean pipelines, which led to the entire pipeline being reset if pickled

* Enable pickling of Vietnamese tokenizer

* Update tokenizer APIs for Chinese, Japanese, Korean, Thai, and Vietnamese so that only the `Vocab` is required for initialization

### Types of change

Bug fix, mainly

## Checklist


adrianeboyd:
(None of the relevant tests actually get run in the CI.)

PR #8975, opened 2021-08-16
Fix making span_group

svlandeg:
## Description
PR https://github.com/explosion/spaCy/pull/8855 introduced a bug that would sometime lead to indexing errors, when there is only one candidate entity:
```
  File "...\spacy\spacy\pipeline\spancat.py", line 421, in _make_span_group
    keeps[ranked[:, max_positive:]] = False
IndexError: index 1 is out of bounds for axis 0 with size 1
```
Upon inspection, it became clear that this line
``` 
keeps[ranked[:, max_positive:]] = False
```
wasn't doing exactly what we wanted (put the correct indices to `False`) as examplified by the (previously failing) unit tests of this PR.

I rewrote the line into a longer `for`-loop to ensure we have the correct behaviour. If anyone knows the correct indexing voodoo to make this shorter ánd keep the tests green, I'm happy to make that adjustement ;-) 

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tests/pipeline/test_spancat.py
+++ b/spacy/tests/pipeline/test_spancat.py
@@ -71,6 +71,57 @@ def test_explicit_labels():
     assert spancat.labels == ("PERSON", "LOC")
 
 
+@pytest.mark.parametrize(
+    "max_positive,nr_results", [(None, 4), (1, 2), (2, 3), (3, 4), (4, 4)]
+)
+def test_make_spangroup(max_positive, nr_results):
+    import numpy
+
+    fix_random_seed(0)
+    nlp = Language()
+    spancat = nlp.add_pipe(
+        "spancat",
+        config={"spans_key": SPAN_KEY, "threshold": 0.5, "max_positive": max_positive},
+    )
+    doc = nlp.make_doc("Greater London")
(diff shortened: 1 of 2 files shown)


polm:
Thanks for fixing this! 

I think the correct index magic way to do this is [numpy.put_along_axis](), but there's no equivalent in cupy (you'd have to do some index conversion), so this should be fine as-is.

polm (approved):
Looks good!

spacy/tests/pipeline/test_spancat.py
```diff
@@ -71,6 +71,57 @@ def test_explicit_labels():
     assert spancat.labels == ("PERSON", "LOC")
 
 
+@pytest.mark.parametrize(
+    "max_positive,nr_results", [(None, 4), (1, 2), (2, 3), (3, 4), (4, 4)]
+)
+def test_make_spangroup(max_positive, nr_results):
+    import numpy
+
```
adrianeboyd, line 79: Is there a particular reason for doing the imports like this? Can we make it more like the rest of the test suite with the imports at the top like the other numpy testing utils?

svlandeg: Definitely!

PR #8976, opened 2021-08-17
Move Japanese Tokenizer extra info to MorphAnalysis

polm:
## Description

This moves most of the extra Japanese tokenizer info that was in `user_data` into a `Token.morph`. 

It also adds the normalized form SudachiPy provides, which previously wasn't exposed at all. 

The one thing that's not moved into `Token.morph` is subtokens. This is a list of full token information, but it's only populated for some words with non-default tokenizer settings. Since it's just a list of `None` with the default tokenizer settings, in that case I changed it to not set `user_data` at all. With non-default settings the behavior regarding subtokens is unchanged.

As a result of this, with default tokenizer settings the Japanese tokenizer no longer modifies `user_data`. 

The docs didn't previously mention any of the `user_data` info, so I'm adding a short section on the current data to the model page.

Since this is an API change I guess it should be held back until v3.2.

### Types of change

API Cleanup?

## Checklist


polm:
Urg, I originally based this off master and changing it to develop made the history weird. Let me try to fix it...

PR #8977, opened 2021-08-17
Moving Japanese tokenizer extra info to Token.morph

polm:
## Description

This moves most of the extra Japanese tokenizer info that was in `user_data` into a `Token.morph`. 

It also adds the normalized form SudachiPy provides, which previously wasn't exposed at all. 

The one thing that's not moved into `Token.morph` is subtokens. This is a list of full token information, but it's only populated for some words with non-default tokenizer settings. Since it's just a list of `None` with the default tokenizer settings, in that case I changed it to not set `user_data` at all. With non-default settings the behavior regarding subtokens is unchanged.

As a result of this, with default tokenizer settings the Japanese tokenizer no longer modifies `user_data`. 

The docs didn't previously mention any of the `user_data` info, so I'm adding a short section on the current data to the model page.

Since this is an API change I guess it should be held back until v3.2.

### Types of change

API cleanup?, minor enhancement

## Checklist

### DIFF
--- a/spacy/lang/ja/__init__.py
+++ b/spacy/lang/ja/__init__.py
@@ -13,7 +13,7 @@ from ...errors import Errors
 from ...language import Language
 from ...scorer import Scorer
 from ...symbols import POS
-from ...tokens import Doc
+from ...tokens import Doc, MorphAnalysis
 from ...training import validate_examples
 from ...util import DummyTokenizer, registry, load_config_from_str
 from ... import util
@@ -41,6 +41,8 @@ class JapaneseTokenizer(DummyTokenizer):
         self.vocab = nlp.vocab
         self.split_mode = split_mode
         self.tokenizer = try_sudachi_import(self.split_mode)
+        # if we're using split mode A we don't need subtokens
+        self.need_subtokens = not (split_mode is None or split_mode == "A")
 
     def __call__(self, text: str) -> Doc:
         # convert sudachipy.morpheme.Morpheme to DetailedToken and merge continuous spaces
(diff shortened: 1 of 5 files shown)


adrianeboyd:
I'd have to search a bit for the thread, but I think I remember a discussion about adding the morphologizer to the default Japanese pipelines, which would cause problems for this. The morphologizer currently clobbers and doesn't have an option to not overwrite or to append, which I suppose we could add, but for now this may not interact well.

polm:
Ah, we did discuss that before, though it didn't come up in a search so maybe it was off Github. 

If it's not too complicated (and it doesn't seem like it would be) I think changing the Morphologizer to allow overwriting would be a good idea. I suspect other languages with tokenizer issues will also have metadata like this, so the change would likely not just benefit Japanese.

adrianeboyd:
Here's the discussion: #8677

I don't think overwriting is what you want, though, you would want it to add the features in addition to the ones from the tokenizer? That wasn't an option I'd considered in the past, but it's definitely sensible for `morph`.

adrianeboyd:
In general, all the pipeline components need to have at least an `overwrite` option because right now it's really inconsistent. Some overwrite, some don't, some have a setting.

polm:
Yeah, I guess it's not overwrite, it's more like... Keep while updating? I guess we could call it "preserve" or "reuse" or something.

Specifically here there are some attributes that can just be left untouched while the Morphologizer updates other things.

adrianeboyd:
This now depends on #9050.

adrianeboyd:
And I think we should add tests to `spacy-models` for this.

polm:
Small update here - there's a "normalized form" that I was putting in the morph attributes instead of `token.norm_`. I think I just forgot about the token field, so I have changed this value to go there instead. 

This can be a little more drastic than the English norm sometimes - an example from the SudachiPy README is that the norm of "SUMMER" is "サマー" - but it is broadly useful.

polm:
Hm, actually... The normalized form is not just normalizing orthography, it also lemmatizes. 

| raw | first token | lemma | norm |
| やった | やっ | やる | 遣る |
| みた | み | みる | 見る |
| 新しく | 新し | 新しい | 新しい |
| 綜合 | 綜合 | 綜合 | 総合 |

I think this is still OK but it is more than I was expecting. The SudachiPy README only covers cases like the last one, with orthographic variation of non-inflecting words.

svlandeg:
Is this ready to merge? (it seems so?)

adrianeboyd:
The tests don't actually test it, so the better test will be with the `spacy-models` tests. It seems fine for me locally.

PR #8980, opened 2021-08-17
Warn and document spangroup.doc weakref

svlandeg:
## Description
`SpanGroup.doc` uses a `weakref.ref` to refer to the `Doc` without causing circular references. Unfortunately this means that something like this will fail:
```
all_spans = [doc.spans for doc in nlp.pipe(texts)]
print(all_spans)
```
because the `SpanGroup` objects within `all_spans` will refer to dead referents.

In this PR, we raise a custom error to explain why this happens & to avoid it + document this behaviour in the API docs.

> RuntimeError: [E866] A SpanGroup is not functional after the corresponding Doc has been garbage collected. To keep using the spans, make sure that the corresponding Doc object is still available in the scope of your function.

I'm not sure there's a solution where we can avoid the whole issue entirely...

### Types of change
UX & docs

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -115,6 +115,8 @@ class Warnings:
     W036 = ("The component '{name}' does not have any patterns defined.")
 
     # New warnings added in v3.x
+    W085 = ("A SpanGroup is not functional after the corresponding Doc has "
+            "been garbage collected.")
     W086 = ("Component '{listener}' will be (re)trained, but it needs the component "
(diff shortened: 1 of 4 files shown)


honnibal:
Yes I think this error is the best thing.

I wouldn't want to try to be clever and play with the `doc.__dealloc__` to try to save the `Doc` object if it has spans that are alive. That will be slow and really buggy, because we'll be resurrecting the `Doc`. The user does just need to change their code, which is easy enough if we tell them what to do.

PR #8983, opened 2021-08-18
Add glossary entry for _SP (fix #8982)

### LINKED ISSUE #8982: _SP tag is not documented
```
spacy.load('en_core_web_trf')('a  b')[1].tag_
```

returns `_SP`, but this tag is not documented in https://spacy.io/models/en as well as lacks an explanation when requested with `spacy.explain('_SP')`.

polm:
## Description

This adds a glossary entry for the `_SP` tag, which is used for whitespace that gets a tag rather than being subsumed. This wasn't documented anywhere, as pointed out in #8982. 

It's possible this was `SP` at some point in the past but was changed. In Chinese OntoNotes SP is used for "sentence [final] particles", which is unrelated.

I guess this should also be added to model docs?

### Types of change

Glossary addition.

## Checklist

### DIFF
--- a/spacy/glossary.py
+++ b/spacy/glossary.py
@@ -95,6 +95,7 @@ GLOSSARY = {
     "XX": "unknown",
     "BES": 'auxiliary "be"',
     "HVS": 'forms of "have"',
+    "_SP": "whitespace",
     # POS Tags (German)
     # TIGER Treebank
     # http://www.ims.uni-stuttgart.de/forschung/ressourcen/korpora/TIGERCorpus/annotation/tiger_introduction.pdf


adrianeboyd:
I think this is fine for `spacy.explain`. Documenting on the models page is going to be hard because it's not a tag from the tagger and adding some kind of ad-hoc `attribute_ruler` tags section with subsections for each possible type of label is probably too messy. Maybe we can add a section here: https://spacy.io/models#design, so at least if you search for `_SP` or `SPACE` in the docs you find something about how the attribute ruler does this in the pretrained pipelines.

PR #8993, opened 2021-08-19
Update default_config.cfg

YashMulchandani:
add language "en"

## Checklist


adrianeboyd:
Hi, where have you run into problems with the language setting as it is in the current default config?

PR #9001, opened 2021-08-19
Pass alignments to Matcher callbacks

kwhumphreys:
## Description

When Matcher is called with `with_alignments=True`, make alignments accessible in `on_match` callback functions.

### Types of change

Enhancement to make Matcher results consistent in callbacks.

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -283,7 +283,7 @@ cdef class Matcher:
                     memset(&matched[start], 1, span_len * sizeof(matched[0]))
         if as_spans:
             final_results = []
-            for i, (key, start, end, *_) in enumerate(final_matches):
+            for key, start, end, *_ in final_matches:
                 if isinstance(doclike, Span):
                     start += doclike.start
                     end += doclike.start
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -284,18 +284,17 @@ cdef class Matcher:
         if with_alignments:
             final_matches_with_alignments = final_matches
             final_matches = [(key, start, end) for key, start, end, alignments in final_matches]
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Hi,

Thanks for the PR, much appreciated! I get what you're trying to do and I think it makes sense to make this change. 

I was just wondering, aesthetically, whether we can avoid copy-pasting the `on_match` application by not returning the results in either of the two `if`/`elif` statements, but rather build an array `final_results` or something such that either holds the `spans` or the `matches`, and which is used to iterate over once for the `on_match` calls right at the end of the function before anything is returned? 

(I think this explanation is more lengthy than the actual chang required- you'd basically just need to rename a few parameters and remove the `return spans` line)

kwhumphreys:
Thanks for the suggestion. I wasn't clear why alignments were being removed here https://github.com/explosion/spaCy/blob/master/spacy/matcher/matcher.pyx#L286 so I didn't want to change it originally, but it seems that the removal is not needed.

svlandeg:
Yea I think you solved it elegantly with the `*_` bit!

spacy/matcher/matcher.pyx
```diff
@@ -281,28 +281,19 @@ cdef class Matcher:
...
-        # perform the callbacks on the filtered set of results
-        for i, (key, start, end) in enumerate(final_matches):
-            on_match = self._callbacks.get(key, None)
-            if on_match is not None:
-                on_match(self, doc, i, final_matches)
         if as_spans:
-            spans = []
-            for key, start, end in final_matches:
+            final_results = []
+            for i, (key, start, end, *_) in enumerate(final_matches):
```
svlandeg, line 286:
```suggestion
            for key, start, end, *_ in final_matches:
```
It doesn't matter much, but I don't think we need to `enumerate` here?

PR #9007, opened 2021-08-20
Fix spancat training on nested entities

svlandeg:
## Description
We had a bug in the training of the spancat: gold entities were aligned without specifying `allow_overlap=True`, effectively preventing the `spancat` from learning nested entities. This PR fixes it and introduces a unit test to ensure correct behaviour.

### Types of change
bug fix

## Checklist


svlandeg:
We'll release this bugfix as part of 3.1.2

PR #9032, opened 2021-08-24
Retain docs' positional information when no exceptions are raised during multiprocessing

baxtree:
Make` nlp.pipe` return None docs when no exceptions are (re-)raised during error handling.

## Description

ATM, `nlp.pipe` returns nothing related to the original texts when it is failed on processing them if errors are handled and not (re-)raised. It is hard to associate the output docs with the input texts without conducting string comparison on the texts (and more complicatedly, their strings could be identical). Say the processing on "text_2" is failed, you will have:
```
[text_1, text_2, text_3] => [doc_1, doc_3]
```
Also, `as_tuples=True` does not come to the rescue and even generates inconsistency based on the current implementation coz you will have:
<pre>
[(text_1, context_1), (text_2, context_2), (text_3, context_3)] => [(doc_1, context_1), (<b>doc_3</b>, <b>context_2</b>)]
</pre>
Thus, `nlp.pipe` could return something without losing the positional information. `None` is used in this PR but subject to discussion or other potential fixes more agreed-on:
```
[text_1, text_2, text_3] => [doc_1, None, doc_3]
[(text_1, context_1), (text_2, context_2), (text_3, context_3)] => [(doc_1, context_1), (None, context_2), (doc_3, context_3)]
```
### Types of change

An enhancement to error handling during multiprocessing.

## Checklist


adrianeboyd:
Hi, we discussed this internally and we think the best option is to use `None` in the internal methods but then filter those docs out in the final `yield` so that the return signature stays the same as `Iterator[Doc]` or `Iterator[Tuple[Doc, _AnyContext]]`.

If you need to keep track of the individual documents to align them with the input, you can do this with custom metadata, either in the `Doc` or the `tuple` info.

I'm not sure whether we want `return None` added to each component method or instead made explicit/required in the error handlers and then having `return / yield from error_handler...`. I think I'd prefer requiring the error handler to return `None` so authors of new components don't need to hard-code this.

baxtree:
Thanks for your review and discussion around the improvement. I will try to find a time slot and see what can be done based on the preference.

adrianeboyd:
We're probably happy to make the adjustments ourselves (we think this is a good idea no matter what), but if you have feedback/ideas, we're happy to hear it.

baxtree:
Oh cool. I think your design is so far so good. The only problem is a failed doc could "contaminate" the entire batch. There shall be a way to get a finer-grained error handling on that and it would be nice to have more clarity on that.

adrianeboyd:
Because of the way documents are processed component by component in batches, I'm not sure there's a graceful way to recover if one doc fails. It does mean that the results are different for `nlp` vs. `nlp.pipe` with the same texts and same error handler.

You could also potentially back off to `batch_size=1` if you run into errors, but I think this is starting to get into the territory that the user should handle, seen as a parallel part to developing the custom error handler.

Oh, wait, I just tried to implement this and the tuples part fails if `Language.pipe` skips the `None` docs, since now there's no alignment. Argh, let's see...

adrianeboyd:
Hi, just as a note, we're still a bit stuck at how to do this without modifying the return signature of `Language.pipe`.

baxtree:
Hi, sure and let me know if I can help. `None` is no longer returned by individual components but returned by error handlers, as suggested in your review. However, since users could make their custom error handler return arbitrary types atm, `yield None` is preserved in the outer multiprocessing pipe.

baxtree:
Hi, @adrianeboyd, can I ask how the designing/fixing is going on your side?

adrianeboyd:
Hi, thinking about this again, I think we can find a solution that takes advantage of #9069 and sets the context internally on the docs during processing. Let me give this a shot...

(The CLI/compat test failures are unrelated.)

adrianeboyd:
This is the basic idea, but the last commit hasn't been merged yet from #9597.

adrianeboyd:
So this is a bit anticlimactic, but after #9069, #9597, and #9606, `as_tuples` works out of the box and we don't need any of the modifications related to returning `None` to get the right alignment.

Aside from the `mock` warning detection with multiprocessing issue that we also didn't figure out in the original custom error handler implementation, the only thing we need to do at this point is add the test. I moved this to a separate PR so the history is a bit easier to manage: #9608.

baxtree:
Glad to see `doc._context` was introduced so the following bug should have been fixed:
`[(text_1, context_1), (text_2, context_2), (text_3, context_3)] => [(doc_1, context_1), (doc_3, context_2)]`

Looks like for the scenario where `as_tuples=False` and exceptions are swallowed, users may still want to find out which texts failed on nlp by themselves. So by design, do they need to set `doc._context` in their custom error handlers? The failed docs will be filtered out anyway so doing that seems to have no benefits. Do they need to set `doc._context` way before the error handling or always prefer the `as_tuples=True` solution?

adrianeboyd:
I think that's a case where the error handler itself can log/save/whatever the failed docs however the user prefers?

adrianeboyd:
In any case, users shouldn't be setting `doc._context`, there's no guarantee about what might happen with private variables.

PR #9033, opened 2021-08-24
Fix verbs list in lang/fr/tokenizer_exceptions.py

davidstrouk:
## Checklist


svlandeg:
Thanks for this! 

@adrianeboyd : do you want to move this to the `develop` branch or can we merge it here on `master`?

adrianeboyd:
I was waffling about this, but I think it's close enough to a bug fix that we should just do it on `master`.

PR #9050, opened 2021-08-25
Add overwrite settings for more components

adrianeboyd:
## Description

For pipeline components where it's relevant and not already implemented, add an explicit `overwrite` setting that controls whether `set_annotations` overwrites existing annotation.

For the `morphologizer`, add an additional setting `extend`, which controls whether the existing features are preserved.

* +overwrite, +extend: overwrite values of existing features, add any new features
* +overwrite, -extend: overwrite completely, removing any existing features
* -overwrite, +extend: keep values of existing features, add any new features
* -overwrite, -extend: do not modify the existing value if set

In all cases an unset value will be set by `set_annotations`.

Preserve current overwrite defaults:

* True: morphologizer, entity linker
* False: tagger, sentencizer, senter

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -498,8 +498,6 @@ class EntityLinker(TrainablePipe):
         deserialize["kb"] = lambda p: self.kb.from_disk(p)
         deserialize["model"] = load_model
         util.from_disk(path, deserialize, exclude)
-        if "overwrite" not in self.cfg:
-            self.cfg["overwrite"] = BACKWARD_OVERWRITE
         return self
 
     def rehearse(self, examples, *, sgd=None, losses=None, **config):
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -20,6 +20,8 @@ from ..util import SimpleFrozenList, registry
 from .. import util
 from ..scorer import Scorer
 
+# See #9050
+BACKWARD_OVERWRITE = True
 
 default_model_config = """
 [model]
@@ -403,7 +405,7 @@ class EntityLinker(TrainablePipe):
         if count_ents != len(kb_ids):
             raise ValueError(Errors.E148.format(ents=count_ents, ids=len(kb_ids)))
         i = 0
-        overwrite = self.cfg.get("overwrite", False)
+        overwrite = self.cfg["overwrite"]
         for doc in docs:
             for ent in doc.ents:
                 kb_id = kb_ids[i]
@@ -452,6 +454,8 @@ class EntityLinker(TrainablePipe):
         deserialize["kb"] = lambda b: self.kb.from_bytes(b)
         deserialize["model"] = load_model
         util.from_bytes(bytes_data, deserialize, exclude)
+        if "overwrite" not in self.cfg:
+            self.cfg["overwrite"] = BACKWARD_OVERWRITE
         return self
 
     def to_disk(
@@ -496,6 +500,8 @@ class EntityLinker(TrainablePipe):
         deserialize["kb"] = lambda p: self.kb.from_disk(p)
         deserialize["model"] = load_model
         util.from_disk(path, deserialize, exclude)
+        if "overwrite" not in self.cfg:
+            self.cfg["overwrite"] = BACKWARD_OVERWRITE
         return self
 
     def rehearse(self, examples, *, sgd=None, losses=None, **config):
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -50,6 +50,7 @@ DEFAULT_NEL_MODEL = Config().from_str(default_model_config)["model"]
         "incl_context": True,
         "entity_vector_length": 64,
         "get_candidates": {"@misc": "spacy.CandidateGenerator.v1"},
+        "overwrite": True,
         "scorer": {"@scorers": "spacy.entity_linker_scorer.v1"},
     },
     default_score_weights={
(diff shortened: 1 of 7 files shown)


adrianeboyd:
One of the motivations for `extend` is a case like #8977.

adrianeboyd:
This is also a case where the duplication between the config and `model/cfg` is a bit confusing / problematic.

spacy/pipeline/sentencizer.pyx
```diff
@@ -152,6 +154,7 @@ class Sentencizer(Pipe):
         """
         cfg = srsly.msgpack_loads(bytes_data)
         self.punct_chars = set(cfg.get("punct_chars", self.default_punct_chars))
+        self.overwrite = cfg.get("overwrite", False)
```
svlandeg, line 157: Do we need this default `False` value here? If the setting is not in `cfg`, we can just keep `self.overwrite` as it was by default after construction? Actually isn't the same true also for `self.punct_chars`?

So you'd have something like 
``` suggestion
        if "punct_chars" in cfg:
            self.punct_chars = set(cfg.get("punct_chars"))
        if "overwrite" in cfg:
            self.overwrite = cfg.get("overwrite", False)
```
and then the same in `from_disk` and potentially some other components.

I just want to be mindful of these hidden defaults creeping back into the code...

adrianeboyd: I agree about hidden defaults, but I don't think this is a hidden default. This is the hard-coded setting from <3.2.0 which isn't going to change, so if you're loading a model that was serialized with v3.1 or earlier, you get the same default in cases where this `cfg` value is missing no matter what you've specified on init. If the `cfg` value is present, it's v3.2+ and you get the stored value. (If you're hacking your `cfg`, well, that's on you.)

I think something similar happened in the `punct_chars` development, but I'd have to check the history to be sure.

I would keep both of these as is.

svlandeg: Ah, I see what you mean. Then I'd personally prefer putting this value in a var somewhere and document that it's for bwd-compat only.

spacy/pipeline/morphologizer.pyx
```diff
@@ -217,14 +226,34 @@ class Morphologizer(Tagger):
...
         for i, doc in enumerate(docs):
             doc_tag_ids = batch_tag_ids[i]
             if hasattr(doc_tag_ids, "get"):
                 doc_tag_ids = doc_tag_ids.get()
             for j, tag_id in enumerate(doc_tag_ids):
                 morph = self.labels[tag_id]
-                doc.c[j].morph = self.vocab.morphology.add(self.cfg["labels_morph"].get(morph, 0))
-                doc.c[j].pos = self.cfg["labels_pos"].get(morph, 0)
+                # set morph
+                if doc.c[j].morph == 0 or overwrite or extend:
```
polm, line 238: I understand what's happening with `morph == 0` but it took me a second to understand it and I still think it looks weird (though it's also the best way to check this). Maybe there should be a comment here?

spacy/pipeline/morphologizer.pyx
```diff
@@ -217,14 +226,34 @@ class Morphologizer(Tagger):
...
+                    if overwrite and extend:
+                        # morphologizer morph overwrites any existing features
+                        # while extending
```
polm, line 241: Why is the comment wrapped but not the code here?

spacy/pipeline/sentencizer.pyx
polm: I think this is fine as-is? If the var is otherwise unused I think we can accomplish the same thing with a comment here.

polm: I think this is fine as-is. If the var isn't used for anything else then I think a comment explaining why this is the way it is will accomplish the same thing.

svlandeg: The same value is used twice within `Sentencizer`'s IO methods, so let's just pull it out into a `default_` var and document the bwd-compat reasoning.

spacy/pipeline/morphologizer.pyx
adrianeboyd: This is how we always check for unset values in `set_annotations`.

spacy/pipeline/morphologizer.pyx
adrianeboyd: I think the answer is: I don't have anything that auto-formats cython?

svlandeg (approved):

adrianeboyd:
Ack, I realized that this isn't going to work like we intended. Since the deserialization just updates the config, this will have the default of the `__init__` and the case where it's unset will never happen. If we want this to actually work in cases where it's missing from the saved model, we'd somehow need a custom `load_cfg` deserializer that replaces the default with `BACKWARD_OVERWRITE`.

adrianeboyd:
So the only setting that matters now is the one provided for the kwarg in `__init__`. Maybe this is okay and we just don't need any of these defaults at all?

svlandeg:
Right, but then if we ever want to change the default behaviour going forward, old models will suddenly behave differently :/

adrianeboyd:
Then we'll have to leave the defaults until v4, I guess, which is probably what we would have wanted to do anyway.

spacy/pipeline/entity_linker.py
```diff
@@ -490,6 +498,8 @@ def load_model(p):
         deserialize["kb"] = lambda p: self.kb.from_disk(p)
         deserialize["model"] = load_model
         util.from_disk(path, deserialize, exclude)
+        if "overwrite" not in self.cfg:
+            self.cfg["overwrite"] = BACKWARD_OVERWRITE
```
svlandeg, line 502: Does this one still make sense?

adrianeboyd: No. It won't do anything, but I meant to remove it.

svlandeg (approved):
I think this looks good now? Definitely an unexpectedly difficult PR :|

PR #9069, opened 2021-08-27
Accept Doc input in pipelines

adrianeboyd:
## Description

Allow `Doc` input to `Language.__call__` and `Language.pipe`, which
skips `Language.make_doc` and passes the doc directly to the pipeline.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/tests/test_language.py
+++ b/spacy/tests/test_language.py
@@ -528,3 +528,15 @@ def test_language_source_and_vectors(nlp2):
     assert long_string in nlp2.vocab.strings
     # vectors should remain unmodified
     assert nlp.vocab.vectors.to_bytes() == vectors_bytes
+
+
+def test_pass_doc_to_pipeline(nlp):
+    texts = ["cats", "dogs", "guinea pigs"]
+    docs = [nlp.make_doc(text) for text in texts]
+    assert not any(len(doc.cats) for doc in docs)
+    doc = nlp(docs[0])
+    assert doc.text == texts[0]
+    assert len(doc.cats) > 0
+    docs = nlp.pipe(docs)
+    assert [doc.text for doc in docs] == texts
+    assert all(len(doc.cats) for doc in docs)
--- a/spacy/tests/test_language.py
+++ b/spacy/tests/test_language.py
@@ -540,8 +540,8 @@ def test_pass_doc_to_pipeline(nlp, n_process):
     assert len(doc.cats) > 0
     if isinstance(get_current_ops(), NumpyOps) or n_process < 2:
         docs = nlp.pipe(docs, n_process=n_process)
-    assert [doc.text for doc in docs] == texts
-    assert all(len(doc.cats) for doc in docs)
+        assert [doc.text for doc in docs] == texts
+        assert all(len(doc.cats) for doc in docs)
 
 
 def test_invalid_arg_to_pipeline(nlp):
--- a/spacy/tests/test_language.py
+++ b/spacy/tests/test_language.py
@@ -530,13 +530,26 @@ def test_language_source_and_vectors(nlp2):
     assert nlp.vocab.vectors.to_bytes() == vectors_bytes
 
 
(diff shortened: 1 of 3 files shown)


adrianeboyd:
I put this in draft mode because there are a few questions, especially related to typing.

Only tangentially related to typing:

Do we want `Language.make_doc` to do the doc skipping or do we want it in `__call__` and `pipe` (as in the initial proposal)?

Related to typing:

If you allow a custom tokenizer to do kind of anything (e.g., take a dict and build a doc out of that), then you have to type `text` as `Any` in `__call__`, `pipe`, and `make_doc`, which is kind of displeasing.

I've been recommending custom tokenizers for things like this because there was no other alternative, so I'm not sure if we want to stop recommending this or allow it type-wise. Otherwise we're recommending a multi-step doc-building / pipe-calling process, which isn't terrible, but would be a shift from what we've recommended so far.

polm:
My inclination is to avoid allowing anything as input so that we can continue to provide meaningful errors when someone passes a list or dataframe or whatever by accident. Could we instead allow passing in a dict to be used as the base for user_data or something?

I think adding this in `__call__` and `pipe` makes sense, in `pipe` it would be similar to the contexts I guess.

adrianeboyd:
I agree that `text: Any` is bad for a lot of reasons.

The `user_data` idea is kind of what this user was suggesting: #9077 . I don't think there's a good general-purpose way to do this, though, and I also think we should not really touch `user_data` in the core library if it all possible so that there aren't unexpected interactions.

And I don't think we should change the API of `__call__` or `pipe`. (For example, I think we've discussed that the `as_tuples` functionality would have been better as a separate method.)

I guess my conclusion is that the multi-step option may be the better recommendation going forward for people who care about typing and custom doc creation.

spacy/tests/test_language.py
```diff
@@ -528,3 +528,28 @@ def test_language_source_and_vectors(nlp2):
...
+def test_pass_doc_to_pipeline(nlp, n_process):
+    texts = ["cats", "dogs", "guinea pigs"]
+    docs = [nlp.make_doc(text) for text in texts]
+    assert not any(len(doc.cats) for doc in docs)
+    doc = nlp(docs[0])
+    assert doc.text == texts[0]
+    assert len(doc.cats) > 0
+    docs = nlp.pipe(docs, n_process=n_process)
+    assert [doc.text for doc in docs] == texts
+    assert all(len(doc.cats) for doc in docs)
```
adrianeboyd, line 543: This test needs to check for CPU ops only or it will hang/crash.

svlandeg: I'll get that check in, but these should really be xfailing tests rather than hardcoding these kind of exceptions :|

svlandeg: (but for another day...)

spacy/tests/test_language.py
```diff
@@ -528,3 +528,29 @@ def test_language_source_and_vectors(nlp2):
...
+    texts = ["cats", "dogs", "guinea pigs"]
+    docs = [nlp.make_doc(text) for text in texts]
+    assert not any(len(doc.cats) for doc in docs)
+    doc = nlp(docs[0])
+    assert doc.text == texts[0]
+    assert len(doc.cats) > 0
+    if isinstance(get_current_ops(), NumpyOps) or n_process < 2:
+        docs = nlp.pipe(docs, n_process=n_process)
+    assert [doc.text for doc in docs] == texts
+    assert all(len(doc.cats) for doc in docs)
```
adrianeboyd, line 544:
```suggestion
        assert [doc.text for doc in docs] == texts
        assert all(len(doc.cats) for doc in docs)
```

spacy/tests/test_language.py
adrianeboyd: This is how I've done it so far, if you know of a more general way to handle this, then that would be welcome, but for another day...

PR #9074, opened 2021-08-27
Fix surprises when asking for the root of a git repo

rspeer:
## Description

In the case of the first asset I wanted to get from git, the data I wanted was the entire repository. I tried leaving "path" blank, which gave a less-than-helpful error, and then I tried `path: "/"`, which started copying my entire filesystem into the project. The path I should have used was "".

I've made two changes to make this smoother for others:

- The 'path' within a git clone defaults to ""
- If the path points outside of the tmpdir that the git clone goes into, we fail with an error

### Types of change

This is an ergonomics/bug fix, with a corresponding change in the documentation to describe the added default.

## Checklist

(Instead of submitting the Contributor Agreement, I have added a `Signed-off-by` line to the commit, which I understand is the new practice.)

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -399,7 +399,7 @@ def git_checkout(
         try:
             source_path = tmp_dir / Path(subpath)
             if not is_subpath_of(tmp_dir, source_path):
-                err = f"{subpath!r} is a path outside of the cloned repository."
+                err = f"'{subpath}' is a path outside of the cloned repository."
                 msg.fail(err, repo, exits=1)
             shutil.copytree(str(source_path), str(dest))
         except FileNotFoundError:
@@ -450,10 +450,10 @@ def git_sparse_checkout(repo, subpath, dest, branch):
         cmd = f"git -C {tmp_dir} checkout {branch} {subpath}"
         run_command(cmd, capture=True)
 
-        # Get a subdirectory of the cloned path, if approriate
+        # Get a subdirectory of the cloned path, if appropriate
         source_path = tmp_dir / Path(subpath)
         if not is_subpath_of(tmp_dir, source_path):
-            err = f"{subpath!r} is a path outside of the cloned repository."
+            err = f"'{subpath}' is a path outside of the cloned repository."
             msg.fail(err, repo, exits=1)
 
         shutil.move(str(source_path), str(dest))
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -397,7 +397,11 @@ def git_checkout(
         run_command(cmd, capture=True)
         # We need Path(name) to make sure we also support subdirectories
         try:
-            shutil.copytree(str(tmp_dir / Path(subpath)), str(dest))
+            source_path = tmp_dir / Path(subpath)
+            if not is_subpath_of(tmp_dir, source_path):
+                err = f"{subpath!r} is a path outside of the cloned repository."
+                msg.fail(err, repo, exits=1)
+            shutil.copytree(str(source_path), str(dest))
         except FileNotFoundError:
             err = f"Can't clone {subpath}. Make sure the directory exists in the repo (branch '{branch}')"
             msg.fail(err, repo, exits=1)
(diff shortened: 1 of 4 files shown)


spacy/cli/_util.py
```diff
@@ -397,7 +397,11 @@ def git_checkout(
         run_command(cmd, capture=True)
         # We need Path(name) to make sure we also support subdirectories
         try:
-            shutil.copytree(str(tmp_dir / Path(subpath)), str(dest))
+            source_path = tmp_dir / Path(subpath)
+            if not is_subpath_of(tmp_dir, source_path):
+                err = f"{subpath!r} is a path outside of the cloned repository."
```
svlandeg, line 402: Is this `!r` notation just to put single quotes around the parameter? In all other places in the repo we'd just use `'{par}'` but I guess we can start adopting this notation if it's commonly used - I hadn't seen it before.

(I do think that a small advantage of using the more simple `'{par}'` is that it'll also put `None` inbetween quotes which I think reads more easily)

spacy/cli/_util.py
```diff
@@ -445,8 +449,14 @@ def git_sparse_checkout(repo, subpath, dest, branch):
         # And finally, we can checkout our subpath
         cmd = f"git -C {tmp_dir} checkout {branch} {subpath}"
         run_command(cmd, capture=True)
-        # We need Path(name) to make sure we also support subdirectories
-        shutil.move(str(tmp_dir / Path(subpath)), str(dest))
+
+        # Get a subdirectory of the cloned path, if approriate
```
svlandeg, line 453:
```suggestion
        # Get a subdirectory of the cloned path, if appropriate
```

spacy/cli/project/assets.py
```diff
@@ -61,7 +61,7 @@ def project_assets(project_dir: Path, *, sparse_checkout: bool = False) -> None:
                         dest.unlink()
             git_checkout(
                 asset["git"]["repo"],
-                asset["git"]["path"],
+                asset["git"].get("path", ""),
```
svlandeg, line 64: I don't think we should have these kind of default values defined deep within the code, it makes it difficult to track where defaults are getting introduced. Could we instead just check whether `subpath is None` in `git_checkout` and raise a custom error there? In the error message we could specify that `""` can be used to denote the entire directory.

svlandeg (commented):
It's definitely surprising that `Path(folder) / "/"` produces the root folder of the system! 
We might want to add a little note in code why the `is_subpath_of` check is needed, because in general I suppose everyone will expect `Path(folder) / subpath` to be a subpath of `folder` (I know I did).

spacy/cli/project/assets.py
rspeer: A couple of issues I see there -- this would violate the type signature of `git_checkout` by passing None in where a str is expected (unless our types are implicitly optional?), and also we'd have a less pertinent error message because it'd be referring to the `subpath` parameter of `git_checkout` instead of the "path" configuration option.

But I could check for this and raise a custom error in `project_assets` instead.

spacy/cli/_util.py
rspeer: One place I find the `!r` notation very helpful is if it's likely for the value to be a string containing an apostrophe -- like if we needed an error referring to the token "'s", it would automatically switch to double quotes. But that's not this situation, so I'll take out the `!r`.

svlandeg: Ah, that's a good point! We might want to consider using it in some other error messages where it could be more relevant :-)

spacy/cli/project/assets.py
```diff
@@ -59,6 +59,10 @@ def project_assets(project_dir: Path, *, sparse_checkout: bool = False) -> None:
                         shutil.rmtree(dest)
                     else:
                         dest.unlink()
+            if "repo" not in asset["git"]:
+                msg.fail("A git asset must include 'repo', the repository address.", exits=1)
+            if "path" not in asset["git"]:
+                msg.fail("A git asset must include 'path' - use \"\" to get the entire repository.", exits=1)
```
svlandeg, line 65:
```suggestion
            if "path" not in asset["git"] or not asset["git"]["path"]:
                msg.fail("A git asset must include 'path' - use \"\" to get the entire repository.", exits=1)
```
I think the original issue was caused by `asset["git"]["path"]` being `None`, which technically means it's present, so we'd need one more additional check.

rspeer: Oh - the original issue I had was a KeyError from leaving out the path, but I can add that check too.

svlandeg: Ah, right, well the extra check will catch cases where the config just reads
```
path: 
```

PR #9081, opened 2021-08-28
benepar usage example has deprecated imports

mjhajharia:
the previous import method was deprecated, changed that to a working code!


polm:
Thanks for the fix!

PR #9084, opened 2021-08-30
Fix inference of epoch_resume

polm:
When an epoch_resume value is not specified individually, it can often
be inferred from the filename. The value inference code was there but
the value wasn't passed back to the training loop.

This also adds a specific error in the case where no epoch_resume value
is provided and it can't be inferred from the filename.

This does not handle the case where a value can be inferred from the filename and an epoch resume value is explicitly passed. In that case the explicitly passed value will be ignored, which seems a bit weird. Should there be a warning or error or something instead?

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -869,6 +869,8 @@ class Errors:
     E1019 = ("`noun_chunks` requires the pos tagging, which requires a "
              "statistical model to be installed and loaded. For more info, see "
              "the documentation:\nhttps://spacy.io/usage/models")
+    E1020 = ("No `epoch_resume` value specified and could not infer one from "
+             "filename. Specify an epoch to resume from.")
 
 
 # Deprecated model shortcuts, only used in errors and warnings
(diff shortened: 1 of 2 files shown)


svlandeg:
> This does not handle the case where a value can be inferred from the filename and an epoch resume value is explicitly passed. In that case the explicitly passed value will be ignored, which seems a bit weird. Should there be a warning or error or something instead?

That doesn't seem right. If it's explicitely passed, we should keep that value?

polm:
> That doesn't seem right. If it's explicitely passed, we should keep that value?

Currently we don't keep it. 

On the one hand, if the filename has a value, I'd trust that first. But if we don't take the explicitly provided epoch number there would be no way to override the filename, so I guess we have to respect it. 

I'll change it to use the provided epoch when specified.

svlandeg (approved):
I do think this makes more sense. Ultimately this will really only affect how much more epochs are trained, and if a user says the provided epoch is 10 out of 30, well, then that's how it is I guess, regardless of the filename. 

I don't think a mismatch between the filename and the provided `epoch_resume` happens very frequently, so I think this is more a bugfix rather than a breaking change (to decide on whether we keep it on `master` or move to `develop`). 

What do you think, Paul? If it's fine by you like this, feel free to merge.

polm:
Yeah I think the old behavior wasn't useful, so this should be fine for `master`.

PR #9102, opened 2021-08-31
Add an Irish lemmatiser, based on BuNaMo

jimregan:
## Description
Initial stab at a lemmatiser for Irish. Depends on data from [BuNaMo](https://github.com/michmech/BuNaMo)

### Types of change
- adds a lookup-based lemmatiser for Irish
- adds two tokeniser exceptions

## Checklist


adrianeboyd:
Thanks for this PR! It is still a WIP or is it ready to be reviewed? It's no problem for us to test it internally with your `spacy-lookups-data` tables even if there's not yet an official release of `spacy-lookups-data` yet.

jimregan:
I _think_ it was just the dependency on the data, but a lot has happened in the last two weeks and I don't trust my memory.

adrianeboyd:
I tested this lemmatizer with a morphologizer trained on `UD_Irish-IDT` without word vectors. The morphologizer performance is not great (no vectors, fairly small training corpus), which I assume is part of the reason that the lemmatizer performance is a bit low, but I hope it will still be useful, and it can always be refined in the future.

```none
TOK     99.60
POS     87.70
MORPH   65.72
LEMMA   82.30
```

jimregan:
> I tested this lemmatizer with a morphologizer trained on `UD_Irish-IDT` without word vectors. The morphologizer performance is not great (no vectors, fairly small training corpus), which I assume is part of the reason that the lemmatizer performance is a bit low, but I hope it will still be useful, and it can always be refined in the future.
> 
> ```
> TOK     99.60
> POS     87.70
> MORPH   65.72
> LEMMA   82.30
> ```

Out of curiosity, which version of UD_Irish-IDT? From what I can determine, early versions had a mixture of expert-annotated and automatically annotated morphological data, so it's possibly a problem with the data itself.

adrianeboyd:
Oh, that could be. I used v2.5 because it's the version currently used by a lot of the existing pretrained pipelines, but it would have been better to use the most recent release, which looks a lot bigger. What's the accuracy that you're expecting?

adrianeboyd:
I tried the most recent release (v2.8) and POS/MORPH looks quite a bit better, but LEMMA is slightly worse. I'm not sure what the cause is? (Fewer automatically annotated lemmas that are actually wrong?)

```
TOK     99.60
POS     92.99
MORPH   84.41
LEMMA   81.29
```

jimregan:
That's probably a school-of-thought thing; verbal substantives are treated as verbs in BuNaMo, they may be treated as nouns in IDT. I think the dust has settled on how they're annotating things, so I'll look into it.

adrianeboyd:
So if IDT is following the UD standards in terms of UPOS, then it would good if this lemmatizer followed the same standards in its lookups. If it's a matter of how the lemmatizer treats particular cases where the UPOS is correct (an example might be exactly how to lemmatize pronouns in English, which can vary between corpora), then we don't necessarily need to follow the exact IDT standard. It was just an easy corpus to test on to make sure everything is running.

jimregan:
> So if IDT is following the UD standards in terms of UPOS, then it would good if this lemmatizer followed the same standards in its lookups. If it's a matter of how the lemmatizer treats particular cases where the UPOS is correct (an example might be exactly how to lemmatize pronouns in English, which can vary between corpora), then we don't necessarily need to follow the exact IDT standard. It was just an easy corpus to test on to make sure everything is running.

Both interpretations are "correct", depending on the linguistic school of thought.

The base case doesn't actually need a lookup; there is a possibility that lookup is needed with genitive case forms, but the only example is in the `dev` set, so it wouldn't be reached anyway (i.e., it can wait).

Having looked at the data, I think the verbal substantive is only responsible for a small number of misses; the rest are places where words were incorrectly lemmatised (like the English 'white' -> 'whi'), inconsistent lemmatisation (with verbal adjectives, some use the form as the lemma, others the verb's lemma), and non-standard forms.

I had considered adding non-standard forms, but it would have involved a two-stage lookup, and it would change a lot, so it's better to wait until the external loading is in place.

adrianeboyd:
I made a few minor adjustments again because the `Token.morph` API isn't quite like a dict. The changes didn't make a very large difference, but you said these cases were rare, right?

```none
POS     92.99
MORPH   84.41
LEMMA   81.35
```

Is there a more suitable corpus just to evaluate the lemmas on, just so we know for internal evaluations?

It's also fine with me to merge it like this. Just be aware that we don't make new releases of spacy that frequently and changes to the lemmatizers would only be made on minor releases (so v3.2.0, v3.3.0), so there can be a pretty long lag before changes are available in an official release (sometimes 3+ months). There's still a little bit of time before v3.2.0, so we can also leave this open if you would still like to work on it.

adrianeboyd (approved):

jimregan:
Hmm... the cases I was aiming at in the last commit are actually extremely common, but there is a rare edge case that I basically ignored. There are a lot of common verbs where it would have been lemmatised correctly 'accidentally'.

tl;dr - this is about as good as it's going to get, for now

The thing with the verbal nouns is quite similar to English; 'reading' can be a (verbal) noun, like 'a reading of a will', or a participle, like 'she is reading'; in Irish, 'reading' is 'léamh', in 'léamh uachta' it's a noun, in 'tá sí ag léamh' it... performs the same function as the participle 'reading', but is still considered a noun. One major difference is that the relationship between 'read' and 'reading' is a predictable inflection, but in Irish the inflectional relationship has been lost over time, so they're learned as a set of related words (in this case, both lemmas happen to be the same). If I combine the two examples, 'tá sí ag léamh uachta', I wouldn't know if it was 'she is reading a will' or 'she is at a reading of a will'.

The only thing that really sets verbal uses of verbal nouns apart is a set of particles, most of which are identical to prepositions, so treating them purely as nouns shifts the ambiguity from POS to MORPH.

adrianeboyd:
Okay, thanks again for your contribution!

PR #9124, opened 2021-09-02
Add the spacy.models_with_nvtx_range.v1 callback

danieldk:
## Description

This callback recursively adds NVTX range markers to the Models in each pipe in
a pipeline. These markers aid in GPU profiling by attributing specific operations
to a `Model`'s forward or backprop passes.

The callback was tested by using it as an `after_init` callback and then profiling
a training run with Nsight.

### Types of change

This is a new feature.

## Checklist


svlandeg:
As the CI will fail until Thinc 8.0.9 is released, I'll put this in draft for now.

adrianeboyd:
I would suggest that `spacy.pipelines.callbacks` may not be the right place to add this. It doesn't interact directly with the `spacy.pipelines` components code and I think it would be better in `spacy.util` or similar.

danieldk:
> I would suggest that `spacy.pipelines.callbacks` may not be the right place to add this. It doesn't interact directly with the `spacy.pipelines` components code and I think it would be better in `spacy.util` or similar.

Ok, I'll move the callback. I had difficulty picking the right place.

danieldk:
That leads to a circular import. This callback needs `TrainablePipe` and `Language` and these in turn need `util`.

adrianeboyd:
Ah, right. Then maybe we do need a new module, but I think somewhere else would be better. This is only for debugging/profiling, right?

danieldk:
> This is only for debugging/profiling, right?

Indeed, primarily for profiling.

adrianeboyd:
Maybe `spacy.training.callbacks`?

danieldk:
> Maybe `spacy.training.callbacks` (which already exists)?

This also introduces a circular dependency, since training is used for `TrainablePipe`. Also, I guess `spacy.training.callbacks` my be a bit odd, since it's also useful for profiling prediction.

Maybe a new `spacy.profiling`?

adrianeboyd:
Hmm, maybe others should weigh in here. I'd be a bit hesitant about adding another top-level module just for this.

If it's not just used for training, then I think the recommendation in the docs should potentially be to use it in `after_creation` rather than `after_init`, since that's not called outside of training.

spacy/pipeline/callbacks.py
```diff
@@ -0,0 +1,20 @@
...
+from thinc.layers import with_nvtx_range
+from thinc.model import wrap_model_recursive
+
+from ..language import Language
+from ..pipeline import TrainablePipe
+from ..util import registry
+
+
+@registry.callbacks("spacy.models_with_nvtx_range.v1")
+def create_models_with_nvtx_range() -> Callable[[Type[Language]], Type[Language]]:
```
adrianeboyd, line 12: I probably just don't understand the details, but: why is this `Type[Language]` instead of `Language`?

danieldk: You are right, this is not a callback that can be used for `before_creation`, so the type signature is wrong. Thanks, I have fixed this.

adrianeboyd: Ah, okay. I think the return type on the callback I wrote in `spacy.training.callbacks` is also wrong, but in the other direction.

danieldk:
I have put this in draft again, I want to update the PR to make it possible to configure the color for forward/backward passes.

danieldk:
I have updated the PR with support for setting colors, however https://github.com/explosion/thinc/pull/537 is required for functioning properly.

adrianeboyd:
I continue to be against `spacy.pipeline.callbacks`, I don't think it belongs here.

svlandeg:
Hm, `spacy.ml.callbacks` might be better? But I'm guess that will also result in a circular dependency. The `Language` typing can probably be resolved by using `"Language"`. To check the `TrainablePipe` instance, the workaround could be to check for a property `is_trainable` + check whether it's `True`?

danieldk:
> If it's not just used for training, then I think the recommendation in the docs should potentially be to use it in after_creation rather than after_init, since that's not called outside of training.

That would be better indeed, though I think we'd have to use `after_pipeline_creation`, since in `after_creation` the trainable pipes do not seem to be available yet.

I was testing this, turns out there is another issue in wrapping that will probably require a change in thinc. So let's put this in Draft again.

danieldk:
Changed to be compatible with https://github.com/explosion/thinc/pull/539.

adrianeboyd:
`spacy.ml.callbacks` sounds fine.

PR #9143, opened 2021-09-05
Update Cython string types

polm:
`basestring` was a compatibility type introduced by Cython to make
dealing with utf-8 strings in Python2 easier. In Python3 it is
equivalent to the unicode (or str) type. 

I replaced all references to basestring with unicode, since that was
used elsewhere, but we could also just replace them with str, which
should also be equivalent.

All tests pass locally.

### Types of change

cleanup

## Checklist


polm:
Note this has been modified beyond its original scope. It now replaces all `basestring` and `unicode` with `str` throughout the codebase. 

As a side effect of this, `from __future__ import unicode_literals` was removed - leaving it in caused a variety of errors.

In theory this should affect anything at all, and all the tests have passed, but it is kind of a big change.

PR #9148, opened 2021-09-05
Validate pos values when creating Doc

polm:
## Description

When you pass pos values to a Doc they aren't validated. Before this change, any value could be set via `Doc.__init__`, but if you tried to access `.pos_` on an invalid value you would get a `KeyError`.

In contrast, if you try to set an invalid pos value directly you get a `KeyError` immediately.

This checks if pos values are valid, and if not throws an error that explains they must be UD values, and other tags should use the `tag` field.

See #9145 for motivating case.

### Types of change

Added validation

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -872,7 +872,7 @@ class Errors:
     E1020 = ("No `epoch_resume` value specified and could not infer one from "
              "filename. Specify an epoch to resume from.")
     E1021 = ("`pos` value \"{pp}\" is not a valid Universal Dependencies tag. "
-             "Non-UD tags should use the `tag` field.")
+             "Non-UD tags should use the `tag` property.")
 
 
 # Deprecated model shortcuts, only used in errors and warnings
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -871,6 +871,8 @@ class Errors:
              "the documentation:\nhttps://spacy.io/usage/models")
     E1020 = ("No `epoch_resume` value specified and could not infer one from "
              "filename. Specify an epoch to resume from.")
+    E1021 = ("`pos` value \"{pp}\" is not a valid Universal Dependencies tag. "
+             "Non-UD tags should use the `tag` field.")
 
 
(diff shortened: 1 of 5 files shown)


omri374 (approved):

spacy/errors.py
```diff
@@ -871,6 +871,8 @@ class Errors:
              "the documentation:\nhttps://spacy.io/usage/models")
     E1020 = ("No `epoch_resume` value specified and could not infer one from "
              "filename. Specify an epoch to resume from.")
+    E1021 = ("`pos` value \"{pp}\" is not a valid Universal Dependencies tag. "
+             "Non-UD tags should use the `tag` field.")
```
adrianeboyd, line 875: If this error is only going to be used for `Doc.__init__`, then I think this should say `tags` instead of `tag`, something like:

```suggestion
             "Non-UD tags should use the `tags` argument.")
```

I wonder if want to use the same error instead of cryptic `KeyError` in `Token.pos_`, too, while we're at it, and then I'd have a different suggestion.

polm: Since we're dealing with it, I can go ahead and replace the `KeyError` too. What would you suggest?

spacy/tokens/doc.pyx
```diff
@@ -285,6 +286,10 @@ cdef class Doc:
                     sent_starts[i] = -1
                 elif sent_starts[i] is None or sent_starts[i] not in [-1, 0, 1]:
                     sent_starts[i] = 0
+        if pos is not None:
+            for pp in pos:
```
svlandeg, line 290: will it be quicker if we loop over `set(pos)` here?

polm: Ah, good point, that should have fewer `not in` checks.

svlandeg:
```suggestion
            for pp in set(pos):
```

adrianeboyd:
I think I would add a test for the `token.pos_ = "blah"` error, but otherwise this is fine.

Issue #8264, opened 2021-06-01
AttributeError: Can't get attribute 'cluster' on <module 'spacy.lang.lex_attrs' 


sebastiz:
## How to reproduce the behaviour

I am have used the fast.ai library to train a model for text classification (fastai2). This produces a pickle file. This was created in google colabs (
```
* **spaCy version:** 2.2.4
* **Platform:** Linux-5.4.109+-x86_64-with-Ubuntu-18.04-bionic
* **Python version:** 3.7.10
* **Models:** en)
```

I then tried to use the trained model in a new script in a jupyter notebook by loading it in to a learner

```
!pip install -Uqq fastbook
import fastbook
fastbook.setup_book()
from fastbook import *
from fastai.text.all import *

import pandas as pd
learner = load_learner('/content/gdrive/MyDrive/Colab_data/BarrettsPragueModel.pkl')
```

## Info about spaCy

```
* **spaCy version:** 2.3.4
* **Platform:** Linux-3.10.0-1160.24.1.el7.x86_64-x86_64-with-glibc2.10
* **Python version:** 3.8.8
```

When I load the learner I get the error as below:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-58-06af04e8023c> in <module>
----> 1 learner = load_learner('BarrettsPragueModel.pkl')

/opt/conda/lib/python3.8/site-packages/fastai/learner.py in load_learner(fname, cpu, pickle_module)
    379     "Load a `Learner` object in `fname`, optionally putting it on the `cpu`"
    380     distrib_barrier()
--> 381     res = torch.load(fname, map_location='cpu' if cpu else None, pickle_module=pickle_module)
    382     if hasattr(res, 'to_fp32'): res = res.to_fp32()
    383     if cpu: res.dls.cpu()

/opt/conda/lib/python3.8/site-packages/torch/serialization.py in load(f, map_location, pickle_module, **pickle_load_args)
    590                     opened_file.seek(orig_position)
    591                     return torch.jit.load(opened_file)
--> 592                 return _load(opened_zipfile, map_location, pickle_module, **pickle_load_args)
    593         return _legacy_load(opened_file, map_location, pickle_module, **pickle_load_args)
    594 

/opt/conda/lib/python3.8/site-packages/torch/serialization.py in _load(zip_file, map_location, pickle_module, pickle_file, **pickle_load_args)
[...]

PR #9150, opened 2021-09-06
Remove two attributes marked for removal in 3.1

polm:
## Description

Two attributes in `tokenizer.pxd` were marked for removal in 3.1 but left in. This just removes them. They were no longer in use anywhere.

### Types of change

Minor cleanup.

## Checklist

### DIFF
--- a/spacy/tokenizer.pxd
+++ b/spacy/tokenizer.pxd
@@ -23,6 +23,10 @@ cdef class Tokenizer:
     cdef object _infix_finditer
     cdef object _rules
     cdef PhraseMatcher _special_matcher
+    # TODO next two are unused and should be removed in v4
+    # https://github.com/explosion/spaCy/pull/9150
+    cdef int _unused_int1
+    cdef int _unused_int2
 
     cdef Doc _tokenize_affixes(self, unicode string, bint with_special_cases)
     cdef int _apply_special_cases(self, Doc doc) except -1
--- a/spacy/tokenizer.pxd
+++ b/spacy/tokenizer.pxd
@@ -23,8 +23,6 @@ cdef class Tokenizer:
     cdef object _infix_finditer
     cdef object _rules
     cdef PhraseMatcher _special_matcher
-    cdef int _property_init_count  # TODO: unused, remove in v3.1
-    cdef int _property_init_max    # TODO: unused, remove in v3.1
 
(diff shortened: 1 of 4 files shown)


honnibal:
Careful here: This will change the struct size of the tokenizer, and so will force recompilation of it and anything that cimports it. So we definitely couldn't make this change in a patch release.

adrianeboyd:
I thought that our policy was that we could make changes like this across minor versions. Do you want to handle this differently in the future?

honnibal:
Yes, we can make changes like this across minor versions. However, if there's little to no inconvenience to retaining the object size, it seems reasonable to do that. I see it as a trade-off.

It's fine if you decide "nah let's just remove them". I did want to point out the object size issue though, because it's not at all intuitive, and the CI doesn't really show the problem.

I definitely wouldn't suggest a strict policy of never changing the object sizes in minor versions, that would be really difficult.

adrianeboyd:
Okay, I agree that there's no real harm in waiting for v4, but we'll probably forget again...

polm:
I would be in favor of removing this sooner rather than later. Even observing semver, the more we put it off / less often we make changes like this the harder it becomes to make them and the more trouble we cause when we do change it (since downstream users are less used to it). Additionally, we removed members like this on a minor version before but as far as I'm aware the only negative side effects were with third-party libs doing something odd with serialization, and that may have been unrelated (#8264).

Separately from that, since it seems like `data_dir` in the Vocab is unused, that's another C member that could need removing. Should I add that to this PR?

adrianeboyd:
It does cause some minor issues for things like my `custom-cython-tokenizer` because there's no way to build a v3-compatible wheel.

polm:
OK, I think for now the solution is to close this, mark unused fields in source (with `TODO` or something), and remove on a major change.

svlandeg:
Actually, I thought the consensus was to rename them from now (see Matt's comment earlier). That will ensure they aren't getting used anymore, up until the point we do decide to remove them when the struct needs other changes.

polm:
Ah OK, I misremembered. 

I'll go ahead and add `data_dir`, since it's a similar case.

polm:
OK, I renamed `data_dir` to `_unused_object`. Wasn't sure if I could safely remove it from pickle-related code.

svlandeg (approved):
Agreed that this is probably safest!

PR #9155, opened 2021-09-06
Ignore prefix in suffix matches

adrianeboyd:
## Description

Ignore the currently matched prefix when looking for suffix matches in the tokenizer. Otherwise a lookbehind in the suffix pattern may match incorrectly due the presence of the prefix in the token string.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/tokenizer/test_tokenizer.py
+++ b/spacy/tests/tokenizer/test_tokenizer.py
@@ -2,7 +2,7 @@ import pytest
 import re
 from spacy.vocab import Vocab
 from spacy.tokenizer import Tokenizer
-from spacy.util import ensure_path
+from spacy.util import ensure_path, compile_prefix_regex, compile_suffix_regex
 from spacy.lang.en import English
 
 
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Evaluation:

* accuracy:
   * f-scores on the default tokenizer settings for ~80 UD corpora for 43 languages are identical
   * no differences in tokenization on 20M English tokens vs. `develop` (although there are some differences in `°F/C.` vs. `master` that I haven't figured out)

* speed: no noticeable difference in my low-tech wall-clock speed tests for the same UD corpora or on 20M English tokens

adrianeboyd:
The `°F.` case is pretty frustrating here. Currently with `master` you get `° F .` and with this you get `° F.` due to the middle-initial-type exceptions, which is even worse. There are no occurrences of `°` in OntoNotes and how it's tokenized across UD corpora varies a lot. (Off the top of my head I would have thought you would want `°F .`)

Additionally, there are several languages that go to various lengths to take `°` out of prefixes because this isn't the tokenization that they want.

For a specific case you could move `°[FCK].` to a tokenizer exception pretty easily, but doing this in the top-level exceptions would potentially screw up a lot of language-specific customizations, and a lot of languages have just copied this suffix regex into their settings.

I haven't tried to look for language-specific settings that include prefixes in lookbehinds.

PR #9167, opened 2021-09-07
🏷 Add Mypy check to CI and ignore all existing Mypy errors

connorbrinton:
## Description

These changes modify SpaCy's CI pipeline to automatically run Mypy on every change. This will help to ensure that static type annotations are respected in other parts of the codebase and suitable for consumption by library users.

In order to introduce this new CI check, all existing Mypy errors have been ignored using the following script:

```python
import re
import subprocess
from pathlib import Path
from typing import List

import typer


# Destructure lines into (i) code, (ii) existing `# type: ignore` directives
# (if any) and (iii) trailing comments (if any)
DIRECTIVE_RE = re.compile(
    r"""(?x)  # Verbose-mode (ignore whitespace and allow comments)
    ^  # We always want to capture beginning at the start of the line
    (?P<code>.*?)  # Everything before trailing comments
    (?P<type_ignore_directive>  # Possible existing `# type: ignore` directive
        \ {1,2}  # Preceded by one or two spaces
        \#\ type:\ ignore\[(?P<error_codes>[^\]]+)\]
    )?
    (?P<trailing_comments>  # Other trailing comments
        \ {1,2}  # Preceded by one or two spaces
        \#\ .*
    )?
    $  # We always want to capture the entire line
    """
)
OLD_RE = """
(?P<directives>  # Directives may occur in any order, so we have a repeating group here
    (?P<separator>\ {1,2})  # Directives may be preceded by one or two spaces
    (?P<directive>
        # Type ignore directives
        (?P<type_ignore_directive>)
        # NoQA directives
        | (?P<noqa_directive>\#\ noqa:[^#]+)
    )
)+
"""

ERROR_RE = re.compile(r"(?P<filepath>.*?):(?P<line_number>\d+): error: .* \[(?P<error_code>[\w\-]+)\]")

def main(
    files: List[Path] = typer.Argument(
        ...,
        exists=True,
        file_okay=True,
        dir_okay=True,
        readable=True,
    )
) -> None:
    """Execute Mypy on the given files and add `# type: ignore[error-code]` directives"""
    # Execute Mypy on the given files
    completed_process = subprocess.run(
        (
            "python",
            "-m",
            "mypy",
            "--show-error-codes",
            *[str(file) for file in files],
        ),
        capture_output=True,
    )

    # Extract structured representation of each error
    for match_groups in ERROR_RE.findall(completed_process.stdout.decode()):
        filepath, line_number, error_code = match_groups

        # Apply # type: ignore[code] directive to the file
        with open(filepath, "r") as source:
            # Read all lines
            lines = source.read().split("\n")

        # Look up the line to modify
        line = lines[int(line_number) - 1]

        # Check for existing directive and corresponding codes
        existing_codes = set()
        trailing_comments = None
        directive_match = DIRECTIVE_RE.match(line)
        if directive_match:
            code = directive_match.group("code")
            existing_codes_str = directive_match.group("error_codes")
            if existing_codes_str:
                existing_codes = set(existing_codes_str.split(", "))
            trailing_comments = directive_match.group("trailing_comments")

        # Add detected code
        error_codes = sorted(existing_codes | {error_code})

        # Build error codes string
        error_codes_str = ", ".join(error_codes)

        # Collect line components
        line_components = [
            code,
            f"# type: ignore[{error_codes_str}]",
        ]
        if trailing_comments:
            line_components.append(trailing_comments)

        # Rebuild line from components
        line = "  ".join(line_components)

        # Store updated line
        lines[int(line_number) - 1] = line

        # Rewrite file
        with open(filepath, "w") as sink:
            sink.write("\n".join(lines))


if __name__ == "__main__":
    typer.run(main)
```

This script may need to be executed multiple times (usually just two) to ignore all errors (since some errors only show up once others have been ignored).

Unfortunately, this PR changes a quite large number of lines. I'm happy to discuss other approaches to integrating Mypy into SpaCy's CI pipeline if this way is too extreme 😅

### Types of change

* CI pipeline change
* `# type: ignore[code]` directives added

## Checklist


connorbrinton:
Looks like SpaCy produces different errors when running using Python 3.7 and 3.8, I'll see if I can add new `# type: ignore` directives for those cases real quick...

connorbrinton:
`# type: ignore` directives have been added for 3.7 and 3.8, now I just need to modify my script to run on 3.6 😅 

Or manually add the `# type: ignore` directives 🤷‍♂️

adrianeboyd:
Thanks for your work on this!

I don't think we'd want to merge this PR as-is, but it's a good starting point to fix the easy cases, which are hopefully the majority. I think a few `types-` libraries are missing, too, it looks like `types-mock` and `types-requests`, which need to be added to `requirements.txt`, and then all the `[import]` cases are fixed. I'll just go ahead and do this...

Locally I get an additional error in `spacy/training/loggers.py` and I'm not sure what the environment difference is between my venv and the CI, so that's a bit confusing.

connorbrinton:
Two issues are currently causing issues with the tests:
* The new `wandb_logger` function is a redefinition of an earlier function. This will be fixed with a `# type: ignore` directive` shortly.
* The way that SpaCy imports `Literal` from `typing` or `typing_extensions` based on availability doesn't work with Mypy. On earlier Python versions, the `Literal` type in `compat.py` is inferred to be `Any`, which causes the "Overloaded function signature 2 will never be matched" error. I'm investigating how `typing_extensions` works to figure out the best way to fix this.
  
  It's possible that the `typing_extensions` module is supposed to be used differently. I'm trying to figure out if we should *always* import `Literal` from `typing_extensions` regardless of Python version, or if `typing_extensions` has a trick that allows its conditional `Literal` declaration to work regardless of Python version 🤔

adrianeboyd:
I think we should rename the method in the logger case. These are going to move to a separate `spacy-loggers` package in v3.2 (where the names are already corrected), but this was a mistake in the PR.

We eventually want to drop `typing_extensions` as a dependency, so I'd be hesitant to use it in general. There are several libraries like this that combine backport + experimental features in a way that makes them frustrating as dependencies where we really only want a backport. Maybe we can figure out another solution?

svlandeg:
Hey @connorbrinton, thanks so much for all your work on this! It's very much appreciated.

I thought I'd have a stab at trying to fix some of the remaining problems, I'll push directly to your branch if that's OK, but feel free to review here!

svlandeg (commented):
@ines : Would you agree with [this change](https://github.com/explosion/spaCy/pull/9167/commits/eda95db09a563a5b5d2a69ebb2ffc71d32d421ee)? `mypy` doesn't like subclassing of `Language.Defaults` and I think it's actually clearer by specifying `BaseDefaults` as base type?

ines:
@svlandeg No idea why we did it this way but yeah, I don't see any problem with using `BaseDefaults` as the type here 👍

svlandeg:
I'm struggling with `VietnameseTokenizer` and `JapaneseTokenizer`. I thought that if I'd make 
```
DummyTokenizer.from_bytes(...) -> Type["DummyTokenizer"]
```
and 
```
JapaneseTokenizer.from_bytes(...) -> "JapaneseTokenizer"
```
that it would work? But it doesn't :'( 
I'm getting
```
spacy\lang\ja\__init__.py:148: error: Return type "JapaneseTokenizer" of "from_bytes" incompatible with return type "Type[DummyTokenizer]" in supertype "Dum
myTokenizer"
```
so I'm probably still not understanding the `Type[]` stuff well enough?

svlandeg:
Also @connorbrinton : at this point it probably makes sense to merge your edits from https://github.com/explosion/spaCy/pull/8773 into this PR directly?

connorbrinton:
> Also @connorbrinton : at this point it probably makes sense to merge your edits from #8773 into this PR directly? We'd still have the old PR for reference for that specific bit. I realise this is becoming a bit of a monolithic PR, but then again a million smaller PRs wouldn't be much better, either.

Yep! :+1: I actually already merged them in with [this commit](https://github.com/explosion/spaCy/pull/9167/commits/58359ca922c902f753ba8f5699041bfcf8081d2f). I didn't pull in the regression test case since the new Mypy check already tests for similar usages within the SpaCy codebase (fixing the type annotations cut out 11 `# type: ignore` directives).

> I'm struggling with VietnameseTokenizer and JapaneseTokenizer. ... so I'm probably still not understanding the Type[] stuff well enough?

Yeah, `Type`s are a bit confusing. Annotating a function parameter or a return type as `"JapaneseTokenizer"` indicates that you expect an _instance_ of the `JapaneseTokenizer` class as the value in that location. But sometimes you don't expect an instance, you expect a _class_ (or "type"). As one example, [the `get_lang_class` function](https://github.com/explosion/spaCy/pull/9167/commits/2fb1d1b245a812bd88b7595f7fdb09dddcb822ea#diff-edf9a5728c2b543e15658f4ab1e7ca86a63b0f18d3751b3e0c08c8f9a147083bR258) returns the class representing the passed language code.

By annotating the return type of `get_lang_class` as `Type[Language]`, we can indicate to function callers that they should expect a class representing that language, such as `English`, `Spanish`, etc. The caller can then instantiate an instance of that class with `cls()`.

What you're trying to do isn't actually solved with `Type`. The idea you want to communicate is that "Calling `from_bytes` on an instance should always return an instance of the same type." There are a couple of ways to do this, but since we're always redefining `from_bytes` in each subclass, the best way to do this is to use a more specific return type (i.e., `-> "JapaneseTokenizer"`) in each subclass.

(The other way to do this if subclasses don't redefine the method is using a type variable annotating `self`, [as discussed in the Mypy docs here](https://mypy.readthedocs.io/en/stable/more_types.html#advanced-uses-of-self-types))

Alrighty, let's annotate `DummyTokenizer.from_bytes`, since `JapaneseTokenizer.from_bytes` is already annotated to return a `JapaneseTokenizer`:
```python
def from_bytes(self, _bytes_data: bytes, **kwargs) -> "DummyTokenizer":
```

Aw man, we got some new errors!
```
spacy/lang/vi/__init__.py:136: error: Signature of "from_bytes" incompatible with supertype "DummyTokenizer"  [override]
spacy/lang/ja/__init__.py:148: error: Signature of "from_bytes" incompatible with supertype "DummyTokenizer"  [override]
```

Mypy is trying to enforce here the Liskov substitution principle, which requires that we can use a subtype instance anywhere we can use a supertype instance. In this case, we need to be able to call `from_bytes` the same way for any `DummyTokenizer` subclass (Mypy expects subclasses to be subtypes of their non-mixin ancestor classes) and the return type must be compatible.

In this case, the return type _is_ compatible, since specializing a return type (using a more specific return type) in a subclass function is fine. The problem is that if we have a `DummyTokenizer`, we can technically do:
```python
tokenizer.from_bytes(_bytes_data=b"...")
```

But if we have a `JapaneseTokenizer`, we can't do the same thing! This is due to the fact that the _names_ of the function parameters have changed, and they're not positional-only arguments.

This can be fixed by renaming the `_bytes_data` parameter to `DummyTokenizer.from_bytes` to just `data` 😄

I just pushed a commit with this change and a couple of other changes to `DummyTokenizer` serialization/deserialization methods to get rid of some similar typing errors :+1:

PR #9175, opened 2021-09-09
Update forte in universe.json

mylibrar:
## Description

Update `pip` and `code_example` of `forte` in `universe.json` to accommodate recent changes in `forte.spacy` package.

### Types of change

Documentation/Website

## Checklist


polm:
Thanks for the update! Is the change to `fortex.spacy` intentional? I tried to run the sample code but got an error about no module named fortex existing. Also you seem to be adding a StaveProcessor twice?

mylibrar:
Thanks for the feedback! The `fortex.spacy` update is intentional and is related to our recent change in `forte.spacy` package. We have released a new version `forte.spacy==0.0.2` which should be compatible with the current sample code. And I actually add two processors, i.e., `SpacyProcessor` and `StaveProcessor`. Sorry for the confusion here.

polm:
OK, thanks for the update, I confirmed the code works. And you're right about the processors, sorry for the confusion.

One minor point, but when opening the browser window it might be a good idea to have more robust output indicating a browser has been opened, take a look at what Streamlit does. 

Anyway, thanks for updating your example!

PR #9180, opened 2021-09-09
Correct parser.py use_upper param info

j-frei:
# Update use_upper param info in parser.py

## Description
Dear SpaCy team,

I think there was a mistake in the documentation: In (e.g.) NER, the parser uses an additional network to predict the action/state scores in the transition-based parsing stage when using a small, non-transformer network.

Kind regards,
Johann

### Types of change
Documentation update

## Checklist


adrianeboyd:
Thanks for the correction!

Issue #9129, opened 2021-09-03
Displacy vizualization for entity linking


shigapov:
Displacy can [vizualize the enitity recognizer](https://spacy.io/usage/visualizers#ent). It would be nice to have vizualization for entity linking pipelines as well. So that the links to entities in a knowledge base are also shown.

polm:
Thanks for the suggestion!

shigapov:
Let's add those links using "manual" option in displacy.

In [Rendering data manually](https://spacy.io/usage/visualizers#manual-usage) there is an example:

```python
ex = [{"text": "But Google is starting from behind.",
       "ents": [{"start": 4, "end": 10, "label": "ORG"}],
       "title": None}]
html = displacy.render(ex, style="ent", manual=True)
```

I propose to add two more keys to "ents" ("kb_id" and "kb_url"):
```python
ex = [{"text": "But Google is starting from behind.",
       "ents": [{"start": 4, "end": 10, "label": "ORG", "kb_id": "Q95", "kb_url": "https://www.wikidata.org/wiki/Q95"}],
       "title": None}]
html = displacy.render(ex, style="ent", manual=True)
```

Then in [templates.py](https://github.com/explosion/spaCy/blob/master/spacy/displacy/templates.py) we need to add one more line to "TPL_ENT":
```python
TPL_ENT = """
<mark class="entity" style="background: {bg}; padding: 0.45em 0.6em; margin: 0 0.25em; line-height: 1; border-radius: 0.35em;">
    {text}
    <span style="font-size: 0.8em; font-weight: bold; line-height: 1; border-radius: 0.35em; vertical-align: middle; margin-left: 0.5rem">{label}</span>
    <a style="text-decoration: none; color: black; font-weight: bold" href="{kb_url}">{kb_id}</a>
</mark>
"""
```

And in [render.py](https://github.com/explosion/spaCy/blob/master/spacy/displacy/render.py#L302-L340) we need to add two more lines (to specify `kb_url` and `kb_id`) and replace `ent_settings = {"label": label, "text": entity, "bg": color}`:
```python
kb_id = str(span.get("kb_id") or "")
kb_url = str(span.get("kb_url") or "")
ent_settings = {"label": label, "text": entity, "bg": color, "kb_id": kb_id, "kb_url": kb_url}
```

If that's fine, I could make PR.

shigapov:
Then it looks like
[...]

PR #9199, opened 2021-09-13
Add NEL visualisation (links to entities in a knowledge base)

shigapov:
## Description
Current NER visualiser cannot show links to entities in a knowledge base. This PR adds NEL visualisation using `manual=True` option in displaCy.

### Types of change

This is enhancement, see [issue 9129](https://github.com/explosion/spaCy/issues/9129).

## Checklist


spacy/displacy/render.py
```diff
@@ -314,6 +314,8 @@ def render_ents(
             label = span["label"]
             start = span["start"]
             end = span["end"]
+            kb_id = str(span.get("kb_id") or "")
+            kb_url = str(span.get("kb_url") or "")
```
svlandeg, line 318: This is an extreme nitpick (sorry) but I'd rewrite this as:
```suggestion
            kb_id = str(span.get("kb_id")) or ""
            kb_url = str(span.get("kb_url")) or ""
```

spacy/displacy/templates.py
```diff
@@ -52,6 +52,7 @@
 <mark class="entity" style="background: {bg}; padding: 0.45em 0.6em; margin: 0 0.25em; line-height: 1; border-radius: 0.35em;">
     {text}
     <span style="font-size: 0.8em; font-weight: bold; line-height: 1; border-radius: 0.35em; vertical-align: middle; margin-left: 0.5rem">{label}</span>
+    <a style="text-decoration: none; color: black; font-weight: bold" href="{kb_url}">{kb_id}</a>
```
svlandeg, line 55: I was wondering - if we include the KB ID into the `span` element, the formatting of the label and the KB ID will be exactly the same. That seems better?
```suggestion
    <span style="font-size: 0.8em; font-weight: bold; line-height: 1; border-radius: 0.35em; vertical-align: middle; margin-left: 0.5rem">{label}
    <a style="text-decoration: none; color: black; font-weight: bold" href="{kb_url}">{kb_id}</a>
    </span>
```

svlandeg: The only disadvantage I see is that there is that there is no "deliminter" between the label and KB ID, but that would kind of complicate the code further. e.g. we could insert a "-" between label and ID, but then only if both are available...

svlandeg (commented):
Thanks for the PR, much appreciated! I have some small comments, but I think this looks good in general. I'll also ping Ines to see whether she might want to review, too :-)

spacy/displacy/render.py
shigapov, line 318: I am not sure. The get-method for dictionary returns None if the key is not in the dictionary, so `str(None) or ""` returns `'None'`. But we want an empty string if "kb_id" and "kb_url" are not in the dictionary.

Just learned that get-method for dictionary has a second argument, so we can simply write:
```suggestion
            kb_id = span.get("kb_id", "")
            kb_url = span.get("kb_url", "")
```

spacy/displacy/templates.py
shigapov: KB ID in the `span` element indeed looks better and more compact!

spacy/displacy/render.py
svlandeg: Yep you're right, that's better!

spacy/displacy/templates.py
```diff
@@ -51,7 +51,9 @@
 TPL_ENT = """
 <mark class="entity" style="background: {bg}; padding: 0.45em 0.6em; margin: 0 0.25em; line-height: 1; border-radius: 0.35em;">
     {text}
-    <span style="font-size: 0.8em; font-weight: bold; line-height: 1; border-radius: 0.35em; vertical-align: middle; margin-left: 0.5rem">{label}</span>
+    <span style="font-size: 0.8em; font-weight: bold; line-height: 1; border-radius: 0.35em; vertical-align: middle; margin-left: 0.5rem">{label}
+    <a style="text-decoration: none; color: black; font-weight: bold" href="{kb_url}">{kb_id}</a>
```
ines, line 55: The problem I see with this is that the markup here will be added to _all_ visualizations, even those that don't have a `kb_id` set (which is very common because not all users train an entity linker). So if we want to include this, we probably want to construct the link outside of the template and pass it in if a `kb_id` is available – otherwise, the variable receives an empty string.

shigapov, line 55: Like this?
```suggestion
    {kb_link}
```

spacy/displacy/render.py
```diff
@@ -323,7 +325,7 @@ def render_ents(
                     markup += "</br>"
             if self.ents is None or label.upper() in self.ents:
                 color = self.colors.get(label.upper(), self.default_color)
-                ent_settings = {"label": label, "text": entity, "bg": color}
+                ent_settings = {"label": label, "text": entity, "bg": color, "kb_id": kb_id, "kb_url": kb_url}
```
shigapov, line 328:
```suggestion
                ent_settings = {"label": label, "text": entity, "bg": color, "kb_link": kb_link}
```

spacy/displacy/render.py
```diff
@@ -314,6 +314,8 @@ def render_ents(
             label = span["label"]
             start = span["start"]
             end = span["end"]
+            kb_id = span.get("kb_id", "")
+            kb_url = span.get("kb_url", "")
```
shigapov, line 318:
```suggestion
            kb_url = span.get("kb_url", "")
            if kb_id:
                kb_link = """<a style="text-decoration: none; color: black; font-weight: bold" href="{}">{}</a>""".format(kb_url, kb_id)
            else:
                kb_link = ""
```

spacy/displacy/templates.py
svlandeg: I did test this with an empty string and it'd basically wouldn't be displayed, it would just be an empty `<a>` element, so I'm not sure it's worth the hassle? But yea it might be cleaner to have this in a conditional.

ines: We shouldn't be generating empty links as part of the default markup, though – it adds a lot of bloat for no reason, is potentially confusing and also produces unsemantic HTML.

PR #9202, opened 2021-09-13
feat: add `spacy.WandbLogger.v3` with optional `run_name` and `entity` parameters

Joozty:
## Description

Just adding `run_name` to WandbLogger. When you run multiple trainings sometimes it's better to assign them own name so it's easier to distinguish them in Wandb Dashboard

https://docs.wandb.ai/guides/track/launch#can-i-just-set-the-run-name-to-the-run-id

## Checklist

### DIFF
--- a/website/docs/api/top-level.md
+++ b/website/docs/api/top-level.md
@@ -507,6 +507,8 @@ remain in the config file stored on your local system.
 | `remove_config_values` | A list of values to include from the config before it is uploaded to W&B (default: empty). ~~List[str]~~                              |
 | `model_log_interval`   | Steps to wait between logging model checkpoints to W&B dasboard (default: None). ~~Optional[int]~~                                    |
 | `log_dataset_dir`      | Directory containing dataset to be logged and versioned as W&B artifact (default: None). ~~Optional[str]~~                            |
+| `run_name`             | The name of the run. If you don't specify a run_name, the name will be created by wandb library. (default: None ). ~~Optional[str]~~                            |
+| `entity`               | An entity is a username or team name where you're sending runs. If you don't specify an entity, the run will be sent to your default entity, which is usually your username.  (default: None). ~~Optional[str]~~                            |
 
 <Project id="integrations/wandb">
 
--- a/website/docs/api/top-level.md
+++ b/website/docs/api/top-level.md
@@ -462,7 +462,7 @@ start decreasing across epochs.
 
  </Accordion>
 
-#### spacy.WandbLogger.v2 {#WandbLogger tag="registered function"}
+#### spacy.WandbLogger.v3 {#WandbLogger tag="registered function"}
 
 > #### Installation
 >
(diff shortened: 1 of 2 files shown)


svlandeg (changes requested):
Thanks for the contribution! We will definitely consider merging this, but we'd want to bump the logger's version number to v3 (a copy of v2 can stay in the code for now, we'll clean it up later). And perhaps then we could also include the `entity_name` addition from https://github.com/explosion/spaCy/pull/8832?

Joozty:
Is bumping logger necessary? `run_name` is an optional parameter and it doesn't break anything. `entity_name` can be an optional parameter as well according to [docs](https://docs.wandb.ai/ref/python/init).

svlandeg:
It wouldn't break backwards compatibility if we didn't bump, but it would still potentially create problems: a config with a v2 in it, created with a new spaCy version and specifying the new parameter(s), suddenly wouldn't run anymore on an older version of spaCy that didn't have the new parameters yet, it would raise an error about the parameter being unavailable, which isn't what you expect if you've used the exact same version of the logger.

Joozty:
Oh I see, thank 🙂 Should I create v3 then?

svlandeg:
Yes, please :-) for now you can just copy-paste the code and keep one copy as v2, and make your changes in the other copy as v3.

website/docs/api/top-level.md
```diff
@@ -507,6 +507,8 @@ remain in the config file stored on your local system.
 | `remove_config_values` | A list of values to include from the config before it is uploaded to W&B (default: empty). ~~List[str]~~                              |
 | `model_log_interval`   | Steps to wait between logging model checkpoints to W&B dasboard (default: None). ~~Optional[int]~~                                    |
 | `log_dataset_dir`      | Directory containing dataset to be logged and versioned as W&B artifact (default: None). ~~Optional[str]~~                            |
+| `run_name`             | The name of the run. If you don't specify a run_name, the name will be created by wandb library. (default: None ). ~~Optional[str]~~                            |
+| `entity`               | An entity is a username or team name where you're sending runs. If you don't specify an entity, the run will be sent to your default entity, which is usually your username.  (default: None). ~~Optional[str]~~                            |
```
Joozty, line 511: I copied these bits for [official documentation](https://docs.wandb.ai/ref/python/init). Maybe I should somehow reference it? Not sure what's the best way to do it. 🤔

svlandeg: I think this is fine!

svlandeg (approved):

adrianeboyd:
I think we should accept this PR in the upcoming [`spacy-loggers`](https://github.com/explosion/spacy-loggers) repo rather than merging it here at all. It looks like this has already been copied over, but I haven't double-checked the code/tests.

svlandeg:
We've discussed a little bit more internally, and we'll still merge this for v.3.1.3.
From 3.2 onwards, we'll be using the separate repo `spacy-loggers` as Adriane noted, and from then on the code will be moved to there.

Thanks again for this contribution!

PR #9226, opened 2021-09-16
Don't serialize user data in DocBin if not saving it (fix #9190)

### LINKED ISSUE #9190: Serialization: store_user_data is ignored in DocBin constructor

When I try to create a `DocBin` out of documents that have `Spans` in custom attributes, I get an exception, even though I set `store_user_data = False`. The exception is thrown in `DocBin.add()` function, line 113 of `_serialize.py` file:
```
self.user_data.append(srsly.msgpack_dumps(doc.user_data))
``` 
The exception is thrown because `srsly.msgpack_dumps(doc.user_data)` cannot serialize `Spans`. By mistake, it is called even when `self.store_user_data` is set to `False`

## The following code reproduces the behavior:
```
import spacy
from spacy.tokens import Doc
from spacy.tokens import DocBin
Doc.set_extension(name = 'test', default = None, force = True)

text = "Hello, world!"
doc = nlp(text)
doc._.test = doc[0].sent # Assign a span to the doc's custom attribute
docs = [doc]

## Create DocBin
attributes = ("ORTH", "TAG", "HEAD", "DEP", "ENT_IOB", "ENT_TYPE", "ENT_ID", "ENT_KB_ID", "LEMMA", "MORPH", "POS", "LOWER", "SENT_START", "SENT_END", "IDX", 'IS_PUNCT', 'LIKE_NUM', 'IS_BRACKET', "IS_LEFT_PUNCT", "IS_RIGHT_PUNCT")
doc_bin = DocBin(attrs = attributes, store_user_data = False, docs = docs)

```

## The error
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-5796ac30343d> in <module>
      1 ## Store
      2 attributes = ("ORTH", "TAG", "HEAD", "DEP", "ENT_IOB", "ENT_TYPE", "ENT_ID", "ENT_KB_ID", "LEMMA", "MORPH", "POS", "LOWER", "SENT_START", "SENT_END", "IDX", 'IS_PUNCT', 'LIKE_NUM', 'IS_BRACKET', "IS_LEFT_PUNCT", "IS_RIGHT_PUNCT")
----> 3 doc_bin = DocBin(attrs = attributes, store_user_data = False, docs = docs)
      4 #serialized_docs = doc_bin.to_bytes()

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/spacy/tokens/_serialize.py in __init__(self, attrs, store_user_data, docs)
     78         self.store_user_data = store_user_data
     79         for doc in docs:
---> 80             self.add(doc)
     81 
     82     def __len__(self) -> int:

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/spacy/tokens/_serialize.py in add(self, doc)
    109             self.strings.add(token.ent_kb_id_)
    110         self.cats.append(doc.cats)
--> 111         self.user_data.append(srsly.msgpack_dumps(doc.user_data))
    112         self.span_groups.append(doc.spans.to_bytes())
    113         for key, group in doc.spans.items():

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/srsly/_msgpack_api.py in msgpack_dumps(data)
     12     RETURNS (bytes): The serialized bytes.
     13     """
---> 14     return msgpack.dumps(data, use_bin_type=True)
     15 
     16 

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/srsly/msgpack/__init__.py in packb(o, **kwargs)
     53     Pack an object and return the packed bytes.
     54     """
---> 55     return Packer(**kwargs).pack(o)
     56 
     57 

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/srsly/msgpack/_packer.pyx in srsly.msgpack._packer.Packer.pack()

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/srsly/msgpack/_packer.pyx in srsly.msgpack._packer.Packer.pack()

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/srsly/msgpack/_packer.pyx in srsly.msgpack._packer.Packer.pack()

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/srsly/msgpack/_packer.pyx in srsly.msgpack._packer.Packer._pack()

~/python-virtualenv/eon-spacy-3/lib/python3.7/site-packages/srsly/msgpack/_packer.pyx in srsly.msgpack._packer.Packer._pack()

TypeError: can not serialize 'spacy.tokens.span.Span' object
```
## Your Environment

- **spaCy version:** 3.0.6
- **Platform:** Darwin-20.3.0-x86_64-i386-64bit
- **Python version:** 3.7.9
- **Pipelines:** en_core_web_lg (3.0.0), en_core_web_md (3.0.0), en_core_web_sm (3.0.0)

polm:
## Description

If you create a DocBin with `store_user_data = False`, the data won't be included in `to_bytes`, but it will still be serialized as soon as it's added to the DocBin. If you have a non-serializable item in user data, this means you can get an error about that even though you don't want to serialize it, which is confusing. See #9190 

This just doesn't save user data if `store_user_data` is `False`. Another option would be to save empty user data. That might be better if someone changes the value on an existing DocBin, but that sounds like a terrible idea in general.

### Types of change

Convenience fix.

## Checklist

### DIFF
--- a/spacy/tests/training/test_training.py
+++ b/spacy/tests/training/test_training.py
@@ -524,6 +524,30 @@ def test_roundtrip_docs_to_docbin(doc):
     assert cats["TRAVEL"] == reloaded_example.reference.cats["TRAVEL"]
     assert cats["BAKING"] == reloaded_example.reference.cats["BAKING"]
 
+def test_docbin_user_data_serialized(doc):
+    doc.user_data["check"] = True
+    nlp = English()
+
+    with make_tempdir() as tmpdir:
+        output_file = tmpdir / "userdata.spacy"
+        DocBin(docs=[doc], store_user_data=True).to_disk(output_file)
+        reloaded_docs = DocBin().from_disk(output_file).get_docs(nlp.vocab)
+        reloaded_doc = list(reloaded_docs)[0]
+
+    assert reloaded_doc.user_data["check"] == True
+
+def test_docbin_user_data_not_serialized(doc):
+    # this isn't serializable, but that shouldn't cause an error
+    doc.user_data["check"] = set()
+    nlp = English()
+
+    with make_tempdir() as tmpdir:
+        output_file = tmpdir / "userdata.spacy"
+        DocBin(docs=[doc], store_user_data=False).to_disk(output_file)
+        reloaded_docs = DocBin().from_disk(output_file).get_docs(nlp.vocab)
+        reloaded_doc = list(reloaded_docs)[0]
+
+    assert "check" not in reloaded_doc.user_data
 
 @pytest.mark.parametrize(
     "tokens_a,tokens_b,expected",
--- a/spacy/tokens/_serialize.py
+++ b/spacy/tokens/_serialize.py
@@ -110,7 +110,8 @@ class DocBin:
             self.strings.add(token.ent_kb_id_)
             self.strings.add(token.ent_id_)
         self.cats.append(doc.cats)
-        self.user_data.append(srsly.msgpack_dumps(doc.user_data))
+        if self.store_user_data:
+            self.user_data.append(srsly.msgpack_dumps(doc.user_data))
         self.span_groups.append(doc.spans.to_bytes())
         for key, group in doc.spans.items():
             for span in group:


svlandeg:
Hm, I see the problem. Up until now, the `DocBin` class has specifically been implemented so that `user_data` was in fact stored on the `Doc` objects, just not serialized to file with `to_bytes`/`to_disk` , as this comment also illustrates:
```
 def get_docs(self, vocab: Vocab) -> Iterator[Doc]:
        """Recover Doc objects from the annotations, using the given vocab.
        Note that the user data of each doc will be read (if available) and returned,
        regardless of the setting of 'self.store_user_data'.
        ...
```

I can definitely see how this causes issues if the data is not serializable and thus can't even be stored on the Doc. To make things worse, the setting is called `store_user_data`, not `serialize_user_data`...

What I worry about, is that this PR would change the behaviour for other users who might actually be using the `user_data` from the `doc_bin`. So I think in general we have three options:
- Leave the behaviour as-is and document more clearly how the internals work, and that `user_data` should be deleted before using the `DocBin` if the data is not `msgpack` serializable
- Change the internal behaviour of the `DocBin` to never store `user_data` (even internally) if `store_user_data` is `False`, which does feel like the more intuitive behaviour. But then this change should go to v.3.2. In this case, to support other use-cases, we could even consider creating a second flag `serialize_user_data` to cater for both behaviours.
- Attempt to store the `user_data` internally even if `store_user_data` is `False`, but catch the `Exception` and do nothing if the flag is `False` - reraise if `True` (+ with a clarification on how to change the flag)

polm:
I hadn't considered that someone might intentionally store data in the DocBin without wanting to serialize it, though the comment suggests the DocBin was designed with that use case in mind. I guess the question is, why? If you're not serializing things does the DocBin offer any advantage over a list? I'm not sure I understand the use case for the DocBin separately from serialization.

For that reason I think it would make sense to make things work like in this patch, and not add a flag for the other use case. The third option is best from a backwards compatibility point of view but just seems like a weird design.

svlandeg:
What bothers me is that there are no unit tests breaking, even though the functionality does change with this PR. I think it might make sense to clear up all different use-cases / various settings of this variable first, to get a better idea of the impact of this decision. Ideally we'd have more unit tests covering more cases, and we would more easily see what breaks and how.

PR #9248, opened 2021-09-20
Allow overriding vars in the project assets subcommand

danieldk:
## Description

`spacy project run` allows overriding arguments through the command-line (e.g. `--vars.gpu=-1`). Overriding variables is not supported by the `spacy project assets` subcommand. However, this would be handy when e.g. a data repository contains multiple corpora and one wants to fetch a different corpus based on the variable set.

This change makes the `project assets` subcommand accept variables to override as well, making the interface more similar to `project run`.

### Types of change

Enhancement

## Checklist

With regards to the last checkbox: from the current documentation I inferred that it would be possible to override variables with `project assets` as well, until I found out you can't 😅.

### DIFF
--- a/spacy/cli/project/assets.py
+++ b/spacy/cli/project/assets.py
@@ -1,18 +1,24 @@
-from typing import Optional
+from typing import Any, Dict, Optional
 from pathlib import Path
 from wasabi import msg
 import re
 import shutil
 import requests
+import typer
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Yep, this makes a lot of sense. Thanks!

PR #6386, opened 2020-11-13
Fix memory issues in Language.evaluate

adrianeboyd:
## Description

Reset annotation in predicted docs before evaluating and store all data in `examples`.

Fixes #6323.

### Types of change

Bug fix

## Checklist

- [x] I have submitted the spaCy Contributor Agreement.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #9305, opened 2021-09-27
Restore tokenization timing in Language.evaluate

adrianeboyd:
## Description

Restore tokenization timing steps that were accidentally removed in #6765.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1369,6 +1369,9 @@ class Language:
             scorer = Scorer(**kwargs)
         # reset annotation in predicted docs and time tokenization
         start_time = timer()
+        # this is purely for timing
+        for eg in examples:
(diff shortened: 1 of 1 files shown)


svlandeg:
Ah, oops, should have caught that while reviewing the original PR!

jpribyl:
@adrianeboyd can you explain why this is required? I see it in the original PR here https://github.com/explosion/spaCy/pull/6386 

But I must be missing something because I'm not seeing what would break without it (guess I could remove it locally and find out :smile: )

adrianeboyd:
Nothing breaks or changes on the annotation performance metrics, but the `SPEED` evaluation isn't counting the time it takes to tokenize because it's starting from tokenized docs in `example.predicted` rather than raw text.

PR #9321, opened 2021-09-28
Adding PhruzzMatcher to spaCy universe

mjvallone:
Adding PhruzzMatcher to spaCy universe

## Description

PhruzzMatcher is a combination of the RapidFuzz library with Spacy PhraseMatcher The goal of this component is to find matches when there were NO "perfect matches" due to typos or abbreviations between a Spacy doc and a list of phrases. To see more about Spacy Phrase Matcher go to https://spacy.io/usage/rule-based-matching#phrasematcher

### Types of change

New component to add in the spaCy universe.

## Checklist


polm:
Thanks for the submission, looks like a cool project! 

About the sample code - it doesn't really do anything, could you make an example that actually has output and shows what data your code sets? It would also be better to use `spacy.blank("es")` so the user doesn't have to download a model if the code doesn't depend on it.

Another thing is your Python package doesn't specify rapidfuzz as a dependency, so if you just `pip install phruzz_matcher` it doesn't work.

mjvallone:
@polm I fixed it, thanks for your suggestions! :)
Please feel free to test it again

polm:
Excellent, thanks for the changes!

lingvisa:
Is this a mature project?

polm:
Hey @lingvisa, Universe is intended as source of information, not as a list of verified components, and while we run a few basic checks before adding things we don't thoroughly vet projects. Making judgments about project maturity is ultimately up to you.

Issue #9263, opened 2021-09-21
Bug in DependencyMatcher--success of a match depends on history of how patterns are added.


mdc713:
This is either a bug or some very unexpected behavior. The results of DependencyMatcher seem to depend on the how exactly the rules are added--in what order, with what names, etc... Examples below.

## How to reproduce the behaviour
The initial set up is this.
```
from spacy.matcher import DependencyMatcher
import spacy

LID = "LEFT_ID"
RID = "RIGHT_ID"
RATTRS = "RIGHT_ATTRS"
RELOP = "REL_OP"
ORTH = "ORTH"
REGEX = "REGEX"
DEP = "DEP"

has_have = {RID: "have/has", RATTRS: {ORTH: {REGEX: "have|has|is"}}}
nsubj = {LID: "have/has", RELOP: ">", RID: "nsubj", RATTRS: {DEP: "nsubj"}}
pobj = {LID: "have/has", RELOP: ">>", RID: "pobj", RATTRS: {DEP: "pobj"}}
dobj = {LID: "have/has", RELOP: ">", RID: "dobj", RATTRS: {DEP: "dobj"}}
relcl = {LID: "have/has", RELOP: "<", RID: "relcl", RATTRS: {DEP: "relcl"}}
conj = {LID: "have/has", RELOP: ">>", RID: "conj", RATTRS: {DEP: "conj"}}

TEST_SENTENCE = "Product A and Product B have one processor, which is one more than Product C."
MODEL_NAME = "en_core_web_sm"
nlp = spacy.load(MODEL_NAME)
TEST_DOC = nlp(TEST_SENTENCE)
matcher = DependencyMatcher(nlp.vocab)
matcher.add('rule-1', [[has_have, nsubj, dobj]])
matcher.add('rule-2', [[has_have, pobj, dobj]])
matcher.add('rule-3', [[has_have, conj, dobj]])
matcher.add('rule-4', [[has_have, nsubj, relcl]])
matcher.add('rule-5', [[has_have, pobj, relcl]])
matcher.add('rule-6', [[has_have, conj, relcl]])
matcher(TEST_DOC)
```

which give the output:
```
[(507345881037203946, [5, 1, 7]),
 (13408152019079346362, [5, 15, 7]),
 (14359486885833749881, [5, 4, 7])]
```
Now things get weird. I try to add a new rule:
```
[...]

PR #9337, opened 2021-09-30
Fix Dependency Matcher Ordering Issue

polm:
## Description

This seems to fix at least some of the problems exposed in #9263, and includes a test that illustrates the issue.

Currently this handles this particular problem: if you have two rules with the same label, behavior is different adding them in one call to `DependencyMatcher.add` vs two calls. The one-call behavior is correct.

It's a little hard to tell from the issue if there are other problems, so this needs more work.

### Types of change

bug fix

## Checklist


polm:
I think the current fix works in this particular situation, but marking this as a draft until it's clear this is the only issue.

polm:
OK, this should be ready for review now. There was another issue, which was that if a rule was removed it wasn't removed from the internal Matcher instance, which could cause extra matches corresponding to rules that had been removed.

PR #9342, opened 2021-09-30
Allow IETF language codes, aliases, and close matches

rspeer:
## Description
spaCy's documentation and docstrings previously asked for "two-letter ISO codes" of languages, for example as the input to `spacy.blank()`, linking to Wikipedia's ISO 639-1 list. However:

- It would not accept the ISO 639-1 code `no` for Norwegian, only `nb`
- It would not accept ISO 639-1 codes that were aliases, such as `iw` for `he`
- Not all of the codes were two letters as documented. In one case (`lij`) it required a three-letter ISO 639-2 code, as that is the only code available for Ligurian, but in all other cases three-letter codes were forbidden.

It seems clear to me that what spaCy is going for is to use the modern generalization of ISO language codes that web browsers and Wikipedia use, where you prefer two-letter codes when they're available, but also allow three-letter codes. This generalization is standardized as [IETF language tags](https://www.w3.org/International/articles/language-tags/), in the standard named [BCP 47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt).

(A BCP is a "Best Current Practice", an updateable pointer to a non-obsolete RFC, so BCP 47 is RFC 5646.)

There are a lot of rules for understanding IETF language tags so that we can be tolerant of aliases, slightly different tag lists, and technically-distinct language codes that are close enough. I already implemented those rules in my [langcodes](https://github.com/rspeer/langcodes) package.

So in this PR, if a language is requested with `spacy.blank` that doesn't exactly match the name of an available language module, it will use the language matching from langcodes to find a matching language if possible. Some effects of this are:

- You can get `spacy.lang.nb.Norwegian` with the code `no`
- Aliases are allowed: `iw` gets `spacy.lang.he.Hebrew`
- Overlong ISO 639-2 codes are allowed: `fra` and `fre` both get `spacy.lang.fr.French`
- It can handle codes that specify their region: `en-UK` gets `spacy.lang.en.English`, and `pt-BR` gets `spacy.lang.pt.Portguese`
- It can handle codes that specify their script: `zh-Hans` (Simplified Chinese) gets `spacy.lang.zh.Chinese`, and `zh-Hant` raises an error because we don't yet have support that matches Traditional Chinese
- This gives us room to support Traditional Chinese in the future
- This one is mostly just to scratch my itch, but `spacy.blank('mul')` gets `spacy.lang.xx.MultiLanguage` -- I haven't changed spaCy's use of the code 'xx', but 'mul' is an IETF language code for this.

An integration that becomes possible because of this: if you choose to, you could guess language tags for multilingual text with fastText's langID model and pass the results as-is into `spacy.blank()` to get an appropriate pipeline for the language, though you'd still have to handle the cases where the language isn't supported by spaCy.

### Types of change
Enhancement

## Checklist


spacy/util.py
```diff
@@ -256,20 +258,85 @@ def lang_class_is_loaded(lang: str) -> bool:
...
+        code = modinfo.name
+        if code == 'xx':
+            # Temporarily make 'xx' into a valid language code
+            code = 'mul'
+        try:
+            language = langcodes.Language.get(code)
+            if language.is_valid():
+                possible_languages.append(code)
+        except langcodes.tag_parser.LanguageTagError:
+            pass
```
honnibal, line 304: Is there a way to do this with a conditional, rather than the try/except?

It's not _really_ a problem because we're only catching the specific library error, but I'd still prefer to have conditions expressed with conditionals, and only use try/except for exceptional branches.

spacy/util.py
```diff
@@ -256,20 +258,85 @@ def lang_class_is_loaded(lang: str) -> bool:
...
+
+    match, _dist = langcodes.closest_match(
+        lang, possible_languages, max_distance=10
+    )
```
honnibal, line 308: Will the 10 here always be fine, or could it come back to haunt us later?

spacy/util.py
```diff
@@ -256,20 +258,85 @@ def lang_class_is_loaded(lang: str) -> bool:
...
 
-    lang (str): Two-letter language code, e.g. 'en'.
+    lang (str): IETF language code, such as 'en'.
     RETURNS (Language): Language class.
     """
     # Check if language is registered / entry point is available
     if lang in registry.languages:
         return registry.languages.get(lang)
     else:
+        # Find the language in the spacy.lang subpackage
```
honnibal, line 328: Definitely support the comment here, thanks :)

honnibal:
Very much support the idea behind this, thanks!

spacy/util.py
rspeer: I admit that this is a wonky way to find out which languages we support. The problem is that if I'm iterating the modules within `spacy.lang`, I'm going to get things like `tokenizer_exceptions`. I can't ask `langcodes.Language.get('tokenizer_exceptions')` if it's valid, because it gets a parse error first.

A way to not encounter this exceptional case would be, perhaps, to put a list of supported languages in `spacy/lang/__init__.py`.

rspeer: ...and that would also cope with the possibility where a new submodule of `spacy.lang` just happens to parse as a language tag.

spacy/util.py
rspeer: I've used a distance of 10 as my threshold for matching for a while, but CLDR changes its language distances a lot, so I can't promise anything.

The most distant match this currently allows is from a deprecated code for Montenegrin (`cnr`) to Croatian (`hr`), a distance of 9. I currently want to rely on matching Bosnian to Croatian, which is a distance of 8.

I can see a vague reason to be worried about a distance of 10: in much earlier versions of CLDR, they would give a distance of 10 from a language to the world language it's being colonized by, like from Wolof to French or from Tamil to English. That was clearly unexpected and unreasonable. They have greatly increased those language distances now (to 34 and 44 respectively). I don't think they'll roll back to being unreasonable, but I could change the maximum to 9 just to be sure?

rspeer:
I should have waited to see all the tests. We have a test expecting an `ImportError` that now gets a non-spaCy-coded `LanguageTagError`, when trying to get `spacy.blank('fjsfijsdof')`. Should I catch that error and turn it into an error with a proper spaCy error code?

spacy/util.py
rspeer: I iterated through CLDR's supported language tags and found one at a distance of 10: `yue-Hans` (Cantonese in simplified characters, as used in Guangzhou) matches `zh` at distance 10. So that's the most borderline case.

rspeer: also, I control updates to langcodes, so if I think that CLDR's language distances have drastically changed in meaning I could save that update for a major version.

spacy/util.py
svlandeg: We could also make the `try` bit even shorter, which I think further clarifies the flow: 
```suggestion
        language = None
        try:
            language = langcodes.Language.get(code)
        except langcodes.tag_parser.LanguageTagError:
            pass
        if language is not None and language.is_valid():
            possible_languages.append(code)
```
Alternatively, `langcodes` could expose a method that deals with the `try`-`catch` internally, something like `is_parsable`.

svlandeg: I do like that this iterates over the actual code base instead of working with a fixed list that we may or may not forget to update at some point ;-)

spacy/util.py
```diff
@@ -256,20 +258,85 @@ def lang_class_is_loaded(lang: str) -> bool:
...
+    None
+    """
+    import spacy.lang  # noqa: F401
+    if lang == 'xx':
+        return 'xx'
+
+    # Find out which language modules we have
+    possible_languages = []
+    for modinfo in pkgutil.iter_modules(spacy.lang.__path__):
+        code = modinfo.name
```
svlandeg, line 295: At this point I would consider adding an additional check if `code` just equals `lang` and if so, return that and exit the method without having to compile the whole list and call `closest_match` etc. For `en` for example, that would break off the loop quite quickly.

spacy/util.py
```diff
@@ -256,20 +258,85 @@ def lang_class_is_loaded(lang: str) -> bool:
...
+            pass
+
+    match, _dist = langcodes.closest_match(
+        lang, possible_languages, max_distance=10
+    )
+    if match == 'mul':
+        # Convert 'mul' back to spaCy's 'xx'
+        return 'xx'
+    elif match != 'und':
+        return match
```
svlandeg, line 313: It's a little unfortunate that we need to check for a hardcoded string value of another library. I don't suppose `"und"` is defined in a variable somewhere? Why doesn't `closest_match` use `None` in these cases?

spacy/util.py
svlandeg: It looks like the default in `langcodes` is 25, that seems pretty high after reading your explanation here? I'd personally prefer not specifying this parameter in `spacy` and just using the default, but I can see that too high a value would be problematic. If it won't be updated in `langcodes`, I think having a more restrictive setting in spaCy would be good, especially when publishing this feature as users will not expect some kind of fuzzy matching behaviour here, and a typo might come to bite them.

svlandeg: That actually brings up the question whether there would be users who'd like to be able to enforce the current, strict behaviour. If we want to cater for that use-case, I think having this variable be adjustable (env var?) would work.

spacy/util.py
svlandeg: Ah wait, we probably don't need it because `get_lang_class` resolves the easy cases, so I think this comment can be ignored.

svlandeg:
Agreed with Matt that this PR is a great idea! I just added some small comments & thoughts.

> We have a test expecting an `ImportError` that now gets a non-spaCy-coded `LanguageTagError`, when trying to get `spacy.blank('fjsfijsdof')`. Should I catch that error and turn it into an error with a proper spaCy error code?

Yes - I would catch this and raise a custom `ValueError` with a new spaCy error msg that you can define in `spacy.errors`.

spacy/util.py
rspeer: In langcodes I wanted to return something that can be parsed as a language code instead of adding an optional type, and 'und' is the BCP 47 code for a language code that provides no information.

A different way to get this string value from langcodes would be `str(langcodes.Language(language=None))`. I could also add that as a constant in the next version (it's due for one when CLDR v40 comes out in early October).

spacy/util.py
rspeer: I could make a top-level `langcodes.is_valid_tag()` that takes in a string, and returns false if it doesn't parse _or_ if it parses to an invalid code.

If that's the plan, then we just need to remember not to make submodules that are accidentally valid tags, which would involve the first component being 2-3 letters. I'm imagining a hypothetical module like "no_support.py" where `no_support` parses as a variant of Norwegian.

spacy/util.py
rspeer: ...in fact, looking back, there was a standard that said I should return 'und' in that case.

But I understand why a function with a different signature would help. Maybe I should add some functions to a version of langcodes now, without waiting for the data to change.

spacy/util.py
rspeer: I think we'd just be setting up for confusion by making that configurable. For what it's worth, here is the complete list of plain language codes that are within a distance of 10 of one of our languages, but not 0:

`['bs', 'gsw', 'no']`

There are a lot more cases that involve regional codes like `en-UK` but I don't think those would come from a typo.

25 is a high default, yeah. I picked 25 as a default for the simple case where i18n consists of looking up strings in a table (which is what langcodes uses for naming languages when you include the optional `language_data` package). There are some distances near 25 that we definitely wouldn't want for NLP.

rspeer:
Okay, this one should resolve a lot of the comments. I made a new version of langcodes for it (but I'm not going to publish it right before the weekend).

spacy/tests/test_misc.py
```diff
@@ -139,7 +139,13 @@ def test_load_model_blank_shortcut():
...
     assert nlp.lang == "en"
     assert nlp.pipeline == []
+
+    # ImportError for loading an unsupported language
     with pytest.raises(ImportError):
+        util.load_model("blank:zxx")
+
+    # ValueError for requesting an invalid language code that isn't registered
+    with pytest.raises(ValueError):
         util.load_model("blank:fjsfijsdof")
```
svlandeg, line 149: Hm actually now that I see it like this, maybe this is confusing behaviour. A user might not necessarily understand why some invalid codes are giving an `ImportError` and others a `ValueError`. Instead, perhaps we can just `pass` when we catch the `LanguageTagError` and then the fall-back `else` will raise the same `ImportError` as before?

rspeer:
In addition to changing the ValueError back to an ImportError, I published langcodes 3.2.0 today. (A Unicode person has found some edge-case bugs in parsing that mean I'll have to release 3.2.1 soon, but they don't realistically affect us.) This should support the things this pull request needs.

spacy/util.py
svlandeg: @rspeer : maybe that's a note that you could add to the [dev docs](https://github.com/explosion/spaCy/tree/master/extra/DEVELOPER_DOCS)?

svlandeg (approved):
Looks like all the review comments were addressed and I think the current implementation is quite a bit cleaner with the new helper method from `langcodes` :-)

PR #9398, opened 2021-10-08
Move core training logic in CLI into standalone function

ines:
## Description

Not sure if we want this but I think it'd make it easier to provide a simple solution for users who don't want to run `spacy train` from the CLI. The CLI function isn't easily callable from Python, so this helper exposes a very simple, easy-to-call function that runs the training exactly like the CLI does.

```python
from spacy.cli.train import train

train("./config.cfg", overrides={"paths.train": "./train.spacy", "paths.dev": "./dev.spacy"})
```

It introduces some minor duplication in the defaults but I'd be okay with that.

### Types of change
enhancement

## Checklist

### DIFF
--- a/spacy/cli/train.py
+++ b/spacy/cli/train.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Dict, Any
 from pathlib import Path
 from wasabi import msg
 import typer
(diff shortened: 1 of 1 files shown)


svlandeg:
Ah yes, I've been wanting to have this, too!

Issue #9387, opened 2021-10-07
Non-deterministic results for Catalan lemmatizer across different runs


BLKSerene:
```
>>> import spacy
>>> [token.lemma_ for token in spacy.load('ca_core_news_sm')('parlada')]
['parlat']
>>> [token.lemma_ for token in spacy.load('ca_core_news_sm')('parlada')]
['parlat']
>>> [token.lemma_ for token in spacy.load('ca_core_news_sm')('parlada')]
['parlat']

================================ RESTART: Shell ================================
>>> import spacy
>>> [token.lemma_ for token in spacy.load('ca_core_news_sm')('parlada')]
['parlad']
>>> [token.lemma_ for token in spacy.load('ca_core_news_sm')('parlada')]
['parlad']
>>> [token.lemma_ for token in spacy.load('ca_core_news_sm')('parlada')]
['parlad']
```

May be related: #8421 

* Operating System: Windows 10 x64
* Python Version Used: 3.8.10
* spaCy Version Used: 3.1.1

polm:
Thanks for the report, that looks like a bug - we'll give it a look!

polm:
OK, I found the root of this problem. The lemmatizing function can find multiple potential lemmas for a word. When it does, they are accumulated in a list, but before returning them the list is made unique by calling `list(set(forms))`. However, the order of a set is not consistent from run to run in Python, and only the first entry in the resulting list is used normally, which causes the inconsistency we see here. 

This looks like it also affects the lemmatizers for French, Dutch, Romanian, and Russian, or at least they have `list(set(xxx))` calls. 

I'll work on a fix for this. Thanks again for reporting!

PR #9405, opened 2021-10-09
Fix inconsistent lemmas

polm:
## Description

As reported in #9387, the Catalan lemmatizer has different lemmas across different processes. The cause is using `list(set(xxx))` to make a list unique. Since the order of a set can vary between process runs, and since the first item in the list is used as the final lemma, that caused inconsistent lemmas. 

This PR adds a function called `unique` to `spacy.util` that keeps order in a list but removes duplicate elements. The implementation is simple and there may be a better way to do it. 

Calls to `list(set())` in lemmatizers and elsewhere were replaced with `unique()`, except where the call was like `sorted(list(set()))`, which was left alone.

One thing about this is that I don't completely understand what's going on when multiple lemma forms are available, and whether being first in the list has any significance. It's possible we don't need a util function and the `sorted` call convention would be enough. 

One other thing is that I'm not really sure how to test this, since the inconsistency can only occur between processes. I could spin up a python child process but it seems like overkill.

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/lang/ca/lemmatizer.py
+++ b/spacy/lang/ca/lemmatizer.py
@@ -2,7 +2,6 @@ from typing import List, Tuple
 
 from ...pipeline import Lemmatizer
 from ...tokens import Token
-from ...util import unique
 
 
 class CatalanLemmatizer(Lemmatizer):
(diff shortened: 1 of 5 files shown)


adrianeboyd:
Ah, I could have sworn I had fixed all these cases already, argh. The solution that we use is:

```python
forms = list(dict.fromkeys(forms))
```

I don't think we need a separate util method for this.

PR #9423, opened 2021-10-11
Support issue marker via pytest

ines:
## Description

This PR introduces an `issue` marker for `pytest` to label tests with one or more issues they're referring to:

```python
@pytest.mark.issue(1234)
def some_test_related_to_1234():
    ...
```

Decorators can also be stacked out-of-the-box:

```python
@pytest.mark.issue(1234)
@pytest.mark.issue(5678)
def some_test_related_to_1234_and_5678():
    ...
```

The `--issue` option on the `pytest` CLI now lets you run all tests referring to one or more issues only. All other tests (without a marker or referencing a different issue) will be skipped:

```bash
python -m pytest spacy --issue 1234,666
```

### Todo

- [ ] decorate more tests
- [ ] check performance? This is the correct usage as [documented](https://docs.pytest.org/en/6.2.x/example/markers.html) and maybe `pytest` does some fancy caching, or it just doesn't matter. But it felt wrong to me to do so much in the test setup 🤷‍♀️
- [x] Do we want a syntax that lets us specify multiple issues inline instead of stacking decorators (which might become a bit verbose)? It's doable but requires more custom logic compared to stacked decorators that work out-of-the-box.

### Types of change
tests

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -122,7 +122,8 @@ exclude =
 
 [tool:pytest]
 markers =
-    slow
+    slow: mark a test as slow
+    issue: reference specific issue
 
 [mypy]
 ignore_missing_imports = True
--- a/spacy/tests/conftest.py
+++ b/spacy/tests/conftest.py
@@ -4,6 +4,7 @@ from spacy.util import get_lang_class
 
 def pytest_addoption(parser):
     parser.addoption("--slow", action="store_true", help="include slow tests")
+    parser.addoption("--issue", action="store", help="test specific issues")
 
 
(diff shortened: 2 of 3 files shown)


polm:
> Do we want a syntax that lets us specify multiple issues inline instead of stacking decorators (which might become a bit verbose)? It's doable but requires more custom logic compared to stacked decorators that work out-of-the-box.

Considering that I don't think many tests cover multiple issues (and we're likely to have consolidated issues in that case anyway) I think just decorator stacking is fine.

PR #9443, opened 2021-10-12
Fix Scorer.score_cats for missing labels

flotothemoon:
Fix Scorer.score_cats returning incorrect scores if gold is missing labels.  

## Description
The current logic for ignoring missing labels is incorrect:
 - As noted by the comment, it assumes that `gold_score is None` if the corresponding gold label is missing. 
 - It uses a default return value when accessing `gold_cats` so it won't be `None` even if the value is missing. 

### Types of change
Bug fix. 

## Checklist
- [X] I ran the (relevant) tests, and all new and existing tests passed.

## Notes
This bug took me a few hours to find so I wanted to submit a quick PR to fix it for everyone. In a perfect world, this PR would include tests for `score_cats` (there are none right now).


flotothemoon:
Thanks for your comments, I hadn't seen `test_textcat`. I've added my test highlighting the issue.

It seems we might not be aligned on what "correct" means here, since you say:

> That doesn't test missing labels explicitely, though a value set to 0.0 should be the same as a missing gold label.

If this were true, it would mean that evaluation and training would handle missing annotations differently (refer to lines 280-285 in `textcat.py`'s `TextCategorizer`).

Logically, I don't think we want to count missing textcat annotations as anything but missing. Am I missing something here?

flotothemoon:
Following up on this, does the provided test case make sense to you @svlandeg?

polm:
I think this is related to #9548.

adrianeboyd:
Also a note: v3.2.0 will have explicit scoring functions for each component in the config, so it'll be easier to make this update while preserving the option of having the previous behavior for those who'd prefer it.

spacy/tests/pipeline/test_textcat.py
```diff
@@ -496,6 +496,38 @@ def test_textcat_evaluation():
...
+    pred2.cats = {"winter": 1.0, "summer": 0.0, "spring": 1.0, "autumn": 1.0}
+    train_examples.append(Example(pred2, ref2))
+
+    scores = Scorer().score_cats(
+        train_examples, "cats", labels=["winter", "summer", "spring", "autumn"]
+    )
+    assert scores["cats_f_per_type"]["winter"]["p"] == 1 / 2
+    assert scores["cats_f_per_type"]["winter"]["r"] == 1 / 1
+    assert scores["cats_f_per_type"]["summer"]["p"] == 0
+    assert scores["cats_f_per_type"]["summer"]["r"] == 0 / 1
```
svlandeg, line 522:
```suggestion
    assert scores["cats_f_per_type"]["summer"]["r"] == 0
```
Nitpick: summer is never `True` in the reference annotation so recall is always 0 - removing the `/1` which seems to imply a missed prediction

spacy/tests/pipeline/test_textcat.py
```diff
@@ -496,6 +496,38 @@ def test_textcat_evaluation():
...
+    assert scores["cats_f_per_type"]["spring"]["p"] == 1 / 1
+    assert scores["cats_f_per_type"]["spring"]["r"] == 1 / 1
+    assert scores["cats_f_per_type"]["autumn"]["p"] == 2 / 2
+    assert scores["cats_f_per_type"]["autumn"]["r"] == 2 / 2
+
+    assert scores["cats_micro_p"] == 4 / 5
+    assert scores["cats_micro_r"] == 4 / 4
+
```
svlandeg, line 530:
```suggestion


```

svlandeg:
Hi again @flotothemoon! 

We've discussed this further internally and have realised that the confusion about the semantics of "missing label" is caused by the difference whether or not a `textcat` component has mutually exclusive labels or not. 

The normal `textcat` component always has exactly one `True` label per document, and so any missing label should be interpreted as `0.0` (assuming that a `True` label is already present). By contrast, the `textcat_multilabel` component should be used to predict zero, one or more true labels per document, and in this case a missing label should really just be missing - ie no "penalty" on accuracy should be induced either. This corresponds to the test case you wrote in this PR.

We've looked at the loss functions and you're right that these are not consistent/correct with respect to these definitions. The `get_loss` method in `TextCategorizer` shouldn't ignore missing values that really are (semantically) `False`, though the current implementation of it is fine for `MultiLabel_TextCategorizer`.

Then, we should make sure that the same definition is used for the `score_cats` function, which already knows about the boolean state of `multi_label`. We probably need a few more tests for this. And finally, we need to revise the documentation to clarify/correct this where need be.

I know this PR has dragged a bit, so please let me know if you'd prefer to continue working on this, or whether you don't have bandwidth for this anymore and would like us to take over at this point.

Either way, thanks again for flagging this - we'll definitely follow up and make sure this gets corrected!

flotothemoon:
Hi @svlandeg, no worries. Thanks for the update - this makes sense.

Regarding the implementation - I'm busy right now, so it depends on when your time horizon for getting this fixed I. If you want it in the next week or so, feel free to take over.

spacy/pipeline/textcat.py
```diff
@@ -294,7 +300,7 @@ def _examples_to_truth(
             for j, label in enumerate(self.labels):
                 if label in eg.reference.cats:
                     truths[i, j] = eg.reference.cats[label]
-                else:
+                elif self.support_missing_values:
```
polm, line 303: What is going on here is kind of subtle and might merit a comment. Something like:

> this is always false by default, but subclasses override it. For the single class case, other values will just be ignored.

polm (approved):
This looks good!

spacy/pipeline/textcat.py
```diff
@@ -158,6 +158,12 @@ def __init__(
         self.cfg = dict(cfg)
         self.scorer = scorer
 
+    @property
+    def support_missing_values(self):
+        # There are no missing values as the textcat should always
+        # predict exactly one label. All other labels are 0.0
+        return False
+
```
svlandeg, line 166:
```suggestion
    @property
    def support_missing_values(self):
        # There are no missing values as the textcat should always
        # predict exactly one label. All other labels are 0.0
        # Subclasses may override this property to change internal behaviour.
        return False

```

spacy/pipeline/textcat.py
svlandeg: I've added the comment to the definition of the property itself, do you think that's sufficient?

polm: That should do it, thanks!

svlandeg:
While this is technically a bug fix that could go on `master`, the bug is kind of subtle and behaviour might be surprisingly different for some users after the merge, which is why I think we should put it on `develop` to keep it for the 3.3.0 release. I'll update `develop` first, close and reopen this PR and hopefully the extra commits will be gone. Then I think this will be good to merge.

svlandeg:
Thanks again for bringing this to our attention and for the initial PR, @flotothemoon !

PR #9449, opened 2021-10-13
Fix JA Morph Values

polm:
## Description

While training a new model some issues with JA morph values were noticed. 

1. Empty values were being set
2. Some weird values, like `=`, were being set

These cause issues with the morph value, so changes were made to avoid setting them. 

Note that currently this looks for the SudachiPy tag 補助記号 ("supporting symbols", ~= sentence punctuation), which should cover all the special morph values in normal use. 

It might be necessary to check for things that contain invalid morph values, like `=_=`. I am not sure exactly which characters  should be avoided, maybe `[=|,]`?

## Checklist


polm:
OK, got it. 

In this case `inflection` should never contain any of those characters, and `reading` will only contain them if the original token does - for punctuation or terms not in the dictionary the text is its own reading. 

The way I was originally setting the values was creating a dict and passing it in for a Morph rather than setting values on a Morph object, so maybe that bypasses some checks?

PR #9471, opened 2021-10-15
Removed a manual file handler pitfall (Pylint R1732 consider-using-with code)

NaelsonDouglas:
## Description
There was a part in the code where a file stream was being open manually.  I just changed the code in order to open the file into a with block.

### Types of change
Refactoring

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,17 +9,17 @@ trigger:
       - "v2.spacy.io"
   paths:
     exclude:
-      - "website/*"
+      - "website/**"
       - "*.md"
 pr:
   paths:
     include:
-      - "*.cfg"
-      - "*.py"
-      - "*.toml"
-      - "*.yml"
+      - "setup.cfg"
+      - "setup.py"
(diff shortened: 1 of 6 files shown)


polm:
Thanks for catching that! Looks good to me.

adrianeboyd:
For background: this code was adapted from `numpy`, but it looks like they're using a totally different approach as of the end of 2020. Maybe we should rework all of this...

NaelsonDouglas:
> Not quite, we want this to continue, probably silently, if it can't write to this file.

@adrianeboyd  I silenced it on [here](https://github.com/explosion/spaCy/pull/9471/commits/7754f11f7607272b1c16f941e6541fc9c018968a) :)

svlandeg:
Hi @NaelsonDouglas , the proposed changes actually change the semantics of the code, as the current version doesn't catch any exception but instead would reraise it. We'll go ahead and close this, as this part of the code base is too risky to make minor style fixes too, risking bugs or changed behaviour. I hope you understand!

Issue #5507, opened 2020-05-26
GPU and multiprocessing leads to ' initialization error'


zhanwen:
```
spacy.prefer_gpu()
nlp = spacy.load('en_core_web_sm', disable=["tagger", "parser"]) 
docs = nlp.pipe(texts, batch_size=128, n_process=2, disable=["tagger", "parser"])
```
NOTE :
if I am using n_process=1 the code runs fine.
If I am not using the gpu the code also runs fine.
It is the combination of GPU and multiprocessing that leads to this error.
```
process Process-2:
Traceback (most recent call last):
  File "/home/ubuntu/anaconda3/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/home/ubuntu/anaconda3/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/spacy/language.py", line 1134, in _apply_pipes
    sender.send([doc.to_bytes() for doc in docs])
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/spacy/language.py", line 1134, in <listcomp>
    sender.send([doc.to_bytes() for doc in docs])
  File "nn_parser.pyx", line 253, in pipe
  File "nn_parser.pyx", line 273, in spacy.syntax.nn_parser.Parser.predict
  File "nn_parser.pyx", line 286, in spacy.syntax.nn_parser.Parser.greedy_parse
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/thinc/neural/_classes/model.py", line 167, in __call__
    return self.predict(x)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/thinc/neural/_classes/model.py", line 131, in predict
    y, _ = self.begin_update(X, drop=None)
  File "_parser_model.pyx", line 243, in spacy.syntax._parser_model.ParserModel.begin_update
  File "_parser_model.pyx", line 293, in spacy.syntax._parser_model.ParserStepModel.__init__
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/thinc/neural/_classes/feed_forward.py", line 46, in begin_update
    X, inc_layer_grad = layer.begin_update(X, drop=drop)
  File "/home/ubuntu/anaconda3/lib/python3.7/site-packages/spacy/_ml.py", line 791, in flatten
[...]

PR #9475, opened 2021-10-15
Print a warning when multiprocessing is used on a GPU

danieldk:
## Description

As reported in #5507, a confusing exception is thrown when
multiprocessing is used with a GPU model and the `fork` multiprocessing
start method:

cupy.cuda.runtime.CUDARuntimeError: cudaErrorInitializationError: initialization error

This change checks whether one of the models uses the GPU when
multiprocessing is used. If so, raise a friendly error message.

Even though multiprocessing can work on a GPU with the `spawn` method,
it quickly runs the GPU out-of-memory on real-world data. Also,
multiprocessing on a single GPU typically does not provide large
performance gains.

### Types of change

Usability improvement.

## Checklist


danieldk:
> Do we need this on `develop` or can it just go to `master`? If this wasn't working before anyway, I don't think it's breaking?

I wasn't too sure about this, since technically multiprocessing did work with the `spawn` method before.

adrianeboyd:
I would be more inclined to warn than to try to forbid this.

adrianeboyd:
(Without adding a hack to the website, updating the compatibility table causes the website models page to try to v3.2.0 models that haven't been published yet.)

danieldk:
> It will deadlock in the CI with fork without `torch.set_num_threads(1)`.

Luckily did not happen yet, but I did this anyway for safety.

svlandeg:
@danieldk : could you merge in the latest `master` and try running the GPU tests again?

PR #9501, opened 2021-10-19
Replace use_ops("numpy") by use_ops("cpu") in the parser

danieldk:
## Description

Replace `use_ops("numpy")` by `use_ops("cpu")` in the parser.

This ensures that the best available CPU implementation is chosen (e.g. Thinc Apple Ops on macOS).

### Types of change

Performance improvement on Apple M1.

## Checklist

### DIFF
--- a/.github/azure-steps.yml
+++ b/.github/azure-steps.yml
@@ -108,3 +108,9 @@ steps:
       python .github/validate_universe_json.py website/meta/universe.json
     displayName: 'Test website/meta/universe.json'
     condition: eq(variables['python_version'], '3.8')
+
+  - script: |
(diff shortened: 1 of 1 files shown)


danieldk:
> I'm always in favour of more tests, but I'm not sure it's useful to duplicate the Thinc tests on this particular occasion? There's no additional functionality provided by spaCy, right?

It would verify whether the `use_ops` in the parser still results in a working parser.

danieldk:
For clarification, with *like we do for thinc* I meant to add a CI test step that runs the spaCy tests with `apple-thinc-ops` installed. Not to port the test for `use_ops` to spaCy.

svlandeg:
Ah, thanks for the clarification, yes, that's a good idea!

PR #9520, opened 2021-10-20
Rename ja morph features to Inflection and Reading

adrianeboyd:
## Description

Rename ja morph features to Inflection and Reading

### Types of change

?

## Checklist


adrianeboyd:

adrianeboyd:
This is more complicated than I thought. You can specify this for the morphologizer with a custom scorer, but then the attribute ruler comes along later and runs the default evaulation for `MORPH`, which overwrites the custom results and gets you back to useless results.

You could generalize this so the scorer can provide a list of things to ignore, but there's no top-level config for the scorer that would make it easy to specify this in a general way. At best you'd have custom factories for `morphologizer` and `attribute_ruler` for `ja` and it could still break down pretty easily with new components vs. your training data that does not contain the additional features from the tokenizer.

PR #9537, opened 2021-10-25
Spanish noun chunks review

DuyguA:
## Description
Quick review on Spanish noun chunks. I checked adjectival clauses, nominal modifiers, compounds that are made by `fixed`, `flat` and `compound`.

Here are some cases to compare to English:

bshirt1

bshirt

Adjectives usually come after the noun, but  `ADJ NOUN` is also quite OK. Using several adjectives to qualify the noun also looks different:

girl1

girl2

Making compounds is a bit different too, English uses `compound` tag and arcs aim left. Spanish parser uses `flat` and in this case we have to extend to right:
flat1
flat2
flat3

Hence for Spanish, I put some code to extend the NPs to right.
A final word is on nominal modifiers and `poss` relation. Spanish doesn't have a `poss` relation and marker, hence possessiveness is shown by `nmod`. Please compare the below constructions, I also included those in tests. 

dest1
dest3
des2

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/tests/lang/es/test_noun_chunks.py
+++ b/spacy/tests/lang/es/test_noun_chunks.py
@@ -38,6 +38,14 @@ import pytest
             ["DET", "NOUN", "ADJ", "CCONJ", "ADJ"],
             [(0,5)]
         ),
+        # el fabuloso gato pardo -> "el fabuloso gato pardo"
+        (
+            ["el", "fabuloso", "gato", "pardo"],
+            [2, 2, 2, 2],
+            ["det", "amod", "ROOT", "amod"],
+            ["DET", "ADJ", "NOUN", "ADJ"],
+            [(0,4)]
+        ),
         # Tengo un gato y un perro -> un gato, un perro
         ( 
             ["Tengo", "un", "gato", "y", "un", "perro"],
--- a/spacy/tests/lang/es/test_noun_chunks.py
+++ b/spacy/tests/lang/es/test_noun_chunks.py
@@ -27,7 +27,7 @@ def test_noun_chunks_is_parsed_es(es_tokenizer):
 
 def test_noun_chunks_basic(doc_basic, es_tokenizer):
     """Test that noun_chunks for basic np determiner+noun"""
-    doc_chunks = list(doc.noun_chunks)
+    doc_chunks = list(doc_basic.noun_chunks)
     assert len(doc_chunks) == 1
     chunk = doc_chunks[0]
     assert chunk in doc_chunks
@@ -38,7 +38,7 @@ def test_noun_chunks_basic(doc_basic, es_tokenizer):
 
 def test_noun_chunks_basic_adj(doc_basic_adj, es_tokenizer):
     """Test that noun_chunks for basic np determiner+noun+adj"""
-    doc_chunks = list(doc.noun_chunks)
+    doc_chunks = list(doc_basic_adj.noun_chunks)
     assert len(doc_chunks) == 1
     chunk = doc_chunks[0]
     assert chunk in doc_chunks
--- a/spacy/tests/lang/es/test_noun_chunks.py
+++ b/spacy/tests/lang/es/test_noun_chunks.py
@@ -17,6 +17,29 @@ def doc_basic_adj(es_vocab):
     pos = ["DET", "NOUN", "ADJ"]
     return Doc(es_vocab, words=words, heads=heads, deps=deps, pos=pos)
 
+@pytest.fixture
+def doc_adj_noun(es_vocab):
+    words = ["Un", "lindo", "gatito"]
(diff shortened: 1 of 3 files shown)


spacy/tests/lang/es/test_noun_chunks.py
```diff
@@ -1,6 +1,148 @@
...
+        ),
+        # un lindo gatito -> "un lindo gatito"
+        (
+            ["Un", "lindo", "gatito"],
+            [2, 2, 2],
+            ["det", "amod", "ROOT"],
+            ["DET", "ADJ", "NOUN"],
+            [(0,3)]
+        ),
+        # una chica hermosa e inteligente -> una chica hermosa e inteligente
```
damian-romero, line 33: So I tried with noun chunks that have modifiers on both sides such as "the fabulous brown cat" and I think the code is not getting those as full chunks. I get an assertion error if I try the following: 
```
        # el fabuloso gato pardo -> "el fabuloso gato pardo"
        (
            ["el", "fabuloso", "gato", "pardo"],
            [3, 3, 3, 3],
            ["det", "amod", "ROOT", "amod"],
            ["DET", "PROPN", "PROPN", "PROPN"],
            [(0, 3)],
        ),
```
Fails with: assert [(2, 3)] == [(0, 3)]

spacy/tests/lang/es/test_noun_chunks.py
```diff
@@ -1,6 +1,148 @@
...
+            [(0,3)]
+        ),
+        (
+            ["Rio", "de", "Janeiro"],
+            [0, 2, 0],
+            ["ROOT", "case", "flat"],
+            ["PROPN", "ADP", "PROPN"],
+            [(0,3)]
+        ),
+        # la destrucción de la ciudad -> la destrucción, la ciudad
```
damian-romero, line 81: I have a question: I understand we want all the noun chunks here but "a destrucción de la ciudad" is a complex noun phrase in itself. We're not interested in marking those as noun chunks as well? Another example of what I have in mind is "The cat that looks nice". Would that phrase just have one noun chunk "The cat"?

damian-romero (approved):
I reviewed the code and tested it with other cases. I think it works well. I just had one comment regarding noun chunks with several modifiers and a question about complex noun phrases.

spacy/tests/lang/es/test_noun_chunks.py
DuyguA: Sure, this is a point I thought about as well. Here, I inspired from how we did for English. For English the NP  `destruction of the city` results in two small NPs `destruction` and `the city`. There are 2 ways to do it here, either a long NP or 2 NPs. I chose the way we do it for English (same for Portuguese). What do you think? If you think the whole NP is more useful for users, I can change the PR a little bit, definitely no problem.

spacy/tests/lang/es/test_noun_chunks.py
DuyguA: > So I tried with noun chunks that have modifiers on both sides such as "the fabulous brown cat" and I think the code is not getting those as full chunks. I get an assertion error if I try the following:
> 
> ```
>         # el fabuloso gato pardo -> "el fabuloso gato pardo"
>         (
>             ["el", "fabuloso", "gato", "pardo"],
>             [3, 3, 3, 3],
>             ["det", "amod", "ROOT", "amod"],
>             ["DET", "PROPN", "PROPN", "PROPN"],
>             [(0, 3)],
>         ),
> ```
> 
> Fails with: assert [(2, 3)] == [(0, 3)]

Argh, can we try changing the -3th line to:
```
["DET", "ADJ', "PROPN", "ADJ"]
```
pos tags doesn't seem correct to me :slightly_smiling_face:

damian-romero: Ah yes! I was using the tags and parse structure from one of the small language models in Spanish. This is better :)

spacy/tests/lang/es/test_noun_chunks.py
damian-romero: I agree. It'd be better to have all languages work in a similar wat if possible. Also we always know what the smallest noun phrases will look like. The complex ones are unbounded so maybe they should be parsed elsewhere if parsed at all, I think.

spacy/tests/lang/es/test_noun_chunks.py
DuyguA: I added this case to the unit tests, thanks for that!

spacy/tests/lang/es/test_noun_chunks.py
honnibal: @DuyguA Yes that's how I'd expect it to be :)

PR #9572, opened 2021-10-30
Add WordDumb to spaCy Universe

xxyzz:
## Description
Add [WordDumb](https://github.com/xxyzz/WordDumb/) to spaCy Universe. WordDumb uses spaCy to find named entities in books to create Kindle X-Ray file.

### Types of change
Update `universe.json`.

## Checklist

I only run `.github/validate_universe_json.py` on `universe.json`, I guess this commit won't affect other tests...

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -3518,7 +3518,21 @@
             },
             "category": ["pipeline", "research", "standalone"],
             "tags": ["spacy", "python", "nlp", "ner"]
-        }        
+        },
+        {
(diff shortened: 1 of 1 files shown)


polm:
Thanks for the submission!

We require a category on entries, which we use to show subsets of projects. It does look like none of the categories really fit WordDumb, but I went ahead and put it in "standalone" since that seemed the closest.

xxyzz:
Thanks!

PR #9593, opened 2021-11-02
Use metaclass to subclass errors to allow better pickling

### LINKED ISSUE #9584: Pickling with dill and recurse=True fails
While trying to use spaCy in preparing a dataset, I found that `dill` is not able to dump a `Language` when using `recurse=True`. In my case it is not possible to set `recurse=None/False` because this handled externally. (Related https://github.com/huggingface/datasets/issues/3178) So this issue serves to ask why this is the case and whether this behavior can be improved. I must admit that I do not quite understand the deep implication of `recurse`, but the [docs](https://dill.readthedocs.io/en/latest/dill.html#dill._dill.dumps) write

> If recurse=True, then objects referred to in the global dictionary are recursively traced and pickled, instead of the default behavior of attempting to store the entire global dictionary. This is needed for functions defined via exec().

```python
import dill
import spacy

if __name__ == '__main__':
    nlp = spacy.load("en_core_web_sm", disable=["tagger", "parser", "ner", "lemmatizer"])
    # fails
    with open("temp.pkl", "wb") as f:
        dill.Pickler(f, recurse=True).dump(nlp)
```

## Your Environment

- **spaCy version:** 3.1.0
- **Platform:** Windows-10-10.0.19041-SP0
- **Python version:** 3.6.8
- **Pipelines:** en_core_web_sm (3.1.0)

BramVanroy:
When trying to pickle with dill, as happens under the hood when processing a HuggingFace `dataset`, I found that spaCy couldn't be pickled (related issue: https://github.com/huggingface/datasets/issues/3178). I found that the cause was `dill.Pickler` that recursively pickled all objects (`Pickler(file, recurse=True).dump(obj)`). This is preferred in `datasets`: such iterative pickling allows for fingerprinting, i.e. keeping track of all the processing done to a dataset, deterministically. If the fingerprint is the same as an earlier processing attempt, then a cache can be used and the processing (e.g. tokenizing with spaCy) does not have to be done again. Very useful and great to save time and processing.

The culprit in the spaCy codebase that could not be pickled in this way, was this part

https://github.com/explosion/spaCy/blob/f1bc655a387a040bbd64d8650eade135215e8a0a/spacy/errors.py#L4-L15

The class `ErrorsWithCodes` cannot be pickled here because it does not have a fixed signature in the global scope (its superclass is not known beforehand). After digging through the Internet (thank you [Stack Overflow](https://stackoverflow.com/a/69808716/1150683)!) I found that a metaclass is the way to go, as implemented in this PR. In my opinion, this is also more structurally clear than before. Regular subclassing a metaclass instead of dynamically created a class based on a subclass with a decorator. The *meta*class defines the magic methods that are required to access the class' attributes, rather than accessing instance attributes.

I modified the error test accordingly, which completes successfully.

After this modification, recursive dill pickling works flawlessly as well.

closes #9584 

### Types of change
Enhancement

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -1,18 +1,13 @@
 import warnings
 
 
-def add_codes(err_cls):
-    """Add error codes to string messages via class attribute names."""
-
-    class ErrorsWithCodes(err_cls):
-        def __getattribute__(self, code):
-            msg = super(ErrorsWithCodes, self).__getattribute__(code)
-            if code.startswith("__"):  # python system attributes like __class__
-                return msg
-            else:
-                return "[{code}] {msg}".format(code=code, msg=msg)
-
-    return ErrorsWithCodes()
+class ErrorsWithCodes(type):
+    def __getattribute__(self, code):
+        msg = super().__getattribute__(code)
+        if code.startswith("__"):  # python system attributes like __class__
+            return msg
+        else:
+            return "[{code}] {msg}".format(code=code, msg=msg)
 
 
 def setup_default_warnings():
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks, the change makes sense to me, and it's nice to support pickling with dill.

PR #9597, opened 2021-11-02
Serialize _context separately in multiprocessing pipe

adrianeboyd:
## Description

Serialize _context separately for multiprocessing pipe

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1631,11 +1631,12 @@ class Language:
             recv.recv() for recv in cycle(bytedocs_recv_ch)
         )
         try:
-            for i, (_, (byte_doc, byte_error)) in enumerate(
+            for i, (_, (byte_doc, byte_context, byte_error)) in enumerate(
                 zip(raw_texts, byte_tuples), 1
             ):
                 if byte_doc is not None:
                     doc = Doc(self.vocab).from_bytes(byte_doc)
+                    doc._context = byte_context
                     yield doc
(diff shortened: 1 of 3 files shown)


adrianeboyd:
But `_multiprocessing_pipe` uses `Doc.to_bytes()`.

honnibal:
Oh Crap. Right, okay.

adrianeboyd:
That might work, yeah.

honnibal:
Nice, LGTM

adrianeboyd:
Well, it would work if `_context` got pickled with the `Doc`, argh...

svonava:
Hey everybody - first of all, spacy is awesome!

I'm seing some weird interplay between as_tuples and multiprocessing.

What I did:
1. Use en_core_web_lg, run nlp.pipe() over 1k docs with batch_size 100, n_process 10.
2. I assume that A & B can have different order here: "for A in nlp.pipe(B)" and I turned on as_tuples to align A & B after the pipe is done.
3. as_tuples=True made my code take an order of magnitude more ram, as if only once I set this true it actually creates 10 different representations of en_core_web_lg in memory.

Is this expected? Does this mean that when as_tuples=False, the multiprocessing doesn't separate the processes properly?

Thank you for any elaboration on this!

adrianeboyd:
Please open a new discussion thread rather than commenting on a older PR.

PR #9610, opened 2021-11-03
Fix tok2vec-less textcat generation in website quickstart

adrianeboyd:
## Description

Something is going wrong with the expression conversion or evaluation on the javascript side, but separating the expression into smaller components seems to at least temporarily fix the problem.

### Types of change

?

## Checklist

### DIFF
--- a/spacy/cli/templates/quickstart_training.jinja
+++ b/spacy/cli/templates/quickstart_training.jinja
@@ -16,8 +16,10 @@ gpu_allocator = null
 
 [nlp]
 lang = "{{ lang }}"
-{%- set no_tok2vec = components|length == 1 and (("textcat" in components or "textcat_multilabel" in components) and optimize == "efficiency")-%}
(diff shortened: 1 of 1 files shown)


almeidava93:
Hi!
I've got the same error here described, but with span categorisation. I didn't find yet a workaround. It seems to run fine with the small portuguese language model, but not with the large one because of tok2vec.

`Bad value substitution: option 'width' in section 'components.spancat.model.tok2vec' contains an interpolation key 'components.tok2vec.model.encode.width' which is not a valid option name. Raw value: '${components.tok2vec.model.encode.width}'`

Here is my code in the CLI:
`python -m prodigy train models/spancat_rfe --base-model pt_core_news_lg --spancat test_3 --lang pt -V`

adrianeboyd:
Hi @almeidava93 , an older/closed spacy PR isn't the best place for a new prodigy question. Since it's probably related to how prodigy is generating the spancat config internally, can you open a new thread in the prodigy support forum instead (https://support.prodi.gy)?

Issue #3940, opened 2019-07-10
ENT_IOB as a token pattern key


Varun-Epi:
## Feature description
Adding `ENT_IOB` attribute as one of the pattern key. Currently, it has not been integrated into entity ruler pipeline.

### Example
```
import spacy
from spacy.pipeline import EntityRuler

nlp = spacy.load('en')
ruler = EntityRuler(nlp, overwrite_ents=True)
pattern = [{'ENT_TYPE': 'GPE', 'OP': '+'}, {'ORTH': 'california'}]
ruler.add_patterns([{"label":'COMB', "pattern":pattern}])
```
```
print('Before Ruler')
print(nlp('Delhi New York california').ents)
Output: (Delhi, New York)
```
```
print('After Ruler')
nlp.add_pipe(ruler)
print(nlp('Delhi New York california').ents)
-- Output: (Delhi New York california,)
```
Correct Output should be **(Delhi, New York california)**

Logical answer would be to annotate only `New York california` as new entity, but here since `ENT_IOB` option is not available, it annotates all the `GPE` present in the given pattern

honnibal:
Fair point, although the `ENT_IOB` values currently have pretty obscure semantics, which makes usability a bit difficult. The field takes a numeric value with the following interpretation:

0: Missing
1: I -- In
2: O -- Out
3: B -- Begin

ines:
@honnibal I guess we could also have it take the strings `I`, `O` and `B` and a `0` for missing?

honnibal:
@ines I've thought a couple of times about how to clean this up. It's pretty annoying: if we take `I`, then what about the hash code of `I`? At least 0 always corresponds to the empty string, so that's easy.

This is one of the only places in the library where we'd using a numeric encoding that's *not* the same as the string-store encoding.

kinghuang:
Any updates on getting `ENT_IOB` added to token patterns?

I'm trying to write a pattern that will pick up and overwrite an existing entity. But, without `ENT_IOB` information, it's hard to accurately pick up all the tokens for an entity.

svlandeg:
Happy to report that this feature will be available in the next release!

Issue #4790, opened 2019-12-10
token.ent_iob should be writable


tamuhey:
I think `token.ent_iob` should be writable because `token.ent_type` is writable:
https://github.com/explosion/spaCy/blob/38e1bc19f4a1c28e27938f46bb3805df4919fedc/spacy/tokens/token.pyx#L733

`token.ent_iob` is not writable now, it can be only changed in cython:
https://github.com/explosion/spaCy/blob/38e1bc19f4a1c28e27938f46bb3805df4919fedc/spacy/tokens/token.pyx#L751

adrianeboyd:
I've wondered about this, too, and it seems like a good idea to make it consistent. It is a little difficult to manage modifying both while making sure that nothing ends up in an invalid state, though.

adrianeboyd:
Hi, we're just reviewing some older issues. As of v3, I think that `Doc.set_ents` is flexible enough to allow you to modify the entity attributes on any given span while making sure that the doc state remains consistent, e.g.:

```python
doc = nlp("An Entity")
doc.ents = [Span(doc, 0, 2, "ENTITY")]
assert doc.ents == ((Span(doc, 0, 2, "ENTITY")),)
doc.set_ents([], outside=[Span(doc, 0, 1)], default="unmodified")
assert doc.ents == ((Span(doc, 1, 2, "ENTITY")),)
```

PR #9620, opened 2021-11-04
WIP: Matcher allow overwriting ent_iob 

DuyguA:
## Description
This issue sourced from #3940 . Basically we want to make it possible to override `token.ent_iob_` by EntityRuler , hence forming longer entities from existing entities is possible. This will work then:

```
ruler.add_patterns([{"label":'GPE', "ent_iobs": "I", "pattern":[{"LOWER":"california"}]}])
doc = nlp("New York california")
doc.ents
("New York california", )
```

where current output is

```
("New York", "california")
```

### Types of change
New feature

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -887,6 +887,10 @@ class Errors(metaclass=ErrorsWithCodes):
     E1021 = ("`pos` value \"{pp}\" is not a valid Universal Dependencies tag. "
              "Non-UD tags should use the `tag` property.")
     E1022 = ("Words must be of type str or int, but input is of type '{wtype}'")
+    E1023 = ("Possible entity IOB values are 0, 1, 2 and 3, but the provided argument '{tag}' is not one of these values.")
+    E1024 = ("Possible entity IOB strings are I, O and B, but the provided argument '{tag}' is not one of these values.")
+
+
 
 
 # Deprecated model shortcuts, only used in errors and warnings
--- a/spacy/pipeline/entityruler.py
+++ b/spacy/pipeline/entityruler.py
@@ -438,13 +438,14 @@ class EntityRuler(Pipe):
             label = f"{label}{self.ent_id_sep}{ent_id}"
         return label
 
-    def _create_iob_label(self, label: Any, ent_iobs: str) -> str:
+    def _create_iob_label(self, label: Any, ent_iobs: Any) -> str:
         """Join Entity label with ent_iobs if the pattern has an `ent_iobs` attribute
         label (str): The label to set for ent.label_
         ent_iobs (str): The ent_iobs string
         RETURNS (str): The ent_label joined with configured `ent_iob_sep`
         """
-        label = f"{label}{self.ent_iob_sep}{ent_iobs}"
+        if isinstance(ent_iobs, str):
+            label = f"{label}{self.ent_iob_sep}{ent_iobs}"
         return label
 
     def from_bytes(
--- a/spacy/pipeline/entityruler.py
+++ b/spacy/pipeline/entityruler.py
@@ -228,7 +228,7 @@ class EntityRuler(Pipe):
                 label, _ = self._split_label(l)
                 all_labels.add(label)
             elif self.ent_iob_sep in l:
-                label, _ = self.split_iob_label(l)
+                label, _ = self._split_iob_label(l)
                 all_labels.add(label)
             else:
                 all_labels.add(l)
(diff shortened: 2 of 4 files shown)


adrianeboyd:
Sorry, I don't think this is a good idea. It's not possible to edit these incrementally and keep the `Doc` annotation valid at all times. Many internal methods assume that the IOB annotation is valid and will fail in unexpected ways if it's not. (However, if people want to muck with the values in cython, that's up to them.)

I'm also a little confused the connection to the linked source issue, since you can add `ENT_IOB` for matcher patterns without making it writable?

adrianeboyd:
Ah, maybe you meant https://github.com/explosion/spaCy/issues/4790?

DuyguA:
> Ah, maybe you meant #4790?

This is a wip, so I didn't add the EntityRuler code yet. 
That one is included too, I wanted to make `ent_id` and `ent_iob` consistent. :wink: 

> Sorry, I don't think this is a good idea. It's not possible to edit these incrementally and keep the Doc annotation valid at all times. Many internal methods assume that the IOB annotation is valid and will fail in unexpected ways if it's not.

OK, that's a fair point.  Basically users will be able to override ents extracted by EntityRuler. Still they can match existing entitites by EntityRuler and make their `ent_iob`s bad.
There's a IOB correction code in `doc`:  

https://github.com/explosion/spaCy/blob/2bf52c44b1cecc62204bf9366a559bcef305b8ee/spacy/tokens/doc.pyx#L750

Even if the users set bad iobs this piece of code corrects many cases.  I'll finish the code, try to screw up the existing entities by EntityRuler in tests and see how much they're corrected. What do you say?

adrianeboyd:
I wanted to say something before you spent more time on this in case writing to `Token.ent_iob` was crucial to your approach.

I think using `Doc.set_ents` is the right approach and I don't think `Token.ent_iob` or `Token.ent_type` should be individually writable in python. Adding `ENT_IOB` to the Matcher would be fine.

DuyguA:
> I wanted to say something before you spent more time on this in case writing to `Token.ent_iob` was crucial to your approach.
> 
> I think using `Doc.set_ents` is the right approach and I don't think `Token.ent_iob` or `Token.ent_type` should be individually writable in python. Adding `ENT_IOB` to the Matcher would be fine.

Sure :slightly_smiling_face:  It's a quick PR, I finish it up soon. Then if there are dangerous points in the code we can speak.

adrianeboyd:
I'm not sure it really makes sense for the entity ruler to do this. The attribute ruler can already do this (and would let you really screw up the IOB tags while doing it; and doesn't convert the string values for ENT_IOB correctly):

```python
import spacy

nlp = spacy.blank("en")
ar = nlp.add_pipe("attribute_ruler")
ar.add(patterns=[[{"ORTH": "a"}, {"ORTH": "cat"}]], attrs={"ENT_IOB": 3, "ENT_TYPE": "CATS"}, index=1)
doc = nlp("a cat")
print(doc.ents)
```

DuyguA:
> I'm not sure it really makes sense for the entity ruler to do this. The attribute ruler can already do this (and would let you really screw up the IOB tags while doing it; and doesn't convert the string values for ENT_IOB correctly):
> 
> ```python
> import spacy
> 
> nlp = spacy.blank("en")
> ar = nlp.add_pipe("attribute_ruler")
> ar.add(patterns=[[{"ORTH": "a"}, {"ORTH": "cat"}]], attrs={"ENT_IOB": 3, "ENT_TYPE": "CATS"}, index=1)
> doc = nlp("a cat")
> print(doc.ents)
> ```

OK, then I suggest we write an appropriate response to  #3940  and don't mess with IOBs :slightly_smiling_face:

adrianeboyd:
But that's not what #3940 wanted to do, is it? They just want `ENT_IOB` as a possible feature. I think that's sensible, and also not as complicated as this would be.

adrianeboyd:
It does have all the issues that Matt mentioned related to the IOB encoding, though, which I think is why it isn't implemented.

adrianeboyd:
Here's the current schema:

https://github.com/explosion/spaCy/blob/2bf52c44b1cecc62204bf9366a559bcef305b8ee/spacy/schemas.py#L214-L248

DuyguA:
> But that's not what #3940 wanted to do, is it? They just want `ENT_IOB` as a possible feature. I think that's sensible, and also not as complicated as this would be.

Afais yes, #3940 wants something else and this PR is for sth else. It'll help newly extracted entities of EntityRuler to merge with entities of NER. Like this one:

```
ruler.add_patterns([{"label":'GPE', "ent_iobs": "I", "pattern":[{"LOWER":"california"}]}])
doc = nlp("New York california")
doc.ents
("New York california", )
```

Current output is:  `(New York, california)`

adrianeboyd:
In that case you've have the entity ruler overwrite existing entities with the longer matches, where the match patterns potentially use `ENT_IOB`. I don't think entity ruler should do anything other than set full entities.

DuyguA:
> In that case you'd have the entity ruler overwrite existing entities with the longer matches, where the match patterns potentially use `ENT_IOB`.

Yes.

> I don't think entity ruler should do anything other than set full entities.

Fair point! Is there another way around we can implement this feature?

adrianeboyd:
Where is this coming from?

DuyguA:
> Where is this coming from?


It's an Ora ticket :blush:

adrianeboyd:
But that's not what the ticket says. Let's take this discussion off this PR...

PR #9622, opened 2021-11-04
Make ent_iob compatible to rest

### LINKED ISSUE #4790: token.ent_iob should be writable

DuyguA:
## Description
Fixes #4790. We make `ent_iob` writeable.

### Types of change
Minor enhancement

## Checklist


adrianeboyd:
Sorry, we still don't want to do this. Users should use `Doc.set_ents` instead.

PR #9629, opened 2021-11-05
Migrate regression and other tests to the new pytest marker

ljvmiranda921:
We now have a new @pytest.mark.issue decorator that we can use to label tests related to a specific issue on the issue tracker.

## Description

We now have a new @pytest.mark.issue decorator that we can use to label tests related to a specific issue on the issue tracker. I added this marker to all regression tests and others that reference a specific issue

### Types of change

Added pytest markers

## Checklist


ljvmiranda921:
Lol it's a lot :O will probably automate the boring stuff haha

spacy/tests/regression/util_add_marker.py
```diff
@@ -0,0 +1,32 @@
+import re
```
ljvmiranda921, line 1: I am not sure if this should even be committed... 🤔

ljvmiranda921 (commented):
A few comments

spacy/tests/regression/util_add_marker.py
adrianeboyd: No, this is only going to be run once, so I don't think it should be committed.

ljvmiranda921: Removed in https://github.com/explosion/spaCy/pull/9629/commits/909177589dcdbde1cd4770f9f744d4d57d08d7e0

svlandeg (approved):
Thanks for taking this on Lj, looks good!

A related task we want to do, is move most of the tests out of `/regression/` and add them to the modules where they belong, but we could do this as a separate PR?

ljvmiranda921:
Hi @svlandeg , awesome thanks! Yeah sure, been meaning to do that in another PR. Might take some time though to sift through each regression. Will do that as well! 💯

PR #9638, opened 2021-11-07
Make Jsonl Corpus reader path optional again

polm:
## Description

In #8396 the path arg to the JsonlCorpus was made optional. This was done so that you could use a config with a pretraining section for training without including the raw text path. However, #9167 made the path non-optional again, apparently inadvertently. 

### Types of change

minor fix

## Checklist

### DIFF
--- a/spacy/training/corpus.py
+++ b/spacy/training/corpus.py
@@ -41,7 +41,7 @@ def create_docbin_reader(
 
 @util.registry.readers("spacy.JsonlCorpus.v1")
 def create_jsonl_reader(
-    path: Union[None, str, Path], min_length: int = 0, max_length: int = 0, limit: int = 0
+    path: Optional[Union[str, Path]], min_length: int = 0, max_length: int = 0, limit: int = 0
 ) -> Callable[["Language"], Iterable[Example]]:
     return JsonlCorpus(path, min_length=min_length, max_length=max_length, limit=limit)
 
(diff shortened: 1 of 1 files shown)


spacy/training/corpus.py
```diff
@@ -41,7 +41,7 @@ def create_docbin_reader(
 
 @util.registry.readers("spacy.JsonlCorpus.v1")
 def create_jsonl_reader(
-    path: Union[str, Path], min_length: int = 0, max_length: int = 0, limit: int = 0
+    path: Union[None, str, Path], min_length: int = 0, max_length: int = 0, limit: int = 0
```
svlandeg, line 44:
```suggestion
    path: Optional[Union[str, Path]], min_length: int = 0, max_length: int = 0, limit: int = 0
```

svlandeg:
Ah, this was probably an oversight on the mypy PR, yes. To make the tests succeed, you'll have to make sure that the `path` argument of the `JsonlCorpus` constructor matches this change though. And let's use `Optional` instead of `Union[None,...]` as we do in other places of the code base.

svlandeg:
Oh I think you pushed the edit to the wrong PR 🙈 https://github.com/explosion/spaCy/pull/9639

polm:
Gah, thanks for catching that! Should be fixed now.

svlandeg (approved):

PR #9639, opened 2021-11-07
Clarify how to fill in init_tok2vec after pretraining

polm:
## Description

This adds a note explaining that if you fill in `init_tok2vec` before actually doing pretraining it will fail because the file doesn't exist. This came up in #9626.

This seems like kind of unsatisfying workflow. Maybe we should just allow paths to files that don't exist here or something?

### Types of change

docs change

## Checklist


website/docs/usage/embeddings-transformers.md
```diff
@@ -714,10 +714,13 @@ subnetwork** within one of your components, and add additional layers to build a
...
-Pretraining produces a **binary weights file** that can be loaded back in at the
-start of training, using the configuration option `initialize.init_tok2vec`.
-The weights file specifies an initial set of weights. Training then proceeds as
-normal.
+Pretraining produces a **binary weights file** that can be loaded back in at
+the start of training, using the configuration option
+`initialize.init_tok2vec`.  The weights file specifies an initial set of
+weights. Training then proceeds as normal. Note that if you fill this value in
+before running pretraining, the command will fail because the file doesn't
+exist yet. You can just update your config file after pretraining finishes, or
```
svlandeg, line 722: This is kind of unfortunate because we tell people to use the exact same config for training and pretraining :( Not sure how better to deal with this though. At which point does it crash when providing an unexisting pretraining path during training, and can we prevent that from happening?

polm: It gives a ConfigValidationError from [this line](https://github.com/explosion/spaCy/blob/24cdd4c88e291d01aff9cd92c37c6cdb03479729/spacy/training/initialize.py#L186). I don't think that line currently has access to whether it's being called in pretraining or not, so the exception might need to be handled higher up.

Also, I think we technically support pretraining with seed tok2vecs. So you can start pretraining with the output of a different pretraining... If we support that, it's hard to know if a non-existing file is a mistake or a file that hasn't been created yet, though I guess we could just print a warning.

polm:
I added a commit that makes it so the `init_tok2vec` setting is ignored in the pretraining step. This accomplishes the goal of allowing the config for pretraining and normal training to be the same. However, if you wanted to use some vectors to initialize in the pretraining step, that would no longer be possible. I think this also means you can't resume training.

One nice thing about this is that the implementation was trivial. If we want to handle things closer to where the error is coming up, that will require a lot of changes - the error is happening inside `load_model_from_config`, which doesn't have any information about whether it's being called from inside pretraining or not, so passing the info down to where the error is being raised will require rewiring many layers of functions.

Another downside of this approach is that we actually can't guarantee the output model has the correct filename, since output paths for the pretraining vectors, like output models in normal training, are specified only on the command line, never in the config. That's not a big deal but I could see a user being confused by it.

Anyway, how do we want to handle this? I think the best options are:

1. Just document the weirdness and maintain current behavior
2. Ignore `init_tok2vec` in pretraining by default, but add a command line flag to allow its use

I would lean slightly towards 2 here because for most users I would expect it to be simpler. I would be more strongly in favor if we could guarantee the output filename is correct (maybe we could check it in the pretrain command?).

svlandeg:
> I added a commit that makes it so the init_tok2vec setting is ignored in the pretraining step. This accomplishes the goal of allowing the config for pretraining and normal training to be the same. 

Conceptually, I think the edit is fine: the pretraining shouldn't need the path for the pretrained model so it seems OK to hardcode this.

> However, if you wanted to use some vectors to initialize in the pretraining step, that would no longer be possible. 

What do you mean exactly?

> I think this also means you can't resume training.

I'm also unsure about this statement - if you want to resume pretraining, you'd use `resume-path` right?

svlandeg:
> I would be more strongly in favor if we could guarantee the output filename is correct (maybe we could check it in the pretrain command?).

Don't we do this already?

polm:
> However, if you wanted to use some vectors to initialize in the pretraining step, that would no longer be possible.

Suppose you do two different pretrainings, pretraining A and pretraining B. Because the tok2vec model is loaded as normal at the start of pretraining, you can specify an `init_tok2vec`. So you can run pretraining A, and the output file for that can be an input file to pretraining B, that produces a different `init_tok2vec` file. I don't think it makes any sense to do this but I think it's possible. 

> I'm also unsure about this statement - if you want to resume pretraining, you'd use resume-path right?

Ah, you're right - I forgot pretraining had special resume logic. So that's not an issue.

> I would be more strongly in favor if we could guarantee the output filename is correct (maybe we could check it in the pretrain command?).

> Don't we do this already?

We don't check that the `init_tok2vec` path is the same as the pretrain output path, and the output path is a directory and not a file anyway.

svlandeg:
> We don't check that the init_tok2vec path is the same as the pretrain output path, and the output path is a directory and not a file anyway.

Oh I see what you mean, but this PR doesn't really influence that? Like we didn't check that before either. We did check whether the path was valid, but not that it corresponded to the output dir mentioned on the command line.

I think I'd be in favour of just removing the value and keep it at that? That's assuming that "pretraining on a pretrained file" is not a use-case anyone is looking for, as we do have the resume mechanism for that. And then we could remove the warning from the docs again, right?

spacy/training/pretrain.py
```diff
@@ -31,6 +31,9 @@ def pretrain(
     allocator = config["training"]["gpu_allocator"]
     if use_gpu >= 0 and allocator:
         set_gpu_allocator(allocator)
+    if config["initialize"]["init_tok2vec"] is not None:
+        # TODO maybe print a warning?
```
svlandeg, line 35: Maybe not a warning, but a note in the [API docs](https://spacy.io/api/data-formats#config-initialize) that this setting is ignored during pretraining itself?

polm:
OK, if we're fine with removing the option for double pretraining then I can just clean this up and it should be good.

website/docs/usage/embeddings-transformers.md
svlandeg: I think we can remove this edit now, no?

polm: Good point! Let me remove that...

svlandeg:
Final question: do we see this as a bug fix and keep on `master`, or as a functionality change and move to `develop`?

polm:
I think we can put it in master because the previous behavior wasn't useful and we don't have any confirmed cases of someone using it.

PR #9649, opened 2021-11-09
Add ENT_IOB key to Matcher

DuyguA:
## Description

We add `ENT_IOB` as a token pattern key, upon request by #3940 . I also added a tiny addition to the documentation for the new key.

### Types of change

Enhancement

## Checklist


spacy/matcher/matcher.pyx
```diff
@@ -796,9 +798,14 @@ def _get_attr_values(spec, string_store):
                 attr = "ORTH"
             if attr == "IS_SENT_START":
                 attr = "SENT_START"
+            if attr == "ENT_IOB":
+                ent_iob = True
             attr = IDS.get(attr)
         if isinstance(value, str):
-            value = string_store.add(value)
+            if orig_attr == "ENT_IOB" and value in iob_strings:
```
adrianeboyd, line 805:
```suggestion
            if attr == ENT_IOB and value in Token.iob_strings():
```

Just use the attr `ENT_IOB` and `Token.iob_strings()` directly here, I don't think there's a real need to store them in additional variables. If it were code that got run while matching, perhaps, but this is just called when the pattern is added and most people won't be using `ENT_IOB` at all, anyway.

DuyguA: 1. OK, fair point. I killed `self._ent_iob-strings`
2.  `attr` turns into an integer after processing it. By the line where we process the `value` , it's an ID already:

```
attr = IDS.get(attr)
        
if isinstance(value, str):
```

That's why I restored the original attribute string.

adrianeboyd: You want to add an import for the attr `ENT_IOB` here:

https://github.com/explosion/spaCy/blob/24cdd4c88e291d01aff9cd92c37c6cdb03479729/spacy/matcher/matcher.pyx#L21

DuyguA: Oh kapish, I do it

spacy/matcher/matcher.pyx
```diff
@@ -796,9 +797,14 @@ def _get_attr_values(spec, string_store):
                 attr = "ORTH"
             if attr == "IS_SENT_START":
                 attr = "SENT_START"
+            if attr == "ENT_IOB":
+                ent_iob = True
```
adrianeboyd, line 801:
```suggestion
```

adrianeboyd:
I think it would be nice to provide better feedback for unsupported values, at least with `validate=True`. I think it should be possible to use `Union` with `Literal` for the values in the schema, and you can add tests for this in `spacy/tests/matcher/test_pattern_validation.py` for what should happen with `validate=True` and `validate=False`.

DuyguA:
> I think it would be nice to provide better feedback for unsupported values, at least with `validate=True`. I think it should be possible to use `Union` with `Literal` for the values in the schema, and you can add tests for this in `spacy/tests/matcher/test_pattern_validation.py` for what should happen with `validate=True` and `validate=False`.

Oki, I'll make some additions!

spacy/tests/matcher/test_pattern_validation.py
```diff
@@ -68,3 +68,20 @@ def test_pattern_errors(en_vocab):
...
+    doc = Doc(en_vocab, words=["I", "visited", "my", "friend", "Anna", "Maria"])
+    doc.ents = [Span(doc, 4, 6, label="PERSON")]
+    matches = [doc[start:end].text for _, start, end in matcher(doc)]
+    assert len(matches) == 0
+
+
+def test_iob_keys_invalid_value_validation(en_vocab):
+    matcher = Matcher(en_vocab)  # Test iob pattern with invalid value, error happens
+    with pytest.raises(MatchPatternError):
+        matcher.add("Rule", [[{"ENT_IOB": "XXX"}]])
```
adrianeboyd, line 87: Can you incorporate this pattern into a parameter for the first test instead?

spacy/schemas.py
```diff
@@ -209,6 +210,7 @@ class TokenPatternOperator(str, Enum):
 UnderscoreValue = Union[
     TokenPatternString, TokenPatternNumber, str, int, float, list, bool
 ]
+IobStringValue = Union[TokenPatternString, Literal[Token.iob_strings()]]
```
adrianeboyd, line 213: Are you sure this `Literal` works and doesn't mean the whole tuple as the literal? (I've barely used `Literal` before, I'm actually not sure myself.)

spacy/matcher/matcher.pyx
```diff
@@ -796,9 +796,13 @@ def _get_attr_values(spec, string_store):
                 attr = "ORTH"
             if attr == "IS_SENT_START":
                 attr = "SENT_START"
+                ent_iob = True
```
adrianeboyd, line 799:
```suggestion
```

(You can commit the suggestion directly from the comment if you want to. Sometimes this is easiest for minor edits.)

spacy/schemas.py
DuyguA: Yes indeed, I tested a bit on the shell and it worked :grimacing: 

Me too, I looked up from the documentation and examples with `List` and `Tuple` are OK

adrianeboyd:
For compatibility with older versions of python you need to import `Literal` from `spacy.compat` instead of from `typing`.

adrianeboyd:
The only remaining question is whether we want to update `intify_attrs` in this PR or in a separate one. Conceptually it's very close and it makes sense for the behavior to be kept in sync, but it is a bit separate from the particular `Matcher` issue.

DuyguA:
> The only remaining question is whether we want to update `intify_attrs` in this PR or in a separate one. Conceptually it's very close and it makes sense for the behavior to be kept in sync, but it is a bit separate from the particular `Matcher` issue.

Agree, I say we do it separately. It's not really related to the functionality of this PR. We can make another small PR.

svlandeg:
@DuyguA : Maybe prepare that additional PR and link it here, so we can merge them simultaneously?

DuyguA:
> @DuyguA : Maybe prepare that additional PR and link it here, so we can merge them simultaneously?

why not, I do now.

DuyguA:
> @DuyguA : Maybe prepare that additional PR and link it here, so we can merge them simultaneously?

Sure, here it is :  https://github.com/explosion/spaCy/pull/9738  :blush:

spacy/schemas.py
```diff
@@ -209,6 +210,8 @@ class TokenPatternOperator(str, Enum):
 UnderscoreValue = Union[
     TokenPatternString, TokenPatternNumber, str, int, float, list, bool
 ]
+IobString = Literal["", "I", "O", "B"]
+IobStringValue = Union[TokenPatternString, IobString]
```
adrianeboyd, line 214: Looking at this again, since the conversion isn't recursive and `IN` etc. aren't going to work, I think it would be better to just support a limited use of `ENT_IOB`:

```suggestion
IobValue = Literal["", "I", "O", "B", 0, 1, 2, 3]
```

I don't think this is all that limiting given that there are only three possible values anyway.

spacy/schemas.py
```diff
@@ -222,6 +225,7 @@ class TokenPattern(BaseModel):
     lemma: Optional[StringValue] = None
     shape: Optional[StringValue] = None
     ent_type: Optional[StringValue] = None
+    ent_iob: Optional[IobStringValue] = None
```
adrianeboyd, line 228:
```suggestion
    ent_iob: Optional[IobValue] = None
```

spacy/schemas.py
```diff
@@ -209,6 +210,8 @@ class TokenPatternOperator(str, Enum):
 UnderscoreValue = Union[
     TokenPatternString, TokenPatternNumber, str, int, float, list, bool
 ]
+IobString = Literal["", "I", "O", "B"]
```
adrianeboyd, line 213:
```suggestion
```

spacy/schemas.py
DuyguA: OK, sure. I was not sure about numbers though, I thought users most probably wouldn't wanna match with number values, letters are more straightforward. Still, I'll add it.

PR #9650, opened 2021-11-09
Add option to batch on the number of characters to Language.pipe

baxtree:
## Description

This PR adds an option to `nlp.pipe` for creating batches based on the total number of characters from the enclosed texts or docs. If present, it averages batch workload and makes memory usage more predictable especially when the text lengths vary significantly across the input stream. 

### Types of change

A new optional argument`batched_chars_threshold` is introduced. If present, it will make sure a multi-item batch contains characters no more than its value. It does not fragment the text or the doc when the number of characters is above the threshold so batching on that will always result in a single-item batch.

## Checklist


adrianeboyd:
This is a reasonable thing to want to do, but I suspect this is pointing to the need for a `pipe_batcher` or similar setting on `[nlp]` instead of an additional hard-coded option, hmm...

baxtree:
Yes. I saw similar batching logic defined in [batchers.py](https://github.com/explosion/spaCy/blob/master/spacy/training/batchers.py). Are they used by training only atm?

adrianeboyd:
There's `[training.batcher]` but not `[nlp.batcher]`.

baxtree:
Oh I see. Another question is does [this block](https://github.com/explosion/spaCy/blob/24cdd4c88e291d01aff9cd92c37c6cdb03479729/spacy/language.py#L1645-L1647) need to be changed if batches vary in the number of enclosed items. Although the tests are passing, it's not clear to me what the consequence is when batch_size is set to 1.

baxtree:
Tried to make `batched_chars_threshold` configurable but found `batch_size` on Language.pipe is also a hard-coded option. Assuming a new `[nlp.batcher]` had abstracted both of them away, would function `_multiprocessing_pipe` end up with a facelift or completely rewriting?

svlandeg:
Hi @baxtree, apologies for the late follow-up, and thanks for this PR!

I agree with Adriane that a `batcher` on `[nlp]` would make sense here. It would change some of the API though, so this would need to go on the `develop` branch. If you've been experimenting with this solution, maybe it's worth opening another PR so we can decide on the best way forward?

baxtree:
Hi, @svlandeg . Agreed and I have updated the base branch accordingly after rebasing `develop` onto my branch. There is seemingly no need to open a new PR but do let me know if that's necessary.

svlandeg:
Either way is fine by me - I just assumed that the `nlp.batcher` option would look quite different than the current PR (hardcoded options).

baxtree:
That's why I am reluctant to make any drastic change on `nlp.pipe`. Its existing implementation has been bullet tested in production by the wider community and sure, new tests could be created but they may not give high enough confidence IMHO.

svlandeg:
Hi @baxtree: while that's true, we also don't want to move forward with the implementation in this specific PR, as we'd prefer a more generic solution. We'll have to figure out what to do with the multiprocessing though, but adding more and more options into the `pipe` function definition doesn't feel like the way to go. Let us know whether you'd prefer starting a new PR or working from this one, or if you don't have time to rework the functionality that's perfectly understandable too, ofcourse!

baxtree:
Hi @svlandeg , Fair enough. I have no preference and think working from this existing one may be fine coz people can see how the issue/implementation evolves based on which requested change and in a better case make contributions to it. I feel it will be more sensical if core developers not only make the best wish on PRs but also give some tangible guidance on refining them, although quite understandably this depends on you guys' bandwidth too.

svlandeg:
> I have no preference and think working from this existing one may be fine coz people can see how the issue/implementation evolves based on which requested change and in a better case make contributions to it.

Sure - I'll put the PR in draft to signal it's a work in progress.

>  I feel it will be more sensical if core developers not only make the best wish on PRs but also give some tangible guidance on refining them, although quite understandably this depends on you guys' bandwidth too.

I see what you mean, it can be frustrating to try and write code that passes a review process if you don't know upfront what the reasoning is underlying that review process / judgement. On the other hand, as software development goes, it's often difficult to imagine up-front how a change will work out across the whole code base after running all tests etc. We often have to iterate on our own PRs too, and often have to circle back to previous ideas when something doesn't work out, etc. 

We definitely do appreciate your contributions, and I hope you haven't found our reviewing process too frustrating. I'll leave this PR open and in draft in case you want to continue working on it. If not, we can close it and we will record the underlying issue in our internal backlog.

svlandeg:
I think we should pick this up at a later point in time and work on a more general solution for the problem. Closing for now.

baxtree:
In that case we can live with a local patch to spaCy coz balancing load across workers is still crucial as per the magnitude of our text notes. Do hope one day the core developers will crystalise the code design so the patch can be happily eliminated.

PR #9659, opened 2021-11-11
Add doc_cleaner component

adrianeboyd:
## Description

Add `doc_cleaner` component

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/pipeline/functions.py
+++ b/spacy/pipeline/functions.py
@@ -155,7 +155,7 @@ class DocCleaner:
     def __call__(self, doc: Doc) -> Doc:
         attrs: dict = self.cfg["attrs"]
         silent: bool = self.cfg["silent"]
-        for attr, value in attrs:
+        for attr, value in attrs.items():
             obj = doc
             parts = attr.split(".")
             skip = False
--- a/spacy/pipeline/functions.py
+++ b/spacy/pipeline/functions.py
@@ -1,6 +1,8 @@
 from typing import Dict, Any
 import srsly
+import warnings
 
+from ..errors import Warnings
 from ..language import Language
 from ..matcher import Matcher
 from ..tokens import Doc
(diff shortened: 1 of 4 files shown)


spacy/pipeline/functions.py
```diff
@@ -136,3 +138,65 @@ def from_disk(self, path, **kwargs):
...
+
+@Language.factory(
+    "doc_cleaner",
+    default_config={"attrs": {"tensor": None, "_.trf_data": None}, "silent": True},
+)
+def make_doc_cleaner(nlp: Language, name: str, *, attrs: Dict[str, Any], silent: bool):
+    return DocCleaner(attrs, silent=silent)
+
+
+class DocCleaner:
```
honnibal, line 151: Are we sure we don't want to inherit from `Pipe` here?

adrianeboyd: Hmm, I just basically copied this from `token_splitter` above. I guess whatever we choose we should do for both?

svlandeg: I guess there's an unclear line between what are helper functions/components or actual proper pipeline components. To me it feels like anything in `spacy.pipeline` should probably inherit from `Pipe`, but maybe not the type of functions that we have here. It doesn't really add much to inherit from `Pipe` either.

svlandeg: There's something else that's drawn my attention though: currently `language.add_pipe` is typed to return  `Pipe`, ever since the mypy PR. Shouldn't this be `Callable[[Doc], Doc]`? Or should we have a protocol for this, just specifying the `__call__` signature? (maybe the latter is overkill)

adrianeboyd: I also don't think it makes a difference to inherit from `Pipe`, although maybe it's slightly better to officially have `is_trainable`.

For components like these the main thing that would be nice is not to have to keep reimplementing the same kind of `cfg` serialization, but it's probably not enough of an issue to want to do anything in the core library about it.

website/docs/api/pipeline-functions.md
```diff
@@ -130,3 +130,25 @@ exceed the transformer model max length.
 | `min_length`   | The minimum length for a token to be split. Defaults to `25`. ~~int~~ |
 | `split_length` | The length of the split tokens. Defaults to `5`. ~~int~~              |
 | **RETURNS**    | The modified `Doc` with the split tokens. ~~Doc~~                     |
+
+## doc_cleaner {#doc_cleaner tag="function" new="3.2.1"}
+
+Clean up `Doc` attributes. Intended for use at the end of pipelines with
+listeners (`tok2vec`, `transformer`) that store tensors and other values that
+can require a lot of memory and frequently aren't needed after the whole
+pipeline has run.
```
svlandeg, line 139: It's not the listeners setting these attributes though? Like, a transformer-based component with inline transformer would run into the same mem usage and would also benefit from this cleaner component, no?

adrianeboyd: I'm pretty sure it's actually just the separate pipeline components that set these. They don't really need to be saved otherwise, at least for the pipeline to work. I'll try it out to double-check, though...

I guess this would be a separate issue with an inline transformer not providing the info later like you might expect. And actually, looking at the `transformer` component again, the `DOC_EXT_ATTR` is ignored for the actual setting part, that's not great.

adrianeboyd: I checked, it's really only from the full pipeline components.

adrianeboyd: So for `en_core_web_sm` the `doc.tensor` is from `tok2vec` and not related to the internal `ner` `tok2vec` that ran second.

svlandeg: Ah, you're right, ofcourse, it's the `tok2vec` and `transformer` components causing the noise. I'd rephrase this explanation, because this happens regardless of the fact whether there are listeners in the pipeline. While probably unuseful, somebody could have a `transformer` component in the pipeline without a corresponding listener. So maybe something like 

```suggestion
Clean up `Doc` attributes. Intended for use at the end of pipelines with
`tok2vec`/`transformer` pipeline components that store tensors and other values that
can require a lot of memory and frequently aren't needed after the whole
pipeline has run.
```
(+ formatting)

adrianeboyd: Sure, that's fine.

Also unrelated, but it's also really hard to make a pipeline with just a transformers because there are some automatic cleanup steps that remove unused listened-to things. (I find it frustrating because you can't play with just a transformer from `init config`.)

Issue #9575, opened 2021-10-31
EntityLinker encounters IndexError if token length of Example predicted and reference Docs aren't equal


kinghuang:
I've been having trouble training an entity_linker component. It's raising an uncaught `IndexError` getting the `kb_id` for an `ent`.

<details>
<summary>IndexError: list index out of range</summary>

```
⚠ Aborting and saving the final best model. Encountered exception:
IndexError('list index out of range')
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.9/site-packages/spacy/__main__.py", line 4, in <module>
    setup_cli()
  File "/usr/local/lib/python3.9/site-packages/spacy/cli/_util.py", line 69, in setup_cli
    command(prog_name=COMMAND)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/typer/main.py", line 500, in wrapper
    return callback(**use_params)  # type: ignore
  File "/usr/local/lib/python3.9/site-packages/spacy/cli/train.py", line 63, in train_cli
    train(nlp, output_path, use_gpu=use_gpu, stdout=sys.stdout, stderr=sys.stderr)
  File "/usr/local/lib/python3.9/site-packages/spacy/training/loop.py", line 122, in train
    raise e
  File "/usr/local/lib/python3.9/site-packages/spacy/training/loop.py", line 105, in train
    for batch, info, is_best_checkpoint in training_step_iterator:
  File "/usr/local/lib/python3.9/site-packages/spacy/training/loop.py", line 203, in train_while_improving
    nlp.update(
[...]

PR #9669, opened 2021-11-13
Fix entity linker batching

polm:
## Description

The Entity Linker doesn't work with Listeners because it violates an (implicit?) requirement that the batch passed to the pipeline `update` must be the same as the batch passed to the model's `forward` using `begin_update`. 

The reason the Entity Linker does this is that it builds sentence docs to get context around relevant entities. The reason this wasn't obviously a problem is that the EntityLinker predates the use of Listeners in spacy.

This PR changes the structure of the EntityLinker so that the batch is handled consistently with other components and works with Listeners. This moves the logic of building context inside the model. 

There are a few issues with this that are still unresolved. The main one is that the EntityLinker only makes a prediction if entities are available. This has two issues. 

1. What happens when no entities are available? This can happen with unlucky batching if enough training docs don't have entities. With no entities some of the calculations just don't work out. Maybe we need to provide some filler data?
2. How do we get the entities? The old code checked gold data for entities and used that to create candidates. In the model we don't have Examples, just Docs without gold data. So we need to either have the Entities set earlier in the pipeline, or copy them from gold docs. Which do we do, and do we make that a config option, or partly automate it, or something else?

With the first commit here, training an NER and NEL model together using listeners is possible using [this user-provided test repo](https://github.com/Joozty/ner_nel_shared_embed_layer). But in order for this to work entities are unconditionally copied over from reference Docs when preparing the batch, which may not be desirable when training with NER, and is not really what we want for a final fix.

### Types of change

Bug fix. Brought up in these issues: #9310, #9291. 

#9575 seems to be a separate problem but it might make sense to fix it in this PR too.

## Checklist


spacy/ml/models/entity_linker.py
```diff
@@ -1,25 +1,27 @@
...
 
 from ...util import registry
 from ...kb import KnowledgeBase, Candidate, get_candidates
 from ...vocab import Vocab
 from ...tokens import Span, Doc
+from ..extract_spans import extract_spans
+from ...errors import Errors
 
 
 @registry.architectures("spacy.EntityLinker.v1")
```
svlandeg, line 15: We probably need to bump this version? (and move v1 to legacy)

svlandeg (commented):
Thanks for looking into this Paul!

About the listener architecture: I think your proposed changes make sense (I'll look into a bit more detail when this is out of draft), though in general we should think about what should happen in these cases. As you point out, there is an implicit assumption here on how listeners should be used. We should either document that, or look into a kind of fall-back mechanism when the listener's cache fails. We can probably defer that to a future discussion/PR though.

About the entities: conceptually, I think we should allow either training the EL on gold entities (as before), or on predicted ones. The latter is perfectly possible with the recent "annotating components" feature. The EL should then take in a parameter that defines whether gold should be used (`True` by default) or no. In the latter case, the user needs to put the NER or any other component that writes to `doc.ents` in `annotating_components`. We can't really check this, though we can output a warning if `annotating_components` is empty at least.

polm:
> The EL should then take in a parameter that defines whether gold should be used (True by default) or no. 

I hadn't considered the possibility of adding a parameter, that sounds great. I'll work on doing that and I think with that change this can be out of draft.

For the broader issue, I think a doc note can be added on the current behavior pretty easily, and we can discuss whether the current requirements are desirable or not later.

polm:
Oh, regarding moving the previous version to `spacy-legacy` - the pipeline code has to change to work with this, so the old model code by itself isn't enough to be usable. How should that be handled?

spacy/ml/models/entity_linker.py
polm: Good point. But because the pipeline code also has to change, the v1 model by itself won't work, so how should that be handled?

svlandeg: Ah, right. Hm. What we can do is to keep v1 in spaCy's codebase for now and raise an exception explaining users need to move to v2 for the new code. We can't move the function and raise this error from within `spacy-legacy`, because then people wouldn't be able to update `spacy-legacy` without updating `spacy`, and we'll get entangled in version misery.

svlandeg: Make sure to go through the code base and update docs & defaults to v2 too.

polm: To be clear, when you say there should be a stripped v1, should I retain the pipeline code to make that functional? Or should it be a non-functional stub that tells people v1 doesn't work?

polm: For now I created a stub v1 that throws an error and shows a short explanation and links to this PR.

svlandeg: Yea I think that makes sense, I don't see a better way to do it. This probably means we need to move this PR to 3.3 (`develop`) though, because it will break people's config files.

svlandeg: @adrianeboyd: do you see a better way? I don't think we've encountered this before, that we actually have to retire a previous version of a registered architecture.

adrianeboyd: There are quite a few older architectures in `spacy-legacy` and I don't see why you can't also include the older component? It might require a lot of copying to do it cleanly (plus all the registry substitutions), but I think that would be better than removing it in a breaking way, even if it's broken/buggy.

polm: So the solution would be to
 
1. copy the old component code somewhere in the component (not in legacy?)
2. In the component, check if the architecture is v1 and use the old component code if so

Is that right? It feels a little inside out and it'll leave a lot of junk around but it's not too tangled.

adrianeboyd: No, I think an entirely independent copy should be in `spacy-legacy` if there's no other way to preserve it.

adrianeboyd: I think we have made minor changes to the parser algorithm across minor versions without the whole copying rigamarole. So I guess doing it over a minor version (also with the architecture version stuff) is the other option.

svlandeg: Oof, that feels like quite the hassle, but you're right that at least it wouldn't be breaking. I'm just not sure whether it's all worth it, because the v2 is just the v1+bugfix, it's not like you're losing anything by removing it. People will just have to change the `1` in their config file to a `2`...

polm: So this does change a lot of code, but since the architecture isn't changing and it really is just a bugfix, could we just skip raising the version entirely? The only change in behavior from outside should be that listeners work now.

polm: Actually, we had to change the config for this, so that's reason enough we can't use the same version I guess. The good news (?) is that besides changing the version number, users can explicitly control where NER annotations come from now.

polm:
Failing tests are because for tests that use the default `get_candidates` for initialization, there are no entities on the example docs. Because there are no entities, no candidates for annotations are produced, and initialization fails. 

In initialization specifically we can check if there are no NER annotations and add some fake ones so the network can run, I'm not sure if that's the right approach though.

spacy/ml/models/entity_linker.py
polm: Going over actually implementing this I am still kind of confused about what this would look like. Assuming the v1 component code is in spacy-legacy, how is it actually accessed? 

1. The non-legacy update/init methods check if the architecture is v1 and change their behavior, importing from legacy (simple implementation but very ugly)
2. Legacy registers a new component, `entity_linker.v1` or something, and users have to update their configs. This is a little weird since components don't usually have version numbers. There is also a question of whether the whole pipeline should be copied over, or whether it should be subclassed and have select methods overridden (probably the latter?). 

Moving the component code inside the model in spacy-legacy doesn't work because the model only receives Docs, not Examples. 

I guess option 2 here is preferable, but since this is the first time we've put a component in legacy we need to think about not just this case but future ones too.

svlandeg: I wouldn't go for option 2 because if users need to update their configs then we might as well tell them to move to v2 :| 

(Honestly I'd still prefer just making v1 obsolete)

polm: Note that as a result of discussion, we decided to keep the v1 component in spacy core in a new legacy area. The old component will be swapped in based on the model, as even if that's messy it's not disruptive to existing users.

polm:
So the tests have been failing because the changes in architecture affect the loss calculation too. 

In the old version of the component, there would be one Doc per gold entity with KB entry in the input to the loss function. The loss function calculates an embedding for each gold entity and compares those. 

The issue is that in the revised version, the model doesn't know what's gold and not, it just uses all entities. This means that it gets predictions for NIL entities, which the loss function doesn't calculate a value for. So that has to be resolved somehow. 

I guess these are the approaches:

1. In the loss calculation, ignore predictions for known NIL entities (my most recent commit tries this)
2. In the loss calculation, add empty/placeholder embeddings for known NIL entities (I tried this but the loss was too high for the test)
3. In training, remove entity annotations for gold entities with no KB ID (not tried yet)

I tried 2 first, but the loss was too high and the test consistently failed.

My most recent commit takes approach 1 and it passes - but only sometimes. Sometimes the loss is a little too high, and more rarely the loss goes to NaN. I'm not sure why it's not reproducible or why NaN would come up - seems there's some kind of overflow?

```
spacy/tests/pipeline/test_entity_linker.py::test_overfitting_IO
  /mnt/pool/code/spacy/env/lib/python3.9/site-packages/thinc/layers/layernorm.py:32: RuntimeWarning: overflow encountered in multiply
    d_xhat = N * dY - sum_dy - dist * var ** (-1.0) * sum_dy_dist

spacy/tests/pipeline/test_entity_linker.py::test_overfitting_IO
  /mnt/pool/code/spacy/env/lib/python3.9/site-packages/numpy/core/_methods.py:47: RuntimeWarning: invalid value encountered in reduce
    return umr_sum(a, axis, dtype, out, keepdims, initial, where)

spacy/tests/pipeline/test_entity_linker.py::test_overfitting_IO
  /mnt/pool/code/spacy/env/lib/python3.9/site-packages/numpy/core/fromnumeric.py:87: RuntimeWarning: invalid value encountered in reduce
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)

spacy/tests/pipeline/test_entity_linker.py::test_overfitting_IO
  /mnt/pool/code/spacy/env/lib/python3.9/site-packages/thinc/backends/ops.py:755: RuntimeWarning: invalid value encountered in multiply
    gradient *= threshold / grad_norm

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

I am not sure about approach 3, I need to think about it more. I guess the model can use the `get_candidates` function to exclude entities from predictions, so maybe that's the right thing to do.

polm:
Thinking about approach 3 some more, I don't think it'll work. The model doesn't know if an entity has a KB annotation or not (because that would imply cheating). It also doesn't currently have access to the get_candidates function, so it can't check that (the function is only in the component). 

What we could do is, the component could "erase" entities that don't have KB annotations, or that would get NIL annotations... but I'm not sure how that would interact with other pipeline components. I guess we could restore entity annotations but then that gets messy. 

I think there's nothing wrong in principle with the approach I'm taking now, though I'm not sure what's up with the math - even if it's wrong, I'd expect it to be consistent on CPU.

spacy/ml/models/entity_linker.py
```diff
@@ -29,6 +31,52 @@ def build_nel_encoder(
...
+            cands.append((start_token, end_token))
+
+        candidates.append(ops.asarray2i(cands))
+    candlens = ops.asarray1i([len(cands) for cands in candidates])
+    candidates = ops.xp.concatenate(candidates)
+    outputs = Ragged(candidates, candlens)
+    # TODO W093 if there's no data. Can we do that here though?
+
+    # because this is just rearranging docs, the backprop does nothing
+    return outputs, lambda x: []
```
svlandeg, line 77: But wait - if we don't backprop back, the `tok2vec` layer won't receive gradients and won't be able to tune itself towards the EL task?

spacy/pipeline/entity_linker.py
```diff
@@ -212,10 +235,23 @@ def initialize(
...
-            doc_sample.append(example.x)
+            doc = example.x
+            if self.use_gold_ents:
+                doc.ents = example.y.ents
+            doc_sample.append(doc)
             vector_sample.append(self.model.ops.alloc1f(nO))
         assert len(doc_sample) > 0, Errors.E923.format(name=self.name)
         assert len(vector_sample) > 0, Errors.E923.format(name=self.name)
+
+        # XXX This is a hack. Also `has_annotation` doesn't seem to work as expected?
```
svlandeg, line 246: What is not working exactly with `has_annotation`?

spacy/pipeline/entity_linker.py
```diff
@@ -212,10 +235,23 @@ def initialize(
...
+        # Check that there are some entities in the batch
+        has_annotations = any([doc.ents for doc in doc_sample])
+        if not has_annotations:
+            doc = doc_sample[0]
+            ent = doc[0:1]
```
svlandeg, line 251: What if this `doc` happens to be empty (ie no tokens)? Can you write a unit test for this edge case?

svlandeg (commented):
Thanks for the work on this Paul. We need to dig into the `nan` issue - I left some review comments. I think the key is in the missing backprop.

In general I'd like to see more unit tests here: test edge cases with no entities (predicted or gold), test with different scenarios of the `use_gold_ents` etc. It would also be good to have a specific test for both the legacy v1 as well as the new v2 architecture, to ensure both get initialized well, can train one or two steps, and can predict. The test can be written by loading an `nlp` from a minimal config that is filled with defaults depending on either architecture.

Some additional comments to your questions:

> What happens when no entities are available? This can happen with unlucky batching if enough training docs don't have entities. With no entities some of the calculations just don't work out. Maybe we need to provide some filler data?

Is there a way to spot this very early on, like we do have empty docs, and prevent the neural network from predicting at all?

>I guess these are the approaches:
> 1. In the loss calculation, ignore predictions for known NIL entities (my most recent commit tries this)
> 2. In the loss calculation, add empty/placeholder embeddings for known NIL entities (I tried this but the loss was too high for the test)
> 3. In training, remove entity annotations for gold entities with no KB ID (not tried yet)

The problem is that the EL model doesn't try to learn the ID for a given entity, it tries to get the embedding of the sentence as close as it can to the entity embedding. There can be no universal NIL embedding - that will mess up normal embeddings too much. So 2. won't work. 

I think 1. should work though: similar to how we would mask missing information when calculating textcat loss, we should mask the NIL entity and not derive loss values from it.

spacy/ml/models/entity_linker.py
polm: I believe that's not relevant here because this is parallel to the tok2vec, not after it. 

```
        model = (
            ((tok2vec >> list2ragged()) & build_span_maker())
            >> extract_spans()
            >> reduce_mean()
            >> residual(Maxout(nO=token_width, nI=token_width, nP=2, dropout=0.0))  # type: ignore[arg-type]
            >> output_layer
        )
```

The `&` is tuplify, so `extract_spans` will backprop to `tok2vec >> list2ragged()` as well as the span maker. But the span maker has no statistical anything - it just generates indices for `extract_spans()` - so it doesn't need to do backprop.

Does that make sense or am I missing something?

svlandeg: Oh yea you're probably right! Hm that still leaves us with the issue of the `nan`'s though. When you print out the internal values at the various input/output steps of the different layers, you can often see 0's creeping into the vectors, eventually zero'ing the entire network.

spacy/pipeline/entity_linker.py
polm: I added a test and check for when the batch has no usable gold examples, which will also cover the no tokens case.

spacy/pipeline/entity_linker.py
polm: I don't remember exactly, so that comment isn't good enough :disappointed: I think that I misunderstood whether it worked on a per-doc or per-pipeline basis.

polm:
I think this should go green this time.

The issue was with the gradient for entities not in the kb. A gradient can't actually be calculated for them, and the model was only returning gradients it could calculate. But because the model doesn't have information about which entities are in the kb, that means there could be a mismatch in the number of calculated embeddings and the size of the gradient.

A mismatch like that should probably cause an error, but it seems it may have been accessing out of bounds memory or doing something else weird. So that's where the nans came from and is another problem that needs a followup.

polm:
Test failure is about garbage collection in textcat...? Might need to update from master, will look more at this later.

svlandeg:
That failing spancat test is really weird. I see that there are some commits added to this branch that aren't supposed to be there. Maybe rebase to the current `master`?

polm:
Test still fails here, but it passes locally for me. Not sure what's up... (Also you're right it's spancat, I mistakenly said textcat before.)

polm:
Looking at the failing test in detail, it looks like it is flaky in master - I ran it 20 times on master and it failed 4 times. Not really sure what's going on here...

svlandeg:
That's really weird. Could you create a seperate PR, disable the flaky test, and create a follow-up ticket on our internal board to look into this in the ner future? Then we can continue with this PR after the test has been disabled on master.

adrianeboyd:
This looks bizarre. My bet is new version of numpy (1.22.0), but we should figure out why.

polm:
OK, I have rebased on master that has the test skip and force-pushed to run the tests again. Let's see if this fixes it...

polm:
It's green! :tada: 

I checked the changes to spacy-legacy and found some issues with them, but those should be resolved now - with this branch and the spacy-legacy one I was able to train the `nel_emerson` example project. Is there anything else that needs checking?

spacy/pipeline/entity_linker.py
```diff
@@ -90,6 +93,22 @@ def make_entity_linker(
         produces a list of candidates, given a certain knowledge base and a textual mention.
     scorer (Optional[Callable]): The scoring method.
     """
+    # Handle legacy model
+    if model.name == "spacy.EntityLinker.v1":
```
svlandeg, line 96: How did you verify that this works? I don't really see how it could, the `model.name` is never set to that specific string as far as I can see.

svlandeg (changes requested):
I added a failing test for the v1 legacy architecture, I tested this locally with the corresponding legacy branch, but it fails because `EntityLinker_v1` is never returned because `model.name` isn't matched.

Similarly, when running the NEL emerson project, the v2 architecture is returned. It looks like the new v2 code just runs in that case, probably because the data is extremely simple. But it seems to run more by coincidence because the legacy fallback isn't actually used.

spacy/pipeline/entity_linker.py
polm: I think I completely overlooked this - it's clear looking at it now it won't work at all, since `model.name` is the automatically constructed Thinc name. 

However looking over this I'm not sure the architecture name is available at all at this point. It's not in the model, it's not in the arguments here, and `nlp.config` doesn't seem to contain the config for the current component (and peeking at that would be nasty anyway). 

I need to think about this more, but can we even check the architecture here?

polm: My solution for the issue here is to look for the name of the new custom component that pulls out the spans in the thinc model name. If it's not there, I assume it's the old model. 

That feels awfully messy, but I think it's the cleanest way to check things here, since the model name isn't actually available.

polm:
Still working on this - in order to get `use_gold_ents = False` working I had to redo the code that aligned predictions and gold KB embeddings, since it relied on the assumption that there was exactly one prediction per gold entity. I figured it out but need to clean it up and write tests for situations where the alignment actually comes up.

polm:
So this works, and I have tests for it, but the code is kind of a mess. 

The issue is that I need aligned entities but I'm not sure that any of the existing alignment functions do what I need. The issue is that the model produces an Entity Linker prediction for each entity on the predicted document. Since those predictions are just a list of predicted embeddings, in order to filter them it's not enough to get the aligned entities, the index relative to the predicted embeddings is necessary. 

I think I can make this cleaner by getting the aligned entities before the forward pass, setting them on the predicted doc temporarily, and then restoring the predicted entities, but I need to check it. If that works it'll have the nice side effect of not doing calculations in the forward pass for instances that can't be backpropagated.

spacy/pipeline/entity_linker.py
svlandeg: It would work for now, but the solution does feel a bit brittle & implicit, making it prone for bugs in the future.

I'm almost tempted to define a new `attr` that would record the factory, maybe something like
```
model.attrs["factory"] = {"EntityLinker": 2}
```
And then any internal check could verify the version, and if the right attribute is not available, assume it's below whatever you're checking, because it's a legacy one. In time I think this might be the most robust/clear solution, but it would require maintenance and making sure things stay consistent whenever we introduce a new model version...

polm:
OK, I have added a `get_matching_ents` function that returns only entities that are aligned between the predicted and reference docs in a function. That's the set of entities that we can backprop for. Now I have some design questions about this.

1. Is `get_matching_ents` a reasonable addition to the Example? Is there a better name for it? Not sure how to differentiate it from "aligned" ents more clearly...
2. When getting matching ents in the specific case of the EntityLinker, should it be necessary that labels of ents match as well as boundaries? My instinct is "no", but I don't have a strong justification for that. We could make it a parameter for the user but I feel like it's not a very interesting one.
3. There is a situation with the internal state in the EntityLinker that I don't know how to test for. If alignment is done incorrectly but the number of entities matches between predicted and gold docs, then the math will work and training can run but the results will be meaningless.

polm:
Tests are now failing because `EntityLinker.v1` isn't available in the registry, because the PR adding it to legacy needs to be coordinated with this. 

I'm not really sure how to handle that... We could do a dev release of `spacy-legacy` and have this PR branch use it but that seems like too many moving parts, and it'll have to be ripped out anyway. I could put in a placeholder but then the test isn't meaningful. I guess we could also just xfail it until `spacy-legacy` is updated...?

svlandeg:
> I'm not really sure how to handle that... We could do a dev release of `spacy-legacy` and have this PR branch use it but that seems like too many moving parts, and it'll have to be ripped out anyway. 

Why should we have to pull it? IIRC there's no issue when a function is declared in both `spacy` and `spacy-legacy` - the former should take precedence. And either way they should be the same thing, right?

Actually, you can go ahead and test locally whether this gives any issues when you have https://github.com/explosion/spacy-legacy/pull/18 installed next to spaCy's `master` branch, and then try running some code with v1. Should hopefully be fine 🤞

polm:
I have been testing it locally - with the patched version of `spacy-legacy` it works fine. 

I thought that maybe using a special release of `spacy-legacy` in this PR branch would be too complicated, but on reflection I guess there's no issue with it - I may have misunderstood how resolution between `spacy` and `spacy-legacy` registries worked. 

If it's not too complicated we can do the test release and update the dependencies in this branch.

svlandeg:
Sorry, I should have been more clear. Ideally, we can just go ahead with a proper release of `spacy-legacy` including the two open PRs there at the moment. This shouldn't interfere with the current, normal `master` branch in `spacy`. Then that release is just a normal one, not a "test" one, and won't need to be pulled (because I agree that would be a hassle)

svlandeg (approved):
This turned out to be a little more involved than we'd expected at the beginning, but with the new release of `spacy-legacy`, this is now finally good to merge 🎉

svlandeg:
Nice work @polm !

PR #9670, opened 2021-11-14
Translation of tutorials to Korean

SY-Mina:
## Description

Hello we are juniors majoring in software at Gachon University in Korea.
Since there is no translation of Korean and to make it easier to access to Korean, we made a translation of tutorials into Korean.

### Types of change

Enhancement to documentation

## Checklist

### DIFF
--- a/website/docs/KOR_doc.MD
+++ b/website/docs/KOR_doc.MD
@@ -1,0 +1,262 @@
+<a href="https://explosion.ai"><img src="https://explosion.ai/assets/img/logo.svg" width="125" height="125" align="right" /></a>
+
+# spaCy: 산업용 자연어처리
+
+spaCy는 Python 및 Cython의 **향상된 자연어 처리**를 위한 라이브러리입니다.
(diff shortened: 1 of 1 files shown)


svlandeg:
Hi!

We really appreciate the effort you've done to make a Korean version of spaCy's readme file 🤩 ! I'm sure this will be helpful to many Korean-speaking users.

From a practical point of view, we currently don't maintain translated versions of our docs, as it would be near-impossible to keep all of those versions in-sync with every update that we do. However, we'd definitely consider linking out to this if you're able to host it yourself?

One other thing I'd like to mention is that @ines is working on the [spaCy course](https://course.spacy.io/en/) and getting that   [updated for v3](https://github.com/ines/spacy-course/tree/v3) and we'd definitely appreciate translations of that, too. Have a look at the FAQ here: https://github.com/ines/spacy-course#i-want-to-help-translate-this-course-into-my-language-how-can-i-contribute and let us know if you're interested in helping out with a Korean version of the course!

PR #9674, opened 2021-11-15
Fix Language-specific factory handling in package command

polm:
## Description

If a component factory is registered like `@French.factory(...)` instead
of `@Language.factory(...)`, the name in the factories registry will be
prefixed with the language code. However in the nlp.config object the
factory will be listed without the language code. The `add_pipe` code
has fallback logic to handle this, but packaging code and the registry
itself don't.

This change makes it so that the factory name in nlp.config is the
language-specific form. It's not clear if this will break anything else,
but it does seem to fix the inconsistency and resolve the specific user
issue that brought this to our attention (#9652). 

### Types of change

Bug fix, probably?

## Checklist

- [] I ran the tests, and all new and existing tests passed.

### DIFF
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from wasabi import Printer, MarkdownRenderer, get_raw_input
 from thinc.api import Config
 from collections import defaultdict
+from catalogue import RegistryError
 import srsly
 import sys
 
@@ -212,9 +213,18 @@ def get_third_party_dependencies(
         if "factory" in component:
             funcs["factories"].add(component["factory"])
     modules = set()
+    lang = config["nlp"]["lang"]
     for reg_name, func_names in funcs.items():
         for func_name in func_names:
-            func_info = util.registry.find(reg_name, func_name)
+            # Try the lang-specific version and fall back
+            try:
+                func_info = util.registry.find(reg_name, lang + "." + func_name)
+            except RegistryError:
+                try:
+                    func_info = util.registry.find(reg_name, func_name)
+                except RegistryError as regerr:
+                    # lang-specific version being absent is not actually an issue
+                    raise regerr from None
             module_name = func_info.get("module")  # type: ignore[attr-defined]
             if module_name:  # the code is part of a module, not a --code file
                 modules.add(func_info["module"].split(".")[0])  # type: ignore[index]
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -671,7 +671,7 @@ class Language:
         resolved = registry.resolve(cfg, validate=validate)
         filled = registry.fill({"cfg": cfg[factory_name]}, validate=validate)["cfg"]
         filled = Config(filled)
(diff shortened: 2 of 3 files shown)


polm:
It seems like this causes one test to fail. The test specifically checks for the name of language-specific components like this, so if that's the only thing that's failing it should be safe to change it. That still doesn't make me confident this is the right fix though.

spacy/language.py
```diff
@@ -681,6 +681,7 @@ def create_pipe(
         # interpolated variables)
         if raw_config:
             filled = filled.merge(raw_config)
+        filled["factory"] = internal_name
```
databill86, line 684: Is there any reason for this line to be duplicated ? (already in line 674)

polm: Yes, the line above it overwrites the value from 674. It might make sense to just delete 674, but I'd need to check.

polm:
OK, so the one test shows that the expectation is that a config is written using generic names (no lang prefix), and that those are resolved to lang-specific forms at load time. If we assume that's what we want, then the solution is for the packaging code only to change. 

My latest commit reverts my previous fix and adds fallback logic in the packaging code. This isn't as bad as I thought it would be, and I think this touches fewer things and shouldn't cause problems elsewhere. 

I think this fix is good now, though I'll look at adding a test too.

spacy/language.py
```diff
@@ -681,6 +681,7 @@ def create_pipe(
         # interpolated variables)
         if raw_config:
             filled = filled.merge(raw_config)
+        filled["factory"] = factory_name
```
adrianeboyd, line 684: This isn't needed, right?

spacy/cli/package.py
```diff
@@ -212,9 +213,18 @@ def get_third_party_dependencies(
...
-            func_info = util.registry.find(reg_name, func_name)
+            # Try the lang-specific version and fall back
+            try:
+                func_info = util.registry.find(reg_name, lang + "." + func_name)
+            except RegistryError:
+                try:
+                    func_info = util.registry.find(reg_name, func_name)
+                except RegistryError as regerr:
+                    # lang-specific version being absent is not actually an issue
+                    raise regerr from None
```
adrianeboyd, line 227: Is there any need for this `try`/`except` vs. the previous behavior?

adrianeboyd:
Maybe a test can be added to `test_cli.py::test_get_third_party_dependencies`, maybe with a `Dutch.factory` or similar with a dummy component?

spacy/language.py
polm: Right, I should remove this.

spacy/cli/package.py
polm: Without this second `try`/`except`, if the fallback fails the traceback here will always report the first lookup failure, which isn't helpful (that's what I wanted to communicate with the comment, anyway). Using the `from None` makes the traceback shorter and to the point so I figured that was preferable.

polm:
Test added, so this should be good if there's no problem with the `try/except` thing above.

PR #9685, opened 2021-11-17
Entity ruler remove pattern

DuyguA:
## Description

Allow removing patterns by their `id` , introduces a new method `remove` to EntityRuler. Fixes #8228

### Types of change

A minor enhancement, also minor addition to docs about the new method.

## Checklist


website/docs/api/entityruler.md
```diff
@@ -210,6 +210,24 @@ of dicts) or a phrase pattern (string). For more details, see the usage guide on
 | ---------- | ---------------------------------------------------------------- |
 | `patterns` | The patterns to add. ~~List[Dict[str, Union[str, List[dict]]]]~~ |
 
+
+## EntityRuler.remove {#remove tag="method"}
+
+Remove a pattern by its ID from the entity ruler. An `Error` is raised if ID does not exist.
```
polm, line 216: This should be `ValueError`, right?

spacy/pipeline/entityruler.py
```diff
@@ -348,6 +348,45 @@ def clear(self) -> None:
...
+            self._create_label(label, eid) for (label, eid) in label_id_pairs
+        ]
+        self.phrase_patterns = defaultdict(
+            list,
+            {
+                label: val
+                for (label, val) in self.phrase_patterns.items()
+                if label not in created_labels
+            },
+        )  # remove the patterns from self.phrase_patterns
```
polm, line 373: Maybe this is a Black formatting thing, but wouldn't this be easier to read if these comments were above the start of the expression rather than the end?

spacy/errors.py
```diff
@@ -887,6 +887,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1021 = ("`pos` value \"{pp}\" is not a valid Universal Dependencies tag. "
              "Non-UD tags should use the `tag` property.")
     E1022 = ("Words must be of type str or int, but input is of type '{wtype}'")
+    E1023 = ("A pattern with {ent_id} id is not present in EntityRuler patterns.")
```
polm, line 890: I would use the format `ID \"{ent_id}\"` instead, in case IDs have spaces or something.

polm (approved):
Tests look good! This will be a nice feature to have.

There are a few small things with comments / text I think could be adjusted but overall this is good.

website/docs/api/entityruler.md
DuyguA: Yes, I think I made a tini-mini typo, thanks for pointing out!

spacy/errors.py
DuyguA: Ah, OK!

svlandeg:
@DuyguA : can you look into the failing tests? It might be that you just need some `# type: ignore` statements to make `mypy` happy.

DuyguA:
> @DuyguA : can you look into the failing tests? It might be that you just need some `# type: ignore` statements to make `mypy` happy.

Erghh, no. This is the error message:

```
spacy/pipeline/entityruler.py:385: error: Unsupported right operand type for in ("PhraseMatcher")
spacy/pipeline/entityruler.py:386: error: "PhraseMatcher" has no attribute "remove"
```

This is the failure line:  https://github.com/explosion/spaCy/blob/6d9370baf0dbc2616756c41ec12b800430b97284/spacy/pipeline/entityruler.py#L386
PhraseMatcher and Matcher both have a remove method, I don't know why it fails. It compiles and works on my local machine, Python 3.8. It also passes the Python 3.10 pipeline.




It's totally fine and works on Python 3.10.

adrianeboyd:
To clarify about the python 3.10 and CI: we're skipping the mypy tests for python 3.10 right now because it is too slow without binary wheels. If you ran it locally with 3.10 you should see the same errors as for the other versions of python.

You probably need to look at the `.pyi` files.

svlandeg:
> You probably need to look at the `.pyi` files.

Yes exactly. @DuyguA : I just checked, and the `remove` method is missing from `phrasematcher.pyi`. You can probably fix this by simply copying the definition from `matcher.pyi`, as the types should be the same:
```
def remove(self, key: str) -> None: ...
```

Locally, you can run `mypy` on just one file or module, to quickly check whether that resolves the error.

spacy/pipeline/entityruler.py
DuyguA: Ah yes, I originally put the comments to the same line then Black moved them to next line. I'll carry them above expressions.

website/docs/api/entityruler.md
```diff
@@ -210,6 +210,24 @@ of dicts) or a phrase pattern (string). For more details, see the usage guide on
 | ---------- | ---------------------------------------------------------------- |
 | `patterns` | The patterns to add. ~~List[Dict[str, Union[str, List[dict]]]]~~ |
 
+
+## EntityRuler.remove {#remove tag="method"}
```
svlandeg, line 214:
```suggestion
## EntityRuler.remove {#remove tag="method" new="3.2.1"}
```

website/docs/api/entityruler.md
```diff
@@ -210,6 +210,24 @@ of dicts) or a phrase pattern (string). For more details, see the usage guide on
...
+
+Remove a pattern by its ID from the entity ruler. A `ValueError` is raised if ID does not exist.
```
svlandeg, line 216:
```suggestion
Remove a pattern by its ID from the entity ruler. A `ValueError` is raised if the ID does not exist.
```

PR #9688, opened 2021-11-17
Allow Matcher to match on ENT_ID and ENT_KB_ID 

nrodnova:
Added  `ENT_ID` and `ENT_KB_ID` attributes so that they can be used in `Matcher`patterns.

## Description
Added  `ENT_ID` and `ENT_KB_ID` attributes so that they can be used in `Matcher`patterns. 
The only change required was in `schemas.py`.
Also, I added tests, one for each attributes and made a change in the documentation.
I ran all tests on MacOS Big Sur (11.2.3)

### Types of change

enhancement, new tests, change to documentation

## Checklist


adrianeboyd:
Thanks for this PR, this is a nice addition!

It's really nice to see thorough tests as part of PRs, but I'm trying to keep an eye on the overall running time of the test suite, since we're starting to run into issues with build/CI timeouts (not directly in the PRs on github, but elsewhere).

I think in this particular case, it would be sufficient if we just added some valid / invalid patterns here instead, since we have other tests for the token attribute getting part of the `Matcher` and these are standard `Token` attributes:

https://github.com/explosion/spaCy/blob/0e93b315f3a5f96f2190d7eae7f6085bafe9c747/spacy/tests/matcher/test_pattern_validation.py#L8-L36

Let me also make a quick formatting suggestion for the docs...

website/docs/api/matcher.md
```diff
@@ -44,6 +44,8 @@ rule-based matching are:
 | `SPACY`                                         | Token has a trailing space. ~~bool~~                                                                                      |
 |  `POS`, `TAG`, `MORPH`, `DEP`, `LEMMA`, `SHAPE` | The token's simple and extended part-of-speech tag, morphological analysis, dependency label, lemma, shape. ~~str~~       |
 | `ENT_TYPE`                                      | The token's entity label. ~~str~~                                                                                         |
+| `ENT_ID`                                      | The token's entity id (ent_id_). ~~str~~                                                                                         |
+| `ENT_KB_ID`                                      | The token's entity kb id (ent_kb_id_). ~~str~~                                                                                         |
```
adrianeboyd, line 48:
```suggestion
| `ENT_ID`                                        | The token's entity ID (`ent_id`). ~~str~~                                                                                 |
| `ENT_KB_ID`                                     | The token's entity knowledge base ID (`ent_kb_id`). ~~str~~                                                               |
```

(It's hard to get the formatting right in the suggestions. We can run `prettier` again later.)

nrodnova:
@adrianeboyd I will change the tests, didn't realize runtime was an issue. Sorry about the formatting. What tools do you use for formatting and also for editing the .md files? I edited the documentation files in Pycharm, and it was pretty painful

adrianeboyd:
No, there's no reason for you to know about the test suite! I really like seeing that the detailed tests succeed.

`.py` files are formatted with `black`, but we haven't found anything that supports cython. The docs are formatted with `prettier`. You can run `npx prettier --write file.md` outside of an IDE, or get `prettier` as an add-on in many IDEs. It's very painful to try to get any of the doc table formatting correct by hand. We don't mind cleaning up the formatting later if it's a pain for you to do yourself.

adrianeboyd:
Just a few formatting edits, and once the CI passes, this can be merged.

PR #9691, opened 2021-11-18
Add note on batch contract for listeners

polm:
Using listeners requires batches to be consistent. This is obvious if
you understand how the listener works, but it wasn't clearly stated in
the Docs, and was subtle enough that the EntityLinker missed it.

There is probably a clearer way to explain what the actual requirement
is, but I figure this is a good start.

This issue was raised by #9669. 

## Checklist


polm:
Good point, I rewrote it to emphasize that this happens because of the caching mechanism.

PR #9703, opened 2021-11-19
New tests for a number of alpha languages

richardpaulhudson:
- Tests for the alpha languages Afrikaans; Albanian; Croatian; Estonian; Icelandic; Latvian; Slovak and Slovenian and for the multi-language module.


adrianeboyd:
The test additions look good! But I think we might want to split this into two PRs, one with the tests for `master` and one with the stop words changes for `develop`, since we only change the language defaults on minor releases (would be v3.3).

richardpaulhudson:
@adrianeboyd good point, I've removed the stop words change here and will resubmit it to `develop`.

spacy/tests/lang/af/test_text.py
```diff
@@ -0,0 +1,14 @@
...
+    text = """
+Hierdie Universele Verklaring van Menseregte as 'n algemene standaard vir die verwesenliking deur alle mense en nasies, 
+om te verseker dat elke individu en elke deel van die gemeenskap hierdie Verklaring in ag sal neem en deur opvoeding, 
+respek vir hierdie regte en vryhede te bevorder, op nasionale en internasionale vlak, daarna sal strewe om die universele 
+en effektiewe erkenning en agting van hierdie regte te verseker, nie net vir die mense van die Lidstate nie, maar ook vir 
+die mense in die gebiede onder hul jurisdiksie.
+
+"""
+    tokens = af_tokenizer(text)
+    assert len(tokens) == 101
```
svlandeg, line 14: Nitpick: Could you format all files with `black`? This ensures the same formatting across the code base, eg it will automatically add an empty line here (there's a few more new files that should be formatted this way, too).

svlandeg:
This looks good, thanks! How did you determine the required number of tokens for the example tests?

richardpaulhudson:
> This looks good, thanks! How did you determine the required number of tokens for the example tests?

I didn't count the tokens for each language, but I did a basic sanity check of the numbers and also had a look at the tokens produced for each language and checked the basic behaviour was as I expected. (It wasn't for the South Indian languages, which I know nothing about anyway, which is why I ended up leaving them out.) There were a few anomalies like Afrikaans `'n` being analysed as two words, but then I regarded that as expected behaviour given the language-independent parser was in use.

svlandeg:
> There were a few anomalies like Afrikaans 'n being analysed as two tokens, but then I regarded that as expected behaviour given the language-independent parser was in use.

I would almost be inclined to add a failing test for cases like these though, to indicate areas that should be improved and clarify that we're aware of them. And maybe we can reduce the test to text that is more clear, i.e. where the number of tokens wouldn't change if we further improve on the language tokenization?

richardpaulhudson:
> > There were a few anomalies like Afrikaans 'n being analysed as two tokens, but then I regarded that as expected behaviour given the language-independent parser was in use.
> 
> I would almost be inclined to add a failing test for cases like these though, to indicate areas that should be improved and clarify that we're aware of them. And maybe we can reduce the test to text that is more clear, i.e. where the number of tokens wouldn't change if we further improve on the language tokenization?

I've changed the tests accordingly.

spacy/tests/lang/af/test_text.py
```diff
@@ -2,14 +2,21 @@
...
     tokens = af_tokenizer(text)
-    assert len(tokens) == 101
+    assert len(tokens) == 100
+
+
+@pytest.mark.xfail
+def test_indefinite_article(af_tokenizer):
+    text = "'n"
+    tokens = af_tokenizer(text)
+    assert len(tokens) == 1
```
svlandeg, line 22: Maybe test with a slightly longer phrase? That looks more natural to me (I'm nitpicking, I know)
```suggestion
    text = "as 'n algemene standaard"
    tokens = af_tokenizer(text)
    assert len(tokens) == 4
```

spacy/tests/lang/et/test_text.py
```diff
@@ -21,4 +16,11 @@ def test_long_text(et_tokenizer):
...
     tokens = et_tokenizer(text)
-    assert len(tokens) == 122
+    assert len(tokens) == 94
+
+
+@pytest.mark.xfail
+def test_ordinal_number(et_tokenizer):
+    text = "10."
+    tokens = et_tokenizer(text)
+    assert len(tokens) == 1
```
svlandeg, line 26: Here I think it's important to include at least some context:
```suggestion
    text = "10. detsembril 1948"
    tokens = et_tokenizer(text)
    assert len(tokens) == 3
```

spacy/tests/lang/hr/test_text.py
```diff
@@ -19,4 +16,11 @@ def test_long_text(hr_tokenizer):
...
     tokens = hr_tokenizer(text)
-    assert len(tokens) == 136
+    assert len(tokens) == 105
+
+
+@pytest.mark.xfail
+def test_ordinal_number(hr_tokenizer):
+    text = "10."
+    tokens = hr_tokenizer(text)
+    assert len(tokens) == 1
```
svlandeg, line 26:
```suggestion
    text = "10. prosinca 1948"
    tokens = hr_tokenizer(text)
    assert len(tokens) == 3
```

spacy/tests/lang/is/test_text.py
```diff
@@ -21,4 +16,11 @@ def test_long_text(is_tokenizer):
...
     tokens = is_tokenizer(text)
-    assert len(tokens) == 157
+    assert len(tokens) == 120
+
+
+@pytest.mark.xfail
+def test_ordinal_number(is_tokenizer):
+    text = "10."
+    tokens = is_tokenizer(text)
+    assert len(tokens) == 1
```
svlandeg, line 26:
```suggestion
    text = "10. desember 1948"
    tokens = is_tokenizer(text)
    assert len(tokens) == 3
```

spacy/tests/lang/lv/test_text.py
```diff
@@ -21,4 +17,11 @@ def test_long_text(lv_tokenizer):
...
     tokens = lv_tokenizer(text)
-    assert len(tokens) == 140
+    assert len(tokens) == 109
+
+
+@pytest.mark.xfail
+def test_ordinal_number(lv_tokenizer):
+    text = "10."
+    tokens = lv_tokenizer(text)
+    assert len(tokens) == 1
```
svlandeg, line 27:
```suggestion
    text = "10. decembrī"
    tokens = lv_tokenizer(text)
    assert len(tokens) == 2
```

spacy/tests/lang/sk/test_text.py
```diff
@@ -43,3 +39,10 @@ def test_lex_attrs_like_number(sk_tokenizer, text, match):
     tokens = sk_tokenizer(text)
     assert len(tokens) == 1
     assert tokens[0].like_num == match
+
+
+@pytest.mark.xfail
+def test_ordinal_number(sk_tokenizer):
+    text = "10."
+    tokens = sk_tokenizer(text)
+    assert len(tokens) == 1
```
svlandeg, line 48:
```suggestion
    text = "10. decembra 1948"
    tokens = sk_tokenizer(text)
    assert len(tokens) == 3
```

spacy/tests/lang/sl/test_text.py
```diff
@@ -21,4 +17,11 @@ def test_long_text(sl_tokenizer):
...
     tokens = sl_tokenizer(text)
-    assert len(tokens) == 153
+    assert len(tokens) == 116
+
+
+@pytest.mark.xfail
+def test_ordinal_number(sl_tokenizer):
+    text = "10."
+    tokens = sl_tokenizer(text)
+    assert len(tokens) == 1
```
svlandeg, line 27:
```suggestion
    text = "10. decembra 1948"
    tokens = sl_tokenizer(text)
    assert len(tokens) == 3
```

svlandeg (commented):
I've suggested some changes to run the tokenizer texts on a bit more context rather than just one token.

spacy/tests/lang/af/test_text.py
richardpaulhudson: Yes, definitely looks much better and is also a better test!

Issue #9706, opened 2021-11-19
Thoughts on  Underscore class for Spans with the same boundaries?


nrodnova:
## How to reproduce the behaviour

The uniqueness of `Span` is determined based on `(start, end, label_, kb_id_)`. Underscore class stores extensions only based on `(start, end)`. Any thoughts on that?

```python
span_1 = Span(doc, 0, 3, label = 'LABEL_1', kb_id = 'KB_ID_1')
span_2 = Span(doc, 0, 3, label = 'LABEL_2', kb_id = 'KB_ID_2')
assert span_1 != span_2
span_1._.test = 'span_1._.text'
print(span_2._.test)
```
The output, as expected, is:
```
 'span_1._.text'
```
Which poses a problem now, when we have  `SpanGroups` and overlapping `Spans`. It can be fixed pretty easily, I think. I can do it, but I want to hear the thoughts of the maintainers first.

adrianeboyd:
I suspect that we're not going to want to make the `Span` extensions more complicated. It's useful that `Doc`, `Span`, and `Token` extensions are all keyed on the same index values and all have the same internal shape.

nrodnova:
Do you have any suggestions/thoughts on how to deal with the situation then? :) Meaning, avoiding overwriting `Span`'s extension values for unequal `Spans` with the same boundaries?

adrianeboyd:
Hmm, the only workarounds I can think of off the top of my head involve keying the internal extension values based on the labels and kb_ids somehow, but I'm not sure how well this would work in practice.

If you've already implemented it, you can open a PR and we can have a look? But I think it's going to be hard to do without affecting the `Doc` and `Token` extensions?

adrianeboyd:
I think this is resolved by #11429.

PR #9708, opened 2021-11-19
Updated Underscore to handle Spans with same boundaries

nrodnova:
Updated Underscore class to properly handle extension values for `Spans` with equal boundaries. 

Resolves issue https://github.com/explosion/spaCy/issues/9706

## Description
The change actually looks reasonably neat. Nothing changed for `Doc` or `Token`. The only thing that needed to be taken care of was the case when a `Span` that have data in `doc.user_data` had its `label` or `kb_id` updated. I also added a test.

### Types of change

enhancement, new test

## Checklist

### DIFF
--- a/spacy/tests/doc/test_underscore.py
+++ b/spacy/tests/doc/test_underscore.py
@@ -189,6 +189,11 @@ def test_underscore_for_span(en_tokenizer) :
     span_1._.span_extension = 'span_1 extension'
     span_2._.span_extension =  'span_2 extension'
 
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'span_2 extension'
+
     assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
     assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
 
@@ -196,9 +201,19 @@ def test_underscore_for_span(en_tokenizer) :
     assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
     assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
 
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'span_2 extension'
+
     span_1.kb_id_ = "KB_ID"
     assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
     assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
+
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'span_2 extension'
 
     assert doc.user_data[('._.', 'doc_extension', None, None)] == 'doc extension'
     assert doc.user_data[('._.', 'token_extension', 0, None)] == 'token extension'
@@ -206,3 +221,9 @@ def test_underscore_for_span(en_tokenizer) :
     span_2._.span_extension = 'updated span_2 extension'
     assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
     assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'updated span_2 extension'
+
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'updated span_2 extension'
+
--- a/spacy/tests/doc/test_underscore.py
+++ b/spacy/tests/doc/test_underscore.py
@@ -171,3 +171,38 @@ def test_underscore_docstring(en_vocab):
     doc = Doc(en_vocab, words=["hello", "world"])
     assert test_method.__doc__ == "I am a docstring"
     assert doc._.test_docstrings.__doc__.rsplit(". ")[-1] == "I am a docstring"
+
+
+def test_underscore_for_span(en_tokenizer) :
+
+    Doc.set_extension(name='doc_extension', default=None)
+    Span.set_extension(name='span_extension', default=None)
+    Token.set_extension(name='token_extension', default=None)
+
+    text = 'Hello, world!'
+    doc = en_tokenizer(text)
+    span_1 = Span(doc, 0, 2, 'SPAN_1')
+    span_2 = Span(doc, 0, 2, 'SPAN_2')
+
+    doc._.doc_extension = "doc extension"
+    doc[0]._.token_extension = "token extension"
+    span_1._.span_extension = 'span_1 extension'
+    span_2._.span_extension =  'span_2 extension'
+
+    assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
+    assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
+
+    span_1.label_ = "NEW_LABEL"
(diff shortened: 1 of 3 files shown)


spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,59 @@ def test_method(doc, arg1=1, arg2=2):
...
+    doc._.doc_extension = "doc extension"
+    doc[0]._.token_extension = "token extension"
+    span_1._.span_extension = 'span_1 extension'
+    span_2._.span_extension =  'span_2 extension'
+
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'span_2 extension'
+
```
svlandeg, line 196: It looks like this code was accidentally duplicated.
```suggestion
```

spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,59 @@ def test_method(doc, arg1=1, arg2=2):
...
+    assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
+
+    span_1.label_ = "NEW_LABEL"
+    assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
+    assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
+
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'span_2 extension'
```
svlandeg, line 207: Is there a reason you copy/pasted this again? I don't think it's necessary?

spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,59 @@ def test_method(doc, arg1=1, arg2=2):
...
+
+    assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
+    assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
+
+    span_1.label_ = "NEW_LABEL"
+    assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
+    assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
```
svlandeg, line 202: I don't think the test for `span_2` should be repeated.
```suggestion
    span_1.label_ = "NEW_LABEL"
    assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
```

spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,59 @@ def test_method(doc, arg1=1, arg2=2):
...
+
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'span_2 extension'
+
+    span_1.kb_id_ = "KB_ID"
+    assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
+    assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'span_2 extension'
```
svlandeg, line 211:
```suggestion
```

spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,59 @@ def test_method(doc, arg1=1, arg2=2):
...
+
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'span_2 extension'
+
```
svlandeg, line 217:
```suggestion
```

spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,59 @@ def test_method(doc, arg1=1, arg2=2):
...
+    assert doc.user_data[('._.', 'doc_extension', None, None)] == 'doc extension'
+    assert doc.user_data[('._.', 'token_extension', 0, None)] == 'token extension'
+
+    span_2._.span_extension = 'updated span_2 extension'
+    assert doc.user_data[('._.', 'span_extension', span_1.start_char, span_1.end_char, span_1.label, span_1.kb_id)] == 'span_1 extension'
```
svlandeg, line 222:
```suggestion
```

spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,59 @@ def test_method(doc, arg1=1, arg2=2):
...
+    assert doc.user_data[('._.', 'span_extension', span_2.start_char, span_2.end_char, span_2.label, span_2.kb_id)] == 'updated span_2 extension'
+
+    doc._.doc_extension == "doc extension"
+    doc[0]._.token_extension == "token extension"
+    span_1._.span_extension == 'span_1 extension'
+    span_2._.span_extension == 'updated span_2 extension'
+
```
svlandeg, line 229:
```suggestion
```

svlandeg (commented):
Thanks a lot for creating this PR to make a potential solution for #9706 more tangible! 

I've been reviewing the PR and had some comments about one of the tests, and wanted to check with you whether you could run `black` for autoformatting.

In general, I'm not 100% convinced yet whether this is the way to go. I definitely see the problem you've outlined in the original issue, but I'm wary of this implicit contract between `Span` and `Underscore` that is introduced here. It feels like the code might be prone to bugs at some point, and keeping the labels/keys in-sync might be a challenge.

But I'm not sure what a better solution would be either. Happy to hear feedback from others!

svlandeg:
@adrianeboyd : would love to hear your ideas as well!

nrodnova:
@svlandeg Regarding formatting - I installed pre-commit step for spacy after your comments on another PR. This one was committed before that, that is why there are formatting issues. I will take a look at the tests if you guys decide this idea is a go. I definitely meant to test something particular on every line, but I will double-check.

nrodnova:
@svlandeg Regarding _this implicit contract between Span and Underscore that is introduced here_. We can pass 2 additional arguments explicitly to the `Underscore` constructor from `Doc` and `Token`  (`None` and  `None`), and then there will be no _implicit contract between Span and Underscore_. Current `Underscore` behavior is inconsistent with `Span`'s equality concept. And can cause confusion when `Spans` overwrite each other 's `_` values.

svlandeg:
> Current `Underscore` behavior is inconsistent with `Span`'s equality concept. And can cause confusion when `Spans` overwrite each other 's `_` values.

Yep, I definitely agree that the current behaviour is inconsistent - this API hasn't been updated since we introduced overlapping spans in `doc.spans`.

> @svlandeg Regarding _this implicit contract between Span and Underscore that is introduced here_. We can pass 2 additional arguments explicitly to the `Underscore` constructor from `Doc` and `Token` (`None` and `None`), and then there will be no _implicit contract between Span and Underscore_. 

I was referring to the fact that `Span` needs to know it should update stuff in the underscore data structure whenever a label or KB ID is set. I'm not sure I understand how 2 additional arguments on the constructor would prevent that?

adrianeboyd:
The main difference is modifiable spans. Until `doc.spans`/`spancat` you couldn't edit any properties of a span at all, which did simplify things in a lot of cases, but also made a lot of `spancat` a hassle to implement.

The other issue we need to keep in mind here in serialization. If you create a `DocBin` or run `Doc.to_bytes` with this modified underscore representation, I don't think you can load it with an earlier version of spacy?

nrodnova:
@adrianeboyd I agree that having mutable `Spans` created some hassle, but it is so much more convenient!

Regarding serialization. I would worry more about being able to read binary representations saved with the older model using the newer one, no? Backward compatibility, not the forward one?

adrianeboyd:
It's definitely important that `.spacy` files are both forward- and backward-compatible, at least within spacy v3.x. I added a format version number to the `.spacy` format in v3.0 because I thought it might be useful in the future, but it's not read/checked/used in any way that might help with this problem. I've looked a bit into the formats and readers without seeing an immediate solution...

svlandeg:
Maintenance update: we're putting this specific PR/issue on hold until we've decided on how to proceed with the `shared_ptr` idea: https://github.com/explosion/spaCy/pull/9869

adrianeboyd:
There are also some additional places where this data is referenced or used directly, which would need to be updated after the types of underscore keys are changed. I'm not sure this is everything, but a few places are:

* `Doc.from_docs`
* `Span.as_doc`

adrianeboyd:
Because of the serialization issues, I think this would be best in `v4`, where we've also put the `shared_ptr` modifications.

adrianeboyd:
As a note, this PR should also now add `Span.id` everywhere similar to `Span.kb_id` and `Span.label` so far.

svlandeg:
Closing in favor of #11429 where we're continuing with this work :-)

PR #9722, opened 2021-11-22
Use reference parse to initialize parser moves

adrianeboyd:
## Description

Use reference parse to initialize parser moves

### Types of change

Bug fix?

## Checklist


adrianeboyd:
This does slightly change the behavior for parser training because it wouldn't be initialized with the exact same move counts, so it might be better to do this in v3.3. If we consider a bug fix (other components initialize from the reference docs), then it could go into v3.2.

PR #9731, opened 2021-11-23
New SpanGroupRuler class that operates on doc.spans rather than doc.ents

nrodnova:
A class that populates `doc.spans` based on specified set of rules.

## Description
I'm going ti keep you busy with my PRs, guys :)
There was a (brief) discussion about this class here: https://github.com/explosion/spaCy/discussions/9353
I wrote it when you introduced `SpanGroup` concept, and find it very useful. It allows flexibility on overlaps, it takes
rules in a dict format (it also understands existing `EntityRuler` format)
```python
patterns = {
    'GROUP': {
        'LABEL' : {
            'KB_ID' :
                [rule_1, rule_2, rule_N],
        }
    }
}
```
It has a method that returns matches as a dictionary of span lists. And in `__call__` methods it populates `doc.spans`.
Since it needs to incorporate `GROUP` in the underlying `matcher` key, and because it allows different ways to overlap spans, I think it deserves to be a separate class. Let me know what you think.

If you find it useful, I will make it a `Pipe`, add serialization and tests, and maybe cythonize it a bit. I can also merge it with `EntityRuler` instead, if you don't think it deserves to be a separate class.

### Use case
```python
import spacy
from spacy.tokens import Span
from spacy.pipeline import SpanGroupRuler
from spacy.pipeline.span_group_ruler import get_overlapping_spans

text = "Fruit pie, apple gallette, cherry and almond pastry."
doc = nlp(text)

fruit = [{'LOWER' : {'IN' : ['apple', 'cherry', 'almond', 'fruit']}}]
pastry = [{'LOWER': {'IN': ['pie', 'gallette', 'pastry']}}]
connector = [{'LOWER' : {'IN': ['and', '-', '/']}, 'OP' : '?'}]

ruler_patterns = {
    'SIMPLE_CONCEPTS' : { # Group
        'FRUIT' : { # Label
            'apple' : [[{'LOWER' : 'apple'}]], #kb_id => list of rules
            'cherry': [[{'LOWER': 'cherry'}]],
            'almond': [[{'LOWER': 'almond'}]],
            '' : [[{'LOWER' : 'fruit'}]],
        },
        'PASTRY' : {
            'pie': [[{'LOWER': 'pie'}]],
            'gallette':[ [{'LOWER': 'gallette'}]],
            '': [[{'LOWER': 'pastry'}]],
        },
    },
    'COMPLEX_CONCEPTS' : {
        'FRUIT_PASTRY' : { # These two won't overlap as they are within the same group
            '' : [fruit + pastry,
                  fruit + connector + fruit + pastry,
                  ],
        },
    },
}

ruler = SpanGroupRuler(nlp, pattern_dict = ruler_patterns)

# Label the spans
# Spans within each group do not overlap, they get filtered with 'LONGEST' policy
doc = ruler(doc)

# Assign fruit and pasrty type to fruit pastries
fruit_pastries = doc.spans['COMPLEX_CONCEPTS']
fruit = [x for x in doc.spans['SIMPLE_CONCEPTS'] if x.label_ == 'FRUIT']
pastry_types = [x for x in doc.spans['SIMPLE_CONCEPTS'] if x.label_ == 'PASTRY']

Span.set_extension(name = 'fruit', default = list())
Span.set_extension(name='type', default=None)
for fruit_pastry in fruit_pastries :
    fruit_pastry._.fruit = [x for x in get_overlapping_spans(fruit_pastry, fruit) if x.kb_id]
    types = [x for x in get_overlapping_spans(fruit_pastry, pastry_types) if x.kb_id]
    fruit_pastry._.type = types[0] if types else 'unknown'

for span in fruit_pastries :
    print("%s: fruit = [%s], type = '%s'" % (span.text.lower(), ', '.join([x.text.lower() for x in span._.fruit]), span._.type))

```
Output:
```
fruit pie: fruit = [], type = 'pie'
apple gallette: fruit = [apple], type = 'gallette'
cherry and almond pastry: fruit = [cherry, almond], type = 'unknown'
```
### Types of change

new feature

## Checklist

- [x]  I ran the tests, and all new and existing tests passed.

### DIFF
--- a/spacy/pipeline/__init__.py
+++ b/spacy/pipeline/__init__.py
@@ -15,6 +15,7 @@ from .spancat import SpanCategorizer
 from .textcat_multilabel import MultiLabel_TextCategorizer
 from .tok2vec import Tok2Vec
 from .functions import merge_entities, merge_noun_chunks, merge_subtokens
+from .span_group_ruler import SpanGroupRuler
 
 __all__ = [
     "AttributeRuler",
--- a/spacy/pipeline/span_group_ruler.py
+++ b/spacy/pipeline/span_group_ruler.py
@@ -1,0 +1,180 @@
+from ..matcher import Matcher
+from ..language import Language
+from typing import Optional, Union, List, Dict, Tuple, Iterable, Any, Callable, Sequence
+from ..tokens import Span
+
+from collections import defaultdict
+
+DEFAULT_MATCH_ID_SEPARATOR = "||"
+PatternType = Dict[str, Union[str, List[Dict[str, Any]]]]
+PatternDictType = Dict[str, Dict[str, Dict[str, List[List[Dict[str, Any]]]]]]
+
+
+class SpanGroupRuler :
+    """
+    This class populates Doc.spans based on spacy matcher rules, similarly to EntityRuler.
+    Pattenrs are accepted in a list form (similar to EntityRuler), or more convenient
+    dictionary form:
+    { 'GROUP_NAME' : {
+            'LABEL_NAME' : {
+                'KB_ID' : [rule_1, rule_2, rule_N],
+            },
+        },
+    }
+    The class provides flexibility controlling span overlaps. By default, it allows overlapping spans
+    between groups. Within each groups, all matched spans are treated similar to entities
+    in entity ruler, i.e. overlapping matches are filtered using 'LONGEST' policy.
+    It can be configured by specifying overlap_filter_level parameter.
+
+    ruler() method returns a dictionary of span lists
+    __call__() method assigns matched spans to doc.spans
+    """
+    def __init__(
+            self,
+            nlp : Language,
+            name : str = 'span_group_ruler',
+            *,
+            validate : bool = True,
+            match_id_separator : str = DEFAULT_MATCH_ID_SEPARATOR,
+            reset_groups : bool = True,
+            overlap_filter_level : str = 'group', # can be 'group', 'label', 'kb_id', 'all' or None # group means overlaps between the groups are allowed. label = between (group, label), kb_id = between (group, label, kb_id)
+            patterns : Optional[List[PatternType]] = None,
+            pattern_dict : Optional[PatternDictType] = None,
+    ):
+        self.nlp = nlp
+        self.name = name
+        self.validate = validate
+        self.match_id_separator = match_id_separator
+        self.reset_groups = reset_groups
+        self.overlap_filter_level = overlap_filter_level # TODO: validate
+        self.matcher = Matcher(self.nlp.vocab, validate = self.validate)
+        self.key2id = {}
+        self.id2key = {}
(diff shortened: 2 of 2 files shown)


adrianeboyd:
I think this is a good idea, but I think it would be better to keep it closer to `EntityRuler`, since `EntityRuler` already contains a lot of the machinery related to supporting both token and phrase patterns.

We'd have to refactor how the `cfg` serialization works a little bit, but then it would be possible to have a `SpanRuler` that has an additional `spans_key` parameter and skips the filtering in `set_annotations`, but otherwise can extend `EntityRuler`.

I'm not sure it's necessary to support more than one spans key in a given `SpanRuler` or whether we want the additional overhead of supporting both `ENT_ID` and `ENT_KB_ID`.

I also think we should try to avoid supporting too many filtering options in the `SpanRuler` itself because it's going to be hard to maintain. (At most I might be in favor of having a registered function for span filtering in `set_annotations` as part of the config. It's also something that's missing from `EntityRuler`.)

nrodnova:
@adrianeboyd Do you mean inheriting `SpanRuler` from `EntityRuler`? 

Also, regarding span filtering options. Ideally, I think, `Matcher` should accept a callback/hook that translates `match_id` into the filtering key, and then just filter matches based on that key. This way, `EntityRuler` won't have to duplicate the filtering code. Filtering is already implemented in `Matcher`, but it is currently hard-coded to be within `match_id`. I was planning a separate PR for that, but maybe we can just talk it over here :)

adrianeboyd:
Sorry for the late response!

I think as part of working on this we should refactor the `EntityRuler` slightly to make the serialization a bit more standard (at the very least with `_get_config` and `_set_config` methods, plus I think there's a bug in the order of the pattern adding and phrase matcher initialization in `from_bytes`) and then `SpanRuler` should inherit from `EntityRuler` and only override `set_annotations` and the serialization methods.

In principle I'd rather have `EntityRuler` inherit from the more general `SpanRuler`, but this is a bit more refactoring than I think we should try to deal with in the near future.

And then I think we should try to make the filtering / sorting more configurable for both at the same time. I think it's okay if the initial version of the `SpanRuler` doesn't do any filtering and just sets all the matches spans for one span key.

I can't promise that we'll have time to work on the refactoring extremely soon, but it is definitely on our to-do list.

svlandeg:
I agree this would be great to have, even it takes some more work :-) I'll put this in draft as we continue discussion/work :-)

adrianeboyd:
I looked at this again and the use of `ent_id` in `EntityRuler` and the non-standard serialization make this really hard to do in any sensible way, so I think a separate class with a small amount of duplicated code related to `add_patterns` is going to be the better option. I'll open a new PR in a minute...

adrianeboyd:
Linking #9880, a proposal for a simpler `SpanRuler` that just uses a single spans key. It currently supports only span labels (not `kb_id`) and doesn't include support for match filtering.

svlandeg:
Closing this PR in favour of #9880, but thanks again @nrodnova for the initial idea!

PR #9738, opened 2021-11-24
Intify IOB

DuyguA:
## Description

Minor enhancement,  IOB strings  `I`, `O`, `B`  are mapped to 1, 2 and 3. 

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/attrs.pyx
+++ b/spacy/attrs.pyx
@@ -158,8 +158,11 @@ def intify_attr(name):
     stringy_attr (string): Attribute string name. Can also be int (will then be left unchanged)
     RETURNS (int): int representation of the attribute, or None if it couldn't be converted.
     """
+    iob_strs = ("", "I", "O", "B")
     if isinstance(name, int):
         return name
+    elif name in "IOB":
+        return iob_strs.index(name)
     elif name in IDS:
         return IDS[name]
     elif name.upper() in IDS:


honnibal:
Ah this will be _super_ breaking, especially changing the integer encoding for `I` will break all the models a lot. We'll also get a different token vector for all words starting with `B` or `O`. We'd only be able to think about this for a v4.

adrianeboyd:
I think this is more of a bug fix than something really breaking, but this initial draft isn't quite right, either. It should be `ENT_IOB` and this is mainly relevant for matcher patterns. The previous version would work with integers but not at all with strings, and unless you can guess some string with the hash 1, you're kind of out-of-luck and it never matches. I don't think it's going to be that breaking in practice.

An example for extending entities with the attribute ruler:

```python
import spacy

nlp = spacy.blank("en")
ar = nlp.add_pipe("attribute_ruler")
ar.add(patterns=[[{"ORTH": "a"}, {"ORTH": "cat"}]], attrs={"ENT_IOB": 3, "ENT_TYPE": "CATS"}, index=1)
doc = nlp("a cat")
print(doc.ents)
```

This currently works with the integer but not with the string. This change should only affect `ENT_IOB` and nothing else.

adrianeboyd:
I don't think this change is in the right place, it should be in `intify_attrs` where you have access to both the name and value.

adrianeboyd:
And maybe there's something I haven't thought through related to `ENT_IOB` and `IN`, but I can't imagine any sensible `IN` uses, either.

adrianeboyd:
And to clarify, we wanted to do this for `Matcher` patterns in the other PR, but this is now for things like the `Token.set_struct_attr` part of attribute ruler, not the matcher pattern part.

spacy/attrs.pyx
```diff
@@ -158,8 +158,11 @@ def intify_attr(name):
     stringy_attr (string): Attribute string name. Can also be int (will then be left unchanged)
     RETURNS (int): int representation of the attribute, or None if it couldn't be converted.
     """
+    iob_strs = ("", "I", "O", "B")
```
adrianeboyd, line 161: It won't be here in the end, but use the same `Token` property rather than a separate list.

DuyguA:
> Ah this will be _super_ breaking, especially changing the integer encoding for `I` will break all the models a lot. We'll also get a different token vector for all words starting with `B` or `O`. We'd only be able to think about this for a v4.

Erggh, yes. I made too much shortcut :grimacing:

spacy/attrs.pyx
DuyguA: OK, sure.

DuyguA:
> I don't think this change is in the right place, it should be in `intify_attrs` where you have access to both the name and value.

Sure, I made too much of a shortcut. :grimacing:  I'll change the code.

spacy/tests/lang/test_attrs.py
```diff
@@ -1,4 +1,6 @@
 import pytest
+from spacy.attrs import intify_attrs, ORTH, NORM, LEMMA, IS_ALPHA, ENT_IOB
+from spacy.lang.lex_attrs import is_punct, is_ascii, is_currency, like_url, word_shape
```
adrianeboyd, line 3:
```suggestion
from spacy.attrs import intify_attrs, ENT_IOB
```

spacy/attrs.pyx
```diff
@@ -1,3 +1,5 @@
+from .tokens.token import Token
```
adrianeboyd, line 1:
```suggestion
from .tokens import Token
```

spacy/attrs.pyx
DuyguA: OK, I imported `Token` into `attrs.py` but it created a problem while building the project. It looks like it creates a cyclic import and doesn't compile at all. If I remove the import and place a list, it works fine. Here's the error:

```
ImportError while loading conftest '/home/duygu/Templates/spaCy/spacy/tests/conftest.py'.
spacy/__init__.py:14: in <module>
    from . import pipeline  # noqa: F401
spacy/pipeline/__init__.py:1: in <module>
    from .attributeruler import AttributeRuler
spacy/pipeline/attributeruler.py:6: in <module>
    from .pipe import Pipe
spacy/pipeline/pipe.pyx:1: in init spacy.pipeline.pipe
    # cython: infer_types=True, profile=True
spacy/vocab.pyx:1: in init spacy.vocab
    # cython: profile=True
spacy/lexeme.pyx:18: in init spacy.lexeme
    from .attrs import intify_attrs
spacy/attrs.pyx:1: in init spacy.attrs
    from .tokens.token cimport Token
spacy/tokens/__init__.py:1: in <module>
    from .doc import Doc
spacy/tokens/doc.pyx:28: in init spacy.tokens.doc
    from ..attrs import intify_attr, IDS
E   ImportError: cannot import name intify_attr
```

Shall I go back to using a list?

spacy/attrs.pyx
adrianeboyd: Ah, hmm. That is an issue. Why don't we move the `iob_strings` tuple definition from `spacy.tokens` to `spacy.attrs` as a global like `IOB_STRINGS`?

DuyguA: I agree, it creates an circular import. It's a good idea, I can do it.

adrianeboyd: Sorry, that wasn't quite what I meant. Leave `Token.iob_strings` as a method (and continue using it in `Doc` and other places), but just define `IOB_STRINGS` in `attrs` as the tuple (we would use a global-looking all caps name for this) and import that in `Token` to return from `Token.iob_strings`.

adrianeboyd (changes requested):
Just changing the status to make sure the requested `iob_strings` changes don't get lost in the comments.

spacy/attrs.pyx
DuyguA: > Sorry, that wasn't quite what I meant. Leave `Token.iob_strings` as a method (and continue using it in `Doc` and other places), but just define `IOB_STRINGS` in `attrs` as the tuple (we would use a global-looking all caps name for this) and import that in `Token` to return from `Token.iob_strings`.

OK, I applied these comment and it looks like as you asked now.

DuyguA:
> Just changing the status to make sure the requested `iob_strings` changes don't get lost in the comments.

I made a clean up now, sorry for the late arrangements :angel:

svlandeg (approved):
Discussed internally with Adriane - this should now be fine to merge!

PR #9742, opened 2021-11-25
Add Japanese kana characters to default tokenizer exceptions (fix #9693)

### LINKED ISSUE #9693: Kana is not included in the default suffix tokenization rules for English models
`nlp('You should learn words such as 犬.')[-1]` produces `.`, which is as expected.

However, `nlp('You should learn words such as いぬ.')[-1]` gives `いぬ.`, apparently due to both the hiragana and the katakana ranges being omitted from the regular expressions in `nlp.Defaults.suffixes`: `...\u3000-\u303F\u31C0-\u31EF...`.

## Your Environment
- **spaCy version:** 3.1.1
- **Platform:** macOS-11.6-x86_64-i386-64bit
- **Python version:** 3.9.6
- **Pipelines:** en_core_web_trf (3.1.0), en_core_web_lg (3.1.0)

polm:
This includes the main kana, or phonetic characters, used in Japanese. Not having these in the tokenizer exception list means that when they are used in non-Japanese text they behave strangely, and can combine with punctuation.

There are some supplemental kana blocks in Unicode outside the BMP that
could also be included, but because their actual use is rare I omitted
them for now, but maybe they should be added. The omitted blocks are:

- Kana Supplement
- Kana Extended (A and B)
- Small Kana Extension

I didn't add a test, not sure if the speed tradeoff is worth it since it seems other similar classes aren't tested.

### Types of change

tokenizer exception adjustment

## Checklist

### DIFF
--- a/spacy/lang/char_classes.py
+++ b/spacy/lang/char_classes.py
@@ -45,6 +45,10 @@ _hangul_syllables = r"\uAC00-\uD7AF"
 _hangul_jamo = r"\u1100-\u11FF"
 _hangul = _hangul_syllables + _hangul_jamo
 
+_hiragana = r"\u3040-\u309F"
+_katakana = r"\u30A0-\u30FFー"
+_kana = _hiragana + _katakana
+
 # letters with diacritics - Catalan, Czech, Latin, Latvian, Lithuanian, Polish, Slovak, Turkish, Welsh
 _latin_u_extendedA = (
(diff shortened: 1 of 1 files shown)


adrianeboyd:
I think this should go in `develop`. For the history, we can probably just wait until `develop` gets updated from `master` and do the base switch back and forth again. I think it's okay without additional tests.

PR #9748, opened 2021-11-26
Displacy serve entity linking support without `manual=True` support.

narayanacharya6:
The PR adds support for visualizing kb_id on entities when run via displacy.serve

## Description
The PR adds support for visualizing kb_id on entities when run via displacy.serve. The recent changes from [PR 9199](https://github.com/explosion/spaCy/pull/9199) only address rendered html produced with the `manual=True` option set. I believe the same should also be afforded when passing `Doc` objects with entities that have the `kb_id` set. 

For the `kb_url` to be set, one can pass the `kb_url_format_template` (can be named better?) in the `options` argument of `displace.serve`. If not set, `kb_url` defaults to `#` similar to the setting in the above linked PR. Please suggest if there is a better way to do this.

### Types of change
Enhancement

## Checklist

### DIFF
--- a/spacy/displacy/__init__.py
+++ b/spacy/displacy/__init__.py
@@ -184,8 +184,12 @@ def parse_ents(doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
     doc (Doc): Document do parse.
     RETURNS (dict): Generated entities keyed by text (original text) and ents.
     """
+    kb_url_format_template = options.get("kb_url_format_template", None)
     ents = [
-        {"start": ent.start_char, "end": ent.end_char, "label": ent.label_}
+        {"start": ent.start_char, "end": ent.end_char, "label": ent.label_,
+         "kb_id": ent.kb_id_ if ent.kb_id_ else "",
+         "kb_url": kb_url_format_template.format(ent.kb_id_) if kb_url_format_template else "#"
+        }
         for ent in doc.ents
     ]
     if not ents:
--- a/spacy/displacy/__init__.py
+++ b/spacy/displacy/__init__.py
@@ -181,7 +181,8 @@ def parse_deps(orig_doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
 def parse_ents(doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
     """Generate named entities in [{start: i, end: i, label: 'label'}] format.
 
-    doc (Doc): Document do parse.
+    doc (Doc): Document to parse.
+    options (Dict[str, Any]): NER-specific visualisation options.
     RETURNS (dict): Generated entities keyed by text (original text) and ents.
     """
     kb_url_format_template = options.get("kb_url_format_template", None)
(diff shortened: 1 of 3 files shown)


spacy/displacy/__init__.py
```diff
@@ -184,8 +184,17 @@ def parse_ents(doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
     doc (Doc): Document do parse.
     RETURNS (dict): Generated entities keyed by text (original text) and ents.
     """
+    kb_url_format_template = options.get("kb_url_format_template", None)
```
svlandeg, line 187: Maybe just `kb_url_template`?

spacy/displacy/__init__.py
```diff
@@ -184,8 +184,17 @@ def parse_ents(doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
     doc (Doc): Document do parse.
```
svlandeg, line 184: Entirely unrelated to your PR, but it bugs me ;-)
```suggestion
    doc (Doc): Document to parse.
    options (Dict[str, Any]): NER-specific visualisation options.
```

spacy/displacy/__init__.py
svlandeg: Also, we need to document this over at https://spacy.io/api/top-level#displacy_options-ent, right?

svlandeg (commented):
Thanks for the PR, much appreciated! Also good to see an additional test for the new functionality. I had just a few small comments.

website/docs/api/top-level.md
```diff
@@ -326,6 +327,13 @@ or pipeline package can also expose a
 [`spacy_displacy_colors` entry point](/usage/saving-loading#entry-points-displacy)
 to add custom labels and their colors automatically.
 
+By default, displaCy links to `#` for entities without a `kb_id` set on their span. 
+If you wish to link an entity to their URL then consider using the `kb_url_template`
+option from above. For e.g. If the `kb_id` on a span is `Q95` and this is a Wikidata
```
svlandeg, line 332:
```suggestion
option from above. For example if the `kb_id` on a span is `Q95` and this is a Wikidata 
```

website/docs/api/top-level.md
```diff
@@ -326,6 +327,13 @@ or pipeline package can also expose a
 [`spacy_displacy_colors` entry point](/usage/saving-loading#entry-points-displacy)
 to add custom labels and their colors automatically.
 
+By default, displaCy links to `#` for entities without a `kb_id` set on their span. 
+If you wish to link an entity to their URL then consider using the `kb_url_template`
+option from above. For e.g. If the `kb_id` on a span is `Q95` and this is a Wikidata 
+identifier then this option can be set to `https://www.wikidata.org/wiki/{}`.
+Clicking on your entity in the rendered HTML should redirect you to their Wikidata page,
+in this case `https://www.wikidata.org/wiki/Q95`
```
svlandeg, line 335:
```suggestion
in this case `https://www.wikidata.org/wiki/Q95`.
```

svlandeg:
Looks good to me, just want to run this by @ines as well :-)

ines (approved):
Yes, looks good 🎉 And really appreciate the test and docs update!

PR #9764, opened 2021-11-29
morphologizer: avoid recreating label tuple for each token

danieldk:
## Description

The `labels` property converts the dictionary key set to a tuple. This property was used for every annotated token, recreating the tuple over and over again.

Construct the tuple once in the set_annotations function and reuse it.

On a Finnish pipeline that I was experimenting with, this results in a speedup of ~15% (~13000 -> ~15000 WPS).

### Types of change

Performance improvement.

## Checklist

### DIFF
--- a/spacy/pipeline/morphologizer.pyx
+++ b/spacy/pipeline/morphologizer.pyx
@@ -231,12 +231,13 @@ class Morphologizer(Tagger):
         cdef Vocab vocab = self.vocab
         cdef bint overwrite = self.cfg["overwrite"]
         cdef bint extend = self.cfg["extend"]
+        labels = self.labels
         for i, doc in enumerate(docs):
             doc_tag_ids = batch_tag_ids[i]
             if hasattr(doc_tag_ids, "get"):
                 doc_tag_ids = doc_tag_ids.get()
             for j, tag_id in enumerate(doc_tag_ids):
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Nice! I bet `tagger` and `senter` have the same issue, too.

danieldk:
> Nice! I bet `tagger` and `senter` have the same issue, too. (You probably can't measure the speed difference with `senter`, though.)

Good catch, `Tagger` does. I noticed this one while profiling embedding stuff, I guess it shows up because the morphologizer has more labels. I'll add a commit for `Tagger`.

adrianeboyd:
Most of the `morphologizer` code aside from the POS/FEATS label conversion is copied directly from `tagger`, so if one has a bug, the other probably does, too.

danieldk:
There is a bit of low hanging fruit in training as well, but I'll make a separate PR for that (needs a bit more testing). It'll also not be as visible due to higher cost elsewhere.

danieldk:
Seems like CI failed because it could not start a macOS runner?

adrianeboyd:
Merge in `master` and try it again.

svlandeg:
I think the CI just timed out. When this happens, it just needs a push on Azure "rerun failed pipelines".

PR #9859, opened 2021-12-14
Update Tokenizer documentation 

antonpibm:
Update the description of `token_match` and `url_match` arguments for Tokenizer
Currently the callables are described as 

> A boolean function

which is not enough as those functions would be serialized with `_get_regex_pattern`. 

## Description
Description changed to match the correct requirement - result of `re.compile(string).match`

### Types of change
a change to the documentation

## Checklist


svlandeg:
Thanks! It's a good idea to have this in-sync with the docs at https://spacy.io/api/tokenizer#init.

PR #9869, opened 2021-12-15
Span/SpanGroup: wrap SpanC in shared_ptr

danieldk:
## Description

When a `Span` that was retrieved from a `SpanGroup` was modified, these changes were not reflected in the `SpanGroup` because the underlying `SpanC` struct was copied.

This change applies the solution proposed by @nrodnova, to wrap `SpanC` in a `shared_ptr`. This makes a `SpanGroup` and Spans derived from it share the same `SpanC`. So, changes made through a `Span` are visible in the `SpanGroup` as well.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -163,6 +163,7 @@ def test_char_span(doc, i_sent, i, j, text):
         assert span.text == text
 
 
+@pytest.mark.issue(9556)
 def test_modify_span_group(doc):
     group = SpanGroup(doc, spans=doc.ents)
     for span in group:
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -4,7 +4,7 @@ from numpy.testing import assert_array_equal
 
 from spacy.attrs import ORTH, LENGTH
 from spacy.lang.en import English
-from spacy.tokens import Doc, Span, Token
+from spacy.tokens import Doc, Span, SpanGroup, Token
 from spacy.vocab import Vocab
 from spacy.util import filter_spans
 from thinc.api import get_current_ops
@@ -161,6 +161,17 @@ def test_char_span(doc, i_sent, i, j, text):
         assert not span
     else:
         assert span.text == text
+
+
+def test_modify_span_group(doc):
+    group = SpanGroup(doc, spans=doc.ents)
+    for span in group:
+        span.start = 0
+        span.label = doc.vocab.strings["TEST"]
+
+    # Span changes must be reflected in the span group
+    assert group[0].start == 0
+    assert group[0].label == doc.vocab.strings["TEST"]
 
 
 def test_spans_sent_spans(doc):
--- a/spacy/tokens/span.pxd
+++ b/spacy/tokens/span.pxd
@@ -25,4 +25,4 @@ cdef class Span:
 
     cpdef np.ndarray to_array(self, object features)
 
-    cdef SpanC *span_c(self)
+    cdef SpanC* span_c(self)
--- a/spacy/tokens/span.pxd
+++ b/spacy/tokens/span.pxd
@@ -24,3 +24,5 @@ cdef class Span:
         return self
 
     cpdef np.ndarray to_array(self, object features)
+
+    cdef SpanC *span_c(self)
--- a/spacy/tokens/span.pxd
+++ b/spacy/tokens/span.pxd
@@ -1,3 +1,4 @@
+from libcpp.memory cimport shared_ptr
 cimport numpy as np
 
 from .doc cimport Doc
@@ -7,17 +8,17 @@ from ..structs cimport SpanC
 
 cdef class Span:
     cdef readonly Doc doc
-    cdef SpanC c
+    cdef shared_ptr[SpanC] c
     cdef public _vector
     cdef public _vector_norm
 
     @staticmethod
-    cdef inline Span cinit(Doc doc, SpanC span):
+    cdef inline Span cinit(Doc doc, const shared_ptr[SpanC] &span):
         cdef Span self = Span.__new__(
             Span,
             doc,
-            start=span.start,
-            end=span.end
+            start=span.get().start,
+            end=span.get().end
         )
         self.c = span
         return self
--- a/spacy/tokens/span.pyx
+++ b/spacy/tokens/span.pyx
@@ -127,41 +127,46 @@ cdef class Span:
                 return False
             else:
                 return True
+
+        cdef SpanC *span_c = self.span_c()
+        cdef SpanC *other_span_c = other.span_c()
+
         # <
         if op == 0:
-            return self.c.get().start_char < other.c.get().start_char
+            return span_c.start_char < other_span_c.start_char
         # <=
         elif op == 1:
-            return self.c.get().start_char <= other.c.get().start_char
+            return span_c.start_char <= other_span_c.start_char
         # ==
         elif op == 2:
             # Do the cheap comparisons first
             return (
-                (self.c.get().start_char == other.c.get().start_char) and \
-                (self.c.get().end_char == other.c.get().end_char) and \
-                (self.c.get().label == other.c.get().label) and \
-                (self.c.get().kb_id == other.c.get().kb_id) and \
+                (span_c.start_char == other_span_c.start_char) and \
+                (span_c.end_char == other_span_c.end_char) and \
+                (span_c.label == other_span_c.label) and \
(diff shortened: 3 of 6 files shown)


danieldk:
Odd, I cannot reproduce the mypy error locally and it did not happen yesterday.

svlandeg:
Maybe it's something to do with the new 0.920 release https://pypi.org/project/mypy/#history? You can either try restricting on the CI, or update locally to see whether that's the culprit.

danieldk:
> Maybe it's something to do with the new 0.920 release https://pypi.org/project/mypy/#history? You can either try restricting on the CI, or update locally to see whether that's the culprit.

Thanks! That's the issue. I think pydantic is not yet compatible with this release. Opened #9873 to pin the mypy version for now.

spacy/tokens/span.pyx
```diff
@@ -128,39 +129,39 @@ cdef class Span:
...
             # Do the cheap comparisons first
             return (
-                (self.c.start_char == other.c.start_char) and \
-                (self.c.end_char == other.c.end_char) and \
-                (self.c.label == other.c.label) and \
-                (self.c.kb_id == other.c.kb_id) and \
+                (self.c.get().start_char == other.c.get().start_char) and \
+                (self.c.get().end_char == other.c.get().end_char) and \
+                (self.c.get().label == other.c.get().label) and \
+                (self.c.get().kb_id == other.c.get().kb_id) and \
```
svlandeg, line 143: Should we try to avoid all these `get()` calls by caching the reference in a variable if it's used a lot? 

Or alternatively, we could have a new property that runs `self.c.get()` so that the code stays more clear?

spacy/tokens/span.pxd
```diff
@@ -7,17 +8,17 @@ from ..structs cimport SpanC
 
 cdef class Span:
     cdef readonly Doc doc
-    cdef SpanC c
+    cdef shared_ptr[SpanC] c
```
svlandeg, line 11: Just leaving a note that we have to decide whether this goes in 3.3 or 4.0 and in case of the latter, whether we keep it on `develop` or a different branch.

svlandeg (commented):
This looks pretty clean overal, nice work!

spacy/tokens/span.pyx
danieldk: > Should we try to avoid all these get() calls by caching the reference in a variable if it's used a lot?

That's a good idea! We could use a local `SpanC *` variable and reuse that. I'll update the pr.

> Or alternatively, we could have a new property that runs self.c.get() so that the code stays more clear?

I think this is a tricky, because once we give back pointers, they could become invalid if the reference count of `shared_ptr` goes to zero.

danieldk: Though maybe we could make a `cdef` method, since then it'd only be visible from Python?

DuyguA: > I think this is a tricky, because once we give back pointers, they could become invalid if the reference count of `shared_ptr` goes to zero.

Agree, then most probably you'll need getters for the above fields. Suppose we implement this small method as @svlandeg  suggested:

```
def get_real_ptr(self):    # This is OK
    return self.c.get() 
```

Then the above lines create trouble:

```
self.get_real_ptr().kb_id          If the self.c is a nullptr, this is trouble
```

As a result we need to check for nullptr for every if via a :

```
def kb(self):
   ptr = self.get_real_ptr()
   if ptr not None:
     return ptr.kb_id
   else return None 
```

Potentially, one might need to implement a `getter` for nullptr checking for all fields. What do you say @danieldk  , is there a shorter way?

danieldk: > ```
> def get_real_ptr(self):    # This is OK
>     return self.c.get() 
> ```

This wouldn't work, since `SpanC *` cannot be converted to a Python object.

> 
> Then the above lines create trouble:
> 
> ```
> self.get_real_ptr().kb_id          If the self.c is a nullptr, this is trouble
> ```
> 
> As a result we need to check for nullptr for every if via a :
> 
> ```
> def kb(self):
>    ptr = self.get_real_ptr()
>    if ptr not None:
>      return ptr.kb_id
>    else return None 
> ```
> 
> Potentially, one might need to implement a `getter` for nullptr checking for all fields. What do you say @danieldk , is there a shorter way?

danieldk: @svlandeg I have added a `cdef` method `Span.span_c` as a shorthand for `Span.c.get` and use that instead now. Since it's `cdef`'ed it doesn't make it any less safe than what we had before.

In methods that have more than one `Span.span_c` call, I have made a local variable as a shorthand:

```
SpanC *span_c = self.span_c()
// Do stuff with span_c.
```

Does that look better to you?

DuyguA: > > ```
> > def get_real_ptr(self):    # This is OK
> >     return self.c.get() 
> > ```
> 
> This wouldn't work, since `SpanC *` cannot be converted to a Python object.

Ah OK!

svlandeg: It takes a little more getting used to, I think, but I think I do like it better on the longer term? It's not difficult to remember that `span_c` will get you the actual `SpanC`. But happy to hear other's opinions!

spacy/tokens/span.pxd
```diff
@@ -7,19 +8,21 @@ from ..structs cimport SpanC
...
-            end=span.end
+            start=span.get().start,
+            end=span.get().end
         )
         self.c = span
         return self
 
     cpdef np.ndarray to_array(self, object features)
+
+    cdef SpanC *span_c(self)
```
adrianeboyd, line 28:
```suggestion
    cdef SpanC* span_c(self)
```

And all the similar cases in the rest of PR, to match our style (which probably isn't documented formally anywhere, not something you would have necessarily known either way).

requirements.txt
```diff
@@ -31,7 +31,7 @@ pytest-timeout>=1.3.0,<2.0.0
 mock>=2.0.0,<3.0.0
 flake8>=3.8.0,<3.10.0
 hypothesis>=3.27.0,<7.0.0
-mypy>=0.910
+mypy==0.910
```
adrianeboyd, line 34: I'd prefer if this change isn't part of this PR. We can cherry-pick the commit from `master` into `develop` (or otherwise update `develop` from `master`) and then merge `upstream/develop` into this branch.

svlandeg: Agreed! Let's update `develop` with `master` first, we'll need that fix for the CI either way.

danieldk: I can rebase it once it's in `develop`. Then it's completely vanished from the PR's history.

spacy/tokens/span.pxd
danieldk: That's a very un-C'ish style. But I am fine with making those changes for consistency.

danieldk: Done!

adrianeboyd: I know, and I have no idea why it ended up that way, but that's how it currently is.

danieldk: Seems like the Cython docs also use this style. Maybe it's standard for Cython?

Still hoping for a great Cython formatter :).

requirements.txt
svlandeg: `develop` has been updated, so if you pul/merge this should hopefully be resolved

spacy/tokens/span_group.pyx
```diff
@@ -149,13 +150,13 @@ cdef class SpanGroup:
...
-                span.end,
-                span.start_char,
-                span.end_char
+                span.get().id,
+                span.get().kb_id,
+                span.get().label,
+                span.get().start,
+                span.get().end,
+                span.get().start_char,
+                span.get().end_char
```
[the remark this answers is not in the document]
danieldk: `SpanGroup` stores a vector of `shared_ptr[SpanC]`, so `span_c` is not available.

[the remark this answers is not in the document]
danieldk: I'll update this with a single `get()`.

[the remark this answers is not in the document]
danieldk: Done.

spacy/tests/doc/test_span.py
```diff
@@ -163,6 +163,17 @@ def test_char_span(doc, i_sent, i, j, text):
         assert span.text == text
 
 
+def test_modify_span_group(doc):
```
svlandeg, line 166:
```suggestion
@pytest.mark.issue(9556)
def test_modify_span_group(doc):
```

PR #9873, opened 2021-12-16
Pin mypy to 0.910 until there is a compatible pydantic version

danieldk:
## Description

Since mypy 0.920 was released, mypy fails with:

```
setup.cfg:140: error: Error importing plugin "pydantic.mypy": cannot import name 'TypeVarDef' from 'mypy.types' (/Users/daniel/.pyenv/versions/torch-1.10/lib/python3.9/site-packages/mypy/types.cpython-39-darwin.so)
```

It seems that we need to wait for a version of pydantic that is compatible
with this new mypy release. So, pin the mypy version for the time being.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,7 +31,7 @@ pytest-timeout>=1.3.0,<2.0.0
 mock>=2.0.0,<3.0.0
 flake8>=3.8.0,<3.10.0
 hypothesis>=3.27.0,<7.0.0
-mypy>=0.910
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Feel free to merge as soon as the CI goes green :-)

adrianeboyd:
(Curses, since I was hoping we could enable the mypy checks for 3.10 once they released wheels.)

PR #9879, opened 2021-12-16
User fewer Vector internals

danieldk:
## Description

Throughout spaCy, we rely on internals of the `Vector` class. This is a first PR to replace a bunch of such uses by methods/properties. In particular:

* `Vectors.data.shape` -> `Vectors.shape`
* `Vectors.data.size` -> `Vectors.size`
*  `self.vocab.vectors.data = ops.asarray(self.vocab.vectors.data)` -> `self.to_ops(ops)`

This was the lower-hanging fruit 🥝, I'll make a separate PR for more difficult cases.

### Types of change

Cleanup.

## Checklist

### DIFF
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -10,7 +10,7 @@ from typing import cast
 import warnings
 from enum import Enum
 import srsly
-from thinc.api import get_array_module, get_current_ops
+from thinc.api import Ops, get_array_module, get_current_ops
 from thinc.backends import get_array_ops
 from thinc.types import Floats2d
 
@@ -510,6 +510,9 @@ cdef class Vectors:
             [[row2key[row] for row in numpy_rows[i] if row in row2key]
                     for i in range(len(queries)) ], dtype="uint64")
         return (keys, best_rows, scores)
+
+    def to_ops(self, ops: Ops):
+        self.data = ops.asarray(self.data)
 
     def _get_cfg(self):
         if self.mode == Mode.default:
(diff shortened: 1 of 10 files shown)


svlandeg (approved):
I love this kind of PR 🤩 
All looks perfect!

Can't wait for the rest 🍎 🍌 🍇

spacy/vectors.pyx
```diff
@@ -511,6 +511,9 @@ cdef class Vectors:
                     for i in range(len(queries)) ], dtype="uint64")
         return (keys, best_rows, scores)
 
+    def to_ops(self, ops: Ops):
```
svlandeg, line 514: Perhaps we can document this one at https://spacy.io/api/vectors ?

svlandeg: I'm not sure how often it would come up that a user would need to call this directly though.

danieldk: I don't know. But I think it doesn't hurt!

danieldk:

adrianeboyd:
I think it would make sense to add the new `to_ops` to the deserialization here, too, which will be needed for the bigendian ops, see: https://github.com/explosion/murmurhash/issues/26#issuecomment-976565538

danieldk:
> I think it would make sense to add the new `to_ops` to the deserialization here, too, which will be needed for the bigendian ops, see: [explosion/murmurhash#26 (comment)](https://github.com/explosion/murmurhash/issues/26#issuecomment-976565538)

I wonder if it wouldn't be better to provide an `Ops.load` which calls `xp.load` and then does an endianness check + conversion. Then all code that loads data would be robust against different endianness.

Issue #7028, opened 2021-02-11
Explicit prioritization for EntityRuler matches


antelope-admin:
Current documentation [mentions](https://spacy.io/usage/rule-based-matching#entityruler-usage) the behaviour of EntityRuler in case of overlapping matches: 

>  If any matches were to overlap, the pattern matching most tokens takes priority. If they also happen to be equally long, then the match occurring first in the Doc is chosen.

However, there is no mention anywhere of cases where the matches occur at same place in document, _and_ are of same length.

Here's a realistic example of when this would happen. We have a few acronyms that are caught exactly as they appear (in this example I included "ITM" which means 'in the money'), and we want anything that is not explicit but has the morphological shape of `XXX` or `XXXX` to be marked as a ticker symbol (e.g. `AAPL`). 

It looks like currently how things get tagged with the following definitions is not user-specifiable nor deterministic:
 

```python
nlp = spacy.load("en_core_web_trf")

## Define custom rule-based matching logic
ruler = nlp.add_pipe("entity_ruler", before="ner")

patterns = [
    dict(
        label="OPTIONS_SLANG",
        pattern="ITM",
    ),
    dict(
        label="TICKER",
        pattern=[
            {"SHAPE": "XXXX"}
        ],
    ),
    dict(
        label="TICKER",
        pattern=[
            {"SHAPE": "XXX"}
        ],
    ),
]

ruler.add_patterns(patterns)

```

this sometimes results in the "right" parse (e.g. `ITM` would be tagged as `OPTIONS_SLANG`) but more often than not ends up applying the more generic `shape: XXX` pattern and tags everything as `TICKER`: 

```python
example = "buying way too many ITM TSLA options. we're not in IPO times here. ROI is low"
doc = nlp(example)
print([
    (ent.text, ent.label_)
    for ent in doc.ents
])
[('ITM', 'TICKER'), ('TSLA', 'TICKER'), ('IPO', 'TICKER'), ('ROI', 'TICKER')]
```

Further, it seems that there are some non-local dependencies at work here; e.g. if I change the above text slightly to remove last match (`ROI`), things seem to work more "correctly" consistently:

```python
example = "buying way too many ITM TSLA options. we're not in IPO times here."
doc = nlp(example)
[
    (ent.text, ent.label_)
    for ent in doc.ents
]
[('ITM', 'OPTIONS_SLANG'), ('TSLA', 'TICKER'), ('IPO', 'TICKER')]
```

TLDR:

[...]

PR #9880, opened 2021-12-16
Add SpanRuler component

adrianeboyd:
## Description

Add a `SpanRuler` component similar to `EntityRuler` that saves a list of matched spans to `Doc.spans[spans_key]`. The matches from the token and phrase matchers are deduplicated and sorted before assignment but are not otherwise filtered.

### Types of change

Feature.

## Checklist


adrianeboyd:
Some open issues / questions:

* Are there better options to combine related functionality in `EntityRuler` and `SpanRuler`? It seems like `SpanRuler` is the more generic option and `EntityRuler` should be a child class, but the kind of non-standard serialization in `EntityRuler` and `ent_id` processing makes this difficult.
* Should there be support for `Span.kb_id` similar to `ent_id`? (I will note that the `ent_id` support in `EntityRuler` is relatively hacky and might be useful to refactor both related to this.)
* Does it make sense to include `patterns` as an init kwarg? I think it should be removed, since this method of initialization, while used by `EntityRuler` and here in the similar tests copied from `EntityRuler`, isn't really how it should be initialized in v3.
* Add tests for scoring (note that passing `spans_key` to the scorer currently requires it to be in `self.cfg`).

Future improvements for `EntityRuler` and `SpanRuler`:

* Registered sorting/filtering method that is called in `.match`

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,345 @@
...
+            (m_id, start, end) for m_id, start, end in matches if start != end
+        )
+        get_sort_key = lambda m: (m[2] - m[1], m[1])
+        final_matches = sorted(deduplicated_matches, key=get_sort_key)
+        return final_matches
+
+    def set_annotations(self, doc, matches):
+        """Modify the document in place"""
+        if not self.cfg["spans_key"] in doc.spans:
+            doc.spans[self.cfg["spans_key"]] = []
```
svlandeg, line 158: We can have a property that accesses `self.cfg["spans_key"]`- for `spancat` it's just `key` so maybe we should take the same?

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,345 @@
...
+            "validate": validate,
+        }
+        self.clear()
+        if patterns:
+            self.add_patterns(patterns)
+        self.scorer = scorer
+
+    def __len__(self) -> int:
+        """The number of all labels added to the span ruler."""
+        return len(set([p["label"] for p in self._patterns]))
```
svlandeg, line 118: The `entity_ruler` defines `__len__` as the number of patterns, so this definition might be unexpected?

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,345 @@
...
+        *,
+        nlp: Optional[Language] = None,
+        patterns: Optional[Sequence[PatternType]] = None,
+    ):
+        """Initialize the pipe for training.
+
+        get_examples (Callable[[], Iterable[Example]]): Function that
+            returns a representative sample of gold-standard Example objects.
+        nlp (Language): The current nlp object the component is part of.
+        patterns Optional[Iterable[PatternType]]: The list of patterns.
```
svlandeg, line 189:
```suggestion
        patterns (Optional[Iterable[PatternType]]): The list of patterns.
```

svlandeg (commented):
I've been wanting this component as well :-)

> Should there be support for `Span.kb_id` similar to `ent_id`? (I will note that the `ent_id` support in `EntityRuler` is relatively hacky and it might be useful to refactor both related to this.)

This has been bothering me a lot too, I agree that there should be a more clean way to support `kb_id` and avoid the hackiness of the separator logic for `ent_id`. In general I think some of the implementation principles you used here should be ported to `entity_ruler` at some point.

> Are there better options to combine related functionality in `EntityRuler` and `SpanRuler`? It seems like `SpanRuler` is the more generic option and `EntityRuler` should be a child class, but the kind of non-standard serialization in `EntityRuler` and `ent_id` processing make this difficult.

Semantically they really could be almost the same component. If you can customize the filtering function before you set annotations, and are able to specify either a key in `doc.spans` or use `doc.ents`, you could cover all use-cases. Two different factories would then simply set different defaults. 

If you'd have two filtering functions (one for `spans`, one for `ents`), then you could even offer storing the "raw hits" in `doc.spans` and the filtered ones in `doc.ents`. Or we should reconsider this entirely and have the filtering from `doc.spans` to `doc.ents` as a separate small component. Unfortunately either way we go, we'll probably have to break bwd-compat, which is how the `entity_ruler` code ended up in this state :|

From a practical point of view, it would be nice to get this new component out sooner rather than later, and maybe consider the refactoring later? What do you think?

> Does it make sense to include `patterns` as an init kwarg? I think it should be removed, since this method of initialization, while used by `EntityRuler` and here in the similar tests copied from `EntityRuler`, isn't really how it should be initialized in v3.

I agree. We should at least not support it in the new component.

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,345 @@
...
+        scorer=scorer,
+    )
+
+
+class SpanRuler(Pipe):
+    """The SpanRuler lets you add spans to the `Doc.spans` using token-based
+    rules or exact phrase matches.
+
+    DOCS: https://spacy.io/api/spanruler
+    USAGE: https://spacy.io/usage/rule-based-matching#spanruler
```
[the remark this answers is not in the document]
adrianeboyd: This was just from a search and replace. The docs would come in this PR after the implemention is more settled.

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,345 @@
...
+
+    def clear(self) -> None:
+        """Reset all patterns."""
+        self._patterns: List[PatternType] = []
+        self.matcher: Matcher = Matcher(
+            self.nlp.vocab, validate=cast(bool, self.cfg["validate"])
+        )
+        self.phrase_matcher: PhraseMatcher = PhraseMatcher(
+            self.nlp.vocab,
+            attr=cast(bool, self.cfg["phrase_matcher_attr"]),
```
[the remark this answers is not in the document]
adrianeboyd: I only addressed cases that `mypy` complained about.

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,345 @@
...
+            matches = self.match(doc)
+            self.set_annotations(doc, matches)
+            return doc
+        except Exception as e:
+            return error_handler(self.name, self, [doc], e)
+
+    def match(self, doc: Doc):
+        self._require_patterns()
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message="\\[W036")
```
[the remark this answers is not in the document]
adrianeboyd: This is separate because you can have just token patterns or just phrase matcher patterns and if one is empty, you get this warning.

spacy/pipeline/span_ruler.py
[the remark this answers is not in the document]
svlandeg: But this shouldn't be cast to `bool`, right?

spacy/pipeline/span_ruler.py
adrianeboyd: Sure.

spacy/pipeline/span_ruler.py
adrianeboyd: It's unclear what else you might count, though. I am open to suggestions.

spacy/pipeline/span_ruler.py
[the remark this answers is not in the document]
adrianeboyd: That is weird, hmm, what was going on...

adrianeboyd:
(No idea what I accidentally clicked to review my own comment...)

spacy/pipeline/span_ruler.py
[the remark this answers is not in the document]
adrianeboyd: Maybe `mypy` accepted a cast from `False` to `None`? I am not sure why this made it through the check at all?

spacy/pipeline/span_ruler.py
adrianeboyd: Ah, I think was originally trying not to track the added patterns to not have things duplicated in the component and the matchers, but added it back at some point to make a few other functions easier. We could reconsider that, I guess, although I think not serializing the patterns in their original input format has some drawbacks for users who might want to edit the patterns directly.

spacy/pipeline/span_ruler.py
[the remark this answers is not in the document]
svlandeg: Yea exactly, me neither. `mypy` weirdness.

spacy/tests/pipeline/test_span_ruler.py
```diff
@@ -0,0 +1,394 @@
...
+        {"label": "BYE", "pattern": [{"LOWER": "bye"}, {"LOWER": "bye"}]},
+        {"label": "HELLO", "pattern": [{"ORTH": "HELLO"}]},
+        {"label": "COMPLEX", "pattern": [{"ORTH": "foo", "OP": "*"}]},
+        {"label": "TECH_ORG", "pattern": "Apple"},
+        {"label": "TECH_ORG", "pattern": "Microsoft"},
+    ]
+
+
+@pytest.mark.issue(8216)
+def test_span_ruler_fix8216(nlp, patterns):
```
svlandeg, line 32:
```suggestion
def test_span_ruler_add_empty(nlp, patterns):
```

adrianeboyd:
I am not sure whether having `spans_key = null` as the setting for setting `doc.ents` is the best idea, but using one setting rather than two potentially incompatible settings seemed slightly better for now.

If there is an additional ID, it could potentially be saved as `span.id` (currently completely unused as far as I can tell) instead of `span.kb_id` to be a bit more generic. It's a little weird that you'd have `span.kb_id` but `token.ent_id` instead of `token.kb_id`. (There are not currently any python properties for accessing `span.id` or `span.id_`. It's easy to add, but the `id` keyword issue is slightly annoying.)

svlandeg:
> but using one setting rather than two potentially incompatible settings seemed slightly better for now.

Would they be incompatible though? I would just save the results both in `doc.spans` as well as in `doc.ents` if both are specified?

svlandeg:
Thinking about this more, I really think that having this component cover all cases could be powerful: you could be running the same matching rules once, store the raw hits in `doc.spans` (i.e. no filter for `spans`) and define some proper filter for `doc.ents` and set those in the same sweep. I would have definitely been able to use that kind of functionality in the past.

adrianeboyd:
Sure, I can see how having it always set `doc.spans` and optionally set `doc.ents` makes sense, although it makes the filtering situation more complicated. I guess you'd need a `spans_filter` and an `ents_filter` setting. I guess that's okay, though?

My plan was for it to fail ungracefully if you define a bad filter for `doc.ents`.

svlandeg:
Yea exactly, I think in that case we'd need to allow for two different filter settings. I do think we can throw a custom error that tells the user to set a proper `ents_filter` when setting `doc.ents` fails. And then the default for this function would be `util.filter_spans`, so users would be have to change this specifically to be able to run into trouble...

adrianeboyd:
I decided instead to make both optional, with a required `ents_filter` set to `util.filter_spans`.

adrianeboyd:
The complications from changing the target branch are minimal and I hope they should go away the next time `develop` is updated.

adrianeboyd:
The last commit changes the behavior of `token.ent_id` slightly. Now setting `doc.ents` clobbers any `ent_id` values in those entity spans. This doesn't seem like a terrible decision, but it might break some people's code, as it did for the entity ruler.

spacy/tokens/span.pyx
```diff
@@ -101,6 +101,8 @@ cdef class Span:
             label = doc.vocab.strings.add(label)
         if isinstance(kb_id, str):
             kb_id = doc.vocab.strings.add(kb_id)
+        if isinstance(id, str):
+            id = doc.vocab.strings.add(id)
```
svlandeg, line 105: I think this code is unideal because `id` is a built-in Python function :/

adrianeboyd: I know, but all the other alternatives for the kwarg here seemed worse?

svlandeg: why not `span_id`?

adrianeboyd: I think `Span(span_id=1234)` is a little weird, but I guess that could work.

adrianeboyd:
This does not yet support `EntityRuler`-type `overwrite_ents=True` where only overlapping entities are removed and not all entities.

adrianeboyd:
This is getting really option-heavy but users have wanted custom sorting of the matches in the past (#7028). The `future_entity_ruler` component has identical options to `entity_ruler`, only the `ent_id_sep` doesn't do anything.

There's no way to get a non-weird sorting in `doc.spans` if you need weird sorting for the custom filtering in `doc.ents`. I don't think this a huge deal in the end, and since `overwrite=False` appends the new spans, I don't just want to sort them.

The main remaining things:

* Whether to revert the `doc.ents` `token.ent_id` behavior discussed in: https://github.com/explosion/spaCy/pull/9880#issuecomment-1020243254
* How to handle the API for the `remove` behavior. `SpanRuler.remove` removes by label (which seems like the best default to me?) but `EntityRuler.remove` removes by `id`. I think the reasoning was that you could remove patterns from the matchers yourself by label, but you couldn't easily remove them by `ent_id`. (If you remove them yourself by label, I think the internal state with the saved patterns is out-of-sync, though, so we don't want to encourage this.)
* Whether we want to keep `SpanRuler.token_patterns` and `SpanRuler.phrase_patterns`. This is so minimal that it doesn't seem like a huge problem to keep it, but it isn't really needed.
* Docs

adrianeboyd:
* Docs are added (and as usual, writing them makes you rethink some API decisions, so the `spans_filter` and `ents_filter` now have the same signature including the existing spans as part of the args provided to the filter).
* The `token_patterns` and `phrase_patterns` behavior wasn't actually the same as the entity ruler and it didn't really seem that helpful vs. `SpanRuler.patterns`, so I removed them.

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,572 @@
...
+    """
+    entities = list(entities)
+    new_entities = []
+    seen_tokens: Set[int] = set()
+    seen_tokens.update(*(range(ent.start, ent.end) for ent in entities))
+    for span in spans:
+        start = span.start
+        end = span.end
+        # check for end - 1 here because boundaries are inclusive
+        if start not in seen_tokens and end - 1 not in seen_tokens:
```
svlandeg, line 168: Same comment about having to check all tokens instead of just `start` and `end-1`

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,572 @@
...
+    return SpanRuler(
+        nlp,
+        name,
+        spans_key=None,
+        spans_filter=None,
+        annotate_ents=True,
+        ents_filter=ents_filter,
+        phrase_matcher_attr=phrase_matcher_attr,
+        validate=validate,
+        overwrite=False,
```
svlandeg, line 59: Shouldn't this also be a setting in the config of `future_entity_ruler`? Otherwise you wouldn't have the option to clear `doc.ents` entirely and only use the new matches?

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,572 @@
...
+    phrase_matcher_attr: Optional[Union[int, str]],
+    validate: bool,
+    overwrite_ents: bool,
+    scorer: Optional[Callable],
+    ent_id_sep: str,
+):
+    if overwrite_ents:
+        ents_filter = overwrite_overlapping_ents_filter
+    else:
+        ents_filter = preserve_existing_ents_filter
```
svlandeg, line 49: I'm not sure whether the names will be intuitive/understandable to users. I wonder whether we should focus on the "priority" aspect instead --> something like `prioritize_matches_ents_filter` and `prioritize_existing_ents_filter`, though these names are equally long :/

svlandeg (commented):
This has come along really nicely. Sure, it's option-heavy, but I really like the flexibility this component brings now!

With respect to the question around `token.ent_id`, I feel like we should avoid to be breaking and fix this in v4 instead.

I haven't finished review entirely yet, but I wanted to get a first batch of comments out.

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,572 @@
...
+    },
+)
+def make_entity_ruler(
+    nlp: Language,
+    name: str,
+    phrase_matcher_attr: Optional[Union[int, str]],
+    validate: bool,
+    overwrite_ents: bool,
+    scorer: Optional[Callable],
+    ent_id_sep: str,
```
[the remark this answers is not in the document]
adrianeboyd: I thought we could potentially remove it in `v4`. For now it makes it easier to replace `entity_ruler` with `future_entity_ruler` without any other changes to your code.

spacy/pipeline/span_ruler.py
adrianeboyd: I was initially trying to restrict it to the exact same config that's currently supported by `entity_ruler`, which only has `overwrite_ents`, and I was worried that both `overwrite` and `overwrite_ents` would be too confusing.

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,572 @@
...
+    """Merge entities and spans into one list without overlaps by allowing
+    spans to overwrite any entities that they overlap with. Priority is given
+    to the earlier spans in the provided spans. Intended to replicate the
+    overwrite_ents=True behavior from the EntityRuler.
+
+    entities (Iterable[Span]): The entities, already filtered for overlaps.
+    spans (Iterable[Span]): The spans to merge, may contain overlaps.
+    RETURNS (List[Span]): Filtered list of non-overlapping spans.
+    """
+    entities = list(entities)
```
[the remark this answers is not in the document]
adrianeboyd: I've been trying avoid any pre-emptive filtering or checking because it will make it slower for all cases.

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,572 @@
...
+    new_entities = []
+    seen_tokens: Set[int] = set()
+    for span in spans:
+        start = span.start
+        end = span.end
+        # check for end - 1 here because boundaries are inclusive
+        if start not in seen_tokens and end - 1 not in seen_tokens:
```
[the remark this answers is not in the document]
adrianeboyd: Ah, I think this works in `util.filter_spans` with the fixed sorting/filtering but not in the general case.

spacy/pipeline/span_ruler.py
svlandeg: I don't think we should try to mimick the old config from `entity_ruler`, as stuff like `overwrite_ents` was confusing/ambiguous before. This is a new factory, so why not introduce it like we want it to be? People likely won't switch over until we make them in v4 anyway?

svlandeg: Ah, I guess that's where we differ in goal. I was thinking we'd want to revisit some of the API decisions around `entity_ruler` as well (but not be breaking until v4)

adrianeboyd: I think a ton of people use the entity ruler with its current options and it's unnecessarily breaking to modify, even in v4, especially if we can support it with this with very little extra code to maintain.

svlandeg: If I understand you correctly, that means that users will either stick to the exact functionality of the old `entity_ruler`, or move to the `span_ruler` which will introduce many more settings at once. I was hoping we'd have some sort of "intermediate" step where the config would be largely the same as the old `entity_ruler`, but added a few features from the new implementation. But I see your point about wanting to be a minimally breaking as possible to avoid forcing people to update their code if they don't need any new features.

adrianeboyd: Yeah, in line with some of the other proposed changes for v4 I think it could be okay to change some of python API a little (maybe deprecate `EntityRuler`, which you shouldn't be importing directly anyway anymore, and `SpanRuler wouldn't have patterns in `__init__` or have `to_disk("patterns.jsonl")`, but I'd rather not break people's configs if it's not necessary. (So far I think the only proposed changes that would break configs are related to `nlp.lang`?)

adrianeboyd:
I didn't put the sorting in the filters for speed reasons in case you're saving both `spans` and `ents` so you wouldn't have to sort in both filters, but I wonder whether this is just too confusing, and a relatively rare use case overall anyway.

I suspect that separating the sorting from the filtering makes it harder conceptually for people adding new filters, too.

svlandeg:
Yea, it feels like decoupling the sorting with the filter might be too bug prone :(

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,572 @@
...
+        scorer=scorer,
+    )
+
+
+def overwrite_overlapping_ents_filter(
+    entities: Iterable[Span], spans: Iterable[Span]
+) -> List[Span]:
+    """Merge entities and spans into one list without overlaps by allowing
+    spans to overwrite any entities that they overlap with. Priority is given
+    to the earlier spans in the provided spans. Intended to replicate the
```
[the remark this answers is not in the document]
adrianeboyd: I think this is less confusing with the sorting moved into the filter.

spacy/pipeline/entityruler.py
```diff
@@ -182,10 +182,7 @@ def set_annotations(self, doc, matches):
             if start not in seen_tokens and end - 1 not in seen_tokens:
                 if match_id in self._ent_ids:
                     label, ent_id = self._ent_ids[match_id]
-                    span = Span(doc, start, end, label=label)
-                    if ent_id:
-                        for token in span:
-                            token.ent_id_ = ent_id
+                    span = Span(doc, start, end, label=label, span_id=ent_id)
```
adrianeboyd, line 185: This edit is no longer required, but it seems cleaner so I left it for now.

spacy/pipeline/span_ruler.py
adrianeboyd: I get confused by `prioritize_matches_ents`.

- `prioritize_existing_ents_filter`
- `prioritize_new_ents_filter`

svlandeg: I could live with those!

adrianeboyd:
Is there a better name for `future_entity_ruler`?

spacy/pipeline/span_ruler.py
```diff
@@ -0,0 +1,568 @@
...
+    return prioritize_existing_ents_filter
+
+
+def overlapping_labeled_spans_score(
+    examples: Iterable[Example], *, spans_key=DEFAULT_SPANS_KEY, **kwargs
+) -> Dict[str, Any]:
+    kwargs = dict(kwargs)
+    attr_prefix = f"spans_"
+    kwargs.setdefault("attr", f"{attr_prefix}{spans_key}")
+    kwargs.setdefault("allow_overlap", True)
```
svlandeg, line 183: Considering the specific naming of this function, I would also specifically set `labeled` to `True`

PR #9903, opened 2021-12-18
Handle "_" value for token pos in conllu data

apjanco:
## Description
If a conllu file has no data for a token's part of speech, `conllu_sentence_to_doc()` throws an error that `'_'` is not a valid upos value.  
 To allow tokens with no pos data, this PR changes the "_" to "" to allow pos assignment during `spacy convert` from conllu. It's a small change that will help our users with less than complete and polished treebanks. 

### Types of change
Handles an exception when UD has a upos value of `"_"`.

## Checklist

### DIFF
--- a/spacy/training/converters/conllu_to_docs.py
+++ b/spacy/training/converters/conllu_to_docs.py
@@ -188,11 +188,10 @@ def conllu_sentence_to_doc(
         id_ = int(id_) - 1
         head = (int(head) - 1) if head not in ("0", "_") else id_
         tag = pos if tag == "_" else tag
+        pos = pos if pos != "_" else ""
         morph = morph if morph != "_" else ""
         dep = "ROOT" if dep == "root" else dep
         lemmas.append(lemma)
-        if pos == "_":
-            pos = ""
         poses.append(pos)
         tags.append(tag)
         morphs.append(morph)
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Thanks, this is a good idea, since spacy definitely supports missing values in annotation. Let me just make a short suggestion (you can commit it directly from the comment) that uses the same style as the surrounding code.

spacy/training/converters/conllu_to_docs.py
```diff
@@ -191,6 +191,8 @@ def conllu_sentence_to_doc(
         morph = morph if morph != "_" else ""
         dep = "ROOT" if dep == "root" else dep
         lemmas.append(lemma)
+        if pos == "_":
+            pos = ""
```
adrianeboyd, line 195:
```suggestion
        pos = pos if pos != "_" else ""
        morph = morph if morph != "_" else ""
        dep = "ROOT" if dep == "root" else dep
        lemmas.append(lemma)
```

adrianeboyd:
And looking at this again, I don't think the morph `_` value is handled correctly here, since this is a valid empty value rather than a missing value, but this would be a separate PR...

apjanco:
Thank you for all your help and style suggestion. My first PR with spaCy😊

PR #9905, opened 2021-12-19
Support get_noun_chunks serialization

antonpibm:
## Description
The ability to set a custom `get_noun_chunks` method for [`Vocab`](https://spacy.io/api/vocab#attributes) object was added in version 3.0, but when either `Vocab` or the `Language` objects are serialized using the methods presented in [Saving and Loaded](https://spacy.io/usage/saving-loading) the custom implementation is overridden with default `get_noun_chunks`.
Note 1: deserialization works when `get_noun_chunks` dump is missing, in `from_bytes` by not calling the setter and in `from_disk` by validating the dump file exist. I'm not familiar with all possible flows, and it's important for backward compatibility so please verify that's suffice.
Note 2: It looks like the definition for pickle/unpickle does have this logic implemented, but the suggested way in all documentation uses the `to_bytes` or `to_disc` serialization methods.

### Types of change

Allow serialization/deserialization of custom `get_noun_chunks` method by adding it to the string(`to_bytes`) or creating a file for it's dump (`to_disk`)
Added a relevant section in the web documentation

## Checklist

Some tests were not ran, summary: `2914 passed, 1099 skipped, 27 xfailed, 1 xpassed, 1 warning in 214.41s (0:03:34)`


adrianeboyd:
This is a very nicely done PR! Unfortunately, we intentionally never use pickle in the `to/from_bytes/disk` methods because it is insecure, so we wouldn't serialize any methods like `get_noun_chunks` directly here. (See: https://spacy.io/usage/saving-loading#pipeline)

Most of the language defaults aside from the tokenizer settings are methods that aren't serialized as part of `to_disk` (also in particular `lex_attr_getters`). The alternatives are:

* use pickle for your purposes, as long as the insecure nature of pickle isn't a problem in your use case (we do make sure that the `nlp` object can be pickled if required; it's a bug if you find a case where pickling doesn't work correctly)
* modify the `syntax_iterators` in the language defaults with a callback in `nlp.before_creation` as described in https://spacy.io/usage/training#custom-code-nlp-callbacks (this requires custom code to be included while training or packaging with `--code` unless you put the registered function in a separate package with an entry point)

antonpibm:
@adrianeboyd Thanks for the suggestions. I get what you say about pickle and have two follow up questions,

1. Would it make sense to add specific `Vocab` config (e.g. `get_noun_chunks`) to the language config (`nlp.config`)?  
2. Simply using `pickle.dump` or `pickle.dumps` does not include the custom function implementation in the serialized object, pickle would only serialize the function's signature, using `srsly` did solve this as it would serialize the implementation itself, btw this is the same for all custom pipe components. When you use pickle, do you expect the functions to be defined already or do you expect the pickle object to encapsulate them?

adrianeboyd:
We implemented versioned language defaults in the config at one point during the development of v3, but it got very verbose and complicated for settings that most users don't modify (aside from the tokenizer settings, which are saved separately), so we decided to revert it in the end.

The basic versioning concept/contract behind the language defaults is that a given spacy minor version will always have the same defaults for a language like `German`, so if you use v3.2.x, you will always get the same `GermanDefaults`. This was a bit less consistent in earlier v2 versions, in particular for the tokenizer settings (again, where it doesn't matter for trained/saved models), but as of v3 we don't modify any language defaults in patch releases (and probably also not in v2.3, but I am not 100% sure).

If you need the exact same language defaults in your config, then you can choose to have a custom language like `CustomGerman` instead of `German` that defines all of the settings from scratch. This is a bit of a hassle (and not frequently tested, so there may some minor bugs, like one that just cropped up related to loading vectors for `de` with the language `custom_de`), but if you want to define all of these settings from scratch be assured that they won't change even if future minor versions, then a custom language is the way to go.

If you only want to modify `get_noun_chunks`, then I'd recommend the callback, but you can use a custom language instead if you'd like. Both would require custom code with `--code`.

I think you're seeing differences between `pickle` and `cloudpickle` (`srsly` wraps/vendors `cloudpickle`): https://github.com/cloudpipe/cloudpickle#overriding-pickles-serialization-mechanism-for-importable-constructs. `pickle` expects the module to be available in the environment where you're unpickling: https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled

adrianeboyd:
I'll go ahead and close this PR, but feel free to open an issue or discussion if you run into any problems.

PR #9917, opened 2021-12-21
Bump sudachipy version

polm:
## Description

Since v0.6.0 SudachiPy uses a new Rust backend, which is much faster. However 0.6.1 has a serious bug and shouldn't be used.

Note that 0.6.2 or a later version should probably be used for training 3.3+ models. This will require a small update to the docs. The dictionary version here should probably also be updated.

### Types of change

dependency bump

## Checklist


setup.cfg
```diff
@@ -106,7 +106,7 @@ apple =
     thinc-apple-ops>=0.0.4,<1.0.0
 # Language tokenizers with external dependencies
 ja =
-    sudachipy>=0.4.9
+    sudachipy>=0.6.2
```
adrianeboyd, line 109: Because of the compatibility/compiling issues, how about `>=0.4.9,!=0.6.1` instead?

The versions used for the models can be managed separately, if necessary.

adrianeboyd: On the other hand the extra barely matters, so it's probably fine.

polm: Yeah, I wouldn't be worried about accidentally installing 6.1, but I think when we bump things like this we might be inclined to use 6.0 as a base here normally, and I wanted to avoid that. 

If we want to keep the lower base and just not do this that's fine too, but the speed bump is probably worth it for most users, so if it's not disruptive to update our models I would suggest we go ahead with it.

Since wheel support is pretty good, I would expect that most users wouldn't even notice the fact that it's using Rust.

adrianeboyd: Well, there aren't any arm64 wheels and having to have rust makes the source install a lot trickier for your average user.

adrianeboyd: I think if the old version still works I would leave it here.

polm: Oh, I wasn't aware of the lack of arm64 support. In that case I guess we can leave the old version.

adrianeboyd: To follow up again, I'd be fine with either leaving it as-is or using `>=0.4.9,!=0.6.1` to make this problem explicit. It probably would be a good idea to make it explicit in the trained pipelines either way.

polm: OK, let's change to the explicit version spec so we don't forget it. I'll update this PR.

adrianeboyd:
@explosionbot please test_slow

adrianeboyd:
@explosion-bot bot please test_slow

adrianeboyd:

polm:
I think it's less important, since we accept some potential drift in dictionary version anyway, but I would be in favor of updating to the most recent version. On the other hand [it looks like](https://pypi.org/project/SudachiDict-core/#:~:text=In%20case%20you%20are%20using%20SudachiPy%20before%20v0.5.2) something changed with 5.2 and if you have an older version of SudachiPy you may not be able to use more recent dictionaries...? 

I'll go ahead and bump the minimum version to 5.2 and use the most recent dictionary, though if we'd rather leave it unchanged that should be fine too.

PR #9930, opened 2021-12-23
Add visualisations for parsed documents

richardpaulhudson:
## Description
This PR enables a tabular representation of all or part of a document to be outputted to the console. The new method is documented [here](https://deploy-preview-9930--spacy.netlify.app/usage/visualizers#pprint). An example of the output is shown below:

displacy_pprint

### Types of change
Enhancement

## Checklist


spacy/visualization.py
```diff
@@ -0,0 +1,601 @@
...
+        # Check sent is really a sentence
+        assert sent.start == sent[0].sent.start
+        assert sent.end == sent[0].sent.end
+        heads: List[Optional[int]] = [
+            None
+            if token.dep_.lower() == "root" or token.head.i == token.i
+            else token.head.i - sent.start
+            for token in sent
+        ]
+        # Check there are no head references outside the sentence
```
ines, line 160: We should replace these naked asserts with actual explicit checks and raise a `ValueError` if needed. I think we can also refactor the conditional to make it more compact and take up fewer lines when auto-formatted.

spacy/visualization.py
```diff
@@ -0,0 +1,601 @@
...
+                ):
+                    # Cross over the existing vertical line, which is owing to a non-projective tree
+                    continue
+                if (
+                    char_matrix[working_token_index][working_horizontal_position]
+                    != SPACE
+                ):
+                    # Draw the arrowhead to the right of what is already there
+                    char_matrix[working_token_index][
+                        working_horizontal_position + 1
```
ines, line 333: Nitpick, but I wonder if we can come up with some shorter variables names across this module? It seems like we're causing a lot of unnecessary line breaks during formatting because the lines get too long. For instance, I'd be fine with shortening `horizontal_position` to `horiz_pos` etc.

setup.cfg
```diff
@@ -48,7 +48,7 @@ install_requires =
     preshed>=3.0.2,<3.1.0
     thinc>=8.0.12,<8.1.0
     blis>=0.4.0,<0.8.0
-    wasabi>=0.8.1,<1.1.0
+    wasabi>=0.9.0,<1.1.0
```
svlandeg, line 51: You need to make sure that the version is consistent with the one from `requirements.txt` - this is actually why the test `test_build_dependencies` fails.

spacy/tests/test_visualization.py
```diff
@@ -0,0 +1,1080 @@
...
+    )
+
+    with pytest.raises(AssertionError):
+        Visualizer.render_dependency_tree(doc[1:3], True)
+
+
+def test_visualization_dependency_tree_highly_nonprojective(en_vocab):
+    """Test a highly non-projective tree (colloquial Polish)."""
+    doc = Doc(
+        en_vocab,
```
svlandeg, line 171: Assuming this is a Polish sentence, lets use the Polish vocab :-)

spacy/tests/test_visualization.py
```diff
@@ -0,0 +1,1080 @@
...
+        "╠══>",
+        "╠>╔═",
+        "╠═══",
+        "║ ╚>",
+        "╚══>",
+    ]
+
+
+def test_visualization_render_native_attribute_int(en_vocab):
+    doc = Doc(
```
svlandeg, line 210: If this exact same `doc` is being used across tests, you could define it as a fixture on top.

svlandeg: Also, maybe some tests can be merged? It's fine to test a few related things in one bigger test.

spacy/tests/test_visualization.py
```diff
@@ -0,0 +1,1080 @@
...
+        AttributeFormat(
+            "dep_",
+        ).render(doc[2], whole_row_fg_color=8, whole_row_bg_color=9)
+        == "\x1b[38;5;8;48;5;9mdep\x1b[0m"
+        if supports_ansi
+        else "dep"
+    )
+
+
+def test_visualization_render_whole_row_colors_with_value_dependent_colors(en_vocab):
```
svlandeg, line 328: Maybe this name is slightly too long ;-)

spacy/tests/test_visualization.py
```diff
@@ -0,0 +1,1080 @@
...
+        "",
+        "",
+        "3",
+        "",
+        "",
+        "7",
+        "7",
+        "3",
+        "",
+    ]
```
svlandeg, line 533: In these cases, when the `black` formatting pushes each item onto one line, it could make sense to define the expected list upfront in a var, then compare to that var. In general I think it would be good to compress some of these tests a bit to make things a bit more readible.

spacy/visualization.py
```diff
@@ -0,0 +1,601 @@
+from os import linesep
+from typing import Dict, List, Optional, Union, cast
```
svlandeg, line 2:
```suggestion
from typing import Dict, List, Optional, Union, cast
from os import linesep
```
This is a convention we should probably document, but the `from typing` imports are typically the very first line.

spacy/visualization.py
```diff
@@ -0,0 +1,601 @@
...
+
+    def __init__(
+        self,
+        attribute: str,
+        *,
+        name: str = "",
+        aligns: str = "l",
+        max_width: Optional[int] = None,
+        fg_color: Union[str, int, None] = None,
+        bg_color: Union[str, int, None] = None,
```
svlandeg, line 55: We typically wouldn't add `None` to the `Union`, but use `Optional` instead.

spacy/visualization.py
```diff
@@ -0,0 +1,601 @@
...
+        render_all_colors_within_values: bool = False,
+        whole_row_fg_color: Union[int, str, None] = None,
+        whole_row_bg_color: Union[int, str, None] = None,
+    ) -> str:
+        """
+        ignore_colors:                      no colors should be rendered, typically because the values are required to calculate widths
+        render_all_colors_within_values:    when rendering a table, self.fg_color and self.bg_color are rendered in Wasabi.
+                                                This argument is set to True when rendering a text to signal that colors should be rendered here.
+        whole_row_fg_color:                 a foreground color used for the whole row. This takes precedence over value_dependent_fg_colors.
+        whole_row_bg_color:                 a background color used for the whole row. This takes precedence over value_dependent_fg_colors.
```
svlandeg, line 100:
```suggestion
        whole_row_bg_color:                 a background color used for the whole row. This takes precedence over value_dependent_bg_colors.
```

spacy/visualization.py
```diff
@@ -0,0 +1,601 @@
...
+            assert (
+                len(
+                    [
+                        i
+                        for i in indices_in_current_column
+                        if i in all_indices_ordered_by_column
+                    ]
+                )
+                == 0
+            )
```
svlandeg, line 188: I'm finding this pretty hard to read - is there a way to rewrite this so it won't be spread out as much?

svlandeg (commented):
Thanks for this PR Richard! I've already had a first look through the code and left some comments. I want to do some more experiments actually running & testing it later.

richardpaulhudson:
One difficulty with this PR is that the proliferation of ANSI control characters and carriage returns mean that the tests are mostly not human-readable. They can however be easily understood by adding a `print()` statement to each test to display the output and running `pytest` with an appropriate option like `-rP` to display standard output.

It would be non-standard, but I wonder whether it might even make sense to retain the `print()`s in the repository?

adrianeboyd:
I think text visualizations will be a great addition!

It feels like this should fit under `displacy` rather than as a method on `Doc`, but I'm not sure about the API/naming.

svlandeg:
Closing for now & adding to our internal backlog

PR #9942, opened 2021-12-27
Warn about entities that cross sentence boundaries

DuyguA:
## Description

We want to warn the users about entities that crosses the sentence boundaries. If the user wants to set the entities themselves this way, most probably it's a mistake and they should be aware of the indices. If our NER returns such entity spans (due to annotation mistakes), we still warn the users.

### Types of change

Minor enhancement

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -192,6 +192,10 @@ class Warnings(metaclass=ErrorsWithCodes):
     W115 = ("Skipping {method}: the floret vector table cannot be modified. "
             "Vectors are calculated from character ngrams.")
     W116 = ("Unable to clean attribute '{attr}'.")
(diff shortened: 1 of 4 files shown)


svlandeg:
The original discussion we had around this (and the original ticket) refered to having a warning when running `debug data` - i.e. catch likely annotation mistakes early. I don't think we'd want this warning as part of the `set_ents` function though.

DuyguA:
> The original discussion we had around this (and the original ticket) refered to having a warning when running `debug data` - i.e. catch likely annotation mistakes early. I don't think we'd want this warning as part of the `set_ents` function though.

Ah OK, then I'll change the PR for sure.

Issue #9696, opened 2021-11-18
Span.__richcomp__ issue


nrodnova:
## How to reproduce the behaviour

When sorting Spans using just `sorted()` method and relying on `Span` to handle it, it sorts only on `Span.start` which makes spans that start with the same token to be in random order (or, rather, the original one). It doesn't look right to me.
Here's the code to demonstrate it

```python
import spacy
from random import Random
from spacy.matcher import Matcher
from spacy.tokens import Span

nlp = spacy.load("en_core_web_sm")

doc = nlp('0 1 2 3 4 5 6')
matcher = Matcher(nlp.vocab, validate=True)

matcher.add('4', [[{}, {}, {}, {}]])
matcher.add('2', [[{}, {}, ]])
matcher.add('1', [[{}, ]])

matches = matcher(doc)
spans = []
for match in matches:
    spans.append(Span(doc, match[1], match[2], nlp.vocab.strings[match[0]]))

Random(42).shuffle(spans)
spans = spans[:10] # Truncate a bit for readability
print("Sorted: [%s]" % "', '".join([x.text for x in sorted(spans)]))

expected = sorted(spans, key = lambda x : (x.start, x.end))
print("Expected: [%s]" % "', '".join([x.text for x in expected]))
```
The output of this fragment is:
```
Sorted: [0 1 2 3', '0 1', '1 2 3 4', '2 3', '3', '3 4 5 6', '3 4', '4', '5 6', '5]
Expected: [0 1', '0 1 2 3', '1 2 3 4', '2 3', '3', '3 4', '3 4 5 6', '4', '5', '5 6]
```
As you can see, in the first line there is no consistency, sometimes longer spans come first, and sometimes shorter.

I suggest to change `Span.__richcomp__` for '<, <=, >, >=' to the following:
(if you agree, I can make the change while I am on it)
```python
        # <
        if op == 0:
            return (self.c.start_char, self.c.end_char) < (other.c.start_char, other.c.end_char)
        # <=
        elif op == 1:
            return (self.c.start_char, self.c.end_char) <=  (other.c.start_char, other.c.end_char)

        ## Similar for > and >=
``` 
## Info about spaCy

- **spaCy version:** 3.0.6
- **Platform:** Darwin-20.3.0-x86_64-i386-64bit
- **Python version:** 3.7.9
- **Pipelines:** en_core_web_lg (3.0.0), en_core_web_md (3.0.0), en_core_web_sm (3.0.0)

[...]

PR #9956, opened 2021-12-29
Span richcmp fix

nrodnova:
When sorting Spans using just `sorted()` method and relying on Span to handle it, it sorts only on `Span.start` which makes spans that start with the same token to be in random order (or, rather, the original one). It should take the Span's length into consideration as well. Per @adrianeboyd's suggestion, I added comparison based on all `Span` attributes.

The detailed description is here:
https://github.com/explosion/spaCy/issues/9696

## Description

I made this update a while ago and somehow forgot to do a PR for it.

## Checklist


svlandeg (approved):
Thanks for another useful PR! The implementation looks good to me. While there's some repetition, I don't see how that can be avoided while keeping the same efficiency. The tests are quite elaborate as well, which is great, and I don't suspect they'd take much time.

I'll just leave this PR open for now to allow Adriane to have a look as well (she's offline this week).

adrianeboyd:
The tests are great here. I think the whole method would be easier to maintain if all the ops compared the same tuples (ordered in the same way for efficiency) for the comparisons, so something like this:

```python
        self_tuple = (self.c.start_char, self.c.end_char, self.c.label, self.c.kb_id, self.doc)
        other_tuple = (other.c.start_char, other.c.end_char, other.c.label, other.c.kb_id, other.doc)
        # <
        if op == 0:
            return self_tuple < other_tuple
        # <=
        elif op == 1:
            return self_tuple <= other_tuple
        # ==
        elif op == 2:
            return self_tuple == other_tuple
        # !=
        elif op == 3:
            return self_tuple != other_tuple
        # >
        elif op == 4:
            return self_tuple > other_tuple
        # >=
        elif op == 5:
            return self_tuple >= other_tuple
```

I gave this a try and the tests all passed. I could have sworn we were already doing this in another similar place (because tuples compare element by element it works the same as the longer fallback version), but I couldn't find it quickly.

nrodnova:
@adrianeboyd I like it, much more readable :)

adrianeboyd:
I think that this should go in `develop` (v3.3) because the behavior will change, but otherwise think this is fine to merge at this point.

adrianeboyd (approved):

svlandeg:
Thanks again for the PR @nrodnova and to @adrianeboyd for making the code so much nicer 🤩

PR #9961, opened 2021-12-30
Fix references to config file in the docs & UX

svlandeg:
## Description
There were some references to `config.cfg` in the docs & error messages that could be pretty confusing if you deliberatly named your config file differently. Also a few places where the docs or comments were just incorrect.

### Types of change
doc fixes.

## Checklist


website/docs/api/cli.md
```diff
@@ -175,7 +175,7 @@ $ python -m spacy init fill-config [base_path] [output_file] [--diff]
 | Name                   | Description                                                                                                                                                                          |
 | ---------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `base_path`            | Path to base config to fill, e.g. generated by the [quickstart widget](/usage/training#quickstart). ~~Path (positional)~~                                                            |
-| `output_file`          | Path to output `.cfg` file. If not set, the config is written to stdout so you can pipe it forward to a file. ~~Path (positional)~~                                                  |
+| `output_file`          | Path to output `.cfg` file or "-" to write to stdout so you can pipe it forward to a file. ~~Path (positional)~~                                                                     |
```
polm, line 178: Does this still write to stdout if not set?

Also I would remove "forward" here.

polm (commented):
Looks good, just had one question.

website/docs/api/cli.md
svlandeg: I've clarified that "-" is the default

polm:
Latest commit fails with mypy errors? That doesn't make any sense, I'll re-run those...

svlandeg:
The `mypy` errors should hopefully be addressed by https://github.com/explosion/spaCy/pull/9979, let me update, push and let's see what the CI says.

PR #9972, opened 2022-01-03
Iceland lang code fix

### LINKED ISSUE #8214: Broken import for Icelandic language data

It looks like importing language data for Icelandic is broken.
E.g. to get stop words:

```python
# This works
import spacy.lang.en.stop_words
spacy.lang.en.stop_words.STOP_WORDS

# Syntax error in import statement
import spacy.lang.is.stop_words
spacy.lang.is.stop_words.STOP_WORDS
```
Error:
```python
>>> import spacy.lang.is.stop_words
  File "<stdin>", line 1
    import spacy.lang.is.stop_words
                      ^
SyntaxError: invalid syntax

```
I have yet to test this on Spacy v3.0.

## Your Environment

* **spaCy version:** 2.3.5
* **Platform:** Linux-5.8.0-53-generic-x86_64-with-glibc2.29
* **Python version:** 3.8.5

DuyguA:
## Description

Icelandic lang code is `is` and it creates trouble in importing this lang and potentially in several places. This PR changes lang code to `isl`.  

Fixes #8214.

### Types of change

Quick fix 

## Checklist


polm:
I would much rather use an established three-letter code than an invented two-letter code. Worst case, the two-letter code gets a different language assigned later...

svlandeg (changes requested):
Some quick comments:
- I think using `isl` is better than `ic` as the latter is not an actual standard code. It wouldn't be the first two-char code we have, there's also `grc`, so I wouldn't worry about that.
- I want to make sure that this is bwd-compatible at least when calling `spacy.blank("is")`. It might already be thanks to `langcodes`, but we should have a unit test to double check
- This will never be fully bwd-compatible because of potentially breaking imports, so this should at least be moved to `develop`
- The tests are currently failing so I unchecked that box. It looks like some imports still need to be fixed in the core code base.
- This will need an update in the docs where we summarize all available languages and their codes.

DuyguA:
> Some quick comments:
> 
>     * I think using `isl` is better than `ic` as the latter is not an actual standard code. It wouldn't be the first two-char code we have, there's also `grc`, so I wouldn't worry about that.

Great then! I didn't notice we had a 3-letter code btw. 

> 
>     * I want to make sure that this is bwd-compatible at least when calling `spacy.blank("is")`. It might already be thanks to `langcodes`, but we should have a unit test to double check

OK, I'll add.

>     * This will never be fully bwd-compatible because of potentially breaking imports, so this should at least be moved to `develop`
I'll rebase!
 
>     * The tests are currently failing so I unchecked that box. It looks like some imports still need to be fixed in the core code base.

I missed that part :) , will fix now. 

>     * This will need an update in the docs where we summarize all available languages and their codes (I also unchecked that box).

I'll add it.

svlandeg:
> I think we don't need to add the lang codes manually, `spaCy/website/src/widgets` does it for us

But doesn't that read its information from `languages.json`?

spacy/tests/tokenizer/test_explain.py
```diff
@@ -4,7 +4,6 @@
 from spacy.tokenizer import Tokenizer
 
 # Only include languages with no external dependencies
-# "is" seems to confuse importlib, so we're also excluding it for now
 # excluded: ja, ru, th, uk, vi, zh, is
```
svlandeg, line 7:
```suggestion
# excluded: ja, ru, th, uk, vi, zh, is
```
```suggestion
# excluded: ja, ru, th, uk, vi, zh
```

DuyguA:
> > I think we don't need to add the lang codes manually, `spaCy/website/src/widgets` does it for us
> 
> But doesn't that read its information from `languages.json`?

Lemme check

spacy/tests/test_language.py
```diff
@@ -550,6 +550,7 @@ def test_spacy_blank():
         ("en", "en"),
         ("fra", "fr"),
         ("fre", "fr"),
+        ("isl", "is"),
```
svlandeg, line 553: This one needs to be reversed
```suggestion
        ("is", "isl"),
```

svlandeg:
> > > I think we don't need to add the lang codes manually, `spaCy/website/src/widgets` does it for us
> > 
> > 
> > But doesn't that read its information from `languages.json`?
> 
> Yes, I took the above lines from `languages.js` I didn't see a lang list in languages.json

`languages.json` is here - it defines the list of languages: https://github.com/explosion/spaCy/blob/master/website/meta/languages.json

DuyguA:
> > > > I think we don't need to add the lang codes manually, `spaCy/website/src/widgets` does it for us
> > > 
> > > 
> > > But doesn't that read its information from `languages.json`?
> > 
> > 
> > (...) I didn't see a lang list in languages.json
> 
> `languages.json` is here - it defines the list of languages: https://github.com/explosion/spaCy/blob/master/website/meta/languages.json

Ah OK, I pushed one more tiny commit then

svlandeg (dismissed):

svlandeg (commented):
Ok great, I think most of the changes are addressed - let's see if the CI goes green now. If it does, we can rebase to `develop` after merging https://github.com/explosion/spaCy/pull/9987.

svlandeg:
This PR is in good shape to fuel the internal discussion now. As it stands, it changes the language code `is` to `isl` throughout the code base. It is not breaking in the sense that `spacy.blank("is")` will still work, but the actual `lang` of `Islandic` has in fact changed, and there might be users checking/using this directly. Their code might still break as a result of this PR.

Alternatively to this PR, we could opt to only move the `lang.is` module to `lang.isl` and keep the actual language code of `Islandic` to `"is"`. This would require some more checks whether we can get that solution to work seemlessly, but with `langcodes` already transparently translating between `is` and `isl`, it should be doable. It would mean no breaking changes at all.

Let's keep this PR open until we've had that internal discussion.

polm:
One issue here that came to mind is that anyone using the current importlib workaround will see this as a breaking change. That seems like something people are actually doing now, so I think it makes sense to keep things stable for them. 

If we want to avoid that, we can re-create the old `lang/is` dir and use an `__init__.py` to import everything from `lang/isl`. Note that `spacy.blank` will load from `lang/is` in that case, so we should use the registry to register the `isl` version to avoid that. If we do this, I think we should put a deprecation warning in the `lang/is` version to suggest people stop using `importlib` or other methods that still manage to load it through that path.

Regarding the actual language code in the class, I am torn. It seems better to keep it unchanged, but keeping it different from the canonical class name seems like a source of future errors that we could avoid by changing it now. Given that it wasn't really working it seems pretty easy to change it now, but that'll only get harder in the future.

DuyguA:
> One issue here that came to mind is that anyone using the current importlib workaround will see this as a breaking change. That seems like something people are actually doing now, so I think it makes sense to keep things stable for them.

Yes, of course :blush: 

> If we want to avoid that, we can re-create the old `lang/is` dir and use an `__init__.py` to import everything from `lang/isl`. Note that `spacy.blank` will load from `lang/is` in that case, so we should use the registry to register the `isl` version to avoid that. If we do this, I think we should put a deprecation warning in the `lang/is` version to suggest people stop using `importlib` or other methods that still manage to load it through that path.

I can change the code a bit

> Regarding the actual language code in the class, I am torn. It seems better to keep it unchanged, but keeping it different from the canonical class name seems like a source of future errors that we could avoid by changing it now. Given that it wasn't really working it seems pretty easy to change it now, but that'll only get harder in the future.

polm:
I don't think there's any direct harm from maintaining an alias. I think it'd be safer to avoid it though for a couple of reasons:

1. having multiple ways to do things without a clear explanation can be confusing (users might not find wherever we document that)
2. I assume we would not use the `importlib` trick ourselves, and it may be hard to test, so we're less likely to notice if problems with it arise later.

Either way is fine though.

adrianeboyd:
Related, there's the issue of what we should do with `xx` (maybe in v4), which should really be `mul`, but some kind of longer-term alias might be useful.

polm:
I took a look at the code needed to maintain an alias and I think I found another issue. I think that if we add extra files in the future, like for syntax iterators or something, we'll have to add equivalent files in the alias directory in order to have it work equivalently. There are ways to avoid that but as far as I can tell it's stuff like messing with `sys.modules` that seems like a really bad idea.

polm:
After testing I'm pretty sure I confirmed my concerns above. Basically we can't get equivalent behavior between `is` and `isl` unless we make sure to always add a dummy file in `is` whenever a real file is added in `isl`. I don't think we want to maintain that, it seems like it would be too easy to forget.

I made a small commit that adds the dummy files to get equivalent behavior now as an example, though I don't like it. I also added a DeprecationWarning as an example of how we would add that (the message would need to be improved). 

One thing that surprised me about DeprecationWarnings is that it was harder to trigger them than expected. [The official docs](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) indicate that the warnings are ignored by default, but triggered if called in `__main__`. But I didn't find that to be the case, though warnings worked normally with `-Wall`.

DuyguA:
> After testing I'm pretty sure I confirmed my concerns above. Basically we can't get equivalent behavior between `is` and `isl` unless we make sure to always add a dummy file in `is` whenever a real file is added in `isl`. I don't think we want to maintain that, it seems like it would be too easy to forget.
> 
> I made a small commit that adds the dummy files to get equivalent behavior now as an example, though I don't like it. I also added a DeprecationWarning as an example of how we would add that (the message would need to be improved).
> 
> One thing that surprised me about DeprecationWarnings is that it was harder to trigger them than expected. [The official docs](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) indicate that the warnings are ignored by default, but triggered if called in `__main__`. But I didn't find that to be the case, though warnings worked normally with `-Wall`.
> 
> My suggestion would be to add the files like in my last commit to avoid breaking things for anyone using the workaround immediately while directing them to use the version of the import that doesn't require a workaround.

It's up to you surely. I didn't like the dummy files as well, though they're not so horrible it's mergeable for me.

adrianeboyd:
My current feeling is that we should wait until v4 and just rename it without trying to maintain anything in parallel, which seems like it has too many potential catches.

svlandeg:
> My current feeling is that we should wait until v4 and just rename it without trying to maintain anything in parallel, which seems like it has too many potential catches.

I agree. We've been discussing this at length internally and there seem to be too many potential risks. I'll go ahead and close this and we'll pick it up for v4.

PR #9979, opened 2022-01-04
Fix type of lexeme.rank

svlandeg:
## Description
On recent PRs, the CI is failing over a `lexeme.rank` assignement. I actually think the CI is right, and that the `pyi` file is wrong, though I don't understand why we didn't see this error before and why it only occurs in older Python versions.

Anyway, let's see if this fix makes the CI happy, because I can't reproduce the original error locally.

### Types of change
bug fix

## Checklist


svlandeg:
Looks like it did - I'll go ahead and merge this in so we can continue with the other PRs

Issue #9904, opened 2021-12-19
Textcat loss is scaled by batch size (and number of classes)


jsnfly:
## How to reproduce the behaviour
In [TextCategorizer#get_loss](https://github.com/explosion/spaCy/blob/93e9bf681fd6fbdd3a34e2c5df6b3a7c93b9596e/spacy/pipeline/textcat.py#L302) the scores are first [divided by the batch size](https://github.com/explosion/spaCy/blob/93e9bf681fd6fbdd3a34e2c5df6b3a7c93b9596e/spacy/pipeline/textcat.py#L316) and then the [mean is taken](https://github.com/explosion/spaCy/blob/93e9bf681fd6fbdd3a34e2c5df6b3a7c93b9596e/spacy/pipeline/textcat.py#L318). So the resulting loss scales (quadratically!) with the inverse of the batch size. Is this a bug?
Example to reproduce:
```python
from spacy.lang.en import English
from spacy.tests.pipeline.test_textcat import make_get_examples_single_label

nlp = English()
nlp.add_pipe("textcat")
get_examples = make_get_examples_single_label(nlp)
nlp.initialize(get_examples)
textcat = nlp.get_pipe("textcat")

examples = get_examples()
scores = textcat.predict([eg.predicted for eg in examples])
loss = textcat.get_loss(examples, scores)[0]
print(f'Textcat loss: {loss}')

examples_double_bs = examples * 2
scores_double_bs = scores.repeat(2, axis=0)
loss_double_bs = textcat.get_loss(examples_double_bs, scores_double_bs)[0]
print(f'Textcat loss with doubled batch size: {loss_double_bs}')
```
This should be an easy fix, by just omitting the first division or is there something I'm missing?

Also, because the [sum is taken before applying the mean](https://github.com/explosion/spaCy/blob/93e9bf681fd6fbdd3a34e2c5df6b3a7c93b9596e/spacy/pipeline/textcat.py#L318), it means that the loss scales linearly with the number of classes, right? But I'm not sure if this is intentional?

## Your Environment
- **spaCy version:** 3.2.1
- **Platform:** Linux-5.11.0-41-generic-x86_64-with-glibc2.33
- **Python version:** 3.9.7

honnibal:
[...]

PR #10002, opened 2022-01-07
Fix texcat loss scaling (#9904)

jsnfly:
This PR fixes issue #9904 by removing the division by the batch size and the summation accross classes before the mean is applied in `TextCategorizer#get_loss`.

## Description
#9904 

### Types of change
enhancement/bug fix

## Checklist


svlandeg (approved):
Hi,

Thanks so much for flagging this issue and for creating a PR to resolve it! I also really like the elegant unit test :-) 

This looks good to me. I'll just wait a bit to see whether any of my colleagues have comments, before merging this in.

polm (approved):
Thanks, this looks great, and the test makes it very clear what the problem was!

Issue #9858, opened 2021-12-14
5-6x slower performance between v2 and v3


ofirnk:
## Regression
With Spacy 2 we've experienced 5-6 times faster processing of documents, with same model "size" (not transformer).

#### With `nlp(text)`
* In Spacy 2: `For text of length 789417 (~780KB) it takes ~56 seconds for full parse`
With pipeline = ['tagger', 'parser', 'ner']
* In Spacy 3: `For text of length 789417 (~780KB) it takes ~303 seconds for full parse`
With pipeline = ['tok2vec', 'tagger', 'parser', 'attribute_ruler', 'lemmatizer', 'ner']

#### With `nlp.pipe([chunks)]` and 16 chunks
* In Spacy 2: `For text of length 789417 (~780KB) it takes ~38 seconds for full parse`
* In Spacy 3: `For text of length 789417 (~780KB) it takes ~203 seconds for full parse`

## How to reproduce the behaviour

Created a Colab with dropdown to choose Spacy <3 or >3
(Please choose "restart runtime and run" after switching between spacy versions)
https://colab.research.google.com/drive/10XUpSQ6ir5-DttWJsUN5SjjVQfCDmglN?usp=sharing

I've plot the analyze time as relative to text length..

![image](https://user-images.githubusercontent.com/8449713/146670941-b43f0ff6-3415-4ae0-9d2c-12e823f09a39.png)
(ratio is attached to right axis, so it goes 2->3->4->5+)

```python
#@title Spacy 3 vs 2: benchmark large text

#@markdown **1. Download a rather large document**

import urllib
import urllib.request
large_text_file_url = "https://www.gutenberg.org/files/1342/1342-0.txt"

urllib.request.urlretrieve(large_text_file_url, "large.txt")  # download
large_txt = urllib.request.urlopen(large_text_file_url).read().decode('utf-8')

print(f"Length of large_txt={len(large_txt)}")
text_to_parse_dropdown = "large" #@param ["large", "small"]
text_to_parse = large_txt if text_to_parse_dropdown == "large" else "Hello"
print(f"Length of text_to_parse={len(text_to_parse)}")

#@markdown **2. Install spacy, download model, import & load model**
spacy_version = "spacy\u003C3" #@param ["spacy\u003C3", "spacy>3"]
import os; os.environ["spacy_version"] = spacy_version
!pip install "$spacy_version"
!python -m spacy download en_core_web_md
import en_core_web_md
nlp = en_core_web_md.load()

[...]

PR #10019, opened 2022-01-10
Speed up the StateC::L feature function

danieldk:
## Description

`StateC::L` gets the n-th most-recent left-arc with a particular head.

Before this change, `StateC::L` would construct a vector of all left-arcs
with the given head and then pick the n-th most recent from that vector.
Since the number of left-arcs strongly correlates with the doc length
and the feature is constructed for every transition, this can make
transition-parsing quadratic.

With this change `StateC::L`:

- Searches left-arcs backwards.
- Stops early when the n-th matching transition is found.
- Does not construct a vector (reducing memory pressure).

This change doesn't avoid the linear search when the transition that is
queried does not occur in the left-arcs. Regardless, performance
improves quite a bit with very long docs (benchmarked by repeating
a Wikipedia paragraph N times):

Before:

```
   N  Time

 400   3.3
 800   5.4
1600  11.6
3200  30.7
```

After:

```
   N  Time

 400   3.2
 800   5.0
1600   9.5
3200  23.2
```

We can probably do better with more tailored data structures, but I
first wanted to make a low-impact PR.

Found while investigating #9858.

### Types of change

Performance improvement.

## Checklist

### DIFF
--- a/spacy/pipeline/_parser_internals/_state.pxd
+++ b/spacy/pipeline/_parser_internals/_state.pxd
@@ -188,20 +188,17 @@ cdef cppclass StateC:
 
         # Work backwards through left-arcs to find the arc at the
         # requested index more quickly.
-        cdef int child = -1
         cdef size_t child_index = 0
         it = this._left_arcs.const_rbegin()
-        while it != this._left_arcs.rend() and child_index != idx:
+        while it != this._left_arcs.rend():
             arc = deref(it)
             if arc.head == head and arc.child != -1 and arc.child < head:
-                child = arc.child
                 child_index += 1
+                if child_index == idx:
+                    return arc.child
             incr(it)
 
-        if child_index == idx:
-            return child
-        else:
-            return -1
+        return -1
 
     int R(int head, int idx) nogil const:
         if idx < 1 or this._right_arcs.size() == 0:
--- a/spacy/pipeline/_parser_internals/_state.pxd
+++ b/spacy/pipeline/_parser_internals/_state.pxd
@@ -1,3 +1,4 @@
+from cython.operator cimport dereference as deref, preincrement as incr
 from libc.string cimport memcpy, memset
 from libc.stdlib cimport calloc, free
 from libc.stdint cimport uint32_t, uint64_t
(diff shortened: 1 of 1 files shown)


ofirnk:
Looks sound to me.
May I suggest continuing the N in the benchmarks a bit further/more resolutions? 
So it could be apparent if we're really linear here after the fix

danieldk:
> Looks sound to me.
> May I suggest continuing the N in the benchmarks a bit further/more resolutions?
> So it could be apparent if we're really linear here after the fix

I'll check how far I can push it, but with `N = 6400` I went over the maximum doc size.

honnibal:
Yep looks solid 👍

honnibal (approved):

danieldk:
Forgot to add:

> So it could be apparent if we're really linear here after the fix

It should not yet be linear in this aspect, because this doesn't improve the search when there are not enough left-arcs with the given head. But I wanted to make a PR that doesn't change any data structures first, that can be merged safely in the `main` branch.

I'll do a PR that improves this for all cases (but probably requires some data structure changes next).

ofirnk:
> Forgot to add:
> 
> > So it could be apparent if we're really linear here after the fix
> 
> It should not yet be linear in this aspect, because this doesn't improve the search when there are not enough left-arcs with the given head. But I wanted to make a PR that doesn't change any data structures first, that can be merged safely in the `main` branch.
> 
> I'll do a PR that improves this for all cases (but probably requires some data structure changes) next.

Yes, I meant that the fix actually linearizes what it should.

In more general note - maybe a long document benchmark should be introduced to the test suite?
Something in similar form to the original issue I posted. It should add 3-5 minutes to the test suite, not more.
( I took a real document and truncated it. Not multiplied a paragraph)

We can move discussion to the original issue or any other place suitable, of course.

And I appreciate a lot the effort here 👏

danieldk:
> Yes, I meant that the fix actually linearizes what it should.

Unfortunately, it doesn't because there are many heads that do not have the given index.

At any rate, I have made a refactor of the state representation that does scale linearly (also with much larger sizes). I have to do some more cleanup, but will probably do a PR tomorrow or Friday.

> In more general note - maybe a long document benchmark should be introduced to the test suite?

It would definitely nice to automatically detect such regressions.

ofirnk:
Just saw the release, thanks!

visaals:
@danieldk What was the size in bytes of the Wikipedia paragraph used in this benchmark (`N=1`)? I'm trying to figure out the optimal size to send to `nlp.pipe([chunks])`, thank you!

danieldk:
This is the test text for N=1:

```Ursa Minor (Little Bear) is a constellation in the northern sky. In North America, it is also known as the Little Dipper since its seven main stars appear to form a ladle shape (diagram pictured).```

(From Wikipedia)

PR #10026, opened 2022-01-11
Check for assets with size of 0 bytes

ryndaniels:
## Description
I ran into an issue where, if downloading an asset (without a checksum) failed in a way that left a 0-byte file on disk, the next run of `spacy project assets` would assume the asset doesn't need to be redownloaded. 0-byte files are probably never what we want, so if that's the case, let's delete it so it will reattempt the download (and warn the user):

Screen Shot 2022-01-11 at 15 42 13

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/cli/project/assets.py
+++ b/spacy/cli/project/assets.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, Optional
 from pathlib import Path
 from wasabi import msg
+import os
 import re
 import shutil
 import requests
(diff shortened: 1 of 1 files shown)


honnibal:
Ah I've wanted this check! Great idea thanks.

spacy/cli/project/assets.py
```diff
@@ -129,10 +130,17 @@ def fetch_asset(
...
-            msg.good(f"Skipping download with matching checksum: {dest}")
+        if checksum:
+            if checksum == get_checksum(dest_path):
+                msg.good(f"Skipping download with matching checksum: {dest}")
+                return
+        else:
+            # If there's not a checksum, make sure the file is a possibly valid size
+            if os.path.getsize(dest_path) == 0:
+                os.remove(dest_path)
+                msg.warn(f"Asset exists but with size of 0 bytes, deleting: {dest}")
```
svlandeg, line 143: Perhaps switch the two statements? Then in case if anything goes wrong with the `remove` function, at least the user knows what was going on.
```suggestion
                msg.warn(f"Asset exists but with size of 0 bytes, deleting: {dest}")
                os.remove(dest_path)
```

svlandeg (commented):
Great idea, thanks Ryn!

svlandeg (approved):

PR #10043, opened 2022-01-12
Update usage example url for GPE span labeling

ColleterVi:
Url extension in https://spacy.io/usage/processing-pipelines#component-example3 is no longer available
Replacing https://restcountries.eu/rest/v2/all for https://restcountries.com/v2/all

## Description
Only changing an URL. No further test is required.

### Types of change
Changed the script within documentation

## Checklist
- [] I ran the tests, and all new and existing tests passed.


polm:
Thanks for catching that!

PR #10048, opened 2022-01-13
Use constant-time head lookups in StateC::{L,R}

danieldk:
## Description

This PR changes the type of left/right-arc collections from `vector[ArcC]` to `unordered_map[int, vector[Arc]]`, so that the arcs are keyed by the head. This allows us to find all the left/right arcs for a particular head in constant time in `StateC::{L,R}`.

Benchmarks with long docs (N is the number of text repetitions):

Before (using #10019):

```
    N  Time (s)

  400   3.2
  800   5.0
 1600   9.5
 3200  23.2
 6400  66.8
12800  220.0
```

After (this commit):

```
   N   Time (s)

  400   3.1
  800   4.3
 1600   6.7
 3200  12.0
 6400  22.0
12800  42.0
```

Related to #9858 and #10019.

### Types of change

Performance improvement.

## Checklist

### DIFF
--- a/spacy/pipeline/_parser_internals/_state.pxd
+++ b/spacy/pipeline/_parser_internals/_state.pxd
@@ -3,6 +3,7 @@ from libc.string cimport memcpy, memset
 from libc.stdlib cimport calloc, free
 from libc.stdint cimport uint32_t, uint64_t
 cimport libcpp
+from libcpp.unordered_map cimport unordered_map
 from libcpp.vector cimport vector
 from libcpp.set cimport set
 from cpython.exc cimport PyErr_CheckSignals, PyErr_SetFromErrno
@@ -30,8 +31,8 @@ cdef cppclass StateC:
     vector[int] _stack
     vector[int] _rebuffer
     vector[SpanC] _ents
-    vector[ArcC] _left_arcs
-    vector[ArcC] _right_arcs
+    unordered_map[int, vector[ArcC]] _left_arcs
+    unordered_map[int, vector[ArcC]] _right_arcs
     vector[libcpp.bool] _unshiftable
     set[int] _sent_starts
     TokenC _empty_token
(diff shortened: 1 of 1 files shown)


honnibal:
Is the memory usage and copy speed still okay? They're less important but they do become relevant for the beam.

danieldk:
> Is the memory usage and copy speed still okay? They're less important but they do become relevant for the beam.

I'll do a heap profile to look at memory usage. Any ideas how to profile the copy speed?

honnibal:
I think you can just use the `StateC.copy()` method -- that's what's used in the beam code

honnibal:
Btw can you also do a quick parser benchmark just to make sure we don't have some unexpected thing (like I don't know, we create lots of state objects because we expect that to be really cheap, and now there's some overhead or something)

I expect this will all be fine, but it's probably good to check.

danieldk:
Best-of-three runs using `en_core_web_lg`:

Before this PR: 22092
With this PR: 23345

So, even seems to be a bit faster on small docs.

danieldk:
Looked again at the profile and `StatC::L`/`R` did indeed take up quite a bit of parsing time before this PR, even with small docs (10 sentences per doc).

Screen Shot 2022-01-13 at 14 06 08

ofirnk:
@danieldk I'm running it on the [very messy] colab I've shared earlier. I'll conclude and share graphs a bit later!

ofirnk:
Ok, so analysis about the amazing progress so far:

1. Before your work, the ratio between v3 and v2 was increasing over document length, reaching 5-6x performance degradation.
2. After first PR - Performance ratio *decreased* to 3-4x :clap: 
3. Including *this* PR - performance is with **constant** degradation of ~30% :1st_place_medal: 

So to summarize - I no longer can see quadratic performance - AMAZING work! 

image

Explanation:
1. Yellow squares is the ratio of v3.2 vs v2 [goes from ~1 towards 5-6 (and probably beyond) ]
2. Red line is analysis time in v2, and reddis-triangles is the analysis time with this PR :clap: 
3. Blue triangles is the ratio of this PR versus v2 - which proves we're back to LINEAR analysis time

danieldk:
Thanks a lot for the analysis 👍🏻, this is really useful!

ofirnk:
> Thanks a lot for the awesome analysis 👍🏻, this is really useful!

With pleasure, the least I can do for this great open source project 🥇

BTW, this can serve as the regression test for future releases.
Speaking of which - is there an ETA for next release with this [rather critical] fix?

honnibal (approved):

svlandeg:
> With pleasure, the least I can do for this great open source project 🥇

🙏 

> BTW, this can serve as the regression test for future releases. Speaking of which - is there an ETA for next release with this [rather critical] fix?

This fix will be released as part of spaCy 3.3.0. We don't have a scheduled release date for that yet, but it likely won't take months ;-) Keep an eye out on the release page!

PR #10051, opened 2022-01-13
Bugfix for similarity return types

richardpaulhudson:
## Description
In the past, calls to `Lexeme.similarity()`, `Span.similarity()` and `Token.similarity()` have returned inconsistent return types: depending on whether or not matrix multiplication was performed to determine the similarity and whether or not Numpy (CPU) or Cupy (GPU) was in use, the return value could be a native Python float, a Numpy float or a Cupy array. This pull request duplicates the pre-existing behaviour of `Doc.similarity()` and ensures that all return values are native floats.

### Types of change
Bug fix

## Checklist

### DIFF
--- a/spacy/lexeme.pyx
+++ b/spacy/lexeme.pyx
@@ -130,8 +130,10 @@ cdef class Lexeme:
             return 0.0
         vector = self.vector
         xp = get_array_module(vector)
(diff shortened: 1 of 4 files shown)


adrianeboyd:

svlandeg (approved):
Good catch! And it looks like the test suite is also happy on GPU so we can go ahead and merge this :-)

Issue #10042, opened 2022-01-12
displacy DEFAULT_LABEL_COLORS schema mismatch en_core_web_trf


davidberenstein1957:
## How to reproduce the behaviour
from spacy.displacy.render import DEFAULT_ENTITY_COLOR, DEFAULT_LABEL_COLORS}
does not align with 
https://spacy.io/models/en#en_core_web_trf

## Your Environment
N.A
* Operating System:
* Python Version Used: 3.8.10
* spaCy Version Used: 3.2.0
* Environment Information: Windows 11

polm:
What do you mean exactly? 

Looking at the list of entities there and the list in the displacy colors, I see that the model has `FAC` and displacy has `FACILITY`. But that's the same for all the English models.

PR #10067, opened 2022-01-16
Rename FACILITY to FAC in color list

polm:
displaCy's color list includes FACILITY instead of FAC, but the English models use FAC. (Based on the presence of FACILITY in some places maybe it used to be the other way?) 

This change makes the colors match the model.

I noticed this because of #10042, which I think is reporting this issue.

### Types of change

bug fix

## Checklist


svlandeg (approved):
Good catch! Discussed with Ines as well - this should be fine to merge.

PR #10069, opened 2022-01-16
Update pydantic version and unpin mypy version

varunsurapaneni:
## Description

#9873 pinned the `mypy` version because `pydantic` was not yet compatible with `mypy` versions > 0.910.

pydantic v1.9.0 was released and includes the following release note:
> Extend pydantic's mypy plugin to support mypy versions 0.910, 0.920, 0.921 & 0.930

This updates `requirements.txt` to unpin the `mypy` version and update the `pydantic` version to allow for version 1.9.0.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ pathy>=0.3.5
 numpy>=1.15.0
 requests>=2.13.0,<3.0.0
 tqdm>=4.38.0,<5.0.0
-pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0
+pydantic>=1.9.0,<1.10.0
 jinja2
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Hi, we added this pin on purpose because there are some mypy failures with the newest versions. I looked into it briefly and I think that both `thinc` and `spacy` will need to be updated. It's on our to-do list, but not with an extremely high priority. If it's something you'd like to look into, PRs would be welcome!

varunsurapaneni:
> 

Thank you for the information (and for being so polite). I apologize for the noise, as I meant to open this PR as a `draft` but forgot to mark it as such.

Issue #9794, opened 2021-12-02
Depending on a model's name in config `spacy debug data` behaves differently


tomateit:
## How to reproduce the behavior
I have two identical config files with `transformer+ner` pipelines. If my NER component (not custom, absolutely tutorial rip-off) is called `entity_recognizer` (or anything else, I checked) and not `ner`, running `spacy debug data` with respective parameters produces different results on the same data:
+ For the case of `entity_recognizer` I get :
```
========================== Named Entity Recognition ==========================
ℹ 0 label(s)
0 missing value(s) (tokens with '-' label)
✔ Good amount of examples for all labels
✔ Examples without occurrences available for all labels
✔ No entities consisting of or starting/ending with whitespace
```
+ And when I rename the component to `ner`:
```
========================== Named Entity Recognition ==========================
ℹ 3 label(s)
0 missing value(s) (tokens with '-' label)
✔ Good amount of examples for all labels
✔ Examples without occurrences available for all labels
✔ No entities consisting of or starting/ending with whitespace
```
Goddamn, I had spent hours of debugging with this debug command until I realized, that my data is OK and I do not have 0 labels. 
## Your Environment
* Operating System: Ubuntu 20.04
* Python Version Used: 3.8
* spaCy Version Used: spacy 3.2

adrianeboyd:
Thanks for the report, that does look like a bug! The messages about having a good amount of examples in both outputs should be correct, since this is looking at `doc.ents` and not the pipeline config, but there's clearly a bug in code that looks up the labels by component name.

aserputov:
Hello, @adrianeboyd @tomateit  I want to work on this issue. Could you please assign me?

adrianeboyd:
Hi, we don't usually formally assign people to issues here. You can open a PR (as a draft if it's not finalized) and link it to this issue by using the "fixes" keyword in the description, so a note will show up here automatically about the PR.

PR #10070, opened 2022-01-17
Add spancat pipeline in spacy debug data

ljvmiranda921:
This PR attempts to add a spancat component to the `spacy debug data` command

## Description

Here are the following checks:
- Check if some model labels are not present in the actual dataset
- Check for whitespace invalid entity spans
- Check for low number of examples per label
- Check if there are examples without occurrences available for all labels

I also updated `compile_gold` to add a spancat key with the following type signature:

```
# Sample look of the golden data
{ ..., "spancat": {"span_key_0": Counter(A=1,B=3,C=4)}, ...} 
```

There are some things I'm still unsure about:
1. Maybe there's a better way I should access the spans and their labels? The way I do it now is through the `reference` attribute of the `Example` object.
2. NER has a `-` placeholder for missing/unknown entities. I also made something similar to SpanCat. Users may not know that they can assign `-` for spancat?
3. I added a new condition on `_get_examples_without_labels` to specify if the pipeline component is for NER / SpanCat. Let me know if there's a better way to do this. 

### Types of change

- Implementation change
- Tests currently WIP

## Checklist


adrianeboyd:
I don't think that `spancat` has the same restriction on whitespace as `ner`. As far as I understand it shouldn't matter at all if there are whitespace tokens anywhere in the spans? And for `ner` it only matters if the whitespace tokens are the first or last (or only) token in the span. (This check would currently check if any token is whitespace, I think?)

I also didn't think that there was a supported missing label for `spancat`?

ljvmiranda921:
> I also didn't think that there was a supported missing label for spancat?

Oh for missing labels, I compared the output of `nlp.get_pipe("spancat").labels` to the ones found in the dataset itself. Or do you mean the `-`? If it's the latter, yeah, I'll think about it more. 🤔  I might remove it since spancat won't always have a `-` for "empty" labels.

> As far as I understand it shouldn't matter at all if there are whitespace tokens anywhere in the spans?

I can probably check if the labeled span only contains a whitespace?

adrianeboyd:
I don't see any reason why you shouldn't be able to have whitespace spans if you want?

I don't see any implementation for handling missing annotation in `spancat`. It might be something we want to think about before making it non-experimental, but I don't think `-` is used as a missing label currently, so this check doesn't make sense.

ljvmiranda921:
>  but I don't think - is used as a missing label currently, so this check doesn't make sense.

Gotcha will remove it!

> I don't see any reason why you shouldn't be able to have whitespace spans if you want?

Oooh I see. Hmm, I might need some help--I can't imagine a use-case where pure-whitespace spans are viable...
that's why I was thinking of adding a check. 🤔 Do they act as negative samples? Either way, sure I will remove the current whitespace implementation!

adrianeboyd:
It would be a weird use case to have whitespace spans, but there's nothing technical preventing it from working, so I don't think it needs the same kind of check as for `ner`, where the training will basically crash if you include this kind of span.

ljvmiranda921:
Gotcha, all clear and makes sense to me!

spacy/cli/debug_data.py
```diff
@@ -193,6 +194,66 @@ def debug_data(
...
+                f"Key: {span_key}, {_format_labels(data_labels.items(), counts=True)}"
+            )
+
+        missing_labels = model_labels - set(itertools.chain(*labels))
+        if missing_labels:
+            msg.warn(
+                "Some model labels are not present in the train data. The "
+                "model performance may be degraded for these labels after "
+                f"training: {_format_labels(missing_labels)}."
+            )
```
adrianeboyd, line 220: I think it would make sense to check this per `spans_key`.

Actually, I think for all these checks it would make sense to run all the checks per `spancat`+`spans_key` as provided in the config, since the main concern for `debug data` should be the spans that are being used for training in this configuration.

This makes the checks a bit different from `ner`, where there's only one possible place to store training data for an `ner` component.

spacy/cli/debug_data.py
```diff
@@ -193,6 +194,66 @@ def debug_data(
...
+        has_low_data_warning = False
+        has_no_neg_warning = False
+        has_ws_ents_error = False
+
+        msg.divider("Span Categorization")
+        msg.info(f"{len(model_labels)} label(s)")
+
+        # Check if some model labels are not present in actual dataset
+        msg.text("Labels in train data: ")
+        for span_key, data_labels in gold_train_data["spancat"].items():
```
adrianeboyd, line 209: Can you use `spans_key` instead of `span_key` to be consistent with `spancat`?

spacy/cli/debug_data.py
```diff
@@ -193,6 +194,66 @@ def debug_data(
...
+                f"least {NEW_LABEL_THRESHOLD} instances of the new label",
+                show=verbose,
+            )
+        if has_no_neg_warning:
+            msg.text(
+                "Training data should always include examples of entities "
+                "in context, as well as examples without a given entity "
+                "type.",
+                show=verbose,
+            )
```
adrianeboyd, line 255: Can you pair these with `if`/`else` for each type of warning?

adrianeboyd:
And let me have a look at #9794 because I think that will be relevant here.

spacy/cli/debug_data.py
ljvmiranda921: Updated in https://github.com/explosion/spaCy/pull/10070/commits/8e2d2e209f8e1619e1fbdcfed6f200ef919fdb1f

spacy/cli/debug_data.py
```diff
@@ -686,14 +757,23 @@ def _format_labels(
     return ", ".join([f"'{l}'" for l in cast(Iterable[str], labels)])
 
 
-def _get_examples_without_label(data: Sequence[Example], label: str) -> int:
+def _get_examples_without_label(
+    data: Sequence[Example], label: str, pipeline: str = "ner"
```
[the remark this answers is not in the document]
ljvmiranda921: Updated in https://github.com/explosion/spaCy/pull/10070/commits/9ed1d146d0688bb00b08b33d8bffb210cd231809

spacy/cli/debug_data.py
ljvmiranda921: Ahh gotcha, will update this!

ljvmiranda921: Hmm, this might be a bit bigger than I imagined. I don't think the `Language.labels` attribute for SpanCat [differentiates between span keys](https://github.com/explosion/spaCy/blob/master/spacy/pipeline/spancat.py#L242) (cf. [how `add_label` is called in SpanCat](https://github.com/explosion/spaCy/blob/master/spacy/pipeline/spancat.py#L408)). When calling `pipe.cfg["labels"]` it will always return a "flattened" list irrespective of the span_key.

It is possible to change `pipe.cfg["labels"]` into a dictionary (i.e., `Dict[str, List[str]]` where key is `spans_key` and value is the corresponding labels) but it might affect/break a few things down the line. 

I can probably do a "hack" where:

```python
# just some sample
model_labels = _get_labels_from_model(nlp, "spancat")  # a flat list irrespective of spans key
for spans_key, labels in data_labels:
    for label in labels:
        if label not in model_labels:
            msg.warn("No {label} found in model labels for key {spans_key}")  # will fix the messaging here
```

I will commit an initial implementation first :)

ljvmiranda921: Updated in https://github.com/explosion/spaCy/pull/10070/commits/78c199cb5b3e6d91287adc5352dc2c953b1e042d

spacy/cli/debug_data.py
ljvmiranda921: Updated in https://github.com/explosion/spaCy/pull/10070/commits/2a45f7191a03ddbf357652e94f390fb992a95e6e

spacy/cli/debug_data.py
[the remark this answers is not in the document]
adrianeboyd: Nevermind about this, see #10079 which will use `factory_name`.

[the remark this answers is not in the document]
adrianeboyd: Ah, since `spancat` wasn't added yet I didn't look at this in #10079, but I think it would also need to use this as a factory name rather than a component name and have a separate implementation for `spancat` that includes the `spans_key`.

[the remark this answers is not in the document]
ljvmiranda921: Updated in https://github.com/explosion/spaCy/pull/10070/commits/3753a8b897a41d4942933c2885562df760188bcd
Not sure what's the recommended way to catch instances of `KeyError`... I was thinking of something like:

```python
try:
    ....
    eg.reference.spans[spans_key]
except KeyError:
   raise
```

But I don't see this pattern in the module itself.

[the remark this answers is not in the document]
adrianeboyd: The reason you don't see this pattern anywhere is that we actively avoid it as part of our code style: https://github.com/explosion/spaCy/blob/master/extra/DEVELOPER_DOCS/Code%20Conventions.md#dont-use-tryexcept-for-control-flow

I'm not sure exactly what you're trying to do in this example, but I think you could just use `if spans_key in eg.reference.spans` instead.

I don't think anything in `debug data` should be actually raising a `KeyError` as part of the checks.

adrianeboyd:
There are going to be conflicts with this vs. #10079, so let's see if we can get that one merged and then continue working on this.

ljvmiranda921:
Noted on the conflict! I'll just be on standby 👍

spacy/cli/debug_data.py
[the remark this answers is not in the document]
ljvmiranda921: Ah I see,  updated PR to include the check!

adrianeboyd:
Okay, now I think you can merge in `master` to continue working (dealing with the merge conflicts: at least initially, please prefer the updated versions of the labels methods from #10079).

I think the key here is that we mainly want to debug the data for a particular config, so we don't need to look at all `doc.spans` entries, we just need to look at the spans keys that are used in any `spancat` components, which may or may not be named `spancat` and there may be more than one.

ljvmiranda921:
Updated!

spacy/cli/debug_data.py
```diff
@@ -193,6 +194,65 @@ def debug_data(
...
+        else:
+            msg.good("Good amount of examples for all labels")
+
+        if has_no_neg_warning:
+            msg.text(
+                "Training data should always include examples of entities "
+                "in context, as well as examples without a given entity "
+                "type.",
+                show=verbose,
+            )
```
adrianeboyd, line 252: The spans aren't necessarily entities so I think these error messages should be updated to say something like "span type" (not sure of exact/best wording).

spacy/cli/debug_data.py
```diff
@@ -193,6 +194,65 @@ def debug_data(
...
+        msg.table(model_labels_spancat, header=["Spans Key", "Labels"], divider=True)
+
+        msg.text("Labels in train data: ")
+        for spans_key, data_labels in gold_train_data["spancat"].items():
+            msg.text(
+                f"Key: {spans_key}, {_format_labels(data_labels.items(), counts=True)}"
+            )
+        # Actual data checks
+        for spans_key, data_labels in gold_train_data["spancat"].items():
+            for label, count in data_labels.items():
```
adrianeboyd, line 212: Would it make sense to do this just for the spans keys in actual spancat components? I think we risk showing warnings related to irrelevant spans keys in the reference data otherwise, for instance, the "low number of examples" warning would be shown for all spans keys, I think?

spacy/cli/debug_data.py
```diff
@@ -686,14 +756,28 @@ def _format_labels(
     return ", ".join([f"'{l}'" for l in cast(Iterable[str], labels)])
 
 
-def _get_examples_without_label(data: Sequence[Example], label: str) -> int:
+def _get_examples_without_label(
+    data: Sequence[Example],
+    label: str,
+    component: str = "ner",
```
adrianeboyd, line 762: If you don't provide "ner" or "spancat" this method is going to crash, I think. I think `Literal` types would be enough to make it clear, but there could be other alternatives / error handling.

spacy/cli/debug_data.py
ljvmiranda921: Span type may be a good term so that it's easier for users to find the analogy between entity types. I updated the wording in this commit: https://github.com/explosion/spaCy/pull/10070/commits/595a87214aa64ce1c8b100046821b8fe206f0817

spacy/cli/debug_data.py
ljvmiranda921: Updated to `Literal` type in this commit: https://github.com/explosion/spaCy/pull/10070/commits/0c289b3faca192c604fd403b50c4ebca3534d49a

I was tempted to do the following:

```python
count = 0
if component not in ["ner", "spancat"]:
    return count
else:
   for eg in data:
       ...
```

But the semantics of returning a count of zero and silently passing an "error" may not be ideal. So I opted for just typing it as a `Literal`.

spacy/cli/debug_data.py
ljvmiranda921: Gotcha, I filtered the `gold_train_data["spancat"]` with the relevant `spans_key` from `_get_labels_from_spancat` here: https://github.com/explosion/spaCy/pull/10070/commits/21c6351c63702a7a07afc51fb2b6f3e0c70ea412 (stored in the `model_labels_spancat` variable).

spacy/cli/debug_data.py
adrianeboyd: It wouldn't be a problem if this particular internal method threw the right kind of error, but an undefined local variable error is going to confuse someone in the future unnecessarily.

And we wouldn't want to just use `Literal` for user-facing methods, but I think this is good enough for this private method right now.

svlandeg (approved):
Adding my review here because Github keeps asking me for it ;-)

It looks good to me from a quick glance, but I trust Adriane has been looking into this in more detail so I'll leave it up to her to merge when she thinks it's ready.

Either way, it will be good to have these checks included in `debug data` !

spacy/cli/debug_data.py
```diff
@@ -6,6 +6,7 @@
 import srsly
 from wasabi import Printer, MESSAGES, msg
 import typer
+import itertools
```
adrianeboyd, line 9:
```suggestion
```

spacy/cli/debug_data.py
```diff
@@ -193,6 +193,69 @@ def debug_data(
...
+        has_no_neg_warning = False
+
+        msg.divider("Span Categorization")
+        msg.table(model_labels_spancat, header=["Spans Key", "Labels"], divider=True)
+
+        msg.text("Labels in train data: ")
+        for spans_key, data_labels in gold_train_data["spancat"].items():
+            msg.text(
+                f"Key: {spans_key}, {_format_labels(data_labels.items(), counts=True)}"
+            )
```
adrianeboyd, line 208: The label counts are only shown for other components in verbose mode, and I think that would make sense here, too. It is still a bit redundant with the table above, but I think that's not a huge deal.

```suggestion
        msg.text("Label counts in train data: ", show=verbose)
        for spans_key, data_labels in gold_train_data["spancat"].items():
            msg.text(
                f"Key: {spans_key}, {_format_labels(data_labels.items(), counts=True)}",
                show=verbose,
            )
```

ljvmiranda921: Ah ok, gotcha!  Will include this diff

PR #10073, opened 2022-01-17
Add spans to doc.to_json

thomashacker:
## Description

This `PR` adds spans to `doc.to_json`.

It is saved as a `dict` with `SpanGroup` names as keys and `list[Span]` as values.
The stored spans contain: `start`, `end`, `label`, and `kb_id`

### Types of change

New feature to existing function

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -1484,6 +1484,14 @@ cdef class Doc:
                 token_data["dep"] = token.dep_
                 token_data["head"] = token.head.i
             data["tokens"].append(token_data)
+        
+        data["spans"] = {}
+        for span_group in self.spans:
+            data["spans"][span_group] = []
+            for span in self.spans[span_group]:
+                span_data = {"start": span.start, "end": span.end, "label": span.label_, "kb_id": span.kb_id}
+                data["spans"][span_group].append(span_data)
+
         if underscore:
             data["_"] = {}
             for attr in underscore:
--- a/spacy/tests/doc/test_to_json.py
+++ b/spacy/tests/doc/test_to_json.py
@@ -1,5 +1,5 @@
 import pytest
-from spacy.tokens import Doc
+from spacy.tokens import Doc, Span
 
 
 @pytest.fixture()
@@ -60,3 +60,13 @@ def test_doc_to_json_underscore_error_serialize(doc):
     Doc.set_extension("json_test4", method=lambda doc: doc.text)
     with pytest.raises(ValueError):
         doc.to_json(underscore=["json_test4"])
+
+
+def test_doc_to_json_span(doc):
+    """Test that Doc.to_json() includes spans"""
+    doc.spans["test"] = [Span(doc, 0, 2, "test"), Span(doc, 0, 1, "test")]
+    json_doc = doc.to_json()
+    assert "spans" in json_doc
+    assert len(json_doc["spans"]) == 1
+    assert len(json_doc["spans"]["test"]) == 2
+    assert json_doc["spans"]["test"][0]["start"] == 0


adrianeboyd:
This opens a can of worms related to whatever we're calling the "JSON format" and options for serializing docs to JSON.

What are the intended uses of `Doc.to_json`?

`Doc.to_json` claims: 

> `RETURNS (dict): The data in spaCy's JSON format.`

but it's not the same format that can be loaded by `spacy.training.converters.json_to_doc` (and therefore not the format documented at https://spacy.io/api/data-formats#json-input), and this may really get confusing, especially since there's also no parallel `Doc.from_json`.

`Doc.to_json` diverges a lot from `spacy.training.gold_io.docs_to_json` and I'm trying to remember what the reason for these to be implemented separately / differently...

spacy/tokens/doc.pyx
```diff
@@ -1484,6 +1484,14 @@ cdef class Doc:
                 token_data["dep"] = token.dep_
                 token_data["head"] = token.head.i
             data["tokens"].append(token_data)
+        
+        data["spans"] = {}
```
svlandeg, line 1488: I'd add a conditional here: `if self.spans` so that `data` doesn't have an empty dict if there is no spans data - that will also keep the behaviour exactly the same for use-cases where there is no `spans` data.

spacy/tokens/doc.pyx
```diff
@@ -1484,6 +1484,14 @@ cdef class Doc:
...
+        for span_group in self.spans:
+            data["spans"][span_group] = []
+            for span in self.spans[span_group]:
+                span_data = {"start": span.start, "end": span.end, "label": span.label_, "kb_id": span.kb_id}
```
svlandeg, line 1492: It looks like the `ents` and `sents` entries actually use the `start_char` and `end_char` instead of `start` and `end` token - so perhaps it makes most sense to mimic that behaviour to be consistent?

svlandeg (commented):
While I agree with Adriane's concerns about "the JSON format", I think that's an issue we'll have to address more properly in  separate PR. Given that we currently do alredy have `doc.to_json` and that we won't be removing this in a next release, we might as well make it support spans.

Good work on also including an appropriate unit test, @thomashacker !

I had a few more small comments, cf below.

spacy/tokens/doc.pyx
```diff
@@ -1486,6 +1486,15 @@ cdef class Doc:
                 token_data["dep"] = token.dep_
                 token_data["head"] = token.head.i
             data["tokens"].append(token_data)
+        
+        if self.spans:
+            data["spans"] = {}
+            for span_group in self.spans:
+                data["spans"][span_group] = []
+                for span in self.spans[span_group]:
+                    span_data = {"start": span.start_char, "end": span.end_char, "label": span.label_, "kb_id": span.kb_id}
```
adrianeboyd, line 1495: Would `kb_id_` be better here?

svlandeg: yes!

spacy/tokens/doc.pyx
```diff
@@ -1486,6 +1486,15 @@ cdef class Doc:
                 token_data["dep"] = token.dep_
                 token_data["head"] = token.head.i
             data["tokens"].append(token_data)
+        
+        if self.spans:
+            data["spans"] = {}
+            for span_group in self.spans:
+                data["spans"][span_group] = []
+                for span in self.spans[span_group]:
+                    span_data = {"start": span.start_char, "end": span.end_char, "label": span.label_, "kb_id_": span.kb_id}
```
adrianeboyd, line 1495:
```suggestion
                    span_data = {"start": span.start_char, "end": span.end_char, "label": span.label_, "kb_id": span.kb_id}
```

thomashacker: > Would `kb_id_` be better here?


I'm a bit confused, did I change it in the wrong place? 😄

svlandeg: Yes ;-) Adriane meant to suggest to query `span.kb_id_` just like we query `span.label_` and not `span.label`, but I guess the suggestion was a little ambiguous ;-)

thomashacker: Aaahhh I didn't see that the suggested change included `span.kb_id_` so sorry! 😞

PR #10079, opened 2022-01-18
Determine labels by factory name in debug data

adrianeboyd:
## Description

For all components, return labels for all components with the
corresponding factory name rather than for only the default name.

For `spancat`, return labels as a dict keyed by `spans_key`.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/cli/debug_data.py
+++ b/spacy/cli/debug_data.py
@@ -715,16 +715,17 @@ def _get_labels_from_model(
 
 
 def _get_labels_from_spancat(
-    nlp: Language, factory_name: str
+    nlp: Language
 ) -> Dict[str, Set[str]]:
     pipe_names = [
         pipe_name
         for pipe_name in nlp.pipe_names
-        if nlp.get_pipe_meta(pipe_name).factory == factory_name
+        if nlp.get_pipe_meta(pipe_name).factory == "spancat"
     ]
     labels: Dict[str, Set[str]] = {}
     for pipe_name in pipe_names:
-        pipe = cast(SpanCategorizer, nlp.get_pipe(pipe_name))
+        pipe = nlp.get_pipe(pipe_name)
+        assert isinstance(pipe, SpanCategorizer)
         if pipe.key not in labels:
             labels[pipe.key] = set()
         labels[pipe.key].update(pipe.labels)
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I'm not entirely sure this test is worth the time it takes. Maybe it should be marked slow?

spacy/cli/debug_data.py
```diff
@@ -699,8 +699,33 @@ def _get_examples_without_label(data: Sequence[Example], label: str) -> int:
...
+    nlp: Language, factory_name: str
+) -> Dict[str, Set[str]]:
+    pipe_names = [
+        pipe_name
+        for pipe_name in nlp.pipe_names
+        if nlp.get_pipe_meta(pipe_name).factory == factory_name
+    ]
+    labels: Dict[str, Set[str]] = {}
+    for pipe_name in pipe_names:
+        pipe = cast(SpanCategorizer, nlp.get_pipe(pipe_name))
```
svlandeg, line 727: I would personally do this as an `assert`. If this depends on user input/settings I would even do an `if` check and raise a `ValueError`, but I don't think that will be the case? 

```suggestion
        pipe = nlp.get_pipe(pipe_name)
        assert isinstance(pipe, SpanCategorizer)
```

svlandeg (approved):
This is definitely an improvement. I just had one small question, but I don't feel strongly so feel free to merge either way.

svlandeg:
Does the test really take that long?

spacy/cli/debug_data.py
adrianeboyd: If mypy is fine with that, it's fine with me. Actually, there's no need for the `factory_name` arg, either...

svlandeg: Yep, `mypy` parses the `assert` and will take that into account for the following lines :-)

svlandeg: You could also have the `factory_name` set default to `"spancat"`?

adrianeboyd:
It's hard to time, but this test seems to add ~1s locally out of ~140s total. We've just had problems on and off with timeouts in CI builds (mainly conda-forge) where it might have succeeded if the test suite were faster, but it's not the only issue, and maybe not running the whole test suite here is one of the options instead. It's just a whole pipeline init x 4 just for this?

Issue #10086, opened 2022-01-18
Contractions incorrectly tokenized when part of an infix substring


johnmccain:
Contractions and other special tokenization rules are not correctly applied when they are part of an infix substring such as words separated by a hyphen or em dash.

## How to reproduce the behaviour
```
import spacy
nlp = spacy.load('en_core_web_sm')

print(list(nlp.tokenizer("don't look--don't see")))
# [do, n't, look, --, don't, see]

print(list(nlp.tokenizer.explain("don't look--don't see")))
# [('SPECIAL-1', 'do'), ('SPECIAL-2', "n't"), ('TOKEN', 'look'), ('INFIX', '--'), ('TOKEN', "don't"), ('TOKEN', 'see')]
```

## Your Environment

- **spaCy version:** 3.2.1
- **Platform:** macOS-11.6.1-x86_64-i386-64bit
- **Python version:** 3.9.9
- **Pipelines:** en_core_web_sm (3.2.0), en_core_web_lg (3.2.0)

## Cause

Based on my reading of the [`explain`](https://github.com/explosion/spaCy/blob/master/spacy/tokenizer.pyx#L613) tokenizer method, it looks like special tokenization rules are meant to be applied on a second pass after separating an infix substring into separate tokens. It does this by using a `PhraseMatcher` looking for the special tokens, but the rules for some special cases like `don't` are not being correctly added to the `PhraseMatcher`.
I believe [tokenizer.pyx:605](https://github.com/explosion/spaCy/blob/ac45ae3779f9b021ebc2227ae6b3104811132979/spacy/tokenizer.pyx#L605) is the culprit. Before adding the PhraseMatcher rule, there is a condition checking if the special case's string contains a prefix, suffix, infix, or space character. If the special case doesn't meet this condition, then the rule will never be added and we get the behavior seen above. When I removed this line and rebuilt spaCy I got the expected output.

Output after removing the condition on line 605:
```
import spacy
nlp = spacy.load('en_core_web_sm')

print(list(nlp.tokenizer("don't look--don't see")))
# [do, n't, look, --, do, n't, see]

print(list(nlp.tokenizer.explain("don't look--don't see")))
# [('SPECIAL-1', 'do'), ('SPECIAL-2', "n't"), ('TOKEN', 'look'), ('INFIX', '--'), ('SPECIAL-1', 'do'), ('SPECIAL-2', "n't"), ('TOKEN', 'see')]
```

[...]

PR #10090, opened 2022-01-19
Fix special tokenization cases not applied when adjacent to an infix #10086

johnmccain:
## Description
Fix tokenizer special cases not being applied when adjacent to an infix. Removes a condition checking that tokenization special cases contain a prefix, suffix, infix, or `" "` prior to adding the case to the tokenizer's `_special_matcher` PhraseMatcher. This condition only impacted the second pass for special tokenization rules, so it caused some special tokenization rules to be skipped when they were adjacent to an infix while they were applied correctly otherwise.

If there is a different preferred way of correcting this behavior, then please feel free to alter or reject this PR.

### Types of change
Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/tokenizer/test_tokenizer.py
+++ b/spacy/tests/tokenizer/test_tokenizer.py
@@ -13,6 +13,16 @@ from spacy.vocab import Vocab
 from spacy.symbols import ORTH
 
 
+@pytest.mark.issue(10086)
+def test_issue10086():
+    """Test special case works when part of infix substring."""
+    text = "No--don't see"
+    tokenizer = English().tokenizer
+    doc = tokenizer(text)
+    assert "n't" in [w.text for w in doc]
+    assert "do" in [w.text for w in doc]
+
+
 @pytest.mark.issue(743)
 def test_issue743():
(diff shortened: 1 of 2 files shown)


adrianeboyd:
The condition is here for efficiency, but I can see that it doesn't work as expected in this case. I suspect that there's no general solution that keeps the current efficiency and has the desired behavior.

The test failure in the CI is unexpected and weird.

I'd need to look into the details here again...

svlandeg:
Hi! 

Apologies for the delay in following up on this. We've been talking internally what the best approach would be here. Because of the performance implications Adriane mentioned, and the fact that this is probably relevant only for some edge cases, we likely wouldn't want to merge the PR as-is. However, we're considering to add a new tokenizer option (disabled by default) to enable the functionality you're suggesting.

johnmccain:
Hey @svlandeg, thanks for the update! That decision makes sense to me. Feel free to close this PR if there is another planned solution.

svlandeg:
Adriane is working on a solution that will allow this behaviour as an option: cf https://github.com/explosion/spaCy/pull/10452. I'll go ahead and close this one in the meantime.

Thanks again for flagging this!

PR #10093, opened 2022-01-19
Support version tags in universe and add note about reporting

ines:
## Description

Display the supported version of spaCy at the top of universe projects. Can be specified in the `universe.json` as either `"spacy_version": 3` or `"spacy_version": [2, 3]` etc. Also add note about reporting non-working projects at the bottom. Open to wording suggestions here.

### Version tag

Screenshot 2022-01-19 at 11 18 22

### Reporting note

Screenshot 2022-01-19 at 11 21 40

### Types of change
docs

## Checklist

### DIFF
--- a/website/src/templates/universe.js
+++ b/website/src/templates/universe.js
@@ -126,7 +126,7 @@ const UniverseContent = ({ content = [], categories, theme, pageContext, mdxComp
                     <H3>Found a mistake or something isn't working?</H3>
                     <p>
                         If you've come across a universe project that isn't working or is
-                        incompatible, let us know by{' '}
+                        incompatible with the reported spaCy version, let us know by{' '}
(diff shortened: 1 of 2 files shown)


website/src/templates/universe.js
```diff
@@ -121,6 +122,18 @@ const UniverseContent = ({ content = [], categories, theme, pageContext, mdxComp
                         </Grid>
                     </Section>
                 )}
+                <section className="search-exclude">
+                    <H3>Found a mistake or something isn't working?</H3>
+                    <p>
+                        If you've come across a universe project that isn't working or is
+                        incompatible, let us know by{' '}
```
svlandeg, line 129:
```suggestion
                        incompatible with the reported spaCy version, let us know by{' '}
```
?

ines: Ah, good point!

PR #10098, opened 2022-01-20
Docker Image for Website Dev

pmbaumgartner:
Adds a `Dockerfile` for a repeatable build of the website with a working version of node/npm and instructions for accessing the images through the github container registry for others to use.

### Types of change
Documentation

## Checklist

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -554,7 +554,7 @@ extensions for your code editor. The
 [`.prettierrc`](https://github.com/explosion/spaCy/tree/master/website/.prettierrc)
 file in the root defines the settings used in this codebase.
 
-## Building & Developing the Site with Docker {#docker}
+## Building & developing the site with Docker {#docker}
 Sometimes it's hard to get a local environment working due to rapid updates to node dependencies,
 so it may be easier to use docker for building the docs.
 
--- a/website/README.md
+++ b/website/README.md
@@ -554,6 +554,40 @@ extensions for your code editor. The
 [`.prettierrc`](https://github.com/explosion/spaCy/tree/master/website/.prettierrc)
 file in the root defines the settings used in this codebase.
 
+## Building & Developing the Site with Docker {#docker}
+Sometimes it's hard to get a local environment working due to rapid updates to node dependencies,
+so it may be easier to use docker for building the docs.
+
+If you'd like to do this,
+**be sure you do *not* include your local `node_modules` folder**,
+since there are some dependencies that need to be built for the image system. 
+Rename it before using.
+
+```bash
+docker run -it \
+  -v $(pwd):/spacy-io/website \
+  -p 8000:8000 \
+  ghcr.io/explosion/spacy-io \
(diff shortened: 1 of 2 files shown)


website/README.md
```diff
@@ -554,6 +554,40 @@ extensions for your code editor. The
 [`.prettierrc`](https://github.com/explosion/spaCy/tree/master/website/.prettierrc)
 file in the root defines the settings used in this codebase.
 
+## Building & Developing the Site with Docker {#docker}
```
svlandeg, line 557: Nitpick: I wouldn't use capitals across the title, most other sections in this readme only capitalize the first letter
```suggestion
## Building & developing the site with Docker {#docker}
```

website/README.md
```diff
@@ -554,6 +554,40 @@ extensions for your code editor. The
...
+  -p 8000:8000 \
+  ghcr.io/explosion/spacy-io \
+  gatsby develop -H 0.0.0.0
+```
+
+This will allow you to access the built website at http://0.0.0.0:8000/ 
+in your browser, and still edit code in your editor while having the site 
+reflect those changes.
+
+**Note**: On M1 Macs you may need to the image tagged `arm64` (`ghcr.io/explosion/spacy-io:arm64`),
```
svlandeg, line 578: there's a verb or something missing here?

website/README.md
```diff
@@ -554,6 +554,40 @@ extensions for your code editor. The
...
+otherwise you'll see `qemu` segfault during the build.
+
+### Building the Docker Image {#docker-build}
```
svlandeg, line 581:
```suggestion
### Building the Docker image {#docker-build}
```

svlandeg (commented):
This will be great to have, thanks so much for looking into this Peter!
I only had some small nitpicking comments :-)

website/README.md
pmbaumgartner: It makes sense to me but it came from my brain so that's not a great test :) I always favor more clarity. How about:

> If you're working on a Mac with an M1 processor, you might see segfault errors from `qemu` if you use the default image. To fix this, use the `arm64` tagged image in the `docker run` command (`ghcr.io/explosion/spacy-io:arm64`).

svlandeg: so it was missing the verb "use"? (I'm trying really hard to parse your original sentence but I really can't, haha, it may just be me)

pmbaumgartner: I'll reformulate to my new suggestion!

To me it's implied that they would "use" it in the original sentence, like if I gave someone an umbrella before they went outside I'd say "You might need this".

svlandeg: But but but it said "may need to the image", not "may need the image" 🙈 
(it's not relevant, at all, I'm just so confused)

pmbaumgartner: Well then this is definitely a case of me failing to read my own writing correctly 🙈

ines (approved):

PR #10109, opened 2022-01-21
Update typing hints

ezorita:
This PR is a follow up on #8427. Improves typing hints for `Matcher`, `PhraseMatcher`, `Language` and `Doc`, and adds new ones for `Underscore` extensions and `DependencyMatcher`.

## Description

Typing hints for the following classes have been created/updated:
### `- Matcher` and `PhraseMatcher`
Overloaded return type of `Matcher.__call__()` depending on whether `as_spans` is `True` or `False`.

### `- Language`
Fixed return type for `Language.component()` decorator.

### `- Doc`
Fixed `np.ndarray` type for `Doc.tensor`.

### `- DependencyMatcher`
Adds typing hint support for `DependencyMatcher`.

### `- Underscore`
Treats underscore extension values as `Any` types. Typing hints for underscore extensions were not set before and all accesses to `Object._.extension` were of type `Unknown`. With the current implementation, access to any extension returns an object of type `Any`, so the user can redefine the type upon retrieval, e.g.
```Python
from typing import List
from spacy.tokens import Span

if not Span.has_extension("myext"):
  Span.set_extension("myext", default=None)

def update_extension(span: Span):
  if span._.myext is None:
    span._.myext = []
  strlist: List[str] = span._.myext # This will not trigger a type check error anymore
  strlist.append("whatever")
```

  

### Types of change

Enhancement of typing hints compatibility.

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyi
+++ b/spacy/matcher/matcher.pyi
@@ -1,4 +1,16 @@
-from typing import Any, List, Dict, Literal, Tuple, Optional, Callable, Union, Iterator, Iterable, overload
+from typing import (
+    Any,
+    List,
+    Dict,
+    Tuple,
+    Optional,
+    Callable,
+    Union,
+    Iterator,
+    Iterable,
+    overload,
+)
+from ..compat import Literal
 from ..vocab import Vocab
 from ..tokens import Doc, Span
 
@@ -12,7 +24,9 @@ class Matcher:
         key: Union[str, int],
         patterns: List[List[Dict[str, Any]]],
         *,
-        on_match: Optional[Callable[[Matcher, Doc, int, List[Tuple[Any, ...]]], Any]] = ...,
+        on_match: Optional[
+            Callable[[Matcher, Doc, int, List[Tuple[Any, ...]]], Any]
+        ] = ...,
         greedy: Optional[str] = ...
     ) -> None: ...
     def remove(self, key: str) -> None: ...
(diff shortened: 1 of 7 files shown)


adrianeboyd:
Thanks for the PR! We have a fair number of users who are still using older versions of python where there aren't wheels for more recent versions of `numpy`, so we wouldn't want change the version requirements here, but all the other additions look like they're on the right track!

ezorita:
Thanks @adrianeboyd, I reverted and solved the numpy fix so that it does not require a version update. Issue is that some of my typing hint updates require `typing.Literal` which is python >= 3.8.

Switching PR as draft to decide what to do with `typing.Literal`.

svlandeg:
Hi @ezorita, thanks for your PR! Since the last PR you created, we now have `mypy` running as part of the test suite (cf https://github.com/explosion/spaCy/pull/9167) which really highlights the power of these pyi files. You'll see that some of the test suite failures are due to this `mypy` run. While some errors may be in the pyi files, there might definitely also be errors in typing in the main repo, we'll just have to dig through them :-) Let me know if you need help with this!

adrianeboyd:
For `Literal`, you can import it from `spacy.compat` to get it as either as a built-in or backport depending on the python version.

ezorita:
@svlandeg happy to hear this has been useful so far. For me it made a huge difference in debugging time to have spacy fully typed!

@adrianeboyd thanks, did not know this was available! I just pushed a fix.

spacy/matcher/matcher.pyi
```diff
@@ -1,4 +1,16 @@
...
+    List,
+    Dict,
+    Tuple,
+    Optional,
+    Callable,
+    Union,
+    Iterator,
+    Iterable,
+    overload,
+)
```
svlandeg, line 12: In these cases we prefer to make the split ourselves instead of having `black` turn this into so many separate lines:
```suggestion
from typing import Any, List, Dict, Tuple, Optional, Callable, Union
from typing import Iterator, Iterable, overload
```

svlandeg (approved):
This looks great to me, thanks again! I only had one nitpicking formatting comment :-)

PR #10131, opened 2022-01-25
Ensure that we pass a CPU array to transition_states

danieldk:
## Description

TransitionSystem::transition_states takes a typed memoryview, so we cannot pass
a CuPy array. Ensure that we are passing a CPU array.

### Types of change

bug fix

## Checklist


danieldk:
Closing. Change was already part of #10141.

PR #10141, opened 2022-01-26
Improve unseen label masking

danieldk:
## Description

Two changes to speed up masking by ~10%:

- Use a bool array rather than an array of float32.

- Let the mask indicate whether a label was seen, rather than
  unseen. The mask is most frequently used to index scores for
  seen labels. However, since the mask marked unseen labels,
  this required computing an intermittent flipped mask.

### Types of change

Performance improvement.

## Checklist

### DIFF
--- a/spacy/ml/tb_framework.py
+++ b/spacy/ml/tb_framework.py
@@ -150,7 +150,7 @@ def forward(model, docs_moves: Tuple[List[Doc], TransitionSystem], is_train: boo
     all_statevecs = []
     all_scores = []
     next_states = [s for s in states if not s.is_final()]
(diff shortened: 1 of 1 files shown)


honnibal:
Sounds good! I felt bad about my previous masking implementation.

Btw to set expectations, I don't think I can find bugs in patches like this very reliably. I assume it's tested at least as well as the previous implementation though, so I'm happy to mark this as approved :)

honnibal (approved):

danieldk:
> I assume it's tested at least as well as the previous implementation though, so I'm happy to mark this as approved :)

I tried intentionally breaking it and the tests do catch it.

PR #10146, opened 2022-01-27
Use paths.vectors for vectors in init config

adrianeboyd:
## Description

So that overriding `paths.vectors` works consistently in generated configs, set vectors model in `paths.vectors` and always refer to this path in `initialize.vectors`.

### Types of change

?

## Checklist


svlandeg:
Good point!

svlandeg (approved):
Yep, definitely makes more sense and easier to override consistently!

Issue #10135, opened 2022-01-25
Inconsistency between span.sent and span.ents when span crosses sentence boundary


nrodnova:
## How to reproduce the behaviour

When a span crosses sentence boundary, that span is not in it's own sentence entity list, which doesn't make sense,

This is because `Span.sent` returns the first sentence, if the span belongs to multiple ones. However, `Span.ents` does not include partial entities.

To make it consistent, `Span.ents` should include a partial entity at the end of the span (not at the beginning, otherwise it will be inconsistent again).

I can make the change, just let me know if you are fine with the logic. Or suggest an alternative :)

The code below demonstrates the problem:

```
import spacy
from spacy.tokens import Span

nlp = spacy.load('en_core_web_sm')
text = "First sentence. Second sentence"
doc = nlp(text)

entity = Span(doc, 2, 4, label = 'LABEL')
print("Entity:", entity)
sentence = entity.sent
print("Entity's sentence:", sentence)

print("Entity belongs to its sentence:",  entity in entity.sent.ents)
```

## Your Environment

* Operating System:
* Python Version Used:
* spaCy Version Used:
* Environment Information:

adrianeboyd:
I think our answer is going to be that the current behavior seems like the least bad option. Partial entity spans would be problematic because they often aren't entities or aren't the same type of entity. We can definitely improve the docs for `Span.ents` to clarify that it only includes the entities that fall completely within the span.

In general, spacy assumes that entities won't cross sentence boundaries. The `ner` component won't predict entities across sentence boundaries, for instance.

nrodnova:
Sorry, I think my wording was confusing. I don't mean including a part of an entity, I meant, include the whole entity that overlaps with the sentence end. This will ensure that an entity is always contained in it's own sentence. From the example above:
```
text = "First sentence. Second sentence"
entities = ["First", "sentence", ". Second", "sentence"]
sentence_1.ents will contain  ["First", "sentence", ". Second"]
sentence_2.ents will contain ["sentence"]

and assert * entity in entity.sent.ents * will always hold
```

adrianeboyd:
I think it's going to be unexpected to get spans from `Span.ents` that include tokens outside of the original span. We discussed this internally and would prefer leave the current behavior as it is. Doc clarifications should be coming soon...

PR #10154, opened 2022-01-28
Clarify Span.ents documentation

ljvmiranda921:
Ref: #10135

## Description

Retain current behaviour. Span.ents will only include entities within
said span. You can't get tokens outside of the original span.

### Types of change

Documentation update

## Checklist

### DIFF
--- a/spacy/tokens/span.pyx
+++ b/spacy/tokens/span.pyx
@@ -471,8 +471,9 @@ cdef class Span:
 
     @property
     def ents(self):
-        """The named entities in the span. Returns a tuple of named entity
-        `Span` objects, if the entity recognizer has been applied.
+        """The named entities in the span. It only includes entities that fall
+        completely within the span. Returns a tuple of named entity `Span`
+        objects, if the entity recognizer has been applied.
 
         RETURNS (tuple): Entities in the span, one `Span` per entity.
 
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Hmm, this comment has gotten a bit stale as there are more components than just the entity recognizer that can set entities. Let me think a bit about a suggestion...

Once we've decided on the wording, there's also a parallel update to make in the `Span` API docs, which should also be included as part of this PR.

spacy/tokens/span.pyx
```diff
@@ -471,8 +471,9 @@ cdef class Span:
 
     @property
     def ents(self):
-        """The named entities in the span. Returns a tuple of named entity
-        `Span` objects, if the entity recognizer has been applied.
+        """The named entities in the span. It only includes entities that fall
+        completely within the span. Returns a tuple of named entity `Span`
+        objects, if the entity recognizer has been applied.
```
adrianeboyd, line 476:
```suggestion
        """The named entities that fall completely within the span. Returns
        a tuple of `Span` objects.
```

ljvmiranda921: Ok will use this and update the corresponding API docs 👍

ljvmiranda921 (commented):
Updated both the docs in the website and the API docstrings! Let me know if I missed anything!

PR #10163, opened 2022-01-28
Write moves costs directly into numpy array

danieldk:
## Description

This avoids elementwise indexing and the allocation of an additional array.
    
Gives a ~15% speed improvement when using batch_by_sequence with size
32.

### Types of change

Performance improvement.

## Checklist


svlandeg:
Awesome!

PR #10164, opened 2022-01-28
Token sent attributes more consistent

svlandeg:
## Description
* Add `is_sent_start` and `is_sent_end` to the list of token attributes in the docs
* For consistency, allow calling `doc.has_annotation("IS_SENT_END")` --> just refer to `SENT_START` as well

### Types of change
docs 

## Checklist


adrianeboyd:
I think supporting `IS_SENT_END` in `Doc.has_annotation` is a little weird, but not enough to object. I think it would be a good idea to add tests for both to `test_has_annotation`, too.

In the docs, then `is_sent_start` is documented as both a property and as an attribute but `is_sent_end` isn't? And most properties are just documented as attributes anyway. I'm not really sure where we want to draw the line for the docs here. Maybe remove the property entry? I think users are more likely to look in the attributes table, although it's then harder to link to directly.

svlandeg:
> I think supporting `IS_SENT_END` in `Doc.has_annotation` is a little weird, but not enough to object.

I was thinking of a use-case where you'd want to do something with a given list of token attributes, but a script would call `has_annotation` on the `doc` first --> while kind of silly for `IS_SENT_END`, it doesn't feel wrong to me to support it, just in case it's useful.

I think it would be a good idea to add tests for both to `test_has_annotation`, too.
> Ah, let me look into that.

> In the docs, then `is_sent_start` is documented as both a property and as an attribute but `is_sent_end` isn't? And most properties are just documented as attributes anyway. I'm not really sure where we want to draw the line for the docs here. Maybe remove the property entry? I think users are more likely to look in the attributes table, although it's then harder to link to directly.

> Yea I'd been wondering whether we had made a conscious decision about this, or whether this just grew organically. Probably the latter. Removing the property seems the most straightforward to do, but that might break links in existing GH threads?

Also, this table has become a bit unweildy, no? There's like an implicit ordering there that I kind of understand, but I'm not sure it's helpful. I feel like we should either split it out in separate tables with clear reasons why, or just put everything alphabetically :|

svlandeg:
@adrianeboyd: I added a new test, because it was kind of difficult to incorporate it into the existing one. The existing test sets `doc[0].head` which calls `set_children_from_heads` which ends up setting sentence information on all tokens, and then the check 
```
assert not doc.has_annotation(attr, require_complete=True)
```
would fail. Additionally, setting sentence information on the first token (index 0) doesn't actually make a difference because `has_annotation` starts from the second token in the check. This is why the new test has 3 tokens.

adrianeboyd:
I couldn't find any occurrences of links to `token#is_sent_start` anywhere I looked so I think I would go ahead and remove it from docs at this point. We can always improve the attributes table more in the future.

PR #10170, opened 2022-01-31
Add whitespace and combined augmenters

adrianeboyd:
## Description

Add whitespace augmenter that inserts a single whitespace token into a doc containing the types of annotation used in core trained pipelines.

Add a combined augmenter that handles lowercasing, orth variants and whitespace augmentation.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/tests/training/test_augmenters.py
+++ b/spacy/tests/training/test_augmenters.py
@@ -1,9 +1,11 @@
 import pytest
-from spacy.training import Corpus
+from spacy.pipeline._parser_internals.nonproj import contains_cycle
+from spacy.training import Corpus, Example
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Will require #10189 to be useful for many components.

adrianeboyd:
As a note, this doesn't necessarily need to be in the core library, either. We can use this separately for the trained pipelines without adding it here. We reset most of the training-related config blocks in the trained pipelines before publishing anyway, to make it easier for users to extend them, so doing it separately shouldn't cause problems.

svlandeg (approved):
Looks good! I think it's fine to have it in the `augment` module.

PR #10175, opened 2022-01-31
Add Pipe.hide_labels to omit labels from pipeline meta

adrianeboyd:
## Description

Add `Pipe.hide_labels` property that will hide `pipe.labels` from `nlp.pipe_labels` and pipeline READMEs in cases where it does not make sense to display the labels to the user.

This is immediately applied to the `senter`, which uses internal labels for its tagger that are not directly applied to the document. It is also intended for use with the edit tree lemmatizer, where the internal labels correspond to integers that are only meaningful within the component and not when displayed to the user.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
I have no attachment to `hide_labels` as a name.

svlandeg (approved):
Looks good to me, I don't think `hide_labels` is so bad and I can't think of anything better.

PR #10185, opened 2022-02-01
Update Ukrainian tokenizer_exceptions

martinjack:
## Description
Added new short words


adrianeboyd:
Thanks for the PR! Let me switch this PR to `develop` so the change goes into v3.3.0 rather than a patch release of v3.2, since we don't modify the language defaults in patch releases.

You don't need to worry if some additional commits show up in the PR temporarily, we will straighten it out on our end before it gets merged.

PR #10187, opened 2022-02-01
YAML multiline tip for project.yml files

pmbaumgartner:
Adds in a tip about the [multi-line syntax](https://yaml-multiline.info/) for YAML in the spaCy Project documentation.

I put it below the table of `project.yml` sections since I thought it was a little awkward to present the tip right next to a the example `project.yml` file that _didn't_ do this, but I think it's fine to move it elsewhere if there's a compelling reason.

image

### Types of change
Docs

## Checklist


svlandeg (commented):
I think the multi-line syntax looks nicer in general for the description, especially in the docs where it generates a huge scrollbar to the right otherwise.

Maybe instead of having two examples, we should just incorporate the multi-line example into the original `project.yml` from the docs, and then have the quick tip next to it? It'll take up less space overall, while making the example more readible.

pmbaumgartner:
> Maybe instead of having two examples, we should just incorporate the multi-line example into the original `project.yml` from the docs, and then have the quick tip next to it? It'll take up less space overall, while making the example more readible.

Are you referencing the example above the table? I think that could look cleaner, but would require a coordinated PR to the `projects` repo since that example is loaded dynamically from there.

svlandeg:
> Are you referencing the example above the table? 

Yep!

> would require a coordinated PR

Sure, that'd be fine!

pmbaumgartner:
Converted the text part of the previous accordion component to a sidebar tip and updated the example `project.yml` to use multi-line for the description and longer commands.

PR #10192, opened 2022-02-01
Raise error in spacy package when model name is not a valid python identifier

pmbaumgartner:
This PR affects `spacy package`. It adds an error when the `--name` argument (or model name from the `meta.json`) is not a valid python identifier.

The model name needs to be a valid model identifier so that it can be included as module in the `spacy_models` entry point from the `setup.py` file.

Prior to this PR, users would get the following message if trying to package a model with an invalid name:

```
error in en_spacy-sst-cpu-efficiency-128 setup command: ("EntryPoint must be in 'name=module:attrs [extras]' format", 'en_spacy-sst-cpu-efficiency-128 = en_spacy-sst-cpu-efficiency-128')
```

## Description

Example output:

```
$ python -m spacy package model-best output/ --name CAPS-HYPHEN-NAME
```
```
ℹ Building package artifacts: sdist

✘ Model name ('CAPS-HYPHEN-NAME') is not a valid python identifier.
This is required to correctly package the pipeline so it can be imported as a
module.
Valid characters for identifiers are ASCII A-Z, a-z, _ (underscore)
and 0-9 (except as the first character)
```

### Types of change
enhancement

## Checklist


spacy/cli/package.py
```diff
@@ -430,6 +446,18 @@ def _format_label_scheme(data: Dict[str, Any]) -> str:
...
+def _is_permitted_package_name(package_name: str) -> bool:
+    permitted_match = re.search(
+        r"^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$", package_name, re.IGNORECASE
+    )
+    return permitted_match is not None
+
+
+def _contains_multiple_underscores(package_name: str) -> bool:
+    permitted_match = re.search(r"_{2,}", package_name, re.IGNORECASE)
+    return permitted_match is not None
```
adrianeboyd, line 458:
```suggestion
    return "__" in package_name
```

You can probably just use this directly instead of in a separate function.

spacy/cli/package.py
```diff
@@ -113,8 +114,18 @@ def package(
...
+                "For specific details see: https://docs.python.org/3/reference/lexical_analysis.html#identifiers",
+                exits=1,
+            )
+        if not _is_permitted_package_name(name):
+            msg.fail(
+                f"Model name ('{name}') is not a permitted package name. "
+                "This is required to correctly package the pipeline.",
+                "We recommend names that use ASCII A-Z, a-z, _ (underscore), "
+                "and 0-9, and start with a character. "
+                "For specific details see: https://www.python.org/dev/peps/pep-0426/#name",
```
adrianeboyd, line 128: The restriction about the first character don't apply because this is concatenated with `lg_` for the final package name.

spacy/cli/package.py
```diff
@@ -109,6 +110,24 @@ def package(
...
+                "This is required to correctly package the pipeline so it can be imported as a module.",
+                "We recommend names that use ASCII A-Z, a-z, _ (underscore), "
+                "and 0-9 (except as the first character). "
+                "For specific details see: https://docs.python.org/3/reference/lexical_analysis.html#identifiers",
+                exits=1,
+            )
+        if not _is_permitted_package_name(name):
+            msg.fail(
+                f"Model name ('{name}') is not a permitted package name. "
+                "This is required to correctly install the pipeline after packaging.",
```
pmbaumgartner, line 125: @adrianeboyd Is this the right specific reason for this error? I could use your help on the wording here. My understanding is that the package should still be able to be created, but it wouldn't be `pip` install-able since it's not a valid package name and it would be normalized if they uploaded it to PyPI?

spacy/cli/package.py
```diff
@@ -113,8 +114,18 @@ def package(
             msg.fail(
                 f"Model name ('{name}') is not a valid python identifier. "
                 "This is required to correctly package the pipeline so it can be imported as a module.",
```
adrianeboyd, line 116: I'm still not very fond of "python identifier" here. Something more like "not a valid module name" and "so it can be imported as a module" and not about packaging per se. I think the link is going to be confusing given the next check, since it doesn't really describe the final constraints.

spacy/cli/package.py
```diff
@@ -430,6 +446,18 @@ def _format_label_scheme(data: Dict[str, Any]) -> str:
     return md.text
 
 
+def _is_permitted_package_name(package_name: str) -> bool:
+    permitted_match = re.search(
+        r"^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$", package_name, re.IGNORECASE
```
adrianeboyd, line 451: Please add a note about where this regex came from.

spacy/cli/package.py
adrianeboyd: Well, we wouldn't usually upload model packages to pypi, so that's not the direct concern. It's a concern that you can package and install it, but then you can't load it with `spacy.load`.

adrianeboyd: I'm not sure whether a least a few of the edge cases with `spacy.load` are our fault and not some packaging internals, but the overall goal is that you want to install something you can load normally.

spacy/cli/package.py
pmbaumgartner: Good point. That actually makes me think we should check this _after_ the language is concatenated to the model name, rather than try and incorporate an exception to these rules to account for the fact that the language will eventually be added.

spacy/cli/package.py
```diff
@@ -173,6 +192,11 @@ def package(
             util.run_command([sys.executable, "setup.py", "bdist_wheel"], capture=False)
         wheel = main_path / "dist" / f"{model_name_v}{WHEEL_SUFFIX}"
         msg.good(f"Successfully created binary wheel", wheel)
+    if "__" in model_name:
+        msg.warn(
+            f"Model name ('{model_name}') contains a run of underscores. "
+            "These will be squashed to a single underscore when normalized upon PyPI upload.",
```
adrianeboyd, line 198: We wouldn't recommend uploading model packages to pypi (they would normally be rejected due to size anyway), so this isn't the explanation. The reason is that pip doesn't treat it as a distinct package name.

```suggestion
            f"Model name ('{model_name}') contains a run of underscores. "
            "Since pip install squashes these to a single underscore, we recommend "
            "package names without sequences of underscores."
```

pmbaumgartner: I admit I wasn't completely understanding the implication here or the cause of the squashing. My understanding is now as follows:

`setuptools`, via `python setup.py build_wheel`, (not `pip`) will squash the run of underscores when the wheel is created. 

e.g.
```
python -m spacy package model-best output/ --name double__underscore --build wheel
```

Results in a wheel file at `output/en_double__underscore-0.0.0/dist/en_double_underscore-0.0.0-py3-none-any.whl` 
 
Which means we would need to fix the success message, because it's using the unsquashed name (this file does not exist):
```
 ✔ Successfully created binary wheel
output/en_double__underscore-0.0.0/dist/en_double__underscore-0.0.0-py3-none-any.whl
```
 
`pip` gets implicated now when a user chooses to install the package if they're pulling it from a repository, because "Comparison of project names is case insensitive and treats arbitrarily-long runs of underscores, hyphens, and/or periods as equal." Meaning, this model package from a repository could be installed with `pip install double__underscore` or `pip install double_underscore`. If there's already a `double_underscore` or `double-underscore` package on that repository, it will result in a conflict when they attempt to upload.

I haven't fully explored the world of python packaging so some of my understanding could still be wrong here. It also does this for the `egg-info` directory that's created, but I'm not sure if that matters.

---

All that said, I think our solution should remain mostly the same:
- We should warn them that `pip` (on install) and `setuptools` (on wheel creation) squash runs of underscores, and link to https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#name
- We should correct the success message for wheel creation to point to the correct file

spacy/cli/package.py
pmbaumgartner: Function removed and used directly in more recent commit.

spacy/cli/package.py
adrianeboyd: I don't think users typically care much whether it's `pip` or `setuptools`, and the project names also still don't have the same restrictions we're focusing on here, so I wouldn't include that link (all the examples with `-` and `.` are confusing) and I would focus on:

* warn that runs of underscores are not significant in installed package names
* fix the reported wheel filename (yeesh on the sdist vs. wheel names here)

pmbaumgartner: Sounds good! I was trying to anticipate where a user would encounter an issue if they ignored the warning, but that might be too clever/detailed. I updated with the general warning you proposed.

spacy/cli/package.py
pmbaumgartner: I'm going to leave the errors where they are because it'll be much simpler to maintain. Otherwise we'll have to go into too much detail trying to communicate what happens to a package name -- e.g. they could name a model `_leading_underscore` and have it be magically fixed by our naming scheme (it would be come `lang_leading_underscore`), but I think we should encourage the best practice here up front.

spacy/tests/test_cli.py
```diff
@@ -16,7 +16,10 @@
 from spacy.cli.debug_data import _get_labels_from_spancat
 from spacy.cli.download import get_compatibility, get_version
 from spacy.cli.init_config import RECOMMENDATIONS, init_config, fill_config
-from spacy.cli.package import get_third_party_dependencies
+from spacy.cli.package import (
+    get_third_party_dependencies,
+    _is_permitted_package_name,
+)
```
adrianeboyd, line 22: We avoid the multiline imports in our code:

```suggestion
from spacy.cli.package import get_third_party_dependencies
from spacy.cli.package import _is_permitted_package_name
```

PR #10197, opened 2022-02-03
Tagger: use unnormalized probabilities for inference

danieldk:
## Description

Using unnormalized softmax avoids use of the relatively expensive exp function,
which can significantly speed up non-transformer models (e.g. I got a speedup
of 27% on a German tagging + parsing pipeline).

### Types of change

Performance improvement

## Checklist

Draft since this requires https://github.com/explosion/thinc/pull/583 and a new Thinc version.


adrianeboyd:
There are users who use the scores directly and (I assume) would be expecting them to be normalized. I'm not say we necessarily shouldn't do this, but this proposal doesn't give these users any way to control this behavior, does it?

danieldk:
> There are users who use the scores directly and (I assume) would be expecting them to be normalized. I'm not saying we necessarily shouldn't do this, but this proposal doesn't give these users any way to control this behavior, does it?

We could make this configurable. Maybe we should also only provide this functionality in a `Tagger.v2` so that if anyone uses the probabilities it doesn't break for them?

spacy/ml/models/tagger.py
```diff
@@ -7,8 +7,21 @@
 
 
 @registry.architectures("spacy.Tagger.v1")
```
adrianeboyd, line 9: I think this should probably be moved to `spacy-legacy` instead.

danieldk: Ok, I have also made a draft pull request for `spacy-legacy`: https://github.com/explosion/spacy-legacy/pull/20

website/docs/api/architectures.md
```diff
@@ -631,11 +632,22 @@ model simply adds a linear layer with softmax activation to predict scores given
 the token vectors.
 
 | Name        | Description                                                                                |
-| ----------- | ------------------------------------------------------------------------------------------ |
+|-------------|--------------------------------------------------------------------------------------------|
 | `tok2vec`   | Subnetwork to map tokens into vector representations. ~~Model[List[Doc], List[Floats2d]]~~ |
 | `nO`        | The number of tags to output. Inferred from the data if `None`. ~~Optional[int]~~          |
+| `normalize` | Normalize probabilities during inference. Defaults to `False`. ~~bool~~           |
```
svlandeg, line 638: Nitpick: could you run `prettier` on the markdown formatting? :-)

svlandeg (approved):
Agreed that it's a good idea to make this configurable in a new version of the `Tagger`.

I find it slightly unintuitive to go from a v1 with always normalization, to a v2 with default normalization off. It might still catch some users off-guard that the behaviour changes between the two versions when you rely on defaults. But then again it is documented in the docs, and I can see why this would be beneficial in most cases / for most users. So I'm leaning towards keeping it like the PR has it currently :-)

website/docs/api/architectures.md
danieldk: Done!

PR #10225, opened 2022-02-07
Add github actions for slow and gpu tests

ryndaniels:
## Description
Adds two github actions to trigger matrices of automated buildkite builds on the `master`, `develop`, and `v4` branches - daily for slow tests, weekly for GPU tests

### Types of change
Tests only

## Checklist

### DIFF
--- a/.github/workflows/slowtests.yml
+++ b/.github/workflows/slowtests.yml
@@ -1,0 +1,19 @@
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  daily-slowtests:
+    strategy:
+      matrix:
+        branch: [master, develop, v4]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger buildkite build
+        uses: buildkite/trigger-pipeline-action@v1.2.0
+        env:
+          PIPELINE: explosion-ai/spacy-slow-tests
+          BRANCH: ${{ matrix.branch }}
+          MESSAGE: ":github: Daily slow tests - triggered from a GitHub Action"
+        secrets:
+          BUILDKITE_API_ACCESS_TOKEN: ${{ secrets.BUILDKITE_SECRET }}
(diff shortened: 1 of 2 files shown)


.github/workflows/slowtests.yml
```diff
@@ -0,0 +1,19 @@
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
svlandeg, line 3: Will this run regardless of whether the code repo has changed? I think that would be fine for the weekly tests, but for the daily tests I was hoping that we could check whether the code has changed in the last 24 hours, and avoid running the tests altogether if it hasn't.

ryndaniels: Right now they are set up to run every day. Are there specific criteria you'd want to use for if the code has changed (eg, changes in the `spacy` directory, or any commits to the branch within the past 24 hours)? There might be a way to do one of those checks as a first step, though it may be a bit hacky :)

svlandeg: I would check for any edit, anywhere. If it can't easily be done, or if it would be too hacky, then I would change the frequency of the `develop` and `v4` branches to weekly (for both GPU & slow), and only run the slow tests on `master` daily.

ryndaniels: Ok, I changed it to something that will (hopefully) check for and only run the tests if there are commits on the relevant branch in the past day... untested for now, but if it looks good we can try it and iterate if needed

svlandeg (approved):
Seems plausible to me :-) Let's merge it in and see what happens! I'll cherry pick to the other 3 branches so that we can see what happens tonight.

svlandeg:
Or maybe we just let this run on `master` tonight and cherry-pick to the other branches tomorrow if that went well.

ryndaniels:
Sounds good! If it needs faster troubleshooting I can also add in a trigger to run on commits to some test branch and then use that for a while for testing.

PR #10231, opened 2022-02-07
Add edit tree lemmatizer

adrianeboyd:
## Description

Add edit tree lemmatizer, converted from [`spacy_experimental.edit_tree_lemmatizer`](https://github.com/explosion/spacy-experimental/tree/d01fd5b479db823772865c362b4e4e1e706cf554/spacy_experimental/edit_tree_lemmatizer)

### Types of change

Enhancement

## Checklist


spacy/pipeline/edit_trees.pxd
```diff
@@ -0,0 +1,92 @@
+from libc.stdint cimport uint32_t, uint64_t
+from spacy.typedefs cimport attr_t, hash_t, len_t
+from spacy.strings cimport StringStore
```
danieldk, line 3: I think that these could now be relative imports?

spacy/tests/pipeline/test_edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,280 @@
...
+import pytest
+from hypothesis import given
+import hypothesis.strategies as st
+from spacy import util
+from spacy.lang.en import English
+from spacy.language import Language
+from spacy.pipeline.edit_trees import EditTrees
+from spacy.training import Example
+from spacy.strings import StringStore
+from spacy.util import make_tempdir
```
danieldk, line 11: These could now be relative imports.

adrianeboyd: The test suite intentionally uses absolute imports so it behaves the same when you run the tests on the installed package.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,376 @@
...
+                if gold_lemma is None:
+                    label = -1
+                else:
+                    tree_id = self.trees.add(predicted.text, gold_lemma)
+                    label = self.tree2label.get(tree_id, 0)
+                eg_truths.append(label)
+
+            truths.append(eg_truths)
+
+        truths = cast(List[Union[Ints1d, Floats2d]], truths)
```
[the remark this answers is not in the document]
svlandeg: Sometimes you can also add in an `assert` if you're 100% of the type at that point in the code, and then `mypy` will use that in the subsequent lines.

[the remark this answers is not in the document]
adrianeboyd: Hmm, it doesn't like that either: 

`Parameterized generics cannot be used with class or instance checks`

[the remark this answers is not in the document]
adrianeboyd: The weird cast doesn't seem worse than ignoring it, but it seems like there ought to be a better solution.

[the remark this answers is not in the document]
danieldk: Maybe it would only work if the loss took `List[V]` where `V` is a typevar with `Ints1d` and `Floats2d` as the possible types. I am fine with keeping it as is.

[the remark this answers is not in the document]
danieldk: My only worry is that if we extend the union in the loss, type checking breaks here.

adrianeboyd:
I'm not sure that `spacy.pipelines` was the best place for `edit_trees`, but there wasn't an immediately obvious better place to put it.

I'm also not sure that `edit_tree_lemmatizer` is the name we should go with in the end, since it's kind of opaque/jargon-y for users (I can't think of the actual name for this ... where the name is from the developers' perspective rather than the users').

My original intention was to have this kind of component as a `mode` of `lemmatizer`, which has tried to follow the `TrainablePipe` API enough to make it possible in the future (mainly having a `model` parameter), but now I think it will be an overall API headache to mix trainable and not-trainable modes.

danieldk:
> I'm also not sure that `edit_tree_lemmatizer` is the name we should go with in the end, since it's kind of opaque/jargon-y for users (I can't think of the actual name for this ... where the name is from the developers' perspective rather than the users').

Maybe `TrainableLemmatizer`/`trainable_lemmatizer`?

adrianeboyd:
So what I mean is the default component name / factory name and not the class name, where I think that `EditTreeLemmatizer` is fine.

Nothing else is `trainable_x` and "trainable" is expressed in other ways. But maybe there aren't any better alternatives...

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,380 @@
...
+        self.top_k = top_k
+
+        self.trees = EditTrees(self.vocab.strings)
+        self.tree2label: Dict[int, int] = {}
+
+        self.cfg: Dict[str, Any] = {"labels": []}
+        self.scorer = scorer
+
+    def get_loss(
+        self, examples: Iterable[Example], scores
```
svlandeg, line 121:
```suggestion
        self, examples: Iterable[Example], scores: List[Floats2d]
```

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,380 @@
...
+    def labels(self) -> Tuple[int, ...]:
+        """Returns the labels currently added to the component."""
+        return tuple(self.cfg["labels"])
+
+    @property
+    def hide_labels(self) -> bool:
+        return True
+
+    @property
+    def label_data(self):
```
svlandeg, line 221:
```suggestion
    def label_data(self)  -> Dict:
```

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,380 @@
...
+        if "trees" not in labels:
+            raise ValueError(Errors.E857.format(name="trees"))
+
+        self.cfg["labels"] = list(labels["labels"])
+
+        # Do not modify the caller's data structures.
+        trees = deepcopy(labels["trees"])
+
+        for tree in trees:
+            if not tree["is_match_node"]:
```
svlandeg, line 336: Should we do a check here on `tree` being a properly structured dict with the correct fields? Or are these internals that can't (shouldn't) go wrong?

spacy/tests/pipeline/test_edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,280 @@
...
+    lemmatizer2.initialize(
+        get_examples=lambda: train_examples, labels=lemmatizer.label_data
+    )
+
+    # Verify that the labels and trees are the same.
+    assert lemmatizer.labels == lemmatizer2.labels
+    assert lemmatizer.trees.to_bytes() == lemmatizer2.trees.to_bytes()
+
+
+def test_dutch():
```
svlandeg, line 189: 🤩 😁

svlandeg:
> I'm not sure that `spacy.pipelines` was the best place for `edit_trees`, but there wasn't an immediately obvious better place to put it.

I was thinking that too. But having `spacy.pipeline._edittree_internals` like we have for the `parser` is probably overkill. 

For the `entity_linker`, we did end up deciding having `kb.pyx` and `kb.pxd` in the root `spacy` module though.

adrianeboyd:
Maybe `_internals` and we move the parser stuff there at some point, too?

svlandeg:
> Nothing else is `trainable_x` and "trainable" is expressed in other ways. But maybe there aren't any better alternatives...

`trainable_x` does make the distinction more clear to users, I think, but I don't like it for the same reason you mention. Alternatively: `neural_lemmatizer`. Though it has the same issues, at least it's shorter :|

svlandeg:
> Maybe `_internals` and we move the parser stuff there at some point, too?

I don't think that'll work, because the internal parser classes are so genericly named that then it might become more difficult to see what all the parser code is...

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,380 @@
...
+                doc_compat_guesses.append(tree_id)
+
+            guesses.append(np.array(doc_compat_guesses))
+
+        return guesses
+
+    def set_annotations(self, docs, batch_tree_ids):
+        if isinstance(docs, Doc):
+            docs = [docs]
+
```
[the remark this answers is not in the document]
danieldk: I think I took this from the `Tagger` pipe, which does the same. But I agree that we should probably not do this.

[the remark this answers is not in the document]
danieldk: Fixed.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,380 @@
...
+            assert len(guesses) == len(docs)
+            return guesses
+
+        scores = self.model.predict(docs)
+        assert len(scores) == len(docs)
+
+        guesses = self._scores2guesses(docs, scores)
+        assert len(guesses) == len(docs)
+
+        return guesses
```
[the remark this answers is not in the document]
danieldk: Fixed.

spacy/pipeline/edit_tree_lemmatizer.py
danieldk: Since the trees can be passed through `initialized`, that would be good! I have written a schema using pydantic, but I am going to try to add this to `EditTrees.from_bytes` as well.

danieldk: I have added validation. I also had to change the on-disk format a bit, so that we can use the same schemas for both. While at it, I also simplified the on-disk format (it looks less like the C structs).

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -0,0 +1,380 @@
...
+            doc_tree_ids = batch_tree_ids[i]
+            if hasattr(doc_tree_ids, "get"):
+                doc_tree_ids = doc_tree_ids.get()
+            for j, tree_id in enumerate(doc_tree_ids):
+                if self.overwrite or doc[j].lemma == 0:
+                    # If no applicable tree could be found during prediction,
+                    # the special identifier -1 is used. Otherwise the tree
+                    # is guaranteed to be applicable.
+                    if tree_id == -1:
+                        # TODO: generalize with a getter setting
```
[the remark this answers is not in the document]
danieldk: I think my idea was to make it customizable what to do when no lemma could be predicted. However, if people want custom behavior, I think it is better that they disable backoff and add their own pipe to do post-processing.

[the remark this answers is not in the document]
danieldk: Removed the TODO comment.

svlandeg (approved):
Looks like this is ready to merge? 

We can revise the cast thing later, it'd require an update of Thinc and I don't want to hold up this PR over something like that.

adrianeboyd:
No, not yet. I would wait a bit until the potential integrations with `Lemmatizer` have been explored a bit more.

svlandeg:
Oh, ok. I thought we didn't want to go there initially, but I'll leave that up to you Adriane. I'll put this in draft to signal that we shouldn't merge yet!

adrianeboyd:
We decided that an integration with `Lemmatizer` or the `lemmatizer` factory is not a particular useful choice. It's already supported to do something like this:

```
nlp.add_pipe("trainable_lemmatizer", name="lemmatizer")
```

adrianeboyd:
Now this needs docs...

danieldk:
> Now this needs docs...

Added, hope I didn't miss anything.

website/docs/usage/linguistic-features.md
```diff
@@ -197,6 +200,20 @@ information, without consulting the context of the token. The rule-based
 lemmatizer also accepts list-based exception files. For English, these are
 acquired from [WordNet](https://wordnet.princeton.edu/).
 
+### Trainable lemmatizer
+
+The [`EditTreeLemmatizer`](/api/edittreelemmatizer) can learn form-to-lemma
+transformations a training set that includes lemma annotations. This removes the
```
adrianeboyd, line 206:
```suggestion
transformations from a training corpus that includes lemma annotations. This removes the
```

website/docs/usage/processing-pipelines.md
```diff
@@ -303,22 +303,23 @@ available pipeline components and component functions.
...
+| `morphologizer`        | [`Morphologizer`](/api/morphologizer)                | Assign morphological features and coarse-grained POS tags.                                |
+| `attribute_ruler`      | [`AttributeRuler`](/api/attributeruler)              | Assign token attribute mappings and rule-based exceptions.                                |
+| `senter`               | [`SentenceRecognizer`](/api/sentencerecognizer)      | Assign sentence boundaries.                                                               |
+| `sentencizer`          | [`Sentencizer`](/api/sentencizer)                    | Add rule-based sentence segmentation without the dependency parse.                        |
+| `tok2vec`              | [`Tok2Vec`](/api/tok2vec)                            | Assign token-to-vector embeddings.                                                        |
+| `trainable_lemmatizer` | [`EditTreeLemmatizer`](/api/edittreelemmatizer)      | Assign base forms to words.                                                               |
```
adrianeboyd, line 321: Let's move this next to `lemmatizer`. (Or possible just sort this table, it is too long.)

website/docs/api/edittreelemmatizer.md
```diff
@@ -0,0 +1,404 @@
...
+| `overwrite`     | Whether existing annotation is overwritten. Defaults to `False`. ~~bool~~                                                                                                                                                                                                                                          |
+| `top_k`         | The number of most probable edit trees to try before resorting to `backoff`. Defaults to `1`. ~~int~~                                                                                                                                                                                                              |
+| `scorer`        | The scoring method. Defaults to [`Scorer.score_token_attr`](/api/scorer#score_token_attr) for the attribute `"lemma"`. ~~Optional[Callable]~~                                                                                                                                                                      |
+
+```python
+%%GITHUB_SPACY/spacy/pipeline/edit_tree_lemmatizer.py
```
adrianeboyd, line 57:
```suggestion
%%GITHUB_SPACY/spacy/pipeline/edit_tree_lemmatizer.pyx
```

website/docs/api/edittreelemmatizer.md
```diff
@@ -0,0 +1,404 @@
...
+| `bytes_data`   | The data to load from. ~~bytes~~                                                            |
+| _keyword-only_ |                                                                                             |
+| `exclude`      | String names of [serialization fields](#serialization-fields) to exclude. ~~Iterable[str]~~ |
+| **RETURNS**    | The `EditTreeLemmatizer` object. ~~EditTreeLemmatizer~~                                     |
+
+## EditTreeLemmatizer.labels {#labels tag="property"}
+
+The labels currently added to the component. **Warning:** the
+`EditTreeLemmatizer` labels are not useful by themselves, since they are
+identifiers of edit trees.
```
adrianeboyd, line 363: Move to a yellow warning block?

website/docs/api/edittreelemmatizer.md
danieldk: It's really `.py`, I changed it sometime when no Cython was needed in the lemmatizer itself.

website/docs/usage/linguistic-features.md
danieldk: Fixed.

website/docs/usage/processing-pipelines.md
danieldk: Fixed.

website/docs/api/edittreelemmatizer.md
danieldk: Yeah, that looks much better.

website/docs/api/edittreelemmatizer.md
```diff
@@ -0,0 +1,409 @@
+---
+title: EditTreeLemmatizer
+tag: class
+source: spacy/pipeline/edit_tree_lemmatizer.pyx
```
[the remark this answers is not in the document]
danieldk: Fixed.

PR #10240, opened 2022-02-08
Update Russian stop_words.py

gremur:
1. Extended list of stopwords
2. Added "integrated" list preprocess (1. remove duplicates if any take place by mistake, 2. remove empty tokens, 3. sort list in alphabetical order, 4. sort list by word length)


polm:
Thanks for the PR! 

The tests are failing, it looks like it's a type issue - the list needs to be a `Set[str]` at the end. So you should be able to fix that by wrapping everything in a `set()` call. (I would fix it but it looks like I don't have permission to modify your branch.)

Also, could you break the nested function calls up a bit? It doesn't have to be one per line, but as it is it's a little hard to follow what's going on. Something like this:

```
STOP_WORDS = filter(None, list(set(STOP_WORDS)))
STOP_WORDS = sorted(STOP_WORDS) # alpha sort
STOP_WORDS = set(sorted(STOP_WORDS, key=len, reverse=True)) # len sort
```

adrianeboyd:
Since this variable is a set in the end, there's no need to sort or deduplicate, and I don't understand how the `filter` is necessary following `str.split()`?

We'll switch this PR to `develop` so that these changes would be in v3.3.0 rather than the next patch release of v3.2. Don't worry if a few additional commits show up temporarily as part of the PR, we'll fix it on our end before it's merged.

PR #10241, opened 2022-02-08
Make core projectivization functions cdef nogil

danieldk:
## Description

While profiling the parser, I noticed that relatively a lot of time is spent in projectivization. This change rewrites the functions in the core loops as `cdef nogil` for efficiency.

In C++-land, we use vector in place of Python lists and absent heads are represented as `-1` in place of `None`.

Profile in the refactored parser, before:

projectivize-before

After:

projectivize-after

We still need to expose `_get_smallest_nonproj_arc` as a Python function for tests, but I couldn't think of anything better than `get_smallest_nonproj_arc_slow` 😅.

### Types of change

Performance improvement.

## Checklist


honnibal:
Looks good to me!

PR #10245, opened 2022-02-09
add tmtoolkit package to spaCy universe

internaut:
Add the [tmtoolkit package](https://tmtoolkit.readthedocs.io/) to the spaCy universe.

## Description

No tests needed.

### Types of change

Documentation only.

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -3792,6 +3792,39 @@
                 "twitter": "jboy"
             },
             "category": ["visualizers", "standalone"]
+        },
+        {
+            "id": "tmtoolkit",
+            "slogan": "Text mining and topic modeling toolkit",
+            "description": "tmtoolkit is a set of tools for text mining and topic modeling with Python developed especially for the use in the social sciences, in journalism or related disciplines. It aims for easy installation, extensive documentation and a clear programming interface while offering good performance on large datasets by the means of vectorized operations (via NumPy) and parallel computation (using Python’s multiprocessing module and the loky package).",
+            "github": "WZBSocialScienceCenter/tmtoolkit",
+            "code_example": [
+                "from tmtoolkit.corpus import Corpus, tokens_table, lemmatize, to_lowercase, dtm",
+                "from tmtoolkit.bow.bow_stats import tfidf, sorted_terms_table",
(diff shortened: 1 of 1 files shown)


polm:
Thanks for the submission, this looks like a cool project! 

That said, I had some trouble running your sample code though. It wouldn't run until I installed `bidict` and `loky` separately - it seems like those should be specified as dependencies. 

Also, if I run it without `en_core_web_sm` installed, I don't get the usual error that `spacy.load` gives about the model not being installed - it looks like you call `spacy.info` before `spacy.load`, and the error there isn't as clear about models that aren't installed. We should improve that error, but you might want to check that the model is installed first.

After installing bidict, loky, and en_core_web_sm the code worked fine. 

Could you add loky and bidict to the project's dependencies, or is there a reason they aren't already in there?

internaut:
Thank you for the quick response. If you only run `pip install tmtoolkit`, a **minimal** version is installed, which only allows a subset of the features to be used (mainly for topic modeling / BoW). The required dependencies for the given example are only installed if you run `pip install -U "tmtoolkit[recommended]"`. After that, you should download the necessary language model for English with the command `python -m tmtoolkit setup en`. All of this is described in the [installation notes](https://tmtoolkit.readthedocs.io/en/latest/install.html).

I chose to do so, because some people may want to use a different NLP package than spaCy and hence I wanted to make spaCy and the other packages (bidict, loky) for the NLP stuff an optional dependency. I wanted to treat the installation options so that `pip install tmtoolkit` installs the *recommended* dependencies and `pip install "tmtoolkit[minimal]"` only the minimal dependencies, but unfortunately this doesn't seem to be possible with setuptools at the moment.

I will however provide an update which gives a hint to the installation instructions when trying to import the NLP stuff without having installed the required dependencies. I will also update the README and place these hints there.

polm:
Ah OK, thanks for the clarification - that makes sense, and I've been bitten by that issue with setuptools myself. 

The use case I'm imagining here is someone who sees the Universe page first, runs `pip install`, and tries your sample code.  What we can do here is just put a comment at the top of the sample code that it requires the "recommended" install.

Having explicit errors when imports fail would be a great addition for the future.

PR #10247, opened 2022-02-09
Extend list of stopwords for ru language

gremur:
Extended list of stopwords.


gremur:
Something went wrong. I will recreate pull request.

PR #10254, opened 2022-02-10
Fix the date command and the matrix failure mode

ryndaniels:


svlandeg (approved):
I'm enjoying this daily ritual :-)

PR #10280, opened 2022-02-14
Extend list of numbers for ru language

gremur:
Extended list of numbers with all forms and cases including short forms, slang variants and roman numerals.


gremur (commented):
Corrected typo

gremur (commented):
May be support of numbers with percentages should be added to 'like_num' function for majority of languages.

thomashacker (approved):
Looks good. Thanks for the `PR`! 🎉

PR #10282, opened 2022-02-14
Extend list of abbreviations for ru language

gremur:
Extended list of abbreviations for ru language those may have influence on tokenization.


thomashacker (approved):
Everything looks good.
Thank you very much for your contribution! 😄

PR #10295, opened 2022-02-15
Delay loading of mecab in Korean tokenizer

adrianeboyd:
## Description

Delay loading of mecab until the tokenizer is called the first time so that it's possible to initialize a blank `ko` pipeline without having mecab installed, e.g. for use with `spacy init vectors`.

In addition, remove `KoreanTokenizer.__del__`, which appears to lead to additional ignored exceptions rather than fewer with the current version of mecab.

### Types of change

?

## Checklist

### DIFF
--- a/spacy/lang/ko/__init__.py
+++ b/spacy/lang/ko/__init__.py
@@ -31,14 +31,23 @@ def create_tokenizer():
 class KoreanTokenizer(DummyTokenizer):
     def __init__(self, vocab: Vocab):
         self.vocab = vocab
-        MeCab = try_mecab_import()  # type: ignore[func-returns-value]
-        self.mecab_tokenizer = MeCab("-F%f[0],%f[7]")
+        self._mecab_tokenizer = None
+
+    @property
+    def mecab_tokenizer(self):
+        # This is a property so that initializing a pipeline with blank:ko is
+        # possible without actually requiring mecab-ko, e.g. to run
+        # `spacy init vectors ko` for a pipeline that will have a different
+        # tokenizer in the end. The languages need to match for the vectors
+        # to be imported and there's no way to pass a custom config to
+        # `init vectors`.
+        if self._mecab_tokenizer is None:
+            MeCab = try_mecab_import()  # type: ignore[func-returns-value]
+            self._mecab_tokenizer = MeCab("-F%f[0],%f[7]")
+        return self._mecab_tokenizer
 
     def __reduce__(self):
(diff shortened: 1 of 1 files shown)


adrianeboyd:
This is ... not great? But the alternative of needing `mecab-ko` to be installed to run `spacy init vectors ko` is a UX headache. The language setting doesn't actually matter for the vectors model, but if the language doesn't match, you run into errors during initialization.

There are other alternatives (config overrides for `init vectors`, alternate checks when loading a vectors model with an existing vocab, etc.) but this seemed like a quick workaround for a relatively niche use case. This is the only remaining language-specific tokenizer that requires non-python packages to load.

adrianeboyd:
Not sure why the CI didn't run, let me close and reopen to see...

spacy/lang/ko/__init__.py
```diff
@@ -31,15 +31,24 @@ def korean_tokenizer_factory(nlp):
 class KoreanTokenizer(DummyTokenizer):
     def __init__(self, vocab: Vocab):
         self.vocab = vocab
-        MeCab = try_mecab_import()  # type: ignore[func-returns-value]
-        self.mecab_tokenizer = MeCab("-F%f[0],%f[7]")
+        self._mecab = try_mecab_import()  # type: ignore[func-returns-value]
```
svlandeg, line 34: Won't this still raise an `ImportError` when `MeCab` is not installed?

adrianeboyd: No, this will raise an error if `natto-py` is not installed, but it doesn't check for mecab yet. I think most users would prefer to be warned early if their pipeline isn't going to work (with `spacy.blank("ko")`) rather than when they first call the tokenizer.

So this warns early and still lets me run `spacy init vectors ko` with just `natto-py` installed (which is easy to install).

adrianeboyd: It does mean the error message isn't 100% accurate.

svlandeg: Right, that makes sense :-) 
So good to merge, or do you want to change anything to the error msg first?

adrianeboyd: No, I can't think of a way to phrase it that wouldn't be even more confusing, and I'm mainly thinking of existing users who are used to the mecab setup. You can still load a `ko` pipeline with a different tokenizer without `natto-py` installed, and `spacy[ko]` includes `natto-py`, so we can point to that if the `init vectors ko` situation if it comes up.

PR #10302, opened 2022-02-16
Fix NER check in CoNLL-U converter

adrianeboyd:
## Description

Leave ents unset if no NER annotation is found in the MISC column.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/training/converters/conllu_to_docs.py
+++ b/spacy/training/converters/conllu_to_docs.py
@@ -44,16 +44,21 @@ def conllu_to_docs(
         yield Doc.from_docs(sent_docs_to_merge)
 
 
-def has_ner(lines, ner_tag_pattern):
+def has_ner(input_data, ner_tag_pattern):
     """
     Check the MISC column for NER tags.
     """
-    for line in lines:
-        parts = line.split("\t")
-        id_, word, lemma, pos, tag, morph, head, dep, _1, misc = parts
-        for misc_part in misc.split("|"):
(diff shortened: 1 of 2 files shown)


spacy/training/converters/conllu_to_docs.py
```diff
@@ -246,7 +251,8 @@ def conllu_sentence_to_doc(
         deps=deps,
         heads=heads,
     )
-    doc_x.ents = [Span(doc_x, ent.start, ent.end, label=ent.label) for ent in doc.ents]
+    if ents is not None:
```
svlandeg, line 254: I find it slightly confusing why this checks the `ents` var - semantically it would be the same as checking `set_ents`, no? I feel like that would make the code easier to read, but maybe I'm missing something.

adrianeboyd: It is this way because it was a per-sentence check in the first draft, and it was the most minimal way to know the results of the check in the second spot without an additional variable. But there are some corpora where `O` is missing, so a top-level check is better. (All the code was there to do this check, so I'm not sure when it got omitted in the past.)

I don't really care either way...

svlandeg, line 254:
```suggestion
    if set_ents:
```

svlandeg: I think it's just a tad more explicit the other way around :-)

PR #10319, opened 2022-02-17
Alignment: use a simplified ragged type for performance

danieldk:
## Description

This introduces the `AlignmentArray` type, which is a simplified version of `Ragged` that performs better on the simple(r) indexing performed for alignment.

When training a biaffine parser pipeline with morphology and tagging on a GPU, this speeds up training by about 19%. This should improve performance for all `TrainablePipe`s that use `get_alignment` to get gold standard annotations.

### Types of change

Performance improvement.

## Checklist


spacy/training/alignment_array.pyx
```diff
@@ -0,0 +1,66 @@
...
+                self._data[idx] = inner
+                idx += 1
+
+    def __getitem__(self, idx):
+        starts = self._starts_ends[:-1]
+        ends = self._starts_ends[1:]
+        if isinstance(idx, int):
+            start = starts[idx]
+            end = ends[idx]
+        elif isinstance(idx, slice):
```
danieldk, line 38: I just noticed that I don't throw an exception when the index is not an int or a slice. I'll add a commit to fix this.

danieldk:
Out of draft: trained a transformer model and didn't see any regressions in accuracy.

spacy/training/alignment_array.pyx
```diff
@@ -0,0 +1,69 @@
...
+            if not (idx.step is None or idx.step == 1):
+                msg = "only a step of 1 is supported when slicing AlignmentArray"
+                raise ValueError(msg)
```
adrianeboyd, line 41: Can you move the error messages into `Errors`?

danieldk: Fixed.

adrianeboyd:
Also tested on a full set of trained pipeline builds.

PR #10348, opened 2022-02-21
Add `path.mkdir` to custom component examples of `to_disk`

pmbaumgartner:
Updates docs to add `path.mkdir()` to the custom component examples that include a `to_disk` method.

### Types of change
Documentation

## Checklist

### DIFF
--- a/website/docs/usage/saving-loading.md
+++ b/website/docs/usage/saving-loading.md
@@ -202,7 +202,7 @@ the data to and from a JSON file.
 > rules _with_ the component data.
 
 ```python
-### {highlight="14-19,21-26"}
+### {highlight="16-23,25-30"}
 from spacy.util import ensure_path
 
 @Language.factory("my_component")
--- a/website/docs/usage/saving-loading.md
+++ b/website/docs/usage/saving-loading.md
@@ -218,6 +218,7 @@ class CustomComponent:
 
     def to_disk(self, path, exclude=tuple()):
         # This will receive the directory path + /my_component
+        path.mkdir()
         data_path = path / "data.json"
         with data_path.open("w", encoding="utf8") as f:
             f.write(json.dumps(self.data))
@@ -468,6 +469,7 @@ component exposes a `to_disk` method, it will be called with the disk path.
 
 ```python
 def to_disk(self, path, exclude=tuple()):
+    path.mkdir()
     snek_path = path / "snek.txt"
     with snek_path.open("w", encoding="utf8") as snek_file:
         snek_file.write(self.snek)
--- a/website/docs/usage/saving-loading.md
+++ b/website/docs/usage/saving-loading.md
@@ -202,7 +202,9 @@ the data to and from a JSON file.
 > rules _with_ the component data.
 
 ```python
-### {highlight="14-18,20-25"}
+### {highlight="14-19,21-26"}
+from spacy.util import ensure_path
+
 @Language.factory("my_component")
 class CustomComponent:
     def __init__(self):
@@ -218,7 +220,9 @@ class CustomComponent:
 
     def to_disk(self, path, exclude=tuple()):
         # This will receive the directory path + /my_component
-        path.mkdir()
+        path = ensure_path(path)
+        if not path.exists():
+            path.mkdir()
         data_path = path / "data.json"
         with data_path.open("w", encoding="utf8") as f:
             f.write(json.dumps(self.data))
@@ -468,8 +472,12 @@ pipeline package. When you save out a pipeline using `nlp.to_disk` and the
 component exposes a `to_disk` method, it will be called with the disk path.
 
 ```python
+from spacy.util import ensure_path
+
 def to_disk(self, path, exclude=tuple()):
-    path.mkdir()
+    path = ensure_path(path)
+    if not path.exists():
+        path.mkdir()
     snek_path = path / "snek.txt"
     with snek_path.open("w", encoding="utf8") as snek_file:
         snek_file.write(self.snek)
--- a/website/docs/usage/processing-pipelines.md
+++ b/website/docs/usage/processing-pipelines.md
@@ -1083,12 +1083,15 @@ on [serialization methods](/usage/saving-loading/#serialization-methods).
 ```python
 ### Custom serialization methods {highlight="6-7,9-11"}
 import srsly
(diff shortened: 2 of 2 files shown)


website/docs/usage/saving-loading.md
```diff
@@ -218,6 +218,7 @@ class CustomComponent:
 
     def to_disk(self, path, exclude=tuple()):
         # This will receive the directory path + /my_component
+        path.mkdir()
```
adrianeboyd, line 221: `path` is `Union[Path, str]` and `path.mkdir()` raises an error if the directory already exists.

```suggestion
        path = spacy.util.ensure_path(path)
        if not path.exists():
            path.mkdir()
```

plus `import spacy` further up. Or you can import/use `Path` directly if you think that's clearer.

And then you most likely need to adjust the highlighted line numbers for the example.

pmbaumgartner: I'm a little lost in the logic of how all these things fit together, and maybe that goes back to my root problem: it seems odd to me that `Language.to_disk` is smart enough to dispatch a path for the component but *not* check if it exists or not for a user, so every user has to implement this`mkdir` boilerplate in every custom component made that wants to implement `to_disk`. 

For types - In this case, isn't `path` always a `Path` due to `Language.to_disk` calling `spacy.util.ensure_path(path)`? At least in the context of these examples we aren't ever calling `to_disk` only on the component and passing a string. I agree it helps to be consistent with the type being `Union[Path, str]`, but is it ever actually not `Path` when called?

I was debating a solution that included `ensure_path`, but the warning we have in the docs for utils scared me away from using this for a public docs example.

> Because utility functions are mostly intended for internal use within spaCy, their behavior may change with future releases.

adrianeboyd: You're right that if you call `Language.to_disk` the directories are created by `util.to_disk` and you get `path` as `Path`, but people may also call `Component.to_disk` directly, which I think was part of your original concern about how the examples didn't include `mkdir`?

If you want to ignore the details for this example (although we never use `Path` as the arg in our examples of calling `to_disk` anywhere in the docs, so I think it's confusing for the case you were trying to improve here), then at the very least:

```python
        if not path.exists():
            path.mkdir()
```

The `to_disk` method definitely shouldn't crash if you provide an existing directory. (As an aside: the directories aren't cleaned, so there are some minor issues, mainly for some older components, where writing to an existing directory might lead to the component not being identical when read back in due to the presence of other files.)

pmbaumgartner: Got it. I think my ultimate solution would be that `Language.to_disk` had something like `if not path.exists(): path.mkdir()` for component directories, and these examples would then work unmodified. Would that cause other issues, though?

adrianeboyd: `Language.to_disk` does include this, but I thought your concern was about people testing this independently?

pmbaumgartner: Sorry, I'll recap since I've been a few different places.

1. The examples we have in the docs for using `to_disk` with custom components don't work out of the box.
2. We can fix this by adding some variation of `if not path.exists(): path.mkdir()` to each.
3. We should make sure that `path` is a `Path`. We could do this with `util.ensure_path`, but we have a warning in there that those functions are preferred for internal use so it feels odd to add it to a public example.
This is necessary so that people can call `Component.to_disk`. However, the context of all of these examples are all in reference to calling `nlp.to_disk`, not `Component.to_disk`. (e.g. https://spacy.io/usage/saving-loading/#serialization-methods and https://spacy.io/usage/processing-pipelines#component-data-initialization), which means that we know `path` is a `Path`, and for the sake of these examples I assumed given the context we can avoid the check.
4. It's odd to me that we pass a path that doesn't exist to a component's `to_disk` method. Yes, you are right that `Language.to_disk` contains a variation of this when `util.to_disk` is called, but that only guarantees the parent directory exists, not that the directory for each writer exists. 

Implementing (2) is no problem for the examples, it's just weird to me given (3) and (4). 

My last suggestion was that we could keep the examples as is and add something to `util.to_disk` to add the conditional for the writer paths we create in there. Sorry if that was confusing, since I was referencing `Language.to_disk` (which calls `util.to_disk`).

e.g. around https://github.com/explosion/spaCy/blob/master/spacy/util.py#L1286

```python
if key.split(".")[0] not in exclude:
    key_path = path / key
    if not key_path.exists():
        key_path.mkdir()
    writer(key_path)
```

adrianeboyd: Ah, I see what you mean now about the component directories not being created. I don't think that `util.to_disk` should create the subdirectories because it's possible for the path to be used as a file rather than a directory (e.g., `tokenizer`).

pmbaumgartner: Hm, in that case could conditional could be `if key_path.is_dir() and not key_path.exists(): key_path.mkdir()`?

adrianeboyd: That check is a little weird: check if it's a directory AND if it doesn't exist (so who knows whether it's a directory? I think `Path` will always say it's not a directory if it doesn't exist?)

I think just creating the directories in the example components is fine. I think using `util.ensure_path` is also fine. This is not at all unstable. It looks like `ensure_path` hasn't really changed in five years.

pmbaumgartner: I'm at a stalemate with myself here. Sure, we could update the examples to do `ensure_path` and `mkdir` to make these examples work, but it seems so obvious to me that if we can pass a directory for the component to the method, we can also make sure that directory exists so someone could use `to_disk` without adding 3 lines of boilerplate every time.

adrianeboyd: But we're not passing a directory, we're passing a path. If we create this as a directory automatically, it will break the serialization for the tokenizer and other custom components.

Internally we avoid the boilerplate by also using `util.to_disk` in the component serialization methods, which is the other option that could be considered in the examples. It's perhaps a bit more opaque for the users, but I think it would also be fine to use it in the examples.

pmbaumgartner: Got it. Should be good to go now. I added `ensure_path` plus the conditional to create the directory to the examples, and updated the corresponding highlights.

adrianeboyd (approved):

PR #10371, opened 2022-02-24
add additional  REL_OP

ninjalu:
## Description
I have added two additional "REL_OP" for the class DependencyMatcher, to include matching, parent or children to the left or the right of the node.

### Types of change
New feature in "REL_OP"

## Checklist

- [x ] I confirm that I have the right to submit this contribution under the project's MIT license.


spacy/matcher/dependencymatcher.pyx
```diff
@@ -423,6 +427,26 @@ cdef class DependencyMatcher:
...
         return [doc[child.i] for child in doc[node].head.children if child.i < node]
 
+    def _right_child(self, doc, node):
+        return [doc[child.i] for child in doc[node].children if child.i > node]
+    
+    def _left_child(self, doc, node):
+        return [doc[child.i] for child in doc[node].children if child.i < node]
+
+    def _right_parent(self, doc, node):
+        if doc[node].head == doc[node]:
```
polm, line 437: Do you need this condition? 

If the node is the root (= is its own head), then it should fail the position check, since equal values don't match.

spacy/matcher/dependencymatcher.pyx
```diff
@@ -423,6 +427,26 @@ cdef class DependencyMatcher:
...
+            return []
+        elif doc[node].head.i > node:
+            return [doc[node].head]
+        return []
+    
+    def _left_parent(self, doc, node):
+        if doc[node].head == doc[node]:
```
polm, line 444: Same as the other.

polm (changes requested):
Thanks, this looks great! We need to talk internally about whether we want to add more operators, but the implementation here is simple and you've explained how these are useful. 

One thing - could you add some basic tests? You should be able to add some simple ones [here](https://github.com/explosion/spaCy/blob/0c15ab7ca1dbc897e07c038daa4bbc1f2e0a7076/spacy/tests/matcher/test_dependency_matcher.py#L315).

polm:
As a bookkeeping thing, note this started in #10283.

svlandeg:
I'll go ahead and put this in draft for the time being :-)

spacy/matcher/dependencymatcher.pyx
ninjalu: Sorry for getting back to this so late. 

> Do you need this condition?
> 
> If the node is the root (= is its own head), then it should fail the position check, since equal values don't match.

No, you are right! I have made these changes and pushed up to this branch.

To answer your questions regarding tests. I have had a look in the test py file you linked above. Having very little experience in testing, I am feeling nervous about writing a competent test. Maybe you could point to some resources I could have a read?  I'm quite happy to contribute to documentation where needed when I find my way around the documentation.

I have also made slight changes in how these REL_OPs are referred to. I used the way right_child and left_child are referred with "++" and "--" indicating positions. I think it is probably easily to remember?

Thanks!

polm:
Pulling out your general questions here rather than in the in-line comment...

> To answer your questions regarding tests. I have had a look in the test py file you linked above. Having very little experience in testing, I am feeling nervous about writing a competent test. Maybe you could point to some resources I could have a read? I'm quite happy to contribute to documentation where needed when I find my way around the documentation.

No worries, these tests in particular are a little odd. The way the big list of four items work is that each item is:

1. the left word lower-case
2. the right word lower-case
3. the operator (so your new operators)
4. the number of matches if you ran that pattern on the sentence "The quick brown fox jumped over the lazy dog"

If that's not clear let me know.

To run the tests, and for more general guidance on writing tests, see [these docs](https://github.com/explosion/spaCy/blob/master/extra/DEVELOPER_DOCS/Code%20Conventions.md#writing-tests). 

For the docs, you can edit [this file](https://github.com/explosion/spaCy/blob/ed561cf428494c2b7a6790cd4b91b5326102b59d/website/docs/api/dependencymatcher.md). The site is kind of hard to build, so don't worry if you can't get that working or figure out the formatting, we can take care of it.

> I have also made slight changes in how these REL_OPs are referred to. I used the way right_child and left_child are referred with "++" and "--" indicating positions. I think it is probably easily to remember?

I don't use these operators a lot (including the existing ones), so I don't have strong feelings either way, but I think the initial patterns made the relations to existing operators pretty clear, which was nice.

pmbaumgartner:
Hey @ninjalu - thanks again for this PR. I can assist with the tests and docs for this. I've looked at the original discussion and the code you've added here for context.

I've started by adding these operators to the documentation table (which would appear [here](https://spacy.io/api/dependencymatcher) once we add these). Would you review these symbols and descriptions and make sure they're accurate?

| Symbol    | Description                                                                                                          |
| --------- | -------------------------------------------------------------------------------------------------------------------- |
| ... | ... |
| `A >++ B` | `B` is a right child of `A`, i.e. `A` is a parent of `B` and `A.i < B.i` _(not in Semgrex)_.                         |
| `A >-- B` | `B` is a left child of `A`, i.e. `A` is a parent of `B` and `A.i > B.i` _(not in Semgrex)_.                          |
| `A <++ B` | `B` is a right parent of `A`, i.e. `A` is a child of `B` and `A.i < B.i` _(not in Semgrex)_.                         |
| `A <-- B` | `B` is a left parent of `A`, i.e. `A` is a child of `B` and `A.i > B.i` _(not in Semgrex)_.                          |

ninjalu:
Thank you so much. I do apologise: it has been on the back burner for a while. I can confirm the symbols and descriptions are correct.

pmbaumgartner:
@ninjalu No worries! I'm glad you're still interested.

Moving into the tests: we need to define a test for each new operator. Our test sentence is "The quick brown fox jumped over the lazy dog". Given that sentence, can we come up with a test for each operator? If you can give me the right inputs, I can write the actual tests if you'd like. I think of the tests like this - we have three input values: the left word lowercase (`A` in the operators table description), a right word lowercase (`B` in the operators table), and finally one of the four new operators. The output value should be the number of matches we expect given that input and the test sentence. 

e.g. (making this up, this is a nonsensical example)

> **Left**: 'the'
> **Right**: 'fox'
> **Operator**: `<++`
> 
> **Matches**: `2`

If it's any help, here is the dependency parse for our test sentence run through displaCy:

image

ninjalu:
> @ninjalu No worries! I'm glad you're still interested.
> 
> Moving into the tests: we need to define a test for each new operator. Our test sentence is "The quick brown fox jumped over the lazy dog". Given that sentence, can we come up with a test for each operator? If you can give me the right inputs, I can write the actual tests if you'd like. I think of the tests like this - we have three input values: the left word lowercase (`A` in the operators table description), a right word lowercase (`B` in the operators table), and finally one of the four new operators. The output value should be the number of matches we expect given that input and the test sentence.
> 
> e.g. (making this up, this is a nonsensical example)
> 
> > **Left**: 'the'
> > **Right**: 'fox'
> > **Operator**: `<++`
> > **Matches**: `2`
> 
> If it's any help, here is the dependency parse for our test sentence run through displaCy:
> 
> image

Jus to confirm, the test sentence has to be:
"The quick brown fox jumped over the lazy dog"
No the example you gave which is:
"The quick brown fox jumped over the lazy dog"

Is that correct?

pmbaumgartner:
> Jus to confirm, the test sentence has to be:
> "The quick brown fox jumped over the lazy dog"
> Not the example you gave which is:
> "The quick brown fox jumped over the lazy fox"
> 
> Is that correct?

That's right, two uses of `fox`. This got me too as I started exploring the example, I was expecting it to be `dog`. I'm guessing this is because generally we want to have tests for the matcher that would match multiple instances of the same lowercase token.

pmbaumgartner:
@ninjalu - have you had a chance to think about the tests for this following the previous discussion?

ninjalu:
Thanks!! I think you confused me even more there 😅. I asked if the fox and dog was the right test. You said that’s coreect, two uses of fox.

Just to confirm: the test sentence is
“The quick brown fox jumped over the lazy fox”. Is that correct or wrong? 

Lu

pmbaumgartner:
This is correct. `“The quick brown fox jumped over the lazy fox”`

What I need from you is the info for the tests (the input and output values). The input values: the left word lowercase (A in the operators table description), a right word lowercase (B in the operators table), and finally one of the four new operators. The output value should be the number of matches we expect given those input values and the test sentence.

For reference, here's the test structure that I'm looking for https://github.com/ninjalu/spaCy/blob/add_rel_op/spacy/tests/matcher/test_dependency_matcher.py#L277

pmbaumgartner:
@ninjalu Thank you! I'm just coming back from holiday and have checked those tests and they're all passing. We'll look at getting this merged soon.

polm (approved):
Looks good to me! Thanks for all your work on this!

ninjalu:
Awesome! Thanks for your assistance!

PR #10413, opened 2022-03-02
Save span candidates produced by spancat suggesters

thomashacker:
## Description

This `PR` adds a new method to the spancat component: `suggest_candidates`.
The new method uses the suggester of the current spancat component to produce a list of span candidate and add them to the docs.

### Types of change

Enhancement

## Checklist


adrianeboyd:
I can understand why you might want this for debugging purposes, but is there a way to do it without adding an additional option to the component config (which shows up in every single config in the future)? I realize it's a bit complicated due to dependencies on preceding pipeline components, but something similar in spirit to `Tokenizer.explain` that can produce the suggestions for a doc individually?

thomashacker:
Instead of adding a new config parameter, I've added a new function that returns the output of the suggester on a list of docs.

adrianeboyd:
For users who are debugging, I think it might make more sense to process a single doc and provide the results as `List[Span]` instead? (I can be convinced otherwise, but my first thought is that many users would have a hard time making heads or tails of this output.)

thomashacker:
That's a great idea! I'll add this to the method

thomashacker:
I've reimplemented the method so that it receives a list of docs and outputs a list of docs with the candidates added as a new spangroup. This should make the output more user-friendly and also give the possibility to compare suggested spans vs predicted spans right away.

adrianeboyd:
This is shaping up nicely! Let's call the span group option `candidates_key` (or something plural with an underscore to match `spans_key`, and I think calling it `spans_key` is going to be too confusing with the main `spans_key` setting) and make it kwarg-only.

spacy/pipeline/spancat.py
```diff
@@ -272,6 +272,24 @@ def predict(self, docs: Iterable[Doc]):
         scores = self.model.predict((docs, indices))  # type: ignore
         return indices, scores
 
+    def suggest_candidates(
+        self, docs: Iterable[Doc], spankey: str = "candidates"
+    ) -> List[Doc]:
+        """Use the spancat suggester to add a list of span candidates to a list of docs
+
+        docs (Iterable[Doc]): The documents to predict.
```
svlandeg, line 280: Agreed with Adriane's proposal to call the parameter `candidates_key` + add it to the docstring here as well.

spacy/pipeline/spancat.py
```diff
@@ -272,6 +272,24 @@ def predict(self, docs: Iterable[Doc]):
...
+        RETURNS: A list of Docs.
+
+        DOCS: WIP
```
svlandeg, line 283: You can add the documentation in this PR and label it as being new since 3.3, then update the link here :-)

spacy/pipeline/spancat.py
```diff
@@ -272,6 +272,24 @@ def predict(self, docs: Iterable[Doc]):
         scores = self.model.predict((docs, indices))  # type: ignore
         return indices, scores
 
+    def suggest_candidates(
```
svlandeg, line 275: Should we call this `store_candidates` instead? The current naming sounds more like you would run `self.suggester` and return the results?

spacy/tests/pipeline/test_spancat.py
```diff
@@ -397,3 +397,21 @@ def zero_suggester(docs, *, ops=None):
...
+        "Just a sentence.",
+        "I like London and Berlin",
+        "I like Berlin",
+        "I eat ham.",
+    ]
+    docs = [nlp(text) for text in texts]
+    output = spancat.suggest_candidates(docs)
+
+    assert len(output) == len(texts)
+    assert type(output[0].spans["candidates"]) == SpanGroup
```
svlandeg, line 417: You could probably also `assert` the actual contents here, using the default suggester?

svlandeg (commented):
Great work Edi, I think it makes a lot of sense to provide this functionality and have it in a way that is directly comparable to the final list of spans. I just had some small comments/questions that I wanted to run by you and Adriane.

spacy/pipeline/spancat.py
```diff
@@ -272,6 +272,24 @@ def predict(self, docs: Iterable[Doc]):
...
+
+        DOCS: WIP
+        """
+        suggester_output = self.suggester(docs, ops=self.model.ops)
+
+        for candidates, doc in zip(suggester_output, docs):
+            doc.spans[spankey] = []
+            for index in candidates.dataXd:
+                doc.spans[spankey].append(doc[index[0] : index[1]])
+        return docs
```
svlandeg, line 291: `mypy` complains here because you're not returning a `List` as the typing indicates. You could either retype this to return `Iterable[Doc]` which is the correct type of `docs`, or you could return `list(docs)` instead.

spacy/pipeline/spancat.py
```diff
@@ -272,6 +272,24 @@ def predict(self, docs: Iterable[Doc]):
...
+        """Use the spancat suggester to add a list of span candidates to a list of docs
+
+        docs (Iterable[Doc]): The documents to predict.
+        RETURNS: A list of Docs.
+
+        DOCS: WIP
+        """
+        suggester_output = self.suggester(docs, ops=self.model.ops)
+
+        for candidates, doc in zip(suggester_output, docs):
```
svlandeg, line 287: `mypy` complains about `zip` not being able to work with `(Ragged, Iterable[Doc])`. I'm not sure whether we can find a quick elegant solution to this - if not you can also just add the directive `  # type: ignore` to the end of this line.

spacy/pipeline/spancat.py
adrianeboyd: Do you need to return the docs? `set_annotations` doesn't return the docs.

svlandeg: You're right, we might as well just not return the docs, given that it's all done in-place. That also resolves that ambiguity for the user.

spacy/pipeline/spancat.py
thomashacker: I added the directive for now :smile:

spacy/pipeline/spancat.py
thomashacker: How can I label the method? Is there an example that does something like this? :smile:

thomashacker: Ah found it!

thomashacker:
I hope I've managed to implement all changes, I also added the changes in the spancat doc with the `new in 3.3 label` :sparkles:

adrianeboyd:
The only remaining thing on my end: I think the docstring and docs should clarify that this is for debugging purposes, otherwise people might expect that it's part of the normal processing somewhere and be surprised when it looks unused?

svlandeg:
Is it only for debugging though? I can think of use-cases where having all the suggestions make sense. e.g. if you're suggesting all dates, but you want the spancat to only label some of them, at the end you might just want to have all those dates available in post-processing.

svlandeg:
Could we put something like "(not enabled by default)"?

adrianeboyd:
It's not like there's a way to enable it, though?

thomashacker:
We can maybe clarify that it's not used by default when running the pipeline and instead has to be run separately. We can also say that it's intended for debugging and analyzing the current used suggester function.

svlandeg:
The more I think about this, the more I actually feel like it was the better approach to have this in the config, as I think there are valid use-cases to this beyond debugging.

adrianeboyd:
I guess I'm still having trouble understanding why this justifies additional config settings. You could also do something similar by setting the threshold to 0?

svlandeg:
That's true Adriane. I'd be fine with merging this as is.

website/docs/api/spancategorizer.md
```diff
@@ -239,6 +239,22 @@ Delegates to [`predict`](/api/spancategorizer#predict) and
...
+>
+> ```python
+> spancat = nlp.add_pipe("spancat")
+> spancat.set_candidates(docs, "candidates")
+> ```
+
+| Name     | Description                                               |
+| -------- | --------------------------------------------------------- |
+| `docs`   | The documents to modify. ~~Iterable[Doc]~~                |
+| `candidates_key` | Key of the Doc.spans dict to save the candidate spans under. ~~str~~ |
```
svlandeg, line 256: Ah this needs another round of `prettier` ;-)

PR #10451, opened 2022-03-07
Fix get_matching_ents

polm:
## Description

`Example.get_matching_ents` was added as part of  #9669, but for some reason master seems to have an old version of the code that just doesn't work. It's already covered by tests that were in #9669 (and fail without this commit) but were succeeding there for some reason. 

### Types of change

Fix

## Checklist


svlandeg (approved):
Let's merge this in as soon as it goes green!

PR #10452, opened 2022-03-07
Add tokenizer option to allow Matcher handling for all rules

adrianeboyd:
## Description

Add tokenizer option `with_faster_rules_heuristics` that determines whether the special cases applied by the internal `Matcher` are filtered by whether they contain affixes or space. If `True` (default), the rules are filtered to prioritize speed over rare edge cases. If `False`, all rules are included in the final `Matcher`-based pass over the doc.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
Something is going very wrong in the new test in the first python 3.6 test, but I am running it again to see if it's consistently very wrong.

svlandeg:
Oof, that is one scary looking failing test :|

adrianeboyd:
I can reproduce it locally, too. The last commit is not going to fix it.

website/docs/api/tokenizer.md
```diff
@@ -44,15 +44,16 @@ how to construct a custom tokenizer with different tokenization rules, see the
...
+| `suffix_search`                | A function matching the signature of `re.compile(string).search` to match suffixes. ~~Optional[Callable[[str], Optional[Match]]]~~                                            |
+| `infix_finditer`               | A function matching the signature of `re.compile(string).finditer` to find infixes. ~~Optional[Callable[[str], Iterator[Match]]]~~                                            |
+| `token_match`                  | A function matching the signature of `re.compile(string).match` to find token matches. ~~Optional[Callable[[str], Optional[Match]]]~~                                         |
+| `url_match`                    | A function matching the signature of `re.compile(string).match` to find token matches after considering prefixes and suffixes. ~~Optional[Callable[[str], Optional[Match]]]~~ |
+| `with_faster_rules_heuristics` | Whether to restrict the final `Matcher`-based pass for rules to those containing affixes or space. Defaults to `True`. ~~bool~~                                               |
```
svlandeg, line 56: Add that this is new in 3.3 ?

adrianeboyd:
I'm still not entirely sure about the naming here...

svlandeg:
> I'm still not entirely sure about the naming here...

If we want to have something shorter, I think either `with` or maybe even `rules` could be removed from the name?

adrianeboyd:
Yeah, I think `faster_heuristics` would be okay. Matt also suggested `prioritize_speed` or something along those lines.

svlandeg:
Let's settle on `faster_heuristics`?

honnibal:
Agree that the name is the least worst option 🙂 👍

PR #10457, opened 2022-03-08
Fix Entity Linker with tokenization mismatches (fix #9575)

### LINKED ISSUE #9575: EntityLinker encounters IndexError if token length of Example predicted and reference Docs aren't equal

polm:
## Description

As outlined in #9575, the Entity Linker has issues when tokenization differs between gold and predicted Docs. This was separate from the issue addressed with the recent fix for tok2vec issues, so this PR should address it. 

The main issue is that while KB IDs are taken aligned to tokens in the predicted Doc, they are indexed into using entity token indices from the gold Doc. This means if there was any change in the tokenization before the token in question, the tokens won't line up, which can result in nonsense or IndexErrors. 

Right now this PR just has a failing test; the fix is in progress. 

My basic strategy for fixing this is to use the `get_matching_ents` function introduced in the previous NEL fix PR to get aligned entities. That should give correct results all cases,  but may be more strict than the previous behavior, since previously KB IDs were taken from entities with a matching first token rather than matching the whole span.

### Types of change

bugfix

## Checklist

### DIFF
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -363,8 +363,7 @@ class EntityLinker(TrainablePipe):
             out = self.model.ops.alloc2f(*sentence_encodings.shape)
             return 0, out
 
-        if (selected_encodings.shape != entity_encodings.shape and
-                selected_encodings.size > 0 and entity_encodings.size > 0):
+        if selected_encodings.shape != entity_encodings.shape:
             err = Errors.E147.format(
                 method="get_loss", msg="gold entities do not match up"
             )
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -234,10 +234,10 @@ class EntityLinker(TrainablePipe):
         nO = self.kb.entity_vector_length
         doc_sample = []
         vector_sample = []
-        for example in islice(get_examples(), 10):
-            doc = example.x
+        for eg in islice(get_examples(), 10):
+            doc = eg.x
             if self.use_gold_ents:
-                ents, _ = example.get_aligned_ents_and_ner()
+                ents, _ = eg.get_aligned_ents_and_ner()
                 doc.ents = ents
             doc_sample.append(doc)
             vector_sample.append(self.model.ops.alloc1f(nO))
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -358,6 +358,11 @@ class EntityLinker(TrainablePipe):
         entity_encodings = self.model.ops.asarray(entity_encodings, dtype="float32")
         selected_encodings = sentence_encodings[keep_ents]
 
+        # if there are no matches, short circuit
+        if not keep_ents:
+            out = self.model.ops.alloc2f(*sentence_encodings.shape)
+            return 0, out
+
         if (selected_encodings.shape != entity_encodings.shape and
                 selected_encodings.size > 0 and entity_encodings.size > 0):
             err = Errors.E147.format(
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -237,7 +237,8 @@ class EntityLinker(TrainablePipe):
         for example in islice(get_examples(), 10):
             doc = example.x
             if self.use_gold_ents:
-                doc.ents = example.y.ents
+                ents, _ = example.get_aligned_ents_and_ner()
+                doc.ents = ents
             doc_sample.append(doc)
             vector_sample.append(self.model.ops.alloc1f(nO))
         assert len(doc_sample) > 0, Errors.E923.format(name=self.name)
@@ -312,7 +313,8 @@ class EntityLinker(TrainablePipe):
 
         for doc, ex in zip(docs, examples):
             if self.use_gold_ents:
-                doc.ents = ex.reference.ents
+                ents, _ = ex.get_aligned_ents_and_ner()
+                doc.ents = ents
             else:
                 # only keep matching ents
                 doc.ents = ex.get_matching_ents()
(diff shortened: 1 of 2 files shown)


polm:
Tests are green now, but leaving as draft until I test it with the sample projects.

polm:
Sample project trained fine so this should be ready for review. The changes ended up being pretty limited.

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -1071,3 +1071,32 @@ def create_kb(vocab):
 
     # this will run the pipeline on the examples and shouldn't crash
     results = nlp.evaluate(train_examples)
+
+def test_tokenization_mismatch():
+    # https://github.com/explosion/spaCy/issues/9575
```
svlandeg, line 1076:
```suggestion
@pytest.mark.issue(9575)
def test_tokenization_mismatch():
```

spacy/pipeline/entity_linker.py
```diff
@@ -345,7 +347,7 @@ def get_loss(self, examples: Iterable[Example], sentence_encodings: Floats2d):
         for eg in examples:
             kb_ids = eg.get_aligned("ENT_KB_ID", as_string=True)
 
-            for ent in eg.reference.ents:
+            for ent in eg.get_matching_ents():
```
svlandeg, line 350: This is not related to this PR, but perhaps we could standardize on 
> for XXX in `examples`

and pick either `example`, `ex` or `eg`. I think most commonly across the repo, we'd use `eg`.

It just caught my attention because the first three edits in this PR use 3 different var names ;-)

svlandeg (commented):
Nice digging and nice unit test! Apart from some small nitpicking, this looks good to me :-)

spacy/pipeline/entity_linker.py
polm: That's a good point. I'm not sure but I think:

- most existing code uses `eg`
- `example` is used in a specific context (maybe before me?)
- I tend to use `ex`

I would be fine with standardizing on `eg`, which is probably more common in existing code.

polm: I standardized on `eg` here.

PR #10471, opened 2022-03-09
small cleanup

Ben-Epstein:
this isn't used until the exception block, so just a bit cleaner to move it there

## Checklist


svlandeg (commented):
The original implementation was to avoid calling `self.get_error_handler()` multiple times within the loop - e.g. when the error handlers is just throwing warnings or printing logs, instead of actually raising an error (it could be anything).

I don't think it's a huge deal either way, but I don't really see the need to make the change here.

adrianeboyd:
Thanks for taking a look, I agree with Sofie that we don't need to make this change here.

PR #10485, opened 2022-03-14
Add a note to the dev docs on mypy

ljvmiranda921:
Mention that we should use mypy when developing and add a few notes on typing (context: from Ora)

### Types of change

Documentation update

## Checklist


svlandeg (approved):
Looks good, thanks!

PR #10487, opened 2022-03-14
Various install docs updates

adrianeboyd:
## Description

Various install docs updates

* Use `pip` and editable install in quickstart instructions:
  * Avoid compiling twice
  * Switch to editable install
* Update pytorch example to more recent versions

In general we should try to move away from `python setup.py` commands, however I am entirely not sure whether we want `--editable` by default in these instructions.

## Checklist

### DIFF
--- a/website/docs/usage/embeddings-transformers.md
+++ b/website/docs/usage/embeddings-transformers.md
@@ -211,23 +211,23 @@ PyTorch as a dependency below, but it may not find the best version for your
 setup.
 
 ```bash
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I think overall people will be more confused if the source install isn't editable, but there are definitely cases where you would want a standard install instead of the editable install.

svlandeg:
Hm. I can see it either way, but I do think the use-case for having it editable will be more frequent?

adrianeboyd:
Yeah, I think so, too. And it's easier for people to see and remove an option than to figure out which one to add to get the editable install.

PR #10502, opened 2022-03-16
Add debug diff command in spaCy CLI

ljvmiranda921:
This PR attempts to add a `diff` subcommand under the `debug` command in spaCy CLI. 
The goal is for (1) users to cross-check their config with respect to some defaults and (2) to help maintainers debug what's wrong in someone's config while in the Discussion forum.

~~**⚠️⚠️ This is dependent upon this PR: https://github.com/ines/wasabi/pull/20**. Ideally, we'd want to review and merge that PR first, then make a new release for wasabi, update spaCy's dependencies with the new wasabi version, etc.~~ Done!

### Sample usage

Users only need to supply their configuration, and it will be compared against the default config in spaCy.

```
spacy debug diff path/to/my_config.cfg  # compares your config wrt the default spaCy installed
spacy debug diff path/to/my_config.cfg --compare-to path/to/another_config.cfg  # compares your config wrt to another cfg file
```

If users have special options when they generated the config (e.g., pretraining, gpu, etc.), they should provide it as a CLI option

```
spacy debug diff path/to/my_config.cfg --pretraining
spacy debug diff path/to/my_config.cfg --optimize accuracy
```

- **Why under the debug subcommand?** I imagine it being used to double-check what has changed in the config since then. Although I don't mind putting it in a different place 👍 
- **What is the "default config" that we compare the user's config against?** Currently we "generate" the default using the `init_config` function based on some parameters found in the user-provided configuration (e.g. `lang`, which pipelines are included, etc.). However, some options like `pretraining`, optimize for `efficiency/accuracy`, and `gpu` must still be supplied manually, and this PR provides CLI options for that.
- **What are possible error scenarios?** Let's say:
    1. User generates a config while on spaCy version 1.
    2. We released a new spaCy version 2 with different defaults / config structure (or more specifically, `init_config` has changed)
    3. User installs new spaCy version 2 and used `debug diff`
    4. The diff will now be large because there's a mismatch between two spaCy config versions.

Although personally, I find the error condition above to be an acceptable behaviour to help pinpoint any breaking changes.

### Types of change

New feature + docs 

## Checklist


ljvmiranda921:
Thinking of the best way to test this functionality 🤔

adrianeboyd:
I think you can go ahead and make the `wasabi` requirements change as part of this PR.

ljvmiranda921:
Done updating the dependencies!

ljvmiranda921:
Updated the dependencies! Just running the tests now...

spacy/cli/debug_diff.py
```diff
@@ -0,0 +1,87 @@
...
+    user = user_config.to_str()
+    other = other_config.to_str()
+
+    if user == other:
+        msg.warn("No diff to show: configs are identical")
+    else:
+        diff_text = diff_strings(other, user, add_symbols=True)
+        if markdown:
+            md = MarkdownRenderer()
+            md.add(md.title("Config Diff"))
```
ines, line 83: The first argument here needs to be the level of the headline – at the moment, running it with `--markdown` raises an error for me?

I'd be happy to just leave it out entirely because I'm not sure it needs a headline? (I we really want to, it should probably level `2`, i.e. `md.title(2. "...")`, since that's what we use in `spacy info`. But when again, I don't think it's needed.)

```suggestion
```

spacy/cli/debug_diff.py
```diff
@@ -0,0 +1,87 @@
...
+                silent=True,
+            )
+
+    user = user_config.to_str()
+    other = other_config.to_str()
+
+    if user == other:
+        msg.warn("No diff to show: configs are identical")
+    else:
+        diff_text = diff_strings(other, user, add_symbols=True)
```
ines, line 80: Mostly a question: Do we want to add the symbols if markdown is disabled? Is there an advantage? IMO, if you just want to view it in your terminal, the colours should be fine (unless there's an edge case I'm missing – if it's colours being unavailable in the terminal, can we handle this via `wasabi`?)-

```suggestion
        diff_text = diff_strings(other, user, add_symbols=markdown)
```

website/docs/api/cli.md
```diff
@@ -626,6 +626,237 @@ will not be available.
...
+$ python -m spacy debug diff [config_path] [--compare-to] [--optimize] [--gpu] [--pretraining] [--markdown]
+```
+
+> #### Example
+>
+> ```cli
+> $ python -m spacy debug diff ./config.cfg
+> ```
+
+<Accordion title="Example output" spaced>
```
ines, line 647: This is really nice, docs-wise! 💯

website/docs/api/cli.md
```diff
@@ -626,6 +626,237 @@ will not be available.
...
+| `gpu`, `-G`          | Whether the config was made to run on a GPU. ~~bool (flag)~~                                                                                                                                                                                       |
+| `pretraining`, `-pt` | Include config for pretraining (with [`spacy pretrain`](/api/cli#pretrain)). Defaults to `False`. ~~bool (flag)~~                                                                                                                                  |
+| `markdown`, `-md` | Generate Markdown for Github issues. Defaults to `False`. ~~bool (flag)~~                                                                                                                                  |
+| **PRINTS**           | Diff between the two config files                                                                                                                                                                                                                  |
```
ines, line 856: Haha, sorry, not review without ultimate nitpicks 😅

```suggestion
| `config_path`        | Path to [training config](/api/data-formats#config) file containing all settings and hyperparameters. ~~Union[Path, str] \(positional)~~                                                                                                            |
| `compare_to`         | Path to another config file to diff against. ~~Optional[Union[Path, str] \(option)~~                                                                                                                                                                |
| `optimize`, `-o`     | `"efficiency"` or `"accuracy"`. Whether the config was optimized for efficiency (faster inference, smaller model, lower memory consumption) or higher accuracy (potentially larger and slower model). Defaults to `"efficiency"`. ~~str (option)~~ |
| `gpu`, `-G`          | Whether the config was made to run on a GPU. ~~bool (flag)~~                                                                                                                                                                                       |
| `pretraining`, `-pt` | Include config for pretraining (with [`spacy pretrain`](/api/cli#pretrain)). Defaults to `False`. ~~bool (flag)~~                                                                                                                                  |
| `markdown`, `-md` | Generate Markdown for Github issues. Defaults to `False`. ~~bool (flag)~~                                                                                                                                  |
| **PRINTS**           | Diff between the two config files.                                                                                                                                                                                                                  |
```

spacy/cli/debug_diff.py
ljvmiranda921: Oh, thanks for catching! I don't mind removing this line as well 👍

website/docs/api/cli.md
ljvmiranda921: Thank you! 🙇

spacy/cli/debug_diff.py
```diff
@@ -0,0 +1,86 @@
...
+from pathlib import Path
+from thinc.api import Config
+
+from ._util import debug_cli, Arg, Opt, show_validation_error, parse_config_overrides
+from ..util import load_config
+from .init_config import init_config, Optimizations
+
+
+@debug_cli.command(
+    "diff",
```
svlandeg, line 14: I wonder whether we should call this `diff-config`, similar to how the `init` subcommand is called `fill-config`? Would love to hear other's opinions on this first though.

website/docs/api/cli.md
```diff
@@ -626,6 +626,237 @@ will not be available.
...
+| `optimize`, `-o`     | `"efficiency"` or `"accuracy"`. Whether the config was optimized for efficiency (faster inference, smaller model, lower memory consumption) or higher accuracy (potentially larger and slower model). Defaults to `"efficiency"`. ~~str (option)~~ |
+| `gpu`, `-G`          | Whether the config was made to run on a GPU. ~~bool (flag)~~                                                                                                                                                                                       |
+| `pretraining`, `-pt` | Include config for pretraining (with [`spacy pretrain`](/api/cli#pretrain)). Defaults to `False`. ~~bool (flag)~~                                                                                                                                  |
+| `markdown`, `-md` | Generate Markdown for Github issues. Defaults to `False`. ~~bool (flag)~~                                                                                                                                  |
```
svlandeg, line 855: Looks like this will need another run of `prettier` :-)

spacy/cli/debug_diff.py
```diff
@@ -0,0 +1,86 @@
...
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
+def debug_diff_cli(
+    # fmt: off
+    ctx: typer.Context,
+    config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
+    compare_to: Optional[Path] = Opt(None, help="Path to another config file to diff against", exists=True, allow_dash=True),
+    optimize: Optimizations = Opt(Optimizations.efficiency.value, "--optimize", "-o", help="Whether the user config was optimized for efficiency or accuracy."),
+    gpu: bool = Opt(False, "--gpu", "-G", help="Whether the original config can run on a GPU"),
+    pretraining: bool = Opt(False, "--pretraining", "--pt", help="Whether to compare on a config with pretraining involved"),
```
svlandeg, line 24: The `optimize`, `gpu` and `pretraining` options are only relevant when `compare_to` is `None`, right? So maybe we should document more clearly that these only affect the "default" config that would be compared to (both here as well as in the docs).

spacy/cli/debug_diff.py
```diff
@@ -0,0 +1,86 @@
...
+    markdown: bool = Opt(False, "--markdown", "-md", help="Generate Markdown for GitHub issues")
+    # fmt: on
+):
+    """Show a diff of a config file with respect to spaCy's defaults. If
```
svlandeg, line 28:
```suggestion
    """Show a diff of a config file with respect to spaCy's defaults or another config file. If
```
(+ reformatting)

spacy/cli/debug_diff.py
```diff
@@ -0,0 +1,86 @@
...
+    additional settings were used in the creation of the config file, then you
+    must supply these as extra parameters to the command. The generated diff
```
svlandeg, line 30:
```suggestion
    must supply these as extra parameters to the command when comparing to the default settings. The generated diff
```

spacy/cli/debug_diff.py
```diff
@@ -0,0 +1,86 @@
...
+
+@debug_cli.command(
+    "diff",
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
+def debug_diff_cli(
+    # fmt: off
+    ctx: typer.Context,
+    config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
+    compare_to: Optional[Path] = Opt(None, help="Path to another config file to diff against", exists=True, allow_dash=True),
```
svlandeg, line 21:
```suggestion
    compare_to: Optional[Path] = Opt(None, help="Path to a config file to diff against, or `None` to compare against default settings", exists=True, allow_dash=True),
```

website/docs/api/cli.md
```diff
@@ -626,6 +626,237 @@ will not be available.
 | overrides                  | Config parameters to override. Should be options starting with `--` that correspond to the config section and value to override, e.g. `--paths.train ./train.spacy`. ~~Any (option/flag)~~                         |
 | **PRINTS**                 | Debugging information.                                                                                                                                                                                             |
 
+### debug diff {#debug-diff tag="command"}
+
+Show a diff of a config file with respect to spaCy's defaults. If
+additional settings were used in the creation of the config file, then you
+must supply these as extra parameters to the command. The generated diff
+can also be used when posting to the discussion forum to provide more
+information for the maintainers.
```
svlandeg, line 635:
```suggestion
Show a diff of a config file with respect to spaCy's defaults or another config file. If
additional settings were used in the creation of the config file, then you
must supply these as extra parameters to the command when comparing to the default settings. The generated diff
can also be used when posting to the discussion forum to provide more
information for the maintainers.
```
+ reformatting

website/docs/api/cli.md
```diff
@@ -626,6 +626,237 @@ will not be available.
...
+
+[initialize.tokenizer]
+```
+</Accordion>
+
+
+| Name                 | Description                                                                                                                                                                                                                                        |
+|----------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `config_path`        | Path to [training config](/api/data-formats#config) file containing all settings and hyperparameters. ~~Union[Path, str] \(positional)~~                                                                                                            |
+| `compare_to`         | Path to another config file to diff against. ~~Optional[Union[Path, str] \(option)~~                                                                                                                                                                |
```
svlandeg, line 851:
```suggestion
| `compare_to`         | Path to another config file to diff against, or `None` to compare against default settings. ~~Optional[Union[Path, str] \(option)~~                                                                                                                                                                |
```

svlandeg (commented):
This will be a great tool to have! And the implementation looks good to me, just had a few comments around documentation.

spacy/cli/debug_diff.py
adrianeboyd: I also share some confusion about what `debug diff` might be diff-ing. `diff-config` doesn't sound too bad, although I wonder a teensy bit whether `diff` is understandable enough. `compare-config`? Hmm...

ljvmiranda921: Ah, I think `diff-config` sounds good! Just to parallel `fill-config`. Going back I realized that `diff` may sound too general and it can mean diffing other things. `diff-config` sounds more explicit. Can change that 👍

website/docs/api/cli.md
ljvmiranda921: Ok will update this! :D

spacy/cli/debug_diff.py
ljvmiranda921: Updated here: https://github.com/explosion/spaCy/pull/10502/commits/79d0aa526770298777192e7e492e877042cf6fe6

spacy/cli/debug_diff.py
ljvmiranda921: UPdated here: https://github.com/explosion/spaCy/pull/10502/commits/ffdb3a017d99bc2e219a088186eea8d0f3c5265f

svlandeg (approved):
Looks great!

PR #10509, opened 2022-03-17
Add support for numbers with percentages

gremur:
Add support for numbers with percentages in the `like_num` function

### Types of change

an enhancement

## Checklist


adrianeboyd:
Hmm, this could apply if you have a custom tokenizer, but the English tokenizer splits `%` from the number by default, so this isn't going to match anything for `spacy.blank("en")`.

PR #10522, opened 2022-03-18
Complete doc about accessing context in pipeline components

percevalw:
Accessing the contexts given as tuples to `nlp.pipe([(text1, context1), (text2, context2)])` from inside
a component is possible using the `doc._context` attribute but this feature is not documented.

For example:

```
import spacy

@Language.component("info_component")
def print_context(doc):
    print(f"Doc context", doc._context)
    return doc

text_tuples = [
    ("This is the first text.", {"text_id": "text1"}),
    ("This is the second text.", {"text_id": "text2"})
]

nlp.add_pipe("print_context", name="print_context")
doc_tuples = nlp.pipe(text_tuples, as_tuples=True)

docs = list(nlp.pipe(doc_tuples))
```

## Description

Document the `doc._context` feature.

### Types of change
Change to the documentation

## Checklist

### DIFF
--- a/website/docs/usage/processing-pipelines.md
+++ b/website/docs/usage/processing-pipelines.md
@@ -95,7 +95,8 @@ You can use the `as_tuples` option to pass additional context along with each
 doc when using [`nlp.pipe`](/api/language#pipe). If `as_tuples` is `True`, then
 the input should be a sequence of `(text, context)` tuples and the output will
 be a sequence of `(doc, context)` tuples. For example, you can pass metadata in
-the context and save it in a [custom attribute](#custom-components-attributes):
+the context and save it in a [custom attribute](#custom-components-attributes)
+or retrieve it from a custom component using the `doc._context` attribute:
 
 ```python
 ### {executable="true"}


adrianeboyd:
Thanks for the PR, but this is considered an internal/private implementation detail (that also might change in the future), so we don't want to include it in the docs.

Instead, you can explicitly set custom extensions to save the same data and access them normally in your components, e.g. as `doc._.my_additional_context` and then you don't need to use `as_tuples` at all.

To support multiprocessing, you do need to re-add any extensions on `__call__` and `pipe` to be sure that they're accessible in the child processes. See `spacy-transformers` for an example (although I don't recommend copying the global variable part of that in particular, just the part that sets any missing extensions):

https://github.com/explosion/spacy-transformers/blob/cfd9955a6ed6e2b057f3e759e253d6f31fb205a3/spacy_transformers/pipeline_component.py#L181-L194

percevalw:
Thank you for these details.

> Instead, you can explicitly set custom extensions to save the same data and access them normally in your components, e.g. as doc._.my_additional_context and then you don't need to use as_tuples at all.

I disagree on this point. Once the `nlp(text)` or `nlp(text_and_context)` has been called, accessing the context from inside the pipeline can only be done using the `._context` attribute, as done in my `print_context` example. I understand however that this may currently be a private implementation detail. Making such an attribute part of the API would be really useful.

adrianeboyd:
In practice it looks like this: https://spacy.io/usage/v3-2#doc-input. Or with `nlp.pipe(docs)`.

PR #10534, opened 2022-03-22
displacy: avoid overlapping arcs in manual mode

### LINKED ISSUE #5447: displacy issues when representing overlapping arcs
Hi, I'm using displacy's manual mode to render weigthed depedency graphs (that is, custom labels and, possibly, loops). In the case I get a two-node loop, labels get overwritten and the arc gets overwritten as well. See output below: 

displacy

Is there any way to offset labels so that I can avoid overlap, or anything similar to use as a workaround? 

## Info about spaCy

* **spaCy version:** 2.2.4
* **Platform:** Darwin-19.4.0-x86_64-i386-64bit
* **Python version:** 3.6.10

Best regards,

jfainberg:
displacy: avoid overlapping arcs in manual mode

Resolves #5447

## Description
Two arcs with the same start and end index will currently display as overlapped. I think this will only occur in manual mode (see issue above). The fix keys the levels assigned to the arcs by `(start, end, label)` (instead of `end-start`). Each arc is compared to the current levels across its span, stored in the list `max_level`.

(I was in two minds about submitting this PR - it takes a nice one-liner and turns it into a chunk of code for a fringe use-case... Happy for ideas for improvements or to drop it entirely 😄).

### Types of change
This PR is a bug fix for #5447.

## Checklist

### DIFF
--- a/spacy/displacy/render.py
+++ b/spacy/displacy/render.py
@@ -270,7 +270,7 @@ class DependencyRenderer:
         RETURNS (str): Rendered SVG markup.
         """
         self.levels = self.get_levels(arcs)
-        self.highest_level = len(self.levels)
+        self.highest_level = max(self.levels.values(), default=0)
         self.offset_y = self.distance / 2 * self.highest_level + self.arrow_stroke
         self.width = self.offset_x + len(words) * self.distance
         self.height = self.offset_y + 3 * self.word_spacing
@@ -330,7 +330,7 @@ class DependencyRenderer:
         if start < 0 or end < 0:
             error_args = dict(start=start, end=end, label=label, dir=direction)
             raise ValueError(Errors.E157.format(**error_args))
-        level = self.levels.index(end - start) + 1
+        level = self.levels[(start, end, label)]
         x_start = self.offset_x + start * self.distance + self.arrow_spacing
         if self.direction == "rtl":
             x_start = self.width - x_start
@@ -346,7 +346,7 @@ class DependencyRenderer:
         y_curve = self.offset_y - level * self.distance / 2
         if self.compact:
             y_curve = self.offset_y - level * self.distance / 6
-        if y_curve == 0 and len(self.levels) > 5:
+        if y_curve == 0 and max(self.levels.values(), default=0) > 5:
             y_curve = -self.distance
         arrowhead = self.get_arrowhead(direction, x_start, y, x_end)
         arc = self.get_arc(x_start, y, y_curve, x_end)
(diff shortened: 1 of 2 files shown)


spacy/displacy/render.py
```diff
@@ -390,15 +390,26 @@ def get_arrowhead(self, direction: str, x: int, y: int, end: int) -> str:
...
+        levels = {}
+
+        for arc in sorted(arcs, key=lambda arc: arc["end"] - arc["start"]):
+            level = max(max_level[arc["start"] : arc["end"]]) + 1
+
+            for i in range(arc["start"], arc["end"]):
+                max_level[i] = level
+
+            levels[(arc["start"], arc["end"], arc["label"])] = level
+
```
ines, line 411: Stylistic nitpick 😅

```suggestion
        for arc in sorted(arcs, key=lambda arc: arc["end"] - arc["start"]):
            level = max(max_level[arc["start"] : arc["end"]]) + 1
            for i in range(arc["start"], arc["end"]):
                max_level[i] = level
            levels[(arc["start"], arc["end"], arc["label"])] = level
```

ines (commented):
Thanks for the PR and for the test! The change makes sense to me. I think the reason we ended up with this implementation is that the dependency renderer was originally designed for standard dependency parses where you don't end up with duplicate args. So I never really considered cases like this.

(I only tested it manually so far and the change seems to be fine and my test examples all rendered correctly. We don't currently have any tests that check if the visualization is rendered correctly because maintaining those is a pain if you're diffing markup and things change. But I'm open to ideas here if you have any!)

> (I was in two minds about submitting this PR - it takes a nice one-liner and turns it into a chunk of code for a fringe use-case... Happy for ideas for improvements or to drop it entirely 😄).

Yeah, I totally know what you mean 😅 But I do think it's worth it because the fix wil also make it easier in the future to adapt displaCy to render more general relation extraction visualizations, which could be pretty cool.

jfainberg:
Very happy to make my first contribution! 😃 

> (I only tested it manually so far and the change seems to be fine and my test examples all rendered correctly. We don't currently have any tests that check if the visualization is rendered correctly because maintaining those is a pain if you're diffing markup and things change. But I'm open to ideas here if you have any!)

Yeah, I tried to do something like this in the unit test, but it got messy fast 😕, so I settled for just checking `renderer.highest_level` which is somewhat indirect...

The only example I've found that maybe doesn't render correctly is if you add duplicate arcs with identical labels, `(start, end, label)` - is this a valid use case? E.g. doing this:

```
{"start": 1, "end": 3, "label": "attr", "dir": "right"},
{"start": 1, "end": 3, "label": "attr", "dir": "right"},
```

will raise both arcs one level, while this:
```
{"start": 1, "end": 3, "label": "attr", "dir": "right"},
{"start": 1, "end": 3, "label": "attr 2", "dir": "right"},
```

will render correctly.

ines:
Yay, thanks again for the very thoughtful contribution 🎉

> The only example I've found that maybe doesn't render correctly is if you add duplicate arcs with identical labels, (start, end, label) - is this a valid use case?

That's a good point but yeah, I don't consider this a valid use case – it's certainly not something the parser should ever produce (and it's also not something I'd expect a relation extraction model to precit). If we wanted to be extra sure here, we could filter out _exact_ duplicates (`(start, end ,label)`) before rendering. I'll go ahead and approve this PR because I want it included in the next release, but this could be a nice and simple follow-up PR!

ines (approved):

PR #10537, opened 2022-03-22
matcher: remove an undefined behavior

danieldk:
## Description

Indexing into a zero-length `std::vector` is an undefined behavior. This change first checks that the vector is not empty before retrieving the (address of) the first element.

## Checklist


svlandeg (approved):
Nice catch!

PR #10538, opened 2022-03-22
Maintain support for empty DocBin span groups

adrianeboyd:
## Description

Maintain support for empty DocBin span groups

(I suspect this only affects `DocBin`s created with spaCy v2.x or pre-release versions of v3, but it's easy enough to avoid breaking the behavior here.)

### Types of change

Bug fix.

## Checklist


svlandeg (approved):
I think there was only a problem with compatibility to beta releases of v3, but you're right that it doesn't hurt to fix :-)

PR #10547, opened 2022-03-24
Preparing span predictor for predicting from gold

kadarakos:
# Preparing span predictor for predicting from gold 

## Description
This PR is midway debugging the `SpanPredictor` and preparing it for training from gold-spans only.

### Types of change
Debugging experimental code.

### DIFF
--- a/spacy/ml/models/coref.py
+++ b/spacy/ml/models/coref.py
@@ -91,7 +91,7 @@ def build_span_predictor(
         # TODO fix device - should be automatic
         device = "cuda:0"
         span_predictor = PyTorchWrapper(
-            SpanPredictor(hidden_size, dist_emb_size, device),
+            SpanPredictor(dim, dist_emb_size, device),
             convert_inputs=convert_span_predictor_inputs
         )
         # TODO use proper parameter for prefix
@@ -148,7 +148,6 @@ def convert_span_predictor_inputs(
     # Normally we shoudl use the input is_train, but for these two it's not relevant
     sent_ids = xp2torch(sent_ids[0], requires_grad=False)
     head_ids = xp2torch(head_ids[0], requires_grad=False)
-
     word_features = xp2torch(tok2vec[0], requires_grad=is_train)
 
     argskwargs = ArgsKwargs(args=(sent_ids, word_features, head_ids), kwargs={})
@@ -557,7 +556,6 @@ class SpanPredictor(torch.nn.Module):
         sent_id = torch.tensor(sent_id, device=words.device)
         heads_ids = heads_ids.long()
         same_sent = (sent_id[heads_ids].unsqueeze(1) == sent_id.unsqueeze(0))
-
         # To save memory, only pass candidates from one sentence for each head
         # pair_matrix contains concatenated span_head_emb + candidate_emb + distance_emb
         # for each candidate among the words in the same sentence as span_head
@@ -568,11 +566,11 @@ class SpanPredictor(torch.nn.Module):
             words[cols],
             self.emb(emb_ids[rows, cols]),
         ), dim=1)
-
+        input(len(heads_ids))
         lengths = same_sent.sum(dim=1)
         padding_mask = torch.arange(0, lengths.max().item(), device=words.device).unsqueeze(0)
         padding_mask = (padding_mask < lengths.unsqueeze(1))  # [n_heads, max_sent_len]
-
+        input(padding_mask.shape)
         # [n_heads, max_sent_len, input_size * 2 + distance_emb_size]
         # This is necessary to allow the convolution layer to look at several
         # word scores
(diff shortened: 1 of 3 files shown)


spacy/ml/models/coref.py
```diff
@@ -37,14 +37,11 @@ def build_wl_coref_model(
     except ValueError:
         # happens with transformer listener
         dim = 768
-    
+
     with Model.define_operators({">>": chain}):
         # TODO chain tok2vec with these models
-        # TODO fix device - should be automatic
-        device = "cuda:0"
```
polm, line 44: General comment about the device handling - is this unecessary because Thinc's `prefer_gpu` takes care of it? I notice there's no extra set up anywhere so I guess it just works?

spacy/pipeline/coref.py
```diff
@@ -511,15 +509,20 @@ def update(
...
 
         total_loss = 0
-
-        for eg in examples:
-            preds, backprop = self.model.begin_update([eg.predicted])
-            score_matrix, mention_idx = preds
-
-            loss, d_scores = self.get_loss([eg], score_matrix, mention_idx)
+        docs = [eg.predicted for eg in examples]
+        for doc, eg in zip(docs, examples):
```
polm, line 513: Minor point but can you do something like this instead:

```
for eg in examples:
    doc = eg.predicted
```

spacy/pipeline/coref.py
```diff
@@ -511,15 +509,20 @@ def update(
...
+            # replicates the EntityLinker's behaviour and
+            # copies annotations over https://bit.ly/3iweDcW
+            # takes 'coref_head_clusters' from the reference.
+            for key, sg in eg.reference.spans.items():
+                if key.startswith(self.input_prefix):
+                    aligned_spans = eg.get_aligned_spans_x2y(sg)
+                    doc.spans[key] = [doc[span.start:span.end] for span in aligned_spans]
```
polm, line 520: I think this can just be:

```
doc.spans[key] = eg.get_aligned_spans_y2x(sg)
```

Note this is `y2x` and not `x2y`.

spacy/pipeline/coref.py
```diff
@@ -511,15 +509,20 @@ def update(
...
-
-        for eg in examples:
-            preds, backprop = self.model.begin_update([eg.predicted])
-            score_matrix, mention_idx = preds
-
-            loss, d_scores = self.get_loss([eg], score_matrix, mention_idx)
+        docs = [eg.predicted for eg in examples]
+        for doc, eg in zip(docs, examples):
+            # replicates the EntityLinker's behaviour and
+            # copies annotations over https://bit.ly/3iweDcW
```
polm, line 515: Even if it's long enough it needs its own line, I would avoid using a bitly link here - just use the direct URL.

spacy/pipeline/coref.py
polm: Another thing is that this because this unconditionally overwrites data on the `predicted` Doc, it modifies that Doc for other components in the pipeline. To avoid that it's necessary to save the state of the Doc (`doc.spans` in this case) and restore it afterwards. Check how `old_ents` is used in the EntityLinker. In this case you can probably just do something like:

```
old_spans = [eg.predicted.spans for eg in examples]
# and to reset...
for spans, eg in zip(old_spans, examples):
    eg.predicted.spans = spans
```

spacy/pipeline/coref.py
```diff
@@ -511,15 +509,20 @@ def update(
...
+            for key, sg in eg.reference.spans.items():
+                if key.startswith(self.input_prefix):
+                    aligned_spans = eg.get_aligned_spans_x2y(sg)
+                    doc.spans[key] = [doc[span.start:span.end] for span in aligned_spans]
+            span_scores, backprop = self.model.begin_update([doc])
+            loss, d_scores = self.get_loss([eg], span_scores)
             total_loss += loss
             # TODO check shape here
-            backprop((d_scores, mention_idx))
+            backprop(d_scores)
```
polm, line 525: Doesn't this need a corresponding change in the backprop function, or was that out of sync?

polm (changes requested):
Looks good! There are some things that need to change with the way the SpanGroups are modified but should be minor.

spacy/pipeline/coref.py
polm: Ah, I realize this is the same thing I'm doing in the EntityLinker :sweat_smile: It's fine as is, might be cleaner with the above pattern.

spacy/ml/models/coref.py
kadarakos: Yes, I kept it around in `wl-coref`, because of some hack, but `thinc` deals with it!

spacy/pipeline/coref.py
kadarakos: The `mention_idx` was already unused. Its an integer array so doesn't do anything in the backprop. The torch `autograd` actually errors out if you give it an int array so it was just wrong, but unused.

spacy/pipeline/coref.py
kadarakos: I did re-write it though :D

polm:
This has run to completion training a full model, great! I'll merge it in before I put in commits with cleanup.

PR #10556, opened 2022-03-27
Reset entity-related attributes when an entity is removed

polm:
Includes test.

## Description

If you have entities on a Doc, and filter the list to remove some entities, and then set the list of entities, `ent_id` and `ent_kb_id` values are not reset. This means that you can get non-entity tokens with those values, for example, which seems undesirable. 

This works as intended but is in draft because

1. We need to discuss if this is desired behavior
2. Docs probably need an update

### Types of change

bug fix?

## Checklist

### DIFF
--- a/spacy/tests/doc/test_add_entities.py
+++ b/spacy/tests/doc/test_add_entities.py
@@ -45,6 +45,24 @@ def test_ents_reset(en_vocab):
     assert [t.ent_iob_ for t in doc] == orig_iobs
 
 
+def test_ent_remove(en_vocab):
+    """Ensure that removing entities clears token attributes"""
+    text = ["Louisiana", "Office", "of", "Conservation"]
+    doc = Doc(en_vocab, words=text)
+    entity = Span(doc, 0, 4, label=391)
+    doc.ents = [entity]
+    for tok in entity:
+        tok.ent_id_ = "TEST"
+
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Linking a related comment: https://github.com/explosion/spaCy/pull/9880#issuecomment-1020243254

adrianeboyd:
I worry that this may be too breaking for v3. In general I do think it makes sense to consider updating `Doc.set_ents` so that it goes through the whole doc to make all `token.ent_` attributes consistent (make `ent_*` consistent for all tokens within each provided span, clear all features in `O` cases, etc.).

But even just setting `token.ent_id` within entity spans more consistently in the span ruler PR broke our own code in the entity ruler. Anyone who's doing this incrementally because you couldn't set all the features before with `doc.ents` may have code that breaks.

polm:
Following internal discussion, it was pointed out that with some major changes to the way `Span`s work in v4 this approach alone won't be sufficient for ensuring consistency. I'm going to think about this more, but it might end up best to close this before working on a more complete approach.

adrianeboyd:
See the proposal in #11328

svlandeg:
Closing in favour of https://github.com/explosion/spaCy/pull/11328 which targets the v4 branch - let's continue discussion over there!

PR #10576, opened 2022-03-29
Return doc offsets in Matcher on spans

adrianeboyd:
## Description

The returned match offsets were only adjusted for `as_spans`, not generally. Because the `on_match` callbacks are always applied to the doc, the `Matcher` matches on spans should consistently use the doc offsets.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -252,6 +252,10 @@ cdef class Matcher:
         # non-overlapping ones this `match` can be either (start, end) or
         # (start, end, alignments) depending on `with_alignments=` option.
         for key, *match in matches:
+            # Adjust span matches to doc offsets
+            if isinstance(doclike, Span):
+                match[0] += doclike.start
+                match[1] += doclike.start
             span_filter = self._filter.get(key)
             if span_filter is not None:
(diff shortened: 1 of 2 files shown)


adrianeboyd:
This is mildly breaking, but it corresponds to the API docs and I think running the matcher on spans is not particular common. (I'm actually surprised no one has complained about `on_match` for spans yet.)

adrianeboyd:
Yeah, it may well be too breaking, but it's dissatisfying that there's no way for callbacks to work on spans at this point. I suppose if no one is doing this, then it doesn't really matter. I suppose we could consider it for v4.

svlandeg:
Yea, fixing this in v4 feels safer to me!

adrianeboyd:
(This will fail due to typer until `v4` is also updated.)

svlandeg (approved):
Looks like this is fine to merge into v4, right Adriane?

adrianeboyd:
If we're sure we want to make this change: sure.

svlandeg:
I do feel like this is a bug fix, and no one else seems to object, so let's get this in for `v4` yea :-)

PR #10573, opened 2022-03-29
Add click pin to avoid typer issues

adrianeboyd:
## Description

Add click pin to avoid typer issues.

### Types of change

?

## Checklist

- [x] I confirm that I have the right to submit this contribution under the project's MIT license.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #10593, opened 2022-03-31
Revert "Add click pin to avoid typer issues (#10573)"

adrianeboyd:
## Description

This reverts commit 9966e08f32cf59963591b589bf816aa114ed39a9.

### Types of change

?

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,6 @@ wasabi>=0.8.1,<1.1.0
 srsly>=2.4.1,<3.0.0
 catalogue>=2.0.6,<2.1.0
 typer>=0.3.0,<0.5.0
-click<8.1.0
 pathy>=0.3.5
 # Third party dependencies
 numpy>=1.15.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,7 +52,6 @@ install_requires =
     srsly>=2.4.1,<3.0.0
     catalogue>=2.0.6,<2.1.0
     typer>=0.3.0,<0.5.0
-    click<8.1.0
     pathy>=0.3.5
     # Third-party dependencies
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -52,7 +52,6 @@ install_requires =
     srsly>=2.4.1,<3.0.0
     catalogue>=2.0.6,<2.1.0
     typer>=0.3.0,<0.5.0
```
illia-v, line 54:
```suggestion
    typer>=0.4.1,<0.5.0
```

requirements.txt
```diff
@@ -11,7 +11,6 @@ wasabi>=0.8.1,<1.1.0
 srsly>=2.4.1,<3.0.0
 catalogue>=2.0.6,<2.1.0
 typer>=0.3.0,<0.5.0
```
illia-v, line 13:
```suggestion
typer>=0.4.1,<0.5.0
```

illia-v (commented):
This may be required to avoid the possibility of the error

setup.cfg
adrianeboyd: We want to try to keep the version ranges as wide as possible because people use spacy in combination with so many other tools and libraries. You can't specify particular pairs of versions of click/typer, so it's difficult to make this as specific as it should be.

Since typer v0.4.0 still works with many other versions of click, I'm not sure we want to be too restrictive here. We could consider specifying `typer>=0.3.0,!=0.4.0,<0.5.0`, but I'm not sure this is that helpful in the end.

The only places where we've forbidden intermediate versions in the range are due to security vulnerabilities, or where the dependency caused problems across the board (most recently: pytest v7.1.0).

PR #10600, opened 2022-04-01
Fix loss serialization issue

danieldk:
## Description

Serialization of a model fails with:

```
TypeError: array(738.3855, dtype=float32) is not JSON serializable
```

Fix this using float conversion.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/.github/azure-steps.yml
+++ b/.github/azure-steps.yml
@@ -64,12 +64,13 @@ steps:
     displayName: "Run GPU tests"
     condition: eq(${{ parameters.gpu }}, true)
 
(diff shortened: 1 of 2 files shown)


svlandeg (approved):
Looks good! I think we can merge this and address the failing CLI test later

danieldk:
> Looks good! I think we can merge this and address the failing CLI test later

I have tacked on a commit that should fix the CI issues. These are steps that use models that require `spacy.TransitionbasedParser.v2`. I have disabled these steps for the time being.

PR #10616, opened 2022-04-04
Update default spans_key to sc in API docs

adrianeboyd:
## Description

Update default spans_key to sc in API docs

### Types of change

Docs.

## Checklist


svlandeg:
Nice catch! Not sure how we ended up with two different defaults, one in the factory and a different one in the constructor, but yea this was documented wrongly.

PR #10622, opened 2022-04-04
Ensure that all gold spans are suggested while training spancat

thomashacker:
## Description

Currently, when training a spancat component, the model completely depends on the spans produced by the suggester. If a suggester is used that doesn't cover all span indices from the reference, the spancat loses valuable training data (primarily for positive classification). 

This `PR` aims to ensure that the span candidates contain **all spans from the reference**.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/pipeline/spancat.py
+++ b/spacy/pipeline/spancat.py
@@ -2,6 +2,7 @@ from typing import List, Dict, Callable, Tuple, Optional, Iterable, Any, cast
 from thinc.api import Config, Model, get_current_ops, set_dropout_rate, Ops
 from thinc.api import Optimizer
 from thinc.types import Ragged, Ints2d, Floats2d, Ints1d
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I can see how this is an issue with a poor/misaligned suggester, but I don't think the answer is to use gold spans during training, since you're creating a situation where you're not training on the kind of input you'd see in actual use. It would be similar to training on gold tokenization. All the built-in pipeline components handle partial and misaligned annotation between the gold and predicted docs, and this is pretty similar in effect.

As an example, you could have one 100-gram in your training data that your `[1, 2, 3]` suggester wouldn't suggest and this drop in recall due to the suggester might be something you want to accept as part of training. Updating your model based on this 100-gram could (would?) lower the performance for cases that it actually would see with its suggester.

PR #10629, opened 2022-04-06
Basic tests for the Tamil language

shadeMe:
## Description

Adds basic tokenizer and sentencizer tests for Tamil.

### Types of change

New test fixture
New tests

## Checklist


spacy/tests/lang/ta/test_tokenizer.py
```diff
@@ -0,0 +1,192 @@
...
+    # in the country) as a single token.
+
+    # This modifies the global fixture instance of the tokenizer, meaning
+    # that the change will persist until the end of the active pytest
+    # session. However, given the scope of this change, it should not cause
+    # any side-effects in any of the other tokenizer tests for this language.
+
+    ta_tokenizer.add_special_case("யு.கே.", [{ORTH: "யு.கே."}])
+
+    tokens = ta_tokenizer(text)
```
adrianeboyd, line 190: To avoid potential side effects, you can just create a new tokenizer instead:

```suggestion
    # Add a special rule to tokenize the initialism "யு.கே." (U.K., as
    # in the country) as a single token.
    nlp = Tamil()
    nlp.tokenizer.add_special_case("யு.கே.", [{ORTH: "யு.கே."}])
    tokens = nlp(text)
```

(In general I tend to avoid the individual language class imports and just use `spacy.blank("ta")` instead so I don't have to keep track of the names, but it doesn't really matter either way here.)

adrianeboyd (changes requested):
Very nice, just a few small comments...

adrianeboyd (approved):

PR #10633, opened 2022-04-08
Add back support for beam parsing to the refactored parser

danieldk:
## Description

Beam parsing was already implemented as part of the `BeamBatch` class.
This change makes its counterpart `GreedyBatch`. Both classes are hooked
up in `TransitionModel`, selecting `GreedyBatch` when the beam size is
one, or `BeamBatch` otherwise.

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/ml/tb_framework.py
+++ b/spacy/ml/tb_framework.py
@@ -3,6 +3,8 @@ from thinc.api import Ops, Model, normal_init, chain, list2array, Linear
 from thinc.api import uniform_init, glorot_uniform_init, zero_init
 from thinc.types import Floats1d, Floats2d, Floats3d, Ints2d, Floats4d
 import numpy
+from ..pipeline._parser_internals import _beam_utils
+from ..pipeline._parser_internals.batch import GreedyBatch
 from ..tokens.doc import Doc
 from ..util import registry
 
@@ -139,6 +141,9 @@ def forward(model, docs_moves: Tuple[List[Doc], TransitionSystem], is_train: boo
     nO = model.get_dim("nO")
     nI = model.get_dim("nI")
 
+    beam_width = model.attrs.get("beam_width", 1)
+    beam_density = model.attrs.get("beam_density", 0.0)
+
     ops = model.ops
     docs, moves = docs_moves
     states = moves.init_batch(docs)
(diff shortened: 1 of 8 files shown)


danieldk:
> Judging from the failing tests in Windows, we might be writing/accessing invalid mem addresses at some point 👀

Yeah, I have been running this through Valgrind and ASAN some times, but I can't seem to trigger it on Linux. I'll try one macOS with ASAN as well. If I can't reproduce it there, I think I have to do it on a Windows machine.

svlandeg:
@explosion-bot please test_slow --thinc-branch pull/633/head

svlandeg:
@explosion-bot please test_gpu --thinc-branch pull/633/head

spacy/ml/tb_framework.py
```diff
@@ -139,6 +141,9 @@ def forward(model, docs_moves: Tuple[List[Doc], TransitionSystem], is_train: boo
     nO = model.get_dim("nO")
     nI = model.get_dim("nI")
 
+    beam_width = model.attrs.get("beam_width", 1)
+    beam_density = model.attrs.get("beam_density", 0.0)
```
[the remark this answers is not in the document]
danieldk: > These values are being defined by the beam_parser factory, right? So we should somehow pass those through?

They are set on the model in the `Parser.{predict, greedy_parse, beam_parse}` methods:

https://github.com/explosion/spaCy/blob/8cc29154acf0a3f38da06c3af010d34f4071df65/spacy/pipeline/transition_parser.pyx#L241

Though, it probably makes sense to make this configurable for `TransitionModel` as well, in case someone uses `TransitionModel` outside these methods? I have changed the `TransitionModel` constructor to accept these setttings.

spacy/pipeline/transition_parser.pyx
```diff
@@ -252,13 +252,14 @@ class Parser(TrainablePipe):
     def greedy_parse(self, docs, drop=0.):
         # TODO: Deprecated
```
svlandeg, line 253:
```suggestion
```
I guess this should be removed?

spacy/pipeline/transition_parser.pyx
```diff
@@ -252,13 +252,14 @@ class Parser(TrainablePipe):
...
         self._resize()
+        self._ensure_labels_are_added(docs)
         with _change_attrs(self.model, beam_width=1):
             states, _ = self.model.predict((docs, self.moves))
         return states
 
     def beam_parse(self, docs, int beam_width, float drop=0., beam_density=0.):
         # TODO: Deprecated
```
svlandeg, line 261:
```suggestion
```

spacy/pipeline/transition_parser.pyx
danieldk: I wasn't sure why it said *deprecated* and not *unimplemented*. But since they are implemented again, it makes sense to remove the comments.

svlandeg: It probably should have said something like "unreimplemented" :p We had the functionality before, lost it in a refactor, but always meant to reimplement it again.

PR #10653, opened 2022-04-13
Add feminine form of word "one" in French

fonfonx:
Add the feminine form of `one`, `first` and `second` (`une`, `première` and `seconde`) in order to make it recognized by the `LIKE_NUM` operator of the `Matcher`.

## Checklist


svlandeg (commented):
Thanks for the PR! While we're at it, perhaps we should also add `première` to the `_ordinal_words`? And `seconde`?

fonfonx:
Good point, I have added them

svlandeg (approved):
Thanks again!

PR #10659, opened 2022-04-14
`Matcher`: Remove superfluous GIL-acquiring check in `get_is_final`

shadeMe:
## Description
This check incurred a significant performance penalty due to implict interactions between the GIL and Cython ref-counting code.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -690,14 +690,14 @@ cdef int8_t get_is_match(PatternStateC state,
     return True
 
 
-cdef int8_t get_is_final(PatternStateC state) nogil:
+cdef inline int8_t get_is_final(PatternStateC state) nogil:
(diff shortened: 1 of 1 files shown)


svlandeg:
We've already discussed this internally, and this looks good to merge? Was there anything else you wanted to do/experiment with still, Madeesh?

shadeMe:
I was going to flag it as ready as soon as the tests successfully ran to completion, but then the breakage due to the allocations PR happened. 

I'd say that the changes in this PR shouldn't haven't any unforeseen side-effects, so it should be safe to merge. But we could also wait until next week, if we want to err on the side of caution.

svlandeg:
Raising the error there was more about being robust to future bugs, I can't imagine removing that check to cause any bugs by itself, really?

shadeMe:
On second thought, you're right. The change is very unlikely to break anything. I'll mark it as ready :)

PR #10668, opened 2022-04-18
Add spacy-span-analyzer to debug data

ljvmiranda921:
This Pull Request adds the [spaCy span analyzer](https://github.com/ljvmiranda921/spacy-span-analyzer) in debug data. Since the frequency has already been included in the original implementation, my PR will add the ff: spans length, span distinctiveness, and boundary distinctiveness.

### DIFF
--- a/spacy/cli/debug_data.py
+++ b/spacy/cli/debug_data.py
@@ -216,11 +216,14 @@ def debug_data(
                 f"Key: {spans_key}, {_format_labels(data_labels.items(), counts=True)}",
                 show=verbose,
             )
+
         # Data checks: only take the spans keys in the actual spancat components
         data_labels_in_component = {
             spans_key: gold_train_data["spancat"][spans_key]
             for spans_key in model_labels_spancat.keys()
         }
+
+        breakpoint()
         for spans_key, data_labels in data_labels_in_component.items():
             for label, count in data_labels.items():
                 # Check for missing labels
@@ -692,14 +695,14 @@ def _compile_gold(
                 elif label == "-":
                     data["ner"]["-"] += 1
         if "spancat" in factory_names:
-            for span_key in list(eg.reference.spans.keys()):
-                if span_key not in data["spancat"]:
-                    data["spancat"][span_key] = Counter()
-                for i, span in enumerate(eg.reference.spans[span_key]):
+            for spans_key in list(eg.reference.spans.keys()):
+                if spans_key not in data["spancat"]:
+                    data["spancat"][spans_key] = Counter()
+                for i, span in enumerate(eg.reference.spans[spans_key]):
                     if span.label_ is None:
                         continue
                     else:
-                        data["spancat"][span_key][span.label_] += 1
+                        data["spancat"][spans_key][span.label_] += 1
         if "textcat" in factory_names or "textcat_multilabel" in factory_names:
             data["cats"].update(gold.cats)
             if any(val not in (0, 1) for val in gold.cats.values()):
@@ -824,3 +827,7 @@ def _get_labels_from_spancat(nlp: Language) -> Dict[str, Set[str]]:
             labels[pipe.key] = set()
         labels[pipe.key].update(pipe.labels)
     return labels
+
+
+def _get_span_characteristics() -> Dict[str, Any]:
+    pass


ljvmiranda921:
I've been thinking about adding "human-facing recommendations" similar to what @/pmbaumgartner suggested. For example:
- `spans_length` can help recommend ngram suggester size. At very high values, we might recommend lowering the batch size
- A boundary distinctiveness / span distinctiveness > 0.8 (closer to PARC and RiQUA dataset) is viable for spancat. Just not sure how to translate it properly in `debug data` ("Spancat should be viable for your dataset!" or something like that).
- I am also thinking of removing the table altogether, and just report the weighted averages instead? 

Given that, I'd like to hear reviews of the technical implementation first before I make major changes!

spacy/cli/debug_data.py
```diff
@@ -247,6 +249,52 @@ def debug_data(
...
+            )
+            metrics = [span_length, span_distinctiveness, sb_distinctiveness]
+            span_char = _compile_span_characteristics(
+                span_data=metrics,
+                labels=data_labels,
+            )
+            # Compute for weighted average (by frequency)
+            wgt_avg = [_wgt_average(metric, data_labels) for metric in metrics]
+            _wgt_avg_row = ["Wgt. Average"] + [str(w) for w in wgt_avg]
+            msg.table(span_char, footer=_wgt_avg_row, header=headers, divider=True)
```
svlandeg, line 296: In the final table, I would round all numbers to two decimals, I don't think more is meaningful?

svlandeg:
>     * `spans_length` can help recommend ngram suggester size. If your average spans length is, say, 3.14, then a good ngram suggester size may be 1 -> 3 or 1 -> 4.

To support this, it would be great to also show the min and max length, right? But I suppose that would become a bit verbose? 

>     * It's also applicable for other settings. Higher spans length tend to have larger ngram sizes, and so we can recommend lowering the batch size. If your spans length is > 7, you might want to lower the batch size to reduce CPU load when training

I really like this idea, but I'm not sure whether `debug data` is the best place for this kind of recommendation. There is no notion of a model or config for this command. Still, we should talk about how we could incorporate these kind of recommendations into the CLI functionality.

>     * From our experiments, a boundary distinctiveness / span distinctiveness > 0.8 (closer to PARC and RiQUA dataset) is viable for spancat. Just not sure how to translate it properly in `debug data` ("Spancat should be viable for your dataset!" or something like that). There's a positive correlation between BD/SD and spancat performance.

By "viable for spancat", you mean that the performance will probably be OK, right? We could also turn this around and say: if the value is too low, indicating "smooth transitions", we could emit a warning saying something to this effect? Because it might also mean that the entities are just not well defined in the annotation schema, right?

spacy/cli/debug_data.py
ljvmiranda921: Agree will do that!

ljvmiranda921:
Hi @svlandeg , thanks for the comments! I updated the PR to make the presentation more compact. 👍 
I also added a few things  (if they don't fit debug data it's easy to just remove them afterwards).
image

ljvmiranda921:
Hi @thomashacker , I added new methods `_get_span_characteristics` and `_print_span_characteristics` in this commit: https://github.com/explosion/spaCy/pull/10668/commits/fbd1074cc7fd817342ae631c490e02e23648ad82

Ideally, you want to pass a list of [Examples](https://spacy.io/api/example) to _compile_gold, but you can also
follow how it's done in the `debug_data` function:

```python
# (Optional) Config loading as copied over from debug-data
# You don't have to follow this, at the end, we only just need a list of Examples to be passed
# to _compile_gold
with show_validation_error(config_path):
    cfg = util.load_config(config_path)
    nlp = util.load_model_from_config(cfg)
    config = nlp.config.interpolate()

dot_names = [T["train_corpus"]]
train_corpus, dev_corpus = resolve_dot_names(config, dot_names)
```

And here's the actual way to get the span characteristics from the private methods

```python
import spacy
from spacy.cli.debug_data import _compile_gold, _get_span_characteristics
from spacy.cli.debug_data import _print_span_characteristics


SPANS_KEY = "sc"
nlp = spacy.blank("en")

train_dataset = list(train_corpus(nlp))  # a list of Examples
gold_train_data = _compile_gold(train_dataset, factory_names=["spancat"], nlp, make_proj=False)
data_labels = gold_train_data["spancat"][SPANS_KEY]
span_characteristics = _get_span_characteristics(train_dataset, gold_train_data, SPANS_KEY, data_labels)  # a dictionary
_print_span_characteristics(span_characteristics)  # optional, for pretty-printing

```

svlandeg (approved):
Looks good to me! It's a really nice addition to have the warning/confirmation about span & boundary distinctiveness. We'll probably need to add some sort of explanation about these measures somewhere in the docs though? Maybe over at https://spacy.io/api/cli#debug-data which can refer to the Papay et al work, what do you think?

ljvmiranda921:
Ah nice that makes sense, will add a few notes in the documentation :)

spacy/cli/debug_data.py
```diff
@@ -247,6 +253,56 @@ def debug_data(
...
+            _print_span_characteristics(span_characteristics)
+
+            msg.info(
+                "It's recommended to use a maximum of "
+                f"{math.ceil(span_characteristics['avg_length'])} ngrams "
+                "when you're using the n-gram suggester function "
+                f"(avg={round(span_characteristics['avg_length'],2)}, "
+                f"min={span_characteristics['min_length']}, max={span_characteristics['max_length']})",
+                show=verbose,
+            )
```
adrianeboyd, line 272: I think this suggestion should be based on recall rather than average length and analyze both min and max. (It's not too hard to construct a case with outliers where the average length leads to <20% recall.)

I think it might make sense to present several min/max settings with recall (something like 75%, 85%, 95%) and provide info/hints about the speed/RAM tradeoffs involved so the users can make a decision themselves.

ljvmiranda921: Hi @adrianeboyd , thanks for the comment! I'm curious how I can best implement this. Do you mean something like:

```
If you want to get:
75% recall, then set min=$A, max=$B,
85% recall, then set min=$C, max=$D,
95% recall, then set min=$E, max=$F
```

where `$A` to `$F` are some integers. I think this might fall under the  `recommend` command? 

Also if that's the case, perhaps I should remove the highlighted line, `It's recommended to use a maximum of ...`, from `debug data`, and insert it to `recommend` instead? 🤔

adrianeboyd: I'm not sure what you mean by "insert to `recommend`"?

I don't know enough about algorithms that would make this easy/fast to do for the general case. Maybe instead of a direct recommendation, you can just present the percentage of spans that fall in each ngram length and point out that there are speed/RAM tradeoffs in including all outliers?

ljvmiranda921: > I'm not sure what you mean by "insert to recommend"?

It's a "new" command that we're considering to include more explicit/direct config recommendations to the user based on their dataset and other factors. 

> Maybe instead of a direct recommendation, you can just present the percentage of spans that fall in each ngram length and point out that there are speed/RAM tradeoffs in including all outliers?

Ah that's doable! I'll update the PR to include this :)

ljvmiranda921: I updated the PR here to improve span length reporting: https://github.com/explosion/spaCy/pull/10668/commits/d0b7db759c53a3f1e37971f00e272bb7666bfc91

image

adrianeboyd: I think presenting the n-gram lengths as strings and in the order is kind of confusing. I think it's still easy enough to understand if it's sorted in the order of n-gram lengths?

ljvmiranda921: Ah sure that makes a lot of sense :) Will update the PR

ljvmiranda921: Updated in: https://github.com/explosion/spaCy/pull/10668/commits/1e796fb8e6532846b19163c1175c1128347be41b

svlandeg (approved):
This (still) looks good to me but I'm happy to defer the decision for merging/further edits to Adriane :-)

adrianeboyd:
I ran this for the `benchmarks/ner_spancat` and it looks like something is off with the math:

```
ℹ 90% of spans have up to 2 n-grams (min=1, max=13). Most common span
lengths are: 1 (42%), 2 (37%). If you are using the n-gram suggester, note that
there are speed / memory tradeoffs when you include all outliers.
```

With `-V` I'd like to see the full distribution of span lengths and not just common ones.

spacy/cli/debug_data.py
```diff
@@ -247,6 +255,63 @@ def debug_data(
...
+            _span_freqs = dict(sorted(_span_freqs.items()))
+            span_freqs = [(str(k), int(v)) for k, v in _span_freqs.items()]
+
+            msg.info(
+                f"{SPAN_LENGTH_THRESHOLD_PERCENTAGE}% of spans have up to "
+                f"{max(_span_freqs.keys())} n-grams "
+                f"(min={span_characteristics['min_length']}, max={span_characteristics['max_length']}). "
+                f"Most common span lengths are: {_format_freqs(span_freqs)}. "
+                "If you are using the n-gram suggester, note that there are speed / memory tradeoffs "
+                "when you include all outliers."
```
adrianeboyd, line 280:
```suggestion
                "If you are using the n-gram suggester, note that omitting "
                "infrequent n-gram lengths can greatly improve speed and " 
                "memory usage with only a small effect on recall."
```

website/docs/api/cli.md
```diff
@@ -466,6 +466,21 @@ takes the same arguments as `train` and reads settings off the
 
 </Infobox>
 
+
+<Infobox title="Notes on span characteristics" emoji="💡">
+
+If your pipeline contains a `"spancat"` component, then this command will also
```
adrianeboyd, line 472:
```suggestion
If your pipeline contains a `spancat` component, then this command will also
```

website/docs/api/cli.md
```diff
@@ -466,6 +466,21 @@ takes the same arguments as `train` and reads settings off the
...
+report span characteristics such as the average span length and the span (or
+span boundary) distinctiveness. The distinctiveness measure shows how different
+the tokens are with respect to the rest of the corpus using the KL-divergence of
+the token distributions. To learn more, you can check Papay et al's work on
```
adrianeboyd, line 476:
```suggestion
the token distributions. To learn more, you can check Papay et al.'s work on
```

ljvmiranda921:
Ah I see thanks for catching @adrianeboyd , is it because the percentages didn't add up to 90%? Perhaps it's because when I add the percentage of the spans with 3 n-grams, it exceeded 90%, so it wasn't included. I'll update the logic here or the way it was presented. 

> With -V I'd like to see the full distribution of span lengths and not just common ones.

Noted on this. Will update the PR

ljvmiranda921:
Updated the displays:
- Used floating-point numbers for the percentages (because sometimes using integers will just show 0%, which I personally find weird and uninformative).
- When `-V` is passed, the full distribution of span lengths is shown

image

spacy/cli/debug_data.py
```diff
@@ -247,6 +255,69 @@ def debug_data(
                     msg.warn(f"No examples for texts WITHOUT new label '{label}'")
                     has_no_neg_warning = True
 
+            with msg.loading("Obtaining span characteristics..."):
+                span_characteristics = _get_span_characteristics(
+                    train_dataset, gold_train_data, spans_key
+                )
+
+            msg.info(f"Span characteristics for spans_key `{spans_key}`")
```
adrianeboyd, line 263:
```suggestion
            msg.info(f"Span characteristics for spans_key '{spans_key}'")
```

spacy/cli/debug_data.py
```diff
@@ -247,6 +255,69 @@ def debug_data(
...
+            )
+            _filtered_span_freqs = _filter_spans_length_freq_dist(
+                _span_freqs, threshold=SPAN_LENGTH_THRESHOLD_PERCENTAGE
+            )
+
+            msg.info(
+                f"At least {SPAN_LENGTH_THRESHOLD_PERCENTAGE}% of spans have up to "
+                f"{max(_filtered_span_freqs.keys())} n-grams "
+                f"(min={span_characteristics['min_length']}, max={span_characteristics['max_length']}). "
+                f"Most common span lengths are: {_format_freqs(_filtered_span_freqs)}. "
```
adrianeboyd, line 278:
```suggestion
                f"The most common span lengths are: {_format_freqs(_filtered_span_freqs)}. "
```

spacy/cli/debug_data.py
```diff
@@ -247,6 +255,69 @@ def debug_data(
...
+                "If you are using the n-gram suggester, note that omitting "
+                "infrequent n-gram lengths can greatly improve speed and "
+                "memory usage with only a small effect on recall."
```
adrianeboyd, line 281: I don't like my own edit!

```suggestion
                "memory usage."
```

spacy/cli/debug_data.py
```diff
@@ -216,6 +223,7 @@ def debug_data(
                 f"Key: {spans_key}, {_format_labels(data_labels.items(), counts=True)}",
                 show=verbose,
             )
+
```
adrianeboyd, line 226:
```suggestion
```

spacy/cli/debug_data.py
```diff
@@ -247,6 +255,69 @@ def debug_data(
...
+            _span_freqs = _get_spans_length_freq_dist(
+                gold_train_data["spans_length"][spans_key]
+            )
+            _filtered_span_freqs = _filter_spans_length_freq_dist(
+                _span_freqs, threshold=SPAN_LENGTH_THRESHOLD_PERCENTAGE
+            )
+
+            msg.info(
+                f"At least {SPAN_LENGTH_THRESHOLD_PERCENTAGE}% of spans have up to "
```
adrianeboyd, line 275: "have up to" seems a little unclear, since it's not obvious that you're necessarily going to count from 1 and it's not that spans contain/have n-grams. I would suggest something more like:

"Over 90% of spans have lengths of 1 -- {NUM}"

"N-gram lengths 1 -- {NUM} At least 90% of spans have lengths 1 -- {NUM}" / "At least 90% of spans are of length 1 -- {NUM}"

ljvmiranda921: Noted on this! Updated in https://github.com/explosion/spaCy/pull/10668/commits/9d3c1389f7a4df6fbf5914369b94cf01fe8f5401

spacy/cli/debug_data.py
```diff
@@ -247,6 +254,69 @@ def debug_data(
...
+            _span_freqs = _get_spans_length_freq_dist(
+                gold_train_data["spans_length"][spans_key]
+            )
+            _filtered_span_freqs = _filter_spans_length_freq_dist(
+                _span_freqs, threshold=SPAN_LENGTH_THRESHOLD_PERCENTAGE
+            )
+
+            msg.info(
+                f"Over {SPAN_LENGTH_THRESHOLD_PERCENTAGE}% of spans have lengths of 1 - "
+                f"{max(_filtered_span_freqs.keys())} n-grams "
```
adrianeboyd, line 275:
```suggestion
                f"Over {SPAN_LENGTH_THRESHOLD_PERCENTAGE}% of spans have lengths of 1 -- "
                f"{max(_filtered_span_freqs.keys())} "
```

website/docs/api/cli.md
```diff
@@ -455,7 +455,7 @@ File       /path/to/thinc/thinc/schedules.py (line 91)
 
 Analyze, debug and validate your training and development data. Get useful
 stats, and find problems like invalid entity annotations, cyclic dependencies,
-low data labels and more.
+low data labels and more.
```
adrianeboyd, line 458:
```suggestion
low data labels and more.
```

website/docs/api/cli.md
```diff
@@ -466,6 +466,21 @@ takes the same arguments as `train` and reads settings off the
...
+span boundary) distinctiveness. The distinctiveness measure shows how different
+the tokens are with respect to the rest of the corpus using the KL-divergence of
+the token distributions. To learn more, you can check Papay et al.'s work on
+[*Dissecting Span Identification Tasks with Performance Prediction* (EMNLP
+2020)](https://aclanthology.org/2020.emnlp-main.396/).
+
+</Infobox>
+
+
+
```
adrianeboyd, line 483:
```suggestion
<Infobox title="Notes on span characteristics" emoji="💡">

If your pipeline contains a `spancat` component, then this command will also
report span characteristics such as the average span length and the span (or
span boundary) distinctiveness. The distinctiveness measure shows how different
the tokens are with respect to the rest of the corpus using the KL-divergence of
the token distributions. To learn more, you can check out Papay et al.'s work on
[*Dissecting Span Identification Tasks with Performance Prediction* (EMNLP
2020)](https://aclanthology.org/2020.emnlp-main.396/).

</Infobox>

```

adrianeboyd (commented):
Minor edits, mostly whitespace

svlandeg:
Thanks again to @ljvmiranda921 for pulling this together and to @adrianeboyd for the careful and considerate reviews! It will be great to have this in spaCy 3.4 🎉

PR #10669, opened 2022-04-18
Fix some issues in Spanish stop-word list and examples

mgrojo:
## Description

Remove significant or not very frequent words from stop word list for Spanish.
    
https://github.com/explosion/spaCy/issues/3052#issuecomment-1100760100

Some issues in Spanish examples.

### Types of change

Bug fix.

## Checklist


richardpaulhudson (commented):
Thank you very much for this, the examples look much better now and the stop words list is more coherent.

In normal circumstances, `develop` would be the right branch to target (as you've done). However, as we are currently in the runup to release 3.3 and want to include this PR there, I'm merging it directly into `master`, which is where we're managing the 3.3 stuff.

PR #10673, opened 2022-04-19
#10672: fixes displacy output for manual unsorted entities

### LINKED ISSUE #10672: Rendering unsorted manual entities in displacy produce incorrect output
When providing manual data for spaCy rendering you must make sure the entities are sorted or it won't work.  This is described in the documentation: https://spacy.io/usage/visualizers#manual-usage. This is however not obvious if you implement it without using the docs, e.g. by copying someone else's code. Also you may not become aware if you have not encounterd unsorted cases in your tests and then it could fail in production.

I ran into this problem and noticed someone else had the same: #4951

@ines I noticed your remark https://github.com/explosion/spaCy/issues/4951#issuecomment-579835246 and figured it is indeed easy to fix. In fact, if everyone who uses manual rendering must do the sorting, I propose to add it to the package itself.  Or am I missing something? 

I'll open a PR with a fix shortly for you to consider.

```
import spacy
doc = {
    "text": "But Google is starting from behind.",
    "ents": [{"start": 14, "end": 22, "label": "SECOND"},
        {"start": 4, "end": 10, "label": "FIRST"}, 
            ],
    "title": None
}

spacy.displacy.render(doc, style="ent", manual=True)
```
image

Expected:
image

## Your Environment
- **spaCy version:** 3.3.0.dev0
- **Platform:** Windows-10-10.0.19043-SP0
- **Python version:** 3.9.12

harmbuisman:
Closes #10672

In addition:
- had to add a dependency for black due to error caused by click package (see https://github.com/psf/black/issues/2964), another option would be to go to newest black version
- I also updated the docs that did not yet mention that a dict is also possible as input (https://spacy.io/api/top-level#displacy.render)
- I saw that List is not used in import line below, but judging from other files in the package I used a noqa to prevent the flake8 fail just in case this import is important for dependent code
```from typing import List, Union, Iterable, Optional, Dict, Any, Callable```

## Description
Now you no longer need to manually sort entities by their start value if providing a dict with ents in it. See #10672 and #4951

### Types of change
Bugfix if you would consider it a bug, enhancement if you consider existing behaviour not to be a bug
Updated docs
Added unit test

## Checklist

- [ x] I confirm that I have the right to submit this contribution under the project's MIT license.

### DIFF
--- a/website/docs/api/top-level.md
+++ b/website/docs/api/top-level.md
@@ -263,7 +263,7 @@ Render a dependency parse tree or named entity visualization.
 
 | Name        | Description                                                                                                                                                                            |
 | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `docs`      | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span]], Doc, Span]~~                                                                                                  |
+| `docs`      | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span, dict]], Doc, Span, dict]~~                                                                                                  |
 | `style`     | Visualization style, `"dep"` or `"ent"`. Defaults to `"dep"`. ~~str~~                                                                                                                  |
 | `page`      | Render markup as full HTML page. Defaults to `True`. ~~bool~~                                                                                                                          |
 | `minify`    | Minify HTML markup. Defaults to `False`. ~~bool~~                                                                                                                                      |
--- a/website/docs/api/top-level.md
+++ b/website/docs/api/top-level.md
@@ -263,7 +263,7 @@ Render a dependency parse tree or named entity visualization.
 
 | Name        | Description                                                                                                                                                                            |
 | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `docs`      | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span, dict]], Doc, Span, dict]~~                                                                                                  |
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by svlandeg** [comment]
Thanks for the PR!

> I saw that List is not used in import line below, but judging from other files in the package I used a noqa to prevent the flake8 fail just in case this import is important for dependent code

I think it'd be fine to just remove `List` from the import statement. Not sure which other files you're referring to, but IMO it would be weird if people would be importing `List` from `displacy` and not from `typing`?

**Comment by harmbuisman** [comment]
> Thanks for the PR!
> 
> > I saw that List is not used in import line below, but judging from other files in the package I used a noqa to prevent the flake8 fail just in case this import is important for dependent code
> 
> I think it'd be fine to just remove `List` from the import statement. Not sure which other files you're referring to, but IMO it would be weird if people would be importing `List` from `displacy` and not from `typing`?

Thanks. Removed the unused import. Tests pass, so seems safe indeed.

File: website/docs/api/top-level.md [hunk 1]
```diff
@@ -263,7 +263,7 @@ Render a dependency parse tree or named entity visualization.
 
 | Name        | Description                                                                                                                                                                            |
 | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `docs`      | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span]], Doc, Span]~~                                                                                                  |
+| `docs`      | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span, dict]], Doc, Span, dict]~~                                                                                                  |
```
svlandeg, line 266:
```suggestion
| `docs`      | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span, dict]], Doc, Span, dict]~~                                                                                      |
```

PR #10676, opened 2022-04-20
Have SpanGroups.setdefault always return the set SpanGroup value

### LINKED ISSUE #10670: SpanGroups.setdefault may return the given default arg--not necessarily the SpanGroup
## Problem:

When calling `.setdefault()` on a `SpanGroups` object (e.g. `Doc().spans`),
* **if:** the key given to setdefault() is not in the `SpanGroups`
* **and:** the default value given to setdefault() is not an instance of `SpanGroup`,
* **then:** `setdefault()` will return (what should likely be considered) an incorrect value: the default (non-`SpanGroup`) value given to setdefault().

This can be particularly problematic if the return value of `setdefault()` is operated upon as if it was the intended `SpanGroup` (and the aforementioned conditions were true).

For example:

(Executed with:

- **spaCy version:** 3.2.4
- **Platform:** Linux-5.17.3-x86_64-with-glibc2.35
- **Python version:** 3.10.4

)

```python
>>> import spacy
>>> from spacy.tokens import Doc, Span, SpanGroup
>>> nlp = spacy.blank('en')
>>> doc = nlp('This is a doc.')
>>> doc.spans
{}
>>> span1 = Span(doc, 0, 1)
>>> span2 = Span(doc, 1, 2)

>>> # Initial update (key not present):
>>> doc.spans.setdefault('mygroup', []).append(span1)
>>> doc.spans
{'mygroup': []}
>>> # ---------------------------------------------
>>> # Expected:
>>> #   set(doc.spans['mygroup']) == {span1}
>>> # ---------------------------------------------
>>> isinstance(doc.spans['mygroup'], SpanGroup) # sanity check
True

>>> # Subsequent update (key present):
>>> doc.spans.setdefault('mygroup', []).append(span2)
>>> doc.spans
{'mygroup': [is]}
>>> # aka:
>>> #   set(doc.spans['mygroup']) == {span2}
>>> # Notice that subsequent calls to setdefault() with an existing key
>>> # operate as expected (i.e. the return value is the SpanGroup).
>>> # ---------------------------------------------
>>> # Expected:
>>> #   set(doc.spans['mygroup']) == {span1, span2}
>>> # ---------------------------------------------
>>> isinstance(doc.spans['mygroup'], SpanGroup) # sanity check
True
```

In general, users may attempt to initially populate a `SpanGroup` via the return value of `setdefault()`, but find that their initial `.append/.extend/...` is not applied to the intended `SpanGroup`; their initial update is essentially ignored.

## Cause:

`SpanGroups` subclasses `UserDict` but does not implement `setdefault` itself, and CPython's `UserDict` uses the `MutableMapping` implementation of `setdefault`.

CPython v3.10.4's `MutableMapping.setdefault` implementation is exactly:
```python
def setdefault(self, key, default=None):
    'D.setdefault(k[,d]) -> D.get(k,d), also set D[k]=d if k not in D'
    try:
        return self[key]
    except KeyError:
        self[key] = default
    return default
```

So, if a `MutableMapping` subclass (e.g. `UserDict`, and thus `SpanGroups`) implements its own `__setitem__` and that `__setitem__` may "alter" the stored value, as is the case with `SpanGroups` (when it changes a non-`SpanGroup` value to a `SpanGroup` via `_make_span_group`), then the value returned by the `setdefault()` call is not the altered value, but the value given as the default arg to `setdefault()`.

(Given Python's dynamic nature (and my remaining ignorance about Python), I was a bit surprised by this behavior in a standard library function. I took a look and found that someone already reported this on BPO a couple years ago ([https://bugs.python.org/issue41647](https://bugs.python.org/issue41647); now mirrored at python/cpython#85813) and it's marked as "not a bug".)

As it works in my example above: when calling e.g. `doc.spans.setdefault('mygroup', [])`, if the 'mygroup' key is not yet present in the `doc.spans` SpanGroups, a new `SpanGroup(doc, name='mygroup', spans=[])` will be created as the value for the new 'mygroup' key, and the empty list `[]` passed to setdefault() will be returned.
Any operations on that returned value apply to that list--not the `SpanGroup` that was created.

## Solution:

(Forthcoming (within a day); I've just written it--it's only a few lines--but I wanted to double-check your contributing guide and consider including a test, etc.)

### Alternative:

Alternatively, this could be considered "not a bug" for spaCy as well, and users would be expected to pass a `SpanGroup` instance as a `setdefault()` default value (or, in general, keep in mind how `UserDict.setdefault` behaves). Though it would be a bit verbose/redudant, e.g.:
```python
doc.spans.setdefault('mygroup',
                     SpanGroup(doc, name='mygroup', spans=[])) \
                    .append(span1)
```
I figured that since `SpanGroups` is noted as a dict-*like* proxy, it didn't necessarily have to inherit quirks of e.g. `UserDict`.

single-fingal:
## Description

Resolves #10670

This changes the behavior so that `SpanGroups.setdefault` will no longer return a non-`SpanGroup` default value passed to setdefault() (if the given key is not present in the SpanGroups); setdefault() will now always return the SpanGroup object that was set as the value for the given key.

### Implementation discussion

The current implementation is open for review. There were a few options that came to mind to resolve #10670, given the current definition of SpanGroups, and each has advantages and disadvantages.

1. One option is to leave `SpanGroups.__setitem__` defined as-is, and to define `SpanGroups.setdefault` as:
```python
def setdefault(
    self, key: str, default: Union[SpanGroup, Iterable["Span"]] = tuple()
) -> SpanGroup:
    if key in self:
        return self[key]
    self[key] = default
    return self[key]
```
This is similar to the underlying `MutableMapping.setdefault`, but it requires performing an additional `__getitem__()` to return the (`SpanGroup`) value that was actually set--since `SpanGroups.__setitem__` may alter the provided value to make it a SpanGroup.

2. To avoid the additional `__getitem__()`, another option would be to essentially have the functionality of `__setitem__`, but also return the SpanGroup value that was set. One way to do this is:
```python
def _setitem(
    self, key: str, value: Union[SpanGroup, Iterable["Span"]]
) -> SpanGroup:
    if not is instance(value, SpanGroup):
        value = self._make_span_group(key, value)
    assert value.doc is self.doc_ref()
    UserDict.__setitem__(self, key, value)
    return value

def __setitem__( ...same signature... ):
    self._setitem(key, value)

def setdefault( ...same signature as in option 1... ):
    if key in self:
        return self[key]
    return self._setitem(key, default)
```
Here, `_setitem` is basically the existing implementation of `__setitem__` with a `return value` tacked on. It does have the benefit of maintaining \_\_setitem\_\_'s current signature, though I'm not particularly keen on having each call to `__setitem__()` simply wrap another function call, particularly since I assume `setdefault()` will be called less frequently than \_\_setitem\_\_(). Alternatively, the current \_\_setitem\_\_ code could basically be duplicated within setdefault, but code duplication has its own issues.

3. (Have I warmed you up to having a \_\_setitem\_\_() return a value?)
Another way to get the same effect as the prior option would just have the existing \_\_setitem\_\_ return a `SpanGroup` itself, instead of `None`. Then `setdefault` would include `return self.__setitem__(key, default)`.

This latter option is the one I've implemented in this pull request's initial commit. Having any class's `__setitem__` return a non-None value is certainly unusual from my experience, but it seemed like it might be an appropriate solution for SpanGroups, when compared against the other options I considered. Every value within a SpanGroups object is always going to be a `SpanGroup`, and I *think* it's reasonable to expect that user code up until this point has not attempted to make use of \_\_setitem\_\_()'s (`None`) return value.
Going forward with this, a risk would then be having user code that uses/relies on SpanGroups.\_\_setitem\_\_ returning its set value. But for `SpanGroups` specifically, maybe that would be an acceptable behavior to maintain.

Note: To accommodate the latter option, I've included a "`type: ignore[override]`", as mypy has rightly pointed out that the `UserDict`/`MutableMapping`'s \_\_setitem\_\_ definitions return `None` and not `SpanGroup`.

Other notes:
1. I've set the default value for `setdefault` to be `tuple()`, since we accept `Iterable[Span]` values, and this would be transformed into an empty `SpanGroup`. I know `SimpleFrozenList()` is used in similar contexts, but I saw an existing `tuple()` default arg in SpanGroups' module and went with that.
2. I avoided using a `try/except` statement in `SpanGroups.setdefault`, as is done in `MutableMapping.setdefault`, because the spaCy developers' guide stated a preference for explicit condition checks.

### Types of change
Bug fix and/or design change/enhancement

## Checklist

Let me know what you think. I know there are other options that would satisfy #10670, but given SpanGroups' current definition, the ones I mentioned above seemed like more-or-less viable choices.

### DIFF
--- a/spacy/tests/doc/test_span_group.py
+++ b/spacy/tests/doc/test_span_group.py
@@ -119,6 +119,18 @@ def test_span_group_set_item(doc, other_doc):
         span_group[index] = span
 
 
+@pytest.mark.issue(10670)
+def test_issue10670(en_tokenizer):
+    # Start with a doc that we know has no 'test' SpanGroup
+    doc = en_tokenizer("0 1 2 3 4 5 6")
+    span1 = doc[0:1]
+    span2 = doc[1:2]
+    doc.spans.setdefault("test", []).append(span1)
+    assert set(doc.spans["test"]) == {span1}
+    doc.spans.setdefault("test", []).append(span2)
+    assert set(doc.spans["test"]) == {span1, span2}
+
+
 def test_span_group_has_overlap(doc):
     span_group = doc.spans["SPANS"]
     assert span_group.has_overlap
--- a/spacy/tokens/_dict_proxies.py
+++ b/spacy/tokens/_dict_proxies.py
@@ -16,7 +16,7 @@ if TYPE_CHECKING:
 # Why inherit from UserDict instead of dict here?
 # Well, the 'dict' class doesn't necessarily delegate everything nicely,
 # for performance reasons. The UserDict is slower but better behaved.
-# See https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/0ww
+# See https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/
 class SpanGroups(UserDict):
     """A dict-like proxy held by the Doc, to control access to span groups."""
 
@@ -28,11 +28,21 @@ class SpanGroups(UserDict):
         self.doc_ref = weakref.ref(doc)
         UserDict.__init__(self, items)  # type: ignore[arg-type]
 
-    def __setitem__(self, key: str, value: Union[SpanGroup, Iterable["Span"]]) -> None:
+    def __setitem__(  # type: ignore[override]
+        self, key: str, value: Union[SpanGroup, Iterable["Span"]]
+    ) -> SpanGroup:
         if not isinstance(value, SpanGroup):
             value = self._make_span_group(key, value)
         assert value.doc is self.doc_ref()
         UserDict.__setitem__(self, key, value)
+        return value
+
+    def setdefault(
+        self, key: str, default: Union[SpanGroup, Iterable["Span"]] = tuple()
+    ) -> SpanGroup:
+        if key in self:
+            return self[key]
+        return self.__setitem__(key, default)
 
     def _make_span_group(self, name: str, spans: Iterable["Span"]) -> SpanGroup:
         doc = self._ensure_doc()


adrianeboyd:
Closing in favor of #10772

PR #10688, opened 2022-04-22
Add Doc.from_json()

rmitsch:
## Description

- Adds a static `Doc.from_json()` method allowing to generate `Doc` instances from JSON output produced by `Doc.to_json()`.
- Added tests for `Doc.from_json()`.
- Moved a function-level `pytest` fixture from `spacy/tests/doc/test_to_json.py` to `spacy/tests/conftest.py`, as it's now also used by `spacy/tests/doc/test_from_json.py`.
- Renamed entities' and spans' `start` entry in JSON generated by `Doc.to_json()` to `start_char`, since it contains the respective `start_char` values. Added `start` entry containing the corresponding `start` values.
- Added brief sections for both `Doc.to_json()` and `Doc.from_json()` to the API documentation.
- Bumped `black` to 22.3.0 due to a fix for [this incompatibility](https://github.com/psf/black/issues/2964).

## Checklist


spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,70 @@ cdef class Doc:
...
+        doc = Doc(
+            vocab,
+            words=words,
+            tags=token_annotations["tag"]["values"],
+            pos=token_annotations["pos"]["values"],
+            morphs=token_annotations["morph"]["values"],
+            lemmas=token_annotations["lemma"]["values"],
+            heads=token_annotations["head"]["values"],
+            deps=token_annotations["dep"]["values"],
+        )
```
svlandeg, line 1491: You'll definitely need to define `spaces` as well - this stores whether or not a space should follow a word. You can create a failing unit test that shows the current behaviour is not correct:

Take a sentence `"This is just brilliant."`. After `to_json` and `from_json` it will currently become `"This is just brilliant ."`

svlandeg (commented):
Thanks for taking this on! It'll definitely be good to have this additional functionality available, and I love your attention to detail with the unit tests & docs :-) 

Some detailed comments below.

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,70 @@ cdef class Doc:
...
+
+        # Complement other document-level properties (cats, spans, ents).
+        doc.cats = doc_json.get("cats", doc.cats)
+
+        if "spans" in doc_json:
+            for span_group in doc_json["spans"]:
+                spans = []
+                for json_span in doc_json["spans"][span_group]:
+                    span = doc.char_span(json_span["start"], json_span["end"], json_span["label"], json_span["kb_id"])
```
svlandeg, line 1500: I would add an `assert span is not None` here, to guard against internal programming errors.

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,70 @@ cdef class Doc:
...
+                    if not doc.has_annotation("DEP"):
+                        for i, token in enumerate(span):
+                            token.is_sent_start = i == 0
```
svlandeg, line 1503: I don't understand the reasoning behind this part? If there are no sentence boundaries on the original `doc`, this part will assign some values to the tokens in the `spans` - but why?

website/docs/api/doc.md
```diff
@@ -478,6 +479,44 @@ Deserialize, i.e. import the document contents from a binary string.
...
+> doc = nlp("All we have to decide is what to do with the time that is given us.")
+> assert doc.to_json()["text"] == doc.text
+> ```
+
+| Name         | Description                                                                                                                                                                                                      |
+| ------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `underscore` | Optional list of string names of custom `doc._.attributes`. Attribute values need to be JSON-serializable. Values will be added to an `"_"` key in the data, e.g. `"_": {"foo": "bar"}`. ~~Optional[List[str]]~~ |
+| **RETURNS**  | The data in JSON format. ~~Dict[str, Any]~~                                                                                                                                                                      |
+
+## Doc.from_json {#from_json tag="method"}
```
svlandeg, line 499: We could add a `new="3.3"` here, or `new="3.3.1"`, depending on when we'll eventually merge this PR and whether it'll make the cut for 3.3 or not :-)

spacy/tokens/doc.pyx
```diff
@@ -1461,16 +1526,14 @@ cdef class Doc:
...
         if self.has_annotation("SENT_START"):
             sents = list(self.sents)
-            data["sents"] = [{"start": sent.start_char, "end": sent.end_char}
-                             for sent in sents]
+            data["sents"] = [{"start": sent.start_char, "end": sent.end_char} for sent in sents]
         if self.cats:
             data["cats"] = self.cats
         data["tokens"] = []
-        attrs = ["TAG", "MORPH", "POS", "LEMMA", "DEP"]
+        attrs = ["TAG", "MORPH", "POS", "LEMMA", "DEP", "SENT_START"]
```
adrianeboyd, line 1536: Let's not duplicate the sentence annotation, since there's some chance it will get out-of-sync. Let's keep `data["sents"]` above with character offsets and then convert them as necessary.

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -0,0 +1,155 @@
+import pytest
+from spacy.tokens import Doc, Span
+
+
+@pytest.fixture()
+def doc(en_vocab):
+    words = ["c", "d", "e"]
+    pos = ["VERB", "NOUN", "NOUN"]
+    tags = ["VBP", "NN", "NN"]
+    heads = [0, 0, 0]
```
adrianeboyd, line 10: Since 0 is the default, let's use at least one non-zero head for the test example?

adrianeboyd (commented):
Just a few comments on some small things I noticed.

We'll have to consider carefully to what extent `Doc.to_json` supports missing annotation and to what extent `Doc.from_json` does error-checking for ill-formed input.

* `ents` in this format doesn't support missing annotation
* the token-level annotation does support missing annotation
  * partial dependency annotation can get tricky, especially in terms of what it means for sentence boundaries

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -70,3 +71,52 @@ def test_doc_to_json_span(doc):
...
     assert json_doc["spans"]["test"][0]["start"] == 0
+
+
+def test_json_to_doc(doc):
+    new_doc = Doc.from_json(doc.vocab, doc.to_json())
+    new_tokens = [token for token in new_doc]
+    assert new_doc.text == doc.text
+    assert len(new_tokens) == len([token for token in doc])
+    assert new_doc.text == doc.text == "c d e "
+    assert len(new_tokens) == 3
```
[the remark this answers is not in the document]
rmitsch: Fixed in next commit.

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -70,3 +71,52 @@ def test_doc_to_json_span(doc):
...
+    assert new_tokens[0].pos_ == "VERB"
+    assert new_tokens[0].tag_ == "VBP"
+    assert new_tokens[0].dep_ == "ROOT"
```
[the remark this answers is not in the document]
rmitsch: Included in next commit. Using `new_tokens[0].head.idx == doc[0].head.idx` to compare the `head` info now, because `Token.__richcmp__()` also checks for equality of the heads' docs, which is not given in our case. Let me know if there's a better way to check for head equality :)

spacy/tokens/doc.pyx
rmitsch: Yes, as this is an instance method now, spaces are set differently. A unit test for space preservation will be added in the next commit.

spacy/tokens/doc.pyx
rmitsch: Included in next commit.

spacy/tokens/doc.pyx
rmitsch: I misunderstood how sentence boundaries are handled. This has been adjusted.

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -70,3 +71,52 @@ def test_doc_to_json_span(doc):
...
+        doc.vocab, doc.to_json(underscore=["json_test1", "json_test2"])
+    )
+    assert all([new_doc.has_extension(f"json_test{i}") for i in range(1, 3)])
+    assert new_doc._.json_test1 == "hello world"
+    assert new_doc._.json_test2 == [1, 2, 3]
+
+
+def test_json_to_doc_span(doc):
+    """Test that Doc.from_json() includes correct.spans."""
+    doc.spans["test"] = [Span(doc, 0, 2, "test"), Span(doc, 0, 1, "test")]
```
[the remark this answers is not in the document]
rmitsch: Included in the next commit.

website/docs/api/doc.md
rmitsch: Went with "3.3.1" - included in the next commit.

spacy/tokens/doc.pyx
rmitsch: Removed in next commit.

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,140 @@ cdef class Doc:
...
+                    if token_annotations[annotation_type]["values"] is None:
+                        token_annotations[annotation_type]["values"] = []
+                    token_annotations[annotation_type]["values"].append(token[annotation_type])
+
+        # Initialize doc instance.
+        start = 0
+        cdef const LexemeC* lex
+        cdef str orth_
+        cdef bint has_space
+        _, spaces = get_words_and_spaces(words, doc_json["text"])
```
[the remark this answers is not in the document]
rmitsch: Included in next commit.

spacy/tests/doc/test_json_doc_conversion.py
rmitsch: Included in next commit.

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,140 @@ cdef class Doc:
...
+                if pp not in parts_of_speech.IDS:
+                    raise ValueError(Errors.E1021.format(pp=pp))
+
+        headings = []
+        values = []
+        annotations = [
+            *[token_annotations[key]["values"] for key in annotation_types],
+            sent_starts
+        ]
+        possible_headings = [POS, HEAD, DEP, LEMMA, TAG, MORPH, SENT_START]
```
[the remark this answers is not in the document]
rmitsch: Included in next commit.

rmitsch:
@adrianeboyd Have we come to a conclusion regarding (1) missing annotation support yet and (2) error-checking?

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,140 @@ cdef class Doc:
...
+        for i, token in enumerate(doc_json["tokens"]):
+            has_space = spaces[i]
+            orth_ = doc_json["text"][token["start"]:token["end"]]
+            lex = self.vocab.get(self.mem, orth_)
+            self.push_back(lex, has_space)
+
+        # Set remaining token-level attributes via Doc.from_array().
+        if token_annotations["head"]["values"] is not None:
+            token_annotations["head"]["values"] = [
+                head - i if head is not None else 0 for i, head in enumerate(token_annotations["head"]["values"])
```
adrianeboyd, line 1499: I don't think that `head` or `dep` can be `None`? You end up with `"dep":""` or `"head":<id>` for what is effectively missing annotation, so you have rely to `"dep"` to see whether dependency annotation is set.

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,140 @@ cdef class Doc:
...
+        for a, annot in enumerate(annotations):
+            if annot is not None:
+                if len(annot) != len(words):
+                    raise ValueError(Errors.E189)
+                headings.append(possible_headings[a])
+                if annot is not token_annotations["head"]["values"]:
+                    values.extend(annot)
+        for value in values:
+            if value is not None:
+                self.vocab.strings.add(value)
```
adrianeboyd, line 1528: I think you can skip this if you switch to `self.vocab.strings.add(annot[i])` below?

adrianeboyd (commented):
Some small comments on the most recent version. I think we'll need some more tests for partial and ill-formed input. I'm still not sure what to do about partial dependency annotation (warn? refuse to handle?).

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,140 @@ cdef class Doc:
...
+            for span_group in doc_json.get("spans", {}):
+                spans = [
+                    self.char_span(span["start"], span["end"], span["label"], span["kb_id"])
+                    for span in doc_json["spans"][span_group]
+                ]
+                assert all([span is not None for span in spans])
+                self.spans[span_group] = spans
+
+        if "ents" in doc_json:
+            self.ents = [self.char_span(ent["start"], ent["end"], ent["label"]) for ent in doc_json["ents"]]
```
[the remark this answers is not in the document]
rmitsch: So line 1574 - `assert all([span is not None for span in spans])` - but with a custom error message in `errors.Errors`?

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,140 @@ cdef class Doc:
...
+        """Convert a JSON document generated by Doc.to_json() to a Doc.
+
+        vocab (Vocab): A storage container for lexical types.
+        RETURNS (Doc): A doc instance corresponding to the specified JSON representation.
+        """
+
+        ### Token-level properties ###
+
+        words = []
+        annotation_types = ("pos", "head", "dep", "lemma", "tag", "morph")
```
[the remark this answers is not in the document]
rmitsch: Agreed. As far as I can tell, there isn't a good reason to use the lowercase strings. I suggest using the variables everywhere in `from_json()`.

spacy/tokens/doc.pyx
rmitsch: `to_json()` generates the token annotation fields as follows:
```python
if include_annotation["DEP"]:
    token_data["dep"] = token.dep_
    token_data["head"] = token.head.i
```
In my understanding: if the coument to be converted to JSON doesn't have the DEP annotation, the token data won't include either "dep" nor "head" entries. Consequently at this point in `from_json()` `token_annotations["head"]["values"]` respectively `token_annotations["dep"]["values"]` will be `None`, since they couldn't been collected from the document JSON.

spacy/tokens/doc.pyx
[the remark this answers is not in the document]
rmitsch: Let me know if the latest changes make sense to you.

spacy/tokens/doc.pyx
```diff
@@ -1451,6 +1451,140 @@ cdef class Doc:
...
+
+        # Set remaining token-level attributes via Doc.from_array().
+        if token_annotations["head"]["values"] is not None:
+            token_annotations["head"]["values"] = [
+                head - i if head is not None else 0 for i, head in enumerate(token_annotations["head"]["values"])
+            ]
+        if token_annotations["dep"]["values"] is not None:
+            MISSING_DEP_ = self.vocab.strings[MISSING_DEP]
+            token_annotations["dep"]["values"] = [
+                dep if dep is not None else MISSING_DEP_ for dep in token_annotations["dep"]["values"]
```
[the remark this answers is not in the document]
rmitsch: Let me know if the latest changes make sense to you.

spacy/tokens/doc.pyx
rmitsch: Let me know if the latest changes make sense to you.

spacy/tokens/doc.pyx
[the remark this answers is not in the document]
rmitsch: Suggestion for using `[POS, ...]` instead of string representations in latest commit. There is still an initial mapping to the string representations, but we won't be able to bypass that if we don't wanna change the output of `Doc.to_json()`.

adrianeboyd:
It looks like some unintentional changes crept in (formatting for `lang/vi`, modifications to `Doc.sents`)?

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+                spans = []
+                for span in doc_json["spans"][span_group]:
+                    char_span = self.char_span(span["start"], span["end"], span["label"], span["kb_id"])
+                    if char_span is None:
+                        raise ValueError(Errors.E1038)
+                    spans.append(char_span)
+                self.spans[span_group] = spans
+
+        if "ents" in doc_json:
+            self.ents = [self.char_span(ent["start"], ent["end"], ent["label"]) for ent in doc_json["ents"]]
```
adrianeboyd, line 1595: The same `char_span -> None` error can happen here. (Same error message, just with a parameter for `spans`/`ents`.)

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+                            if attrs.ndim == 1:
+                                attrs[i] = morph_key
+                            else:
+                                attrs[i, j] = morph_key
+                    else:
+                        for i in range(len(words)):
+                            if attrs.ndim == 1:
+                                attrs[i] = self.vocab.strings[annot[i]]
+                            else:
+                                attrs[i, j] = self.vocab.strings[annot[i]]
```
adrianeboyd, line 1569: You do need `self.vocab.strings.add(annot[i])` here instead. It's confusing, but `self.vocab.strings[x]` only gets the hash, it doesn't add it to the string store.

spacy/tokens/doc.pyx
adrianeboyd: I think that `token_annotations["dep"]["values"]` is `None`, but the individual values (if they are set) have no way of being `None`. `"dep"` will always be a string and `"head"` will always be an int.

spacy/tokens/doc.pyx
rmitsch: So 
```python
if attrs.ndim == 1:
    attrs[i] = self.vocab.strings[annot[i]]
else:
    attrs[i, j] = self.vocab.strings[annot[i]]
self.vocab.strings.add(annot[i])
```
despite `attrs` being passed to `.from_array()` immediately after?

rmitsch:
> It looks like some unintentional changes crept in (formatting for `lang/vi`, modifications to `Doc.sents`)?

Have been reverted, thanks for the hint.

spacy/tokens/doc.pyx
rmitsch: I see, so I a proper check would be like
```python
head - i if token_annotations["dep"]["values"] is != "" else 0 for i, head in enumerate(token_annotations["head"]["values"]) 
```
Also - this is handled the same way in `__init__()`. Should it be changed accordingly there too?

spacy/tokens/doc.pyx
adrianeboyd: It can look more like `morph_key` above? The `attrs` are passed as `uint64` and if the strings weren't already added to the string store there's no way to recover them.

adrianeboyd: (And it's not your fault, but this use of `attrs` as a variable name here is extremely confusing.)

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,152 @@ cdef class Doc:
...
+
+        vocab (Vocab): A storage container for lexical types.
+        RETURNS (Doc): A doc instance corresponding to the specified JSON representation.
+        """
+
+        ### Token-level properties ###
+
+        words = []
+        # Map annotation type IDs to their string equivalents.
+        annot_types = {POS: "pos", HEAD: "head", DEP: "dep", LEMMA: "lemma", TAG: "tag", MORPH: "morph"}
```
adrianeboyd, line 1476:
```suggestion
        annot_types = {t: self.vocab.strings[t].lower() for t in (POS, HEAD, DEP, LEMMA, TAG, MORPH)}
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,152 @@ cdef class Doc:
...
+        token_annotations = {
+            # For each annotation type: store (1) annotation type required to include it, (2) list of annotations for
+            # this type.
+            annot_type_id: {"req": annot_type_id if annot_type_id != HEAD else DEP, "values": None}
+            for annot_type_id in annot_types
+        }
```
adrianeboyd, line 1482: I think the code will be more readable if this is split into two separate dicts, so that we don't have `["values"]` everywhere below.

spacy/tokens/doc.pyx
rmitsch: Got it. Is the check for an empty string in individual `"dep"` values as shown in the last comment what you meant?

adrianeboyd:
The comments and change requests have gotten pretty hard to follow above. The main things to refactor right now are:

* reshape once instead of using `ndim ==1` special cases
* rename `headings` to `attrs` and `attrs` to `array` (or maybe `values` if you prefer?).
* refactor to two separate token-info dicts to remove repeated `["values"]`

Then we can figure out how to handle partial dependency annotation.

spacy/tokens/doc.pyx
rmitsch: Handled the vocabulary addition like `morph_key` now:
```python
vocab_key = self.vocab.strings.add(annot[i])
if array.ndim == 1:
    array[i] = vocab_key
else:
    array[i, j] = vocab_key
```
Is that what you meant?

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,162 @@ cdef class Doc:
...
+                                array[i] = morph_key
+                            else:
+                                array[i, j] = morph_key
+                    else:
+                        for i in range(len(words)):
+                            vocab_key = self.vocab.strings.add(annot[i])
+                            if array.ndim == 1:
+                                array[i] = vocab_key
+                            else:
+                                array[i, j] = vocab_key
```
adrianeboyd, line 1579: I think this can be simplified to:

```suggestion
                    if annot is token_annotations[HEAD]:
                        for i in range(len(words)):
                            array[i, j] = annot[i]
                    elif annot is token_annotations[MORPH]:
                        for i in range(len(words)):
                            array[i, j] = self.vocab.morphology.add(token_annotations[MORPH][i])
                    else:
                        for i in range(len(words)):
                            array[i, j] = self.vocab.strings.add(annot[i])
```

spacy/tokens/doc.pyx
adrianeboyd: `token_annotations["dep"]["values"]` is checking the list value, so I think you meant `token_annotations["dep"]["values"][i] != ""` (not `is !=`)? (I considered this to be part of "how to handle partial dependency annotation".)

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+        """
+
+        ### Token-level properties ###
+
+        words = []
+        # Map annotation type IDs to their string equivalents.
+        annot_types = {t: self.vocab.strings[t].lower() for t in (POS, HEAD, DEP, LEMMA, TAG, MORPH)}
+        token_annotations_reqs = {
+            annot_type_id: annot_type_id if annot_type_id != HEAD else DEP for annot_type_id in annot_types
+        }
```
adrianeboyd, line 1479: I think we can skip this if we assume no partial token annotation and do the `HEAD`/`DEP` checks below.

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+        token_annotations = {annot_type_id: None for annot_type_id in annot_types}
```
adrianeboyd, line 1480: Can you refactor this so that you only have keys for the annotation that's present instead of using `None`?

I think this makes the checks below clearer and doesn't potentially confuse `None` vs. `[]` which will both evaluate to `False` but not mean the same thing.

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+
+        vocab (Vocab): A storage container for lexical types.
+        RETURNS (Doc): A doc instance corresponding to the specified JSON representation.
+        """
+
+        ### Token-level properties ###
+
+        words = []
+        # Map annotation type IDs to their string equivalents.
+        annot_types = {t: self.vocab.strings[t].lower() for t in (POS, HEAD, DEP, LEMMA, TAG, MORPH)}
```
adrianeboyd, line 1476: I think "types" is a bit confusing as a name here and would prefer `attr` or `attr_id` for the attr symbols like `POS` (we have the type `attr_id_t` for these types in `attrs.pxd`).

```suggestion
        token_attrs = {
            t: self.vocab.strings[t].lower()
            for t in (POS, HEAD, DEP, LEMMA, TAG, MORPH)
        }
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+        cdef str orth_
+        cdef bint has_space
+        reconstructed_words, spaces = get_words_and_spaces(words, doc_json["text"])
+        assert words == reconstructed_words
+
+        for i, token in enumerate(doc_json["tokens"]):
+            has_space = spaces[i]
+            orth_ = doc_json["text"][token["start"]:token["end"]]
+            lex = self.vocab.get(self.mem, orth_)
+            self.push_back(lex, has_space)
```
adrianeboyd, line 1513:
```suggestion
        for word, has_space in zip(words, spaces):
            lex = self.vocab.get(self.mem, word)
            self.push_back(lex, has_space)
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+
+        # Set remaining token-level attributes via Doc.from_array().
+        if token_annotations[HEAD] is not None:
```
adrianeboyd, line 1516: All these would become `if HEAD not in token_annotations` etc.

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+                    raise ValueError(Errors.E1021.format(pp=pp))
+
+        attrs = []
+        annotations = [token_annotations[key] for key in annot_types]
+        possible_headings = [POS, HEAD, DEP, LEMMA, TAG, MORPH]
+        for a, annot in enumerate(annotations):
+            if annot is not None:
+                if len(annot) != len(words):
+                    raise ValueError(Errors.E189)
+                attrs.append(possible_headings[a])
```
adrianeboyd, line 1544:
```suggestion
        for attr in token_attrs.keys():
            if attr in token_annotations:
                if len(token_annotations[attr]) != len(words):
                    raise ValueError(Errors.E189)
                attrs.append(attr)
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+                    if annot is token_annotations[HEAD]:
+                        for i in range(len(words)):
+                            array[i, j] = annot[i]
+                    elif annot is token_annotations[MORPH]:
+                        for i in range(len(words)):
+                            array[i, j] = self.vocab.morphology.add(token_annotations[MORPH][i])
+                    else:
+                        for i in range(len(words)):
+                            array[i, j] = self.vocab.strings.add(annot[i])
+                    j += 1
```
adrianeboyd, line 1564:
```suggestion
            for j, (attr, annot) in enumerate(token_annotations.items()):
                if attr is HEAD:
                    for i in range(len(words)):
                        array[i, j] = annot[i]
                elif attr is MORPH:
                    for i in range(len(words)):
                        array[i, j] = self.vocab.morphology.add(annot[i])
                else:
                    for i in range(len(words)):
                        array[i, j] = self.vocab.strings.add(annot[i])
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+        for i, token in enumerate(doc_json["tokens"]):
+            has_space = spaces[i]
+            orth_ = doc_json["text"][token["start"]:token["end"]]
+            lex = self.vocab.get(self.mem, orth_)
+            self.push_back(lex, has_space)
+
+        # Set remaining token-level attributes via Doc.from_array().
+        if token_annotations[HEAD] is not None:
+            token_annotations[HEAD] = [
+                head - i if head is not None else 0 for i, head in enumerate(token_annotations[HEAD])
```
adrianeboyd, line 1518:
```suggestion
                head - i for i, head in enumerate(token_annotations[HEAD])
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+            ]
+        if token_annotations[DEP] is not None:
+            deps = []
+            for dep in token_annotations[DEP]:
+                if dep is None:
+                    raise ValueError(Errors)
+                deps.append(dep)
+            token_annotations[DEP] = deps if len(deps) else None
```
adrianeboyd, line 1526: I'm not sure we'll need this exact check in the end, but I think you can simplify this to just check for `None`:

```suggestion
        if DEP in token_annotations[DEP]:
            if None in token_annotations[DEP]:
                raise ValueError(Errors.E###)
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+        if "ents" in doc_json:
+            ents = []
+            for ent in doc_json["ents"]:
+                char_span = self.char_span(ent["start"], ent["end"], ent["label"])
+                if char_span is None:
+                    raise ValueError(Errors.E1038.format(obj="entity"))
+                ents.append(char_span)
+            self.ents = ents
+
+        # Add custom attributes.
```
adrianeboyd, line 1597: We need to document better that this is only for `Doc` extensions. I suspect adding support for `Token` and `Span` extensions will be needed, but not in this PR.

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+
+        # Gather token-level properties.
+        for token in doc_json["tokens"]:
+            words.append(doc_json["text"][token["start"]:token["end"]])
+            for annot_type_id in token_annotations:
+                # Assuming all tokens have exactly the same set of attributes.
+                if annot_types[token_annotations_reqs[annot_type_id]] in token:
+                    if token_annotations[annot_type_id] is None:
+                        token_annotations[annot_type_id] = []
+                    token_annotations[annot_type_id].append(token[annot_types[annot_type_id]])
```
adrianeboyd, line 1490: I think it gets confusing if we use `token` for something that's not a `Token`.

```suggestion
        for token_json in doc_json["tokens"]:
            words.append(doc_json["text"][token_json["start"]:token_json["end"]])
            for attr, attr_json in token_attrs.items():
                if attr_json in token_json:
                    if token_json["id"] == 0 and attr not in token_annotations:
                        token_annotations[attr] = [] 
                    elif attr not in token_annotations:
                        raise ValueError("Something about how partial token annotation is not supported.")
                    token_annotations[attr].append(token_json[attr_json])
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+            if token_annotations[annot_type_id]:
+                if len(token_annotations[annot_type_id]) != len(words):
+                    inconsistent_props.add(annot_types[annot_type_id])
+        if len(inconsistent_props):
+            raise ValueError(Errors.E1039.format(inconsistent_props=inconsistent_props))
+
+        # Initialize doc instance.
+        start = 0
+        cdef const LexemeC* lex
+        cdef str orth_
```
adrianeboyd, line 1504:
```suggestion
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+                    else:
+                        for i in range(len(words)):
+                            array[i, j] = self.vocab.strings.add(annot[i])
+                    j += 1
+            self.from_array(attrs, array)
+
+        ### Span/document properties ###
+
+        # Complement other document-level properties (cats, spans, ents).
+        self.cats = doc_json.get("cats", self.cats)
```
adrianeboyd, line 1570: I don't think we want to have any current doc attributes as defaults.

```suggestion
        self.cats = doc_json.get("cats", {})
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+
+        # Set sentence boundaries, if dependency parser not available but sentences are specified in JSON.
+        if not self.has_annotation("DEP"):
+            for sent in doc_json.get("sents", {}):
+                for t in self.char_span(sent["start"], sent["end"]):
+                    self[t.i].is_sent_start = t.i == 0
+
+        if "spans" in doc_json:
+            for span_group in doc_json.get("spans", {}):
```
adrianeboyd, line 1579:
```suggestion
            for span_group in doc_json.get["spans"]:
```

Or the spans/ents could be unindented with `for span_group in doc_json.get("spans", {})`?

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,146 @@ cdef class Doc:
...
+            deps = []
+            for dep in token_annotations[DEP]:
+                if dep is None:
+                    raise ValueError(Errors)
+                deps.append(dep)
+            token_annotations[DEP] = deps if len(deps) else None
+
+        if token_annotations[DEP] and not token_annotations[HEAD]:
+            token_annotations[HEAD] = [0] * len(token_annotations[DEP])
+        if token_annotations[HEAD] and not token_annotations[HEAD]:
```
adrianeboyd, line 1530:
```suggestion
        if HEAD in token_annotations and not DEP in token_annotations:
```

adrianeboyd (commented):
Let me make a few more suggestions for refactoring so the naming is more consistent with the codebase, so the dicts/checks for the annotation are a little easier to follow, and to simplify and fix minor bugs in a few places.

adrianeboyd:
I think we should consider adding a schema to `spacy.schemas` for this format.

spacy/tokens/doc.pyx
rmitsch: Do we have an issue for this already or should we create one?

spacy/tokens/doc.pyx
rmitsch: Did you mean `doc_json.get("spans")` or `doc_json["spans"]`?

adrianeboyd: I meant either:

```python
if "spans" in doc_json:
    for span_group in doc_json["spans"]:
```

or

```python
for span_group in doc_json.get("spans", {}):
```

(and unindent one level)

The second version looks more like `"cats"` and `_`, so I'd suggest that?

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,135 @@ cdef class Doc:
                 remove_label_if_necessary(attributes[i])
                 retokenizer.merge(span, attributes[i])
 
+    def from_json(self, doc_json):
+        """Convert a JSON document generated by Doc.to_json() to a Doc.
+
+        vocab (Vocab): A storage container for lexical types.
+        RETURNS (Doc): A doc instance corresponding to the specified JSON representation.
+        """
```
adrianeboyd, line 1470: I think we can add an option `, *, validate=False` option to use the schema for additional validation and then this is stable enough that you can go head and also update the docstring.

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1462,135 @@ cdef class Doc:
...
+
+        # Set remaining token-level attributes via Doc.from_array().
+        if HEAD in token_annotations:
+            token_annotations[HEAD] = [
+                head - i for i, head in enumerate(token_annotations[HEAD])
+            ]
+        if DEP in token_annotations:
+            if DEP in token_annotations[DEP]:
+                if None in token_annotations[DEP]:
+                    raise ValueError(Errors.E1040)
```
adrianeboyd, line 1519: I think we can skip this and let the validation handle it.

adrianeboyd (commented):
The schema is a great addition and I think we can add optional validation to this method (similar to the various matcher pattern validation) so that users who run into issues can validate, but it's not done by default for speed reasons.

rmitsch:
Commens/suggestions have been addressed.

spacy/errors.py
```diff
@@ -913,6 +913,10 @@ class Errors(metaclass=ErrorsWithCodes):
     E1034 = ("Node index {i} out of bounds ({length})")
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
+    E1037 = ("Annotation is ill-formed.")
```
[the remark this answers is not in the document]
adrianeboyd: Sorry, I meant we can use this error code, not this exact message. Can we provide useful information from the validation to the user in the error message?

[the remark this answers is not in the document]
rmitsch: Passing on the message from `schema.validate()` now in case of an ill-formed object. If you feel like something else might make more sense here, feel free to reopen the conversation.

spacy/errors.py
```diff
@@ -913,6 +913,12 @@ class Errors(metaclass=ErrorsWithCodes):
     E1034 = ("Node index {i} out of bounds ({length})")
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
+    E1037 = ("Document JSON representation doesn't adhere to schema: {message}")
+    E1038 = ("The {obj} start or end annotations (start: {start}, end: {end}) "
+             "could not be aligned to token boundaries.")
+    E1039 = ("Doc.from_json requires all tokens to have the same attributes. "
+             "Some tokens do not contain annotation for: {partial_attrs}")
+    E1040 = ("DEP attribute mustn't be None in specified token annotations.")
```
adrianeboyd, line 921:
```suggestion
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1463,132 @@ cdef class Doc:
                 remove_label_if_necessary(attributes[i])
                 retokenizer.merge(span, attributes[i])
 
+    def from_json(self, doc_json, *, validate=False):
+        """Convert a JSON document generated by Doc.to_json() to a Doc.
+
+        vocab (Vocab): A storage container for lexical types.
```
adrianeboyd, line 1469: This is still out-of-date?

spacy/errors.py
```diff
@@ -913,6 +913,11 @@ class Errors(metaclass=ErrorsWithCodes):
     E1034 = ("Node index {i} out of bounds ({length})")
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
+    E1037 = ("Document JSON representation doesn't adhere to schema: {message}")
```
adrianeboyd, line 916: To make this more similar to existing error messages:

```suggestion
    E1037 = ("Invalid JSON input: {message}")
```

website/docs/api/doc.md
```diff
@@ -481,6 +481,43 @@ Deserialize, i.e. import the document contents from a binary string.
...
+> from spacy.tokens import Doc
+> doc = nlp("All we have to decide is what to do with the time that is given us.")
+> doc_json = doc.to_json()
+> deserialized_doc = Doc(nlp.vocab).from_json(doc_json)
+> assert deserialized_doc.text == doc.text == doc_json["text"]
+> ```
+
+| Name        | Description                                                                |
+| ----------- | -------------------------------------------------------------------------- |
+| `doc_json`  | The Doc data in JSON format. ~~Dict[str, Any]~~                            |
```
adrianeboyd, line 518: Also add `validate` here.

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -0,0 +1,191 @@
...
+    assert new_doc.text == doc.text == "c d e "
+    assert len(new_tokens) == len([token for token in doc]) == 3
+    assert new_tokens[0].pos == doc[0].pos
+    assert new_tokens[0].tag == doc[0].tag
+    assert new_tokens[0].dep == doc[0].dep
+    assert new_tokens[0].head.idx == doc[0].head.idx
+    assert new_tokens[0].lemma == doc[0].lemma
+    assert len(new_doc.ents) == 1
+    assert new_doc.ents[0].start == 1  # character offset!
+    assert new_doc.ents[0].end == 2  # character offset!
```
adrianeboyd, line 112: I realize that these comments weren't from your PR, but let's go ahead and remove them since they're incorrect.

```suggestion
    assert new_doc.ents[0].start == 1
    assert new_doc.ents[0].end == 2
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1463,132 @@ cdef class Doc:
                 remove_label_if_necessary(attributes[i])
                 retokenizer.merge(span, attributes[i])
 
+    def from_json(self, doc_json, *, validate=False):
+        """Convert a JSON document generated by Doc.to_json() to a Doc.
+
+        doc_json (Dict): JSON representation of doc object to restore.
```
adrianeboyd, line 1469:
```suggestion
        doc_json (Dict): JSON representation of doc object to load.
```

spacy/tokens/doc.pyx
```diff
@@ -1462,6 +1463,132 @@ cdef class Doc:
...
+        validate (bool): Flag indicating whether the the schema of `doc_json` should be validated.
```
adrianeboyd, line 1470:
```suggestion
        validate (bool): Whether to validate `doc_json` against the expected schema.
            Defaults to False.
```

website/docs/api/doc.md
```diff
@@ -481,6 +481,44 @@ Deserialize, i.e. import the document contents from a binary string.
...
+> doc_json = doc.to_json()
+> deserialized_doc = Doc(nlp.vocab).from_json(doc_json)
+> assert deserialized_doc.text == doc.text == doc_json["text"]
+> ```
+
+| Name        | Description                                                                                                              |
+| ----------- | ------------------------------------------------------------------------------------------------------------------------ |
+| `doc_json`  | The Doc data in JSON format. ~~Dict[str, Any]~~                                                                          |
+| `validate`  | Flag indicating whether the the schema of `doc_json` should be validated against `spacy.schemas.DocJSONSchema`. ~~bool~~ |
+| **RETURNS** | A doc instance corresponding to the specified JSON representation. ~~Doc~~                                               |
```
adrianeboyd, line 520: (plus reformatting)

```suggestion
| `doc_json`  | The Doc data in JSON format from [`Doc.to_json`](#to_json). ~~Dict[str, Any]~~                                                                          |
| _keyword-only_ | |
| `validate` | Whether to validate the JSON input against the expected schema for detailed debugging. Defaults to `False`. ~~bool~~ |
| **RETURNS** | A `Doc` corresponding to the provided JSON. ~~Doc~~                                               |
```

website/docs/api/doc.md
```diff
@@ -481,6 +481,45 @@ Deserialize, i.e. import the document contents from a binary string.
...
+> #### Example
+>
+> ```python
+> doc = nlp("All we have to decide is what to do with the time that is given us.")
+> assert doc.to_json()["text"] == doc.text
+> ```
+
+| Name         | Description                                                                                                                                                                                                      |
+| ------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `underscore` | Optional list of string names of custom `doc._.attributes`. Attribute values need to be JSON-serializable. Values will be added to an `"_"` key in the data, e.g. `"_": {"foo": "bar"}`. ~~Optional[List[str]]~~ |
```
adrianeboyd, line 498:
```suggestion
| `underscore` | Optional list of string names of custom `Doc` attributes. Attribute values need to be JSON-serializable. Values will be added to an `"_"` key in the data, e.g. `"_": {"foo": "bar"}`. ~~Optional[List[str]]~~ |
```

adrianeboyd:
I think this is nearly ready to go. Can you resolve the conflicts related to the error numbers?

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -0,0 +1,191 @@
...
+    """Test that Doc.from_json() includes correct.sents."""
+    for test_doc in (doc, doc_without_deps):
+        json_doc = test_doc.to_json()
+        new_doc = Doc(doc.vocab).from_json(json_doc, validate=True)
+        assert [sent.text for sent in doc.sents] == [
+            sent.text for sent in new_doc.sents
+        ]
+        assert [token.is_sent_start for token in doc] == [
+            token.is_sent_start for token in new_doc
+        ]
```
adrianeboyd, line 158:
```suggestion
        assert [sent.text for sent in test_doc.sents] == [
            sent.text for sent in new_doc.sents
        ]
        assert [token.is_sent_start for token in test_doc] == [
            token.is_sent_start for token in new_doc
        ]
```

spacy/tokens/doc.pyx
```diff
@@ -1467,6 +1470,133 @@ cdef class Doc:
...
+        ### Span/document properties ###
+
+        # Complement other document-level properties (cats, spans, ents).
+        self.cats = doc_json.get("cats", {})
+
+        # Set sentence boundaries, if dependency parser not available but sentences are specified in JSON.
+        if not self.has_annotation("DEP"):
+            for sent in doc_json.get("sents", {}):
+                for t in self.char_span(sent["start"], sent["end"]):
+                    self[t.i].is_sent_start = t.i == 0
```
adrianeboyd, line 1571: The token indices in the span (`t.i`) aren't relative to the span, so this is not working (test corrected above to show failure).

adrianeboyd (changes requested):
It looks like there was a bug in the `doc_without_deps` test for sentence boundaries, and there's still a bug in setting sentences from JSON `sents`.

spacy/tokens/doc.pyx
```diff
@@ -1467,6 +1470,133 @@ cdef class Doc:
...
+        ### Span/document properties ###
+
+        # Complement other document-level properties (cats, spans, ents).
+        self.cats = doc_json.get("cats", {})
+
+        # Set sentence boundaries, if dependency parser not available but sentences are specified in JSON.
+        if not self.has_annotation("DEP"):
+            for sent in doc_json.get("sents", {}):
+                for t in self.char_span(sent["start"], sent["end"]):
+                    self[t.i].is_sent_start = t.idx == sent["start"]
```
adrianeboyd, line 1571: Ah, let's also check for invalid spans here, since this could return `None`?

I'm also tempted to suggest the less compact version that looks more like `span[0].is_sent_Start = True`, `span[1:].is_sent_start = False`?

spacy/tokens/doc.pyx
```diff
@@ -1467,6 +1470,138 @@ cdef class Doc:
...
+
+        # Set sentence boundaries, if dependency parser not available but sentences are specified in JSON.
+        if not self.has_annotation("DEP"):
+            for sent in doc_json.get("sents", {}):
+                char_span = self.char_span(sent["start"], sent["end"])
+                if char_span is None:
+                    raise ValueError(Errors.E1039.format(obj="sentence", start=sent["start"], end=sent["end"]))
+                self[char_span[0].i].is_sent_start = True
+                for t in self.char_span(sent["start"], sent["end"])[1:]:
+                    self[t.i].is_sent_start = False
```
adrianeboyd, line 1575: A `Span` is just a view on the tokens in the doc, so you can modify the span's tokens directly without more indexing: 

```suggestion
                char_span[0].is_sent_start = True
                for token in char_span[1:]:
                    token.is_sent_start = False
```

(This does rely on the fact that `Doc.char_span` returns `None` instead of an empty span for cases with no tokens. Empty spans are possible in general, but I think this is okay.)

rmitsch: Wasn't aware of that. Thanks!

spacy/tokens/doc.pyx
```diff
@@ -1467,6 +1470,138 @@ cdef class Doc:
...
+
+        # Gather token-level properties.
+        for token_json in doc_json["tokens"]:
+            words.append(doc_json["text"][token_json["start"]:token_json["end"]])
+            for attr, attr_json in token_attrs.items():
+                if attr_json in token_json:
+                    if token_json["id"] == 0 and attr not in token_annotations:
+                        token_annotations[attr] = []
+                    elif attr not in token_annotations:
+                        raise ValueError(Errors.E1041)
```
adrianeboyd, line 1503:
```suggestion
                        raise ValueError(Errors.E1040.format(partial_attrs=attr))
```

spacy/errors.py
```diff
@@ -919,6 +919,12 @@ class Errors(metaclass=ErrorsWithCodes):
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
     E1037 = ("Invalid attribute value '{attr}'.")
+    E1038 = ("Invalid JSON input: {message}")
+    E1039 = ("The {obj} start or end annotations (start: {start}, end: {end}) "
+             "could not be aligned to token boundaries.")
+    E1040 = ("Doc.from_json requires all tokens to have the same attributes. "
+             "Some tokens do not contain annotation for: {partial_attrs}")
+    E1041 = ("Partial token annotation is not supported.")
```
adrianeboyd, line 927:
```suggestion
```

adrianeboyd (commented):
I think E1040 is a clearer error message for users and still covers the same issue well enough.

adrianeboyd (approved):

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -0,0 +1,191 @@
...
+def test_json_to_doc_attribute_consistency(doc):
+    """Test that Doc.from_json() raises an exception if tokens don't all have the same set of properties."""
+    doc_json = doc.to_json()
+    doc_json["tokens"][1].pop("morph")
+    with pytest.raises(ValueError):
+        Doc(doc.vocab).from_json(doc_json)
+
+
+def test_json_to_doc_validation_error(doc):
+    """Test that Doc.from_json() raises an exception if tokens don't all have the same set of properties."""
```
adrianeboyd, line 187:
```suggestion
    """Test that Doc.from_json() raises an exception when validating invalid input."""
```

svlandeg (commented):
Nice work, @rmitsch and @adrianeboyd !

PR #10707, opened 2022-04-25
Fix: De/Serialize `SpanGroups` including the SpanGroup keys

### LINKED ISSUE #10685: SpanGroup()s of a SpanGroups() are lost upon de/serialization--if having the same .name
## Problem:

If a `SpanGroup.name` conflicts with another SpanGroup.name in a `SpanGroups` object--but the `SpanGroup`s are values of different keys of the SpanGroups object--then when the SpanGroups object is serialized and then deserialized, only 1 of the `SpanGroup`s will be present in the `SpanGroups` object.

This occurs when e.g. serializing/deserializing `Doc`s via `DocBin`--whether using DocBin's `get_docs()` or `to_bytes()`/`from_bytes()`.

A Doc's `SpanGroups` might be in such a state when (at least):
* a user creates a `SpanGroup` object and sets the `.name` attribute to the name of another SpanGroup within the Doc's `SpanGroups`; or
* a user changes an existing `SpanGroup`'s `.name` to the name of another SpanGroup within the Doc's `SpanGroups`.

### How to reproduce the behaviour:

(Executed with:)
* **spaCy version:** 3.2.4
* **Platform:** Linux-5.17.3-x86_64-with-glibc2.35
* **Python version:** 3.10.4

For example:

```python
import spacy
from spacy.tokens import SpanGroup, DocBin
nlp = spacy.blank("en")

doc = nlp("Will it blend?")
doc.spans['test'] = [doc[0:1]]
... # perhaps at some other point in code:
spangroup = SpanGroup(doc, name='test', spans=[doc[1:2]])
doc.spans['test2'] = spangroup

# Now choose your own adventure:
# ***A:
>>> doc.spans
{'test': [Will], 'test2': [it]}
>>> doc.spans.from_bytes(doc.spans.to_bytes())
{'test': [it]}

# ***B:
# (`doc.spans` is defined as it was prior to adventure "A":)
>>> db = DocBin(docs=[doc])
>>> redocs = list(db.get_docs(nlp.vocab))
>>> redocs[0].spans
{'test': [it]}

# and consequently:
>>> redb = db.from_bytes(db.to_bytes())
>>> redoc = list(redb.get_docs(nlp.vocab))[0]
>>> redoc.spans
{'test': [it]}

# which similarly occurs for `DocBin`'s `to_disk` and `from_disk`.
```

And it's also problematic if e.g. (as a special case):
```python
doc.spans['bad']  = SpanGroup(doc, spans=[doc[0:1]])
doc.spans['user'] = SpanGroup(doc, spans=[doc[1:2]])
```
since `SpanGroup`s without a given 'name' all default to the name "", and if/when they're set as values in a Doc's `SpanGroups`, only 1 will be loaded/set upon deserialization.

While I wouldn't expect most users to write code exactly like the simple/direct examples above, they do have the power to create "naked" `SpanGroup` objects (i.e. via `SpanGroup()`--outside of a `doc.spans`) and use/modify/assign them (to a `doc.spans`) as desired.

## Cause:

`SpanGroups.to_bytes` serializes its `SpanGroup`s as a list, and doesn't include their corresponding keys (in the `SpanGroups` container). When deserializing a `SpanGroups` object, each `SpanGroup` in the list is assigned as the value of the key corresponding to its `.name`. When multiple `SpanGroup`s have the same `.name`, only the last `SpanGroup` with that `.name` will be present in the `SpanGroups` object.

`DocBin.add()` serializes `SpanGroups` using their `.to_bytes()`, so de/serialization of `Doc`s this way inherits this behavior.

## Solution:

Checks are already in place for ensuring agreement between a `SpanGroup`'s `Doc` and a `SpanGroups`' `Doc`.
To prevent the above issues, we may want to at least have:
1. (both of the following:)
    1. `SpanGroups` ensure that no added `SpanGroup` has the same name as another SpanGroup in the SpanGroups object. This could be done by e.g.:
        1. iterate over the `SpanGroup`s, or cache their names (probably: neither of which you want to do); or:
        2. require that a `SpanGroup.name` matches the key it is the value of in `SpanGroups`; in other words, maintaining the invariant:
            ```python
            for name, span_group in doc.spans.items():
                assert span_group.name == name
            ```
    2. `SpanGroup`s prevent changing their `.name` (e.g. via attribute assignment)--*if the `SpanGroup` object is a value of its `Doc`'s `SpanGroups` object* (which `SpanGroup`s currently don't have to be)--:
        1. to a name of another SpanGroup in its same `SpanGroups` object; or:
        2. (if we required bullet "1.i.b" (noted above):) to any name other than the key it is the value of in its `Doc`'s `.spans`.
2. (or:) Change how `SpanGroups` are serialized/deserialized:
    * Include the key names.
    * though this will still allow multiple SpanGroup()s within a `SpanGroups` to have the same `.name`. (This may not be a problem, but I haven't done an exhaustive search of the codebase to see how `SpanGroup.name`s are used--other than for de/serialization.)

Those are the two general approaches that came to mind to address this, but let me know what you think.
I'm leaning towards option 2, because it's both simpler and I think more in line with how you're treating `SpanGroup`s currently. I don't think I'd have a problem implementing that solution, if you want to go forward with it or something like it.

single-fingal:
## Description

Fixes #10685

This de/serializes `SpanGroups` objects as dicts, so that the keys corresponding to `SpanGroup`s are included.

This prevents the loss of `SpanGroup`s that have the same `.name` as other `SpanGroup`s within the same `SpanGroups` object (upon de/serialization of the `SpanGroups`).

### Notes / Questions

I figured I'd offer a solution to the problem. This pull request implements the "option 2" solution mentioned in #10685.

1. I was wondering if we should output a warning if we deserialized a `SpanGroups` that has more than 1 `SpanGroup` with the same `.name` (which would still occur when deserializing `SpanGroups` that were serialized in the old format). That's up to you, but in the 4th commit of this PR, I included what that might look like.
    * The warning text (and its identifer--currently `Warnings.W119`) are certainly open for changes, however.
        * I considered if it would be worth including the `Doc` object's text, as a way to identify what Doc the warning applies to, but from a glance at the existing warning messages, that kind of verbosity seemed a bit out of place. Maybe it's not, though.
    * Also: when deserializing, I'm not e.g. counting the number of times a given `.name` is duplicated; currently, I'm just checking, for each SpanGroup in the serialized `SpanGroups`, if its `.name` is already a key of the `SpanGroups`. The performance of that check seems pretty negligible. And my guess is that that additional information--the number of duplicated `.name`s--wouldn't be particularly useful for users.
2. I added tests in a new module, in `tests/serialize/test_serialize_spangroups.py`. There wasn't an existing test module that tested `SpanGroups` specifically (`tests/doc/test_span_group.py` focuses on `SpanGroup`), and it seemed like my tests would fit best within the `tests/serialize/` dir.
3. I maintained the existing definition of `SpanGroups._EMPTY_BYTES`--as a `msgpack_dumps([])`--even though I changed `SpanGroups` to be serialized as a `dict` (not a `list`). I don't think this harms anything, and for `SpanGroups` that are already serialized using the old format (as a list), this makes deserialization just a bit easier.
4. (I left a remaining "TODO" in the 4th commit:) I wonder if we should bother to check if a serialized `SpanGroups` is neither a `dict` nor a `list`, and if that's true (i.e. if it's an invalid serialization), then perhaps do something like raise an exception, and if so, what exception that should be. For now, I don't do any such checks/exception-raising.
5. Part of me would like to see more (preferably "real-world") deserialization tests for backwards-compatibility--if only for my own peace of mind. This pull request fixes a data loss issue, and I want to make sure that deserializing `SpanGroups` (e.g. via `Doc`s/`DocBin`s) with the new approach doesn't cause any new data loss. Let me know if you have any additional tests in mind (e.g. to be tested by the new `test_deserialize_spangroups_compat` function), or if things are looking fine as they are.

> As a side note, I have a general question about spaCy's test fixtures:
>     Can we expect them to stay defined as they are--forever?
> The main reason I ask is: at least in some cases, I'm inclined to define my own e.g. `Doc` object because I'm not sure if an existing Doc fixture might change in a way that would (ultimately) break (or not) a test. (For example, a Doc fixture might not define a `.spans` for the Doc--currently--but can I assume it will always stay that way? That is, once a fixture is created, is it set in stone?)

I'm open for discussion.

### Type of change
Bug fix

## Checklist


adrianeboyd:
Thanks for the PR! I think we need to decide whether it makes sense to support `SpanGroup` names that don't match the `SpanGroups` key. Given how the serialization is designed, I don't think it was intended to support mismatched names, but I see how it isn't enforced.

single-fingal:
Sure thing; totally agreed. I could probably take a stab at the other approach, but I'll wait for a decision on the matter. I just figured it was simple enough to implement this fix (if only temporarily), while any other solution would require a more fundamental change to how `SpanGroups`/`SpanGroup`s work.

adrianeboyd:
I don't see any reasonable way to enforce that the `SpanGroups` keys and `SpanGroup.name` match in the general case, so I think this looks like a good backwards-compatible solution.

The only issue I still have is what happens in a case like this:

```python
sg = SpanGroup(doc, spans=[doc[0:2], doc[1:3]])
doc.spans["a"] = sg
doc.spans["b"] = sg
```

Where before serialization you have a single object and afterwards you have two copies of the same object.

Aside from that, I'd modify a few minor things and add a few additional tests, which I'm happy to do myself on this branch if that's okay with you?

single-fingal:
Good point--about the single object -> 2+ copies. Thankfully (more or less), users shouldn't currently code for that kind of behavior, because of the way SpanGroups currently work. As it stands in production now, `SpanGroup`s that exist as values to 2+ keys in a `SpanGroups` would only be serialized/deserialized as the value to 1 key in the resulting `SpanGroups` object. So there's currently no way for the same SpanGroup to be deserialized as the value to more than 1 key, anyway.
Given that, I think a valid way going forward would be to just inform users that, (if they're going to have a SpanGroup associated with multiple keys,) when a `SpanGroups` is deserialized, a different SpanGroup object will always be created for the value of each key.
(And if, in the future, you require `SpanGroups` keys to match with `SpanGroup.name`s, then the business about not having 2+ copies for a single SpanGroup in a `SpanGroups` will be moot (along with the code that enforced it)--since there will be no way for a SpanGroup to be the value of multiple keys.)

What you have in mind is nicely symmetric, though, and I definitely see the value in it. That's how I would do things, too, personally. It's just that from my experience, a good chunk of the serialization/deserialization I've seen in software doesn't exhibit that kind of behavior, and I suspect many users wouldn't expect it. (Counterpoint: e.g. Python's `pickle` has gotten this right, at least for the data that I've thrown at it.)

For example, I took a look, and there are portions of spaCy code that similarly deserialize single objects as multiple (different) objects:

* `Doc`s:
    This occurs for `user_data` of Docs:
    ```python
    >>> import spacy
    >>> from spacy.tokens import Doc
    >>> nlp = spacy.blank('en')

    >>> my_dict = {'key1': 'value 1'}

    >>> user_data = {'ud1': my_dict, 'ud2': my_dict}

    >>> doc = Doc(nlp.vocab, words='I have user data'.split(),
                  user_data=user_data)

    >>> doc.user_data
    {'ud1': {'key1': 'value 1'}, 'ud2': {'key1': 'value 1'}}
    
    >>> doc.user_data['ud1'] is doc.user_data['ud2']
    True

    >>> # Now serialize/deserialize
    >>> # (this behavior would also occur from Doc.to_dict()/.from_dict()
	>>> #  and .to_disk()/.from_disk()):

    >>> redoc = Doc(nlp.vocab)
    >>> redoc.from_bytes(doc.to_bytes())

    >>> redoc.user_data
    {'ud1': {'key1': 'value 1'}, 'ud2': {'key1': 'value 1'}}
    
    >>> redoc.user_data['ud1'] is redoc.user_data['ud2']
    False
    ```
* `DocBin`s:
    This occurs for Doc objects themselves, within DocBins:
    ```python
    >>> from spacy.tokens import DocBin
    
    >>> docs = [nlp('This is a single Doc object.')] * 2
    
    >>> docs[0] is docs[1]
    True

    >>> db = DocBin(docs=docs)
    >>> redocs = list(db.get_docs(nlp.vocab))
    
    >>> redocs[0] is redocs[1]
    False
    ```
* and `Lookups`/`Table`s:
    This occurs for Table objects within `Lookups` objects:
    ```python
    >>> from spacy.lookups import Lookups
    >>> lookups = Lookups()

    >>> table_a = lookups.add_table('a', {'foo': 'bar'})
    >>> lookups.set_table('b', table_a)
    >>> lookups.tables
    ['a', 'b']

    >>> table_a, table_b = lookups.get_table('a'), lookups.get_table('b')
    >>> table_a is table_b
    True

    >>> # Now serialize/deserialize:

    >>> lookups.from_bytes( lookups.to_bytes() )

    >>> lookups.tables
    ['a', 'b']

    >>> table_a, table_b = lookups.get_table('a'), lookups.get_table('b')
    >>> table_a is table_b
    False
    ```

If any of these you think are issues that should be changed, let me know and I'll create GitHub issues for them.

---

But, of course, just because something is a certain way, doesn't mean it has to stay that way.

I've created a commit that implements the kind of "symmetric" deserializing being discussed (though I currently have it on a separate branch, just because I want to let you decide if encoding that behavior is the way to go (after I had given my spiel above)): [fix/spangroups_serialization-ids](https://github.com/single-fingal/spaCy/tree/fix/spangroups_serialization-ids)

#### Implementation discussion:

I saw 2 main ways to do this:
1. include `SpanGroup` `id()`s within each serialized SpanGroup;
2. or: have `SpanGroup` containers (i.e. `SpanGroups`) do the work of identifying identical objects
    (e.g. by including SpanGroup id()s *alongside* each serialized SpanGroup)

I lean towards the latter because:
1. adding the id() to each serialized SpanGroup is overhead that's not always necessary.
2. if you have the container do the work, you can also (optionally) avoid serializing identical objects more than once in the first place.
3. however, you would have to do this for each container of such an object (where required). But, from what I can see, I suspect that `SpanGroup`s will likely have their one and only spaCy-approved container be `SpanGroups`, at least for a while.
4. we're already changing the de/serialization of `SpanGroups`, so keeping the changes in 1 place is a nice bonus.

However, with any approach, some extra processing will have to be done.

There were roughly 2 ways I saw to implement the latter method:
1. Serialize `SpanGroups` as dicts, mapping each str key to a pair: `(id(spangroup), spangroup.to_bytes())`, for every SpanGroup in the `SpanGroups`.
    The serialization would look like this:
    ```python
    ...
    msg = {key: (id(value), value.to_bytes()) for key, value in self.items()}
    return srsly.msgpack_dumps(msg)
    ```
    Deserialization would look like this:
    ```python
    ...
    msg = srsly.msgpack_loads(bytes_data)
    if isinstance(msg, dict):
        spangroup_ids: Dict[int, SpanGroup] = {}
        for key, (sg_id, value_bytes) in msg.items():
            group = spangroup_ids.get(sg_id)
            if group is None:
                group = SpanGroup(doc).from_bytes(value_bytes)
                spangroup_ids[sg_id] = group
            self[key] = group
    ...
    ```
    Advantages:
    * Very straightforward to serialize.

    Disadvantages:
    * (Only slightly) more complex to deserialize.
    * A SpanGroup will be serialized once for each key that it is a value of (though, when the `SpanGroups` is deserialized, such an object will only appear once in memory--meeting the requirement).

2. Determine the unique SpanGroup values and then serialize them--once, each--e.g. as a dict mapping SpanGroup bytes to their keys:
    Serialization:
    ```python
    ...
    id_bytes_keys = {}
    for key, value in self.items():
        value_id = id(value)
        if value_id not in id_bytes_keys:
            id_bytes_keys[value_id] = [value.to_bytes(), key]
        else:
            id_bytes_keys[value_id].append(key)
    msg = {value_bytes: keys for value_bytes, *keys in id_bytes_keys.values() }
    return srsly.msgpack_dumps(msg)
    ```
    Deserialization:
    ```python
    ...
    msg = srsly.msgpack_loads(bytes_data)
    if isinstance(msg, dict):
        for value_bytes, keys in msg.items():
            group = SpanGroup(doc).from_bytes(value_bytes)
            for key in keys:
                self[key] = group
    ...
    ```
    Advantages:
    * `SpanGroup`s that are values of more than 1 key in a `SpanGroups` aren't even serialized more than once. (Though I suspect that in practice, there will rarely be such `SpanGroup`s, anyway.)
    * Fairly straightforward to deserialize.

    Disadvantages:
    * Slightly more complex to serialize.
    * (With this particular implementation:) the order of the dict keys is not guaranteed to be the same upon deserialization. But this could be corrected for, if required.

The second approach is the one currently implemented on that branch. Though I think the first approach is just as valid, with the assumption that users aren't likely to have a `SpanGroup` be the value of more than 1 key.

In both cases, I take advantage of the fact that a `dict` is being used as the outermost container for the `SpanGroups` serialization: since it is not a `list` (which was previously used), we easily know what "version" of `SpanGroups` serialization a given bytes is. But if `SpanGroups` serialization changes again in the future, more involved processing may be required (as is the case for versioning serializations in general).

---

Bottom line: I think it makes sense to (basically) deserialize identical objects once, in general, and if you still want to go forward with that, I'll happily include the implementation I discussed--or one like it--on this branch. Just say the word.

And as far as changing the branch goes, be my guest.

I did just notice that the spaCy Developer Code Conventions doc mentioned issuing warnings in loops should be avoided, but it's getting late here. I can adjust the warnings generated in this PR, if you don't get to it first.

adrianeboyd:
I agree that it doesn't really seem necessary to worry too much about deserializing with two copies. I think we can just describe `SpanGroup.name` as being more like `SpanGroup.attrs` as an optional thing that isn't connected to `SpanGroups`.

I think it's okay to have relatively noisy warnings in cases where you're unexpectedly losing data.

adrianeboyd:
Thinking about the details some more, this has the same issue as #9708, since older versions of spacy won't be able to read newer `.spacy` files. It's going to be slightly tricky to avoid a lot of copying, but I think it would be better to continue using a list but correct the name values within the serialized version so it's restored as intended. I will have a look...

single-fingal:
I like that approach, especially since you're essentially able to improve old spaCy's behavior for the problematic cases (when given new spaCy-serialized data), but the one (relatively minor) issue I see with it is that `SpanGroup.name`s are lost after serialization if they differ from the corresponding keys; matching `.name`s to keys is enforced only for serialized data, and not for `SpanGroups` while they're in memory. e.g.:
```python
>>> doc.spans['hi'] = SpanGroup(doc, name='there')
>>> doc.spans['hi'].name
'there'
>>> doc.spans.from_bytes(doc.spans.to_bytes())
>>> doc.spans['hi'].name
'hi'
```
And this would occur for the newer versions of spaCy as well as the older versions. (I wouldn't consider it an issue if only the older versions had that behavior, since they have a bigger problem.)

The only other alternative I can think of (apart from an attempt at ensuring that SpanGroup names always match their keys), for forward-compatibility, would be something like the solution I had on the aforementioned [fix/spangroups_serialization-ids](https://github.com/single-fingal/spaCy/tree/fix/spangroups_serialization-ids) branch (i.e. the last bit of code I included in my prior comment):

Serialization could still be as a dict, e.g.:
```python
...
bytes_keys = {}
for key, value in self.items():
    # Cleaned up
    # (or use a defaultdict(list) for bytes_keys--same effect:)
    bytes_keys.setdefault(value.to_bytes(), []).append(key)
return srsly.msgpack_dumps(bytes_keys)
```
because the deserialization would be:
```python
...
msg = srsly.msgpack_loads(bytes_data)
if isinstance(msg, dict):
    for value_bytes, keys in msg.items():
        group = SpanGroup(doc).from_bytes(value_bytes)
        # (Modified from my previous GitHub post, so that
        #  each deserialized SpanGroup is always a distinct object:)
        self[keys[0]] = group
        for key in keys[1:]:
            self[key] = group.copy()
...
```
which would still be compatible with how older versions of spaCy deserialize, i.e.:
```python
msg = [] if bytes_data == self._EMPTY_BYTES else srsly.msgpack_loads(bytes_data)
self.clear()
doc = self._ensure_doc()
for value_bytes in msg:
    group = SpanGroup(doc).from_bytes(value_bytes)
    self[group.name] = group
return self
```
since iterating over the dict (keys) is equivalent to iterating over the list elements--because they're both serialized `SpanGroup`s objects. (We would just have to be comfortable with the fact that, for earlier versions of spaCy 3.x that deserialize newer data, the type of `msg` may not be a list, but another iterable.)

The two minor advantages this approach has is: it maintains `SpanGroup.name`s pre- and post-serialization, and it happens to serialize `SpanGroup`s that have the same `bytes` representation only once. But a disadvantage is: I think it doesn't result in as-nice output for older spaCy versions that your version does--for the problematic inputs.

<details><summary>An example might better show the difference between the two:</summary>

I'll call your current approach Approach A and the one above, Approach F (because Adriane gets an A and Fingal gets an F). The setup is this:
```python
>>> doc = nlp('How now, brown cow?')
>>> # Some variety:
>>> #   1 SpanGroup where its name matches its key
>>> #   2 SpanGroups that have the same name--which is not a key
>>> #   2 SpanGroups that have the same name--which is a key
>>> #   1 SpanGroup that is a value for 2 different keys (where its name is a key)
>>> doc.spans['key1'] = SpanGroup(doc, name='key1', spans=[doc[0:1]])
>>> doc.spans['key2'] = SpanGroup(doc, name='too', spans=[doc[1:2]])
>>> doc.spans['key3'] = SpanGroup(doc, name='too', spans=[doc[2:3]])
>>> doc.spans['key4'] = SpanGroup(doc, name='key4', spans=[doc[3:4]])
>>> doc.spans['key5'] = SpanGroup(doc, name='key4', spans=[doc[4:5]])
>>> sg6 = SpanGroup(doc, name='key6', spans=[doc[5:6]])
>>> doc.spans['key6'] = sg6
>>> doc.spans['key7'] = sg6

>>> # This is what the user entered and so likely expects:
>>> doc.spans
{'key1': [How], 'key2': [now], 'key3': [,], 'key4': [brown], 'key5': [cow], 'key6': [?], 'key7': [?]}
>>> [ group.name for group in doc.spans.values() ]
['key1', 'too', 'too', 'key4', 'key4', 'key6', 'key6']
```

Serializing with Approach A and deserializing with A results in:

```python
>>> redoc = nlp('How now, brown cow?')
>>> redoc.spans.from_bytes(doc.spans.to_bytes()) # this is fine
{'key1': [How], 'key2': [now], 'key3': [,], 'key4': [brown], 'key5': [cow], 'key6': [?], 'key7': [?]}
>>> [group.name for group in redoc.spans.values()]
['key1', 'key2', 'key3', 'key4', 'key5', 'key6', 'key7']
```
The only noticeable thing is that the names post-serialization are different, as noted above.

Serializing with A and deserializing with current/"old" spaCy results in:

```python
>>> # `spangroups_bytes` is set to the value of `doc.spans.to_bytes()`
>>> redoc.spans.from_bytes( spangroups_bytes ) # this is fine--or better than fine
{'key1': [How], 'key2': [now], 'key3': [,], 'key4': [brown], 'key5': [cow], 'key6': [?], 'key7': [?]}
>>> [ group.name for group in redoc.spans.values() ]
['key1', 'key2', 'key3', 'key4', 'key5', 'key6', 'key7']
```
This is what I like in particular about A: old spaCy deserializes this new spaCy serialization the same way that A/new spaCy deserializes it.

---

Now, serializing with Approach F and deserializing with F results in:

```python
>>> redoc = nlp('How now, brown cow?')
>>> redoc.spans.from_bytes(doc.spans.to_bytes()) # this is fine
{'key1': [How], 'key2': [now], 'key3': [,], 'key4': [brown], 'key5': [cow], 'key6': [?], 'key7': [?]}
>>> [group.name for group in redoc.spans.values()]
['key1', 'too', 'too', 'key4', 'key4', 'key6', 'key6']
```
The only notable difference with A is that the `.name`s match their values pre-serialization.

Serializing with F and deserializing with old spaCy results in:

```python
>>> redoc.spans.from_bytes( spangroups_bytes )
{'key1': [How], 'too': [,], 'key4': [cow], 'key6': [?]}
>>> [ group.name for group in redoc.spans.values() ]
['key1', 'too', 'key4', 'key6']
```
This is where F isn't as nice as A. There are missing keys/values.

But it's basically the same as how old spaCy would deserialize the same example `doc.spans` if it was also serialized by old spaCy in the first place (not A or F):

```python
>>> redoc.spans.from_bytes( doc.spans.to_bytes() )
{'key1': [How], 'too': [,], 'key4': [cow], 'key6': [?]}
>>> [ group.name for group in redoc.spans.values() ]
['key1', 'too', 'key4', 'key6']
```
(Older spaCy versions will only deserialize 1 of the `SpanGroup`s for each unique `SpanGroup.name`--which is the bug we're addressing here (at least for new versions of spacy).)
So basically, with F, older spaCy versions will stay as broken as they already are for such data.

(That difference between the approaches could basically be explained by what takes priority when determining a SpanGroup's key for old spaCy: `SpanGroup.name`s or `SpanGroups` keys. In old spaCy and Approach F, the `.name`s determine the keys. Approach A has the keys take priority.)

</details>

I've committed an update to that other [fix/spangroups_serialization-ids](https://github.com/single-fingal/spaCy/tree/fix/spangroups_serialization-ids) branch, which reflects this other approach (though I still need to include your earlier refactoring). I'm a bit on the fence, though, because I do like how your approach makes older spaCy work better for the problematic cases, but I would prefer it if it maintained `SpanGroup.name`s how they were pre-serialization (for new spaCy).

Any preferences?

adrianeboyd:
I mainly switched from the last to first key because I thought it made the error message clearer for the user, but I see how it changes the behavior.

I like the idea of using the bytes as the dict keys to make it backwards-compatible.

I think this now addresses all your concerns?

single-fingal:
I just made one main change: to always deserialize `SpanGroup`s as copies (for each
key), since it's possible that two different `SpanGroup`s have the same bytes, so we can't assume two `SpanGroup`s with the same bytes were the same (identical) object.

This doesn't result in "symmetric" serialization (to use the term from earlier), but it turns out it's impossible to have both symmetric serialization and forward-compatibility (for prior versions of spaCy) at the same time, here. But from our earlier discussion (and the fact that our "A" and "F" approaches (referenced in the prior comment) behave this way), it seems like it's acceptable to not be symmetric.

The only other thing I would like to do is add a few more tests, unless you're comfortable with where it's at. I can likely come up with some more tomorrow.

spacy/tokens/_dict_proxies.py
```diff
@@ -53,20 +54,51 @@ def setdefault(self, key, default=None):
...
-        msg = [value.to_bytes() for value in self.values()]
+        msg: Dict[bytes, List[str]] = {}
+        for key, value in self.items():
+            msg.setdefault(value.to_bytes(), []).append(key)
         return srsly.msgpack_dumps(msg)
 
     def from_bytes(self, bytes_data: bytes) -> "SpanGroups":
-        msg = [] if bytes_data == self._EMPTY_BYTES else srsly.msgpack_loads(bytes_data)
+        # backwards-compatibility: bytes_data may be one of:
+        # b'', a serialized empty list, or a serialized list of SpanGroup bytes
```
adrianeboyd, line 69:
```suggestion
        # b'', a serialized empty list, a serialized list of SpanGroup bytes
        # or a serialized dict of SpanGroup bytes -> keys
```

adrianeboyd:
The edits look good, thanks! I would approve this with the current tests, but if you'd like to add more tests I'm not going to object.

single-fingal:
Done.

And hopefully the long bytestring that I added isn't too bothersome... Eventually, I imagine there might be test data that's stored in separate files, but I figured starting that trend for just one (still relatively small) example might not be worth it.

adrianeboyd (approved):

spacy/tests/serialize/test_serialize_span_groups.py
```diff
@@ -0,0 +1,160 @@
...
+            {
+                "key1": {"name": "key1", "spans": [(0, 1), (1, 2)]},
+                "too": {"name": "too", "spans": [(1, 2), (0, 1)]},
+                "key4": {"name": "key4", "spans": [(0, 1)]},
+            },
+            True,
+        ),
+    ],
+)
+def test_deserialize_span_groups_compat(
```
svlandeg, line 115: I really love the fact that you've included this bwd-compat test in here with the bytestrings!

spacy/tests/serialize/test_serialize_span_groups.py
```diff
@@ -0,0 +1,160 @@
...
+    en_tokenizer, spans_bytes, doc_text, expected_spangroups, expected_warning
+):
+    """Test backwards-compatibility of `SpanGroups` deserialization.
+    This uses serializations (bytes) from a prior version of spaCy.
```
svlandeg, line 119:
```suggestion
    This uses serializations (bytes) from a prior version of spaCy (before 3.3.1).
```

spacy/tests/serialize/test_serialize_span_groups.py
```diff
@@ -0,0 +1,160 @@
...
+    expected_warning (bool): Whether a warning is to be expected from .from_bytes()
+        --i.e. if more than 1 SpanGroup has the same .name within the `SpanGroups`.
+    """
+    doc = en_tokenizer(doc_text)
+
+    if expected_warning:
+        with pytest.warns(UserWarning):
+            doc.spans.from_bytes(spans_bytes)
+    else:
+        doc.spans.from_bytes(spans_bytes)
```
svlandeg, line 137: Correct me if I'm wrong, but this doesn't catch the behaviour if `expected_warning` is `False` but the deserialization does in fact throw a warning, right?

(I wouldn't hold off on merging because of this though, it's just a very minor thing)

svlandeg (approved):
Thanks so much @single-fingal for raising this issue and for the clear examples where the current behaviour is unsatisfying, and for creating this PR with such care!

I really appreciate all the thought that has gone into ensuring compatibility with older spacy versions and the many tests.

This looks good to merge by me. I had just one or two very minor nitpicks but if @adrianeboyd wants to get this out in 3.3.1 we should go ahead and get this merged :-)

spacy/tests/serialize/test_serialize_span_groups.py
```diff
@@ -0,0 +1,160 @@
...
```
adrianeboyd, line 137:
```suggestion
        # TODO: explicitly check for lack of a warning
        doc.spans.from_bytes(spans_bytes)
```

spacy/tests/serialize/test_serialize_span_groups.py
single-fingal: Oof. Good catch. I thought I had tested for that when I first wrote that test function... (Maybe my pytest config was turning those warnings into errors? Doubtful; it was probably just PEBKAC. Either way, it shouldn't happen again.)

You're right, and I've written a couple of lines to ensure that warnings aren't ~raised when we don't expect them.

single-fingal: (I pushed a commit with those changes to this same branch, but let me know if they should be elsewhere.)

adrianeboyd: We have several places in the test suite that check for no warnings that should probably all be refactored together with a correspond change in the pytest requirements. I've added a link to this discussion in the comments for our internal task so this can be done in the combination with the existing cases. Thanks for your attention to this!

PR #10718, opened 2022-04-27
Remove vestigial debug print statement in `walk_head_nodes`

shadeMe:
## Description

Removes one of the last, potential GIL-Refnanny related regressions.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/tokens/graph.pyx
+++ b/spacy/tokens/graph.pyx
@@ -19,7 +19,7 @@ from .token import Token
 cdef class Edge:
     cdef readonly Graph graph
     cdef readonly int i
-
+    
     def __init__(self, Graph graph, int i):
         self.graph = graph
         self.i = i
@@ -35,7 +35,7 @@ cdef class Edge:
     @property
     def head(self) -> "Node":
         return Node(self.graph, self.graph.c.edges[self.i].head)
-
+    
     @property
     def tail(self) -> "Tail":
         return Node(self.graph, self.graph.c.edges[self.i].tail)
@@ -61,7 +61,7 @@ cdef class Node:
     def __init__(self, Graph graph, int i):
         """A reference to a node of an annotation graph. Each node is made up of
         an ordered set of zero or more token indices.
-
+        
         Node references are usually created by the Graph object itself, or from
         the Node or Edge objects. You usually won't need to instantiate this
         class yourself.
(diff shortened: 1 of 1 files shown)


adrianeboyd:
I realize that it's a really small change, but all the reformatting makes this PR kind of hard to follow and is going to make the history really confusing in the future. (I had to use a text search to find the place where the print statement was removed in the commit.)

We don't have automatic processes for formatting cython, so if there's a good solution that you're aware of, it would be useful to add it. If it's not, then I would prefer if this kind of PR didn't do any unrelated reformatting, or at least not beyond a few lines at most. What does everyone think?

shadeMe:
Ah, I didn't notice all the whitespace changes on my end. I can revert them, not an issue. That said, the GitHub diff interface does let one hide whitespace changes: Click on the gear icon at the top of the _Files Changed_ page, check "Hide whitespace" and hit _Apply_.

I don't have a specific formatter for Cython, but my editor is configured to automatically trim trailing whitespaces (which is what has happened here).

svlandeg:
> I realize that it's a really small change, but all the reformatting makes this PR kind of hard to follow and is going to make the history really confusing in the future. (I had to use a text search to find the place where the print statement was removed in the commit.)
> 
> We don't have automatic processes for formatting cython, so if there's a good solution that you're aware of, it would be useful to add it. If it's not, then I would prefer if this kind of PR didn't do any unrelated reformatting, or at least not beyond a few lines at most. What does everyone think?

Agreed! As long as we don't have a standard way of formatting Cython, I'd avoid making too many formatting changes in parts of the code that would otherwise be untouched by a PR. Mostly for `git blame`'s sake ;-)

svlandeg (commented):
The `graph.pyx` stuff is not currently used by anyone or anything, as far as I know. There's actually plenty of other `print` statements that should be cleaned up in this same file, or at least put behind some `debug` variable or something.

shadeMe:
I only see two other `print` statements in the `graph.pyx` file on my end. Both of them are in regular Python functions, so they shouldn't concern us at least with respect to RefNanny. Still, I could remove them in this PR if that's what we want to do (feature-gating them seems like overkill in this case given how ad-hoc they appear to be).

In any event, I do think it makes sense to merge the proposed change as it will eliminate a potential bottleneck in the future, especially given that the GIL lock was being acquired in a graph traversal method.

svlandeg:
Yea, let's just remove the other `print` statements too (you're right that there are currently 3 in total, I seemed to recall there were more)

Issue #10643, opened 2022-04-11
Removing overlapping rule from PhraseMatcher leads to segfault


pypae:
## Minimal Reproducible Example
Removing rules from a `Phrase Matcher` that overlap with terms of other rules leads to a segfault.
For my usecase, I cannot add and remove the terms in the same rule, because I need to keep track of the individual `match_id`s. 

```python
import spacy
from spacy.matcher import PhraseMatcher

segment = "Keep in mind that this will only save out the binary data for the individual components."
terms = ["binary", "binary data"]

nlp = spacy.load("en_core_web_sm")
matcher = PhraseMatcher(nlp.vocab)
for match_id, term in enumerate(nlp.pipe(terms)):
    print(f"Adding {match_id=}, {term=}")
    matcher.add(str(match_id), [term])

doc = nlp(segment)
matches = matcher(doc)

for match_id in range(len(terms)):
    print(f"Removing {match_id=}")
    matcher.remove(str(match_id))
```

## Environment
- **spaCy version:** 3.2.4
- **Platform:** macOS-12.3-arm64-arm-64bit, Linux-5.4.0-97-generic-x86_64-with-glibc2.29
- **Python version:** 3.9.12, 3.8.10
- **Pipelines:** de_core_news_sm (3.2.0), en_core_web_sm (3.2.0)

PR #10734, opened 2022-04-30
Fix PhraseMatcher remove overlapping terms

pypae:
## Fix PhraseMatcher remove overlapping terms
When removing keywords from a `PhraseMatcher`, the node after the last token was not considered a `path_node`. 
On line 140, we check if the current node contains other successors, and remove it if it doesn't.
```python
if node_pointer.filled == 1:
    # remove node
else:
    # keep node
```
If we don't include the node after the last token, we never check, if there are tokens of other keywords following this token.  As a result, removing the shorter of two overlapping keywords resulted in the removal of both terms.

The illustration below shows the issue using the terms of the newly added test case.

```mermaid
graph LR
    id1(self.c_map)-- binary --> id2( )
    id2 -- "self._terminal_hash" --> id3( )
    id2 -- "data" --> id4( )
    id4 -- "self._terminal_hash" --> id5( )

    subgraph "path_nodes (fixed)"
    id2
        subgraph "path_nodes (master)"
        id1
        end
    end
``` 

### Types of change
Fix issue #10643.

## Checklist

### DIFF
--- a/spacy/matcher/phrasematcher.pyx
+++ b/spacy/matcher/phrasematcher.pyx
@@ -118,6 +118,8 @@ cdef class PhraseMatcher:
                     # if token is not found, break out of the loop
                     current_node = NULL
                     break
+            path_nodes.push_back(current_node)
+            path_keys.push_back(self._terminal_hash)
             # remove the tokens from trie node if there are no other
             # keywords with them
             result = map_get(current_node, self._terminal_hash)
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Thanks for the PR! This looks like exactly the kind of error I was expecting to find given the bug report, but I hadn't had time to look into the details yet. (And that's a cool-looking markdown graph!)

adrianeboyd:
This looks good to me, thanks again for the PR!

PR #10741, opened 2022-05-02
Fix StringStore.__getitem__ return type depending on parameter types

ldorigo:
Small fix using  `@overload` so that `StringStore.__getitem__` returns an `int` when given a `str` or `bytes` and a `str` when given an `int`.

### Types of change

Small typing improvement

## Checklist


adrianeboyd:
This change makes sense, but it looks like mypy doesn't support it, at least not in the version we're currently using (until `thinc` updates are complete to support newer mypy and pydantic version). Does the mypy check look different on your end?

ldorigo:
Sorry, I haven't tested it in mypy (only in pyright), and I didn't think that types were part of the tests (I just submitted the PR as a small patch via Github's online editor). I did check and it seems like @overload is used in other parts of Spacy though - will get a proper dev environment locally now and see what's up :-)

spacy/strings.pyi
```diff
@@ -7,7 +7,11 @@ class StringStore:
     def __init__(
         self, strings: Optional[Iterable[str]] = ..., freeze: bool = ...
     ) -> None: ...
-    def __getitem__(self, string_or_id: Union[bytes, str, int]) -> Union[str, int]: ...
+    @overload
+    def __getitem__(self, string_or_id: Union[bytes, str]) -> int: ...
+    @overload
+    def __getitem__(self, string_or_id: int) -> str: ...
+    def __getitem__(self, string_or_id: Union[bytes, str, int]) -> Union[str, int]: ...
```
adrianeboyd, line 14: Ah, I think the problem is this line, which shouldn't be needed anymore.
```suggestion
```

ldorigo:
Sorry, had some trouble compiling locally - thanks for fixing it yourself!

PR #10746, opened 2022-05-03
Parser `StateC` optimizations

shadeMe:
## Description

This PR concerns itself with the following:
* It removes the `gil`-based error handling in the `__init_` method to prevent RefNanny from acquiring a lock during function invocation. This was done by replacing the C-array used to track heads with a `vector`.
* Reserves memory for the internal buffer at init-time to reduce allocation overhead at insertion-time.
* Replaces `std` container `at` calls (which can throw) with `noexcept` subscript operator invocations. This should reduce the amount of EH-specific code emitted by the C++ compiler. We were doing the requisite bounds checking ourselves anyway.

### Types of change

Performance enhancement

## Checklist

### DIFF
--- a/spacy/pipeline/_parser_internals/_state.pxd
+++ b/spacy/pipeline/_parser_internals/_state.pxd
@@ -255,7 +255,7 @@ cdef cppclass StateC:
             return 0
         elif this._sent[word].sent_start == 1:
             return 1
-        elif this._sent_starts.count(word) >= 1:
+        elif this._sent_starts.const_find(word) != this._sent_starts.const_end():
             return 1
         else:
             return 0
--- a/spacy/pipeline/_parser_internals/_state.pxd
+++ b/spacy/pipeline/_parser_internals/_state.pxd
@@ -47,7 +47,7 @@ cdef cppclass StateC:
         this._heads.resize(length)
         this._unshiftable.resize(length)
 
-        # Reserve memoray ahead of time to minimize allocations during parsing.
+        # Reserve memory ahead of time to minimize allocations during parsing.
         # The initial capacity set here ideally reflects the expected average-case/majority usage.
         init_capacity = 20
         this._stack.reserve(init_capacity)
--- a/spacy/pipeline/_parser_internals/_state.pxd
+++ b/spacy/pipeline/_parser_internals/_state.pxd
@@ -26,7 +26,7 @@ cdef struct ArcC:
 
 
 cdef cppclass StateC:
-    int* _heads
+    vector[int] _heads
     const TokenC* _sent
     vector[int] _stack
     vector[int] _rebuffer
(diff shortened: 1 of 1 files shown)


spacy/pipeline/_parser_internals/_state.pxd
```diff
@@ -41,25 +40,32 @@ cdef cppclass StateC:
...
-    __init__(const TokenC* sent, int length) nogil:
+    __init__(const TokenC* sent, int length) nogil except +:
+        cdef int init_capacity
+
+        this._heads.resize(length)
+        this._unshiftable.resize(length)
+
+        # Reserve memory ahead of time to minimize allocations during parsing.
+        # The initial capacity set here ideally reflects the expected average-case/majority usage.
+        init_capacity = 20
```
danieldk, line 51: Maybe this is too much of a micro-optimization, but since most non-Windows impls of `vector` use doubling, using a power of two will keep the capacities of power of two. This might be beneficial, since a lot of memory management use sizes that are powers of two (pages, different pools, etc).

spacy/pipeline/_parser_internals/_state.pxd
```diff
@@ -41,25 +40,32 @@ cdef cppclass StateC:
...
-            with gil:
-                PyErr_SetFromErrno(MemoryError)
-                PyErr_CheckSignals()
         this.offset = 0
         this.length = length
         this._b_i = 0
         for i in range(length):
             this._heads[i] = -1
-            this._unshiftable.push_back(0)
+            this._unshiftable[i] = 0
```
danieldk, line 65: Doesn't `this._unshiftable.resize(length)` already set this to all-zeros? (Since the default constructor is used.)

Also, the resize for `_heads` could be changed to:

```
this._heads.resize(length, -1)
```

and then this loop could be eliminated completely.

danieldk (commented):
Looks like a good set of optimizations. Added some comments.

spacy/pipeline/_parser_internals/_state.pxd
shadeMe: Indeed. Since the default size is zero, the `resize` call should default-insert all new values.

spacy/pipeline/_parser_internals/_state.pxd
```diff
@@ -132,19 +135,24 @@ cdef cppclass StateC:
...
+        if i >= stack_size or i < 0:
             return -1
-        return this._stack.at(this._stack.size() - (i+1))
+        else:
+            return this._stack[stack_size - (i+1)]
 
     int B(int i) nogil const:
+        cdef int buf_size
+
+        buf_size = this._rebuffer.size()
```
danieldk, line 149:
```suggestion
        cdef int buf_size = this._rebuffer.size()
```

spacy/pipeline/_parser_internals/_state.pxd
```diff
@@ -41,25 +40,29 @@ cdef cppclass StateC:
...
-    __init__(const TokenC* sent, int length) nogil:
+    __init__(const TokenC* sent, int length) nogil except +:
+        cdef int init_capacity
+
+        this._heads.resize(length, -1)
+        this._unshiftable.resize(length, False)
+
+        # Reserve memory ahead of time to minimize allocations during parsing.
+        # The initial capacity set here ideally reflects the expected average-case/majority usage.
+        init_capacity = 32
```
danieldk, line 51: I'd just use

```
    cdef int init_capacity = 32
```

here. Luckily this is not C89 :).

spacy/pipeline/_parser_internals/_state.pxd
```diff
@@ -132,19 +135,24 @@ cdef cppclass StateC:
                 ids[i] = -1
 
     int S(int i) nogil const:
-        if i >= this._stack.size():
-            return -1
-        elif i < 0:
+        cdef int stack_size
+
+        stack_size = this._stack.size()
```
danieldk, line 140: Sorry, forgot this in the previous review, but I think if the assignment is also in the outer scope, it's nicer/shorter to assign at the declaration site.

```suggestion
        cdef int stack_size = this._stack.size()
```

spacy/pipeline/_parser_internals/_state.pxd
shadeMe: Ah, I see. I wasn't sure that it was allowed, so I just followed the style I saw in the codebase. Will change these assignments.

danieldk (approved):
Looks good to me!

PR #10751, opened 2022-05-03
Add test for old architectures

svlandeg:
## Description
Add (slow) tests for old architectures so we can spot it easily when something goes wrong with stuff in `spacy-legacy`. Now that we run the slow test suite regularly, I think it's fine to mark some of these tests as slow.

Also includes a doc fix as 2 textcat architectures were under the subheading "layers" instead of "architectures". The diff looks weird but basically I just moved some subsections around.

### Types of change
adding (slow) tests

## Checklist


adrianeboyd:

PR #10762, opened 2022-05-06
Document different ways to create a pipeline

rmitsch:
## Goals

Document/emphasize different ways to create pipelines in docs.

## Description

Moved up and slightly modified a snippet describing the usage of `spacy.blank()` and calling the `Language` object directly.

### Types of change

Changes to existing documentation.

## Checklist

### DIFF
--- a/website/docs/usage/models.md
+++ b/website/docs/usage/models.md
@@ -44,9 +44,9 @@ languages that don't yet come with a trained pipeline, you have to import them
 directly, or use [`spacy.blank`](/api/top-level#spacy.blank):
 
 ```python
-from spacy.lang.uk import Ukrainian
-nlp = Ukrainian()  # use directly
-nlp = spacy.blank("uk")  # blank instance
+from spacy.lang.yo import Yoruba
+nlp = Yoruba()  # use directly
+nlp = spacy.blank("yo")  # blank instance
 ```
 
 A blank pipeline is typically just a tokenizer. You might want to create a blank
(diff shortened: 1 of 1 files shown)


website/docs/usage/models.md
```diff
@@ -59,6 +38,31 @@ your data.
...
+If a trained pipeline is available for a language, you can download it using the
+[`spacy download`](/api/cli#download) command as shown above. In order to use
+languages that don't yet come with a trained pipeline, you have to import them
+directly, or use [`spacy.blank`](/api/top-level#spacy.blank):
+
+```python
+from spacy.lang.fi import Finnish
+nlp = Finnish()  # use directly
+nlp = spacy.blank("fi")  # blank instance
+```
```
adrianeboyd, line 50: I think this example will be less confusing if we switch to something that currently doesn't have a pipeline, so not Finnish.

rmitsch: Changed to Ukrainian.

website/docs/usage/models.md
```diff
@@ -59,6 +38,31 @@ your data.
...
+
+Initializing the language object directly yields the same result as generating
+it using `spacy.blank()`. In both cases the default configuration for the chosen
+language is loaded, and no pretrained components will be available.
```
polm, line 54: I would explain what a blank pipeline actually does and why you'd want it. Something like:

> A blank pipeline is typically just a tokenizer. You could create a blank pipeline when you only need a tokenizer, when you want to add more components from scratch, or for testing purposes.

website/docs/usage/models.md
adrianeboyd: Sorry, there's no way for you to know, but `Ukrainian` may be added soon-ish. How about `Yoruba`?

PR #10772, opened 2022-05-09
Override SpanGroups.setdefault to provide default SpanGroup

adrianeboyd:
## Description

Override `SpanGroups.setdefault` to provide `default` values from `SpanGroup`/`Iterable` as `SpanGroup`
Minor:

* fix a mistake in the `SpanGroup` docs

### Types of change

?

## Checklist

### DIFF
--- a/spacy/tokens/_dict_proxies.py
+++ b/spacy/tokens/_dict_proxies.py
@@ -45,7 +45,11 @@ class SpanGroups(UserDict):
 
     def setdefault(self, key, default=None):
         if not isinstance(default, SpanGroup):
-            raise ValueError("SpanGroups.setdefault only supports default values with type SpanGroup, received type '{default_type}'.".format(default_type=type(default)))
+            if default is None:
+                spans = []
+            else:
+                spans = default
+            default = self._make_span_group(key, spans)
         return super().setdefault(key, default=default)
 
     def to_bytes(self) -> bytes:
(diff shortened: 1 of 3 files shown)


spacy/tokens/_dict_proxies.py
```diff
@@ -43,6 +43,11 @@ def copy(self, doc: Optional["Doc"] = None) -> "SpanGroups":
             doc = self._ensure_doc()
         return SpanGroups(doc).from_bytes(self.to_bytes())
 
+    def setdefault(self, key, default=None):
+        if not isinstance(default, SpanGroup):
```
svlandeg, line 47: What if it's an `Iterable["Span"]` which we do support for `__setitem__`? We could also call the `self._make_span_group` functionality?

spacy/tokens/_dict_proxies.py
```diff
@@ -43,6 +43,11 @@ def copy(self, doc: Optional["Doc"] = None) -> "SpanGroups":
...
+            raise ValueError("SpanGroups.setdefault only supports default values with type SpanGroup, received type '{default_type}'.".format(default_type=type(default)))
```
svlandeg, line 48: Add message to `errors.py`?

Also I'd add in a "but" after the comma to make the message even more crystal clear

spacy/tokens/_dict_proxies.py
adrianeboyd: Ah, I think I was originally thinking we'd have to enforce the `name` and then there's be no way to support a list, but if we've abandoned all hope there, then we could use `_make_span_group`. (I kind of want to disable it entirely because it's less efficient that other options once you have this expensive / unused object creation here (vs. `None` or `[]`), but I guess people can make their own decisions.)

adrianeboyd:
If we do it like this then it probably doesn't need to be restricted to v3.4.

svlandeg (approved):
This approach looks fine by me!

PR #10774, opened 2022-05-09
Enable multiprocessing groups within project config

richardpaulhudson:
## Description
Enable the specification of a group of commands within a spaCy project workflow that are to be executed in parallel.

### Features

1. Each spaCy projects command is made up of `n` operating-system-level commands that are executed in series. The spaCy projects commands within a parallel group are executed in parallel, but the operating-system-level commands within each spaCy projects command are still executed in series.
2. spaCy project workflows support the definition of dependencies and outputs (files created by commands) in order to ensure that spaCy projects commands are not re-executed unnecessarily on consecutive workflow runs. All this functionality works for each command in a parallel group with respect to the rest of the project file in the same way as it does for a serial command. However, the management of dependencies *between* the members of a parallel group is out of scope: the user is responsible for ensuring that no problems occur.
3. It is possible to specify a project-file-wide maximum number of parallel processes. If a parallel group contains more commands than this maximum `n`, only the first `n` commands within the group are started. Whenever a command completes, the next command in the group is started, and so on until all the commands in the group have executed.
4. A table is displayed with real-time status information about the executing commands. The status table works in a similar fashion to the `docker pull` command-line interface.
6. If any command within a group returns a non-zero return code, the execution of the other commands within the group is terminated.
7. Console output relating to a command as well as outputted by the OS-level commands it comprises is initially redirected to a logfile in a temporary directory that a user can monitor in real time if they so wish. When all the commands in a parallel group have completed, the contents of these logfiles are reproduced in the console. This ensures that output from different commands is not mixed up in the console. It also means that the console output after a parallel group has executed is very similar to the console output that would have resulted from the group being executed in series.

### Architecture

1. The **main process** that is executing the project file uses the `multiprocessing` module to spawn two or more **worker processes**, each of which is responsible for a single spaCy project command, and each of which executes — in series, as **subprocesses**, and using the `subprocess` module — the operating-system-level commands that it contains. 
2. Each worker process sends status information about the command and operating-system-level commands that it is executing back to the main process via a [queue](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue). This status information includes the process ids of the subprocesses that the worker process starts for each operating-system level command.
3. If a subprocess has failed or been terminated, the worker process that started it does not execute any remaining operating-system-level commands within the spaCy project command for which it is responsible.
4. If the main process receives notice from a worker process that a subprocess has failed, it terminates directly any other subprocesses that are currently running. Each worker process controlling one of these subprocesses detects that its subprocess has been terminated and notifies the main process of this fact. Doing things this way round has the advantage that subprocesses terminated by the main process are detected, treated and reported identically to subprocesses terminated from outside the library, i.e. directly from the operating system console.
5. The main process maintains a simple state machine that reflects its knowledge of the activities of the worker processes. It uses this knowledge to decide when to start new worker processes or terminate subprocesses, and as the basis for the status table. Note that the state machine maintains a distinction between `failed` and `terminated` processes that is only meainingful on POSIX platforms; on Windows, both states are reported as `failed/terminated`.

The background to the high-level decision to use `subprocess` in conjunction with `multiprocessing` and a status queue is documented [here](https://github.com/richardpaulhudson/spacy_projects_multiprocessing).

### Design decisions in need of review

These design decisions reflect various tradeoffs that others may well judge differently; all of them could be altered with minimal effort.

| **Nr** | **Decision** | **Rationale** |
| --- | --- | --- |
| 1. | The console output for each command, having initially been redirected to its own logfile, is reproduced on the console once execution of the parallel group is complete, and there is no option to switch this off. | Reproducing the log output has the advantage that the console output is virtually unchanged whether commands are executed in series or in parallel; an option to switch off reproducing log output would complicate the project file syntax; in most cases commands do not log significant amounts of output, so reproducing it is unproblematic. |
| 2. | The logfiles for the individual commands are written to a temporary directory; the name of each logfile is derived from the name of the command whose output it contains (this is possible because parallel groups are not allowed to contain the same command twice); this behaviour is fixed. | The output is reproduced on the console which is where most users will view it; adding options around the location of the log files would complicate the project file syntax. |
| 3. | The console output from each worker process is sent to the main process via the queue, although it would also be possible for the main process to read it directly from the file system. | Although this procedure means the output has to be serialized an extra time, the files are expected to remain small, so the improved encapsulation the procedure provides is more important; I confirmed with a short load test that the queue can handle much larger amounts of data than it would ever realistically need to transfer. |
| 4. | `STDERR` output is treated identically to `STDOUT` output and there is no option to change this. | In most cases, the most important requirement is that commands being executed in parallel do not output to the console in real time as this messes up the status table display. `STDERR` output is only likely to be important when debugging complex issues, in which case the user is in any case likely to get the problematic commands working in series before moving them into a parallel group. |
| 5. | Execution of a parallel group is halted if any command returns a non-zero return code, although there are situations in which a non-zero return code might be expected. | The same is true of spaCy projects in general when commands are executed in series. |
| 6. | Processes are terminated using the `SIGTERM` signal and there is no option to use alternatives like `SIGKILL`. | Terminating the other commands in a group when one has failed is a convenience feature. The user's next step is likely to be to debug the failed command on its own, i.e. serially outside a parallel group. It seems most unlikely that a user's response to a process within a parallel group failing and it not having been possible to terminate some other process in the group with `SIGTERM` would be to specify a different signal. |
| 7. | The keepalive message interval, maximum width of a parallel command group name within a divider and temporary logfile directory name are all hardcoded in the `parallel.py` module without any options to change them. | Options to change these parameters would complicate the syntax of the project file; it does not seem likely that anyone would have a reason to want values different from the defaults. |
| 8. | The main process spawns worker processes regardless of the platform, although forking is more efficient on Unix/Linux. | Because the main process does not have a significant memory footprint, the additional cost of spawning is not relevant; on the other hand, it makes sense to keep the behaviour as consistent as possible across platforms. |
| 9. | The execution of serial commands is left as it was, meaning that there are differences in how a command is executed depending on whether it is part of a parallel group or not. | It would be possible to execute each serial command as a 1-member parallel group, but this would unnecessary complicate what is still the standard way of executing commands, and would also increase the risk of this PR. |

### Demonstration
This demonstration has been tested on Linux, macOS and Windows 10. To try out the functionality, create these two files in a directory:

`script.py`:

```
import sys
from time import sleep
_, sleep_secs, rc = sys.argv
print("Output before sleep to stdout")
sleep(int(sleep_secs))
print("Output after sleep to stderr", file=sys.stderr)
sys.exit(int(rc))
```

`project.yml`:

```
workflows:
  all:
    - sleepC
    - parallel: [sleepC, sleepA, sleepB, sleepD, sleepE]
    - parallel: [sleepE, sleepA, fail, sleepC, sleepD]
    - sleepB

max_parallel_processes: 2
commands:

  - name: sleepA
    script:
      - "python script.py 2 0"
      - "python script.py 2 0"
      - "python script.py 3 0"
      - "python script.py 4 0"

  - name: sleepB
    script:
     - "python script.py 1 0"
     - "python script.py 2 0"

  - name: sleepC
    script:
    - "python script.py 4 0"

  - name: sleepD
    script:
    - "python script.py 2 0"
    outputs:
    - someOutput

  - name: sleepE
    script:
    - "python script.py 1 0"
    - "python script.py 1 0"
    - "python script.py 1 0"
    - "python script.py 1 0"
    - "python script.py 1 0"

  - name: fail
    script:
    - "python script.py 1 1"
```

Then type from within the directory:

1) `spacy project run all` (successful and then unsuccessful execution of a parallel group; running commands are terminated when another command fails in the second group; command with unchanged output is not run the second time; status table is displayed)
2) `spacy project run all --force`
3) `spacy project run all --dry`
4) `spacy project run`
5) `spacy project run all --help`
6) `spacy project document`
7) `spacy project dvc` (can only be attempted if `dvc` is already installed; exits with an error message as `dvc` does not support parallel groups)

### Types of change
Enhancement or new feature

## Checklist


richardpaulhudson:

spacy/cli/_util.py
```diff
@@ -221,24 +221,45 @@ def validate_project_commands(config: Dict[str, Any]) -> None:
 
     config (Dict[str, Any]): The loaded config.
     """
+
+    def verify_workflow_step(workflow_name: str, step: str) -> None:
```
svlandeg, line 225: We shouldn't let this depend on the definition of `command_names` that is below it - it makes the code less readible and more prone to errors. 

Instead, I propose to make this a private (non-nested) function that takes in `config` and `step`, or alternatively `workflow`, `commands`, and `step` (shorter variable names help with readability, too)

spacy/cli/_util.py
```diff
@@ -221,24 +221,45 @@ def validate_project_commands(config: Dict[str, Any]) -> None:
...
+            )
+
     command_names = [cmd["name"] for cmd in config.get("commands", [])]
     workflows = config.get("workflows", {})
     duplicates = set([cmd for cmd in command_names if command_names.count(cmd) > 1])
     if duplicates:
         err = f"Duplicate commands defined in {PROJECT_FILE}: {', '.join(duplicates)}"
         msg.fail(err, exits=1)
-    for workflow_name, workflow_steps in workflows.items():
+    for workflow_name, workflow_step_or_lists in workflows.items():
```
svlandeg, line 240: Let's find a shorter name for `workflow_step_or_lists`, maybe just `commands`? Then what is currently `command_names` could be `all_commands` or such.

spacy/cli/_util.py
```diff
@@ -221,24 +221,45 @@ def validate_project_commands(config: Dict[str, Any]) -> None:
...
-                    f"Unknown command specified in workflow '{workflow_name}': {step}",
-                    f"Workflows can only refer to commands defined in the 'commands' "
-                    f"section of the {PROJECT_FILE}.",
-                    exits=1,
-                )
+        for step_or_list in workflow_step_or_lists:
+            if isinstance(step_or_list, str):
+                verify_workflow_step(workflow_name, step_or_list)
+            else:
+                workflow_list = cast(List[str], step_or_list)
```
svlandeg, line 248: Before casting here, we want to add an assert to check the type just for robustness sake.

svlandeg (commented):
As a generic suggestion, when introducing new functionality like this it might be worth showing how it can be used in the PR description. This is useful for reviewers, but also for anyone who might stumble upon this PR at some point in the future :-) 

More specifically, I have been trying to experiment with this functionality but couldn't define a good example (in yaml) that works - can you provide one just for quick testing purposes?

spacy/cli/project/run.py
```diff
@@ -93,13 +119,30 @@ def project_run(
...
+def _get_workflow_steps(workflow_items: List[Union[str, List[str]]]) -> List[str]:
+    steps: List[str] = []
+    for workflow_item in workflow_items:
+        if isinstance(workflow_item, str):
+            steps.append(workflow_item)
+        else:
+            assert isinstance(workflow_item, list)
+            assert isinstance(workflow_item[0], str)
+            steps.extend(workflow_item)
+    return steps
```
richardpaulhudson, line 147: In these two cases (user does not enter a command or asks for `--help` about a command) I have abstracted away the parallel execution and just listed the commands in the workflow. In the output of `spacy project document`, on the other hand, I have made the parallel execution explicit. I think this is a good compromise between making everything too complicated and providing the necessary information; IMHO if someone just wants to know what a command does they are probably not that interested in implementation details like serial or parallel execution. I am open to being persuaded otherwise though :-).

spacy/cli/project/dvc.py
```diff
@@ -123,6 +125,11 @@ def update_dvc_config(
             dvc_cmd.append("--always-changed")
         full_cmd = [*dvc_cmd, *deps_cmd, *outputs_cmd, *outputs_nc_cmd, *project_cmd]
         dvc_commands.append(join_command(full_cmd))
+    if not processed_step:
```
richardpaulhudson, line 128: I realised this error handling was missing when testing this PR. Strictly speaking it could have gone in a separate PR, but because it was so small and I thought it might conflict with this PR I decided to add it in here.

spacy/cli/project/dvc.py
```diff
@@ -105,13 +105,15 @@ def update_dvc_config(
         dvc_config_path.unlink()
     dvc_commands = []
     config_commands = {cmd["name"]: cmd for cmd in config.get("commands", [])}
-    for name in workflows[workflow]:
+    processed_step = False
+    for name in get_workflow_steps(workflows[workflow]):
```
richardpaulhudson, line 109: As far as I could find out DVC does not offer anything comparable to multiprocessing groups, so that they need to be 'flattened out' when converting a spaCy config file to a DVC config file.

svlandeg:
> I have been trying to experiment with this functionality but couldn't define a good example (in yaml) that work

Retrying my previous experiment with the current PR code does work - some of the error handling you improved must have fixed it. Also thanks for the cool example in the PR description! That's definitely helpful. One comment I have there is that when you execute `spacy project run` it shows

> Available workflows in project.yml
> Usage: python -m spacy project run [WORKFLOW]
> 
> all   setup -> commandA -> commandB

which implies an order between `commandA` and `commandB` -> could you look into fixing that as well?

spacy/cli/project/run.py
svlandeg: I personally use `spacy project run` like ALL THE TIME when working with a spacy project, and I'd personally like to see the parallellization represented there. Would it be difficult to implement?

richardpaulhudson: There would be no difficulty at all in implementing it. The question is rather how it could be represented in a way that informs rather than confuses the user who may not necessarily be interested in such technical details as parallel vs. serial execution. Do you have a suggestion for _how_ you would like to see this information shown?

svlandeg:
One more thing I was realising is that this change does introduce a potential for breaking compatibility: when users create a .yml with this parallelization for a new version of spaCy, that same .yml will result in errors in older versions of spaCy, and the error is actually quite cryptic:
```
✘ Invalid project.yml. Double-check that the YAML is correct.
[workflows -> all -> 1] str type expected
```
I don't think we can do much about that though...

spacy/cli/project/run.py
svlandeg: Right now it would print
> all   setup -> commandA -> commandB

for me personally, something like

> all   setup -> commandA
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;-> commandB

(with the `->` aligned) would make it more clear that this is not a sequence but a parallel execution. But I'm not sure how feasible it is to actually implement that and make sure the vertical alignment works across different terminals etc. 

Alternatively, it could be something like 
> all   setup -> [commandA, commandB]

which would also make it easier to chain with subsequent steps.

richardpaulhudson: That is a clear syntax for the case where somebody types `spacy project run`. However, what I was really unsure how to tackle was the following situation:

```
> spacy project run all --help
Usage: python -m spacy project run all 

Workflow consisting of 3 commands:

1. setup      
2. commandA   
3. commandB   
```

Perhaps this syntax:

```
> spacy project run all --help
Usage: python -m spacy project run all 

Workflow consisting of 3 commands:

1.  setup      
2=. commandA   
2=. commandB   
```
?

svlandeg: Could we list them as
```
1. setup      
2a. commandA   
2b. commandB   
```
?

richardpaulhudson: Another option would be to add a single-character column to the table to the right of the numbers and use Ascii art to bracket together the multiprocessing groups (see `visualisation.py` [here](https://github.com/explosion/spaCy/pull/9930/files#diff-beda9f8c93177c5d5367ce855433a713f8ab9f70f781ec4cf1826aa46f52e30f) for an example of the technique I mean)

pmbaumgartner: I have an idea on syntax, but before I get there, I have a question about the `project.yml` syntax for this: is the only way to designate parallelism by nesting commands under `workflows`? I think that could be confusing to a user who wants to nest commands for a workflow but _doesn't_ intend for them to be run in parallel. Nesting would be a nice feature on it's own for organizing more complex project workflows.

Building off of that, if we first were to incorporate serial nested workflows, you could use some sort of syntax in those nested workflows to indicate parallelism - something like surrounding the commands with `(` and `)`. So for example, this would still be a serial project workflow:

```yaml
workflows:
  all:
    - setup
    - nestedCommandGroup
       - commandA
       - commandB
```

And would be the same as executing without the nested group while providing the user a way to indicate `commandA` and `commandB` are related.

Then, if you wanted to indicate those could be run in parallel, you could have something like:

```yaml
workflows:
  all:
    - setup
    - nestedCommandGroup
       - (commandA)
       - (commandB)
```

Then for `spacy project run` I'd probably prefer a hybrid of what's already been proposed, specifically:

```
1.  setup      
2. (commandA)
2. (commandB)
```

This way the indicator of parallelism lies with the command name, _not_ the numbering—which feels like we might be overloading it with this information.

richardpaulhudson: The big problem here is that `(commandA)` and `(commandB)` are valid command identifiers in the pre-existing system. This means that such identifiers may already be present in config files that people are using in a non-nested context, so that the only feature reliably indicating parallelism in the above `spacy project run` output would be the numbering.

It would be theoretically possible to implement a rule where bracketed names denoted parallelism in a nested context but were processed verbatim in a non-nested context, but this would get very confusing in config files with both types of bracketing, be complex to explain and to code and would also require rules to deal with cases like 

```
workflows:
  all:
    - setup
    - nestedCommandGroup
       - (commandA)
       - commandB
       - (commandC)
```

Because of the backwards-compatibility problem, it's probably best to avoid any solution based around the command name.

I think generally that nesting is as good a way as any to denote parallelism. It was Matt's original suggestion in the Ora ticket, although I don't know to what extent it was a considered suggestion and to what extent it was just an example. But I think it's going to get confusing for users if we introduce nesting as a supported option at the same time as introducing parallelism but without a strict relationship between the two. And if users just want to group commands they can do so using spacing between the command names.

rmitsch: **Logging**: from the suggested options I like the ASCII art one best. What about having both a command ID and a _timestep_ ID? Like, `setup` would be 1/1, `commandA` would be 2/2, `commandB` would be 3/2. Perhaps splitting the logical step from the time step would clear things up.
```
Workflow consisting of 3 commands:
| Command ID | Time step | Command  |
|------------|-----------|----------|
| 1          | 1         | setup    |
| 2          | 2         | commandA |
| 3          | 2         | commandB |
```

**Syntax in project.yml**: Like the nesting idea, but agree with @richardpaulhudson on potential confusion and the downsides of using brackets to denote parallelism. How about we make this more explicit on the level of the name of the nested group? For example:
```yaml
workflows:
  all:
    - setup
    - nestedCommandGroup::parallel
       - commandA
       - commandB
```
This would indicate that all commands in the nested group would be executed in parallel (I'd go for all or nothing in one nested group, I think mixing would complicate things too much).
Having a nested command group without `::parallel` would imply serial execution.

The downside to this would be adding an arbitrary marker like `::parallel`. Alternatively we could handle in proper  .yaml-style with dicts:
```yaml
workflows:
  all:
    - setup
    - nestedCommandGroup
        mode: parallel
        commands:
          - commandA
          - commandB
```

Both approaches are backwards-compatible, since these suggestions only affect nested command groups.

richardpaulhudson: This looks good. However, I'm finding it hard to see that people wouldn't be confused by nested command groups that didn't actually have any additional semantics, and wonder whether it shouldn't be a general principle of config files that different syntax corresponds to different semantics?

rmitsch: I'd enforce the same syntax for command groups, regardless of whether they are intended to be executed in parallel or not. For nested sequential command groups:
```yaml
workflows:
  all:
    - setup
    - nestedCommandGroup
        mode: sequential
        commands:
          - commandA
          - commandB
```
Or, if we have `mode` default to `sequential`:
```yaml
workflows:
  all:
    - setup
    - nestedCommandGroup
        commands:
          - commandA
          - commandB
```
Does this address your point?

richardpaulhudson: That much was clear, but I'm troubled by the fact that (unless I've misunderstood something):

```
workflows:
  all:
    - setup
    - nestedCommandGroup
        commands:
          - commandA
          - commandB
```

 would have identical semantics to
 
```
workflows:
  all:
    - setup
    - commandA
    - commandB
```

Generally in a config file I see it as being undesirable for there to be multiple ways of achieving the same behaviour.

rmitsch: Ah, I see - misunderstood you before. Yes, I agree. Unless nested command groups have some other user-modifiable properties (should they? Not in the loop with this one), it doesn't make too much sense to have them in a sequential version IMO.

pmbaumgartner: Yeah, I agree that's potentially confusing to have the multiple behaviors for the same end result, but it's also a totally optional change from the flat to nested syntax. I think this pattern of short "string" vs. long "dict" values is pretty common - e.g. [`docker-compose` ports: Short and Long syntax](https://docs.docker.com/compose/compose-file/compose-file-v3/#ports). To me the implication that a nested group is automatically parallel is more confusing because the user never explicitly states they want parallelism. It also locks us into that behavior for any nesting of commands, having to make _future_ changes to nesting backwards-compatible with the parallelism—which is why I proposed distinguishing serial from parallel in some way other than the implication from nesting.

I have an alternative idea from the closest thing I've used to the spaCy project config, which is [Taskfile](https://taskfile.dev/usage/#task-dependencies). The model there is a little different because you can set dependent tasks, and if a task has multiple dependencies, they'll automatically be run in parallel. They have an option to run specific non-dependent commands in parallel when they're executed, so translating to our use case: rather than defining parallelism in the flag, we could make `spacy project run` take a `--parallel` arg and a variable number of args that are commands to run in parallel. 

The main downside with that I see is that there's no way to document that commands _can/should_ be run in parallel in the config. However, part of me thinks that parallelism semantically makes more sense as an execution option (with the `run` command), rather than something that is in the config, mostly because I can see it being easy to modify a command in your config and have it silently fail because it's now dependent on another command.

honnibal: I can see the complaint that the 'unnamed nesting' thing is too cryptic and the parallelism will be surprising. But I think the naming feels surprising, it doesn't do anything if it's done serially like that and it feels too overlapping with the subcommands as has been said.

I think I'd like this better than the named subgrounds with an option 'parallel':

```
workflows:
  all:
    - setup
    - parallel:
      - commandA
      - commandB
```

In terms of the grammar of the config it feels a little ugly to me that you can have this object that has one key, (`parallel`), having a list there feels a little more aesthetic to me. But the object would maybe make things a bit more flexible in future, and does make it more obvious what's going on.

spacy/cli/project/dvc.py
honnibal: We could also consider raising an error rather than flattening in this case?

spacy/cli/project/run.py
```diff
@@ -67,21 +71,62 @@ def project_run(
...
+                ):
+                    num_processes = max_parallel_processes
+                process_iterator = iter(processes)
+                for _ in range(num_processes):
+                    next(process_iterator).start()
+                for _ in range(len(steps_list)):
+                    completion_queue.get()
+                    next_process = next(process_iterator, None)
+                    if next_process is not None:
+                        next_process.start()
```
njsmith, line 128: Wouldn't it be simpler to use a `multiprocessing.Pool`, or even `concurrent.futures.ProcessPoolExecutor`? That way you just tell the library how many processes you want running at once, and it takes care of scheduling the individual tasks. Or is the idea that you specifically want to kill each process after it's finished one job, and spin up a new fresh process for the next one? I don't think `concurrent.futures` exposes that level of control, but with `multiprocessing.Pool` it's just a matter of passing `maxtasksperchild=1` to the constructor.

Also, if you're using multiprocessing, you should probably spend a few moments thinking about which [start method](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) you want. On Windows and macOS the default is to re-invoke python from scratch, which is a bit heavy but safe, and there aren't really any other options. But on Linux, the default is to use `fork`, which is fast, *but* if the parent process is doing anything complex (in particular, using multithreading, either directly or via some library like BLAS), then `fork` can cause all kinds of nasty problems. `forkserver` or `spawn` are safer defaults.

richardpaulhudson: I've investigated the various ways of parallelising commands and [documented the results with code snippets](https://github.com/richardpaulhudson/spacy_projects_multiprocessing). if you follow the argumentation there, the current architecture with multiprocessing queues represents the most appropriate way of solving the problem, but changes are necessary to better manage error handling and logging and to ensure cross-platform consistency.

I'd be very grateful for feedback, especially on the specific issues set out in the last section.

richardpaulhudson:
I've investigated the various ways of parallelising commands and [documented the results with code snippets](https://github.com/richardpaulhudson/spacy_projects_multiprocessing). if you follow the argumentation there, the current architecture with multiprocessing queues represents the most appropriate way of solving the problem, but changes are necessary to better manage error handling and logging and to ensure cross-platform consistency.

I'd be very grateful for feedback, especially on the specific issues set out in the last section.

rmitsch:
> I'd be very grateful for feedback, especially on the specific issues set out in the last section.

That's a very thorough write-up! :clap:  

> `multiprocessing.Pool` provides a convenient method for collecting the output from multiple parallel processes and returning when they have all completed.

I think `Pool` actually does return/proceed whenever a child process has finished, not just when _all_ of them do. This doesn't change any of the other points you make.

I agree that `Queue` seems like a reasonable choice.

> 4.4 Managing console output

Fully agree that child processes shouldn't log to the console directly, that's just a mess (`stderr` is perhaps fine, as it shouldn't happen often anyway). Output to be displayed in the console could also be stored in a queue though - while it's not "real-time", at least we wouldn't have to wait for the process to finish in order to see the log and on average not as much log would be lost for killed processes.

> 4.5 Executing serial commands

Agreed. Might also make debugging easier.

richardpaulhudson:
Thanks for your very timely feedback @rmitsch! 

>  Output to be displayed in the console could also be stored in a queue though - while it's not "real-time", at least we wouldn't have to wait for the process to finish in order to see the log and on average not as much log would be lost for killed processes.

I completely agree that we should try and avoid losing `stdout` output for stored processes. Up until now I thought I had to do something else in the worker process where the subprocess is started, but it turns out that `example_queueing_with_output_management.py` already does everything it can to catch `stdout` output. The problem is rather in the subprocess code itself: if you start a Python script without `PYTHONUNBUFFERED=1`, it buffers its `stdout` output and anything that's in the buffer is lost when the process is killed; if you call subprocess code that contains a `sys.stdout.flush()`, the output is retained by the worker process even if the subprocess is killed. And because we can't control what subprocess code does, there's nothing we can really do about this problem.

For subprocesses that do write to `stdout` regularly, do you (and other people) think it's better to save up all the output in one go, or to write it in chunks? Writing it in chunks from the main process would avoid the effect where logging output from different processes ends up on one line, but it would still make the logging output harder to read than if all the output for each process were grouped together. At the same time, though, writing it in chunks would be closer to real-time output.

richardpaulhudson:
This is generally an excellent idea, although I'm not sure about the `Status` column as this is log output rather than an interactive GUI.

rmitsch:
> This is generally an excellent idea, although I'm not sure about the `Status` column as this is log output rather than an interactive GUI. I'm not sure it's really necessary though, because: ...

Yeah, it would have to be updated in-place `tqdm`-style - or re-printed every x seconds. You are correct, it's not strictly necessary, as it can be inferred from the logging output. However, I think it might be easy to lose track of which processes have failed/are done at any point if there's a lot of logging output (which we can't/don't want to control).

richardpaulhudson:
> Yeah, it would have to be updated in-place `tqdm`-style - or re-printed every x seconds

Printing it `tqdm` style would be great from a usability point of view — can anyone else think of any reasons that would speak against doing that? (`tqdm` isn't used within spaCy projects at present, although it is already a spaCy dependency).

Btw

> printing all the output to stdout once all processes are done as suggested by you

was not exactly my suggestion, rather printing the output to stdout for each process when that process completes (although if a process fails, other processes in the group will be killed which will lead to their output being printed as well shortly afterwards). This means that if the status table is printed out `tqdm`-style, it will need to be rerendered after the stdout block for each completed process.

rmitsch:
> Btw
> 
> > printing all the output to stdout once all processes are done as suggested by you
> 
> was not exactly my suggestion...

Ah, ok. Actually, does it even make sense to (messily) log to `stdout` _if_ we log to file and have a status indicator?

richardpaulhudson:
> Actually, does it even make sense to (messily) log to `stdout` if we log to file and have a status indicator on `stdout`?

Hmm, I see what you mean. I think we need to distinguish between the output from the worker process when a job finishes and any output the subprocess might have logged to `stdout`:

- as the worker process output won't end up in the subprocess log files, we definitely want to retain it, also for consistency with serially executed commands. I agree that if we have a status indicator it doesn't make much difference whether the worker process output for a given subprocess is displayed when the subprocess terminates or when the whole group does, though.
- I can see that if a subprocess logs lots of output to `stdout` it probably doesn't make sense to duplicate that output to the main console. But if it only logs a couple of lines (which many subprocesses probably do) it would be nice to see those lines alongside the worker process output for that subprocess. So one possibility would be to duplicate the output on the console if it's less than e.g. 20 lines, and to display a message that `n` lines of output were logged if it's more than 20 lines. Or is this getting overengineered?

rmitsch:
> Hmm, I see what you mean. I think we need to distinguish between...

The distinction makes sense, and I fully agree on the first point. Regarding subprocess logs in the main console: perhaps we could just start with a binary option (display subprocess logs - yes/no) and then see if users want a line limit? If we don't even want a binary choice*, I'd go without no log replication to the console.

* Configuring this would require changes to the syntax, not sure that's worth it.

richardpaulhudson:
> Configuring this would require changes to the syntax, not sure that's worth it.

The PR already introduces changes to the projects file syntax. I don't think more are a problem, especially if there's default behaviour so that existing project files remain valid.

spacy/cli/_util.py
```diff
@@ -556,3 +602,121 @@ def setup_gpu(use_gpu: int, silent=None) -> None:
         local_msg.info("Using CPU")
         if has_cupy and gpu_is_available():
             local_msg.info("To switch to GPU 0, use the option: --gpu-id 0")
```
richardpaulhudson, line 604: These methods have been moved here from `run.py` as they are now also called from `parallel.py`.

spacy/cli/project/run.py
```diff
@@ -204,107 +256,3 @@ def validate_subcommand(
             ". ".join(help_msg),
             exits=1,
         )
-
```
richardpaulhudson, line 207: This code has been moved to `_util.py` as it is also called from  `parallel.py`.

richardpaulhudson:

richardpaulhudson:

richardpaulhudson:

spacy/cli/_util.py
rmitsch: Are there any changes in `check_rerun()`, `update_lockfile()`, `check_deps()`, `_get_lock_entry()`, `_get_fileinfo()`?

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+PARALLEL_LOGGING_DIR_NAME = "parrTmp"
+
+# The maximum permissible width of divider text describing a parallel command group
+MAX_WIDTH_DIVIDER = 60
+
+# Whether or not to display realtime status information
+DISPLAY_STATUS_TABLE = sys.stdout.isatty() and supports_ansi()
+
+
+class _ParallelCommandState:
```
rmitsch, line 51: Maybe as `dataclass` instead?

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+    ):
+        self.name = name
+        self.display_color = display_color
+        if transitions is not None:
+            self.transitions = transitions
+        else:
+            self.transitions = []
+
+
+_PARALLEL_COMMAND_STATES = (
```
rmitsch, line 67: Is it necessary to have this separately or can it be defined directly as `_ParallelCommandInfo.state_dict`?

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+        self.cmd_index = cmd_index
+        self.state = _ParallelCommandInfo.state_dict["pending"]
+        self.pid: Optional[int] = None
+        self.last_keepalive_time: Optional[int] = None
+        self.running_os_cmd_index: Optional[int] = None
+        self.rc: Optional[int] = None
+        self.console_output: Optional[str] = None
+
+    def change_state(self, new_state: str):
+        assert new_state in self.state.transitions, (
```
rmitsch, line 112: Not sure about the convention, but I think we want all errors to be properly defined error codes instead of `asserts`.

spacy/cli/_util.py
richardpaulhudson: OK, sorry, I see now that my original comment wasn't really very helpful :-). So I can confirm there are no changes to the behaviour of any of this code, it's unchanged except for:

- the names of `get_lock_entry()` and `get_fileinfo()` have been changed to `_get_lock_entry()` and `_get_fileinfo()`
- `check_deps()` has been moved into a separate function (it was previously part of `run.project.run()`
- within `check_deps()` the name of the `subcommand` variable has been changed to `cmd_name`

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+    # group is not allowed to contain the same command more than once
+    log_file_name = os.sep.join(
+        (current_dir, PARALLEL_LOGGING_DIR_NAME, cmd_info.cmd_name + ".log")
+    )
+    file_not_found = False
+    # buffering=0: make sure output is not lost if a subprocess is terminated
+    with open(log_file_name, "wb", buffering=0) as logfile:
+        for os_cmd_index, os_cmd in enumerate(cmd_info.cmd["script"]):
+            command = split_command(os_cmd)
+            if len(command) and command[0] in ("python", "python3"):
```
rmitsch, line 335: This seems a bit brittle to me (if I understand it correctly). Isn't it possible that the interpreter could also be python3.x?

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+                                "type": "keepalive",
+                                "cmd_index": cmd_info.cmd_index,
+                            }
+                        )
+                    else:
+                        break
+
+                if sp.returncode != 0:
+                    if os.name == "nt":
+                        status = "Failed/terminated"
```
rmitsch, line 394: If we have overlapping status values, we could also consider bit masking. If this is the only case though (I think it is?), then it doesn't make sense/would be overengineering.

spacy/schemas.py
```diff
@@ -458,10 +458,11 @@ class ProjectConfigSchema(BaseModel):
     vars: Dict[StrictStr, Any] = Field({}, title="Optional variables to substitute in commands")
     env: Dict[StrictStr, Any] = Field({}, title="Optional variable names to substitute in commands, mapped to environment variable names")
     assets: List[Union[ProjectConfigAssetURL, ProjectConfigAssetGit]] = Field([], title="Data assets")
-    workflows: Dict[StrictStr, List[StrictStr]] = Field({}, title="Named workflows, mapped to list of project commands to run in order")
-    commands: List[ProjectConfigCommand] = Field([], title="Project command shortucts")
+    workflows: Dict[StrictStr, List[Union[StrictStr, Dict[Literal["parallel"], List[StrictStr]]]]] = Field({}, title="Named workflows, mapped to list of project commands to run in order")
```
rmitsch, line 461: I didn't know about `Literal`. Neat!

rmitsch (commented):
Great work! I'll need more time to digest the core parallelization logic and run a few tests, but looks reasonable so far. These are just a few nitpicks.

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+                    other_cmd_info.last_keepalive_time is not None
+                    and time() - other_cmd_info.last_keepalive_time
+                    > PARALLEL_GROUP_STATUS_INTERVAL * 20
+                ):
+                    # a specific command has hung
+                    other_cmd_info.change_state("hung")
+            if mess["type"] == "started":
+                cmd_info.change_state("running")
+                cmd_info.running_os_cmd_index = cast(int, mess["os_cmd_index"])
+                cmd_info.pid = cast(int, mess["pid"])
```
rmitsch, line 250: IMO the `cast` isn't necessary here. I don't see any wrong type hints without it, at least.

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+                # a command in the group hasn't succeeded, so terminate/cancel the rest
+                for other_cmd_info in (
+                    c for c in cmd_infos if c.state.name == "running"
+                ):
+                    try:
+                        os.kill(cast(int, other_cmd_info.pid), SIGTERM)
+                    except:
+                        # there could be a race condition between the process completing
+                        # and the message arriving to notify about it
+                        pass
```
rmitsch, line 277: Could you elaborate on what happens with the stray process in this case?

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+                for other_cmd_info in (
+                    c for c in cmd_infos if c.state.name == "pending"
+                ):
+                    other_cmd_info.change_state("cancelled")
+            if mess["type"] != "keepalive" and DISPLAY_STATUS_TABLE:
+                if first:
```
rmitsch, line 283: Maybe a more informative variable name than `first`.

spacy/tests/test_parallel.py
```diff
@@ -0,0 +1,238 @@
+import os
+from time import time
+import pytest
+import srsly
+from spacy.cli._util import load_project_config
+from spacy.cli.project.run import project_run
+from .util import make_tempdir
+
+
```
rmitsch, line 9: Could you add descriptions to the docstrings describing the corresponding tests briefly?

spacy/tests/test_parallel.py
```diff
@@ -0,0 +1,238 @@
...
+        project_run(d, "all")
+        assert os.path.exists(os.sep.join((str(d), "A")))
+        assert os.path.exists(os.sep.join((str(d), "B")))
+        assert os.path.exists(os.sep.join((str(d), "C")))
+        assert os.path.exists(os.sep.join((str(d), "D")))
+        assert os.path.exists(os.sep.join((str(d), "E")))
+
+
+@pytest.mark.parametrize("failing_command", ["python", "someNotExistentOSCommand"])
+def test_project_run_multiprocessing_failure(failing_command:str):
```
rmitsch, line 187:
```suggestion
def test_project_run_multiprocessing_failure(failing_command: str):
```

spacy/cli/_util.py
```diff
@@ -215,30 +220,71 @@ def validate_project_version(config: Dict[str, Any]) -> None:
...
+            f"Unknown command specified in workflow '{workflow_name}': {step}",
+            f"Workflows can only refer to commands defined in the 'commands' "
+            f"section of the {PROJECT_FILE}.",
+            exits=1,
+        )
+
+
 def validate_project_commands(config: Dict[str, Any]) -> None:
     """Check that project commands and workflows are valid, don't contain
     duplicates, don't clash  and only refer to commands that exist.
```
rmitsch, line 250:
```suggestion
    duplicates, don't clash and only refer to commands that exist.
```

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+from subprocess import STDOUT, Popen, TimeoutExpired
+from wasabi import msg
+from wasabi.util import color, supports_ansi
+
+from .._util import check_rerun, check_deps, update_lockfile, load_project_config
+from ...util import SimpleFrozenDict, working_dir, split_command, join_command
+from ...util import check_bool_env_var, ENV_VARS
+from ...errors import Errors
+
+"""
```
rmitsch, line 19: IIRC it's convention to add the docstring at the very top of the file.

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+            check_deps(cmd_info.cmd, cmd_info.cmd_name, project_dir, dry)
+            if (
+                not check_rerun(
+                    current_dir, cmd_info.cmd, check_spacy_commit=check_spacy_commit
+                )
+                and not force
+            ):
+                cmd_info.change_state("not rerun")
+        rmtree(PARALLEL_LOGGING_DIR_NAME, ignore_errors=True)
+        os.mkdir(PARALLEL_LOGGING_DIR_NAME)
```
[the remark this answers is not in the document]
rmitsch: Also - I might have missed it, but do we check whether this directory has been created by the user before we delete it? In general I'm in favor for using `tempfile.TemporaryDirectory()` for something like this, since (a) the output is printed to the console anyway and (b) if users want to keep the log files, they can copy them to a non-temporary location afterwards.

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+                cmd_info.console_output = cast(str, mess["console_output"])
+            if any(
+                c for c in cmd_infos if c.state.name in ("failed", "terminated", "hung")
+            ):
+                # a command in the group hasn't succeeded, so terminate/cancel the rest
+                for other_cmd_info in (
+                    c for c in cmd_infos if c.state.name == "running"
+                ):
+                    try:
+                        os.kill(cast(int, other_cmd_info.pid), SIGTERM)
```
rmitsch, line 273: Why are you `cast`ing here?

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+    cmd_info: _ParallelCommandInfo,
+    *,
+    dry: bool,
+    current_dir: str,
+    parallel_group_status_queue: Queue,
+) -> None:
+    """Run a single spaCy projects command as a worker process.
+
+    Communicates with the main process via queue messages that consist of
+    dictionaries whose type is determined by the entry 'type'. Possible
```
rmitsch, line 322: Nitpick: I'm not sure if `type` is ideal here due to its connotation to variable types. Maybe `status`? Feel free to ignore though.

spacy/tests/test_parallel.py
```diff
@@ -0,0 +1,238 @@
...
+    with open(os.sep.join((d_path, first_flag_filename)), 'w') as first_flag_file:
+        first_flag_file.write("")
+else: # should never happen because of skipping
+    with open(os.sep.join((d_path, second_flag_filename)), 'w') as second_flag_file:
+        second_flag_file.write("")
+        """
+
+        pscript_loc = os.sep.join((str(d), "pscript.py"))
+        with open(pscript_loc, "w") as pscript_file:
+            pscript_file.write(pscript)
```
rmitsch, line 99: Think it'd be better to do this as temporary file.

spacy/tests/test_parallel.py
```diff
@@ -0,0 +1,238 @@
...
+from time import sleep
+
+_, sleep_secs, rc = sys.argv
+sleep(int(sleep_secs))
+sys.exit(int(rc))
+        """
+
+        pscript_loc = os.sep.join((str(d), "pscript.py"))
+        with open(pscript_loc, "w") as pscript_file:
+            pscript_file.write(pscript)
```
rmitsch, line 201: Same as above: I'm in favor of temporary files wherever possible.

spacy/cli/project/parallel.py
richardpaulhudson: I agree it's better off defined within `_ParallelCommandInfo`, but want to generate `state_dict` from it to ensure we can't get into a situation where a name points to a state in the dictionary that doesn't actually have that name.

spacy/cli/project/parallel.py
richardpaulhudson: Oh yes I see — I had copied this from `run.py` (ll. 220 onwards) but the original motivation for the feature there was a bit different. It's safe to assume that if someone specifies e.g. `python3.9` they know that's exactly what they want, but in the parallel situation we should still add `-u`.

Do you have any ideas about how we could handle the alias situation?

spacy/cli/project/parallel.py
richardpaulhudson: This is the only case, so I think bit masking would be too much.

spacy/cli/_util.py
```diff
@@ -215,30 +220,71 @@ def validate_project_version(config: Dict[str, Any]) -> None:
...
-                )
+        for workflow_item in workflow_items:
+            if isinstance(workflow_item, str):
+                verify_workflow_step(workflow_name, commands, workflow_item)
+            else:
+                assert isinstance(workflow_item, dict)
+                assert len(workflow_item) == 1
+                steps_list = workflow_item["parallel"]
+                assert isinstance(steps_list[0], str)
+                steps = cast(List[str], steps_list)
```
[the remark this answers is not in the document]
richardpaulhudson: You're right that this would be better, but this also raises the general point you made above about asserts: perhaps we should be throwing a `ValueError` here instead?

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+
+# The dirname where the temporary logfiles for a parallel group are written
+# before being copied to stdout when the group has completed
+PARALLEL_LOGGING_DIR_NAME = "parrTmp"
+
+# The maximum permissible width of divider text describing a parallel command group
+MAX_WIDTH_DIVIDER = 60
+
+# Whether or not to display realtime status information
+DISPLAY_STATUS_TABLE = sys.stdout.isatty() and supports_ansi()
```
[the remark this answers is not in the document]
richardpaulhudson: If there status table cannot be displayed in real time, it isn't displayed at all because it wouldn't serve any useful purpose. If ANSI is not available, nothing is displayed unless the `parallel[..., ...]` description was too wide to be displayed in the divider, in which case it is outputted again in full.

spacy/cli/project/parallel.py
richardpaulhudson: Strange, my `mypy` complains. The problem is that the dictionary entries in the message have the type `Union[str, int]` and have to be cast to whichever type the target variable has.

spacy/cli/project/parallel.py
richardpaulhudson: I'll update the comment in the code, but the list of processes to kill is based on the main process' knowledge of what the worker processes are doing. It could be the case that a subprocess has already completed but that the message from the worker process to the main process notifying about this is still in the queue.

spacy/cli/project/parallel.py
richardpaulhudson: The type of the variable is `Optional[int]`.

spacy/cli/project/parallel.py
richardpaulhudson: I've converted it to a dataclass and don't understand why this doesn't now have the `Outdated` tag.

spacy/cli/project/parallel.py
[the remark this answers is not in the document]
richardpaulhudson: Or do you feel it would make sense to display the status table a single time once the group has completed?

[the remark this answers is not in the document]
rmitsch: No, I think not displaying it at all if it can't be shown in real time is reasonable.

spacy/cli/project/parallel.py
rmitsch: Thanks for the context and the updated comment.

spacy/cli/project/parallel.py
rmitsch: Ah, `mypy` complains indeed otherwise.

spacy/cli/_util.py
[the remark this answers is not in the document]
rmitsch: I looked it up since I wasn't sure about how strict we are about this. Our code conventions mention that
> As a general rule of thumb, all error messages raised within the core library should be added to `Errors`.

So I'd say let's convert all raw asserts into `ValueError`s.

spacy/cli/project/parallel.py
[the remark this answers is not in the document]
rmitsch: Forgot about (c), which is also important: it automatically avoids all conflicts with pre-existing file/directory names.

spacy/cli/project/parallel.py
rmitsch: Ah, I see. Since this is the same behavior as in the existing logic I'd ignore it anyway.

spacy/cli/_util.py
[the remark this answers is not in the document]
richardpaulhudson: In fact the project file has already been validated against the schema, so these `assert`s are not really necessary in the first place: I'll get rid of them (as well as implementing your suggestion).

spacy/tests/test_parallel.py
richardpaulhudson: Is that necessary given that we are already writing to a temporary directory? Does a temporary file have additional benefits in this situation?

rmitsch: Ah, I completely missed you're doing `make_tempdir()` anyway :slightly_smiling_face:

spacy/cli/project/parallel.py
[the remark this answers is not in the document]
rmitsch: I think you have some good points there - having default path(s) is certainly useful, and it's rather unlikely users have a `parrTmp` directory. Regarding the last point: a temporary directory can be configured to not be deleted once the corresponding context has been closed.

In general I'd go for the standard way nonetheless, as I think the advantages don't really justify the non-standard approach here. I don't mind this approach either though as long as we are checking and warning (with confirmation maybe) that the existing `parrTmp` directory will be deleted. Maybe a second opinion would be useful for this.

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,442 @@
...
+
+
+def project_run_parallel_group(
+    project_dir: Path,
+    cmd_names: List[str],
+    *,
+    overrides: Dict[str, Any] = SimpleFrozenDict(),
+    force: bool = False,
+    dry: bool = False,
+) -> None:
```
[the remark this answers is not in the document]
richardpaulhudson: I've refactored it into two functions and also moved some of the functionality in the main function around and am interested in whether you think it's clearer now.

spacy/cli/project/parallel.py
richardpaulhudson: I've changed the name to `mess_type`.

spacy/cli/project/parallel.py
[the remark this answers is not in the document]
richardpaulhudson: I'd rather go the temporary directory route, but:

1) I don't want to delete the temporary directory after the parallel group has completed in case a user wants to access the files
2) I don't want temporary directories to be created with each invocation of a workflow and never deleted

Perhaps a good compromise would be to delete the temporary directory if everything succeeded and retain it if something went wrong? What do you think?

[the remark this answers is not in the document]
rmitsch: 1. AFAIK on Linux temporary files and directories are deleted after a certain number of days or on system restart and in Windows never by default.
2. This is probably a dealbreaker then for temporary directories, see 1.

I think logs should be kept independently from success or failure. Perhaps we do have a convention regarding temporary files already though?

spacy/cli/project/parallel.py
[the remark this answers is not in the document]
rmitsch: Yes, absolutely :slightly_smiling_face:

spacy/cli/project/parallel.py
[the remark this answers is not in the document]
richardpaulhudson: @rmitsch and I discussed this offline and decided to go for standard temporary directories; I have changed the code accordingly.

richardpaulhudson:

richardpaulhudson:

rmitsch (commented):
A last round of comments, then I'd approve the PR.

rmitsch:
Not approving yet because of the Windows/ANSI/`wasabi` issue, but otherwise I'm done with my review. LGTM!

richardpaulhudson:
When a command within a parallel group has failed, the other commands in the group are terminated. Here a race condition can occur where the worker process to be terminated has just completed of its own accord and no longer exists.  One of the last batches of changes based on review comments was to narrow the scope of the exception caught when handling this race condition.

One CI run failed on a subsequent run, suggesting that the new, narrower exception does not apply on Windows and that a different exception would be necessary there. I reran the CI to test the assumption that this race condition (which occurs fairly rarely - I've never seen it anywhere outside the context of a CI runner, presumably because the CI runner does not have consistent access ) was the probable culprit and indeed it passed the second time. I think it is very likely to have been the race condition that caused the exception although it could also have been some strange one-off with the CI runner.

I'm on vacation until mid-August and when I get back I'll do more tests on a Windows machine to find out exactly what's going on. Presuming I'm right about the race condition, I'll then wider the scope of the `try - except` block to handle the Windows exception and also double-check what happens on macOS.

Apart from this issue and the Wasabi/Windows issue, this PR is ready for further reviews.

spacy/cli/project/parallel.py
```diff
@@ -0,0 +1,467 @@
...
+
+# How often the worker processes managing the commands in a parallel group
+# send keepalive messages to the main process (seconds)
+_PARALLEL_GROUP_STATUS_INTERVAL = 1
+
+# The maximum permissible width of divider text describing a parallel command group
+_MAX_WIDTH_DIVIDER = 60
+
+# Whether or not to display realtime status information
+_DISPLAY_STATUS_TABLE = sys.stdout.isatty() and supports_ansi()
```
svlandeg, line 45: Unfortunately it looks like `isatty()` doesn't always work correctly on Windows, e.g. with Git Bash: https://stackoverflow.com/questions/40912072/git-for-windows-mintty-sys-stdout-isatty-returns-false

richardpaulhudson: Do you have any ideas about how this could be improved? The aim of `isatty()` is to check we're on a console rather than writing to a log file to decide whether or not to use the richer GUI with cursor repositioning. In this link it seems that `isatty()` is returning false negative results, which is not great but I think we can live with: it means a user who could use the richer GUI gets the simpler one. IMHO it would be a bigger problem if `isatty()` returned false positive results because a user would then get the richer GUI in a situation where it wouldn't be rendered properly.

richardpaulhudson:

spacy/cli/project/parallel.py
richardpaulhudson: It seems this was actually fixed in 2019: https://github.com/mintty/mintty/issues/56?

svlandeg: Unfortunately I did run into this issue back in August 2022 so I'd like to keep this comment open for now to potentially explore other alternatives or workarounds, if there are any.

richardpaulhudson: Do you remember whether you specifically tested the value of `is_atty()` and found it to be falsely negative, or whether you observed that the whole statement

```
sys.stdout.isatty() and supports_ansi()
```

returned a false-negative result and hence that you were not seeing the richer GUI on Windows? If the second is the case, this will have been because the second condition in the statement, `supports_ansi()`, was returning a false-negative result because of a problem with Wasabi and Colorama that has since been fixed: https://github.com/explosion/wasabi/pull/31. I can confirm that I got the fallback GUI on Windows before this was merged and the rich GUI on Windows after it was merged.

svlandeg: Thanks for the context! I'll have to go back and double check with the new version of `wasabi`.

svlandeg:
I'm adding this one to our backlog - when we continue working on this, we'll need to move this into 'weasel'.

PR #10784, opened 2022-05-11
`enable` argument for spacy.load()

rmitsch:
## Goals

Introduce `enable` argument to `spacy.load()`, behaving analogously to and `disable`.

## Description

`enable` takes precedence over their negative counterparts. I.e.: if `enable` is set, `disable` has to be consistent. Otherwise an error is raised. Internally, `enable` is resolved to `disable` by disabling those components in the pipeline that are not part of `include`.

Example:
```python
spacy.load("en_core_web_sm", enable=["tagger", "parser"])
```
This disables all components other than the tagger and the parser.

### Types of change

New feature, new tests.

## Checklist


spacy/__init__.py
```diff
@@ -42,14 +44,24 @@ def load(
     disable (Iterable[str]): Names of pipeline components to disable. Disabled
         pipes will be loaded but they won't be run unless you explicitly
         enable them by calling nlp.enable_pipe.
+        enable (Iterable[str]): Names of pipeline components to enable. All other
+            pipes will be disabled (and can be enabled using `nlp.enable_pipe`).
     exclude (Iterable[str]): Names of pipeline components to exclude. Excluded
         components won't be loaded.
+    include (Iterable[str]): Names of pipeline components to include. All other
+        components will be excluded.
```
svlandeg, line 52: I think it's slightly more clear to say "All other components will not be loaded".

spacy/language.py
```diff
@@ -1835,8 +1844,17 @@ def from_config(
         # Restore the original vocab after sourcing if necessary
         if vocab_b is not None:
             nlp.vocab.from_bytes(vocab_b)
-        disabled_pipes = [*config["nlp"]["disabled"], *disable]
+
+        # Resolve disabled/enabled settings.
+        # todo @RM ensure "enabled" is in config
```
svlandeg, line 1849: Marking TODO comment here to resolve before PR gets merged

spacy/language.py
```diff
@@ -1988,11 +2016,52 @@ def to_disk(
...
+        assert arg_type in ("exclude", "disable")
+
+        if active:
+            if inactive:
+                warnings.warn(
+                    Warnings.W120.format(
+                        arg1="include" if arg_type == "exclude" else "enable",
+                        arg2=arg_type,
+                    )
+                )
```
svlandeg, line 2052: This is a reasonable strategy, but I wonder whether it would be confusing because we do something slightly different in `nlp.select_pipes`: There we raise an exception if the two settings are both set but notcompatible:
```
        if enable is not None:  
            ...
            to_disable = [pipe for pipe in self.pipe_names if pipe not in enable]
            
            # raise an error if the enable and disable keywords are not consistent
            if disable is not None and disable != to_disable:
                raise ValueError(
                    Errors.E992.format(
                        enable=enable, disable=disable, names=self.pipe_names
                    )
                )
```

svlandeg (commented):
I'd love to hear some thoughts/reviews by other colleagues as well, but in principle I think this will be really useful. It makes the code & docs slightly more complex, but it will have a huge impact on many use-cases where you really just want one or 2 components and disable everything else. It feels like such a common use-case that we might as well cater for it...

spacy/language.py
rmitsch: Hm, I didn't consider that. I agree that we should keep the warnings consistent. Then I'll scrap `W120` and go with `E992` on incompatibility?

rmitsch: Using a similar mechanism and error message now - let me know what you think. Latest commit also includes a minor test refactoring. Apparently I also broke something, since `test_load_without_strings()` fails now. Will take a look at that next week.

And sorry for rebasing - force of habit, noticed too late :grimacing: Last time :crossed_fingers:

rmitsch:
> The Python 3.8 mac build includes some additional CLI tests that aren't run in the other builds (just to avoid excessive downloads/time), so these failures do matter, and aren't particular to python 3.8 / macos.
> 
> I didn't realize that the plan was to add this setting to the config? I thought this was just adding new keyword arguments to `spacy.load`?
> 
> Edited to add: I don't think it makes sense to have two conflicting features in the config like `disable` and `include`. Someone writing a config can figure out what the right settings are with `pipeline` and `disable`. Someone loading a model with `spacy.load` has a good reason to want an `include` option that makes it easier to selectively load things.

Adding it to the config isn't necessary. It was my assumption we'd want to handle `enable` the same way as `disable`, which would mean adding it to the config. I think you have a point there, I'd be ok with removing it from the config again. What does @svlandeg think?

svlandeg:
Agreed with Adriane's points!

rmitsch:
Removed `enabled` from the config. [This](https://github.com/explosion/spaCy/pull/10784#discussion_r873705104) is the last remaining open question from my side.

spacy/language.py
adrianeboyd: There's not a complete list of fields supported by `exclude` for serialization and it's allowed for components to define their own, although I doubt it's done much in practice. `exclude` is passed to all the serialization methods so this needs to be supported.

I think you have to allow arbitrary extra things in `exclude`.

adrianeboyd: Then I think the contrast between `exclude` (anything) and `include` (pipeline components) is going to be weird. How weird would it be to just allow extra ignored things in `include` like with `exclude`?

adrianeboyd: Although I guess `disable` is also (pipeline components), so maybe it's not that weird. But the checks are going to need to allow arbitrary things in `exclude`.

rmitsch: I quite dislike this inconsistency, I think it's rather confusing from a usage perspective. I don't see how we can avoid it however as long as it's not possible to retrieve all relevant fields.

I relaxed the check now so that arbitrary terms are allowed in `exclude`, but ignored once `include` is not empty. This leads to inconsistent component exclusions:
```python
# "strings" is not in nlp.pipe_names

# Will exclude ("strings",)
spacy.load(tmp_dir, exclude=["strings"])  
# Will exclude all components in pipeline except "tagger", but not "strings"
spacy.load(tmp_dir, include=["tagger"], exclude=["strings"])  
```
I believe this to be an rather unexpected result. This could be handled differently by adding arbitrary elements in `exclude` (such as `"strings"`) to the exclusions even if `included` is set. This would be inconsistent in different ways though:
```python
# "strings" is not in nlp.pipe_names

# Will exclude "strings" + all components in pipeline except "tagger" 
spacy.load(tmp_dir, include=["tagger"], exclude=["strings"]) 
# Will exclude all components in pipeline "except" tagger, but not "strings" (as we are not aware of it)
spacy.load(tmp_dir, include=["tagger"])  
```
Again not ideal IMO, as I'd expect `include` to exclude _everything_ else.

If I understand correctly, option (1) is more closely aligned with the behaviour of `select_pipes()`, so I'd keep this and adjust the documentation, if we agree on this. Perhaps creating an issue for looking into deeper into this and fixing the `include`/`exclude` behaviour to be consistent would be reasonable?

rmitsch:
Removed `include`. Updated website docs.

spacy/__init__.py
```diff
@@ -42,14 +43,21 @@ def load(
     disable (Iterable[str]): Names of pipeline components to disable. Disabled
         pipes will be loaded but they won't be run unless you explicitly
         enable them by calling nlp.enable_pipe.
+        enable (Iterable[str]): Names of pipeline components to enable. All other
+            pipes will be disabled (and can be enabled using `nlp.enable_pipe`).
```
adrianeboyd, line 47:
```suggestion
    enable (Iterable[str]): Names of pipeline components to enable. All other
        pipes will be disabled (but can be enabled later using nlp.enable_pipe).
```

spacy/language.py
```diff
@@ -2004,6 +2053,7 @@ def from_disk(
 
         path (str / Path): A path to a directory.
         exclude (Iterable[str]): Names of components or serialization fields to exclude.
+
```
adrianeboyd, line 2056:
```suggestion
```

spacy/errors.py
```diff
@@ -919,6 +919,8 @@ class Errors(metaclass=ErrorsWithCodes):
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
     E1037 = ("Invalid attribute value '{attr}'.")
+    E1038 = ("Function was called with `{arg1}`={arg1_values} and `{arg2}`={arg2_values} but that information is "
+             "conflicting.")
```
adrianeboyd, line 923: Still not 100% happy with the phrasing, but:

```suggestion
    E1038 = ("Function was called with `{arg1}`={arg1_values} and "
             "`{arg2}`={arg2_values} but these arguments are conflicting.")
```

spacy/language.py
```diff
@@ -1991,6 +2002,44 @@ def to_disk(
...
+        disabled components.
+        `enable` takes precedence over `disable` - so if the former is set, all other components are assumed to be
+        disabled, independently of the value of `disable`. If `enable` is empty, only those component whose names are
+        included in `disable` are assumed to be disabled.
+
+        disable (Iterable[str]): Names of components or serialization fields to disable.
+        enable (Iterable[str]): Names of pipeline components to enable.
+        pipe_names (Iterable[str]): Names of all pipeline components.
+
+        RETURNS (Tuple[str, ...]): Names of components to exclude from pipeline w.r.t. specified includes and excludes.
```
adrianeboyd, line 2020: Can you wrap the docstrings at 80 chars and start immediately after the initial `"""`?

website/docs/usage/processing-pipelines.md
```diff
@@ -362,6 +362,18 @@ nlp = spacy.load("en_core_web_sm", disable=["tagger", "parser"])
...
+In addition to `disable`, `spacy.load()` also accepts `enable`. If `enable` is
+set, all components except for those in `enable` are disabled.
+
+```python
+# Load the complete pipeline, but disable all components except the tagger
+nlp = spacy.load("en_core_web_sm", enable=["tagger"])
+# Has the same effect, as the parser is already not part of enabled set of components
+nlp = spacy.load("en_core_web_sm", enable=["tagger"], disable=["parser"])
+# Will raise an error, as the sets of enabled and disabled components are conflicting
+nlp = spacy.load("en_core_web_sm", enable=["tagger"], disable=["tagger"])
```
adrianeboyd, line 374: These examples are going to be confusing with the `tok2vec` situation, but maybe it would better to use it so it's clear. Maybe use the pairs `["tok2vec", "tagger"]` + `["ner"]` for the example so it's one with `tok2vec` and one without?

rmitsch: Not sure what you mean. Like this?
```python
nlp = spacy.load("en_core_web_sm", enable=["tok2vec", "tagger"])
nlp = spacy.load("en_core_web_sm", enable=["tok2vec", "tagger"], disable=["ner"])
nlp = spacy.load("en_core_web_sm", enable=["tok2vec", "tagger"], disable=["tagger"])
```

adrianeboyd: Yes, or `["ner"]` / `["ner"]` if that's clearer in the last example.

polm:
I'm kind of late here, but I agree this would be a great feature to have - for short bits of code this can be clearer and more succint than disable.

spacy/language.py
```diff
@@ -1991,6 +2002,45 @@ def to_disk(
...
+    @staticmethod
+    def _resolve_component_status(
+        disable: Iterable[str], enable: Iterable[str], pipe_names: Collection[str]
+    ) -> Tuple[str, ...]:
+        """Derives whether (1) `disable` and `enable` values are consistent and (2)
+        resolves those to a single set of disabled components. `enable` takes precedence
+        over `disable` - so if the former is set, all other components are assumed to be
+        disabled, independently of the value of `disable`. If `enable` is empty, only
+        those component whose names are included in `disable` are assumed to be
+        disabled.
```
svlandeg, line 2014: This doc string is out-of-date with the current functionality that checks compatibility between the `disable` and `enable`

svlandeg (commented):
Looks good to me! Just had a few more minor comments but should otherwise be good to merge :-)

svlandeg:
Nice work, Raphael!

PR #10786, opened 2022-05-12
Fix issues for Mypy 0.950 and Pydantic 1.9.0

richardpaulhudson:
This spaCy PR depends on https://github.com/explosion/thinc/pull/599, so that this spaCy PR will not turn green until a version of Thinc is released that incorporates that Thinc PR.

### Types of change
Mypy / Pydantic fixes

## Checklist


spacy/util.py
```diff
@@ -294,7 +294,7 @@ def find_matching_language(lang: str) -> Optional[str]:
 
     # Find out which language modules we have
     possible_languages = []
-    for modinfo in pkgutil.iter_modules(spacy.lang.__path__):  # type: ignore
+    for modinfo in pkgutil.iter_modules(spacy.lang.__path__): # type: ignore[attr-defined]
```
adrianeboyd, line 297: Extremely picky detail: but what's reformatting the whitespace in cases like this? Can you check the formatting with the current version of black?

richardpaulhudson: Thanks, you're right, somehow the version checked in in the PR seems not to have had the correct formatting. It should be fixed now.

realabja:
Do you think this, will be available in the next release?

svlandeg:
> Do you think this, will be available in the next release?

Hi! We're planning to finish & release this for Thinc 8.1 and spaCy 3.4 🎉

spacy/errors.py
```diff
@@ -1,3 +1,4 @@
+from typing import Literal
```
svlandeg, line 1: `Literal` needs to be imported from `spacy.compat`

setup.cfg
```diff
@@ -57,7 +57,7 @@ install_requires =
     tqdm>=4.38.0,<5.0.0
     numpy>=1.15.0
     requests>=2.13.0,<3.0.0
-    pydantic>=1.7.4,!=1.8,!=1.8.1,<1.9.0
+    pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0
```
svlandeg, line 60: @adrianeboyd: do you remember why we have these specified both in `thinc` as well as in `spacy`?

svlandeg (commented):
It's encouraging to see that we can remove so many ignores & casts, nice work Richard!

setup.cfg
adrianeboyd: Both `thinc` and `spacy` include `pydantic` imports.

svlandeg: Yes, but why? Why doesn't `spacy` just rely on `Thinc` pulling in `pydantic`?

svlandeg: I can see an argument in favour of this technique to have slightly more independent release cycles between the two repo's, but in general it does feel like we're just creating overhead / duplicate work for ourselves :|

spacy/ml/models/textcat.py
```diff
@@ -59,7 +59,7 @@ def build_simple_cnn_text_classifier(
             resizable_layer=resizable_layer,
         )
     model.set_ref("tok2vec", tok2vec)
-    model.set_dim("nO", nO)  # type: ignore  # TODO: remove type ignore once Thinc has been updated
+    model.set_dim("nO", cast(int, nO))
```
danieldk, line 62: Could you explain this cast? `nO` could still be `None` right? If so, I'd expect this to be:

```suggestion
    if nO is not None:
        model.set_dim("nO", cast(int, nO))
```

Also, I am not sure I understand why this `set_dim` is necessary. Does the chaining fail to set the value incorrectly when `nO` is provided?

spacy/ml/models/textcat.py
```diff
@@ -85,15 +85,15 @@ def build_bow_text_classifier(
...
+        resizable_layer: Model[Floats2d, Floats2d] = resizable(
             sparse_linear,
             resize_layer=partial(resize_linear_weighted, fill_defaults=fill_defaults),
         )
         model = extract_ngrams(ngram_size, attr=ORTH) >> resizable_layer
         model = with_cpu(model, model.ops)
         if output_layer:
             model = model >> with_cpu(output_layer, output_layer.ops)
-    model.set_dim("nO", nO)  # type: ignore[arg-type]
+    model.set_dim("nO", cast(int, nO))
```
danieldk, line 96: Same comment/question as above.

spacy/ml/models/textcat.py
```diff
@@ -130,7 +130,7 @@ def build_text_classifier_v2(
         model = (linear_model | cnn_model) >> output_layer
         model.set_ref("tok2vec", tok2vec)
     if model.has_dim("nO") is not False:
-        model.set_dim("nO", nO)  # type: ignore[arg-type]
+        model.set_dim("nO", cast(int, nO))
```
danieldk, line 133: See other comment about `nO`.

danieldk (commented):
This looks like a really nice improvement! I added a question about casting in some `set_dim` calls.

spacy/ml/models/textcat.py
danieldk: Fair enough! We can leave the `set_dim`s here.

spacy/ml/models/textcat.py
richardpaulhudson: I've made a corresponding change in the branch, but for some reason it isn't being picked up here.

danieldk: Some caching invalidation issue in GitHub? I'll resolve this :).

danieldk (approved):

spacy/util.py
```diff
@@ -1671,11 +1672,10 @@ def to_ternary_int(val) -> int:
 # See licenses/3rd_party_licenses.txt
 def packages_distributions() -> Dict[str, List[str]]:
     """Return a mapping of top-level packages to their distributions. We're
-    inlining this helper from the importlib_metadata "backport" here, since
     it's not available in the builtin importlib.metadata.
```
svlandeg, line 1675: Looks like you removed this line by accident?

svlandeg (commented):
This looks great Richard - nice work!

spacy/util.py
svlandeg: Haha don't worry about it, accidents happen and that what reviews are for! I would not have had a single comment otherwise, so at least now I feel useful ;-)

Issue #10727, opened 2022-04-28
Init Config from Python: typer.Option default value not working


janvdp:
## Goal
Create the config.cfg from Python and not from the command line.

From the source:
https://github.com/explosion/spaCy/blob/master/spacy/cli/init_config.py#L28

### Preparation
```python
from pathlib import Path
from spacy.cli.init_config import init_config_cli, Optimizations

my_config = Path("./my_config.cfg")
```

### This should work, ...
```Python
# This code 
init_config_cli(output_file=my_config, lang=lang, pipeline="ner",optimize=Optimizations.efficiency) 
```

### ..., however:
```Python

⚠ To generate a more effective transformer-based config (GPU-only),
install the spacy-transformers package and re-run this command. The config
generated now does not use transformers.
---------------------------------------------------------------------------
UndefinedError                            Traceback (most recent call last)
/tmp/ipykernel_664/2418944264.py in <module>
      3 
      4 #init_config_cli(output_file=my_config, lang=lang, pipeline="ner",optimize=Optimizations.efficiency, gpu=False, pretraining=False)
----> 5 init_config_cli(output_file=my_config, lang=lang, pipeline="ner",optimize=Optimizations.efficiency) #, gpu=False, pretraining=False)
      6 
      7 

/opt/conda/lib/python3.8/site-packages/spacy/cli/init_config.py in init_config_cli(output_file, lang, pipeline, optimize, gpu, pretraining, force_overwrite)
     60             exits=1,
     61         )
---> 62     config = init_config(
     63         lang=lang,
     64         pipeline=pipeline,

/opt/conda/lib/python3.8/site-packages/spacy/cli/init_config.py in init_config(lang, pipeline, optimize, gpu, pretraining, silent)
    171         )
    172         variables["transformer_data"] = None
--> 173     base_template = template.render(variables).strip()
    174     # Giving up on getting the newlines right in jinja for now
    175     base_template = re.sub(r"\n\n\n+", "\n\n", base_template)

/opt/conda/lib/python3.8/site-packages/jinja2/environment.py in render(self, *args, **kwargs)
   1289             return concat(self.root_render_func(ctx))  # type: ignore
   1290         except Exception:
-> 1291             self.environment.handle_exception()
   1292 
   1293     async def render_async(self, *args: t.Any, **kwargs: t.Any) -> str:

/opt/conda/lib/python3.8/site-packages/jinja2/environment.py in handle_exception(self, source)
[...]

PR #10788, opened 2022-05-12
Workaround for Typer optional default values with Python calls

rmitsch:
## Goals

Allow calling `init_config_cfg()` from Python with optional arguments not being set. This currently fails due to the behaviour of [Typer](https://github.com/tiangolo/typer) for handling optional arguments in Python calls of functions decorated with `@init_cli.command`. See https://github.com/explosion/spaCy/issues/10727 for a more in-depth description.

## Description

Modified `init_config_cli()` to force default values for optional arguments if they are still of types `typer.models.OptionInfo` or `typer.models.ArgumentInfo`. This seems to be done by Typer automatically on CLI calls, but not on Python calls (see also [the corresponding discussion in the Typer repo](https://github.com/tiangolo/typer/issues/106)).

If such a type is detected, the corresponding variable is set to the configured primitive default value.

### Types of change

Bug fix/enhancement, new test.

## Checklist


svlandeg (commented):
Thanks for digging into this! I might have a look at the `typer` side of things later this week, too, but either way it's a good idea to make sure our code base is robust against this type of failure.

rmitsch:
Comments/suggestions have been addressed.

svlandeg (commented):
Looking into this in a bit more detail, I'm not sure this is the path we want to take. For one, it makes the implementation of the `_cli` function quite ugly and error-prone. Eg. having to think about using `optional_args` instead of the parameters directly -> I could see this leading to errors in the future.

I've started wondering about the original issue as well - I'm not convinced that functions with `_cli` in their name **should** work from a Python call. We definitely don't advertise this API anywhere.

However, the problem remains that even if users call into `init_config` and `save_config` directly, as you'd probably want to recommend them to, then they don't get access to the default values that you'd get from the CLI.

The current implementation does not repeat the default values in e.g. `init_config`, as this also avoids them going out-of-sync across functions. But what if we define some of these default variables as a global on top? Then both functions can access them, they'd always be in sync, and we'd have functions that you can call directly from within Python (though not the `_cli` ones)

rmitsch:
> Looking into this in a bit more detail...
- Ugliness and error proneness of `optional_args`: agreed. The original intention was to not do this for each argument manually. A possible alternative would be to wrap the arguments in a function forcing the default value where necessary:
  ```python
      ...
      config = init_config(
          lang=get_typer_option_value(lang),
         ...
      )
  ```
  It's not much better, but at least avoids a parallel data store for argument values.

- Support of `_cli` calls from Python: it's IMO totally legitimate not to support that. 
- Default values for `init_config()`, `save_config()`: agreed, we shouldn't repeat default values. I'm not a huge fan of setting the default values as globals, but to be fair, I don't have a better idea.

So tl;dr: I think your suggestion of making the default values global is the least bad way to go forward. Should we bundle them in an object so they don't clutter the module too much, like this?
```python
class InitDefaultValues:
    output_file = "-"
    pretraining = False
    ...
    
```

rmitsch:
Removed forcing of optional default values, added synchronized default values for `init_config_cli()` and `init_cli()`.

spacy/cli/init_config.py
```diff
@@ -44,6 +59,7 @@ def init_config_cli(
 
     DOCS: https://spacy.io/api/cli#init-config
     """
+
```
adrianeboyd, line 62: I don't think these newlines were intentional or added by black?

spacy/cli/init_config.py
```diff
@@ -1,6 +1,7 @@
 from typing import Optional, List, Tuple
 from enum import Enum
 from pathlib import Path
+
```
svlandeg, line 4:
```suggestion
```

spacy/cli/init_config.py
```diff
@@ -52,6 +67,7 @@ def init_config_cli(
             "The provided output file already exists. To force overwriting the config file, set the --force or -F flag.",
             exits=1,
         )
+
```
svlandeg, line 70:
```suggestion
```

spacy/tests/test_cli.py
```diff
@@ -750,6 +751,28 @@ def test_debug_data_compile_gold():
     assert data["boundary_cross_ents"] == 1
 
 
+@pytest.mark.issue(10727)
+def test_init_config_from_python_without_optional_args():
```
svlandeg, line 755: I'm not sure we need this test? It ensures that an `AttributeError` is thrown when `init_config_cli` is called, but this isn't really the behaviour we're targeting, it's more like an unfortunate side effect we can work around by calling the other internal methods that now do have defaults? So maybe we should check the latter instead?

svlandeg (commented):
Looking good! Removed some of the introduced newlines and just had one more question on the test.

spacy/tests/test_cli.py
rmitsch: True, that's a leftover. What would we check for `init_config()` though? Seems to me like the final change of adding default params is so trivial that we can forego the new test completely and rely on the existing ones.

spacy/tests/test_cli.py
```diff
@@ -23,6 +23,7 @@
 from spacy.cli.debug_data import _get_spans_length_freq_dist
 from spacy.cli.download import get_compatibility, get_version
 from spacy.cli.init_config import RECOMMENDATIONS, init_config, fill_config
+from spacy.cli.init_config import init_config_cli, Optimizations
```
svlandeg, line 26:
```suggestion
```

spacy/cli/init_config.py
```diff
@@ -12,7 +12,7 @@
 from ..schemas import RecommendationSchema
```
svlandeg, line 12:
```suggestion
from ..schemas import RecommendationSchema
from ..util import SimpleFrozenList
```

spacy/cli/init_config.py
```diff
@@ -12,7 +12,7 @@
...
 from ._util import init_cli, Arg, Opt, show_validation_error, COMMAND
 from ._util import string_to_list, import_code
```
svlandeg, line 14:
```suggestion
from ._util import string_to_list, import_code

```

spacy/cli/init_config.py
```diff
@@ -10,9 +10,11 @@
 from .. import util
 from ..language import DEFAULT_CONFIG_PRETRAIN_PATH
 from ..schemas import RecommendationSchema
+from ..util import SimpleFrozenList
 from ._util import init_cli, Arg, Opt, show_validation_error, COMMAND
 from ._util import string_to_list, import_code
 
+from ..util import SimpleFrozenList
```
svlandeg, line 17:
```suggestion
```

svlandeg (approved):
Should be good to merge once the CI goes green!

rmitsch:
@svlandeg I think we can merge this?

PR #10789, opened 2022-05-12
Vectorize `example.get_aligned_parses`

shadeMe:
## Description

Previously, we were aligning heads and dependencies between candidate and gold tokens by manually iterating over elements in the example/alignment array. This wasn't optimal as it didn't leverage `numpy`'s vectorization. This PR rewrites the alignment code with `numpy` ops, reducing parser loss calculation time by ~45-50 ms per mini-batch.

### Before vectorization
parse-align-no-vec

### After vectorization
parse-align-vec

### Types of change
Performance enhancement

## Checklist

### DIFF
--- a/spacy/training/example.pyx
+++ b/spacy/training/example.pyx
@@ -1,5 +1,5 @@
 from collections.abc import Iterable as IterableInstance
-import numpy as np
+import numpy
 from murmurhash.mrmr cimport hash64
 
 from ..tokens.doc cimport Doc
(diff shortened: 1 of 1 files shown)


spacy/training/example.pyx
```diff
@@ -176,24 +176,47 @@ cdef class Example:
...
+
+        # Fetch indices of all gold heads for the aligned gold tokens.
+        heads = np.asarray(heads, dtype='i')
+        gold_head_i = heads[gold_i]
+
+        # Select all gold tokens that are heads of the previously selected 
+        # gold tokens (and are aligned to a single candidate token).
+        g2c_len_heads = gold_to_cand.lengths[gold_head_i]
+        g2c_len_heads = np.where(g2c_len_heads == 1)[0]
+        g2c_i = np.vectorize(lambda x: gold_to_cand[int(x)][0])(gold_head_i[g2c_len_heads]).squeeze()
```
shadeMe, line 209: Is there a better/more performant method of performing a vectorized gather op here?

spacy/training/example.pyx
```diff
@@ -1,5 +1,5 @@
 from collections.abc import Iterable as IterableInstance
-import numpy
+import numpy as np
```
adrianeboyd, line 2: Let's not change this. The codebase consistently uses:

- `cimport numpy as np`
- `import numpy`

spacy/training/example.pyx
```diff
@@ -176,24 +176,47 @@ cdef class Example:
...
-                    aligned_deps[cand_i] = deps[gold_i]
-        return aligned_heads, aligned_deps
+
+        # Select all candidate tokens that are aligned to a single gold token.
+        c2g_single_toks = numpy.where(cand_to_gold.lengths == 1)[0]
+
+        # Fetch all aligned gold token incides.
+        if c2g_single_toks.shape == cand_to_gold.lengths.shape:
+            # This the most likely case.
+            gold_i = cand_to_gold[0:cand_to_gold.lengths.shape[0]].squeeze()
```
danieldk, line 197: Just for understanding: if `c2g_single_toks.shape == cand_to_gold.lengths.shape`, then every gold token is aligned to a predicted token (all the lengths are 1). Wouldn't that also mean that `cand_to_gold` had the same length? Is

`cand_to_gold[0:cand_to_gold.lengths.shape[0]]`

the same as

`cand_to_gold[:]`

?

danieldk (commented):
Looks correct to me, nice work! Besides passing tests, did you train/evaluate some models to check that there are no regressions?

spacy/training/example.pyx
shadeMe: Ah, correct. I can in fact elide the start and end indices in that slice.

shadeMe:
Yeah, training and evaluating with the above changes resulted in no regressions in UAS/LAS accuracy on my end.

PR #10793, opened 2022-05-12
Update spaCy Universe: spacytextblob (code example)

Schero1994:
## Description
Updating the code example of `spacytextblob` from the spaCy Universe.
The additions are from the [stackoverflow ](https://stackoverflow.com/questions/70720126/oserror-e053-could-not-read-config-cfg-spacy-on-colab) thread, which suggested a workaround to solve an `OSError` with the `config.cfg` file. 
The change includes additional installations to prevent the `OSError` from occurring.

### Types of change
updating `universe.json`

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -298,6 +298,10 @@
             "github": "SamEdwardes/spacytextblob",
             "pip": "spacytextblob",
             "code_example": [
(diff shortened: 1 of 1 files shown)


polm (approved):
Looks good, thanks for improving that!

PR #10798, opened 2022-05-13
Tagger/morphologizer alignment performance optimizations

shadeMe:
## Description

* The offset calculation in `AlignmentArray` was incurring a lot of overhead as it was using single element index/add ops to populate the underlying `numpy` array, leading to a lot of interop chatter that invoked short-living array allocations, etc. Using a native Python list as an intermediate buffer removes this overhead, cutting down execution time by half.

* In `example.get_aligned`, indexing into the attribute array of aligned gold tokens yielded `numpy` arrays of unit length and `numpy` integer types (representing the attribute hashes). This was being passed to `StringStore` to retrieve the string value corresponding to the hash. 

  In the `StringStore.__getitem__` method, there exists a fast path where a LUT is used to check for/return special symbol. This was done by comparing the incoming input against the length of the ordered LUT. Under normal circumstances, this would be a high-throughput op, but since the values passed to this method were not native Python types, the generated C++ code ended up invoking the `numpy`-specific override for the internal `RichCompare` `PyObject` virtual function. This incurred a fair amount of overhead, which was further exacerbated by the calls not being vectorized.

  This overhead was eliminated by explicitly unwrapping unit-length arrays and coercing `numpy` integer types to native Python types before passing them to `StringStore`.

* Finally, the most likely case of alignment w.r.t the `TAG`, `POS`, and `DEP` resulted in candidate tokens forming a 1-to-1 correspondence with the gold tokens. This allowed for the implementation of a vectorized fast path.

### Before
tag-morph-before

### After 
tag-morph-after
 
(The above figures do not include the parser loss speed-up from #10789) 

### Types of change
Performance enhancement

## Checklist

### DIFF
--- a/spacy/training/alignment_array.pyx
+++ b/spacy/training/alignment_array.pyx
@@ -6,24 +6,26 @@ import numpy
 cdef class AlignmentArray:
     """AlignmentArray is similar to Thinc's Ragged with two simplfications:
     indexing returns numpy arrays and this type can only be used for CPU arrays.
-    However, these changes make AlginmentArray more efficient for indexing in a
+    However, these changes make AlignmentArray more efficient for indexing in a
     tight loop."""
 
     __slots__ = []
 
     def __init__(self, alignment: List[List[int]]):
-        self._lengths = None
-        self._starts_ends = numpy.zeros(len(alignment) + 1, dtype="i")
-
         cdef int data_len = 0
         cdef int outer_len
         cdef int idx
+
+        starts_ends = [0] * (len(alignment) + 1)
         for idx, outer in enumerate(alignment):
             outer_len = len(outer)
-            self._starts_ends[idx + 1] = self._starts_ends[idx] + outer_len
+            starts_ends[idx + 1] = starts_ends[idx] + outer_len
             data_len += outer_len
 
+        self._lengths = None
+        self._starts_ends = numpy.asarray(starts_ends, dtype='i')
         self._data = numpy.empty(data_len, dtype="i")
+
         idx = 0
         for outer in alignment:
             for inner in outer:
--- a/spacy/training/example.pyx
+++ b/spacy/training/example.pyx
@@ -153,24 +153,74 @@ cdef class Example:
     def get_aligned(self, field, as_string=False):
         """Return an aligned array for a token attribute."""
         align = self.alignment.x2y
+        gold_values = self.reference.to_array([field])
+
+        def _vectorized():
+            # Fast path for Doc attributes/fields that are predominantly a single value,
+            # i.e., TAG, POS, MORPH.
+            x2y_single_toks = []
+            x2y_single_toks_i = []
+
+            x2y_multiple_toks = []
+            x2y_multiple_toks_i = []
+
+            # Gather indices of gold tokens aligned to the candidate tokens into two buckets.
+            #   Bucket 1: All tokens that have a one-to-one alignment.
+            #   Bucket 2: All tokens that have a one-to-many alignment.
+            for idx, token in enumerate(self.predicted):
+                aligned_gold_i = align[token.i]
+                if len(aligned_gold_i) == 1:
+                    x2y_single_toks.append(aligned_gold_i.item())
+                    x2y_single_toks_i.append(idx)
+                else:
+                    x2y_multiple_toks.append(aligned_gold_i)
+                    x2y_multiple_toks_i.append(idx)
+
+            # Map elements of the first bucket directly to the output array.
+            output = numpy.full(len(self.predicted), None)
+            output[x2y_single_toks_i] = gold_values[x2y_single_toks].squeeze()
+
+            # Collapse many-to-one alignments into one-to-one alignments if they
+            # share the same value. Map to None in all other cases.
+            for i in range(len(x2y_multiple_toks)):
+                aligned_gold_values = gold_values[x2y_multiple_toks[i]]
+
+                # If all aligned tokens have the same value, use it.
+                x2y_multiple_toks[i] = aligned_gold_values[0].item() \
+                    if len(set(list(aligned_gold_values))) == 1 else None
+
+            output[x2y_multiple_toks_i] = x2y_multiple_toks
+
+            return output.tolist()
+
+        def _non_vectorized():
+            # Slower path for fields that return multiple values (resulting
+            # in ragged arrays that cannot be vectorized trivially).
+            output = [None] * len(self.predicted)
+
+            for token in self.predicted:
+                aligned_gold_i = align[token.i]
+                values_org = gold_values[aligned_gold_i]
+                values = values_org.ravel()
+                if len(values) == 0:
+                    output[token.i] = None
+                elif len(values) == 1:
+                    output[token.i] = values.item()
+                elif len(set(list(values))) == 1:
+                    # If all aligned tokens have the same value, use it.
+                    output[token.i] = values[0].item()
+                else:
+                    output[token.i] = None
+
+            return output
+
 
         vocab = self.reference.vocab
-        gold_values = self.reference.to_array([field])
-        output = [None] * len(self.predicted)
-        for token in self.predicted:
-            values = gold_values[align[token.i]]
-            values = values.ravel()
-            if len(values) == 0:
-                output[token.i] = None
-            elif len(values) == 1:
-                output[token.i] = values.item()
-            elif len(set(list(values))) == 1:
-                # If all aligned tokens have the same value, use it.
-                output[token.i] = values[0].item()
-            else:
-                output[token.i] = None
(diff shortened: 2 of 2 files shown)


spacy/training/example.pyx
```diff
@@ -153,24 +153,74 @@ cdef class Example:
...
-        for token in self.predicted:
-            values = gold_values[align[token.i]]
-            values = values.ravel()
-            if len(values) == 0:
-                output[token.i] = None
-            elif len(values) == 1:
-                output[token.i] = values[0]
-            elif len(set(list(values))) == 1:
+
+        def _vectorized():
```
svlandeg, line 158: I'd personally prefer making these into separate private functions, as that helps with readability.

spacy/training/example.pyx
```diff
@@ -150,27 +150,82 @@ cdef class Example:
...
-            values = gold_values[align[token.i]]
-            values = values.ravel()
+            aligned_gold_i = align[token.i]
+            values_org = gold_values[aligned_gold_i]
+            values = values_org.ravel()
             if len(values) == 0:
                 output[token.i] = None
             elif len(values) == 1:
-                output[token.i] = values[0]
+                output[token.i] = values.item()
```
svlandeg, line 205:
```suggestion
                output[token.i] = values[0].item()
```

spacy/training/example.pyx
```diff
@@ -150,27 +150,82 @@ cdef class Example:
...
+        x2y_multiple_toks = []
+        x2y_multiple_toks_i = []
+
+        # Gather indices of gold tokens aligned to the candidate tokens into two buckets.
+        #   Bucket 1: All tokens that have a one-to-one alignment.
+        #   Bucket 2: All tokens that have a one-to-many alignment.
+        for idx, token in enumerate(self.predicted):
+            aligned_gold_i = align[token.i]
+            if len(aligned_gold_i) == 1:
+                x2y_single_toks.append(aligned_gold_i.item())
```
svlandeg, line 169:
```suggestion
                x2y_single_toks.append(aligned_gold_i[0].item())
```

spacy/training/example.pyx
```diff
@@ -150,27 +150,82 @@ cdef class Example:
...
+        output[x2y_single_toks_i] = gold_values[x2y_single_toks].squeeze()
+
+        # Collapse many-to-one alignments into one-to-one alignments if they
+        # share the same value. Map to None in all other cases.
+        for i in range(len(x2y_multiple_toks)):
+            aligned_gold_values = gold_values[x2y_multiple_toks[i]]
+
+            # If all aligned tokens have the same value, use it.
+            x2y_multiple_toks[i] = aligned_gold_values[0].item() \
+                if len(set(list(aligned_gold_values))) == 1 else None
```
svlandeg, line 186: If the `if`-`else` gets split up across separate lines anyway, let's just use the more verbose structure with separate `if` and `else` statements.

svlandeg (commented):
Looks good to me on first view, with some smaller nitpicks below.

Have you double checked the coverage of the test suite for this functionality?

spacy/training/example.pyx
shadeMe: The removal of the subscript operator was an intentional change in that line as it not necessary: Calling `item()` on unit-length `numpy` arrays will essentially call the method on the first element of the array with the extra overhead incurred by the subscript operation.

spacy/training/example.pyx
shadeMe: Same as above.

shadeMe:
Re. test coverage - Yeah, the new functionality is covered by the test suite.

spacy/training/alignment_array.pyx
```diff
@@ -6,24 +6,26 @@ import numpy
...
-
         cdef int data_len = 0
         cdef int outer_len
         cdef int idx
+
+        starts_ends = [0] * (len(alignment) + 1)
         for idx, outer in enumerate(alignment):
             outer_len = len(outer)
-            self._starts_ends[idx + 1] = self._starts_ends[idx] + outer_len
+            starts_ends[idx + 1] = starts_ends[idx] + outer_len
```
danieldk, line 22: Sometimes in these cases (since we are in Cython-land), it can also help to treat the numpy array as a raw C array. Since we know the length, it's safe and it avoids all the overhead of NumPy array indexing, but is probably faster than first constructing a Python list and then converting that.

spacy/training/example.pyx
```diff
@@ -150,27 +150,84 @@ cdef class Example:
...
+        output = numpy.full(len(self.predicted), None)
+        output[x2y_single_toks_i] = gold_values[x2y_single_toks].squeeze()
+
+        # Collapse many-to-one alignments into one-to-one alignments if they
+        # share the same value. Map to None in all other cases.
+        for i in range(len(x2y_multiple_toks)):
+            aligned_gold_values = gold_values[x2y_multiple_toks[i]]
+
+            # If all aligned tokens have the same value, use it.
+            if len(set(list(aligned_gold_values))) == 1:
```
danieldk, line 185: I wonder if `groupby` isn't faster, since it doesn't involve intermediate set construction. See the `all_equal` recipe in

https://docs.python.org/3/library/itertools.html

danieldk, line 185: Forgot to add that it if it is not faster, I think you could simplify this to:

```suggestion
            if len(set(aligned_gold_values)) == 1:
```

since the `set` constructor takes an iterable.

spacy/training/example.pyx
```diff
@@ -150,27 +150,84 @@ cdef class Example:
...
+        # Slower path for fields that return multiple values (resulting
+        # in ragged arrays that cannot be vectorized trivially).
         output = [None] * len(self.predicted)
+
         for token in self.predicted:
-            values = gold_values[align[token.i]]
-            values = values.ravel()
+            aligned_gold_i = align[token.i]
+            values_org = gold_values[aligned_gold_i]
+            values = values_org.ravel()
```
danieldk, line 203: I think I liked the original more

```suggestion
            values = gold_values[aligned_gold_i].revel()
```

Because then the intermediate value can be removed before the end of the scope. If the Python interpreter is smart enough, it could infer this because `values_org` is not referenced after this point, but I wouldn't be too confident about this.

spacy/training/example.pyx
```diff
@@ -150,27 +150,84 @@ cdef class Example:
...
             if len(values) == 0:
                 output[token.i] = None
             elif len(values) == 1:
-                output[token.i] = values[0]
+                output[token.i] = values.item()
             elif len(set(list(values))) == 1:
```
danieldk, line 208: See above :).

danieldk (commented):
Looks good to me, some small nitpicks.

spacy/training/alignment_array.pyx
shadeMe: Great point! I find it's very easy to forget that C arrays are an option too :)

spacy/training/example.pyx
shadeMe: Oh, I think I renamed that variable to print it out in whilst debugging and then forgot to revert that change.

spacy/training/example.pyx
shadeMe: As in "directly derefrence the `data()` pointer and explicitly cast to `int`?

danieldk: No, no need for construction `list` first, since `set` takes an iterable. (Or possibly `groupby`.)

shadeMe: Ah, right. (Why can't GitHub just highlight the line to which a review comment is attached...)

shadeMe:
Did a quick profile: It's a wash - the newly-introduced changes do not appear to noticeably improve performance, nor do they cause any regressions (w.r.t the original changes).

shadeMe:

shadeMe:

spacy/training/alignment_array.pyx
```diff
@@ -6,28 +6,33 @@ import numpy
...
     def __init__(self, alignment: List[List[int]]):
-        self._lengths = None
-        self._starts_ends = numpy.zeros(len(alignment) + 1, dtype="i")
-
         cdef int data_len = 0
         cdef int outer_len
         cdef int idx
+
+        self._starts_ends = numpy.zeros(len(alignment) + 1, dtype='int32')
+        cdef int* starts_ends_ptr = <int*>self._starts_ends.data
```
danieldk, line 20: I know that `int` is pretty much fixed, but it always makes me a bit nervous. Maybe `int32_t`? Cython provides [definitions](https://github.com/cython/cython/blob/master/Cython/Includes/libc/stdint.pxd) for these as well.

shadeMe:
> _⏳ Triggering 'TestSuite.GPU' for 'spaCy' on Buildkite..._

PR #10826, opened 2022-05-20
Add note about multiple patterns

polm:
## Description

If you have multiple patterns associated with a single label in the Matcher, there's no way to tell which of the patterns was responsible for the match directly. This is why the EntityRuler internally stores Entity IDs as different labels, for example. 

This change makes this limit explicit in the documentation. 

This point was raised in #10818.

### Types of change

doc enhancement

## Checklist


website/docs/api/matcher.md
```diff
@@ -113,6 +113,12 @@ string where an integer is expected) or unexpected property names.
 
 Find all token sequences matching the supplied patterns on the `Doc` or `Span`.
 
+Note that if a single label has multiple patterns associated with it, the
+returned matches don't provide a way to tell which pattern was responsible for
+the match. If that information is necessary, separate labels should be used for
+each pattern. For example, the [`EntityRuler`](/api/entityruler) uses a
+separator to store labels and entity IDs as the labels passed to the `Matcher`.
```
svlandeg, line 120: I'm not sure the additional information about `EntityRuler` helps or just further confuses things? People might not really know how everything is implemented internally and that's fine. The suggestion to use separate labels could be enough?

polm: I'd be fine with leaving this out, but I thought pointing out an example implementation would be helpful. I think without direction people will tend to do something like `LABEL1`, `LABEL2`, etc, when they could do `LABEL||1`, `LABEL||2` in a way that makes it easy to group things in post-processing.

svlandeg: Maybe we could give the example more directly then? I'm afraid that the additional explanation only parses/helps when you already know what it looks like :-)

polm: Got it. In that case it gets a little long for this place in the documentation maybe... Should I put it in a sidebar, or make an FAQ post in Discussion or something...?

svlandeg: Ah maybe an FAQ post would make sense! Then you can explain all it, including the `EntityRuler` stuff if you like ;-)

polm: OK, I'll write an FAQ post and link to it from here.

PR #10829, opened 2022-05-20
Add case-insensitive candidate generator to NEL

polm:
This is a first draft of what this might look like. The implementation is pretty simple but works.

A basic test is included.

This is functional, but currently it only works if you remember to
lowercase things when creating the KB. Probably the most efficient way
to do it but maybe there's a cleaner way.

Documentation has also not been added yet.

### Types of change

enhancement

## Checklist

### DIFF
--- a/spacy/kb.pyx
+++ b/spacy/kb.pyx
@@ -77,6 +77,15 @@ def get_candidates(KnowledgeBase kb, span) -> Iterator[Candidate]:
     return kb.get_alias_candidates(span.text)
 
 
+def get_candidates_lowercase(KnowledgeBase kb, span) -> Iterator[Candidate]:
+    """
+    Return candidate entities for a given span using the text of the span in lowercase.
+
+    This is the same as `get_candidates`, but is intended for case-insensitive retrieval.
+    """
+    return kb.get_alias_candidates(span.text.lower())
+
+
 cdef class KnowledgeBase:
     """A `KnowledgeBase` instance stores unique identifiers for entities and their textual aliases,
     to support entity linking of named entities to real-world concepts.
@@ -290,7 +299,7 @@ cdef class KnowledgeBase:
         """
         Return candidate entities for an alias. Each candidate defines the entity, the original alias,
         and the prior probability of that alias resolving to that entity.
-        If the alias is not known in the KB, and empty list is returned.
+        If the alias is not known in the KB, an empty list is returned.
         """
         cdef hash_t alias_hash = self.vocab.strings[alias]
         if not alias_hash in self._alias_index:
--- a/spacy/ml/models/entity_linker.py
+++ b/spacy/ml/models/entity_linker.py
@@ -98,3 +98,8 @@ def empty_kb(entity_vector_length: int) -> Callable[[Vocab], KnowledgeBase]:
 @registry.misc("spacy.CandidateGenerator.v1")
 def create_candidates() -> Callable[[KnowledgeBase, Span], Iterable[Candidate]]:
     return get_candidates
+
+
+@registry.misc("spacy.CaselessCandidateGenerator.v1")
+def create_candidates() -> Callable[[KnowledgeBase, Span], Iterable[Candidate]]:
+    return get_candidates_lowercase
--- a/spacy/tests/pipeline/test_entity_linker.py
+++ b/spacy/tests/pipeline/test_entity_linker.py
@@ -6,7 +6,7 @@ from numpy.testing import assert_equal
 from spacy import registry, util
 from spacy.attrs import ENT_KB_ID
 from spacy.compat import pickle
-from spacy.kb import Candidate, KnowledgeBase, get_candidates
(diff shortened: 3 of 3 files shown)


svlandeg:
> it only works if you remember to lowercase things when creating the KB

I don't think this solution is sufficient, because of this dependency on how the KB was created. Users of this functionality might (wrongfully) assume that it'll do case-independent matching between mention and KB, however either is cased. So I feel like implementing it only from one end (only lowercasing the mention) would be too confusing.

polm:
I'm also not happy with that requirement, but I'm not sure what the best alternative is. 

It's possible to make this work as-is, but doing so without altering the index or making any assumptions about it at all would require iterating over all possible case variations for the input. That doesn't seem reasonable.

The better thing to do is to either build the index in a particular way (case-insensitive), or to run a prep function on the index. 

If the index is built properly, that should be fine. But building the index is not handled in the config or by spaCy, it's up to the user.

Running a prep function, which would go over the index and add lowercase entries for any not already lowercase, is easy to do, but would result in extra entries (assuming we don't add a delete function), using extra memory for no benefit. We could rebuild the index instead, which would take a bit longer, but have no downsides after that. The problem is that there's no concept of a function like this anywhere at present, so we would either have to add a new option somewhere or do state-based trickery in the `get_candidates` function, which seems like a bad idea.

We could make the `get_candidates` function a callable object and have the component call `init()` on it or something.

Basically it seems like the strategy used by the `get_candidates` function is likely to depend pretty heavily on how the index was built, and even for what seems a simple change like ignoring case here, just changing the retrieval function in isolation won't work.

rmitsch:
Considering the discussed changes to the NEL workflow, I think it'd make sense to move the responsibility for building the index away from the user. If we e.g. really do candidate generation via NN search in embedding space, we'll need a particular kind of optimized indexing anyway to ensure reasonable performance.
Similar for external ontologies: we'll probably expect a specific format or even data source.

For now though I wouldn't worry too much about reasonable performance or getting the config right. I expect that the system architecture/config design will also be heavily influenced by the requirements of the approach we'll go with in the end (for example: maybe it turns out that using ontologies in combination with embeddings is rubbish, so we don't need to consider using both).

Hence my suggestion: let's build the benchmark first and jerry-rig whatever necessary to test our proposed alternative models. Once we've decided which one to put into production, we'll think about the modifications necessary to the config/sys. architecture. What's your take on this?

polm:
For the benchmark I definitely agree this isn't important and we can just do whatever.

For the general case I am open to the idea of taking index creation away from the user, but I think there are some things that make it not clear how that would work.

1. What if the user doesn't have a list of items, but a process that generates them? In the simple case, imagine they use the current KB and have some function that corrupts input to generate aliases. That means they can't just specify data as input.
2. Where does KB creation go in the config? `[training]` I guess? Do we need a function that runs at load (not just training) in case it checks for updates or something? I think we need to design this. 

If we had a concrete proposal for what that would look like it might make it easier to discuss.

Separately from those more general issues, for users right now who want to use a case-insensitive lookup, I think this PR is the best we can do without drastic changes. I guess we could add something to check that the KB was lowercase at startup time, but that would make the config more complicated and be slow.

rmitsch:
1. Stupid question, but is this something we need/want to support? What are the downsides to always requiring the input to be data?
2. Intuitively it would have placed it in `[corpora]`. I don't know the workflow well enough though to judge how much sense that makes from a internal point of view. 
A concrete proposal would definitely be helpful. I'll file an issue for that.

Regrading the case-insensitive lookup: is this something that has been requested frequently? If not, I'm not sure whether it makes sense to introduce a change here that might be replaced by a different approach to candidate generation altogether a month later.

polm:
> Stupid question, but is this something we need/want to support? What are the downsides to always requiring the input to be data?

That's something we have to decide. I don't think it's critical to support processes, but want to be aware of that decision.

> Regrading the case-insensitive lookup: is this something that has been requested frequently?

It's been requested a few times, which I guess makes it one of the more-requested features. I don't think there's any urgency to it, though I didn't expect the implemetation to have issues like this.

svlandeg:
Hi team,

it's clear that we need to hash out some config extensions to address this properly, and it's good that this PR has brought those issues to the surface.

Given the issues that were raised, I'd be reluctant to add this PR to spaCy now, as this new generator might set the wrong expectations for any user who doesn't realise the dependency with the index. As I said in my earlier comment - they might (understandably) expect that the mention is compared to entries in the KB in a fully case insensitive fashion, which isn't actually the case (pun intended).

With respect to users who have asked for this functionality, basically the solution from this PR is the same as what I wrote in the original [PR description](https://github.com/explosion/spaCy/pull/5876) when we made the candidate generation configurable: it's straightforward to implement a custom `Callable` candidate generator and plug that into your pipeline, as long as you also take accountability of creating the `kb` correctly to support lowercase matches.

In that sense, and in the philosophy of "Let them write code", I propose we close out this PR, and continue looking at the larger issue of index/KB/candidate_generator configurability & compatibility internally.

PR #10830, opened 2022-05-20
Better errors for has_annotation and Matcher

kadarakos:
More informative errors when wrong `Doc` attribute is provided

## Description
Noticed that when providing a wrong attribute to `Doc.has_annotation` the error does not get caught and the `None` returned by `intify_attr` propagates down to leading to a cryptic message:

```
~/spaCy/spacy/tokens/doc.pyx in genexpr()

TypeError: an integer is required

```

The `Matcher` object handles this user error, but it first assigns `None` with `attr = IDS.get(attr)` and then tells the user that `None` is an invalid input. This PR stores the `input_attr = attr` early on and then reports the error with the input rather than `None`.  

### Types of change
Small fix to the error messages.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -914,9 +914,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1034 = ("Node index {i} out of bounds ({length})")
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
-    E1037 = ("\"{attr}\" is not a valid value for attr.\n"
-             "Please consult https://spacy.io/api/token#attributes "
-             "for more information!")
+    E1037 = ("Invalid attribute value '{attr}'.")
 
 # Deprecated model shortcuts, only used in errors and warnings
 OLD_MODEL_SHORTCUTS = {
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -914,7 +914,9 @@ class Errors(metaclass=ErrorsWithCodes):
     E1034 = ("Node index {i} out of bounds ({length})")
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
-
+    E1037 = ("\"{attr}\" is not a valid value for attr.\n"
+             "Please consult https://spacy.io/api/token#attributes "
+             "for more information!")
 
 # Deprecated model shortcuts, only used in errors and warnings
 OLD_MODEL_SHORTCUTS = {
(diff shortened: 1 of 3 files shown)


spacy/tokens/doc.pyx
```diff
@@ -423,6 +424,12 @@ cdef class Doc:
         elif attr == "IS_SENT_END" or attr == self.vocab.strings["IS_SENT_END"]:
             attr = SENT_START
         attr = intify_attr(attr)
+        if attr is None:
+            raise ValueError(
+                f"\"{input_attr}\" is not a valid value for attr.\n"
+                "Please consult https://spacy.io/api/token#attributes "
+                "for more information!"
```
adrianeboyd, line 431: Can you move this to `spacy.errors`?

spacy/errors.py
```diff
@@ -914,7 +914,9 @@ class Errors(metaclass=ErrorsWithCodes):
     E1034 = ("Node index {i} out of bounds ({length})")
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
-
+    E1037 = ("\"{attr}\" is not a valid value for attr.\n"
+             "Please consult https://spacy.io/api/token#attributes "
+             "for more information!")
```
adrianeboyd, line 919: It is a problem that there is no good list of the attr values in the docs. We should probably add one, but in a separate PR. (We usually avoid "!" and also "please" for the most part.)

```suggestion
    E1037 = ("Invalid attribute value '{attr}'.")
```

spacy/errors.py
```diff
@@ -914,7 +914,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1034 = ("Node index {i} out of bounds ({length})")
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
-
+    E1037 = ("Invalid attribute value '{attr}'.")
```
adrianeboyd, line 917: Sorry, missed this the first time:

```suggestion
    E1037 = ("Invalid attribute value '{attr}'.")

```

adrianeboyd:
While we're here, can you update both E152 and E153 to add AttributeRuler to the list of things you can add `validate=True` to? It's a little tricky in terms of the phrasing `validate=True` because you're not instantiating it in the same way, but hopefully it's enough to point people in the right direction.

spacy/matcher/matcher.pyx
```diff
@@ -786,6 +786,7 @@ def _preprocess_pattern(token_specs, vocab, extensions_table, extra_predicates):
 def _get_attr_values(spec, string_store):
     attr_values = []
     for attr, value in spec.items():
+        input_attr = attr
```
svlandeg, line 789: I guess I would have called this `orig_attr` but I can see either way.

svlandeg (approved):
Thanks for the fixes, this is definitely more user-friendly :-)

PR #10836, opened 2022-05-23
Add API docs for token attribute symbols

adrianeboyd:
## Description

Add API docs for token attribute symbols.

### Types of change

Docs.

## Checklist

### DIFF
--- a/website/docs/api/attributes.md
+++ b/website/docs/api/attributes.md
@@ -48,33 +48,33 @@ e.g. `token.dep` while the string value is accessed by appending `_` as in
 A map from string attribute names to internal attribute IDs is stored in
 `spacy.attrs.IDS`.
 
-| Attribute     |  Description                                                                                                                                                  |
-| ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-|  `DEP`        | The token's dependency label. ~~str~~                                                                                                                         |
-| `ENT_ID`      | The token's entity ID (`ent_id`). ~~str~~                                                                                                                     |
-| `ENT_IOB`     | The IOB part of the token's entity tag. Uses custom integer vaues rather than the string store: unset is `0`, `I` is `1`, `O` is `2`, and `B` is `3`. ~~str~~ |
-| `ENT_KB_ID`   | The token's entity knowledge base ID. ~~str~~                                                                                                                 |
-| `ENT_TYPE`    | The token's entity label. ~~str~~                                                                                                                             |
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I'm not entirely sure where to put this. It's not really cython-specific, but it's also not a class like all the other non-cython API docs.

You could potentially additionally some cython-specific methods such as `Token.get_struct_attr`, but I'm not sure we want to add this, or if we want to add this in this section.

I also can't figure out what's going on with the prettier formatting, which indents some number of items in the table by one space, and I have no idea what's going on. I'd ignore it, except that it's visible in the rendered table.

website/docs/api/attributes.md
```diff
@@ -0,0 +1,80 @@
...
+and the internal integer symbols (`DEP`). The internal IDs can be imported from
+`spacy.attrs` or retrieved from the [`StringStore`](/api/stringstore).
+
+The corresponding [`Token` attributes](/api/token#attributes) can be accessed
+using the attribute name in lowercase, e.g. `token.orth` or `token.length`.
+
+String attributes are stored internally as integer hash values from the
+[`StringStore`](/api/stringstore). The integer value is accessed as `Token.attr`
+e.g. `token.dep` while the string value is accessed by appending `_` as in
+`token.dep_`.
```
svlandeg, line 46: There's something off with the flow of this paragraph to me, like it should be incorporated in the previous 2 paragraphs somehow? Mentioning the `StringStore` twice in slightly different context is a bit confusing, and the same is true for explaining how to access `token.attr` twice (kind of)

svlandeg (commented):
I think it's a good idea to document some of this more explicitly, though I worry a little about the overlap with `api/matcher#patterns` and `api/token#attributes`. 

It doesn't really help that we're calling both "Token attributes", one being the definition and the other being the actual attribute of an actual token :|

svlandeg:
> I also can't figure out what's going on with the prettier formatting, which indents some number of items in the table by one space, and I have no idea what's going on. I'd ignore it, except that it's visible in the rendered table.

There's a literal `NBSP` in the original table in the [Matcher doc](https://spacy.io/api/matcher) - you can see it rendered wrongly there too. I've got this open now anyway so I'll push the fix for this file and fix some others across the repo in a separate PR.

adrianeboyd:
One of the problem with phrasing this is avoiding saying "symbols" anywhere, since that has another specific meaning.

svlandeg (approved):
Anyway I think this is good to merge, Adriane?

adrianeboyd:
It's a start at least...

PR #10837, opened 2022-05-23
Handle Cyrillic combining diacritics

richardpaulhudson:
## Description
Some languages that are written in the Cyrillic alphabet permit the use of diacritic marks (́á à a̍) to mark the stressed syllable in words where the stress is distinctive. Unlike in the Latin alphabet, there are often no separate Unicode points for Cyrillic accented vowels; instead, the standard Cyrillic letter is followed by a letter from the Unicode block [Combining Diacritical Marks](https://unicode-table.com/en/blocks/combining-diacritical-marks/).

This PR adds support for these marks to the tokenizers for Russian, Ukrainian and Bulgarian. Because this functionality is potentially useful for any language written in Cyrillic with distinctive prosodic stress — including languages not yet supported — it has been implemented in such a way that the feature can easily be added to further languages without copying code.

### Types of change
Bugfix

## Checklist

### DIFF
--- a/spacy/lang/punctuation.py
+++ b/spacy/lang/punctuation.py
@@ -1,5 +1,5 @@
 from .char_classes import LIST_PUNCT, LIST_ELLIPSES, LIST_QUOTES, LIST_CURRENCY
-from .char_classes import LIST_ICONS, HYPHENS, CURRENCY, UNITS
+from .char_classes import LIST_ICONS, HYPHENS, CURRENCY, UNITS, COMBINING_DIACRITICS
 from .char_classes import CONCAT_QUOTES, ALPHA_LOWER, ALPHA_UPPER, ALPHA, PUNCT
 
 
@@ -39,8 +39,20 @@ TOKENIZER_INFIXES = (
         r"(?<=[{al}{q}])\.(?=[{au}{q}])".format(
             al=ALPHA_LOWER, au=ALPHA_UPPER, q=CONCAT_QUOTES
         ),
-        r"(?<=[{a}]),(?=[{a}])".format(a=ALPHA),
+        r"(?<=[{a}])(?:{p})(?=[{a}])".format(a=ALPHA, p=PUNCT),
         r"(?<=[{a}])(?:{h})(?=[{a}])".format(a=ALPHA, h=HYPHENS),
         r"(?<=[{a}0-9])[:<>=/](?=[{a}])".format(a=ALPHA),
     ]
 )
+
+
+# Some languages written with the Cyrillic alphabet permit the use of diacritics
+# to mark stressed syllables in words where stress is distinctive. Such languages
+# should use 'CYRILLIC_TOKENIZER_INFIXES' in place of 'TOKENIZER_INFIXES.
+CYRILLIC_TOKENIZER_INFIXES = list(TOKENIZER_INFIXES) + [
+    r"(?<=[{a}][{d}])\.".format(a=ALPHA, d=COMBINING_DIACRITICS),
+    r"(?<=[{a}][{d}])(?:{p})(?=[{a}])".format(a=ALPHA, d=COMBINING_DIACRITICS, p=PUNCT),
+    r"(?<=[{a}][{d}])(?:{h})(?=[{a}])".format(
+        a=ALPHA, d=COMBINING_DIACRITICS, h=HYPHENS
+    ),
+]
--- a/spacy/lang/bg/__init__.py
+++ b/spacy/lang/bg/__init__.py
@@ -2,6 +2,7 @@ from .stop_words import STOP_WORDS
 from .tokenizer_exceptions import TOKENIZER_EXCEPTIONS
 from .lex_attrs import LEX_ATTRS
 from ..tokenizer_exceptions import BASE_EXCEPTIONS
+from ..punctuation import CYRILLIC_TOKENIZER_INFIXES
 
 from ...language import Language, BaseDefaults
 from ...attrs import LANG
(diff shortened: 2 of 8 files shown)


spacy/lang/punctuation.py
```diff
@@ -39,8 +39,20 @@
         r"(?<=[{al}{q}])\.(?=[{au}{q}])".format(
             al=ALPHA_LOWER, au=ALPHA_UPPER, q=CONCAT_QUOTES
         ),
-        r"(?<=[{a}]),(?=[{a}])".format(a=ALPHA),
+        r"(?<=[{a}])(?:{p})(?=[{a}])".format(a=ALPHA, p=PUNCT),
         r"(?<=[{a}])(?:{h})(?=[{a}])".format(a=ALPHA, h=HYPHENS),
         r"(?<=[{a}0-9])[:<>=/](?=[{a}])".format(a=ALPHA),
     ]
 )
+
```
richardpaulhudson, line 47: I decided to put this here rather than copying it into the three language-specific directories because it is potentially relevant to any language with variable stress written in Cyrillic, of which there are many (e.g. several Finno-Ugric languages). If any of these languages are supported by spaCy in the future, it will be useful to have this definition here.

spacy/lang/punctuation.py
```diff
@@ -39,8 +39,20 @@
         r"(?<=[{al}{q}])\.(?=[{au}{q}])".format(
             al=ALPHA_LOWER, au=ALPHA_UPPER, q=CONCAT_QUOTES
         ),
-        r"(?<=[{a}]),(?=[{a}])".format(a=ALPHA),
+        r"(?<=[{a}])(?:{p})(?=[{a}])".format(a=ALPHA, p=PUNCT),
```
adrianeboyd, line 42: Why is this change included in this PR?

richardpaulhudson: Sorry, I'm not ready for this to be reviewed yet. I'll take it out of draft once I am.

spacy/lang/punctuation.py
```diff
@@ -44,3 +44,16 @@
         r"(?<=[{a}0-9])[:<>=/](?=[{a}])".format(a=ALPHA),
     ]
 )
+
+
+# Some languages written with the Cyrillic alphabet permit the use of diacritics
+# to mark stressed syllables in words where stress is distinctive. Such languages
+# should use 'CYRILLIC_TOKENIZER_INFIXES' in place of 'TOKENIZER_INFIXES'.
+CYRILLIC_TOKENIZER_INFIXES = list(TOKENIZER_INFIXES) + [
```
adrianeboyd, line 52: The naming could be more general here, right? It could be something like "COMBINING_DIACRITICS_TOKENIZER_INFIXES`?

spacy/lang/punctuation.py
```diff
@@ -44,3 +44,16 @@
...
+    r"(?<=[{a}][{d}])\.".format(a=ALPHA, d=COMBINING_DIACRITICS),
```
adrianeboyd, line 53: The regexes do let you do all kinds of things, but this is a suffix rather than an infix?

spacy/lang/punctuation.py
```diff
@@ -24,7 +24,7 @@
         r"(?<=°[FfCcKk])\.",
         r"(?<=[0-9])(?:{c})".format(c=CURRENCY),
         r"(?<=[0-9])(?:{u})".format(u=UNITS),
-        r"(?<=[0-9{al}{e}{p}(?:{q})])\.".format(
+        r"(?<=[0-9{al}{e}{p}{q}])\.".format(
```
richardpaulhudson, line 27: This non-capturing group didn't seem to serve any useful purpose, so I removed it and confirmed that doing so didn't seem to have any impact on the tests. I'd be grateful for confirmation that there doesn't seem to be any reason for keeping it.

spacy/lang/punctuation.py
```diff
@@ -44,3 +44,23 @@
...
     ]
 )
+
+
+# Some languages e.g. written with the Cyrillic alphabet permit the use of diacritics
+# to mark stressed syllables in words where stress is distinctive. Such languages
+# should use the COMBINING_DIACRITICS... suffix and infix regex lists in
+# place of the standard ones.
+COMBINING_DIACRITICS_TOKENIZER_SUFFIXES = list(TOKENIZER_SUFFIXES) + [
+    r"(?<=[{a}][{d}])\.".format(a=ALPHA, d=COMBINING_DIACRITICS),
```
richardpaulhudson, line 54: This regex includes single upper-case letters followed by a combining diacritic before a period. This is different from the behaviour for standard upper-case letters, where only two before a period are treated as a suffix. The difference is motivated by two things: the presence of the diacritic strongly suggests we are dealing with a real word rather than a random jumble of characters; and only handling two upper-case letters before a period taking diacritics into account would require several separate regexes because either or both could have a diacritic.

svlandeg (commented):
Looks good to me if we split that one edit into a separate PR :-)

spacy/lang/punctuation.py
svlandeg: Has this been included in a separate PR? I saw two but those were closed again? Just asking out of curiosity, if there was a reason not to go forward with this edit.

richardpaulhudson: I discovered that the groups did serve a purpose in the `CURRENCY` and `UNITS` regexes and thought perhaps there was an argument for maintaining the same structure throughout the regexes in the group, so that it was better to leave this as it is.

adrianeboyd (approved):

PR #10841, opened 2022-05-23
Update documentation for displacy style kwargs

maxTarlov:
## Description
Added `"span"` to the accepted values for the visualization styles in the `displacy.serve` and `displacy.render` top-level functions. The `"span"` style is new as of spaCy 3.3, so I added the "new" tag for that option only. The `"span"` visualization style is listed between the `"dep"` and `"ent"` visualization styles to avoid the appearance of the "new" tag applying to all visualization styles while maintaining the default-value-first convention.

### Types of change
Update documentation

## Checklist

### DIFF
--- a/website/docs/api/top-level.md
+++ b/website/docs/api/top-level.md
@@ -239,7 +239,7 @@ browser. Will run a simple web server.
 | Name      | Description                                                                                                                                                       |
 | --------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `docs`    | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span]], Doc, Span]~~                                                                             |
-| `style`   | Visualization style, `"dep"`, `"span"` <Tag variant="new">3.3</Tag> or `"ent"`. Defaults to `"dep"`. ~~str~~                                                                                             |
+| `style`   | Visualization style, `"dep"`, `"ent"` or `"span"` <Tag variant="new">3.3</Tag>. Defaults to `"dep"`. ~~str~~                                                                                             |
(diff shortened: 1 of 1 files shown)


maxTarlov:
I was unable to set up the website on my local machine with either npm or docker. Since the changes are so minimal, I manually checked the style by editing the web page in my browser and confirmed that it looked as it should to the best of my knowledge.

ljvmiranda921 (approved):
Hi @maxTarlov  thanks for this PR! For posterity, here's how it'll look like:

image

website/docs/api/top-level.md
```diff
@@ -239,7 +239,7 @@ browser. Will run a simple web server.
 | Name      | Description                                                                                                                                                       |
 | --------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `docs`    | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span]], Doc, Span]~~                                                                             |
-| `style`   | Visualization style, `"dep"` or `"ent"`. Defaults to `"dep"`. ~~str~~                                                                                             |
+| `style`   | Visualization style, `"dep"`, `"span"` <Tag variant="new">3.3</Tag> or `"ent"`. Defaults to `"dep"`. ~~str~~                                                                                             |
```
adrianeboyd, line 242: Nitpicky, but I think I'd prefer the order `dep`, `ent`, `span` which keeps it alphabetical and makes it easier to see the existing `ent` option, which gets kind of buried behind the new tag in this version.

maxTarlov: To be honest, I was a little unsure of the order I chose. I have changed it back to alphabetical order

adrianeboyd (approved):

PR #10844, opened 2022-05-24
Add guards around torch import for coref

polm:
Torch is required for the coref/spanpred models but shouldn't be
required for spaCy in general.

The one tricky part of this is that one function in coref_util relied on
torch, but that file was imported in several places. Since the function
was only used in one place I moved it there.

Tests are not working because they're a bit out of sync with other re-organizations, but that is not specific to the changes in this PR.

### Types of change

bug fix? for in-progress PR

## Checklist


kadarakos (approved):
lgtm!

PR #10847, opened 2022-05-25
luganda language extension

tobiusaolo:
This is an intiative to add luganda language from Uganda,East Africa to spacy

### DIFF
--- a/spacy/lang/lg/__init__.py
+++ b/spacy/lang/lg/__init__.py
@@ -11,11 +11,13 @@ Tokenizer exception: Special-case rules for the tokenizer, for example, contract
 
 from .stop_words import STOP_WORDS
 from .lex_attrs import LEX_ATTRS
+from .punctuation import TOKENIZER_INFIXES
 from ...language import Language, BaseDefaults
 
 
 class LugandaDefaults(BaseDefaults):
     lex_attr_getters = LEX_ATTRS
+    infixes = TOKENIZER_INFIXES
     stop_words = STOP_WORDS
 
 
--- a/spacy/lang/lg/__init__.py
+++ b/spacy/lang/lg/__init__.py
@@ -10,12 +10,12 @@ Tokenizer exception: Special-case rules for the tokenizer, for example, contract
 '''
 
 from .stop_words import STOP_WORDS
-#from .lex_attrs import LEX_ATTRS
+from .lex_attrs import LEX_ATTRS
 from ...language import Language, BaseDefaults
 
 
 class LugandaDefaults(BaseDefaults):
-    #lex_attr_getters = LEX_ATTRS
+    lex_attr_getters = LEX_ATTRS
     stop_words = STOP_WORDS
 
 
--- a/spacy/lang/lg/__init__.py
+++ b/spacy/lang/lg/__init__.py
@@ -1,0 +1,25 @@
+'''
+
+__init__.py:This defines the defaults and Language subclass
+Further we need the following
+Punctuation: Punctuation chars(?) for Luganda
+Stopwords: Frequent words that appear often in Luganda
+Syntax iterators: Functions that compute views of a Doc object based on its syntax, e.g. Noun chunks
+Tokenizer exception: Special-case rules for the tokenizer, for example, contractions like “ebye'mizannyo” and abbreviations with punctuation, like “Owek.”.
+
+'''
+import spacy
+from spacy.language import Language
+from stop_words import STOP_WORDS
+
+class LugandaDefaults(Language.Defaults):
+    stop_words = STOP_WORDS
+
+@spacy.registry.languages("lg")
+class Luganda(Language):
+    """docstring for ."""
+    lang="lg"
+    Defaults=LugandaDefaults
+# nlp2 = Luganda()
+# print(nlp2.lang, [token.is_stop for token in nlp2("wa ne lwa")])
+__all__=["Luganda"]
--- a/spacy/lang/lg/__init__.py
+++ b/spacy/lang/lg/__init__.py
@@ -1,14 +1,3 @@
-'''
-
-__init__.py:This defines the defaults and Language subclass
-Further we need the following
-Punctuation: Punctuation chars(?) for Luganda
-Stopwords: Frequent words that appear often in Luganda
-Syntax iterators: Functions that compute views of a Doc object based on its syntax, e.g. Noun chunks
-Tokenizer exception: Special-case rules for the tokenizer, for example, contractions like “ebye'mizannyo” and abbreviations with punctuation, like “Owek.”.
-
-'''
-
 from .stop_words import STOP_WORDS
 from .lex_attrs import LEX_ATTRS
 from .punctuation import TOKENIZER_INFIXES
--- a/spacy/lang/lg/__init__.py
+++ b/spacy/lang/lg/__init__.py
@@ -8,18 +8,20 @@ Syntax iterators: Functions that compute views of a Doc object based on its synt
 Tokenizer exception: Special-case rules for the tokenizer, for example, contractions like “ebye'mizannyo” and abbreviations with punctuation, like “Owek.”.
 
 '''
-import spacy
-from spacy.language import Language
-from stop_words import STOP_WORDS
 
-class LugandaDefaults(Language.Defaults):
+from .stop_words import STOP_WORDS
+#from .lex_attrs import LEX_ATTRS
+from ...language import Language, BaseDefaults
+
+
+class LugandaDefaults(BaseDefaults):
+    #lex_attr_getters = LEX_ATTRS
     stop_words = STOP_WORDS
 
-@spacy.registry.languages("lg")
+
 class Luganda(Language):
-    """docstring for ."""
-    lang="lg"
-    Defaults=LugandaDefaults
-# nlp2 = Luganda()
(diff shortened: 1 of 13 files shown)


polm:
You can click on the red X to see why a build failed, though in this case the most recent build succeeded so you're fine. We'll take a look at this, thanks for the submission!

spacy/lang/lg/lex_attrs.py
```diff
@@ -0,0 +1,74 @@
+_num_words=[
```
adrianeboyd, line 1:
```suggestion
from ...attrs import LIKE_NUM


_num_words=[
```

spacy/lang/lg/lex_attrs.py
```diff
@@ -0,0 +1,74 @@
...
+"ekyekyenda",#ninetieth
+"ekyekikumi",#hundredth
+"ekyolukumi",#thousandth
+"ekyakakadde",#millionth
+"ekyakawumbi",#billionth
+"ekyakase",#trillionth
+"ekyakatabalika",#quadrillionth
+"ekyakeesedde",#gajillionth
+"ekyakafukunya"#bazillionth
+]
```
adrianeboyd, line 74: This is a suggestion for a `like_num` method that checks whether you have a number out of digits like `1000` or a word like `thousand` (taken from your list above). I'm not sure exactly what the right checks would be for Luganda, but hopefully this will help you get started:


```suggestion
]


def like_num(text):
    if text.startswith(("+", "-", "±", "~")):
        text = text[1:]
    text = text.replace(",", "").replace(".", "")
    if text.isdigit():
        return True
    if text.count("/") == 1:
        num, denom = text.split("/")
        if num.isdigit() and denom.isdigit():
            return True
    if text.lower() in _num_words:
        return True
    return False


LEX_ATTRS = {LIKE_NUM: like_num}
```

spacy/lang/lg/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+'''
+
+from .stop_words import STOP_WORDS
+#from .lex_attrs import LEX_ATTRS
+from ...language import Language, BaseDefaults
+
+
+class LugandaDefaults(BaseDefaults):
+    #lex_attr_getters = LEX_ATTRS
```
adrianeboyd, line 18: Once you've updated `lex_attrs.py` you can add the imports and defintions here:


```suggestion
from .lex_attrs import LEX_ATTRS
from ...language import Language, BaseDefaults


class LugandaDefaults(BaseDefaults):
    lex_attr_getters = LEX_ATTRS
```

adrianeboyd (commented):
Thanks for the PR, this is a great start! I added some comments about how to configure the `LIKE_NUM` modifications for `LEX_ATTRS`.

spacy/lang/lg/lex_attrs.py
tobiusaolo: Hey I have worked on the above comment

spacy/lang/lg/lex_attrs.py
tobiusaolo: worked on it

spacy/lang/lg/lex_attrs.py
```diff
@@ -0,0 +1,97 @@
...
+    if text.isdigit():
+        return True
+    if text.count("/") == 1:
+        num, denom = text.split("/")
+        if num.isdigit() and denom.isdigit():
+            return True
+    text_lower = text.lower()
+    if text_lower in _num_words:
+        return True
+    if text_lower.endswith(("st", "nd", "rd", "th")):
```
adrianeboyd, line 91: This if-statement is meant for English tokens like `1st`, `2nd`, `3rd`, `4th` and I doubt it's correct for Luganda?

spacy/lang/lg/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+__init__.py:This defines the defaults and Language subclass
+Further we need the following
+Punctuation: Punctuation chars(?) for Luganda
+Stopwords: Frequent words that appear often in Luganda
+Syntax iterators: Functions that compute views of a Doc object based on its syntax, e.g. Noun chunks
+Tokenizer exception: Special-case rules for the tokenizer, for example, contractions like “ebye'mizannyo” and abbreviations with punctuation, like “Owek.”.
+
+'''
+
```
adrianeboyd, line 11:
```suggestion
```

spacy/lang/lg/stop_words.py
```diff
@@ -0,0 +1,20 @@
+#stopwords as whitespace-seperated list
```
adrianeboyd, line 1: Do you have a source (link or citation?) for the stop words?

spacy/lang/lg/stop_words.py
```diff
@@ -0,0 +1,20 @@
...
+	"ye nyinza ateekeddwa tetuteekeddwa neetaaga seetaaga nedda edda kati ku gumu gujja oba ekirala wabweru waggulu"
+	"nnina byebimu n'olwekyo ekyo bo abava bingi abangi ojja bangi waliyo bino bwabwe bandi bajja ajja wansi bulijjo kaseera ba"
+	"balina kino ebyo ku nnyo ennyo okutuusa  bwayo yabadde ffe tu-yina kyekimu"
+	"oyo babadde baali tebaali ki kiki ddi wa ani lwaki ne gwe wandi oli oyina kikyo e mu wange ku bwe wa bajja"
+	"newankubade sinakindi n'olwekyo okuggyako gunno guno bateekeddwa oba gwe mwe"
+ 	"gyabwe erina tolina ebimu mingi zijja ffe nanti anti naye ate"
+ 	"wamu awamu baweebwa aweebwa weebwa era wadde mpozzi ekyo oyo kati  kyekyo oluvannyuma kwegamba nandiyagadde wadde kubanga"
+ 	 "olwokuba wabula nnyo nnyini nnyinza tuyina tulina tayina balina bali okuwa twetaaga okugenda bayina alina mulina"
+	"oyina olina abamu bano ye otya ki ono gwa nabadde mbadde".split()
+)
```
svlandeg, line 20: I would format this as 
```
STOP_WORDS = set(
    """
wa lwa si ...
...
... nabadde mbadde
""".split()
)
```
and sort this alphabetically

adrianeboyd:
Thanks for the updates, this is looking good! In a second I'll try to make a few minor edits and reformat so this is ready to merge...

adrianeboyd:
Actually, one more question: what is the intended tokenization of strings like `'ab’emmamba'` and `"ky'ebyenjigiriza"`? When I try out the examples (thanks for adding a few!), I get the tokenization:

```none
Abooluganda ab’emmamba ababiri ['Abooluganda', 'ab’emmamba', 'ababiri']
Ekisaawe ky'ebyenjigiriza kya mugaso nnyo ['Ekisaawe', "ky'ebyenjigiriza", 'kya', 'mugaso', 'nnyo']
```

From the stop words, it looks like you're expecting `"ky'` to be a separate token?

If I know what the tokenization is intended to be, I can add a few tokenizer tests and help adjust the tokenizer settings.

tobiusaolo:
Thank you for your feedback, first about  tokenization i have consulted the
language expert and if we follow the sentence below:

sentence:Abooluganda ab’emmamba ababiri

We can tokenize the above sentence as
['Abooluganda', 'ab’emmamba', 'ababiri']

About the 'ky' i will remove it and update the repo otherwise thank
you for the guidance

Regards

adrianeboyd:
Do you have a source for the stop words?

I'm still a bit confused about the tokenizer settings vs. stop words.

Is `ky'` ever a separate token and not just a prefix? With the current tokenizer settings, none of the stop words with `'` will end up as separate tokens, so the stop words with apostrophes might not make sense.

For example:

```python
import spacy

nlp = spacy.blank("lg")

doc = nlp("Ekiwandiiko ky'olunaku")
print([t.text for t in doc]) # ['Ekiwandiiko', "ky'olunaku"]
```

I will add some basic tokenizer tests in a minute with the example above.

tobiusaolo:
You're right, the ky' is not supposed it a prefix. According to the
discussion with the Luganda Experts, they indicated that the word should
remain "ky'olunaku" when tokenized.
For the source of stop words we are its don't yet publish but it was list
generated by the Experts here.

adrianeboyd:
I'm worried that users will be confused in the future because "ky'" is a stop word but never a separate token that could be marked as a stop word. Does it make sense to remove all these stop words?

```
contractions = [
    "b'",
    "bw'",
    "by'",
    "eky'",
    "ey'",
    "ez'",
    "g'",
    "gw'",
    "gy'",
    "ky'",
    "lw'",
    "ly'",
    "n'",
    "ng'",
    "olw'",
    "ow'",
    "w'",
    "y'",
    "z'",
]
```

tobiusaolo:
I sent in a new stopwords list in the latest PR which does not include the
"ky',b'". Those words were transferred to the contractions.
I suggest that the stopwords should stand  since the contractions are
distinct .
Kind regards

adrianeboyd:
Sorry for the delay, I thought I should wait on an update because in the current version the contractions are still added to the stop words. If the contractions are removed, then I think this is fine to merge. Let me go ahead and do that...

We're actually planning to remove the default stop word lists for v4, but I was hoping to leave all the stop words in v3 as a useful reference for users.

adrianeboyd:
Thanks again for the PR! We'll mention Luganda in the release notes for the next release (probably v3.4.2).

Issue #10839, opened 2022-05-23
Pipe() process time is very slow with as_tuples attribute


hugoobauer:
When I set `as_tuples=True` attribute to the `.pipe` method, the process time is very very slow.

Results without tuples :
```
0:00:02.997644 Starting pipe
text processed
[...]
text processed
Pipe finished in 0:00:13.833113
```

Results with tuples :
```
0:00:02.879235 Starting pipe
text processed
[...]
text processed
Pipe finished in 0:02:19.793344
```

## How to reproduce the behaviour
Here a code sample :
```
import datetime

import spacy


TIME = datetime.datetime.now()


def time_profiler():
    global TIME
    diff = datetime.datetime.now() - TIME
    TIME = datetime.datetime.now()
    return diff


if __name__ == "__main__":
    nlp = spacy.load("fr_core_news_lg")
    # text_to_processed = nlp.pipe(
    #     [f"<p>Testing without tuple</p>" * 100] * 50,
    #     n_process=2, batch_size=1
    # )
    text_to_processed = nlp.pipe(
        [(f"<p>Testing tuple</p>" * 100, -1)] * 50,
        n_process=2, batch_size=1, as_tuples=True
    )
    print(time_profiler(), "Starting pipe")
    # for text_processed in text_to_processed:
    #     print("text processed")
    for text_processed, _id in text_to_processed:
        print("text processed", _id)
    print("Pipe finished in", time_profiler())
```

## Your Environment
- **spaCy version:** 3.3.0
- **Platform:** macOS-12.4-arm64-arm-64bit
- **Python version:** 3.9.11
- **Pipelines:** fr_core_news_lg (3.3.0)

shadeMe:
Thanks for the report and the code sample! I'm able to reproduce the slow down on my end (on Linux) as well. Looks like the code spends an inordinate amount of time serializing and deserializing data during IPC (inter-process communication).

PR #10864, opened 2022-05-27
Avoid pickling `Doc` inputs passed to `Language.pipe()`

shadeMe:
## Description

A user [reported](https://github.com/explosion/spaCy/issues/10839) significantly slower performance when calling `Language.pipe()` with `as_tuples=True` (as compared to `as_tuples=False`). Closer investigation revealed the following:
- When `as_tuples` is set to `True`, the inputs passed as `texts` are preemptively converted to `Doc` objects (to store their respective context data in the `_context` attribute). Then, `pipe()` is called again with the augmented inputs.
- When multiprocessing is enabled (`n_process > 1`), `Doc` objects passed to `pipe()` are pickled as-is and sent to child processes along with their shared `Vocab` and `Vectors` instances.
- This incurs a significant serialization/deserialization overhead, which is further exacerbated by small batch sizes.

This PR address the issue through the following changes:
- Serialize `Doc` objects in the input to byte arrays using `Doc.to_bytes()` when multiprocessing is enabled. The overhead of pickling a byte array is negligible.
- In child processes, deserialize incoming byte arrays to their correspond `Doc` representations before executing the pipeline.
- Ensure that the `_context` attribute is serialized when calling `Doc.to_bytes/to_dict()` in order to prevent the loss of context information.

### Relevant issues
- #10839

### Types of change

Performance enhancement

## Checklist


shadeMe:

adrianeboyd:
`_context` is just internals used for multiprocessing with `pipe`, so we don't want to add it to `Doc.to_bytes()`, which affects every serialized `DocBin`/`.spacy`.

The existing implementation added it to `pickle_doc` to send docs and added it to the tuples that are sent back to receive docs.

https://github.com/explosion/spaCy/blob/87adb3257600b5d0cc3540f6c7246ad2ccd0c0f9/spacy/language.py#L2192

For this, I think we can take it out of `pickle_doc` since we're not using pickle anymore, and use the same approach (some kind of tuple?) when sending so that `_context` stays more as internals.

shadeMe:
If `_context` is only ever used in `pipe` and we want to pass the context data directly as part of a tuple, shall we then remove `_context` from `Doc` entirely?

adrianeboyd:
Hmm, maybe we don't actually need `Doc._context` at all here anymore?

shadeMe:
Posted here for context: Seems like we can't get rid of `_context` so easily. `pipe` is allowed to return processed batches that are smaller then the input ([this PR](https://github.com/explosion/spaCy/pull/8004) talks about custom error handlers potentially skipping specific documents). So, maintaining the alignment between the source `Doc`s and their corresponding contexts will require either storing the context object on the document itself (like we do currently) or using a secondary mapping of unique `Doc` IDs (that cannot be mutated after init) to context objects. So, we'll continue using the `Doc._context` field when passing `Doc`s to the pipeline.

shadeMe:
I've simplified the code further.

shadeMe:

spacy/language.py
```diff
@@ -1597,16 +1614,17 @@ def _multiprocessing_pipe(
         n_process: int,
         batch_size: int,
     ) -> Iterator[Doc]:
+        serialized_texts_with_ctx = self._prepare_doc_for_ipc(texts)  # type: ignore
         # raw_texts is used later to stop iteration.
-        texts, raw_texts = itertools.tee(texts)
+        texts_itr, raw_texts = itertools.tee(serialized_texts_with_ctx)
```
adrianeboyd, line 1619:
```suggestion
        texts, raw_texts = itertools.tee(serialized_texts_with_ctx)
```

spacy/language.py
```diff
@@ -1597,16 +1614,17 @@ def _multiprocessing_pipe(
...
         # for sending texts to worker
         texts_q: List[mp.Queue] = [mp.Queue() for _ in range(n_process)]
         # for receiving byte-encoded docs from worker
         bytedocs_recv_ch, bytedocs_send_ch = zip(
             *[mp.Pipe(False) for _ in range(n_process)]
         )
 
-        batch_texts = util.minibatch(texts, batch_size)
+        batch_texts = util.minibatch(texts_itr, batch_size)
```
adrianeboyd, line 1627:
```suggestion
        batch_texts = util.minibatch(texts, batch_size)
```

spacy/tokens/doc.pyx
```diff
@@ -1393,6 +1393,7 @@ cdef class Doc:
                 self.vocab.strings.add(s)
         if "has_unknown_spaces" not in exclude and "has_unknown_spaces" in msg:
             self.has_unknown_spaces = msg["has_unknown_spaces"]
+
```
adrianeboyd, line 1396:
```suggestion
```

spacy/language.py
```diff
@@ -1576,9 +1580,22 @@ def pipe(  # noqa: F811
...
+    ) -> Iterable[Tuple[Union[str, bytes], _AnyContext]]:
+        # Serialize Doc inputs to bytes to avoid incurring pickling
+        # overhead when they are passed to child processes. Also yield
+        # any context objects they might have separately (as they are not serialized).
+        for doc_like in texts:
+            if isinstance(doc_like, Doc):
+                yield (doc_like.to_bytes(), cast(_AnyContext, doc_like._context))
+            else:
+                yield (doc_like, cast(_AnyContext, None))
+
```
adrianeboyd, line 1598: I think this could be defined within `_multiprocessing_pipe` and then it could have a simpler name like `prepare_doc`?

adrianeboyd (commented):
Looks good! A few very minor naming/formatting suggestions...

spacy/language.py
```diff
@@ -1597,8 +1601,21 @@ def _multiprocessing_pipe(
         n_process: int,
         batch_size: int,
     ) -> Iterator[Doc]:
+        def prepare_doc(
```
adrianeboyd, line 1604: I'm still not happy with my own naming suggestion. `prepare_input`? `split_context`?

shadeMe: I'll go for `prepare_input`. The comment inside the function should describe the intent more concretely, in any event.

adrianeboyd (approved):

spacy/errors.py
```diff
@@ -919,6 +919,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1035 = ("Token index {i} out of bounds ({length})")
     E1036 = ("Cannot index into NoneNode")
     E1037 = ("Invalid attribute value '{attr}'.")
+    E1038 = ("Expected a string, 'Doc', or 'bytes' as input, but got: {type}.")
```
adrianeboyd, line 922:
```suggestion
    E1038 = ("Expected a string, Doc, or bytes as input, but got: {type}")
```

PR #10867, opened 2022-05-29
Fix misspelt keyword in StringStore docs example

freddyheppell:
## Description

Fixes a syntax error in an example in the StringStore docs caused by a misspelt keyword

### Types of change
* Documentation example code fix

## Checklist


polm:
Thanks for the fix!

PR #10873, opened 2022-05-30
Remove English exceptions with mismatched features

adrianeboyd:
## Description

Remove English contraction exceptions with mismatched features that lead to exceptions like "theses" and "thisre".

### Types of change

Bug fix.

## Checklist


adrianeboyd:

spacy/lang/en/tokenizer_exceptions.py
```diff
@@ -139,26 +139,27 @@
...
-            {ORTH: "'s", NORM: "'s"},
-        ]
+        if morph != "Number=Plur|Person=3":
+            _exc[orth + "'s"] = [
+                {ORTH: orth, NORM: word},
+                {ORTH: "'s", NORM: "'s"},
+            ]
 
-        _exc[orth + "s"] = [{ORTH: orth, NORM: word}, {ORTH: "s"}]
+            _exc[orth + "s"] = [{ORTH: orth, NORM: word}, {ORTH: "s"}]
```
svlandeg, line 162: You could still have `thiss`, right? Which would likely be more like a misspelling/typo than two actual tokens?

adrianeboyd: This PR only prevents `these s` or `this re` from being split into two tokens.

A lot of these exceptions (in particular `id`) have many contexts where they weren't intended as two tokens, but this PR is just trying to fix this/that/these/those where there's no analysis that makes sense as two tokens.

PR #10874, opened 2022-05-30
Make pipe method working with stream of texts

hugoobauer:
## Description
This RP follows [this discussion](https://github.com/explosion/spaCy/discussions/10838) where I explain the problem in detail.
The `pipe()` method does not work with an infinite stream of documents.
Here is an example of code that shows that the code remains blocked for the first few documents until the number of documents is sufficient :
```
from threading import Thread
from queue import Queue

import spacy

q = Queue()


def infinite_sequences():
    while True:
        yield q.get()


def pipeline_runner(_nlp: spacy.language.Language):
    my_pipe = _nlp.pipe(infinite_sequences(), n_process=2, batch_size=1)
    print("NLP pipe started")
    for doc in my_pipe:
        print("Doc processed", doc[:10])
    print("NLP pipe ended")


if __name__ == "__main__":
    nlp = spacy.load("fr_core_news_lg")
    pipeline_thread = Thread(target=pipeline_runner, args=(nlp, ))
    pipeline_thread.start()
    # Will be blocked and doc will never be returned
    # With 4 texts, will work and return only the first 2 texts and then block and not return the next 2 documents until other documents have been received by the pipe methode.
    for i in range(1):  # <-- change this number to 4
        elem = f"<p>testing {i}</p>" * 1000
        q.put(elem)
        print("Elem inserted to queue", elem[:10])
    pipeline_thread.join()

```

The main loop is blocked as long as the number of documents does not match the number `n_process*batch`
My changes allow to parallelize the reading and writing in the queues without blocking the main loop.

### Types of change
Enhancement of `pipe` methode.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -9,6 +9,7 @@ import functools
 from contextlib import contextmanager
 from copy import deepcopy
 from pathlib import Path
+from threading import Thread
 import warnings
 from thinc.api import get_current_ops, Config, CupyOps, Optimizer
 import srsly
@@ -1600,7 +1601,7 @@ class Language:
         # raw_texts is used later to stop iteration.
         texts, raw_texts = itertools.tee(texts)
         # for sending texts to worker
-        texts_q: List[mp.Queue] = [mp.Queue() for _ in range(n_process)]
+        texts_q: List[mp.Queue] = [mp.Queue(maxsize=batch_size) for _ in range(n_process)]
         # for receiving byte-encoded docs from worker
         bytedocs_recv_ch, bytedocs_send_ch = zip(
             *[mp.Pipe(False) for _ in range(n_process)]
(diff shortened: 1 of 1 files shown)


adrianeboyd:
I do agree that there are some places where `nlp.pipe` could work better for infinite generators (I suspect you may also run into memory issues for long-running processes with `tee`), but I think it would be better to solve this particular use case outside of `nlp.pipe`, which is meant for processing large amounts of data in batches.

If you're using `batch_size=1`, you can use `nlp` instead of `nlp.pipe` and handle multiprocessing at a higher level in your code. If you're loading the model and then using fork in linux, most of the model data shouldn't be duplicated in memory.

I'll have a look to see if there's a sensible alternative to the current way it steps the batches, but I'm not sure at this point.

svlandeg:
Considering Adriane's original reply, I'll go ahead and close this PR for now, but we'll add this issue to our internal backlog. Thanks!

PR #10875, opened 2022-05-30
Refactor Coval Scoring code

polm:
## Description

This moves the scoring code that was taken from [coval]() into `scorer.py`, renames some things to be less generic, and adds the license to our list. 

### Types of change

Refactor

## Checklist


spacy/scorer.py
```diff
@@ -1143,3 +1143,144 @@ def _auc(x, y):
         # regular numpy.ndarray instances.
         area = area.dtype.type(area)
     return area
+
+
```
adrianeboyd, line 1147: Can you add a comment header here to separate this code from the preceding scikit-learn code (just similar to what's there for ROC AUC)?

polm:
Note: I'm going to go ahead and refactor the scorers into registered scorers in this PR too.

spacy/scorer.py
polm: Added in the most recent commit.

spacy/ml/models/coref.py
```diff
@@ -56,7 +56,9 @@ def convert_coref_clusterer_inputs(model: Model, X: List[Floats2d], is_train: bo
     def backprop(args: ArgsKwargs) -> List[Floats2d]:
         # convert to xp and wrap in list
         gradients = torch2xp(args.args[0])
-        assert isinstance(gradients, Floats2d)
+        #TODO why did this change? This was fine before merging master.
+        # It's still 2d but this assert fails.
+        #assert isinstance(gradients, Floats2d)
```
polm, line 61: Merging in `master` broke this `assert` for some reason, I'm not sure why. I need to look at this more.

svlandeg: Probably because of a change in `8.1.0.dev0`?

polm: I considered the possibility of Thinc changes, but I haven't updated my dev environment, so that's not the case. I'm using Thinc 8.0.16.

spacy/scorer.py
```diff
@@ -1289,3 +1346,22 @@ def lea(input_clusters, output_clusters, mention_to_gold):
         den += len(c)
 
     return num, den
+
+
+# This is coref related, but not from coval.
+# def doc2clusters(doc: Doc, prefix) -> MentionClusters:
```
polm, line 1352: It would be nice to have MentionClusters here but I don't know where to stick it to avoid circular imports.

spacy/scorer.py
```diff
@@ -873,6 +875,67 @@ def score_deps(
...
+                        pred_starts.append(pred_mention.start)
+                        pred_ends.append(pred_mention.end)
+
+
+            # TODO check logic
+            # see how many are perfect
+            cs = [a == b for a, b in zip(starts, pred_starts)]
+            ce = [a == b for a, b in zip(ends, pred_ends)]
+            correct = [int(a and b) for a, b in zip(cs, ce)]
+            accuracy = sum(correct) / len(correct)
```
polm, line 933: @kadarakos I had to rewrite this because we can't pass the xp / numpy thing here. I think it's OK but let me know if anything is off about it.

spacy/ml/models/coref.py
polm: This is the only thing here I still don't totally understand, but it doesn't need to hold up this PR. (I'm also not sure why the assert was here, @kadarakos was there an error case or something this was checking for?

polm:
I ended up fixing the tests (mostly by skipping when Torch isn't present) and types (mostly by ignoring or removing them) :/ . It is green now at least, and should be ready to merge in the coref branch for further cleanup.

spacy/ml/models/coref.py
kadarakos: I don't think I've put the assert there I also have no idea.

spacy/scorer.py
polm: In the scorer we only pass the examples and the config, so we don't have access to the `np` associated with the model, and we can't put it in the config because it's not serializable. (Maybe I'm missing something but that's why I assumed this was necessary.) Look at the other functions around this one.

spacy/tests/pipeline/test_coref.py
```diff
@@ -77,6 +82,7 @@ def test_initialized_short(nlp):
     print(doc.spans)
```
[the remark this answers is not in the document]
polm: Must have been leftover debugging, I'll remove it.

spacy/ml/models/coref.py
polm: OK, I'll just remove it.

spacy/tests/pipeline/test_coref.py
```diff
@@ -45,18 +47,20 @@ def snlp():
...
 def test_add_pipe(nlp):
     nlp.add_pipe("coref")
     assert nlp.pipe_names == ["coref"]
 
 
+@pytest.mark.skipif(not has_torch, reason="Torch not available")
 def test_not_initialized(nlp):
     nlp.add_pipe("coref")
     text = "She gave me her pen."
     with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
polm: Added `match="E109"`, which corresponds to that error.

PR #10878, opened 2022-05-30
Reimplement parser rehearsal function

danieldk:
## Description

Before the parser refactor, rehearsal was driven by a loop in the `rehearse` method itself. For each parsing step, the loops would:

1. Get the predictions of the teacher.
2. Get the predictions and backprop function of the student.
3. Compute the loss and backprop into the student.
4. Move the teacher and student forward with the predictions of
   the student.

In the refactored parser, we cannot perform search stepwise rehearsal anymore, since the model now predicts all parsing steps at once. Therefore, rehearsal is performed in the following steps:

1. Get the predictions of all parsing steps from the student, along with its backprop function.
2. Get the predictions from the teacher, but use the predictions of the student to advance the parser while doing so.
3. Compute the loss and backprop into the student.

To support the second step a new method, `advance_with_actions` is added to `GreedyBatch`, which performs the provided parsing steps.

---

Another thing to be aware of while reviewing this PR is that an issue surfaced while reinstating the rehearsal test is that the Thinc optimizers do not support resizing of existing parameters. The solution that we came up after some discussion is that we should work around this by wrapping the weights/biases of the upper layer of the parser model in `Linear` (like the old parser). When the upper layer is resized, we copy over the existing parameters into a new `Linear` instance. This does not trigger an error in Optimizer, because it sees the resized layer as a new set of parameters.

### Types of change

Refactor

## Checklist

### DIFF
--- a/spacy/ml/tb_framework.pyx
+++ b/spacy/ml/tb_framework.pyx
@@ -1,5 +1,5 @@
 # cython: infer_types=True, cdivision=True, boundscheck=False
-from typing import List, Tuple, Any, Optional, cast
+from typing import List, Tuple, Any, Optional, TypeVar, cast
 from libc.string cimport memset, memcpy
 from libc.stdlib cimport calloc, free, realloc
 from libcpp.vector cimport vector
@@ -10,12 +10,14 @@ from thinc.api import uniform_init, glorot_uniform_init, zero_init
 from thinc.api import NumpyOps
 from thinc.backends.linalg cimport Vec, VecVec
 from thinc.backends.cblas cimport CBlas
-from thinc.types import Floats1d, Floats2d, Floats3d, Ints2d, Floats4d
+from thinc.types import Floats1d, Floats2d, Floats3d, Floats4d
+from thinc.types import Ints1d, Ints2d
 
 from ..errors import Errors
 from ..pipeline._parser_internals import _beam_utils
 from ..pipeline._parser_internals.batch import GreedyBatch
-from ..pipeline._parser_internals.transition_system cimport c_transition_batch, TransitionSystem
+from ..pipeline._parser_internals.transition_system cimport c_transition_batch, c_apply_actions
+from ..pipeline._parser_internals.transition_system cimport TransitionSystem
 from ..pipeline._parser_internals.stateclass cimport StateC, StateClass
 from ..tokens.doc import Doc
 from ..util import registry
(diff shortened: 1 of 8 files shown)


spacy/ml/tb_framework.pyx
```diff
@@ -43,18 +45,28 @@ def TransitionModel(
...
     tok2vec_projected.set_dim("nO", hidden_width)
 
+    # Fixme: we use `upper` as a container for the upper layer's
+    # weights and biases. Thinc optimizers cannot handle resizing
+    # of parameters. So, when the parser model is resized, we
+    # construct a new `upper` layer, which has a different key in
+    # the optimizer. Once the optimizer supports parameter resizing,
+    # we can replace the `upper` layer by `upper_W` and `upper_b`
+    # parameters in this model.
+    upper = Linear(nO=None, nI=hidden_width, init_W=zero_init)
```
shadeMe, line 55: Would it more preferable to rename `upper` to `output` here? Using a positional name that implicitly depends on how one visualises the topology of the layers can be a bit confounding, I think (on first glance, I implicitly associated `upper` with `input_layer` as I tend to visualise a network in a top-down manner).

spacy/ml/tb_framework.pyx
```diff
@@ -113,9 +127,7 @@ def init(
     inferred_nO = _infer_nO(Y)
     if inferred_nO is not None:
         current_nO = model.maybe_get_dim("nO")
-        if current_nO is None:
-            model.set_dim("nO", inferred_nO)
-        elif current_nO != inferred_nO:
+        if current_nO is None or current_nO != inferred_nO:
```
shadeMe, line 130: Why was `model.set_dim("nO", inferred_nO)` removed here?

spacy/ml/tb_framework.pyx
```diff
@@ -127,72 +139,84 @@ def init(
...
-                        np.ndarray[np.npy_bool, ndim=1] seen_mask):
-    cdef vector[StateC *] c_states
+                np.ndarray[np.npy_bool, ndim=1] seen_mask, actions: Optional[List[Ints1d]]=None):
+    cdef vector[StateC*] c_states
     cdef StateClass state
     for state in states:
         if not state.is_final():
             c_states.push_back(state.c)
-    weights = get_c_weights(model, <float *> feats.data, seen_mask)
+    cdef object featsp = feats
```
shadeMe, line 187: This variable appears to be unused.

spacy/ml/tb_framework.pyx
```diff
@@ -127,72 +139,84 @@ def init(
...
     cdef CBlas cblas = model.ops.cblas()
-    scores = _parseC(cblas, moves, &c_states[0], weights, sizes)
+    scores = _parseC(cblas, moves, &c_states[0], weights, sizes, actions=actions)
 
     def backprop(dY):
         raise ValueError(Errors.E1038)
 
     return (states, scores), backprop
 
 cdef list _parseC(CBlas cblas, TransitionSystem moves, StateC** states,
```
shadeMe, line 200: Not a part of this PR per-se, but this method seems like it can benefit from a more descriptive name.

spacy/pipeline/_parser_internals/transition_system.pyx
```diff
@@ -155,6 +155,15 @@ cdef class TransitionSystem:
         action.do(state.c, action.label)
         state.c.history.push_back(action.clas)
 
+    def apply_actions(self, states, const int[::1] actions):
+        assert len(states) == actions.shape[0]
+        cdef StateClass state
+        cdef vector[StateC*] c_states
+        for state in states:
+            c_states.push_back(state.c)
```
shadeMe, line 163: We can allocate the memory for the entire vector using `resize` and index-init the pointers to avoid the EH overhead of `push_back`.

spacy/ml/tb_framework.pyx
danieldk: It's set `resize_output`.

spacy/ml/tb_framework.pyx
danieldk: Agreed.

spacy/ml/tb_framework.pyx
danieldk: Thanks, fixed!

spacy/pipeline/_parser_internals/transition_system.pyx
danieldk: Great idea. Fixed.

spacy/ml/tb_framework.pyx
```diff
@@ -127,72 +139,83 @@ def init(
...
     feats, backprop_feats = _forward_precomputable_affine(model, tokvecs, is_train)
     seen_mask = _get_seen_mask(model)
 
+    # Fixme: support actions in forward_cpu
     if beam_width == 1 and not is_train and isinstance(model.ops, NumpyOps):
-        return _forward_greedy_cpu(model, moves, states, feats, seen_mask)
+        return _forward_greedy_cpu(model, moves, states, feats, seen_mask, actions=actions)
     else:
-        return _forward_fallback(model, moves, states, tokvecs, backprop_tok2vec, feats, backprop_feats, seen_mask, is_train)
+        return _forward_fallback(model, moves, states, tokvecs, backprop_tok2vec, feats, backprop_feats, seen_mask, is_train, actions)
```
svlandeg, line 177:
```suggestion
        return _forward_fallback(model, moves, states, tokvecs, backprop_tok2vec, feats, backprop_feats, seen_mask, is_train, actions=actions)
```
(for symmetry but more importantly I think it's good practice to use the `keyword=` style for optional arguments)

spacy/pipeline/transition_parser.pyx
```diff
@@ -359,7 +361,42 @@ class Parser(TrainablePipe):
...
         """Perform a "rehearsal" update, to prevent catastrophic forgetting."""
-        raise NotImplementedError
+        if losses is None:
+            losses = {}
+        for multitask in self._multitasks:
+            if hasattr(multitask, 'rehearse'):
+                multitask.rehearse(examples, losses=losses, sgd=sgd)
+        if self._rehearsal_model is None:
+            return None
+        losses.setdefault(self.name, 0.)
```
svlandeg, line 371:
```suggestion
        losses.setdefault(self.name, 0.0)
```
(I think we've been slightly inconsistent in the past but should probably standardize on `0.0`?)

svlandeg (approved):
Looks good to me! I only had a few nitpicks.

shadeMe (approved):

spacy/ml/tb_framework.pyx
```diff
@@ -74,29 +86,31 @@ def TransitionModel(
...
+        old_b = upper.get_param("b")
         new_W[:old_nO] = old_W  # type: ignore
         new_b[:old_nO] = old_b  # type: ignore
         for i in range(old_nO, new_nO):
             model.attrs["unseen_classes"].add(i)
-        model.set_param("upper_W", new_W)
-        model.set_param("upper_b", new_b)
+        upper = new_upper
+        model.layers[-1] = upper
+        model.set_ref("upper", upper)
```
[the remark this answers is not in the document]
danieldk: No good reason :). Fixed.

PR #10904, opened 2022-06-03
Use thinc-apple-ops>=0.1.0.dev0 with `apple` extras

danieldk:
## Description

This change ensures that installing spaCy with `apple` uses thinc-apple-ops version 0.1.0.dev0 or later, which has the proper CBLAS passthrough. Also test with at least this version.

### Types of change

Enhancement

## Checklist

- [x] I confirm that I have the right to submit this contribution under 
the project's MIT license.
- [x] My changes don't require a change to the documentation, or if they 
do, I've added all required information.

### DIFF
--- a/.github/azure-steps.yml
+++ b/.github/azure-steps.yml
@@ -111,7 +111,7 @@ steps:
     condition: eq(variables['python_version'], '3.8')
 
   - script: |
-      ${{ parameters.prefix }} python -m pip install "thinc-apple-ops>=0.1.0.dev0"
+      ${{ parameters.prefix }} python -m pip install --pre thinc-apple-ops
       ${{ parameters.prefix }} python -m pytest --pyargs spacy
     displayName: "Run CPU tests with thinc-apple-ops"
     condition: and(startsWith(variables['imageName'], 'macos'), eq(variables['python.version'], '3.10'))
--- a/.github/azure-steps.yml
+++ b/.github/azure-steps.yml
@@ -111,7 +111,7 @@ steps:
     condition: eq(variables['python_version'], '3.8')
 
   - script: |
-      ${{ parameters.prefix }} python -m pip install thinc-apple-ops
+      ${{ parameters.prefix }} python -m pip install "thinc-apple-ops>=0.1.0.dev0"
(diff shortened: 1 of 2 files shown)


.github/azure-steps.yml
```diff
@@ -111,7 +111,7 @@ steps:
     condition: eq(variables['python_version'], '3.8')
 
   - script: |
-      ${{ parameters.prefix }} python -m pip install thinc-apple-ops
+      ${{ parameters.prefix }} python -m pip install "thinc-apple-ops>=0.1.0.dev0"
       ${{ parameters.prefix }} python -m pytest --pyargs spacy
     displayName: "Run CPU tests with thinc-apple-ops"
     condition: and(startsWith(variables['imageName'], 'macos'), eq(variables['python.version'], '3.9'))
```
adrianeboyd, line 117: I think this is out-of-date:

```suggestion
    condition: and(startsWith(variables['imageName'], 'macos'), eq(variables['python.version'], '3.10'))
```

.github/azure-steps.yml
```diff
@@ -111,7 +111,7 @@ steps:
     condition: eq(variables['python_version'], '3.8')
 
   - script: |
-      ${{ parameters.prefix }} python -m pip install thinc-apple-ops
+      ${{ parameters.prefix }} python -m pip install "thinc-apple-ops>=0.1.0.dev0"
```
adrianeboyd, line 114: The only slightly more general option might be to use `--pre` here instead?

danieldk: I didn't know about that option. So, then this should be

```
      ${{ parameters.prefix }} python -m pip install --pre "thinc-apple-ops>=0.1.0"
```

?

adrianeboyd: Just

```
python -m pip install --pre thinc-apple-ops
```

adrianeboyd: I'm not sure if we'd have cases where we specifically wouldn't want `--pre`, but it's probably okay as a default here for this test?

danieldk: Yeah, I think for CI this should be completely safe. I've added a commit to use `--pre`.

PR #10911, opened 2022-06-03
precomputable_biaffine: avoid concatenation

danieldk:
## Description

The `forward` of `precomputable_biaffine` performs matrix multiplication and then `vstack`s the result with padding. This creates a temporary array used for the output of matrix concatenation.

This change avoids the temporary by pre-allocating an array that is large enough for the output of matrix multiplication plus padding and fills the array in-place.

This gave me a small speedup (a bit over 100 WPS) on de_core_news_lg on M1 Max (after changing thinc-apple-ops to support in-place gemm as BLIS does).

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/ml/_precomputable_affine.py
+++ b/spacy/ml/_precomputable_affine.py
@@ -22,9 +22,11 @@ def forward(model, X, is_train):
     nP = model.get_dim("nP")
     nI = model.get_dim("nI")
     W = model.get_param("W")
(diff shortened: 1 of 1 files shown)


shadeMe (approved):

svlandeg (commented):
Reviewing your PRs is like following a refresher course on efficient programming 🤓  Thanks Daniël! 🙏

PR #10916, opened 2022-06-05
Fix coref size inference

polm:
## Description

Currently coref works because it has the size of `roberta-base` hardcoded. Ideally we would be able to infer the shape from the tok2vec, but for Transformer layers, the size isn't available when the model object is created. However, if we declare sublayers in a Torch Module after it's created, then serialization stops working because it loses track of the sublayers. 

I hope there is a better way to do this, but at the moment the solution I have is to require specifying the tok2vec size in the config. This feels repetitive, but it's not complicated and it works. Additionally, if the size is wrong, Torch should give a helpful error immediately.

### Types of change

fix?

## Checklist

### DIFF
--- a/spacy/ml/models/coref.py
+++ b/spacy/ml/models/coref.py
@@ -19,19 +19,15 @@ def build_wl_coref_model(
     # pairs to keep per mention after rough scoring
     antecedent_limit: int = 50,
     antecedent_batch_size: int = 512,
+    tok2vec_size: int = 768,  # tok2vec size
 ):
     # TODO add model return types
-    # TODO fix this
-    try:
-        dim = tok2vec.get_dim("nO")
-    except ValueError:
-        # happens with transformer listener
-        dim = 768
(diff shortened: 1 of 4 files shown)


spacy/ml/models/coref.py
```diff
@@ -56,7 +52,7 @@ def convert_coref_clusterer_inputs(model: Model, X: List[Floats2d], is_train: bo
     def backprop(args: ArgsKwargs) -> List[Floats2d]:
         # convert to xp and wrap in list
         gradients = torch2xp(args.args[0])
-        assert isinstance(gradients, Floats2d)
+        # assert isinstance(gradients, Floats2d)
```
kadarakos, line 55: Why comment out the `assert`?

polm: Same reason as in this other PR - it's breaking with recent changes, though in fact the code is still fine. https://github.com/explosion/spaCy/pull/10875#discussion_r888757693

kadarakos (approved):

PR #10919, opened 2022-06-06
Add min_max operator in matcher {n,m}

### LINKED ISSUE #5603: Is it possible to support quantifier (n,m) for rule matching?
```
! | Negate the pattern, by requiring it to match exactly 0 times.
-- | --
? | Make the pattern optional, by allowing it to match 0 or 1 times.
+ | Require the pattern to match 1 or more times.
* | Allow the pattern to match zero or mo
```
In addition to the operators above, is it possible to support the {n,m} quantifier in the rule matcher? Or least {n} times. The reason is that, for linear matching, there is often a need to restrict how far two tokens occur in a piece of text. The shorter the distance, the more reliable the pattern. For example:

```
pattern = [{"LOWER": "hello"},
           {"LOWER": "world"}] 
```
For this pattern, I want to add an additional condition to require that the two words 'hello' and 'world' will be within a distance of no more than 12 words. Is there a way to express this logic for the current matcher? If not, how likely this can be supported by introducing the RE quantifier {n,m}? This is very useful for pattern matching in practical applications.

shen-qin:
## Description
Resolved #5603

Added min max quantifier {n,m}
Screenshot 2022-06-07 at 1 27 19 AM

Summary
1. Modified matcher.pyx to include min_max function {n,m} and its variants
2. Modified schemas.py to include min_max validation error
3. Added test cases to test_matcher_api.py, test_matcher_logic.py and test_pattern_validation.py
4. Modified API and Usage in spaCy website to include min_max operator

### Types of change
Min_max operator {n,m} for matcher

## Checklist


polm:
Thanks for the PR! As a note, it looks like the test failures are due to typing issues, and the actual tests pass. You can fix the typing issues, but if they don't make sense, don't worry, we can fix it.

It may take us a bit but we'll take a look at this.

shen-qin:
> Thanks for the PR! As a note, it looks like the test failures are due to typing issues, and the actual tests pass. You can fix the typing issues, but if they don't make sense, don't worry, we can fix it.
> 
> It may take us a bit but we'll take a look at this.


Hi polm, I have tried looking at the typing issue but I could not figure it out.

PR #10935, opened 2022-06-09
Rename some identifiers in the parser refactor

danieldk:
## Description

The naming of some identifiers came up in the reviews of #10878. This PR tries to address these issues:

* `lower`/`upper` are renamed to `hidden`/`output`.
* `_parseC` is renamed to `_parse_batch`.
* Some functions in `tb_framework` are prefixed by an underscore to make it more explicit that they are private.

### Types of change

Improve source code readability.

## Checklist

### DIFF
--- a/spacy/ml/tb_framework.pyx
+++ b/spacy/ml/tb_framework.pyx
@@ -188,15 +188,15 @@ def _forward_greedy_cpu(model: Model, TransitionSystem moves, states: List[State
     cdef int n_tokens = feats.shape[0] - 1
     sizes = get_c_sizes(model, c_states.size(), n_tokens)
     cdef CBlas cblas = model.ops.cblas()
-    scores = _parseC(cblas, moves, &c_states[0], weights, sizes, actions=actions)
+    scores = _parse_batch(cblas, moves, &c_states[0], weights, sizes, actions=actions)
 
     def backprop(dY):
         raise ValueError(Errors.E1038)
 
     return (states, scores), backprop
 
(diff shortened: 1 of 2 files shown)


spacy/ml/tb_framework.pyx
```diff
@@ -183,23 +183,23 @@ def _forward_greedy_cpu(model: Model, TransitionSystem moves, states: List[State
     for state in states:
         if not state.is_final():
             c_states.push_back(state.c)
-    weights = get_c_weights(model, <float*>feats.data, seen_mask)
+    weights = _get_c_weights(model, <float*>feats.data, seen_mask)
     # Precomputed features have rows for each token, plus one for padding.
     cdef int n_tokens = feats.shape[0] - 1
-    sizes = get_c_sizes(model, c_states.size(), n_tokens)
+    sizes = _get_c_sizes(model, c_states.size(), n_tokens)
```
shadeMe, line 189: What does the `c` in `_get_c_weights/sizes` signify? Is it just a shorthand for `class` (in which case, I think we can probably do away with it entirely)?

danieldk: This comes from the old parser, but I think `c_` here means: lives in C-land as opposed to Python-land.

danieldk: I do see an upside to the naming, since `tb_framework` contains both Thinc models (which we could consider being Python-land) and the fast C path.

svlandeg (approved):

honnibal:
Agree with this, thanks!

PR #10938, opened 2022-06-09
`StringStore`-related optimizations

shadeMe:
Assorted type-checking changes to avoid calling into 3rd party code unnecessarily.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/strings.pyx
+++ b/spacy/strings.pyx
@@ -3,6 +3,7 @@ cimport cython
 from libc.string cimport memcpy
 from libcpp.set cimport set
 from libc.stdint cimport uint32_t
+from libcpp cimport bool
 from murmurhash.mrmr cimport hash64, hash32
 
 import srsly
@@ -14,6 +15,13 @@ from .symbols import NAMES as SYMBOLS_BY_INT
 from .errors import Errors
 from . import util
 
+# Not particularly elegant, but this is faster than `isinstance(key, numbers.Integral)`
+cdef inline int try_coerce_to_hash(object key, hash_t* out_hash):
+    try:
+        out_hash[0] = key
+        return True
+    except:
+        return False
 
 def get_string_id(key):
     """Get a string ID, handling the reserved symbols correctly. If the key is
@@ -22,15 +30,28 @@ def get_string_id(key):
     This function optimises for convenience over performance, so shouldn't be
     used in tight loops.
     """
-    if not isinstance(key, str):
-        return key
-    elif key in SYMBOLS_BY_STR:
-        return SYMBOLS_BY_STR[key]
+    cdef hash_t hash    
+    if isinstance(key, str):
+        symbol = SYMBOLS_BY_STR.get(key, None)
+        if symbol is not None:
+            return symbol
+        else:
+            chars = key.encode("utf8")
+            return hash_utf8(chars, len(chars))
+    elif try_coerce_to_hash(key, &hash):
+        # Coerce the integral key to the expected primitive hash type.
+        # This ensures that custom/overloaded "primitive" data types
+        # such as those implemented by numpy are not inadvertently used 
+        # downsteam (as these are internally implemented as custom PyObjects 
+        # whose comparison operators can incur a significant overhead).
+        return hash
     elif not key:
         return 0
     else:
-        chars = key.encode("utf8")
-        return hash_utf8(chars, len(chars))
+        return key
+        raise ValueError(Errors.E1039.format(input=key,
+                                             expected_types=', '.join(['str', 'int']),
+                                             invalid_type=type(key)))
 
 
 cpdef hash_t hash_string(str string) except 0:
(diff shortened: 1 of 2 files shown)


spacy/strings.pyx
```diff
@@ -22,15 +30,28 @@ def get_string_id(key):
...
+        # such as those implemented by numpy are not inadvertently used 
+        # downsteam (as these are internally implemented as custom PyObjects 
+        # whose comparison operators can incur a significant overhead).
+        return hash
     elif not key:
         return 0
     else:
-        chars = key.encode("utf8")
-        return hash_utf8(chars, len(chars))
+        return key
```
danieldk, line 51:
```suggestion
```

Otherwise the exception is unreachable?

spacy/strings.pyx
```diff
@@ -110,28 +130,35 @@ cdef class StringStore:
...
+        elif try_coerce_to_hash(string_or_id, &id):
+            if id == 0:
+                return ""
+            elif id < len(SYMBOLS_BY_INT):
+                return SYMBOLS_BY_INT[id]
             else:
-                return decode_Utf8Str(utf8str)
+                utf8str = <Utf8Str*>self._map.get(id)
+        else:
+            # TODO: Raise an error instead
```
danieldk, line 155: Not in this PR? Does this change the API?

spacy/strings.pyx
```diff
@@ -174,30 +201,29 @@ cdef class StringStore:
...
+        """Check whether a string or ID is in the store.
 
-        string (str): The string to check.
+        string_or_id (str or int): The string to check.
         RETURNS (bool): Whether the store contains the string.
         """
-        cdef hash_t key
-        if isinstance(string, int) or isinstance(string, long):
-            if string == 0:
+        cdef hash_t id
```
danieldk, line 210: See above.

spacy/strings.pyx
```diff
@@ -174,30 +201,29 @@ cdef class StringStore:
...
+            elif string_or_id in SYMBOLS_BY_STR:
+                return True
+            id = hash_string(string_or_id)
+        elif try_coerce_to_hash(string_or_id, &id):
+            pass
         else:
-            string = string.encode("utf8")
-            key = hash_utf8(string, len(string))
-        if key < len(SYMBOLS_BY_INT):
+            # TODO: Raise an error instead
```
danieldk, line 220: Some question here, is it a TODO because otherwise the API would change?

danieldk (commented):
Mostly nitpicking

spacy/strings.pyx
shadeMe: Correct.

spacy/strings.pyx
```diff
@@ -14,6 +15,13 @@ from .symbols import NAMES as SYMBOLS_BY_INT
 from .errors import Errors
 from . import util
 
+# Not particularly elegant, but this is faster than `isinstance(key, numbers.Integral)`
+cdef inline bool try_coerce_to_hash(object key, hash_t* out_hash):
```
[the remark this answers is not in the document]
shadeMe: Good points; will change.

spacy/strings.pyx
danieldk: Maybe we should make a ticket for v4 to change this behavior?

shadeMe: Yeah, that's the plan (just haven't manage to do it yet 😅).

danieldk (approved):
Looks good to me, but could use a review from someone who is more familiar with `StringStore` internals.

spacy/strings.pyx
```diff
@@ -179,19 +179,22 @@ cdef class StringStore:
         string (str): The string to add.
         RETURNS (uint64): The string's hash value.
         """
+        cdef hash_t hash
```
polm, line 182: I think the reason the original code avoided this variable name is that `hash` is a builtin in Python. `hash` is the best name for it, and it should be harmless as a local here, but it's something to be aware of.

spacy/strings.pyx
```diff
@@ -14,6 +14,13 @@ from .symbols import NAMES as SYMBOLS_BY_INT
 from .errors import Errors
 from . import util
 
+# Not particularly elegant, but this is faster than `isinstance(key, numbers.Integral)`
```
polm, line 17: Is this faster under some conditions, like if you mostly don't hit the except block?

spacy/strings.pyx
```diff
@@ -110,28 +129,35 @@ cdef class StringStore:
...
+        cdef hash_t hash
+        cdef Utf8Str* utf8str = NULL
+
         if isinstance(string_or_id, str):
-            key = hash_string(string_or_id)
-            return key
+            if len(string_or_id) == 0:
+                return 0
+
+            symbol = SYMBOLS_BY_STR.get(string_or_id, None)
```
polm, line 139: The default is `None` here so that the `hash_string` call is only made if required, right? I would leave a comment about that.

spacy/strings.pyx
polm: Do we know where the no-error behavior is relied on, if anywhere?

polm (approved):
I have some minor comments but this basically looks good to me. You might want to change the variable name of `hash` since it's a builtin. (Also sorry it took me so long to get to this!)

spacy/strings.pyx
shadeMe: If I recall correctly, this version was always faster than the `isinstance` check (the former barely even showed up in the profile). I think it might have to do with how the `Integral` abstract base class is implemented.

spacy/strings.pyx
shadeMe: Correct; will do.

spacy/strings.pyx
polm: Ah OK, sounds good then.

polm (approved):
Looks good!

shadeMe:

shadeMe:
@explosion-bot please test_slow_gpu

adrianeboyd:
I don't understand why the slow GPU tests succeed and the slow tests fail while building? Is there some difference in the setup that leads to this? (Accidentally not testing the right branches or something like that?)

shadeMe:
> I don't understand why the slow GPU tests succeed and the slow tests fail while building? Is there some difference in the setup that leads to this? (Accidentally not testing the right branches or something like that?)

Yeah, I find that rather curious too. The error does point to some branch mismatch between the spacy and thinc codebases, but the version numbers ostensibly seem like they _ought_ to line up? I'll take a closer look at it.

shadeMe:
@explosion-bot please test_slow_gpu

PR #10940, opened 2022-06-09
Merge the parser refactor into `v4`

danieldk:
## Description

This is a large refactor of the parsing model, started by @honnibal. The refactor turns the paring model into a more canonical Thinc model. As a result, GPU training performance is much-improved, since backpropagation can be done for all states at once, rather than per parsing step.

### Types of change

Enhancement.

## Checklist


.github/azure-steps.yml
```diff
@@ -64,12 +64,13 @@ steps:
...
-      python -c "import spacy; nlp=spacy.load('ca_core_news_sm'); doc=nlp('test')"
-    displayName: 'Test download CLI'
-    condition: eq(variables['python_version'], '3.8')
+  # Disabled in parser refactor branch, requires spacy.TransitionBasedParser.v2.
+  #- script: |
+  #     python -m spacy download ca_core_news_sm
+  #     python -m spacy download ca_core_news_md
+  #     python -c "import spacy; nlp=spacy.load('ca_core_news_sm'); doc=nlp('test')"
+  #   displayName: 'Test download CLI'
+  #   condition: eq(variables['python_version'], '3.8')
```
shadeMe, line 73: Can we uncomment these lines here, or should we wait until this PR is merged?

spacy/tests/test_misc.py
```diff
@@ -81,32 +78,33 @@ def test_util_get_package_path(package):
...
+#     ids[1, 2] = 0
+#     ids[1, 1] = -1
+#     ids[1, 0] = -1
+#     dY[1] = 1
+#     ids[2, 0] = -1
+#     dY[2] = 5
+#     d_pad = _backprop_precomputable_affine_padding(model, dY, ids)
+#     assert d_pad[0, 0, 0, 0] == 6
+#     assert d_pad[0, 1, 0, 0] == 1
+#     assert d_pad[0, 2, 0, 0] == 0
```
shadeMe, line 107: Can this be uncommented?

spacy/tests/parser/test_parse.py
```diff
@@ -402,10 +460,11 @@ def test_overfitting_IO(pipe_name):
...
     "parser_config",
     [
-        # TransitionBasedParser V1
-        ({"@architectures": "spacy.TransitionBasedParser.v1", "tok2vec": DEFAULT_TOK2VEC_MODEL, "state_type": "parser", "extra_state_tokens": False, "hidden_width": 64, "maxout_pieces": 2, "use_upper": True}),
-        # TransitionBasedParser V2
-        ({"@architectures": "spacy.TransitionBasedParser.v2", "tok2vec": DEFAULT_TOK2VEC_MODEL, "state_type": "parser", "extra_state_tokens": False, "hidden_width": 64, "maxout_pieces": 2, "use_upper": True}),
+        # TransitionBasedParser V1, FIXME: re-enable
+        # ({"@architectures": "spacy.TransitionBasedParser.v1", "tok2vec": DEFAULT_TOK2VEC_MODEL, "state_type": "parser", "extra_state_tokens": False, "hidden_width": 64, "maxout_pieces": 2, "use_upper": True}),
+        # TransitionBasedParser V2, FIXME: re-enable
+        # ({"@architectures": "spacy.TransitionBasedParser.v2", "tok2vec": DEFAULT_TOK2VEC_MODEL, "state_type": "parser", "extra_state_tokens": False, "hidden_width": 64, "maxout_pieces": 2, "use_upper": True}),
```
shadeMe, line 466: Same question as above.

spacy/pipeline/_parser_internals/ner.pyx
```diff
@@ -385,9 +387,9 @@ cdef class Begin:
         elif st.B_(1).ent_iob == 3:
             # If the next word is B, we can't B now
             return False
-        elif st.B_(1).sent_start == 1:
-            # Don't allow entities to extend across sentence boundaries
-            return False
+        #elif st.B_(1).sent_start == 1:
+        #    # Don't allow entities to extend across sentence boundaries
+        #    return False
```
shadeMe, line 392: Can be removed?

spacy/pipeline/_parser_internals/ner.pyx
```diff
@@ -465,9 +467,9 @@ cdef class In:
                 # Otherwise, force acceptance, even if we're across a sentence
                 # boundary or the token is whitespace.
                 return True
-        elif st.B(1) != -1 and st.B_(1).sent_start == 1:
-            # Don't allow entities to extend across sentence boundaries
-            return False
+        #elif st.B(1) != -1 and st.B_(1).sent_start == 1:
+        #    # Don't allow entities to extend across sentence boundaries
+        #    return False
```
shadeMe, line 472: Same as above.

spacy/pipeline/_parser_internals/ner.pyx
danieldk: I think these need to be re-instated. So, I did so :). Thanks!

spacy/tests/test_misc.py
danieldk: `PrecomputableAffine` doesn't exist as a separate layer anymore, so I removed this.

spacy/tests/parser/test_add_label.py
```diff
@@ -123,6 +123,7 @@ def test_ner_labels_added_implicitly_on_predict():
     assert "D" in ner.labels
 
 
+@pytest.mark.skip(reason="Not yet supported")
```
shadeMe, line 126: We probably want to use `xfail` than `skip` for this test (with the reason set `"no beam parser yet"`), I think.

spacy/tests/parser/test_parse.py
```diff
@@ -355,6 +412,7 @@ def test_overfitting_IO(pipe_name):
         train_examples.append(Example.from_dict(nlp.make_doc(text), annotations))
         for dep in annotations.get("deps", []):
             parser.add_label(dep)
+    # train_examples = train_examples[:1]
```
shadeMe, line 415: Seems like a statement from a debugging session; can probably be removed.

spacy/tests/parser/test_ner.py
```diff
@@ -533,7 +534,7 @@ def test_overfitting_IO(use_upper):
     for i in range(50):
         losses = {}
         nlp.update(train_examples, sgd=optimizer, losses=losses)
-    assert losses["ner"] < 0.00001
+    assert losses["ner"] < 0.001
```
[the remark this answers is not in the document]
danieldk: Matt changed the threshold during development. Since the seed is fixed and the original threshold works again, I changed it back to `0.00001`.

spacy/tests/parser/test_parse.py
danieldk: Removed.

shadeMe:
I think we can now remove some of the commented out lines that reference the `v2` parser, seeing how we won't be porting it back to `spacy-legacy`?

danieldk:
> I think we can now remove some of the commented out lines that reference the `v2` parser, seeing how we won't be porting it back to `spacy-legacy`?

Removed all the references to the old versions.

adrianeboyd:
When did you make the decision not to have legacy support for `spacy.TransitionParser.v1` and `spacy.TransitionParser.v2`? This is extremely breaking.

We can deal with having the output can be slightly different given that the factory behavior is versioned on the spacy version (so `parser`+v3.3.0 may not produce the same output as `parser`+v3.4.0 with identical configs), but I think we should try extremely hard not to break existing v3 configs in v4.

danieldk:
> When did you make the decision not to have legacy support for `spacy.TransitionParser.v1` and `spacy.TransitionParser.v2`? This is extremely breaking.

First the idea was to bring back the older versions, possibly as part of `spacy-legacy`. IIRC we discussed this with Matt and he thought it'd be best to just remove the old version.

Since there are changes to the layout of the parameters, etc. it'd be hard to support the `v1`/`v2` versions with the new code base. So, we'll have to fork what is now in spaCy 3.

adrianeboyd:
I don't care about the model compatibility, just the config compatibility. So this continues to work:

```
[components.parser]
factory = "parser"
learn_tokens = false
min_action_freq = 30
moves = null
scorer = {"@scorers":"spacy.parser_scorer.v1"}
update_with_oracle_cut_size = 100

[components.parser.model]
@architectures = "spacy.TransitionBasedParser.v2"
state_type = "parser"
extra_state_tokens = false
hidden_width = 64
maxout_pieces = 2
use_upper = true
nO = null

[components.parser.model.tok2vec]
@architectures = "spacy.Tok2VecListener.v1"
width = ${components.tok2vec.model.encode:width}
upstream = "tok2vec"
```

svlandeg:
We had a similar situation for the `entity_linker`, where a new implementation of the component wasn't compatible anymore with older config versions. What we decided at the time is to keep the old code around in `spacy.pipeline.legacy` and to have a "hack" in the factory to decide when to use this legacy version: https://github.com/explosion/spaCy/blob/master/spacy/pipeline/entity_linker.py#L102

It's ugly, but does mean the old configs can still be used. I also imagine that the legacy parser would probably entail a lot more than just the one Python file like we have for the `entity_linker` 😬

danieldk:
> It's ugly, but does mean the old configs can still be used. I also imagine that the legacy parser would probably entail a lot more than just the one Python file like we have for the `entity_linker` 😬

I'll try and see how far we can get. I think configuration-wise the largest difference is that in the new parser we always use the upper layer. So I guess that means that if the user specifies `use_upper = false`, we have to emit a warning and use the upper layer anyway.

danieldk:
I have added back `spacy.TransitionParser.v2` and (besides adding it to the tests) tried training a model from a configuration generated using [the quickstart](https://spacy.io/usage/training#quickstart).

`spacy.TransitionParser.v1` lives in `spacy-legacy`, so I think we have to update that version in legacy later.

I am not sure what to do about:

* `spacy.PrecomputableAffine.v1`, since this is not a separate model anymore.
* `spacy.TransitionModel.v1`, the model and arguments changed a fair bit.

Though I think these are rarely used directly in spaCy model configurations?

spacy/pipeline/_parser_internals/search.pyx
```diff
@@ -0,0 +1,306 @@
...
+            all_probs = _softmax(p_scores + g_scores)
+            p_probs = all_probs[:len(p_scores)]
+            g_probs_all = all_probs[len(p_scores):]
+            g_probs = _softmax(g_scores)
+
+            self.cost = pred.loss
+            self.delta = d
+            self.p_hist = p_hist
+            self.g_hist = g_hist
+            # TODO: These variables are misnamed! These are the gradients of the loss.
```
svlandeg, line 288: adding a review comment here so we don't forget to address this

svlandeg:
> I am not sure what to do about:
> 
>     * `spacy.PrecomputableAffine.v1`, since this is not a separate model anymore.
> 
>     * `spacy.TransitionModel.v1`, the model and arguments changed a fair bit.
> 

`_precomputable_affine.py` could just be moved in its entirety to `spacy-legacy`, no? It doesn't have any internal dependencies I think?

Similarly, can't we just move the the code for `spacy.TransitionModel.v1` to legacy?

It would be good to start that PR for `spacy-legacy` while wrapping up this PR, then we can also bump legacy's version here.

spacy/tests/parser/test_parse.py
```diff
@@ -451,10 +512,9 @@ def test_distill():
 @pytest.mark.parametrize(
     "parser_config",
     [
-        # TransitionBasedParser V1
-        ({"@architectures": "spacy.TransitionBasedParser.v1", "tok2vec": DEFAULT_TOK2VEC_MODEL, "state_type": "parser", "extra_state_tokens": False, "hidden_width": 64, "maxout_pieces": 2, "use_upper": True}),
```
svlandeg, line 455: Did we make a decision on this - I thought we could just move the appropriate code to legacy and keep v1 as is?

svlandeg: The test in legacy itself is not exactly the same thing though right? Having the test also in spaCy ensures that the architecture is in fact available (from legacy) and behaves similar as the other architectures (i.e. is a drop-in). Or am I missing something?

danieldk: Ah, I see! I will try to add it back to the test.

svlandeg: Thanks! Basically the legacy/compat discussion is the main thing I wanted to understand fully before merging this in :-)

danieldk: I added back the test, but commented it out. What I think we need to do:

- Merge the refactor into v4.
- Do a v4 dev release.
- Create a `spacy-legacy` branch for v4.
- This PR against `spacy-legacy`: https://github.com/explosion/spacy-legacy/pull/36
- Do a `spacy-legacy` release.
- Update `v4` to require this legacy release and uncomment the test.

Tested locally and with the `spacy-legacy` PR mentioned above the test passes.

svlandeg: Fun! 😁 

Thanks for spelling this out Daniël, that sounds like a perfect plan!

danieldk: I'll convert it into a ticket, so that we don't forget to do it.

svlandeg (approved):
This has been a truly epic team effort! So great to finally be able to merge this in 🥳  Nice job, team!

PR #10944, opened 2022-06-10
Add option to disable pipes in EntityRuler

polm:
## Description

The EntityRuler uses reference to the calling `Language` internally to generate docs for matching. Depending on how the pipeline is configured, this can call pipes that aren't needed, resulting in unused work. This PR adds a `disable_pipes` option to the EntityRuler that allows specifying pipes that should be disabled when creating Docs internally. 

There are other ways this could be done. For example, we could try automatic inference of which pipes are needed, since pipes specify the attributes they set. But I think this approach is better, since we would want a manual override for any inference in any case.

Another thing to note is that pipes after the EntityRuler are disabled by default. Maybe it's enough if we make that more prominent in the documentation?

Currently this is just at a proposal stage, and these things still need doing:

- [ ] same feature in span ruler
- [ ] doc updates

### Types of change

enhancement

## Checklist

### DIFF
--- a/spacy/pipeline/entityruler.py
+++ b/spacy/pipeline/entityruler.py
@@ -87,6 +87,7 @@ class EntityRuler(Pipe):
         ent_id_sep: str = DEFAULT_ENT_ID_SEP,
         patterns: Optional[List[PatternType]] = None,
         scorer: Optional[Callable] = entity_ruler_score,
+        disable_pipes: Optional[List[str]] = [],
     ) -> None:
         """Initialize the entity ruler. If patterns are supplied here, they
         need to be a list of dictionaries with a `"label"` and `"pattern"`
@@ -128,6 +129,7 @@ class EntityRuler(Pipe):
         if patterns is not None:
             self.add_patterns(patterns)
         self.scorer = scorer
+        self.disable_pipes = disable_pipes
 
     def __len__(self) -> int:
         """The number of all patterns added to the entity ruler."""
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
It's a good point to raise - users are not necessarily realising the performance implication of moving the `entity_ruler` to a different place in the `pipeline`. I do think having to define the other pipes manually could be both error prone and tedious. Then again, automatically inferring might be error prone, too. One other alternative is to have the user define which pipeline components should run, but then dependencies (like on the `tok2vec`) might get forgotten.

Interested to hear what others think!

rmitsch:
I'd be in favor of having the user disable pipes. I think it might be unexpected behavior if we automatically infer pipes only in this context. Using includes instead of excludes - that's what you meant, @svlandeg? - might be bothersome to implement for other reasons, if you remember the include/exclude/enable/disable PR (although maybe that's different here).

adrianeboyd:
I mean more that if the user knows they need tags and they know they need `["tok2vec", "tagger"]` for this, then it's easier to use `source` to add the entity ruler to various pipelines (where you can't currently easily change config settings like this with `source`).

Then it doesn't matter whether there's `parser` or `ner` in the pipeline that they're adding the entity ruler to. It's not 100% foolproof in terms of which components might be providing tags (like whether you need `tok2vec` or where POS is coming from), but it's more flexible that specifying things to disable.

polm:
I didn't know about the script workaround, which I agree almost makes this unnecessary, but it is a good point that won't help when loading a model. 

I think that specifying pipes to enable is much clearer than ones to disable. Since #10784 has been merged we can use the behavior there already too.

svlandeg:
After internal discussion, we decided not to clutter up the config files with this setting, which we assume would only be relevant/useful for a limited number of users. 

Happy to reconsider at any point, and if we do decided we want this feature, we should make sure it works both for the `entity_ruler` as well as the `span_ruler`.

PR #10950, opened 2022-06-10
Docs: displaCy documentation - data types, `parse_{deps,ents,spans}`, spans example

pmbaumgartner:
Adds an example for the `spans` displaCy format as well as references to the (undocumented) `parse_{deps,ents,spans}` functions for converting docs to this format.

**Edit:** Expanding the scope of this PR to float the idea of using `TypedDict` to document the data expected when `manual=True`, since we do not have this documented anywhere. Will also include further documentation of the `parse_{deps,ents,spans}` functions [here](https://spacy.io/api/top-level#displacy).

### Types of change
Documentation

## Checklist


website/docs/usage/visualizers.md
```diff
@@ -389,6 +392,16 @@ in displaCy's format as a dictionary (instead of `Doc` objects).
 }
 ```
 
+```python
+### SPANS input with knowledge base links
+{
+    "text": "But Google is starting from behind.",
+    "spans": [{"start": 4, "end": 10, "start_token": 0, "end_token": 1, "label": "ORG", "kb_id": "Q95", "kb_url": "https://www.wikidata.org/entity/Q95"}],
```
svlandeg, line 399: Does it need the `start_token` and `end_token`? I guess I'm mainly asking whether the differences between the `ENT input` (above) and this `SPANS input` are due to fundamental differences in the spans/ents functionality?

Also, I wonder whether we should have the same example as for ents, or perhaps have an example with overlapping spans to highlight why `spans` would be useful.

pmbaumgartner: Let me look into it more. I basically took the output of `displacy.parse_spans` and went with that, but I'm not sure how those are used in the actual render function.

Also, agreed on the example, I'll try and come up with one that has overlaps.

pmbaumgartner:
It's not a hard dependency, but https://github.com/explosion/spaCy/pull/10994 is probably a prior PR we want before this because it fixes displaCy for spans. 

@svlandeg this is somewhat related, but maybe we also want to solve in this PR: do we feel that `displacy.parse_{deps,ents,spans}` functions, and their outputs, need documentation? I tried to look into your question and displaCy only uses the `start_token` and `end_token` attributes (which are confusingly called `start` and `end` for ents). There may be other things that use the character indices I could be missing. Either way, my main point is that not really clear what we mean when we say a user can pass data `in displaCy's format as a dictionary`, but "displaCy's format" is only documented through these examples (or if someone happens to stumble upon the `parse_x` functions).

svlandeg:
@pmbaumgartner : yes - that was indirectly the point of my question: it's unclear to a reader how to define these dictionaries and they'll end up trying to use the examples to extract the API design for the various options but that's not ideal.

Ideally we should probably document this more extensively over at https://spacy.io/api/top-level#displacy

pmbaumgartner:
@svlandeg This might be overkill and the beginning of a longer discussion about using `TypedDict` in more places across the repo (i.e. matcher rules), but I've gone ahead and pushed a commit that uses `TypedDict` to provide the schema for these dictionaries. I figure we might as well do things the "right" way rather than writing out the types in the docstrings and documentation files. Another advantage here is that people can use these types in their IDEs, and we can use these type objects to quickly and consistently generate the documentation for them, e.g.

```python
from wasabi import table
from typing import TypedDict  # import from spacy.compat in reality

DisplacyDepsArcs = TypedDict(
    "DisplacyDepsArcs", {"start": int, "end": int, "label": str, "dir": str}
)

print(table([(k, v.__name__) for k, v in DisplacyDepsArcs.__annotations__.items()], header=("KEY", "TYPE")))
```

```
KEY     TYPE
start   int 
end     int 
label   str 
dir     str 
```

Is it worth pursuing this further as a solution, or do we want something else for this?

pmbaumgartner:
We're scoping this one down to providing documentation on the dispaCy data structures.

website/docs/usage/visualizers.md
pmbaumgartner: Well, it's much easier to answer these questions after I worked on https://github.com/explosion/spaCy/pull/11092 + https://github.com/explosion/spaCy/pull/10994. 😄 

We only use `start_token` and `end_token`. However, there's an inconsistency in how we refer to the token start/end between NER and Spans. If we wanted to be consistent, I can think of a few options: 
1. Rename `start` and `end` in the displaCy NER parser /  data structure to be `start_token` and `end_token`.
2. Remove `start` and `end` from the span data structure, then rename current (`start_token`, `end_token`) to (`start`, `end`) for consistency with NER. If we ever need the character-level information from a span, this could create an issue.

pmbaumgartner:
Alright, this PR now just updates the documentation on using displacy's `parse` functions. This happens in two spots:

1. `usage/visualizers` - Updated to make reference to the now documenated `parse_{deps,ents,spans}` functions, as well as provide an example of how to use those functions to get data into displaCy's format.
2. `api/top-level` documents `parse_{deps,ents,spans}` with the correct arguments and returns values, as well as provides an example for each.

There was a minor code change in `spacy/displacy/__init__.py`, where I changed the name of the `orig_doc` argument in `parse_deps` to `doc` in order to be consistent.

As for documenting the data structures themselves, I have a feeling we should leave those out of the documentation for now. I think most users will use the `parse_` functions rather than manually create the data structures necessary. Internally, `render` uses these `parse` functions as well, so their use will guarantee that they're passing the correct data to `render` with `manual=True` if they've pre-parsed their docs. Another reason to avoid this for now is that, particularly in the case of the deps structure, its a bit of a nested data structure issue and I'm afraid we'd just end up with a lot of tables w/ arguments added to the documentation for this not oft-used data structure. I've documented these data structures types in a [gist](https://gist.github.com/pmbaumgartner/85b3a0256131e86fbdd78a8b4060d361) for now, so we can always come back to them later.

spacy/displacy/__init__.py
```diff
@@ -120,15 +120,14 @@ def app(environ, start_response):
     return [res]
 
 
-def parse_deps(orig_doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
+def parse_deps(doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
```
svlandeg, line 123: Unfortunately, this change would be breaking for users who'd call the function with `parse_deps(orig_doc=...)`, so we'll need to move this particular edit to `v4`. In the meantime, we could update the docstring to be consistent (and edit the typo "do parse" to "to parse")

website/docs/usage/visualizers.md
svlandeg: We should update the example here to contain only the information needed, which will at least highlight the difference/inconsistency. And it would be nice to have an example with overlapping spans to better illustrate the usefulness of this option.

If we'd want to revise the actual underlying data structures, it's probably best/safest to defer that to v4 - let me add a note to our internal backlog for that.

svlandeg (commented):
Looking good! Just had a few more comments.

website/docs/usage/visualizers.md
pmbaumgartner: Sounds good, I'll update to the [example we have in the visualizers documentation](https://spacy.io/usage/visualizers#span), because it's nice and short.

spacy/displacy/__init__.py
pmbaumgartner: Didn't think of that as a breaking change, good catch! I'll revert.

spacy/displacy/__init__.py
```diff
@@ -123,17 +123,18 @@ def app(environ, start_response):
 def parse_deps(orig_doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
     """Generate dependency parse in {'words': [], 'arcs': []} format.
 
-    doc (Doc): Document do parse.
+    orig_doc (Doc): Document to parse.
+    options (Dict[str, Any]): Dependency parse specific visualisation options.
     RETURNS (dict): Generated dependency parse keyed by words and arcs.
     """
-    doc = Doc(orig_doc.vocab).from_bytes(
+    orig_doc = Doc(orig_doc.vocab).from_bytes(
```
svlandeg, line 130: I don't think you needed to rewrite all of these references to `doc`?

pmbaumgartner: Whoops. That was my IDE trying to be smart (and me being dumb) with Replace Symbol, I'll undo.

website/docs/api/top-level.md
```diff
@@ -265,14 +265,81 @@ Render a dependency parse tree or named entity visualization.
 | Name        | Description                                                                                                                                                                            |
 | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `docs`      | Document(s) or span(s) to visualize. ~~Union[Iterable[Union[Doc, Span, dict]], Doc, Span, dict]~~                                                                                      |
-| `style`     | Visualization style,`"dep"`, `"ent"` or `"span"` <Tag variant="new">3.3</Tag>. Defaults to `"dep"`. ~~str~~                                                                                                                  |
+| `style`     | Visualization style,`"dep"`, `"ent"` or `"span"` <Tag variant="new">3.3</Tag>. Defaults to `"dep"`. ~~str~~                                                                            |
```
svlandeg, line 268:
```suggestion
| `style`     | Visualization style, `"dep"`, `"ent"` or `"span"` <Tag variant="new">3.3</Tag>. Defaults to `"dep"`. ~~str~~                                                                            |
```

svlandeg (approved):

PR #10960, opened 2022-06-13
account for NER labels with a hyphen in the name

svlandeg:
## Description
This came up when running `spacy debug data` on data from [AnEM](https://raw.githubusercontent.com/juand-r/entity-recognition-datasets/master/data/AnEM/CONLL-format/data/AnEM.train), which contains an entity label `"Multi-tissue_structure"`. The `ner` pipe would parse this as `"Multi"` and the `debug data` command would do the same internally, because both were using
```
label.split("-")[1]
```
instead of
```
label.split("-", 1)[1]
```
when stripping of the BILU prefix (e.g. `"B-"`)

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -38,6 +38,7 @@ from .underscore import Underscore, get_ext_args
 from ._retokenize import Retokenizer
 from ._serialize import ALL_ATTRS as DOCBIN_ALL_ATTRS
 from ..util import get_words_and_spaces
+from ..training import split_bilu_label
 
 DEF PADDING = 5
 
@@ -330,7 +331,7 @@ cdef class Doc:
                 else:
                     if len(ent) < 3 or ent[1] != "-":
                         raise ValueError(Errors.E177.format(tag=ent))
-                    ent_iob, ent_type = ent.split("-", 1)
+                    ent_iob, ent_type = split_bilu_label(ent)
                     if ent_iob not in iob_strings:
                         raise ValueError(Errors.E177.format(tag=ent))
                     ent_iob = iob_strings.index(ent_iob)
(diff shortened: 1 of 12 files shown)


spacy/training/iob_utils.py
```diff
@@ -218,6 +218,10 @@ def tags_to_entities(tags: Iterable[str]) -> List[Tuple[str, int, int]]:
     return entities
 
 
+def strip_bilu_prefix(label: str):
```
shadeMe, line 221: We might want to add the type annotation for the result too, just for the sake of completeness.

shadeMe (approved):

spacy/training/iob_utils.py
svlandeg, line 221:
```suggestion
def strip_bilu_prefix(label: str) -> str:
```

spacy/tokens/doc.pyx
```diff
@@ -38,6 +38,7 @@ from .underscore import Underscore, get_ext_args
 from ._retokenize import Retokenizer
 from ._serialize import ALL_ATTRS as DOCBIN_ALL_ATTRS
 from ..util import get_words_and_spaces
+from ..training import split_bilu_label
```
svlandeg, line 41:
```suggestion
```

spacy/tokens/doc.pyx
```diff
@@ -330,7 +331,7 @@ cdef class Doc:
                 else:
                     if len(ent) < 3 or ent[1] != "-":
                         raise ValueError(Errors.E177.format(tag=ent))
-                    ent_iob, ent_type = ent.split("-", 1)
+                    ent_iob, ent_type = split_bilu_label(ent)
```
svlandeg, line 334:
```suggestion
                    ent_iob, ent_type = ent.split("-", 1)
```

PR #10965, opened 2022-06-14
Add NVTX ranges to `TrainablePipe` components

shadeMe:
## Description

Adds a callback-based wrapper for `Pipe`/`TrainablePipe` components to annotate their methods in nVidia Nsight profiles.

Screenshot_20220614_152739

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/ml/callbacks.py
+++ b/spacy/ml/callbacks.py
@@ -3,12 +3,12 @@ import functools
 import inspect
 import types
 import warnings
-from ..errors import Warnings
 
 from thinc.layers import with_nvtx_range
 from thinc.model import Model, wrap_model_recursive
 from thinc.util import use_nvtx_range
 
+from ..errors import Warnings
 from ..util import registry
 
 if TYPE_CHECKING:
@@ -16,7 +16,7 @@ if TYPE_CHECKING:
     from ..language import Language  # noqa: F401
 
 
-DEFAULT_ANNOTATABLE_PIPE_METHODS = [
+DEFAULT_NVTX_ANNOTATABLE_PIPE_METHODS = [
     "pipe",
     "predict",
     "set_annotations",
@@ -81,7 +81,7 @@ def pipes_with_nvtx_range(
         else:
             extra_funcs = []
 
-        for name in DEFAULT_ANNOTATABLE_PIPE_METHODS + extra_funcs:
+        for name in DEFAULT_NVTX_ANNOTATABLE_PIPE_METHODS + extra_funcs:
             func = getattr(pipe, name, None)
             if func is None:
                 if name in extra_funcs:
--- a/spacy/ml/callbacks.py
+++ b/spacy/ml/callbacks.py
@@ -1,9 +1,9 @@
+from typing import Type, Callable, Dict, TYPE_CHECKING, List, Optional
 import functools
 import inspect
 import types
-from typing import Type, Callable, Dict, TYPE_CHECKING, List, Optional
 import warnings
-from spacy.errors import Warnings
+from ..errors import Warnings
 
 from thinc.layers import with_nvtx_range
 from thinc.model import Model, wrap_model_recursive
@@ -37,7 +37,7 @@ def models_with_nvtx_range(nlp, forward_color: int, backprop_color: int):
         if hasattr(pipe, "is_trainable") and pipe.is_trainable
     ]
 
-    # We need process all models jointly to avoid wrapping callbacks twice.
+    # We need to process all models jointly to avoid wrapping callbacks twice.
     models: Model = Model(
         "wrap_with_nvtx_range",
         forward=lambda model, X, is_train: ...,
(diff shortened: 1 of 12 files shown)


shadeMe:
Both good points! I'll rewrite it as a configurable callback (with the option of passing additional function names to annotate).

spacy/ml/callbacks.py
```diff
@@ -1,8 +1,13 @@
-from functools import partial
-from typing import Type, Callable, TYPE_CHECKING
+import functools
+import inspect
+import types
+from typing import Type, Callable, Dict, TYPE_CHECKING, List, Optional
```
svlandeg, line 4: We generally put `from typing` statements on the very first line of the file

spacy/ml/callbacks.py
```diff
@@ -1,8 +1,13 @@
...
+import warnings
+from spacy.errors import Warnings
```
svlandeg, line 6: This should be a relative import, right?

spacy/ml/callbacks.py
```diff
@@ -11,29 +16,109 @@
...
+
+
+def models_with_nvtx_range(nlp, forward_color: int, backprop_color: int):
+    pipes = [
+        pipe
+        for _, pipe in nlp.components
+        if hasattr(pipe, "is_trainable") and pipe.is_trainable
+    ]
+
+    # We need process all models jointly to avoid wrapping callbacks twice.
```
svlandeg, line 40:
```suggestion
    # We need to process all models jointly to avoid wrapping callbacks twice.
```

spacy/ml/callbacks.py
```diff
@@ -11,29 +16,109 @@
...
+    models: Model = Model(
+        "wrap_with_nvtx_range",
+        forward=lambda model, X, is_train: ...,
+        layers=[pipe.model for pipe in pipes],
+    )
```
svlandeg, line 45: This feels slightly hacky to me, to make a `Model` just to be able to use the `walk` method. Can't we keep a `Set[int]` of seen IDs across models just like the `walk` method would do, but without the additional overhead of creating an actual artificial `Model`?

spacy/ml/callbacks.py
```diff
@@ -11,29 +16,109 @@
...
+        if additional_pipe_functions:
+            extra_funcs = additional_pipe_functions.get(pipe.name, [])
+        else:
+            extra_funcs = []
+
+        for name in DEFAULT_ANNOTATABLE_PIPE_METHODS + extra_funcs:
+            func = getattr(pipe, name, None)
+            if func is None:
+                if name not in DEFAULT_ANNOTATABLE_PIPE_METHODS:
+                    warnings.warn(Warnings.W121.format(method=name, pipe=pipe.name))
```
svlandeg, line 88: Considering that this is a user warning about the input they provide, I would specifically check 
```
if name in additional_pipe_functions:
```
It would usually boil down to the same thing, but the code feels more explicit this way. Also, if the user specified a function that is in `DEFAULT_ANNOTATABLE_PIPE_METHODS` but somehow not found, we'd also want to warn about that.

spacy/ml/callbacks.py
```diff
@@ -11,29 +16,109 @@
...
+                    name,
+                    wrapped_func,
+                )
+            except AttributeError:
+                warnings.warn(Warnings.W122.format(method=name, pipe=pipe.name))
+
+    return nlp
+
+
+@registry.callbacks("spacy.models_and_pipes_with_nvtx_range.v1")
```
svlandeg, line 113: I do think in this case it makes sense to have the ensemble baked in ;-)

svlandeg (commented):
I really like the current design of this! Had a few smaller comments but in general this will be awesome to have!

spacy/ml/callbacks.py
shadeMe: This code wasn't originally introduced in this PR - I just had to move it out of the original callback function to accommodate the new ensemble callback.

Seems like the usage of `Model` is intended, according to [this commit](https://github.com/explosion/spaCy/pull/9124/commits/071d930bf222a9768a903ff0bbacb5411c68222d).

svlandeg: Before we had 
```
models = wrap_model_recursive(models, with_nvtx_range)
```
with `wrap_model_recursive` calling 
```
    for node in list(models.walk()):
        models.replace_node(node, wrapper(node))
```
Now, we call
```
    for node in models.walk():
        with_nvtx_range(
            node, forward_color=forward_color, backprop_color=backprop_color
        )
```
directly, using the artificial `models` of type `Model` only to use the `walk` method, which is nothing more than
```
        queue = [self]
        seen: Set[int] = set()
        for node in queue:
            if id(node) in seen:
                continue
            seen.add(id(node))
            yield node
            queue.extend(node.layers)
```
So basically what I'm saying is, can't we do this more directly, without the artificial `Model`, walk over all the nodes in the different `pipe.model for pipe in pipes`, keep a set of `seen` IDs and avoid calling the  `with_nvtx_range` method if that layer has already been seen?

I'm proposing this only because there's a `Model` constructor used here but we never really a `Model` - we just want a convenient way to iterate over nodes without duplication, right?

shadeMe: Ah, gotcha! Dunno why I got confused about that :sweat_smile: Yeah, you're right - we can get rid of that wrapper instance of `Model`.

svlandeg (approved):
LGTM! I'll just leave this unmerged for a bit in case Daniël wanted to re-review.

danieldk (approved):

PR #10978, opened 2022-06-17
Make KB entities/vector/aliases initialization methods private 

rmitsch:
## Goal

Make KB initialization methods for entities/vectors/aliases private.

## Description

Refactoring `initialize_entities()`, `initialize_vectors()`, `initialize_aliases()` by adding a `_` prefix.l

### Types of change

Refactoring.

## Checklist

### DIFF
--- a/spacy/kb.pyx
+++ b/spacy/kb.pyx
@@ -93,14 +93,14 @@ cdef class KnowledgeBase:
         self.vocab = vocab
         self._create_empty_vectors(dummy_hash=self.vocab.strings[""])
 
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks!

In principle this could be a breaking change, but these methods have never been [documented](https://spacy.io/api/kb) so I think it shouldn't affect anyone as the recommended way to set a bunch of entities is through [`set_entities`](https://spacy.io/api/kb#set_entities), which calls these methods internally.

PR #10981, opened 2022-06-18
Addition of min_max quantifier in matcher {n,m}

### LINKED ISSUE #5603: Is it possible to support quantifier (n,m) for rule matching?

shen-qin:
## Description

Reupload of #10919 
Resolved #5603 

Added min max quantifier {n,m}
image

Summary
1. Modified API and Usage for spaCy website to include min_max operator
2. Modified matcher.pyx to include min_max function {n,m} and its variants
3. Modified schemas.py to include min_max validation error
4. Added test cases to test_matcher_api.py, test_matcher_logic.py and test_pattern_validation.py

### Types of change

Min_max operator {n,m} for matcher
## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -86,10 +86,14 @@ cdef class Matcher:
         is a dictionary mapping attribute IDs to values, and optionally a
         quantifier operator under the key "op". The available quantifiers are:
 
-        '!': Negate the pattern, by requiring it to match exactly 0 times.
-        '?': Make the pattern optional, by allowing it to match 0 or 1 times.
-        '+': Require the pattern to match 1 or more times.
-        '*': Allow the pattern to zero or more times.
+        '!':      Negate the pattern, by requiring it to match exactly 0 times.
+        '?':      Make the pattern optional, by allowing it to match 0 or 1 times.
+        '+':      Require the pattern to match 1 or more times.
+        '*':      Allow the pattern to zero or more times.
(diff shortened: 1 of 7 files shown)


svlandeg:
Researched the `mypy` failure a bit and updated the branch directly according to [this](https://stackoverflow.com/a/67871116/7961860) advice from StackOverflow.

adrianeboyd (changes requested):
Very nice PR!

I tried and failed to break the validation. I added some additional bad cases and a suggestion to move these test cases to the section of the list for cases that are always flagged (nothing is technically incorrect).

adrianeboyd (approved):

shen-qin:
Nice! Thank you :D

PR #10992, opened 2022-06-20
vectors: avoid expensive comparisons between numpy ints and Python ints

danieldk:
## Description

Avoid expensive comparisons between numpy ints and Python ints when mapping hashes to rows in the embedding table. When doing some repeated benchmarks, this seems to improve performance by 800-900WPS on an 5950X.

### Types of change

Performance improvement.

## Checklist

### DIFF
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -336,7 +336,7 @@ cdef class Vectors:
         xp = get_array_module(self.data)
         if key is not None:
             key = get_string_id(key)
-            return self.key2row.get(key, -1)
+            return self.key2row.get(int(key), -1)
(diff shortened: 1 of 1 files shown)


svlandeg:
@danieldk : could you have a look into the failing test?

spacy/vectors.pyx
```diff
@@ -339,7 +339,7 @@ cdef class Vectors:
             return self.key2row.get(key, -1)
```
adrianeboyd, line 339: I realize it's only one key, but doesn't it make sense to make the same change here, too?

danieldk: Done.

PR #11002, opened 2022-06-22
Store activations in `Doc`s when `save_activations` is enabled

danieldk:
## Description

This change adds the new `activations` attribute to `Doc`. This attribute can be used by trainable pipes to store their activations, probabilities, and guesses for downstream users.

As an example, this change modifies the `tagger` and `senter` pipes to add an `store_activations` option. When this option is enabled, the probabilities and guesses are stored in `set_annotations`.

**Edit:** out of draft.

### Types of change

Proposal.

## Checklist


danieldk:
> Adding a field to `Doc` is breaking, so I think this would need to wait until v4?

Yeah. I was going to rebase this to `v4`, but we first need to update `v4` to avoid that all the commits end up in this PR:

https://github.com/explosion/spaCy/pull/11034

danieldk:
Something went awry with the merge, will fix this later.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -186,8 +197,12 @@ def _scores2guesses(self, docs, scores):
 
         return guesses
 
-    def set_annotations(self, docs: Iterable[Doc], batch_tree_ids):
+    def set_annotations(self, docs: Iterable[Doc], activations: ActivationsT):
+        batch_tree_ids = activations["guesses"]
         for i, doc in enumerate(docs):
+            doc.activations[self.name] = {}
+            for activation in self.store_activations:
```
shadeMe, line 204: Don't we need an `isinstance(self.store_activations, bool)` check here (and in similar locations in the other files)?

danieldk: `store_activations` is a property in `TrainablePipe`, we ensure in the setter that we have the correct data structure:

https://github.com/explosion/spaCy/blob/df513beddb9bc2e2407c15bd0d083df11f56f6bb/spacy/pipeline/trainable_pipe.pyx#L351

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -91,6 +97,7 @@ def __init__(
         overwrite: bool = False,
         top_k: int = 1,
         scorer: Optional[Callable] = lemmatizer_score,
+        store_activations=False,
```
svlandeg, line 100:
```suggestion
        store_activations: Union[bool, List[str]] = False,
```
Also needs to be documented in the doc string

svlandeg (commented):
Nice idea! 

Left a first batch of some comments.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -116,6 +123,7 @@ def __init__(
 
         self.cfg: Dict[str, Any] = {"labels": []}
         self.scorer = scorer
+        self.store_activations = store_activations  # type: ignore
```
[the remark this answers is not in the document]
danieldk: `store_activations` is a property of `TrainablePipe`. The setter guarantees that we always get `List[str]` when we use the getter. So, `for activation in self.store_activations` is safe, because `self.store_activations` will never return a `bool`.

The issue is that we MyPy is currently not able to handle setters with a different type of argument than the property getter and therefore the `type: ignore` is needed here. It's a long thread, but upstream now wants to support this use case, but not one has implemented it yet.

https://github.com/python/mypy/issues/3004

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -21,6 +21,9 @@
 from .. import util
 
 
+ActivationsT = Dict[str, Union[List[Floats2d], List[Ints1d]]]
```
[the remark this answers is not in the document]
danieldk: But this is only the type used by the `predict` method of the edit tree lemmatizer? Doesn't it make more sense to use specific types here?

spacy/errors.py
```diff
@@ -934,6 +935,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1041 = ("Expected a string, Doc, or bytes as input, but got: {type}")
     E1042 = ("Function was called with `{arg1}`={arg1_values} and "
              "`{arg2}`={arg2_values} but these arguments are conflicting.")
+    E1043 = ("store_activations attribute must be set to List[str] or bool")
```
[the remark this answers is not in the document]
danieldk: Fixed.

spacy/pipeline/edit_tree_lemmatizer.py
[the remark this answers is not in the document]
svlandeg: Aaah, right, I hadn't seen the setter bit, thanks!

danieldk:
Fixed merge conflicts.

spacy/pipeline/entity_linker.py
```diff
@@ -598,3 +645,30 @@ def rehearse(self, examples, *, sgd=None, losses=None, **config):
...
+        docs_scores.append(
+            Ragged(ops.flatten(doc_scores), ops.asarray1i(doc_scores_lens))
+        )
+        docs_ents.append(
+            Ragged(
+                ops.flatten(doc_ents, dtype="uint64"), ops.asarray1i(doc_scores_lens)
+            )
+        )
+
+    def _add_activations(self, doc_scores, doc_scores_lens, doc_ents, scores, ents):
```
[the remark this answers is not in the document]
svlandeg: Or maybe we should work with some small, specific datastructures to make this code less error-prone? To be honest I'm currently finding it quite hard to follow what exactly is being stored when and how throughout the entity linker's code.

spacy/pipeline/edit_tree_lemmatizer.py
svlandeg: (there's some other places throughout the PR where I think the type could be added similarly)

spacy/pipeline/spancat.py
```diff
@@ -290,7 +298,7 @@ def set_candidates(
             for index in candidates.dataXd:
                 doc.spans[candidates_key].append(doc[index[0] : index[1]])
 
-    def set_annotations(self, docs: Iterable[Doc], indices_scores) -> None:
+    def set_annotations(self, docs: Iterable[Doc], activations) -> None:
```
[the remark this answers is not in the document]
svlandeg: Also needs to be updated in the docstring few lines below this (I can't make a review comment on that line)

spacy/pipeline/tagger.pyx
```diff
@@ -155,21 +167,25 @@ class Tagger(TrainablePipe):
             guesses.append(doc_guesses)
         return guesses
 
-    def set_annotations(self, docs, batch_tag_ids):
+    def set_annotations(self, docs, activations):
```
svlandeg, line 170:
```suggestion
    def set_annotations(self, docs: Iterable[Doc], activations: Dict):
```

spacy/pipeline/tagger.pyx
```diff
@@ -155,21 +167,25 @@ class Tagger(TrainablePipe):
...
         """Modify a batch of documents, using pre-computed scores.
 
         docs (Iterable[Doc]): The documents to modify.
         batch_tag_ids: The IDs to set, produced by Tagger.predict.
```
svlandeg, line 174:
```suggestion
        activations (Dict): The activations used for setting annotations, produced by Tagger.predict.
```

spacy/pipeline/textcat.py
```diff
@@ -208,6 +221,9 @@ def set_annotations(self, docs: Iterable[Doc], scores) -> None:
         DOCS: https://spacy.io/api/textcategorizer#set_annotations
```
svlandeg, line 221: Even though the `textcat` only has one relevant field here, I would make the signature consistent and have `predict` return a dict, and `set_annotations` take that dict instead of the raw `scores`.

spacy/pipeline/senter.pyx
```diff
@@ -107,19 +121,23 @@ class SentenceRecognizer(Tagger):
     def label_data(self):
         return None
 
-    def set_annotations(self, docs, batch_tag_ids):
+    def set_annotations(self, docs, activations):
```
svlandeg, line 124:
```suggestion
    def set_annotations(self, docs: Iterable[Doc], activations: Dict):
```

spacy/pipeline/senter.pyx
```diff
@@ -107,19 +121,23 @@ class SentenceRecognizer(Tagger):
...
         """Modify a batch of documents, using pre-computed scores.
 
         docs (Iterable[Doc]): The documents to modify.
         batch_tag_ids: The IDs to set, produced by SentenceRecognizer.predict.
```
svlandeg, line 128:
```suggestion
        activations (Dict): The activations used for setting annotations, produced by SentenceRecognizer.predict.
```

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -1172,3 +1173,79 @@ def create_kb(vocab):
 
     assert len(doc.ents) == 1
     assert doc.ents[0].kb_id_ == entity_id if meet_threshold else EntityLinker.NIL
+
+
+def test_store_activations():
```
svlandeg, line 1178: This feels like mainly copy/paste from another test, is that right? If so, can we expand on the other one instead?

spacy/tests/pipeline/test_morphologizer.py
```diff
@@ -197,3 +199,29 @@ def test_overfitting_IO():
     gold_pos_tags = ["NOUN", "NOUN", "NOUN", "NOUN"]
     assert [str(t.morph) for t in doc] == gold_morphs
     assert [t.pos_ for t in doc] == gold_pos_tags
+
+
+def test_store_activations():
+    # Simple test to try and quickly overfit the morphologizer - ensuring the ML models work correctly
```
svlandeg, line 205: The docstring here is copied from another test but should read something else (also for other tests of other components)

website/docs/api/doc.md
```diff
@@ -751,22 +751,23 @@ The L2 norm of the document's vector representation.
...
+| `user_hooks`                         | A dictionary that allows customization of the `Doc`'s properties. ~~Dict[str, Callable]~~                                                       |
+| `user_token_hooks`                   | A dictionary that allows customization of properties of `Token` children. ~~Dict[str, Callable]~~                                               |
+| `user_span_hooks`                    | A dictionary that allows customization of properties of `Span` children. ~~Dict[str, Callable]~~                                                |
+| `has_unknown_spaces`                 | Whether the document was constructed without known spacing between tokens (typically when created from gold tokenization). ~~bool~~             |
+| `_`                                  | User space for adding custom [attribute extensions](/usage/processing-pipelines#custom-components-attributes). ~~Underscore~~                   |
+| `activations`                        | A dictionary of activations per trainable pipe (available when the `store_activations` option of a pipe is enabled). ~~Dict[str, Option[Any]]~~ |
```
svlandeg, line 770: add "new" tag for 4.0

svlandeg: Also for the other updates to the docs :-)

website/docs/api/edittreelemmatizer.md
```diff
@@ -44,14 +44,15 @@ architectures and their arguments and hyperparameters.
...
+| `top_k`             | The number of most probable edit trees to try before resorting to `backoff`. Defaults to `1`. ~~int~~                                                                                                                                                                                                              |
+| `scorer`            | The scoring method. Defaults to [`Scorer.score_token_attr`](/api/scorer#score_token_attr) for the attribute `"lemma"`. ~~Optional[Callable]~~                                                                                                                                                                      |
+| `store_activations` | Store activations in `Doc` when annotating. Supported activations are `"probs"` and `"guesses"`. ~~Union[bool, list[str]]~~                                                                                                                                                                                        |
```
svlandeg, line 55: Alternatively, we could also say that the available/supported activations are defined by the `activations` property on the component. I would consider doing that if we envision this list to change. If we think it'll remain roughly the same for all the components, perhaps the current direct solution is more clear. (I'm just worrying about maintenance burden)

svlandeg (commented):
This will be really good to have in v4. I've gone through the PR in detail and left some comments. Most of them minor things.

spacy/pipeline/edit_tree_lemmatizer.py
danieldk: The issue is that MyPy then infers that `store_activations` is a `Union[bool, List[str]]`. But maybe I should get rid of this setter and replace it by a `set_store_activations` method.

danieldk: Replaced all setter uses by a `set_store_activations` method and added the type to the `__init__` signatures.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -186,8 +199,12 @@ def _scores2guesses(self, docs, scores):
 
         return guesses
 
-    def set_annotations(self, docs: Iterable[Doc], batch_tree_ids):
+    def set_annotations(self, docs: Iterable[Doc], activations: ActivationsT):
+        batch_tree_ids = activations["guesses"]
         for i, doc in enumerate(docs):
+            doc.activations[self.name] = {}
+            for activation in self.store_activations:
+                doc.activations[self.name][activation] = activations[activation][i]
```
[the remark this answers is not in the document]
danieldk: Fixed.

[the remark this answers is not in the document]
danieldk: I reverted the change. With the dict comprehension mypy infers that the type is `Union[Floats2d, Ints1d]`, unfortunately it's not able to widen it to the necessary `Union[ArrayXd, Ragged]]` in the assignment to `doc.activations`. I think my original code passes the test because the initial

```
doc.activations[self.name] = {}
```

makes MyPy infer the dictionary type as `Union[ArrayXd, Ragged]]`.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -377,3 +394,7 @@ def _pair2label(self, form, lemma, add_label=False):
             self.tree2label[tree_id] = len(self.cfg["labels"])
             self.cfg["labels"].append(tree_id)
         return self.tree2label[tree_id]
+
+    @property
+    def activations(self):
+        return ["probs", "guesses"]
```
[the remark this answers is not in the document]
danieldk: We didn't and we should :).

adrianeboyd:
I think that `set_store_activations` is a really confusing name. Are you setting? Are you storing? Are you setting "store activations"? What are "store activations"? Are they "stored activations"?

spacy/pipeline/entity_linker.py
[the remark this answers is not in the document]
danieldk: I have added type annotations to `_add_activations` and `_add_doc_activations`, this also required adding type signatures in the body of `set_annotations`, so I think now it's at least easier to understand.

Tests will probably fail until we have https://github.com/explosion/thinc/pull/739 though :(.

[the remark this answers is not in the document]
danieldk: Also made these methods kwarg-only as you suggested.

spacy/pipeline/entity_linker.py
```diff
@@ -410,13 +422,21 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
         if not docs:
-            return final_kb_ids
+            return {"kb_ids": final_kb_ids, "ents": docs_ents, "scores": docs_scores}
         if isinstance(docs, Doc):
             docs = [docs]
-        for i, doc in enumerate(docs):
+        for doc in docs:
+            doc_ents = []
+            doc_scores = []
+            doc_scores_lens: List[int] = []
```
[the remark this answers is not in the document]
danieldk: Removed this, since `_add_doc_activations` can get the score/entity array lengths itself.

svlandeg:
> I think that `set_store_activations` is a really confusing name. Are you setting? Are you storing? Are you setting "store activations"? What are "store activations"? Are they "stored activations"?

😁 

`set_persisten_activations` and maybe use the name "persistent" in other places as well?

danieldk:
> `set_persistent_activations` and maybe use the name "persistent" in other places as well?

`set_persisted_activations` or maybe `set_saved_activations`?

spacy/pipeline/morphologizer.pyx
```diff
@@ -217,14 +229,15 @@ class Morphologizer(Tagger):
         assert len(label_sample) > 0, Errors.E923.format(name=self.name)
         self.model.initialize(X=doc_sample, Y=label_sample)
 
-    def set_annotations(self, docs, batch_tag_ids):
+    def set_annotations(self, docs, activations):
```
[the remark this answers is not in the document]
danieldk: Fixed. Made the `Dict` more specific.

spacy/pipeline/entity_linker.py
```diff
@@ -479,27 +508,45 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
         """Modify a batch of documents, using pre-computed scores.
 
         docs (Iterable[Doc]): The documents to modify.
-        kb_ids (List[str]): The IDs to set, produced by EntityLinker.predict.
+        activations (List[str]): The activations used for setting annotations, produced
+                                 by EntityLinker.predict.
 
         DOCS: https://spacy.io/api/entitylinker#set_annotations
         """
+        kb_ids = cast(List[str], activations["kb_ids"])
```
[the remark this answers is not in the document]
danieldk: We can't do `asserts` on generic types as far as I understand? (I would require traversing the data structure to validate it.)

spacy/pipeline/textcat.py
danieldk: Fixed.

spacy/pipeline/edit_tree_lemmatizer.py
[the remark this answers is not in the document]
adrianeboyd: I have to say that the `List[str]` option and the `predict`/`set_annotations` dict format does not really ... inspire confidence about the long-term stability of this setting.

adrianeboyd:
Isn't it more along the lines of `set_activations_to_store` or `set_activations_to_save` or `set_store_activations_setting`?

spacy/pipeline/edit_tree_lemmatizer.py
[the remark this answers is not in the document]
svlandeg: I have a similar gut feeling, but not sure what a better alternative is. We could have some sort of generic `Enum` storing potential values across all components, even though that will contain a few highly component-specific ones?

spacy/pipeline/entity_linker.py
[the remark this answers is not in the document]
svlandeg: Oh :'(

spacy/pipeline/edit_tree_lemmatizer.py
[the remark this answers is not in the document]
danieldk: Also, a 'sealed' enum makes it harder for community to provide activations for their own pipe. Maybe a per-pipe enum? All the tagger-based components could share the same enum.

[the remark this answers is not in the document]
adrianeboyd: How bad would it be for this setting to just be `bool` instead so that we don't have to expose this?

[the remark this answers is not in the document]
svlandeg: Hm, so basically an all-or-nothing and let the users iterate over the dict to get the keys instead? It's an option - like how often would people actually go in and adjust which activations would be stored...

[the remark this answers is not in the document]
danieldk: The `Union[bool, List[str]]` came out of an earlier discussion. One of the benefits of giving the user the option of what to store is that particular activations for some pipes may be large arrays and would only be garbage collected when the `Doc` is deallocated (or the user explicitly deletes the corresponding keys). This could be a waste if they need another activation which is much smaller.

I don't think the size is an issue for our current components yet. `probs` is larger than `guesses`, but I think when people want activations, they want the larger `probs` anyway (since they are already getting the classes).

I think we could make it a `bool` now. If it turns out to be too limiting, we could always make it `Union[bool, List[str]]` or `Union[bool, List[SomeEnumType]]` later.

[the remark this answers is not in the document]
adrianeboyd: I think trying to make this too configurable is making it too complicated and hard to maintain. It's so much easier if this stays an internal implementation detail.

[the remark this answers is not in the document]
svlandeg: Yea, I know we discussed this earlier but it does make things quite a bit more complex in the code + some worries about maintainability/stability. So it's good to consider the trade-off to how often users would actually need to customize in more detail than on/off. 

Fine by me to have it as a boolean for now - it would also mean that existing configs wouldn't break if we expand to `Union[bool, List[str]]` later.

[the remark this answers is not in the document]
danieldk: I still have to rename the property, but it is a `bool` now.

spacy/pipeline/spancat.py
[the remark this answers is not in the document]
danieldk: Done.

spacy/pipeline/tagger.pyx
danieldk: Fixed.

spacy/pipeline/senter.pyx
danieldk: Fixed.

danieldk:
> Isn't it more along the lines of `set_activations_to_store` or `set_activations_to_save` or `set_store_activations_setting`?

Changed to `save_activations`. Since it can now only be enabled/disabled, it is a regular property again (no `set_.*`). So, this functionality can now be enabled with:

```
pipe.save_activations = True
```

website/docs/api/doc.md
danieldk: Added to all `activations` and `save_activations` attributes/options.

spacy/tests/pipeline/test_morphologizer.py
danieldk: Fixed.

spacy/tests/pipeline/test_entity_linker.py
danieldk: I am a bit two-minded about that. On the one hand it's less redundant, on the other hand, it's nice when you get something like

```
FAILED spacy/tests/pipeline/test_entity_linker::test_save_activations
```

because you get a nice semantic description of what failed. Getting a `FAILED ... test_overfitting_IO` is not so informative when the activation code broke.

I guess the shared code could be factored out to a separate function, but maybe that's too much?

svlandeg: I agree that it's generally nice to have a single test for a single functionality and to immediately understand which functionality exactly broke when a test fails. But we want to make the trade-off against running time of the test suite. Like, I don't care so much about actually copy/pasting code in the test suite, but we do want to be mindful of blowing up the test suite too much.

Though now that we run the slow tests more regularly, we can also move more stuff to `slow`.

danieldk: Ah, right, I didn't consider running time (the downside of having fast local machines). Yeah, it'd probably make sense to mark these activation saving tests that do optimization as `slow`.

danieldk: Oh, I just realized that we don't actually need to train the entity linker in this test. So, I removed the optimization loop.

danieldk:
Note: tests are failing because we need Thinc with https://github.com/explosion/thinc/pull/739.

website/docs/api/edittreelemmatizer.md
danieldk: I removed the `activations` property, since it was primarily there so that the earlier `set_store_activations` could validate the queried attributes.

I'd expect that these names would largely stay the same (maybe only extended over time?).

spacy/pipeline/entity_linker.py
```diff
@@ -479,27 +515,48 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
         if count_ents != len(kb_ids):
             raise ValueError(Errors.E148.format(ents=count_ents, ids=len(kb_ids)))
         i = 0
         overwrite = self.cfg["overwrite"]
-        for doc in docs:
+        for j, doc in enumerate(docs):
+            if self.save_activations:
+                doc.activations[self.name] = {}
+                for act_name, acts in activations.items():
+                    if act_name != "kb_ids":
```
svlandeg, line 556: Let's maybe put `"kb_ids"` in a var on top of the module here? I might consider the same for the other components, but this particular string is used quite a few times verbatim.

spacy/pipeline/morphologizer.pyx
```diff
@@ -52,7 +53,13 @@ DEFAULT_MORPH_MODEL = Config().from_str(default_model_config)["model"]
 @Language.factory(
     "morphologizer",
     assigns=["token.morph", "token.pos"],
-    default_config={"model": DEFAULT_MORPH_MODEL, "overwrite": True, "extend": False, "scorer": {"@scorers": "spacy.morphologizer_scorer.v1"}},
+    default_config={
+        "model": DEFAULT_MORPH_MODEL,
+        "overwrite": True,
+        "extend": False,
+        "scorer": {"@scorers": "spacy.morphologizer_scorer.v1"},
+        "save_activations": False
```
svlandeg, line 61:
```suggestion
        "save_activations": False,
```
(to mimick `black` formatting)

spacy/pipeline/senter.pyx
```diff
@@ -38,11 +39,21 @@ DEFAULT_SENTER_MODEL = Config().from_str(default_model_config)["model"]
 @Language.factory(
     "senter",
     assigns=["token.is_sent_start"],
-    default_config={"model": DEFAULT_SENTER_MODEL, "overwrite": False, "scorer": {"@scorers": "spacy.senter_scorer.v1"}},
+    default_config={
+        "model": DEFAULT_SENTER_MODEL,
+        "overwrite": False,
+        "scorer": {"@scorers": "spacy.senter_scorer.v1"},
+        "save_activations": False
```
svlandeg, line 46:
```suggestion
        "save_activations": False,
```

spacy/pipeline/spancat.py
```diff
@@ -290,19 +300,29 @@ def set_candidates(
...
+        activations: ActivationsT: The activations, produced by SpanCategorizer.predict.
 
         DOCS: https://spacy.io/api/spancategorizer#set_annotations
         """
         labels = self.labels
-        indices, scores = indices_scores
+
+        indices = activations["indices"]
+        assert isinstance(indices, Ragged)
+        scores = cast(Floats2d, activations["scores"])
```
svlandeg, line 315: Why is one being asserted, and the other casted?

spacy/pipeline/tagger.pyx
```diff
@@ -322,4 +341,4 @@ class Tagger(TrainablePipe):
         self._allow_extra_label()
         self.cfg["labels"].append(label)
         self.vocab.strings.add(label)
-        return 1
+        return 1
```
svlandeg, line 344: Looks like an empty line/return was deleted here

spacy/pipeline/transition_parser.pyx
```diff
@@ -665,4 +665,4 @@ cdef class Parser(TrainablePipe):
                     golds.append(gold)
                 if state.is_final():
                     break
-        return states, golds, max_length
+        return states, golds, max_length
```
svlandeg, line 668: I think this change can just be reverted

spacy/tests/pipeline/test_senter.py
```diff
@@ -101,3 +103,26 @@ def test_overfitting_IO():
     # test internal pipe labels vs. Language.pipe_labels with hidden labels
     assert nlp.get_pipe("senter").labels == ("I", "S")
     assert "senter" not in nlp.pipe_labels
+
+
+def test_save_activations():
+    # Simple test to try and quickly overfit the senter - ensuring the ML models work correctly
```
svlandeg, line 109: TODO: update docstring

spacy/tests/pipeline/test_spancat.py
```diff
@@ -419,3 +419,23 @@ def test_set_candidates():
     assert len(docs[0].spans["candidates"]) == 9
     assert docs[0].spans["candidates"][0].text == "Just"
     assert docs[0].spans["candidates"][4].text == "Just a"
+
+
+def test_save_activations():
+    # Simple test to try and quickly overfit the spancat component - ensuring the ML models work correctly
```
svlandeg, line 425: TODO: update docstring

spacy/tests/pipeline/test_tagger.py
```diff
@@ -211,6 +213,26 @@ def test_overfitting_IO():
     assert doc3[0].tag_ != "N"
 
 
+def test_save_activations():
+    # Simple test to try and quickly overfit the tagger - ensuring the ML models work correctly
```
svlandeg, line 217: TODO: update docstring

spacy/tests/pipeline/test_textcat.py
```diff
@@ -871,3 +872,43 @@ def test_textcat_multi_threshold():
 
     scores = nlp.evaluate(train_examples, scorer_cfg={"threshold": 0})
     assert scores["cats_f_per_type"]["POSITIVE"]["r"] == 1.0
+
+
+def test_save_activations():
+    fix_random_seed(0)
```
svlandeg, line 878: I don't think we need to fix the random seed here (same for `test_save_activations_multi`)

kadarakos:
Please allow me to ask this :trollface:  question. Isn't the `store_activations` mechanism kind of a temporary stand-in for a more general `GlobalListener` that is kinda like an `ActivationPool` that components read, write and clear?

danieldk:
> Please allow me to ask this :trollface: question. Isn't the `store_activations` mechanism kind of a temporary stand-in for a more general `GlobalListener` that is kinda like an `ActivationPool` that components `read` from, `write` to and `clear`?
> Like the global memory of the `pipeline` is the `Doc` and this would be the global memory for the `model` part of the `pipeline`? Isn't that the kind of stuff we are missing? I guess ideally any later `model` could query and tensor produced by earlier `model`s and `backprop` should be supported through this `ActivationPool`.

But in my use case, it is not a model that uses the activations, but a pipe. Also, I think activations could be useful to downstream applications that are not pipes or models (e.g. some custom code that the user wrote). So, it seems most logical to associate them with a `Doc`? Otherwise the `Language.pipe` method should also return other things than `Doc`s.

kadarakos:
> > Please allow me to ask this :trollface: question. Isn't the `store_activations` mechanism kind of a temporary stand-in for a more general `GlobalListener` that is kinda like an `ActivationPool` that components `read` from, `write` to and `clear`?
> > Like the global memory of the `pipeline` is the `Doc` and this would be the global memory for the `model` part of the `pipeline`? Isn't that the kind of stuff we are missing? I guess ideally any later `model` could query and tensor produced by earlier `model`s and `backprop` should be supported through this `ActivationPool`.
> 
> But in my use case, it is not a model that uses the activations, but a pipe. Also, I think activations could be useful to downstream applications that are not pipes or models (e.g. some custom code that the user wrote). So, it seems most logical to associate them with a `Doc`? Otherwise the `Language.pipe` method should also return other things than `Doc`s.

Okok I see (we can talk about it later just don't want to spam more :D )

spacy/pipeline/spancat.py
danieldk: We cannot assert in the case of `scores`, since it is a NumPy array, which is not a subclass of `Floats2d` :(.

spacy/pipeline/entity_linker.py
danieldk: Fixed.

spacy/pipeline/transition_parser.pyx

spacy/tests/pipeline/test_spancat.py
danieldk: Updated all these comments.

spacy/tests/pipeline/test_textcat.py
danieldk: Removed fixing of the random seed in both cases.

svlandeg (approved):
We'll still need to release explosion/thinc#739, bump the Thinc requirement and have the CI go green, but in principle this seems good to me :-)

Will be a great feature to have!

PR #11013, opened 2022-06-22
Update vector handling in similarity methods

adrianeboyd:
## Description

Distinguish between vectors that are 0 vs. missing vectors when warning about missing vectors.

Update `Doc.has_vector` to match `Span.has_vector` and `Token.has_vector` for cases where the vocab has vectors but none of the tokens in the container have vectors.

### Types of change

Bug fix? (I was mainly trying to remove warnings from the test suite, but then noticed this inconsistency in the definitions of `has_vector` that made it difficult to fix the warning for all cases.)

## Checklist


svlandeg (approved):
Makes sense!

PR #11016, opened 2022-06-23
NEL abstention threshold

rmitsch:
## Goals

Add a configurable confidence threshold for NEL predictions.

## Description

Introduces the parameter `abstention_threshold` (defaults to 0) to `EntityLinker` and `EntityLinker_v1`. Predictions are discarded if their `score` doesn't meet the specified threshold. If no prediction does, `EntityLinker.NIL` / `EntityLinker_v1.NIL` is set.

### Types of change

New feature.

## Checklist

### DIFF
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -442,7 +442,7 @@ class EntityLinker(TrainablePipe):
                     if not candidates:
                         # no prediction possible for this entity - setting to NIL
                         final_kb_ids.append(self.NIL)
-                    elif len(candidates) == 1 and self.abstention_threshold is None:
+                    elif len(candidates) == 1 and self.abstention_threshold == 0:
                         # shortcut for efficiency reasons: take the 1 candidate
                         final_kb_ids.append(candidates[0].entity_)
                     else:
@@ -472,13 +472,13 @@ class EntityLinker(TrainablePipe):
                             if sims.shape != prior_probs.shape:
                                 raise ValueError(Errors.E161)
                             scores = prior_probs + sims - (prior_probs * sims)
-                        best_index = (
-                            xp.where(scores >= self.abstention_threshold)[0]
-                            .argmax()
-                            .item()
+                        scores = xp.where(scores >= self.abstention_threshold)[0]
+                        best_candidate_entity_ = (
+                            candidates[scores.argmax().item()].entity_
+                            if scores.size
+                            else EntityLinker.NIL
                         )
-                        best_candidate = candidates[best_index]
-                        final_kb_ids.append(best_candidate.entity_)
+                        final_kb_ids.append(best_candidate_entity_)
         if not (len(final_kb_ids) == entity_count):
             err = Errors.E147.format(
                 method="predict", msg="result variables not of equal length"
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -56,6 +56,7 @@ DEFAULT_NEL_MODEL = Config().from_str(default_model_config)["model"]
         "overwrite": True,
         "scorer": {"@scorers": "spacy.entity_linker_scorer.v1"},
         "use_gold_ents": True,
+        "abstention_threshold": None,
     },
     default_score_weights={
         "nel_micro_f": 1.0,
@@ -77,6 +78,7 @@ def make_entity_linker(
     overwrite: bool,
     scorer: Optional[Callable],
     use_gold_ents: bool,
+    abstention_threshold: float = 0,
 ):
     """Construct an EntityLinker component.
 
@@ -91,6 +93,10 @@ def make_entity_linker(
     get_candidates (Callable[[KnowledgeBase, "Span"], Iterable[Candidate]]): Function that
         produces a list of candidates, given a certain knowledge base and a textual mention.
     scorer (Optional[Callable]): The scoring method.
+    use_gold_ents (bool): Whether to copy entities from gold docs or not. If false, another
+        component must provide entity annotations.
+    abstention_threshold (Optional[float]): Confidence threshold for entity predictions. If confidence is below the
+        threshold, prediction is discarded.
     """
 
     if not model.attrs.get("include_span_maker", False):
@@ -107,6 +113,7 @@ def make_entity_linker(
             get_candidates=get_candidates,
             overwrite=overwrite,
             scorer=scorer,
+            abstention_threshold=abstention_threshold,
         )
     return EntityLinker(
         nlp.vocab,
@@ -121,6 +128,7 @@ def make_entity_linker(
         overwrite=overwrite,
         scorer=scorer,
         use_gold_ents=use_gold_ents,
+        abstention_threshold=abstention_threshold,
     )
 
 
@@ -156,6 +164,7 @@ class EntityLinker(TrainablePipe):
         overwrite: bool = BACKWARD_OVERWRITE,
         scorer: Optional[Callable] = entity_linker_score,
         use_gold_ents: bool,
+        abstention_threshold: float = 0,
     ) -> None:
         """Initialize an entity linker.
 
@@ -174,9 +183,17 @@ class EntityLinker(TrainablePipe):
             Scorer.score_links.
         use_gold_ents (bool): Whether to copy entities from gold docs or not. If false, another
             component must provide entity annotations.
+        abstention_threshold (Optional[float]): Confidence threshold for entity predictions. If confidence is below the
+            threshold, prediction is discarded. If None, predictions are made regardless of confidence.
 
         DOCS: https://spacy.io/api/entitylinker#init
         """
+
+        # todo @RM replace with proper Error.
+        assert (
+            0 <= abstention_threshold <= 1
+        ), "Abstention threshold has to be in range [0, 1]."
+
         self.vocab = vocab
         self.model = model
         self.name = name
@@ -192,6 +209,7 @@ class EntityLinker(TrainablePipe):
         self.kb = empty_kb(entity_vector_length)(self.vocab)
         self.scorer = scorer
         self.use_gold_ents = use_gold_ents
+        self.abstention_threshold = abstention_threshold
 
     def set_kb(self, kb_loader: Callable[[Vocab], KnowledgeBase]):
         """Define the KB of this pipe by providing a function that will
@@ -424,9 +442,8 @@ class EntityLinker(TrainablePipe):
                     if not candidates:
                         # no prediction possible for this entity - setting to NIL
                         final_kb_ids.append(self.NIL)
-                    elif len(candidates) == 1:
+                    elif len(candidates) == 1 and self.abstention_threshold is None:
                         # shortcut for efficiency reasons: take the 1 candidate
-                        # TODO: thresholding
                         final_kb_ids.append(candidates[0].entity_)
                     else:
                         random.shuffle(candidates)
@@ -455,8 +472,11 @@ class EntityLinker(TrainablePipe):
                             if sims.shape != prior_probs.shape:
                                 raise ValueError(Errors.E161)
                             scores = prior_probs + sims - (prior_probs * sims)
-                        # TODO: thresholding
-                        best_index = scores.argmax().item()
+                        best_index = (
+                            xp.where(scores >= self.abstention_threshold)[0]
+                            .argmax()
+                            .item()
+                        )
                         best_candidate = candidates[best_index]
                         final_kb_ids.append(best_candidate.entity_)
         if not (len(final_kb_ids) == entity_count):
--- a/spacy/pipeline/legacy/entity_linker.py
+++ b/spacy/pipeline/legacy/entity_linker.py
@@ -7,7 +7,7 @@ from pathlib import Path
 from itertools import islice
 import srsly
 import random
-from thinc.api import CosineDistance, Model, Optimizer, Config
+from thinc.api import CosineDistance, Model, Optimizer
 from thinc.api import set_dropout_rate
 import warnings
 
@@ -20,7 +20,7 @@ from ...language import Language
 from ...vocab import Vocab
 from ...training import Example, validate_examples, validate_get_examples
 from ...errors import Errors, Warnings
-from ...util import SimpleFrozenList, registry
+from ...util import SimpleFrozenList
 from ... import util
 from ...scorer import Scorer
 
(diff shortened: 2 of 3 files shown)


svlandeg:
I'm unsure whether "abstention" is such a widely used term - I haven't encountered it so much myself personally. What about `NIL_threshold` or `min_threshold` or `score_threshold` or such?

spacy/pipeline/legacy/entity_linker.py
```diff
@@ -54,6 +54,7 @@ def __init__(
         get_candidates: Callable[[KnowledgeBase, Span], Iterable[Candidate]],
         overwrite: bool = BACKWARD_OVERWRITE,
         scorer: Optional[Callable] = entity_linker_score,
+        abstention_threshold: Optional[float] = None,
```
svlandeg, line 57: We don't need to make this change in the `v1` version. In fact, I'd argue that we want a new version (`v3`) that introduces this new parameter. Basically this ensures that config files are as cross-compat as possible: we wouldn't want a situation where somebody has a `v2` with this new threshold, which works well on one system with a new version of spaCy, and then the same config with the same `v2` breaking down on another system with an older spaCy version.

spacy/pipeline/entity_linker.py
```diff
@@ -455,10 +469,19 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
+                            else xp.arange(len(scores))
+                        )
+                        best_candidate_entity_ = (
+                            candidates[
+                                meets_abst_threshold[scores.argmax().item()]
+                            ].entity_
+                            if meets_abst_threshold.size
+                            else EntityLinker.NIL
+                        )
+                        final_kb_ids.append(best_candidate_entity_)
```
svlandeg, line 484: It feels to me like the rewrite is unnecessarily complex. You're comparing every score to the threshold and still taking a max, though we might as well just take the max first (as in the original code) and only do one final check against the `max_score` and either define & append `best_candidate.entity_`, or append `NIL` if the highest score is below the threshold.

spacy/pipeline/legacy/entity_linker.py
polm: I would definitely put this in a new version.

spacy/pipeline/entity_linker.py
```diff
@@ -56,6 +56,7 @@
         "overwrite": True,
         "scorer": {"@scorers": "spacy.entity_linker_scorer.v1"},
         "use_gold_ents": True,
+        "abstention_threshold": None,
```
polm, line 59: Is there a difference in behavior between `None` and 0 here? If not, shouldn't it just default to 0?

spacy/pipeline/entity_linker.py
```diff
@@ -174,9 +183,14 @@ def __init__(
             Scorer.score_links.
         use_gold_ents (bool): Whether to copy entities from gold docs or not. If false, another
             component must provide entity annotations.
-
+        abstention_threshold (Optional[float]): Confidence threshold for entity predictions. If confidence is below the
+            threshold, prediction is discarded. If None, predictions scores are not checked.
         DOCS: https://spacy.io/api/entitylinker#init
         """
+
+        if abstention_threshold is not None and not (0 <= abstention_threshold <= 1):
```
polm, line 191: Should this be strictly less than 1?

polm (commented):
This is a good start! I agree with Sofie that 1. the legacy component doesn't need to be updated 2. It's better to take the max and then filter. 

As an extra point, I'm not sure why the default threshold is `None`, since it seems to work the same as 0.

polm:
> I'm unsure whether "abstention" is such a widely used term - I haven't encountered it so much myself personally. What about NIL_threshold or min_threshold or score_threshold or confidence_threshold or such?

I actually think it's a good name. I'm not sure there's a single commonly-used term for this in the literature, but I think it's very clear about the purpose is. However, I think it makes the most sense to make this just `threshold`, which is consistent with our other components, like textcat and spancat.

spacy/pipeline/legacy/entity_linker.py
rmitsch: What does "new version" entail exactly? A new `EntityLinker` class with the `confidence_trehold` argument + rerouting `make_entity_linker()`?

spacy/pipeline/entity_linker.py
```diff
@@ -424,9 +439,8 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
                     if not candidates:
                         # no prediction possible for this entity - setting to NIL
                         final_kb_ids.append(self.NIL)
-                    elif len(candidates) == 1:
+                    elif len(candidates) == 1 and self.abstention_threshold == 0:
```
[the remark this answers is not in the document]
rmitsch: Yes, should have been `None` anyway - `0` was a leftover from then default value was `0`.

spacy/pipeline/entity_linker.py
rmitsch: That's absolutely correct :sweat_smile:

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -766,7 +766,7 @@ def create_kb(vocab):
         doc = nlp(text)
         for ent in doc.ents:
             predictions.append(ent.kb_id_)
-    assert predictions == GOLD_entities
+    assert sorted(predictions) == sorted(GOLD_entities)
```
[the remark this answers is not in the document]
rmitsch: It's not, it's a leftover from when I was trying to track down a bug. Removed.

spacy/pipeline/entity_linker.py
rmitsch: It's possible for the scores to be below zero. It's likely practically not relevant, but can happen (e.g. before running a single `update()`). In that case these scores would be discarded, even though no `threshold` might have been set. I think it'd be more reasonable to expect that no threshold check is run if the user doesn't set it, hence `None` instead of `0`.

spacy/pipeline/entity_linker.py
rmitsch: Don't have a strong opinion on this. Do you think it should be? I think we could leave that up to the user.

spacy/pipeline/legacy/entity_linker.py
polm: Ah, right, this is a pipeline change, not a model class change...

In that case I guess this is like the `overwrite` changes to components when we added that across the board. I'd need to go back and look at those PRs, but I think in that case you could add it with a default that doesn't change behavior. 

Because the v1 NEL component is a little weird this is a something we'd be deciding for the first time, but my thought is that it should stay basically frozen, so there's no need to add support for a new feature to it.

rmitsch:
> I actually think it's a good name. I'm not sure there's a single commonly-used term for this in the literature, but I think it's very clear about the purpose is. However, I think it makes the most sense to make this just `threshold`, which is consistent with our other components, like textcat and spancat.

I also like the name due to its specificity. It's true though that I didn't find much literature on "abstention" - sometimes authors refer to "abstaining from making predictions", but almost never to "abstention". I think `threshold` is a good alternative, especially for its consistency with other components.

spacy/pipeline/legacy/entity_linker.py
rmitsch: > ...so there's no need to add support for a new feature to it.

I'm good with v1 staying frozen. 

> ...but I think in that case you could add it with a default that doesn't change behavior.

Meaning adding it to `EntityLinker`, but not adding a new class?

spacy/pipeline/entity_linker.py
polm: OK, makes sense. It would be good to mention that somewhere in a comment or docs.

spacy/pipeline/entity_linker.py
polm: No, I don't think it makes sense to strictly enforce that. Using 1 would be kind of silly but the user has to choose it.

spacy/pipeline/legacy/entity_linker.py
polm: > Meaning adding the argument to EntityLinker, but not adding a new class?

Yes. When we added the `overwrite` option to all the components, we used a default that didn't change behavior and therefore didn't add a new architecture version. See #9050. I think we can handle this the same way.

adrianeboyd: Yes, in the past we've added new optional parameters that preserve the existing defaults without modifying the names.

It's an issue that the factory names aren't versioned so the only thing you can really do is add new settings where the default preserves the existing behavior. It does mean that it will complain if you use the auto-filled with a previous version of spacy, but if you always auto-fill from a base config it should still work with old versions. I think this is acceptable?

spacy/pipeline/entity_linker.py
rmitsch: It's mentioned in the docstrings, will also write about it in the docs.

spacy/pipeline/legacy/entity_linker.py
svlandeg: Oh sorry, I totally confused things as I was mixing up between an architecture change and a factory change - as Adriane pointed out we don't typically version the latter (`entity_linker_v1` is a weird exception but as discussed that can just be frozen)

spacy/errors.py
```diff
@@ -934,6 +934,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1041 = ("Expected a string, Doc, or bytes as input, but got: {type}")
     E1042 = ("Function was called with `{arg1}`={arg1_values} and "
              "`{arg2}`={arg2_values} but these arguments are conflicting.")
+    E1043 = ("Expected None or a value in range [0, 1] for abstention threshold, but got {value}.")
```
svlandeg, line 937: Maybe instead of "abstention threshold" we can make this "EntityLinker threshold" but actually give the component name to `E1043` as a parameter? (kind of like how `W088` is used)

spacy/pipeline/entity_linker.py
```diff
@@ -91,6 +93,10 @@ def make_entity_linker(
     get_candidates (Callable[[KnowledgeBase, "Span"], Iterable[Candidate]]): Function that
         produces a list of candidates, given a certain knowledge base and a textual mention.
     scorer (Optional[Callable]): The scoring method.
+    use_gold_ents (bool): Whether to copy entities from gold docs or not. If false, another
+        component must provide entity annotations.
+    threshold (Optional[float]): Confidence threshold for entity predictions. If confidence is below the threshold,
+        prediction is discarded. If None, predictions scores are not checked.
```
svlandeg, line 99:
```suggestion
    threshold (Optional[float]): Confidence threshold for entity predictions. If confidence is below the threshold,
        prediction is discarded. If None, predictions are not filtered by any threshold.
```

spacy/pipeline/entity_linker.py
```diff
@@ -174,9 +182,14 @@ def __init__(
             Scorer.score_links.
         use_gold_ents (bool): Whether to copy entities from gold docs or not. If false, another
             component must provide entity annotations.
-
+        threshold (Optional[float]): Confidence threshold for entity predictions. If confidence is below the
+            threshold, prediction is discarded. If None, predictions scores are not checked.
```
svlandeg, line 186:
```suggestion
            threshold, prediction is discarded. If None, predictions are not filtered by any threshold.
```

svlandeg (commented):
Looks good to me! Only had a few minor nitpicks/suggestions.

spacy/pipeline/legacy/entity_linker.py
rmitsch: Can this remain as is then or are further changes necessary/recommended?

svlandeg: I reviewed again yesterday, the legacy version has remained unchanged which was my main point, so this particular comment can be resolved :-)

rmitsch:
Updated docs.

website/docs/api/entitylinker.md
```diff
@@ -47,22 +47,24 @@ architectures and their arguments and hyperparameters.
...
+| `overwrite` <Tag variant="new">3.2</Tag> | Whether existing annotation is overwritten. Defaults to `True`. ~~bool~~                                                                                                                                                                                                                       |
+| `scorer` <Tag variant="new">3.2</Tag>    | The scoring method. Defaults to [`Scorer.score_links`](/api/scorer#score_links). ~~Optional[Callable]~~                                                                                                                                                                                        |
+| `threshold` <Tag variant="new">3.4</Tag> | Confidence threshold for entity predictions. The default of `None` implies that all predictions are accepted, otherwise those with a score beneath the treshold are discarded. If no predictions with scores above it are inferred, the linked entity is `NIL`. ~~Optional[float]~~ |
```
polm, line 67: Suggest change:

> If no predictions with scores above it are inferred,

to:

> if there are no predictions with scores above the threshold,

website/docs/api/entitylinker.md
```diff
@@ -95,20 +97,21 @@ custom knowledge base, you should either call
...
```
polm, line 114: Same suggestion as the other one.

spacy/errors.py
```diff
@@ -934,6 +934,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1041 = ("Expected a string, Doc, or bytes as input, but got: {type}")
     E1042 = ("Function was called with `{arg1}`={arg1_values} and "
              "`{arg2}`={arg2_values} but these arguments are conflicting.")
+    E1043 = ("Expected None or a value in range [{range_start}, {range_end}] for {metric}, but got {value}.")
```
polm, line 937: So in general this would be a good idea to keep the number of error codes down, but this is only used with the EntityLinker threshold, right? In that case I think it should be fine to just change `{metric}` to `"entity linker threshold"`. This does have the advantage that the error code indicates a specific failure and not a generic one.

polm (commented):
Few questions and nitpicks but looking good!

adrianeboyd:
If there's no difference between `threshold=None` and `threshold=0.0`, I think it might be better to type `threshold` as `float` in the same way as `spancat`.

rmitsch:
> If there's no difference between `threshold=None` and `threshold=0.0`, I think it might be better to type `threshold` as `float` in the same way as `spancat`.

There is, see [this convo](https://github.com/explosion/spaCy/pull/11016#discussion_r905922730). So IMO it makes sense to keep the `None` - open to suggestions though.

svlandeg (approved):
This looks good to me!

PR #11017, opened 2022-06-23
Remove `simply`

ericholscher:
## Description
This just removes the word `simply` from the architecture docs. I was reading this page, and as a relative beginner, nothing about it was simple. There might be other ways to express some of these ideas, eg `only`, but I'm not actually sure what most of these sentences are saying :) 

### Types of change
Doc fix

## Checklist


svlandeg (commented):
Thanks, I see your point. In the two last cases, "simply" was used to capture the fact that more complicated architectures/algorithms could be implemented as well, and that the default spaCy implementation is a relatively straightforward one, compared to the realm of all possibilities. I think the cases that you highlighted all read fine without this word though.

PR #11024, opened 2022-06-24
`Morphology`/`Morphologizer` optimizations and refactoring

shadeMe:
## Description

This PR refactors the following classes and their associated types:

### `Morphology` 
- Replace `Pool`-based memory management of features with C++ vector-based containers.
- Replace `PreshMap` with C++ `unordered_map`.
- Reduce memory allocations all around by eliding unnecessary copies of Python objects and containers.
- Preserve a normalized `dict` of features that's reused for transformations into `str`.
- Remove special-case handling for POS tag features during normalization as it was essentially superfluous (the value was being converted to a string in the next step).
- Remove unused `normalize_attrs` method.

### `Morphologizer`
- Do not unnecessarily sort feature `dict`s.
- Change `labels` to return an `Iterable` instead of a `Tuple`. This was causing a lot of copies.
- Only convert labels to `Tuple` when we need random access. 

The optimizations introduced in this PR result in significant performance gains in the morphologizer loss calculation part of the training pipeline: throughput per mini-batch increases by ~85% (~43% reduction in CPU time).

#### Before
Screenshot_20220624_125028

### After
Screenshot_20220624_125110

Since some of the above API changes are breaking, this PR targets the `v4` branch.

### Types of change

Refactoring, performance enhancement

## Checklist


shadeMe:
> Btw, does unordered_map perform better than the PreshMap on this problem size? If so that's good to know and might open up other optimisations in future. PreshMap was originally tested on quite large arrays so maybe on small data it performs poorly.
I didn't explicitly benchmark the two to compare, as the primary reason behind the switch was to eschew the semi-manual memory management it required for the inserted values (`cymem` overhead was actually perceptible in this scenario). But I agree that we should benchmark our support libraries in different scenarios to see if we can eke out more performance (I'll add it to the TODO list).

danieldk:
I'd also be interested in the performance of `unordered_map`. It has a reputation for being fairly slow in order to support the standard to the letter (e.g. iterator invalidation rules). Though I am not sure how much it matters for our use cases. If it does make a difference, we could consider a license-compatible header-only replacement.

danieldk:
For the build failures: I think you need to merge in the `v4` branch to get the `CBlas` changes that we recently made.

spacy/morphology.pxd
```diff
@@ -1,21 +1,40 @@
...
+        this.key = 0
+
 cdef class Morphology:
-    cdef readonly Pool mem
     cdef readonly StringStore strings
-    cdef PreshMap tags # Keyed by hash, value is pointer to tag
-
-    cdef MorphAnalysisC create_morph_tag(self, field_feature_pairs) except *
-    cdef int insert(self, MorphAnalysisC tag) except -1
+    cdef unordered_map[hash_t, unique_ptr[MorphAnalysisC]] tags
```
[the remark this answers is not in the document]
shadeMe: Just wanted to avoid the overhead of atomic increment/decrement ops required to operate the reference counter in `std::shared_ptr`. I considered avoiding the pointer indirection entirely, but Cython's limited support for move semantics meant that we can't construct the `MorphAnalysisC` object in-place in the `map` and will have to suffer allocations, deallocations multiple temporaries and at least one copy.

I do agree with your reasoning, though. I can do some more profiling to determine if the `shared_ptr` overhead is small enough to not matter.

spacy/morphology.pyx
```diff
@@ -24,134 +23,183 @@ cdef class Morphology:
...
+
+
+    cdef hash_t _add(self, features):
         """Insert a morphological analysis in the morphology table, if not
         already present. The morphological analysis may be provided in the UD
         FEATS format as a string or in the tag map dict format.
         Returns the hash of the new analysis.
         """
-        cdef MorphAnalysisC* tag_ptr
+        cdef hash_t hash = 0
```
svlandeg, line 95: To avoid confusion with the builtin, let's call this `tag_hash` as well?

spacy/morphology.pyx
```diff
@@ -24,134 +23,183 @@ cdef class Morphology:
...
-        cdef MorphAnalysisC tag = self.create_morph_tag(field_feature_pairs)
+
         # the hash key for the tag is either the hash of the normalized UFEATS
         # string or the hash of an empty placeholder
-        norm_feats_string = self.normalize_features(features)
-        tag.key = self.strings.add(norm_feats_string)
-        self.insert(tag)
-        return tag.key
+        norm_feats_string = self._normalized_feat_dict_to_str(features)
+        hash = self.strings.add(norm_feats_string)
```
svlandeg, line 116: Same comment about builtin

spacy/morphology.pyx
```diff
@@ -24,134 +23,183 @@ cdef class Morphology:
...
+        # Feat dict is normalized at this point.
+        for field, values in feats.items():
+            field_key = self.strings.add(field)
+            if isinstance(values, list):
+                for value in values:
+                    value_key = self.strings.add(field + self.FIELD_SEP + value)
+                    field_feature_pairs.append((field_key, value_key))
+            else:
+                value_key = self.strings.add(field + self.FIELD_SEP + values)
+                field_feature_pairs.append((field_key, value_key))
```
svlandeg, line 140: Rather than copying this code, we could also check if `values` is not a list, then do 
```
values = [values]
```
and continue with the `for value in values` loop.

spacy/pipeline/morphologizer.pyx
```diff
@@ -231,26 +231,29 @@ class Morphologizer(Tagger):
         cdef Vocab vocab = self.vocab
         cdef bint overwrite = self.cfg["overwrite"]
         cdef bint extend = self.cfg["extend"]
-        labels = self.labels
+
+        # We need random access for the upcoming ops, so we need to 
+        # we need to allocate a compatible container out of the iterable.
```
svlandeg, line 236:
```suggestion
        # We need random access for the upcoming ops, 
        # so we need to allocate a compatible container out of the iterable.
```

svlandeg (commented):
Great work Madeesh!

spacy/morphology.pyx
```diff
@@ -24,134 +23,183 @@ cdef class Morphology:
     EMPTY_MORPH = symbols.NAMES[symbols._]
 
     def __init__(self, StringStore strings):
-        self.mem = Pool()
         self.strings = strings
-        self.tags = PreshMap()
```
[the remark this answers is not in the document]
shadeMe: No, that's not necessary as the generated C++ code will automatically call the default constructor.

spacy/morphology.pyx
shadeMe: Ah, the current implementation is actually an explicit choice to reduce memory allocation. If I recall the results of my last profiling session correctly, the second branch (where we only have one value) is taken often enough that adding an allocation there - and consequent the enumeration code - incurs perceptible overhead.

spacy/morphology.pyx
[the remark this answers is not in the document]
svlandeg: That's what I was assuming, thanks for the verification!

spacy/morphology.pyx
svlandeg: Oh, that's surprising! But sure, if that's what you find then by all means let's keep the code as is. Maybe add a quick comment so it doesn't get rewritten in the future?

spacy/morphology.pyx
```diff
@@ -37,7 +38,7 @@ cdef class Morphology:
         else:
             return shared_ptr[MorphAnalysisC]()
 
-    def _normalize_attr(self, attr_key, attr_value):
+    def _normalize_attr(self, attr_key : Union[int, str], attr_value : Union[int, str]) -> Tuple[str, Union[str, List[str]]]:
```
svlandeg, line 41: We do return `None` if the input is of unexpected types, so then this should be `Optional`? Or should that code path raise instead of warn?

spacy/morphology.pyx
```diff
@@ -24,134 +24,187 @@ cdef class Morphology:
...
+
+        return attr_key, attr_value
+
+    def _str_to_normalized_feat_dict(self, feats: str) -> Dict[str, str]:
+        if not feats or feats == self.EMPTY_MORPH:
+            return {}
+
+        out = []
+        for feat in feats.split(self.FEATURE_SEP):
+            field, values = feat.split(self.FIELD_SEP)
```
danieldk, line 63: I wonder if we should use `feat.split(self.FIELD_SEP, 1)` or do data validation here? If we can assume that `FIELD_SEP` occurs once, then `feat.split(self.FIELD_SEP, 1)` might be a bit faster, since `split` doesn't have to search the rest of the string for matches (and can perhaps opportunistically allocate).

danieldk (commented):
Really nice PR 🎉. I really like how besides improving performance it also increases safety while at it.

Only two small comments/questions about two `split`s.

PR #11026, opened 2022-06-24
`Example.get_aligned_parse`: Handle unit and zero length vectors correctly

shadeMe:
## Description

We were unnecessarily squeezing the gold token index vectors, causing unit-length vectors to yield a scalar and breaking indexing (the root cause seems to point to a reference sentence containing just one token). We additionally provide the output type to `np.vectorize` to correctly handle corner cases where no alignment can be made between the gold and candidate tokens.

These changes were sanity checked by locally training English and German parser models for multiple epochs.

### Types of change

Bugfix

## Checklist

### DIFF
--- a/spacy/tests/training/test_training.py
+++ b/spacy/tests/training/test_training.py
@@ -689,17 +689,16 @@ def test_projectivize(en_tokenizer):
     assert proj_labels == deps
 
     # Test documents with no alignments
-    doc_a = Doc(doc.vocab).from_json(
-        json.loads(
-            '{"text": "Double-Jointed", "tokens": [{"id": 0, "start": 0, "end": 14, "head": 0, "dep": "ROOT"}]}'
-        )
+    doc_a = Doc(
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Could you also add a test for this case?

spacy/tests/training/test_training.py
```diff
@@ -930,7 +929,7 @@ def test_split_sents(merged_dict):
     assert token_annotation_2["SENT_START"] == [1, 0, 0, 0]
 
 
-def test_alignment():
+def test_alignment(en_tokenizer):
```
adrianeboyd, line 932:
```suggestion
def test_alignment():
```

spacy/tests/training/test_training.py
```diff
@@ -1,5 +1,5 @@
 import random
-
+import json
```
danieldk, line 2: Leftover

adrianeboyd (approved):

PR #11042, opened 2022-06-28
Fix tokenization mismatch handling in coref

polm:
## Description

Coref was written without any consideration for mismatches in tokenization between the gold and predicted docs. This PR makes it so such issues are handled correctly. 

One consideration here is, what to do when a target token (the head of a mention span) isn't shared between the two tokenizations? It may be desirable in some cases to have a mitigation strategy, like the expand/contract strategy present in `doc.char_span`. However at present this PR just throws an error if a target can't be mapped.

This is still in progress.

### Types of change

Bug fix.

## Checklist


polm:
@explosion-bot Please test_slow_gpu

polm:
I believe this should be working well enough to merge now. 

There's still a question of what to do when annotations don't have compatible boundaries between gold and predicted Docs. Currently coref gives up and the span predictor will ignore those spans (zero gradients); I don't have strong feelings about which of these are better. As mentioned in the initial post here I could also see offering mitigation strategies make sense (I feel like `expand` would usually be appropriate), though it would make the component more complicated.

spacy/ml/models/coref_util.py
```diff
@@ -201,3 +203,17 @@ def create_gold_scores(
 
     # caller needs to convert to array, and add placeholder
     return out
+
+
+def spans2ints(doc):
```
adrianeboyd, line 208: To follow the naming conventions in the IOB utils and other conversions I'd suggest renaming this to at least `spans_to_offsets`.

It's unfortunate that we use `offsets` for both token and character offsets, but `spans_to_char_offsets` gets kind of clunky, or even clunkier `unlabeled_spans_to_char_offsets` so it's not confused with the utilities for labeled spans.

Maybe it should just be private, especially if it's effectively only a test util?

adrianeboyd:
There is one remaining catch with using character offsets with `Example`, and that's that the whitespace isn't required to align between the predicted and reference docs, so working with character offsets could lead to unexpected/incorrect results.

```python
eg = Example(nlp.make_doc("\n\nThis   is  an    example."), nlp("This is an example."))
```

I think this is rare in practice and it's okay if the component can't handle it, but there should be explicit errors in all the relevant spots that this component can't handle this kind of example.

polm:
Oh wow, I was not aware that was allowed. I'll work on adding checks that the character spans have the same contents after being translated from the reference to the predicted doc.

spacy/ml/models/coref_util.py
polm: Thanks, I'll rename it to `_spans_to_offsets` - originally this was just in one of the tests, but I moved it to util because it was relevant for both components. 

I think we do need a util function that can say "these SpanGroups from different docs are equivalent" in a robust way, but that isn't needed for these tests / in this PR. (Maybe there's already something like that for spancat?)

adrianeboyd:
I think you could just check that the texts of the predicted and reference docs are the same? (You could also lowercase them if you wanted to allow little more of the original leeway.)

The actual requirement is:

https://github.com/explosion/spaCy/blob/7c1bf2fa1fd60004a4646068e7c281978c374d60/spacy/training/align.pyx#L16-L20

(We ran into problems at one point because Turkish dotless i is not the same number of characters uppercase and lowercase.)

polm:
Thanks, that makes it clear, and checking up front would definitely be easier. I was thinking that if the checks were only in the required places it would make them easier to remove / fix later, but it's probably not that big a difference.

polm:
OK, the point about differing whitespace should be addressed now.

adrianeboyd:

adrianeboyd:

spacy/pipeline/coref.py
```diff
@@ -218,6 +218,13 @@ def update(
         total_loss = 0
 
         for eg in examples:
+            if eg.x.text != eg.y.text:
```
kadarakos, line 221: Not sure how expensive is this in `Python` for longer documents. Is this something we must do?

spacy/ml/models/coref_util.py
```diff
@@ -143,16 +143,18 @@ def create_head_span_idxs(ops, doclen: int):
 
 
 def get_clusters_from_doc(doc) -> List[List[Tuple[int, int]]]:
-    """Given a Doc, convert the cluster spans to simple int tuple lists."""
+    """Given a Doc, convert the cluster spans to simple int tuple lists. The
+    ints are char spans, to be tokenization independent.
+    """
     out = []
     for key, val in doc.spans.items():
         cluster = []
```
kadarakos, line 151:
```suggestion
        cluster = set()
```

spacy/ml/models/coref_util.py
```diff
@@ -143,16 +143,18 @@ def create_head_span_idxs(ops, doclen: int):
...
         for span in val:
-            # TODO check that there isn't an off-by-one error here
-            # cluster.append((span.start, span.end))
-            # TODO This conversion should be happening earlier in processing
+
             head_i = span.root.i
-            cluster.append((head_i, head_i + 1))
+            head = doc[head_i]
+            char_span = (head.idx, head.idx + len(head))
+            cluster.append(char_span)
```
kadarakos, line 157:
```suggestion
            cluster.add(char_span)
```

spacy/ml/models/coref_util.py
```diff
@@ -143,16 +143,18 @@ def create_head_span_idxs(ops, doclen: int):
...
 
         # don't want duplicates
         cluster = list(set(cluster))
```
kadarakos, line 160:
```suggestion
        cluster = list(cluster)
```

spacy/ml/models/coref_util.py
```diff
@@ -143,16 +143,18 @@ def create_head_span_idxs(ops, doclen: int):
 
 
 def get_clusters_from_doc(doc) -> List[List[Tuple[int, int]]]:
-    """Given a Doc, convert the cluster spans to simple int tuple lists."""
+    """Given a Doc, convert the cluster spans to simple int tuple lists. The
+    ints are char spans, to be tokenization independent.
```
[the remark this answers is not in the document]
adrianeboyd: I think the original is better. (I'm not sure "robust" is the right framing, and even then probably "robust to differences in tokenization" but not "robust to tokenization".)

spacy/pipeline/coref.py
adrianeboyd: It is something that you have to do because otherwise you can't know whether you can actually compare the character offsets. If speed turns out to be an issue, it could be cached in `Example` instead, maybe?

spacy/tests/pipeline/test_span_predictor.py
```diff
@@ -0,0 +1,227 @@
...
+def test_tokenization_mismatch(nlp):
+    train_examples = []
+    for text, annot in TRAIN_DATA:
+        eg = Example.from_dict(nlp.make_doc(text), annot)
+        ref = eg.reference
+        char_spans = {}
+        for key, cluster in ref.spans.items():
+            char_spans[key] = []
+            for span in cluster:
+                char_spans[key].append((span[0].idx, span[-1].idx + len(span[-1])))
```
kadarakos, line 157: could this also be `span.start_char` and `span.end_char` for simplicity?

spacy/pipeline/coref.py
polm: I believe that Python caches the hash for strings and only compares that, but it's good to keep in mind.

spacy/ml/models/coref_util.py
polm: One of the failure modes of coref is to say `[John, John, John, John]` (exactly the same span) are all coreferences, so eliminating duplicate indices with a set would obscure that.

spacy/ml/models/coref_util.py
```diff
@@ -201,3 +203,17 @@ def create_gold_scores(
...
+
+
+def _spans_to_offsets(doc):
+    """Convert doc.spans to nested list of ints for comparison.
+    The ints are character indices, and the spans groups are sorted by key first.
+
+    This is useful for checking consistency of predictions.
+    """
+    out = []
+    keys = sorted([key for key in doc.spans])
```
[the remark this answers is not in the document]
polm: When comparing Docs, the SpanGroups should be compared to SpanGroups with the same name. The names are gone here so it's important the keys are accessed in the same order. 

In my actual tests I found that without this the keys were in different orders sometimes, so without this the tests could fail.

spacy/tests/pipeline/test_coref.py
```diff
@@ -84,73 +89,115 @@ def test_initialized_short(nlp):
...
 
-    nlp.add_pipe("coref")
+    nlp.add_pipe("coref", config=CONFIG)
     optimizer = nlp.initialize()
     test_text = TRAIN_DATA[0][0]
     doc = nlp(test_text)
-    print("BEFORE", doc.spans)
 
-    for i in range(5):
+    # Needs ~12 epochs to converge
```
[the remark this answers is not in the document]
polm: That might work, but running the whole file takes 5s on my machine, so I'm not sure it's all that slow anyway?

spacy/tests/pipeline/test_span_predictor.py
polm: Good point, not sure why I wasn't doing that already.

spacy/tests/pipeline/test_coref.py
```diff
@@ -84,73 +89,115 @@ def test_initialized_short(nlp):
...
+
+
+@pytest.mark.skipif(not has_torch, reason="Torch not available")
+def test_tokenization_mismatch(nlp):
+    train_examples = []
+    for text, annot in TRAIN_DATA:
+        eg = Example.from_dict(nlp.make_doc(text), annot)
+        ref = eg.reference
+        char_spans = {}
+        for key, cluster in ref.spans.items():
```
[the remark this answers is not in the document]
polm: `get_clusters_from_doc` doesn't preserve keys, and we need to restore keys later, so it won't work here. It could be good to have a function for that though.

spacy/pipeline/span_predictor.py
```diff
@@ -231,24 +238,44 @@ def get_loss(
...
             n_classes = start_scores.shape[1]
             start_probs = ops.softmax(start_scores, axis=1)
             end_probs = ops.softmax(end_scores, axis=1)
             start_targets = to_categorical(starts, n_classes)
             end_targets = to_categorical(ends, n_classes)
             start_grads = start_probs - start_targets
             end_grads = end_probs - end_targets
-            grads = ops.xp.stack((start_grads, end_grads), axis=2)
+            # now return to original shape, with 0s
+            final_start_grads = ops.alloc2f(*span_scores[:, :, 0].shape)
```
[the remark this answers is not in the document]
polm: If the array was full size, with fields that were to be masked out later, wouldn't that change the softmax caculation? Besides that I think these are equivalent.

spacy/tests/pipeline/test_coref.py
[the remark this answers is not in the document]
polm: I replaced `_spans_to_offsets` with `get_clusters_from_doc`. It ended up working without sorting keys too, so I must have imagined the issue, though I guess we'll see if it comes up again.

spacy/ml/models/coref_util.py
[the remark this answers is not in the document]
polm: I ended up replacing this function with `get_clusters_from_doc`, which worked despite not sorting, so I guess I misremembered this. We'll have to see if it comes up again though.

polm:
Merged this to get all the coref changes in one place rather than having several small PRs.

PR #11056, opened 2022-06-30
dont use get_array_module

kadarakos:
Update for new `thinc` compatibility

## Description
The behavior of `get_array_module` is changing in `thinc` and it will `raise ValueError` when passing an argument that is neither a `numpy` object nor a `cupy` object.

### Types of change
Mini update

## Checklist


svlandeg:
@explosion-bot please test_slow --thinc-branch refs/pull/703/head

adrianeboyd:
Ah, nevermind, I see more what's going on. But are there other places in the code where this could be wrong?

danieldk (approved):
I think using `self.model.ops.xp` is also semantically more correct. After all, it's source of arrays we get out of the model as well in the successful case.

PR #11058, opened 2022-06-30
Non greedy quantifier

shen-qin:
## Checklist


polm:
Thanks for the PR, this would be a wonderful feature to have! 

I notice you didn't fill in the PR message at all - could you confirm that you have the right to submit this under the project's license? Also it looks like you updated the docs already.

shen-qin:
Hi! This is an accident let me close the PR, will send the official PR later!

PR #11064, opened 2022-07-01
Non-greedy quantifiers ['*?', '+?']

shen-qin:
This PR adds non-greedy quantifiers ['*?',  '+?'] to spaCy matcher. 

### Abstract
This pull request implements the non-greedy `*?` and `+?` quantifiers which **returns as few token match instances as possible**. Within the current spaCy matcher, `*` and `+` quantifiers returns **all possible matches** which might be more than the desired results.  

As a part of this pull request, quantifier `ZERO_MINUS` is added and the existing quantifier `ONE` is modified to achieve `*?: (ZERO_MINUS,)` and `+?: (ONE, ZERO_MINUS)`.

The non-greedy behaviour is achieved mainly via `cast_to_non_greedy_action()` which changes actions from `get_action()` into one that demonstrates non-greedy behaviour i.e. remove matches that are of greedy behaviour.

Additionally, this PR also can be further developed into the non-greedy version of the min_max quantifiers `{m,n} ?`.

#### Accuracy
The current spacy matcher demonstrates "get all" behaviour for `*` and `+` where all possible matches are returned. However, this may create undesired results and require the user to filter the matches.

```
pattern = [{"IS_DIGIT": True}, {"ORTH": "."}, {"OP": "+?"}, {"ORTH": "\n"}]
matcher.add("Pattern", [pattern])
doc = nlp("""
These are the ingredients to make a soufflé: 
1. 200g of flour 
2. 2 eggs 
3. 4 pods of vanilla 
4. water
""")
```
```
Output: 
"1. 200g of flour ", "2. 2 eggs", "3. 4 pods of vanilla", "4. water"
```
In the above example, the desired output is to return each line of ingredients to make the soufflé. If `+` quantifier was used, it would have also returned other results such as `"1. 200g of flour 2. 2 eggs"` and even the whole string.

Minimal or non-greedy quantifiers which match as few instances as possible would be able to achieve the desired output of only the 4 lines of ingredients as seen above.

#### Speed Improvements
Furthermore, the termination of branches once there is a match results in a faster run time when using non-greedy quantifiers.

Using the [sample text](https://github.com/explosion/spaCy/files/9056451/text.txt) and the code below, the greedy pattern consisting of syncategorematic words with greedy wildcards is always slower than the non-greedy counterpart.
```
Greedy pattern
{"TEXT": "a"}, {"OP": "+"}, {"TEXT": "the"}, {"OP": "+"}, {"TEXT": "of"}, {"OP": "+"}, {"TEXT": "to"}, {"OP": "+"}, {"TEXT": "be"}
Non-greedy pattern
{"TEXT": "a"}, {"OP": "+?"}, {"TEXT": "the"}, {"OP": "+?"}, {"TEXT": "of"}, {"OP": "+?"}, {"TEXT": "to"}, {"OP": "+?"}, {"TEXT": "be"}
```
One instance of the test using the sample text which is 670 words long is as follows:
Greedy + :            0.47761 seconds
Non-greedy +? :  0.39530 seconds

##### Speed test
```
import timeit

import_module = '''
import spacy
'''

testcode1 = '''
matcher = Matcher(nlp.vocab)
file_name = 'text.txt'
text = open(file_name).read()
pattern = [{"TEXT": "a"}, {"OP": "+"}, {"TEXT": "the"}, {"OP": "+"}, {"TEXT": "of"}, {"OP": "+"}, {"TEXT": "to"}, {"OP": "+"}, {"TEXT": "be"}]
matcher.add("Pattern", [pattern])
doc = nlp(text)
matches = matcher(doc)
'''

testcode2 = '''
matcher = Matcher(nlp.vocab)
file_name = 'text.txt'
text = open(file_name).read()
pattern = [{"TEXT": "a"}, {"OP": "+?"}, {"TEXT": "the"}, {"OP": "+?"}, {"TEXT": "of"}, {"OP": "+?"}, {"TEXT": "to"}, {"OP": "+?"}, {"TEXT": "be"}]
matcher.add("Pattern", [pattern])
doc = nlp(text)
matches = matcher(doc)
'''

print("Greedy + :", timeit.timeit(stmt=testcode1, setup=import_module, number=1))
print("Non-greedy +? : ", timeit.timeit(stmt=testcode2, setup=import_module, number=1))
```

The addition of the non-greedy quantifiers will thus simplify the process of using spaCy matcher and it would be a beneficial addition to spaCy matcher.

### `*?` vs `*`  Illustration
(Refer to spacy/tests/matcher/test_matcher_logic.py for more examples)
```
*?
string: "a a b" 
pattern: a*? b*
result = ["a a b",  "a b",  "b"] 
```
vs 
```
*
string: "a a b" 
pattern: a* b*
result = ["a", "a a", "a", "a a b",  "a b",  "b"] 
```

## Main changes

**matcher.pyx and matcher.pxd**
1. Added Quantifier "ZERO_MINUS", modified "ONE"
2. Added actions "EXTEND", "RETRY_OR_EXTEND" and "MATCH_ADVANCE"
3. cast_to_non_greedy_action() function added

**test_matcher_logic.py**
1. test_matcher_non_greedy_operator() added to demonstrate the non-greedy behaviour of "*?" and "+?"

### Types of change

New feature

## Checklist


adrianeboyd:
Thanks for the PR! Since this changes the `Matcher` algorithm, we may need some time to do a careful evaluation, especially to make sure it doesn't affect the current performance. To make it easier to review, could you undo all the unrelated formatting changes in `matcher.pyx`?

shen-qin:
> Thanks for the PR! Since this changes the `Matcher` algorithm, we may need some time to do a careful evaluation, especially to make sure it doesn't affect the current performance. To make it easier to review, could you undo all the unrelated formatting changes in `matcher.pyx`?

Removed

adrianeboyd:
Thanks for the quick reply, we'll have a look!

adrianeboyd:
In my first round of simple comparisons, the speed of the matcher for patterns involving `*` and `+` looks to be unchanged.

Let me make a few minor formatting edits to make the diff a bit cleaner so it's easier to review all the changes to the algorithm.

shen-qin:
Hello! Are there any updates for this? Thank you :)

spacy/tests/matcher/test_matcher_logic.py
```diff
@@ -786,3 +798,63 @@ def test_matcher_with_alignments_non_greedy(en_vocab):
...
+        (1, "aabbab", "a+? b", ["a b", "a a b", "a b"]),
+        (2, "aabbab", "a*? b+?", ["a a b", "a b", "b", "b", "a b", "b"]),
+        (3, "aabbab", "a*? b*?", []),
+        (4, "aabbab", "b*? b*?", []),
+        (5, "aabbab", "b+? b+?", ["b b"]),
+        (6, "aabbab", "a* b*?", ["a", "a a", "a", "a"]),
+        (7, "aabbab", "a*? b*", ["a a b", "a b", "b", "a a b b", "a b b", "b b", "b", "a b", "b"]),
+        (8, "aabbc", "a* b*? c*?", ["a", "a a", "a"]),
+        (9, "aabbc", "a* b*? c", ["a a b b c", "a b b c", "b b c", "b c", "c"]),
+        (10, "abc", "a* b*? c*", ["a", "a b c", "b c", "c"]),
```
adrianeboyd, line 816: I'm a little confused by this test case (with four matches) vs. the comments (with three matches). I'd (naively) expect to also see a match for `"a b"` if `"b c"` is also a match?

adrianeboyd: Wait, I think might have backwards. If so, why is `"b c"` also a match?

shen-qin: Hi there, apologies for the delay, was quite preoccupied with finals and only had the time to reply now.

spaCy matcher's behaviour is slightly different from regex where the spaCy `*` implementation returns **all possible matches** instead of regex `*` returning only 1 match.
### Case of `a*b*c*` 
#### Regex matching
Screenshot 2022-12-09 at 11 46 55 PM

For example in the case of the pattern `a*b*c*` on the text `"abc"`, regular expression will return only abc as seen in the screenshot above.

#### spaCy matching
Screenshot 2022-12-08 at 1 50 31 AM
As for spaCy matcher, if we were to apply `pattern = [{'ORTH': 'a', 'OP': '*'}, {'ORTH': 'b', 'OP': '*'},{'ORTH': 'c', 'OP': '*'}]` onto ` a b c` as seen in the screenshot above, it returns **all possible matches** in contrast to regular regex.

### Case of `a*b*?c*` in my implementation
spaCy matcher will transverse over the tokens `"a b c"` with matching pattern `a*b*?c*.

Spacy should seek to return the following
```
a* | b*? | c* 
0  | ... | 0
1  | ... | 0
0  | ... | 1
1  | ... | 1
```
Where `...` is the shortest possible match, 0 is absent and 1 is present.
In the below cases, `x` represents cases where the result is removed as it did not fulfil the shortest possible match criteria.
#### matcher machine at position `"a"`
```
a b c
^
```
```
a* | b*? | c* 
0  |  0  | 0
1  |  0  | 0 --> return "a"
1  |  x  | 0 --> branch "terminated"
1  |  1  | 1 --> return "a b c"
```

#### matcher machine at position "b"
```
a b c
  ^
```
```
b*? | c* 
0   | 0
x   | 0
1   | 1 --> return "b c"
```
#### matcher machine at position "c"
```
a b c
    ^
```
```
c* 
0
1 --> return "c"
```

Hope this helps! Thanks!!!

svlandeg:
Hi @shen-qin, I want to thank you again for all the work & careful explanations you've put into this proposal.

Unfortunately, we've decided that we won't be merging this functionality into spaCy's core code base right now. We've had some internal discussion around this, but found the overal behaviour of these new quantifiers, especially when combined with spaCy's existing greedy quantifiers, simply too confusing & unexpected. The fact that they behave in an asymetric way and that you really can only understand the behaviour once you start thinking about all the details of the algorithmic implementation, makes it very difficult to come up with easy-to-explain use-cases for documentation purposes. We think it'll be too confusing to be practically useful and to warrant the maintenance costs that will be associated to having this feature.

I'm sorry that this is the conclusion we've reached, and I want to repeat that we really do appreciate your efforts on this, so we hope this won't discourage you from contributing in the future!

shen-qin:
Hi @svlandeg and team, thank you so much for the explanation. Really sad that this did not go through but I do understand. Once again thank you!

PR #11068, opened 2022-07-04
precompute_hiddens/Parser: look up CPU ops once (v4)

danieldk:
## Description

Avoid looking up CPU ops multiple times to get the CBlas struct.

### Types of change

Performance enhancement.

## Checklist

- [x] I confirm that I have the right to submit this contribution under the project's MIT license.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #11069, opened 2022-07-04
precompute_hiddens/Parser: do not look up CPU ops (3.4)

danieldk:
## Description

`get_ops("cpu")` is quite expensive. To avoid this, we want to cache the result as in #11068. However, for 3.x we do not want to change the ABI. So we avoid the expensive lookup by using NumpyOps. This should have a minimal impact, since `get_ops("cpu")` was only used when the model ops were `CupyOps`. If the ops are `AppleOps`, we are still passing through the correct BLAS implementation.

### Types of change

Performance fix.

## Checklist

### DIFF
--- a/spacy/pipeline/transition_parser.pyx
+++ b/spacy/pipeline/transition_parser.pyx
@@ -9,7 +9,7 @@ from libc.stdlib cimport calloc, free
 import random
 
 import srsly
-from thinc.api import get_ops, set_dropout_rate, CupyOps
+from thinc.api import get_ops, set_dropout_rate, CupyOps, NumpyOps
 from thinc.extra.search cimport Beam
 import numpy.random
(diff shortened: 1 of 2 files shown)


spacy/pipeline/transition_parser.pyx
```diff
@@ -30,6 +30,9 @@ from ..errors import Errors, Warnings
 from .. import util
 
 
+_NUMPY_OPS = NumpyOps()
```
svlandeg, line 33: I suggest to call this `NUMPY_OPS` for consistency across Thinc & spaCy? We've never used the underscore before...

danieldk: I'll change this. I though to underscore it to prevent this to be considered public.

PR #11076, opened 2022-07-05
Add Softmax scorer for spancat

ljvmiranda921:
**WIP**: This PR adds a Softmax scorer for the spancat component that should be used in cases where classes are exclusive. 

## Description

Spancat currently has a `LinearLogistic` layer that creates multilabel annotations on text. However, this is not necessary when classes are exclusive or when spans are non-overlapping. Switching from a `LinearLogistic` to a `Softmax` layer is recommended. This issue came up when we're doing the spancat experiments. 

### Usage

In order for the user to switch to `spacy.Softmax.v1`, they have to update the config and specify the architecture in the `scorer`. 
**Alternative usage (not yet implemented)**: Add a parameter, `exclusive_classes:bool` (a la `textcat`), in the `SpanCategorizer` architecture to use Softmax automatically. 

### Types of change
New feature

## Checklist

### DIFF
--- a/spacy/ml/models/spancat.py
+++ b/spacy/ml/models/spancat.py
@@ -1,11 +1,13 @@
 from typing import List, Tuple, cast
-from thinc.api import Model, with_getitem, chain, list2ragged, Logistic
-from thinc.api import Maxout, Linear, concatenate, glorot_uniform_init
-from thinc.api import reduce_mean, reduce_max, reduce_first, reduce_last
-from thinc.types import Ragged, Floats2d
 
-from ...util import registry
+from thinc.api import Linear, Logistic, Maxout, Model, Softmax, chain
+from thinc.api import concatenate, glorot_uniform_init, list2ragged
+from thinc.api import reduce_first, reduce_last, reduce_max, reduce_mean
+from thinc.api import with_getitem
+from thinc.types import Floats2d, Ragged
+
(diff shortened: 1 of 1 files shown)


spacy/ml/models/spancat.py
```diff
@@ -17,6 +19,14 @@ def build_linear_logistic(nO=None, nI=None) -> Model[Floats2d, Floats2d]:
     return chain(Linear(nO=nO, nI=nI, init_W=glorot_uniform_init), Logistic())
 
 
+@registry.layers("spacy.Softmax.v1")
```
adrianeboyd, line 22: Can't you use the `Softmax.v1` layer from thinc without needing an extra definition here?

svlandeg:
> In order for the user to switch to `spacy.Softmax.v1`, they have to update the config and specify the architecture in the `scorer`. **Alternative usage (not yet implemented)**: Add a parameter, `exclusive_classes:bool` (a la `textcat`), in the `SpanCategorizer` architecture to use Softmax automatically.

It would make sense to make the implementation & usage consistent with how we do it in `textcat`.

svlandeg:
After internal conversations: we'll want to implement this as a new component, somewhat similar to `textcat` vs `textcat_multilabel`. I'll close this PR while we work on a more extensive PR to address that.

PR #11087, opened 2022-07-06
Update Coref Docs

polm:
## Description

This updates a few areas in the coref docs that had not been filled in or had gotten out of sync. There is also some cleanup of old TODOs and error messages. This should not affect any tests.

### Types of change

Cleanup.

## Checklist

- [x] I confirm that I have the right to submit this contribution under the project's MIT license.
- [ ] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #11089, opened 2022-07-06
Dimension inference in Coref

polm:
## Description

This PR is for doing dimension inference in coref instead of requiring manually specifying input size. PyTorch based layers need to know their size when their `__init__` function is called or serialization doesn't work. 

The current status is that for the coref component this works now, but serialization still doesn't work. If this can be figured out the span predictor will still need similar changes.

## Checklist


polm:
As a note, the failing tests here are just type issues, which should be resolved by #11087.

polm:
With the other PRs merged, I'll run the tests here before merging it in.

polm:
This PR added some variables that didn't have types, which caused mypy tests to fail. 

Besides that a textcat test that shouldn't be affected by any of these PRs seems to be failing. I'll look into that more.

polm:

polm:

polm:
OK, tests passed, so merging to get all changes in one place.

PR #11090, opened 2022-07-06
Github Action for spaCy universe project alert

nipunsadvilkar:
## Description

This is internal requirement of Community Team @ Explosion.

We want to have an alert system to get list of new spaCy universe projects when a PR is raised against spaCy repo involving changes to `website/meta/universe.json` file only. 

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -2943,6 +2943,7 @@
                 "from pysbd.utils import PySBDFactory",
                 "",
                 "nlp = spacy.blank('en')",
+                "# works with spaCy<=2.x.x",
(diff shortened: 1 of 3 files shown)


website/meta/universe.json
```diff
@@ -2943,6 +2943,7 @@
                 "from pysbd.utils import PySBDFactory",
                 "",
                 "nlp = spacy.blank('en')",
+                "# works with spaCy<=2.x.x",
```
svlandeg, line 2946: It would be better to annotate this with the `spacy_version` meta data as well (this is a new feature that ideally we'd add for all the entries). Maybe better to do that in a separate PR?

PR #11119, opened 2022-07-12
Fix span typings

### LINKED ISSUE #11118: Span.id_ and Span.id missing from Span typings
`Span` is missing `id` and `id_` attribute in [`span.pyi`](https://github.com/explosion/spaCy/blob/master/spacy/tokens/span.pyi#L120) typings

## Info about spaCy

- **spaCy version:** 3.3.1
- **Platform:** macOS-12.4-arm64-arm-64bit
- **Python version:** 3.9.9

mr-bjerre:
## Description

`Span` is missing `id` and `id_` attributes in the typing file, i.e. `span.pyi`. But it is [available properties](https://github.com/explosion/spaCy/blob/master/spacy/tokens/span.pyx#L784)

Fixes #11118

### Types of change
Bug fix on typings

## Checklist

### DIFF
--- a/spacy/tokens/span.pyi
+++ b/spacy/tokens/span.pyi
@@ -119,6 +119,8 @@ class Span:
     kb_id: int
     ent_id: int
     ent_id_: str
+    id: int
+    id_: str
     @property
     def orth_(self) -> str: ...
     @property
--- a/spacy/tokens/span.pyi
+++ b/spacy/tokens/span.pyi
@@ -119,8 +119,10 @@ class Span:
     kb_id: int
     ent_id: int
     ent_id_: str
-    id: int
-    id_: str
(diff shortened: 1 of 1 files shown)


spacy/tokens/span.pyi
```diff
@@ -119,6 +119,8 @@ class Span:
     kb_id: int
     ent_id: int
     ent_id_: str
+    id: int
+    id_: str
```
rmitsch, line 123: Since `id` and `ìd_` are properties, it's cleaner to declare them like this:
```suggestion
    @property
    def id(self) -> int: ...
    @property
    def id_(self) -> str: ...
```
(We are a bit inconsistent here with the other properties, but will clean that up.)

PR #879, opened 2017-03-09
Fix for Issue #840 - URL pattern too broad

rappdw:
Changed _URL_PATTERN in language/tokenizer_exceptions.py, using url regex validation pattern from: https://mathiasbynens.be/demo/url-regex

## Description

This change will screen out a number of patterns that are not URLs as well as picking up some patterns that were missed.
Test cases were added to test_urls to expose the issue and validate the fix.

## Types of changes

- [ x] **Bug fix** (non-breaking change fixing an issue)
- [ ] **New feature** (non-breaking change adding functionality to spaCy)
- [ ] **Breaking change** (fix or feature causing change to spaCy's existing functionality)
- [ ] **Documentation** (addition to documentation of spaCy)

## Checklist:

- [ ] My change requires a change to spaCy's documentation.
- [ ] I have updated the documentation accordingly.
- [x ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #840, opened 2017-02-16
URL regex in tokenizer_exceptions.py too broad


rappdw:
The regex for the URL used for token_match during tokenization treats strings that are not URLs as URLs. For example, I'd expect the following text:

"This is the ticker symbol for Google, (NASDAQ:GOOG). Google\'s homepage is http://www.google.com"

to produce 'NASDAQ', ':', 'GOOG' as separate tokens while producing 'http://www.google.com' as a single token.

Using the URL regex proposed by https://gist.github.com/dperini/729294 yields better results, e.g.

    token_match = re.compile(
        r"^"
        # protocol identifier
        r"(?:(?:https?|ftp)://)"
        # user:pass authentication
        r"(?:\S+(?::\S*)?@)?"
        r"(?:"
        # IP address exclusion
        # private & local networks
        r"(?!(?:10|127)(?:\.\d{1,3}){3})"
        r"(?!(?:169\.254|192\.168)(?:\.\d{1,3}){2})"
        r"(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})"
        # IP address dotted notation octets
        # excludes loopback network 0.0.0.0
        # excludes reserved space >= 224.0.0.0
        # excludes network & broadcast addresses
        # (first & last IP address of each class)
        r"(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])"
        r"(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}"
        r"(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))"
        r"|"
        # host name
        r"(?:(?:[a-z\u00a1-\uffff0-9]-?)*[a-z\u00a1-\uffff0-9]+)"
        # domain name
        r"(?:\.(?:[a-z\u00a1-\uffff0-9]-?)*[a-z\u00a1-\uffff0-9]+)*"
        # TLD identifier
        r"(?:\.(?:[a-z\u00a1-\uffff]{2,}))"
        r")"
        # port number
        r"(?::\d{2,5})?"
        # resource path
        r"(?:/\S*)?"
        r"$"
    ).match

## Your Environment
* Operating System: OSX & Linux
* Python Version Used: 3.6.0 
* spaCy Version Used: 1.6.0
* Environment Information:

honnibal:
Thanks, we'll definitely patch this case.

[...]

PR #11121, opened 2022-07-12
Match private networks as URLs

antonpibm:
Currently, URLs under private networks won't be matched as URLs by the tokenizer
e.g.
```
https://10.140.12.13/foo/bar?arg1=baz&arg2=taz # not a URL
https://142.140.12.13/foo/bar?arg1=baz&arg2=taz # A URL
```

From my understanding, at some point the URL matching was changed,
initial issue: https://github.com/explosion/spaCy/issues/840
relevant PR: https://github.com/explosion/spaCy/pull/879

It looks like the regex was taken from some validation code, where maybe it made sense to neglect private networks, but let me argue that for our usage it doesn't. In my usecase I have a dump of corporate logs where some of the systems URLs are reported as their private networks, one can refer to the URL structure's [Wikipedia page](https://en.wikipedia.org/wiki/URL) and see that there is no exception for private networks. 

In terms of NLP, both examples above are URLs, but only one would be matched as single token, the prefix of URL with `10.140.12.13` won't even have the `like_url` attribute set to true (the token is `https://10.140.12.13`).

**Note**: I haven't removed the exception for broadcast addresses as they are unlikely to appear as URL, but I don't know whether a NLP lib should make this validation or also match the broadcast address.

### Types of change

enhancement

## Checklist


rmitsch:
Thanks for submitting this PR! We'll evaluate the changes.

svlandeg:
Hi @antonpibm, thanks again for the PR! As the changes could influence tokenization in existing pipelines/applications, we've decided to add these changes to the `v4` branch that we're using to prepare the upcoming 4.0.0. We'll also be looking into some other exceptions that perhaps can be removed - if you'd find any other odd cases we'd be happy to consider those as well!

We'll merge this as soon as the tests go green again.

antonpibm:
@svlandeg thx for the update. Would gladly help with any relevant consideration

PR #11123, opened 2022-07-12
Matchers documentation update (error in example)

radenkovic:
## Description
Simple update of documentation (example for Matchers) to reflect the code.  Function `on_match` in the example was not called, therefore the sample was not doing anything. This PR fixes it (added necessary code in the example).

```python
def on_match(matcher, doc, id, matches): # ⚠️ NOTE: this function is unused in the current example
    print('Matched!', matches)

matcher = Matcher(nlp.vocab)
patterns = [
   [{"LOWER": "hello"}, {"LOWER": "world"}],
   [{"ORTH": "Google"}, {"ORTH": "Maps"}]
]
matcher.add("TEST_PATTERNS", patterns) #  ⚠️ NOTE: on_match argument is not passed! (fixed in the PR)
doc = nlp("HELLO WORLD on Google Maps.")
matches = matcher(doc)
```

### Types of change
Documentation update.

## Checklist

### DIFF
--- a/website/docs/api/matcher.md
+++ b/website/docs/api/matcher.md
@@ -199,7 +199,7 @@ will be overwritten.
 >    [{"LOWER": "hello"}, {"LOWER": "world"}],
 >    [{"ORTH": "Google"}, {"ORTH": "Maps"}]
 > ]
-> matcher.add("TEST_PATTERNS", patterns)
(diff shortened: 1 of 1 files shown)


polm:
Thanks for the PR. In sample code it's generally fine if it doesn't do anything, since it's just to give a general idea of how something works, but this is better if it shows how to actually pass an `on_match` function like in your modifications.

PR #11125, opened 2022-07-13
Support custom attributes for tokens and spans in json conversion

thomashacker:
## Description

Currently, `Doc.to_json()` and `Doc.from_json()` only support custom attributes set on the whole `Doc` object. This `PR` adds support for custom attributes in tokens and spans.

### Types of change

New Feature

## Checklist


spacy/tokens/doc.pyx
```diff
@@ -1602,12 +1602,26 @@ cdef class Doc:
                 ents.append(char_span)
             self.ents = ents
 
-        # Add custom attributes. Note that only Doc extensions are currently considered, Token and Span extensions are
-        # not yet supported.
-        for attr in doc_json.get("_", {}):
+        # Add custom attributes for the whole doc
+        for attr in doc_json["_"]["docs"]:
```
shadeMe, line 1606: Will the index/subscript operator return an empty iterator if `docs` isn't found in `_` (or if `_` isn't found)? If not, we'll want to revert to using the `get` method with an empty dict as the default value.

thomashacker: Yes, you're exactly right! Will be changed in the next commit

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -57,18 +99,68 @@ def test_doc_to_json(doc):
     assert json_doc["ents"][0]["end"] == 3  # character offset!
     assert json_doc["ents"][0]["label"] == "ORG"
     assert not schemas.validate(schemas.DocJSONSchema, json_doc)
+    assert srsly.json_loads(srsly.json_dumps(json_doc))
```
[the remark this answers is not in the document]
thomashacker: Oh, you're right!

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -57,18 +99,68 @@ def test_doc_to_json(doc):
...
+    assert json_doc["_"]["json_test1"] == "hello world"
+    assert json_doc["_"]["json_test2"] == [1, 2, 3]
+    assert "user_data" in json_doc
+    assert json_doc["user_data"]["token_test"]["attribute_value"] == 117
+    assert json_doc["user_data"]["span_test"]["attribute_value"] == "span_attribute"
+    assert not schemas.validate(schemas.DocJSONSchema, json_doc)
+    assert srsly.json_loads(srsly.json_dumps(json_doc)) == json_doc
+
+
+def test_doc_to_json_with_token_span_attributes_missing(doc):
```
svlandeg, line 150: Nitpick on the name: should it be "token_attribute_missing"? Because the span attribute is included, right?

svlandeg (commented):
Nice work Edi! I particularly like all the unit tests & extensions :-) 

Just had a few minor comments/questions.

spacy/tokens/doc.pyx
```diff
@@ -1650,22 +1666,36 @@ cdef class Doc:
...
+            user_keys = []
+            if self.user_data:
+                data["user_data"] = {}
+                for data_key in self.user_data:
+                    attr = data_key[1]
+                    if attr not in data["user_data"] and attr in underscore and data_key[2] is not None:
+                        value = self.user_data[data_key]
+                        if not srsly.is_json_serializable(value):
+                            raise ValueError(Errors.E107.format(attr=attr, value=repr(value)))
+                        data["user_data"][attr] = {"start_index": data_key[2], "end_index": data_key[3], "attribute_value":value}
```
adrianeboyd, line 1682: Let me think about the naming here a bit...

spacy/tokens/doc.pyx
```diff
@@ -1650,22 +1666,36 @@ cdef class Doc:
...
+                    span_data = {"start": span.start_char, "end": span.end_char, "label": span.label_, "kb_id": span.kb_id_}
                     data["spans"][span_group].append(span_data)
 
         if underscore:
+            user_keys = []
+            if self.user_data:
+                data["user_data"] = {}
+                for data_key in self.user_data:
+                    attr = data_key[1]
+                    if attr not in data["user_data"] and attr in underscore and data_key[2] is not None:
```
adrianeboyd, line 1678: This check for `attr not in data["user_data"]` isn't doing the right thing for the common case where you've defined an extension with the same name for `Token`, `Span`, and `Doc`.

Consider:

```python
Token.set_extension("my_ext", default=0)
Span.set_extension("my_ext", default=0)
Doc.set_extension("my_ext", default=0)
doc = nlp("cats")
doc._.my_ext = 1
doc[0:1]._.my_ext = 2
doc[0]._.my_ext = 3
```

produces:

```
{
  "text": "cats",
  "sents": [
    {
      "start": 0,
      "end": 4
    }
  ],
  "tokens": [
    {
      "id": 0,
      "start": 0,
      "end": 4
    }
  ],
  "_": {
    "my_ext": 1
  }
}
```

You only get one of them in the output here.

thomashacker:
I've adjusted the naming and ensured that added keys in the future wouldn't break the code. I also ensured that the code checks whether the `user_data` is a `tuple` + attributes with the same name for `Doc`, `Token`, and `Span` are now supported. 

For the new logic, I also added some additional tests.

spacy/tokens/doc.pyx
```diff
@@ -1650,22 +1669,48 @@ cdef class Doc:
...
+                    if not srsly.is_json_serializable(value):
+                        raise ValueError(Errors.E107.format(attr=attr, value=repr(value)))
+                    data["_"][attr] = value
+                elif attr not in user_keys:
                     raise ValueError(Errors.E106.format(attr=attr, opts=underscore))
-                value = self._.get(attr)
-                if not srsly.is_json_serializable(value):
-                    raise ValueError(Errors.E107.format(attr=attr, value=repr(value)))
-                data["_"][attr] = value
+
```
adrianeboyd, line 1710:
```suggestion
```

spacy/tokens/doc.pyx
```diff
@@ -1650,22 +1669,48 @@ cdef class Doc:
...
+                                doc_keys.add(attr)
+                                continue
+                            # Check if token attribute
+                            elif end is None:
+                                if attr not in data["user_data"]["tokens"]:
+                                    data["user_data"]["tokens"][attr] = {"token_start": start, "value":value}
+                            # Else span attribute
+                            else:
+                                if attr not in data["user_data"]["spans"]:
+                                    data["user_data"]["spans"][attr] = {"token_start": start, "token_end": end, "value":value}
```
adrianeboyd, line 1698: I would represent both token and span as the same way they are internally with `token_end` as `None` for tokens.

spacy/tokens/doc.pyx
```diff
@@ -1650,22 +1669,48 @@ cdef class Doc:
...
+            user_keys = set()
+            if self.user_data:
+                data["user_data"] = {"spans":{}, "tokens":{}}
+                for data_key in self.user_data:
+                    if type(data_key) == tuple and data_key[0] == "._.":
+                        _, attr, start, end = data_key
+                        if attr in underscore:
+                            value = self.user_data[data_key]
+                            if not srsly.is_json_serializable(value):
+                                raise ValueError(Errors.E107.format(attr=attr, value=repr(value)))
```
adrianeboyd, line 1686: `Errors.E107` needs to be updated to be more general.

spacy/tokens/doc.pyx
```diff
@@ -1650,22 +1669,48 @@ cdef class Doc:
...
+                            user_keys.add(attr)
+
             data["_"] = {}
             for attr in underscore:
-                if not self.has_extension(attr):
+                if self.has_extension(attr):
+                    value = self._.get(attr)
+                    if not srsly.is_json_serializable(value):
+                        raise ValueError(Errors.E107.format(attr=attr, value=repr(value)))
+                    data["_"][attr] = value
```
adrianeboyd, line 1707: I think this can combined with the previous loop?

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -57,18 +99,116 @@ def test_doc_to_json(doc):
...
+    doc[0]._.my_ext = 117
+    json_doc = doc.to_json(underscore=["my_ext"])
+
+    assert "_" in json_doc
+    assert json_doc["_"]["my_ext"] == "hello world"
+    assert "user_data" in json_doc
+    assert json_doc["user_data"]["tokens"]["my_ext"]["value"] == 117
+    assert json_doc["user_data"]["spans"]["my_ext"]["value"] == "span_attribute"
+    assert not schemas.validate(schemas.DocJSONSchema, json_doc)
+    assert srsly.json_loads(srsly.json_dumps(json_doc)) == json_doc
```
adrianeboyd, line 195: For all the tests that include all the available extensions, I'd suggest adding a full doc round-trip that checks `doc.to_bytes() == new_doc.to_bytes`.

spacy/schemas.py
```diff
@@ -517,3 +517,7 @@ class DocJSONSchema(BaseModel):
     _: Optional[Dict[StrictStr, Any]] = Field(
         None, title="Any custom data stored in the document's _ attribute"
     )
+    user_data: Optional[Dict[StrictStr, Dict[StrictStr, Any]]] = Field(
+        None, title="Any user data stored in the document"
```
adrianeboyd, line 521:
```suggestion
        None, title="Any custom data stored in token and span extensions"
```

spacy/schemas.py
```diff
@@ -517,3 +517,7 @@ class DocJSONSchema(BaseModel):
...
+    )
+
```
adrianeboyd, line 523:
```suggestion
```

spacy/tokens/doc.pyx
```diff
@@ -1602,13 +1602,32 @@ cdef class Doc:
...
+                token_end = doc_json["user_data"]["spans"][span_attr]["token_end"]
+                value = doc_json["user_data"]["spans"][span_attr]["value"]
+
+                if not Span.has_extension(span_attr):
+                    Span.set_extension(span_attr)
+                self[token_start:token_end]._.set(span_attr, value)
+
+       
+
         return self
```
adrianeboyd, line 1631:
```suggestion
        return self
```

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -57,18 +99,116 @@ def test_doc_to_json(doc):
     assert json_doc["ents"][0]["end"] == 3  # character offset!
     assert json_doc["ents"][0]["label"] == "ORG"
     assert not schemas.validate(schemas.DocJSONSchema, json_doc)
+    assert srsly.json_loads(srsly.json_dumps(json_doc)) == json_doc
 
 
 def test_doc_to_json_underscore(doc):
-    Doc.set_extension("json_test1", default=False)
-    Doc.set_extension("json_test2", default=False)
+    if not Doc.has_extension("json_test1"):
```
adrianeboyd, line 106: Instead of checking for the presence of extensions repeatedly, use this as an import and it will automatically clear after each function:

```
from .test_underscore import clean_underscore  # noqa: F401
```

spacy/tokens/doc.pyx
```diff
@@ -1650,22 +1669,48 @@ cdef class Doc:
...
                     data["spans"][span_group].append(span_data)
 
         if underscore:
+            doc_keys = set()
+            user_keys = set()
+            if self.user_data:
+                data["user_data"] = {"spans":{}, "tokens":{}}
+                for data_key in self.user_data:
+                    if type(data_key) == tuple and data_key[0] == "._.":
+                        _, attr, start, end = data_key
```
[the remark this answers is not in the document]
thomashacker: Ah, sorry! I've misread that

spacy/tests/doc/test_json_doc_conversion.py
thomashacker: This actually throws assertion errors

thomashacker: I also added the assert line to the `test_json_to_doc` which doesn't include any custom attributes. Would this indicate that something is wrong with `to_json()` / `from_json()` or rather `to_bytes()` ?

adrianeboyd: Can you figure out what the differences are between the docs?

thomashacker: As far as I can tell, this seems to be the only difference between the two docs.

```
E         -  b'\xa1d\xa3ORG\xa4ROOT\xa7Feat1=B\xb2has_unknown_spaces\xc2',
E         ?                                                            ^
E         +  b'\xa1d\xa3ORG\xa4ROOT\xa7Feat1=B\xb2has_unknown_spaces\xc3',
E         ?                                                            ^
```

thomashacker: The end byte of the doc seems to differ, maybe related to encoding?

adrianeboyd: I can't tell which one, but it looks like one of them was created with `spaces` and one without. I'm not sure if it's a problem in the test doc/fixture or in `Doc.from_json`.

`Doc.to_json` doesn't really handle `has_unknown_spaces=True`, though, so it could be a more general problem.

adrianeboyd: It's not the encoding, it's `has_unknown_spaces`. I think we can pretty much that we only want to handle `has_unknown_spaces=False`, but I'm not sure how far we want to enforce it. For the tests it should be possible create input/output docs that both have `has_unknown_spaces=False`.

thomashacker: I got it fixed by setting both `has_unknown_spaces=False`. However, there is still a different assertion error when comparing both `to_bytes()`

```
E         -  b'r_data_values\xc4!\x94\xabhello world\x93\x01\x02\x03u\xaespan_attribute',
E         ?                                                         -
E         +  b'r_data_values\xc4!\x94\xabhello world\x93\x01\x02\x03\xaespan_attributeu',
E         ? 
```      

It has to do something with the custom span attribute, the `new_doc` seems to add an `u` at the end? I haven't figured out yet what it means / how it is caused.

spacy/tokens/doc.pyx
```diff
@@ -1650,20 +1666,40 @@ cdef class Doc:
...
+                            if start is None:
+                                data["_"][attr] = value
+                            # Check if token attribute
+                            elif end is None:
+                                if attr not in data["user_data"]["tokens"]:
+                                    data["user_data"]["tokens"][attr] = {"token_start": start, "token_end": end, "value":value}
+                            # Else span attribute
+                            else:
+                                if attr not in data["user_data"]["spans"]:
+                                    data["user_data"]["spans"][attr] = {"token_start": start, "token_end": end, "value":value}
```
adrianeboyd, line 1697: You shouldn't need the `"tokens"` and `"spans"` keys here?

adrianeboyd: I'm also having a naming crisis about `user_data` here again.

thomashacker: I added the "tokens" / "spans" keys because they can have custom extensions with the same identifier, so my initial thought was to separate both

adrianeboyd: But you can distinguish them by the `end_token` value just like it's stored internally? Or there's no need for `token_end` for the token spans?

Let me continue my naming crisis, too (with a side of "this looks like prodigy except that `token_end` is inclusive in prodigy" on top).

I wouldn't want to call it this except for `underscore` in the kwarg, but let me think a bit...

In an ideal world we would have:

```
data["_"][attr] = [{"token_start": , "token_end": , "value": }, ]
```

We are not in this ideal world because we are stuck with:

```
data["_"][attr] = doc_ext_value
```

So something that doesn't overlap with other uses of `user_data` but still does something similar:

```
data["underscore_token"][attr] = [{"token_start": , "value": }, ]
data["underscore_span"][attr] = [{"token_start": , "token_end": , "value": }, ]
```

This leaves room to actually add other `user_data` in the future (although I doubt we would).

thomashacker: I agree, I've implemented your suggestion and adjusted all tests accordingly

spacy/schemas.py
```diff
@@ -517,3 +517,6 @@ class DocJSONSchema(BaseModel):
     _: Optional[Dict[StrictStr, Any]] = Field(
         None, title="Any custom data stored in the document's _ attribute"
     )
+    user_data: Optional[Dict[StrictStr, Dict[StrictStr, Any]]] = Field(
```
adrianeboyd, line 520: The schema wasn't updated but all the tests pass. There should be a schema test that is expected to pass with all types of extensions.

spacy/tokens/doc.pyx
```diff
@@ -1650,20 +1667,40 @@ cdef class Doc:
...
+                            value = self.user_data[data_key]
+                            if not srsly.is_json_serializable(value):
+                                raise ValueError(Errors.E107.format(attr=attr, value=repr(value)))
+                            # Check if doc attribute
+                            if start is None:
+                                data["_"][attr] = value
+                            # Check if token attribute
+                            elif end is None:
+                                if attr not in data["underscore_token"]:
+                                    data["underscore_token"][attr] = {"token_start": start, "token_end": end, "value":value}
```
adrianeboyd, line 1695:
```suggestion
                                    data["underscore_token"][attr] = {"token_start": start, "value":value}
```

spacy/schemas.py
thomashacker: What would be the best way to write a unit test for that? Since these fields are `Optional`?

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -57,18 +102,104 @@ def test_doc_to_json(doc):
...
+
+def test_doc_to_json_with_token_span_same_identifier(doc):
+    Doc.set_extension("my_ext", default=False)
+    Token.set_extension("my_ext", default=False)
+    Span.set_extension("my_ext", default=False)
+
+    doc._.my_ext = "hello world"
+    doc[0:1]._.my_ext = "span_attribute"
+    doc[0]._.my_ext = 117
+    json_doc = doc.to_json(underscore=["my_ext"])
```
adrianeboyd, line 178: This will fail until the schema is updated:

```suggestion
    json_doc = doc.to_json(underscore=["my_ext"])
    
    assert schemas.validate(schemas.DocJSONSchema, json_doc)
```

spacy/schemas.py
```diff
@@ -517,3 +517,9 @@ class DocJSONSchema(BaseModel):
     _: Optional[Dict[StrictStr, Any]] = Field(
         None, title="Any custom data stored in the document's _ attribute"
     )
+    underscore_token: Optional[Dict[StrictStr, Any]] = Field(
+        None, title="Any custom data stored in the tokens's _ attribute"
```
adrianeboyd, line 521:
```suggestion
        None, title="Any custom data stored in the token's _ attribute"
```

spacy/schemas.py
```diff
@@ -517,3 +517,9 @@ class DocJSONSchema(BaseModel):
...
+    )
+    underscore_span: Optional[Dict[StrictStr, Any]] = Field(
+        None, title="Any custom data stored in the spans's _ attribute"
```
adrianeboyd, line 524:
```suggestion
        None, title="Any custom data stored in the span's _ attribute"
```

spacy/tokens/doc.pyx
```diff
@@ -1650,20 +1667,40 @@ cdef class Doc:
...
+                            value = self.user_data[data_key]
+                            if not srsly.is_json_serializable(value):
+                                raise ValueError(Errors.E107.format(attr=attr, value=repr(value)))
+                            # Check if doc attribute
+                            if start is None:
+                                data["_"][attr] = value
+                            # Check if token attribute
+                            elif end is None:
+                                if attr not in data["underscore_token"]:
+                                    data["underscore_token"][attr] = {"token_start": start, "value":value}
```
adrianeboyd, line 1695:
```suggestion
                                    data["underscore_token"][attr] = {"token_start": start, "value": value}
```

spacy/tokens/doc.pyx
```diff
@@ -1650,20 +1667,40 @@ cdef class Doc:
...
+                            # Else span attribute
+                            else:
+                                if attr not in data["underscore_span"]:
+                                    data["underscore_span"][attr] = {"token_start": start, "token_end": end, "value":value}
```
adrianeboyd, line 1699:
```suggestion
                                    data["underscore_span"][attr] = {"token_start": start, "token_end": end, "value": value}
```

spacy/tests/doc/test_json_doc_conversion.py
thomashacker: I looked into the `validate` command, [which returns an empty `list` if the schema is correct](https://github.com/explosion/spaCy/blob/master/spacy/schemas.py#L42). However, `assert []` causes an `assertion error` even though the schema is correct. 

I think the assertion test should be:
`assert not schemas.validate(schemas.DocJSONSchema, json_doc)` or 
`assert schemas.validate(schemas.DocJSONSchema, json_doc) == []`

adrianeboyd: Ah, I misunderstood how the validation works, and I also thought that extra fields were forbidden when they aren't. I'm not sure we want to make extra fields forbidden, but based on trying things out it turns out that `_` causes problems in the schema, so maybe this would be better:

```
    underscore_doc: Optional[Dict[StrictStr, Any]] = Field(
        None, title="Any custom data stored in the document's _ attribute",
        alias="_",
    )
```

thomashacker: Updated the code 👍

adrianeboyd: Can you still add the right version of this validation to the test?

adrianeboyd: I'm also a little concerned that the schema makes it look like it supports `underscore_doc` when it doesn't, but otherwise the schema kind of doesn't work properly. And we do the same thing in `TokenPattern`, too.

thomashacker: What do you mean by "right version"? 😇

adrianeboyd: Whatever works at this line above if the validation passes.

thomashacker: Do you mean this validation line?
`assert schemas.validate(schemas.DocJSONSchema, json_doc)`

adrianeboyd: And I think we should change all the validation lines in the tests because this is too confusing:

```python
assert not schemas.validate()
```

Maybe:

```python
assert len(schemas.validate()) == 0
```

PR #11127, opened 2022-07-14
[Refactoring]Removing for loops and using filter instead

maldil:
## Description 
This is a code refactoring to make the code more concise by removing `for loops` and using the filter operator instead. 

### Types of change
Not a new feature or bug fixing. This is a code refactoring.

## Checklist

### DIFF
--- a/spacy/lang/es/lemmatizer.py
+++ b/spacy/lang/es/lemmatizer.py
@@ -351,10 +351,9 @@ class SpanishLemmatizer(Lemmatizer):
             possible_lemma = re.sub(old + "$", new, word)
             if possible_lemma != word:
                 possible_lemmas.append(possible_lemma)
+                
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Hi, thanks for taking a closer look at the code here! One of spacy's code conventions is to avoid `filter` in favor of other alternatives. Some more detailed code conventions are in the developer docs here: https://github.com/explosion/spaCy/blob/master/extra/DEVELOPER_DOCS/Code%20Conventions.md#iteration-and-comprehensions

I'll add that I expected to find a link to this in `CONTRIBUTING.md` but it seems to be missing, which is an oversight on our part.

adrianeboyd:
I'll close this PR for now. We'll add a link to the detailed code conventions in #11171.

PR #11159, opened 2022-07-19
Added BERTopic to Spacy Universe

MaartenGr:
## Description
A proposal to update `universe.json` file to add [BERTopic](https://github.com/MaartenGr/BERTopic) to the spaCy Universe.

### Types of change
Documentation change

## Checklist

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -564,7 +564,8 @@
                 "topic_model = BERTopic(embedding_model=nlp)",
                 "topics, probs = topic_model.fit_transform(docs)",
                 "",
-                "topic_model.visualize_topics()"
+                "fig = topic_model.visualize_topics()",
+                "fig.show()"
             ],
             "category": ["visualizers", "training"],
             "author": "Maarten Grootendorst",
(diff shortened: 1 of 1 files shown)


polm:
Thanks for the submission, this looks good!

I have one question - I ran this on a remote machine I was using via SSH, and the script just doesn't output anything. I think if it's a local machine the visualization method is supposed to open a window or something, but is it expected for it to just do nothing if there's no X session?

MaartenGr:
@polm Ah, I totally forgot about that. The visualization function returns a plotly figure `plotly.graph_objects.Figure`. Since most users of BERTopic run this in a notebook, the visualization automatically gets rendered. This is not the case when you run it from the terminal or in a different IDE as you may need to do the following instead:

```python
fig = topic_model.visualize_topics()
fig.show()
```

For demonstration purposes, the above might be a bit more robust across environments/IDEs. What do you think?

polm:
Using `fig.show()` sounds good - in a situation like my SSH one (which I don't think is common) it'll give a clear error too.

PR #11162, opened 2022-07-19
Updates to Slovenian language

orglce:
## Description
This is a general PR for the support of Slovenian language. 

### Types of change

- **stop_words.py** has been updated from [this list](https://github.com/stopwords-iso/stopwords-sl)
- **lex_attributes.py** contains a semi exhaustive list of numbers and currencies in different cases (could contain more, but i'm unsure of how that would impact performance)
- **punctuation.py** and **tokenizer_exceptions.py** have been updated to more closely follow the tokenization of the Slovenian gold standard dataset [UD_SSJ](https://github.com/UniversalDependencies/UD_Slovenian-SSJ) 

The dataset also considers roman numerals as a single token, but I have been unable to figure out how to include that regex in the punctuation.py. 

## Checklist


richardpaulhudson:
Thanks for the PR, I shall have a look over the next few days.

richardpaulhudson:
Thank you very much indeed for this PR @orglce, it will be great to have better support for Slovenian!

I'll add some specific comments in a minute, but one general request is whether you could possibly add some tests? They would be especially helpful for features that you have developed from scratch rather than copying them from other languages. Examples of such features are the `HYPHENS_PERMITTED` list or the splitting of initials in names like `J.K. Rowling`. As well as proving that the features work as expected, tests will also help people who might come in the future to understand them. A good starting point for tests is probably the langauge-specific test directory for a well-resourced language such as English (https://github.com/explosion/spaCy/tree/master/spacy/tests/lang/en): you can copy individual tests and modify them as required.

spacy/lang/sl/lex_attrs.py
```diff
@@ -0,0 +1,116 @@
+from ...attrs import LIKE_NUM
+from ...attrs import IS_CURRENCY
+import unicodedata
+
+
+_num_words = set(
```
richardpaulhudson, line 6: I get your point about performance, but I feel at least the inflected forms of the numbers up to 10 should be included here, e.g. `dveh` `trem` `petimi`.

spacy/lang/sl/punctuation.py
```diff
@@ -0,0 +1,64 @@
+from ..char_classes import LIST_ELLIPSES, LIST_ICONS, HYPHENS, LIST_PUNCT, LIST_QUOTES, CURRENCY, UNITS, PUNCT, \
+    LIST_CURRENCY, CONCAT_QUOTES
+from ..char_classes import CONCAT_QUOTES, ALPHA_LOWER, ALPHA_UPPER, ALPHA
+from ..char_classes import merge_chars
+from ..punctuation import TOKENIZER_PREFIXES as BASE_TOKENIZER_PREFIXES
+
+_prefixes = ["•", "\-", "\/", "'"] + BASE_TOKENIZER_PREFIXES
+
+_suffixes = (
+        ["—", "–", "\-", "\:", "\+", "\/", "=", "×", "\$", "•", "¯", "%"]
```
richardpaulhudson, line 10: These should either be double-escaped (e.g. `"\\+"` rather than `"\+"`) or the strings should be preceded with `r`, e.g. `r"\+"`

richardpaulhudson, line 10: There seems to be some overlap with `LIST_PUNCT`, e.g. `:`.

spacy/lang/sl/lex_attrs.py
```diff
@@ -0,0 +1,116 @@
...
+)
+
+
+def like_num(text):
+    if text.startswith(("+", "-", "±", "~")):
+        text = text[1:]
+    text = text.replace(",", "").replace(".", "")
+    if text.isdigit():
+        return True
+    if text.count("/") == 1:
```
richardpaulhudson, line 95: As far as I can see, this code can never be executed because numbers like `2/3` are split into three parts by the infix rules.

spacy/lang/sl/punctuation.py
```diff
@@ -0,0 +1,64 @@
...
+        + LIST_ELLIPSES
+        + LIST_QUOTES
+        + LIST_ICONS
+        + [
+            r"(?<=°[FfCcKk])\.",
+            r"(?<=[0-9])(?:{c})".format(c=CURRENCY),
+            r"(?<=[0-9])(?:{u})".format(u=UNITS),
+            r"(?<=[{al}{e}{p}(?:{q})])\.".format(al=ALPHA_LOWER, e=r"%²\-\+", q=CONCAT_QUOTES, p=PUNCT),
+            r"(?<=[{au}][{au}])\.".format(au=ALPHA_UPPER),
+            r"(?<=[{a}][{au}])\.".format(a=ALPHA, au=ALPHA_UPPER),
```
richardpaulhudson, line 21: What is the reasoning behind the `(a=ALPHA, au=ALPHA_UPPER)` suffix rule?

It makes the rule preceding it redundant; and the rules we have ended up with mean that:

- a period is split off following any lowercase letter
- a period is split off following an uppercase letter preceded by any other letter

If this is the desired behaviour, it might be worth considering changing `al` to `a` in the fourth rule from the top, as this is simpler than having three separate rules and there are very few situations it would catch that the three rules as they stand would not catch.

spacy/lang/sl/punctuation.py
```diff
@@ -0,0 +1,64 @@
...
+                    "(jevcimi)|(ovka)|(ovke)|(ovki)|(ovko)|(ovk)|(ovkama)|(ovkah)|(ovkam)|(ovkami)|" \
+                    "(evka)|(evke)|(evki)|(evko)|(evk)|(evkama)|(evkah)|(evkam)|(evkami)|(jevka)|" \
+                    "(jevke)|(jevki)|(jevko)|(jevk)|(jevkama)|(jevkah)|(jevkam)|(jevkami)|(timi)|" \
+                    "(im)|(ima)|(a)|(imi)|(e)|(o)|(ega)|(ti)|(em)|(tih)|(emu)|(tim)|(i)|(tima)|" \
+                    "(ih)|(ta)|(te)|(to)|(tega)|(tem)|(temu))"
+
+_infixes = (
+        LIST_ELLIPSES
+        + LIST_ICONS
+        + [
```
richardpaulhudson, line 51: It would be helpful to have an explanation and preferably also a test for the first two of these rules. In the case of the first one, it might also be worth considering replacing the list of punctuation with an existing character class?

spacy/lang/sl/tokenizer_exceptions.py
```diff
@@ -0,0 +1,260 @@
+from typing import Dict, List
+from ..tokenizer_exceptions import BASE_EXCEPTIONS
+from ...symbols import ORTH, NORM
+from ...util import update_exc
+
+_exc: Dict[str, List[Dict]] = {}
+
+_other_exc = {
+    "...": [{ORTH: "."}, {ORTH: "."}, {ORTH: "."}],
```
richardpaulhudson, line 9: Is there a particular reason why we would want to split `...` into three tokens for Slovenian? We normally regard it as a single token with its own semantics: the fact it contains three periods could be seen as an implementation detail.

spacy/lang/sl/tokenizer_exceptions.py
```diff
@@ -0,0 +1,260 @@
...
+    {ORTH: "glag.", NORM: "glagolski"},
+    {ORTH: "glasb.", NORM: "glasba"},
+    {ORTH: "gled.", NORM: "gledališče"},
+    {ORTH: "gost.", NORM: "gostinstvo"},
+    {ORTH: "gozd.", NORM: "gozdarstvo"},
+    {ORTH: "gr.", NORM: "grški, grško"},
+    {ORTH: "grad.", NORM: "gradbeništvo"},
+    {ORTH: "hebr.", NORM: "hebrejsko"},
+    {ORTH: "hrv.", NORM: "hrvaško"},
+    {ORTH: "ide.", NORM: "praindoevropsko"},
```
richardpaulhudson, line 88: Can the abbreviation not refer to modern Indo-European languages or peoples rather than necessarily to Proto-Indo-European?

spacy/lang/sl/tokenizer_exceptions.py
```diff
@@ -0,0 +1,260 @@
...
+    "le-ta": [{ORTH: "le"}, {ORTH: "-"}, {ORTH: "ta"}],
+    "Le-ta": [{ORTH: "Le"}, {ORTH: "-"}, {ORTH: "ta"}],
+    "le-tega": [{ORTH: "le"}, {ORTH: "-"}, {ORTH: "tega"}],
+    "Le-tega": [{ORTH: "Le"}, {ORTH: "-"}, {ORTH: "tega"}]
+}
+
+_exc.update(_other_exc)
+
+
+for exc_data in [
```
richardpaulhudson, line 37: In general the `NORM` value has to be a single form, It's fine if that form consists of a phrase, like `to je`, but it should be possible to replace the abbreviation with the `NORM` form in any text and for the text to still make sense. Some of these `NORM` forms are more like dictionary entries, either because they have multiple meanings (`agronomija, agrotehnika`) or grammatical forms (`angleški, angleško`) or because they contain words that are not directly part of the abbreviation (`prehodna raba (glagola)`). 

In general if it's not possible to assign a single `NORM` form to an abbreviation, it's probably best to leave it without one.

spacy/lang/sl/tokenizer_exceptions.py
```diff
@@ -0,0 +1,260 @@
...
+    {ORTH: "okt.", NORM: "oktober"},
+    {ORTH: "or.", NORM: "orodnik"},
+    {ORTH: "os.", NORM: "oseba"},
+    {ORTH: "otr.", NORM: "otroško"},
+    {ORTH: "oz.", NORM: "oziroma"},
+    {ORTH: "pal.", NORM: "paleontologija"},
+    {ORTH: "papir.", NORM: "papirništvo"},
+    {ORTH: "ped.", NORM: "pedagogika"},
+    {ORTH: "pisar.", NORM: "pisarniško"},
+    {ORTH: "podr.", NORM: "vez"},
```
richardpaulhudson, line 145: I don't know enough Slovenian to understand this entry, but if it's really correct it would be helpful to have a comment to explain it.

richardpaulhudson (commented):
Thank you very much indeed for this PR @orglce, it will be great to have better support for Slovenian!

I'll add some specific comments in a minute, but one general request is whether you could possibly add some tests? They would be especially helpful for features that you have developed from scratch rather than copying them from other languages. Examples of such features are the HYPHENS_PERMITTED list or the splitting of initials in names like J.K. Rowling. As well as proving that the features work as expected, tests will also help people who might come in the future to understand them. A good starting point for tests is probably the langauge-specific test directory for a well-resourced language such as English (https://github.com/explosion/spaCy/tree/master/spacy/tests/lang/en): you can copy individual tests and modify them as required.

spacy/lang/sl/tokenizer_exceptions.py
orglce: It was a bug in my automation script. It was originally 'podr. vez. - podredna vez'. Now I just put it under abbreviations without the `NORM` under podr.

spacy/lang/sl/tokenizer_exceptions.py
orglce: I changed everything to a single form. In the future I will also try to add more `NORM` forms.

spacy/lang/sl/tokenizer_exceptions.py
orglce: Yes, you are right. It makes much more sense.

spacy/lang/sl/tokenizer_exceptions.py
orglce: There is not, I just followed our gold standard dataset that seems to split it. But if Spacy in general does not split it, I think it also makes sense to be that way for Slovenian. I removed the exception.

spacy/lang/sl/punctuation.py
orglce: It was just the fact that I was following the gold standard dataset that seems to split much more than the Spacys default tokenizer. Things like `naslovom»Pogovor` or `odlomek)Ciljna` would be split by » and ). But after second thought I don't think the Slovenian tokenizer should work any differently than the default tokenizer on non language specific strings. What do you think? Should we follow the dataset (and its tokenizer [Obeliks](https://github.com/clarinsi/obeliks)) or leave as much of the default tokenizer as possible?

I have cleaned the rules and commented the out for now. The only rule that would be different from the default ones then would be this one

`r"(?<=[{a}0-9])(?:{h})(?!{hp}$)".format(a=ALPHA, h=HYPHENS, hp=HYPHENS_PERMITTED),`

Is there a way to only add this one to the default infixes instead of copy pasting everything and just modifying one rule?

spacy/lang/sl/lex_attrs.py
orglce: I've removed the rule to split numbers on ``

spacy/lang/sl/punctuation.py
orglce: The lists have been cleaned and duplicates removed.

spacy/lang/sl/lex_attrs.py
orglce: After looking at the inflicted forms, basically few different forms cover all the inflicted forms of the masculine, feminine and neutral cases so I have just included all of them. I have also updated the inflicted forms of the currencies.

spacy/lang/sl/punctuation.py
orglce: I have noticed a few words in the dataset ended with a capital for some reason and the default tokenizer doesn't split it. I have just included this rule but I don't know if it reallly makes sense.

e.g. the default tokenizer doesn't split `BiH.` but the dataset splits the dot.

spacy/lang/sl/lex_attrs.py
orglce: I have tried to note down all the inflicted forms and it turns out that a lot of them overlap. So everything combined isn't all that much. That should pretty much be everything. I should have looked at it sooner. :)

spacy/lang/sl/punctuation.py
adrianeboyd: If you think it's kind of a fluke in the data that those tokens are split, it's fine to leave this out from the default spacy settings. It's useful to be able to say that the default tokenizer settings for a particular language are following the guidelines from an existing corpus, but it doesn't have to be customized for every single case in the corpus if you don't think this makes sense for general spacy users.

It's not hard to add a new pattern to the default infixes, but editing or removing an existing pattern is harder, so we usually copy/paste in those cases. You can extend the infixes like this (the order can matter if there are overlapping patterns):

https://github.com/explosion/spaCy/blob/93960dc4b59510b011c12079fbba09eb8219f74e/spacy/lang/ko/punctuation.py#L1-L12

Is it possible that adding the custom rule first will cover the cases you're trying to address? (I don't actually think so, but at this point I'm not sure I understand enough to the details to be sure?)

spacy/lang/sl/lex_attrs.py
richardpaulhudson: The rule still appears to be in the current code?

orglce: The rule in the punctuation.py or somewhere else? I think I removed it. The infixes are now basically the same as default with the exception of the HYPHENS-PERMITTED rule.

spacy/lang/sl/punctuation.py
orglce: I've tried to add the rule before the `BASE_TOKENIZER_INFIXES` and it doesn't seem to follow it, because there is another hyphen rule after it. The rule I am trying to add just says don't split any words that end with a hyphen and the any of the suffixes from the list HYPHENS_PERMITTED. 

e.g. BTC-ju, UKC-jev, tam-ovci, SFOR-ju ...

These shouldn't be split because they contain -ju, -jev, -ovci, -ju that is in the list. Everything else containing a hyphen should be split as normal.

spacy/lang/sl/lex_attrs.py
richardpaulhudson: Sorry, my mistake, I was looking at the code in `lex_attrs.py` (which obviously doesn't split anything)!

spacy/lang/sl/punctuation.py
richardpaulhudson: Yes, I think in this case we have to copy/paste and alter the standard group (as you have done), because you're not defining new infixes but restricting the scope of one of the current infixes. However, I would change the `HYPHENS_PERMITTED` index to look like this:

`r"(?<=[{a}0-9])(?:{h})(?!{hp}$)(?=[{a}])".format(a=ALPHA, h=HYPHENS, hp=HYPHENS_PERMITTED)`

because the existing infix will match e.g. `BTC-3`.

spacy/lang/sl/punctuation.py
richardpaulhudson: I would make the decision based on whether or not you think this a feature of Slovenian that is different from most other languages — then I'd add the rule. If however you think it's just a few specific words like `BiH`, I would add these as tokenizer exceptions.

If we do want this rule, it would be worth considering changing the existing rule to 

`r"(?<=[{a}{e}{p}(?:{q})])\.".format(a=ALPHA, e=r"%²\-\+", q=CONCAT_QUOTES, p=PUNCT),` 
and getting rid of the two rules that follow it as the changed rule would catch everything they catch as well without catching very much else. 

On the other hand, if you extend the suffix patterns as Adriane suggested in the conversation below, it would be fine to just add the new rule and leave the old ones as they are.

spacy/lang/sl/punctuation.py
orglce: Yes, thanks. I've missed that one.

spacy/lang/sl/punctuation.py
orglce: It is not a feature of Slovenian rather a typo, so I have remove the rule. However I have left the rule that splits the initials because there is quite a few examples in the dataset that do this. I am sorry about the confusing number of commits, I am going back and forth on some of these rules, but I think now it should be OK. However I cannot simply extend the rules because I have modified this one to not include numbers, because this is a feature of Slovenian.

`r"(?<=[{al}{e}{p}(?:{q})])\.".format(al=ALPHA_LOWER, e=r"%²\-\+", q=CONCAT_QUOTES, p=PUNCT)`

The original rule looks like this:

`r"(?<=[0-9{al}{e}{p}(?:{q})])\.".format(al=ALPHA_LOWER, e=r"%²\-\+", q=CONCAT_QUOTES, p=PUNCT)`

adrianeboyd:
Thanks for the updates, I think this looks ready to go when we have a branch ready for v3.5.0.

In a minute I'll push a couple commits that re-enable a test and format.

PR #11175, opened 2022-07-21
Add a way to get the URL to download a pipeline to the CLI

polm:
## Description

This adds a `--url` flag to the `spacy info [model]` command that prints the URL of the most recent compatible version of the named pipeline. This URL is printed without any formatting for use in scripts, like:

```
url=$(spacy info en_core_web_sm --url)
wget "$url"
```

Because this requires a compatibility check, it requires an Internet connection.

This also adds the download URL for installed packages to the output of `spacy info [model]` if available. The information is typically available for pipelines installed with pip, and is accessed through `pkg_resources`. This does not require an Internet connection.

-----
Original description: 

This is a simple implementation of a `--dry-run` flag for `spacy download` that makes it print the URL of the package to be installed without downloading anything. 

I don't have strong opinions about the way this is done, but there are some questions:

- would it be better to print just the URL?
- should there be a method to get the URL for outside use?
- should we be concerned about the flags overlapping with pip (they don't currently)?

I'm not sure how universal this is, but I used `-n` for the short flag, which I understand to come from `noop`. 

### Types of change

minor enhancement

## Checklist

### DIFF
--- a/spacy/cli/download.py
+++ b/spacy/cli/download.py
@@ -19,7 +19,8 @@ def download_cli(
     ctx: typer.Context,
     model: str = Arg(..., help="Name of pipeline package to download"),
     direct: bool = Opt(False, "--direct", "-d", "-D", help="Force direct download of name + version"),
-    sdist: bool = Opt(False, "--sdist", "-S", help="Download sdist (.tar.gz) archive instead of pre-built binary wheel")
+    sdist: bool = Opt(False, "--sdist", "-S", help="Download sdist (.tar.gz) archive instead of pre-built binary wheel"),
+    dry_run: bool = Opt(False, "--dry-run", "-n", help="Print the URL of the file to download without downloading it")
     # fmt: on
 ):
     """
@@ -29,13 +30,22 @@ def download_cli(
     additional arguments provided to this command will be passed to `pip install`
     on package installation.
 
+    If the --dry-run flag is set, the URL of the pipeline to download will be
+    printed, but nothing will actually be downloaded.
+
     DOCS: https://spacy.io/api/cli#download
     AVAILABLE PACKAGES: https://spacy.io/models
     """
-    download(model, direct, sdist, *ctx.args)
+    download(model, direct, sdist, dry_run, *ctx.args)
 
 
-def download(model: str, direct: bool = False, sdist: bool = False, *pip_args) -> None:
+def download(
+    model: str,
+    direct: bool = False,
+    sdist: bool = False,
+    dry_run: bool = False,
+    *pip_args,
+) -> None:
     if (
         not (is_package("spacy") or is_package("spacy-nightly"))
         and "--no-deps" not in pip_args
@@ -55,7 +65,6 @@ def download(model: str, direct: bool = False, sdist: bool = False, *pip_args) -
         components = model.split("-")
         model_name = "".join(components[:-1])
         version = components[-1]
-        download_model(dl_tpl.format(m=model_name, v=version, s=suffix), pip_args)
     else:
         model_name = model
         if model in OLD_MODEL_SHORTCUTS:
@@ -66,7 +75,14 @@ def download(model: str, direct: bool = False, sdist: bool = False, *pip_args) -
             model_name = OLD_MODEL_SHORTCUTS[model]
         compatibility = get_compatibility()
         version = get_version(model_name, compatibility)
-        download_model(dl_tpl.format(m=model_name, v=version, s=suffix), pip_args)
+
+    filename = dl_tpl.format(m=model_name, v=version, s=suffix)
+    download_url = about.__download_url__ + "/" + filename
+    if dry_run:
+        msg.good(f"The download URL for {model} is:", download_url)
+        return
+
+    download_model(filename, pip_args)
     msg.good(
         "Download and installation successful",
         f"You can now load the package via spacy.load('{model_name}')",
--- a/spacy/cli/download.py
+++ b/spacy/cli/download.py
@@ -79,9 +79,12 @@ def download(
 
 
 def get_model_filename(model_name: str, version: str, sdist: bool = False) -> str:
-    dl_tpl = "{m}-{v}/{m}-{v}{s}#egg={m}=={v}"
+    dl_tpl = "{m}-{v}/{m}-{v}{s}"
+    egg_tpl = "#egg={m}=={v}"
     suffix = SDIST_SUFFIX if sdist else WHEEL_SUFFIX
     filename = dl_tpl.format(m=model_name, v=version, s=suffix)
+    if sdist:
+        filename += egg_tpl.format(m=model_name, v=version)
     return filename
 
 
--- a/spacy/cli/download.py
+++ b/spacy/cli/download.py
@@ -20,7 +20,6 @@ def download_cli(
     model: str = Arg(..., help="Name of pipeline package to download"),
     direct: bool = Opt(False, "--direct", "-d", "-D", help="Force direct download of name + version"),
     sdist: bool = Opt(False, "--sdist", "-S", help="Download sdist (.tar.gz) archive instead of pre-built binary wheel"),
-    dry_run: bool = Opt(False, "--dry-run", "-n", help="Print the URL of the file to download without downloading it")
     # fmt: on
 ):
     """
@@ -30,20 +29,16 @@ def download_cli(
     additional arguments provided to this command will be passed to `pip install`
     on package installation.
 
-    If the --dry-run flag is set, the URL of the pipeline to download will be
-    printed, but nothing will actually be downloaded.
-
     DOCS: https://spacy.io/api/cli#download
     AVAILABLE PACKAGES: https://spacy.io/models
     """
-    download(model, direct, sdist, dry_run, *ctx.args)
+    download(model, direct, sdist, *ctx.args)
 
 
 def download(
     model: str,
     direct: bool = False,
     sdist: bool = False,
-    dry_run: bool = False,
     *pip_args,
 ) -> None:
     if (
@@ -59,8 +54,6 @@ def download(
             "dependencies, you'll have to install them manually."
         )
         pip_args = pip_args + ("--no-deps",)
-    suffix = SDIST_SUFFIX if sdist else WHEEL_SUFFIX
-    dl_tpl = "{m}-{v}/{m}-{v}{s}#egg={m}=={v}"
     if direct:
         components = model.split("-")
         model_name = "".join(components[:-1])
@@ -76,17 +69,20 @@ def download(
         compatibility = get_compatibility()
         version = get_version(model_name, compatibility)
 
-    filename = dl_tpl.format(m=model_name, v=version, s=suffix)
-    download_url = about.__download_url__ + "/" + filename
-    if dry_run:
-        msg.good(f"The download URL for {model} is:", download_url)
-        return
+    filename = get_model_filename(model_name, version, sdist)
 
     download_model(filename, pip_args)
     msg.good(
         "Download and installation successful",
         f"You can now load the package via spacy.load('{model_name}')",
     )
+
+
+def get_model_filename(model_name: str, version: str, sdist: bool = False) -> str:
+    dl_tpl = "{m}-{v}/{m}-{v}{s}#egg={m}=={v}"
+    suffix = SDIST_SUFFIX if sdist else WHEEL_SUFFIX
+    filename = dl_tpl.format(m=model_name, v=version, s=suffix)
+    return filename
 
 
 def get_compatibility() -> dict:
@@ -115,6 +111,11 @@ def get_version(model: str, comp: dict) -> str:
             exits=1,
         )
     return comp[model][0]
+
+
+def get_latest_version(model: str) -> str:
+    comp = get_compatibility()
+    return get_version(model, comp)
 
 
 def download_model(
--- a/spacy/cli/info.py
+++ b/spacy/cli/info.py
@@ -18,6 +18,7 @@ def info_cli(
     silent: bool = Opt(False, "--silent", "-s", "-S", help="Don't print anything (just return)"),
     exclude: str = Opt("labels", "--exclude", "-e", help="Comma-separated keys to exclude from the print-out"),
     url: bool = Opt(False, "--url", "-u", help="Print the URL to download the pipeline from"),
+    requirements: bool = Opt(False, "--requirements", "-r", help="Print requirements info for a pipeline"),
     # fmt: on
 ):
     """
@@ -28,7 +29,14 @@ def info_cli(
     DOCS: https://spacy.io/api/cli#info
     """
     exclude = string_to_list(exclude)
-    info(model, markdown=markdown, silent=silent, exclude=exclude, url=url)
+    info(
+        model,
+        markdown=markdown,
+        silent=silent,
+        exclude=exclude,
+        url=url,
+        requirements=requirements,
+    )
 
 
 def info(
@@ -38,16 +46,24 @@ def info(
     silent: bool = True,
     exclude: Optional[List[str]] = None,
     url: bool = False,
+    requirements: bool = False,
 ) -> Union[str, dict]:
     msg = Printer(no_print=silent, pretty=not silent)
     if not exclude:
         exclude = []
     if url:
+        # TODO should we check that requirements and url are not used together?
         if model is not None:
             title = f"Download info for pipeline '{model}'"
             data = info_model_url(model)
             print(data["download_url"])
             return data
+        else:
+            msg.fail("--url option requires a pipeline name", exits=1)
+    elif requirements:
+        if model is not None:
+            title = f"Requirements info for pipeline '{model}'"
+            data = info_model_requirements(model)
         else:
             msg.fail("--url option requires a pipeline name", exits=1)
     elif model:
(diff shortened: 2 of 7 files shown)


adrianeboyd:
I think there's too much danger for this to conflict with the pip flags. I noticed that this was just merged, maybe we can use this somehow instead? https://github.com/pypa/pip/pull/10771

Or alternatively this should be in a separate command (not `download`) and without any interaction with pip flags.

adrianeboyd:
There's no way for pip to resolve things without downloading so `pip install --dry-run` does the download but then doesn't install in the end, which isn't great here.

polm:
OK, we definitely shouldn't use this if pip is adding a flag with the same name. Since pip has a lot of options it seems like the only safe option is adding another command. I'm not sure what it should be called, maybe `download-url`?

adrianeboyd:
I think a separate command sounds easier, but the naming is hard. `download_url` still sounds too much to me like it might be downloading something.

In general I think people would like to be able to do something like:

```bash
spacy whatever en_core_web_sm >> requirements.txt
```

polm:
What if we had commands like `spacy pipelines url` and `spacy pipelines list` or something? `list` giving the full list and `url` giving the url of the listed (maybe one or more?) pipelines?

adrianeboyd:
Let's see, I'd never noticed this but we already have `spacy info model`, which is only intended for installed models, but maybe we can add a bit of info about how to specify requirements, also for our own trained pipelines that aren't installed. It would require some level of failing gracefully without internet access.

svlandeg:
I like this idea - could it be 
```
spacy info en_core_web_sm --url
```
that would basically work regardless of whether the model is already installed or not?

adrianeboyd:
I was imagining something that could show a bit more about how to specify it as a dependency, so at least some option with more details like:

```
$ spacy info en_core_web_sm --requirements

setup.cfg:

install_requires = 
    en_core_web_sm @ URL

poetry:

   something with {url = "URL"} (I'd have to look it up)
```

Also a plain `--url` seems useful. Or maybe this should just point to the docs with more details, but something that makes this easier in general.

polm:
OK, I have updated this to support this syntax:

```
spacy info en_core_web_sm --url
```

Initially I had this work like other info commands and print a table by default, but consdering the comment about piping the output upthread, and that there's only one relevant field, I instead modified to print just the URL of the most recent compatible version of the pipeline when successful. 

As far as failure modes:

- if not online, the compatibility check will fail gracefully with a clear message (this is just in the compatibility function)
- if the pipeline name doesn't exist, an error mentioning that will be raised (same as downloading a name that doesn't exist)
- if you didn't pass a pipeline name there's an error for that

There are some unresolved questions. There are lots of things we could add but I think it's better to keep this a simple, minimal feature that can be integrated in other things, so I've kept things simple when in doubt about the below. (In particular, because this is overloading the existing `info` command, the flag interactions would get weird. If we want more sophisticated options I think it'd be better to have a new command or mode.)

- this always provides the wheel URL rather than the `.tar.gz` URL. I don't think that's a problem but I'm not sure when one would be preferable over the other.
- I didn't add a feature to directly specify the version rather than grabbing the latest compatible one.
- we could also provide more info about models that aren't installed, but that feels like a different feature.

For the `--requirements` option, it wouldn't be difficult to add that, I'll look at doing it tomorrow.

adrianeboyd:
I think providing just the wheel URL is fine and just providing the same URL as you would get from `spacy download` is also fine.

The no-internet error is pretty ugly here, especially since it's not that clear to users that it's going to need internet access to provide the URL.

adrianeboyd:
Actually if the model is installed locally, it make more sense to provide that version, or possibly to provide both?

polm:
> The no-internet error is pretty ugly here, especially since it's not that clear to users that it's going to need internet access to provide the URL.

I don't think it's that bad? It clearly states what the failure was (couldn't look up the compatibility table). It would be nice if we could check compatibility offline, but since we can't... The ideal solution would probably be a URL that could be generated locally that would redirect to the latest compatible version. (There is also the issue that looking up the compatibility table offline doesn't throw an exception, it exits the program, so it would need to be modified to handle it elegantly.)

> Actually if the model is installed locally, it might make more sense to provide that version, or possibly to provide both?

I'm not sure I understand the use case? These are the uses I was imagining:

1. Download a model in CI for some reason. You probably want the most recent compatible version, and if you don't you can cache the output somehow.
2. You need the URL to download the model in a particular way due to a corporate firewall. So you just need the default download URL. 

I guess for locally installed models, you might want the URL if you want to declare dependencies for a project you've been working on - is that what you had in mind? For that use case, would it be better to add the download URL to the `spacy info model_name` (without the --url option)?

spacy/cli/download.py
```diff
@@ -66,13 +68,23 @@ def download(model: str, direct: bool = False, sdist: bool = False, *pip_args) -
...
+
+    download_model(filename, pip_args)
     msg.good(
         "Download and installation successful",
         f"You can now load the package via spacy.load('{model_name}')",
     )
 
 
+def get_model_filename(model_name: str, version: str, sdist: bool = False) -> str:
+    dl_tpl = "{m}-{v}/{m}-{v}{s}#egg={m}=={v}"
```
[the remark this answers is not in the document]
polm: Added that behaviour.

spacy/cli/info.py
```diff
@@ -16,17 +17,18 @@ def info_cli(
     markdown: bool = Opt(False, "--markdown", "-md", help="Generate Markdown for GitHub issues"),
     silent: bool = Opt(False, "--silent", "-s", "-S", help="Don't print anything (just return)"),
     exclude: str = Opt("labels", "--exclude", "-e", help="Comma-separated keys to exclude from the print-out"),
+    url: bool = Opt(False, "--url", "-u", help="Print the pipeline download URL"),
```
svlandeg, line 20: This is likely a matter of personal preference, so feel free to ignore.
```suggestion
    url: bool = Opt(False, "--url", "-u", help="Print the URL to download the pipeline from"),
```

spacy/cli/info.py
```diff
@@ -35,11 +37,20 @@ def info(
...
 ) -> Union[str, dict]:
     msg = Printer(no_print=silent, pretty=not silent)
     if not exclude:
         exclude = []
-    if model:
+    if url:
+        if model is not None:
+            title = f"Download info for pipeline '{model}'"
+            data = info_model_url(model)
+            print(data["download_url"])
```
svlandeg, line 49: Shouldn't we print through `msg` here (and at L67)?

polm: I was originally using `msg`, but I changed it because I expect a use like this:

```bash
download_url=$(spacy info en_core_news_sm)
wget $download_url
```

So any extra formatting would be undesirable. 

I think L67 is that way for similar reasons, but I'm not sure.

polm:
I just pushed a change to add output for `spacy info [pipeline] --requirements`. The code is a bit messy because there are two output formats here. 

This is the one I prefer:

```
============== Requirements info for pipeline 'en_core_web_sm' ==============



    Release info page:

    https://github.com/explosion/spacy-models/releases/tag/en_core_web_sm-3.4.0

    setup.cfg:

    install_requires =
        en_core_web_sm @ https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.4.0/en_core_web_sm-3.4.0-py3-none-any.whl

    poetry:

    en_core_web_sm = { url = "https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.4.0/en_core_web_sm-3.4.0-py3-none-any.whl" }
```

This is a little hacky though - it's using the same `msg` output as other pre-existing commands, but with an empty key and one value in order to preserve formatting. (This is probably why the markdown option uses `print`.)

This is output with one key per entry. This is the same way other commands work but I think it's too hard to read, mainly because of the newline for `setup.cfg` format. 

```
============== Requirements info for pipeline 'en_core_web_sm' ==============

Release info page   https://github.com/explosion/spacy-models/releases/tag/en_core_web_sm-3.4.0
setup.cfg           install_requires =
        en_core_web_sm @ https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.4.0/en_core_web_sm-3.4.0-py3-none-any.whl
poetry              en_core_web_sm = { url = "https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.4.0/en_core_web_sm-3.4.0-py3-none-any.whl" }
```

I also feel like at this point `spacy info` is getting a little overloaded, and it would be better to have subcommands (like `debug data`). But since we already support model names I guess we'd need to have a new keyword if we wanted to do that.

For this PR I think we should leave `--requirements` for later and figure out the formatting for the output. My preference is to provide just a URL for ease when scripting, but I think it would also make sense to provide both the download URL and the release info URL ([example](https://github.com/explosion/spacy-models/releases/tag/zh_core_web_trf-3.4.0)) in the typical table format used by other existing options.

polm:
Just to clarify the state of this a bit, what I think we should do now is:

1. leave `--requirements` for a separate PR
2. output just the URL, without formatting, for use in scripting

About generating different URLs based on installed packages, I'm still not clear what the use case for that is for downloading (it makes sense for `--requirements`). 

I have left the `--requirements` related suggestions in the PR to make it easier to check them, but would delete them before taking them out of draft if we do want to do that separately.

spacy/cli/info.py
```diff
@@ -16,17 +19,24 @@ def info_cli(
     markdown: bool = Opt(False, "--markdown", "-md", help="Generate Markdown for GitHub issues"),
     silent: bool = Opt(False, "--silent", "-s", "-S", help="Don't print anything (just return)"),
     exclude: str = Opt("labels", "--exclude", "-e", help="Comma-separated keys to exclude from the print-out"),
+    url: bool = Opt(False, "--url", "-u", help="Print the URL to download the most recent version of the pipeline"),
```
adrianeboyd, line 22: everywhere: "recent compatible version"

polm:
I think this is OK to merge now?

adrianeboyd:
I'll take another look. Can you update the PR description?

polm:
Updated the PR description to reflect the current behavior.

spacy/cli/info.py
```diff
@@ -16,17 +19,27 @@ def info_cli(
...
     exclude: str = Opt("labels", "--exclude", "-e", help="Comma-separated keys to exclude from the print-out"),
+    url: bool = Opt(False, "--url", "-u", help="Print the URL to download the most recent compatible version of the pipeline"),
     # fmt: on
 ):
     """
     Print info about spaCy installation. If a pipeline is specified as an argument,
     print its meta information. Flag --markdown prints details in Markdown for easy
     copy-pasting to GitHub issues.
 
+    Flag --url prints only the download URL of the most recent comptaible
```
adrianeboyd, line 30:
```suggestion
    Flag --url prints only the download URL of the most recent compatible
```

website/docs/api/cli.md
```diff
@@ -77,14 +77,15 @@ $ python -m spacy info [--markdown] [--silent] [--exclude]
...
+| Name                                             | Description                                                                                                  |
+| ------------------------------------------------ | ------------------------------------------------------------------------------------------------------------ |
+| `model`                                          | A trained pipeline, i.e. package name or path (optional). ~~Optional[str] \(option)~~                        |
+| `--markdown`, `-md`                              | Print information as Markdown. ~~bool (flag)~~                                                               |
+| `--silent`, `-s` <Tag variant="new">2.0.12</Tag> | Don't print anything, just return the values. ~~bool (flag)~~                                                |
+| `--exclude`, `-e`                                | Comma-separated keys to exclude from the print-out. Defaults to `"labels"`. ~~Optional[str]~~                |
+| `--url`, `-u` <Tag variant="new">3.5.0</Tag>     | Print the URL to download the most recent compatible version of the pipeline. Requires a pipeline name. ~~bool (flag)~~ |
```
svlandeg, line 86: This needs another round of `prettier` ;-)

polm: Thanks, should be OK now.

adrianeboyd:
I think the functionality is looking good. Do we want to include the changes to the docs in this PR or a follow-up PR?

There are at least these sections:

https://spacy.io/usage/models#download-pip
https://spacy.io/usage/models#production

And maybe easier instructions on the models pages themselves? (I think having to dig through releases for the asset URL is the too-difficult part.)

svlandeg:
I'd prefer having the docs changing with the PR!

polm:
OK, I updated the docs. In addition to adding mention of the new command to the places linked above, I put download links directly in the tables on the model pages. How's that?

polm:
Here's what it looks like on the model page:

2022-08-30T18-26-35

website/docs/usage/models.md
```diff
@@ -365,15 +365,30 @@ pipeline package can be found.
...
+> [Github Releases](https://github.com/explosion/spacy-models/releases), and you
+> can find download links there, as well as on the model page. You can also get download URLs directly
+> from the command line by using `spacy info` with the `--url` flag, which may be useful for automation.
+>
+> ```bash
+> spacy info en_core_web_sm --url
+> ```
+>
+> This command will print the download URL for the latest version of a pipeline
+> compatible with the version of spaCy you're using. Note that in order to look up the compatability information an Internet connection is required.
```
adrianeboyd, line 382: Do we capitalize internet?

website/docs/usage/models.md
```diff
@@ -365,15 +365,30 @@ pipeline package can be found.
 To download a trained pipeline directly using
 [pip](https://pypi.python.org/pypi/pip), point `pip install` to the URL or local
 path of the wheel file or archive. Installing the wheel is usually more
-efficient. To find the direct link to a package, head over to the
-[releases](https://github.com/explosion/spacy-models/releases), right click on
-the archive link and copy it to your clipboard.
+efficient.
+
+> #### How to Find Download URLs {#download-urls}
```
adrianeboyd, line 370: ?

```suggestion
> #### Pipeline Package URLs {#pipeline-urls}
```

website/docs/usage/models.md
```diff
@@ -365,15 +365,30 @@ pipeline package can be found.
...
+>
+> Pretrained pipeline distributions are hosted on
+> [Github Releases](https://github.com/explosion/spacy-models/releases), and you
+> can find download links there, as well as on the model page. You can also get download URLs directly
```
adrianeboyd, line 374:
```suggestion
> can find download links there, as well as on the model page. You can also get URLs directly
```

website/docs/usage/models.md
```diff
@@ -365,15 +365,30 @@ pipeline package can be found.
...
+> from the command line by using `spacy info` with the `--url` flag, which may be useful for automation.
+>
+> ```bash
+> spacy info en_core_web_sm --url
+> ```
+>
+> This command will print the download URL for the latest version of a pipeline
```
adrianeboyd, line 381:
```suggestion
> This command will print the URL for the latest version of a pipeline
```

website/docs/usage/models.md
```diff
@@ -515,8 +530,8 @@ Because pipeline packages are valid Python packages, you can add them to your
 application's `requirements.txt`. If you're running your own internal PyPi
 installation, you can upload the pipeline packages there. pip's
 [requirements file format](https://pip.pypa.io/en/latest/reference/pip_install/#requirements-file-format)
-supports both package names to download via a PyPi server, as well as direct
-URLs.
+supports both package names to download via a PyPi server, as well as [direct
+URLs](#download-urls).
```
adrianeboyd, line 534: (update if changed above)

website/src/templates/models.js
```diff
@@ -240,6 +249,7 @@ const Model = ({
         { label: 'Components', content: components, help: MODEL_META.components },
         { label: 'Pipeline', content: pipeline, help: MODEL_META.pipeline },
         { label: 'Vectors', content: meta.vectors, help: MODEL_META.vecs },
+        { label: 'Download URL', content: meta.download_url, help: MODEL_META.download_url },
```
adrianeboyd, line 252:
```suggestion
        { label: 'Package URL', content: meta.download_url, help: MODEL_META.download_url },
```

website/src/templates/models.js
```diff
@@ -134,6 +135,13 @@ function formatAccuracy(data, lang) {
         .filter(item => item)
 }
 
+function formatDownloadUrl(lang, name, version) {
+  const fullName = `${lang}_${name}-${version}`
+  const filename = `${fullName}-py3-none-any.whl`
+  const url = `https://github.com/explosion/spacy-models/releases/download/${fullName}/${filename}`
+  return <Link to={url} hideIcon>{filename}</Link>
```
adrianeboyd, line 142: If it's called a URL I think it should look like a URL, or it should be called something else, like a "link".
```suggestion
  return <Link to={url} hideIcon>{url}</Link>
```

adrianeboyd (commented):
I think the summary of this is that I don't like "Download URL" as the name for this. It's just a URL? Possibly a "(pipeline) (package) URL"?

website/src/templates/models.js
polm: I originally had the URL but changed it to the filename because I felt it was too long. We can use the URL here but I feel it's not helpful...

website/docs/usage/models.md
polm: I usually do, but it looks like the two other mentions in the docs don't, so I can change it.

polm: Changed.

website/src/templates/models.js
polm: If you have a better name please feel free to change it. I do feel that if we called it something like "Package link" it might not be clear that clicking on it would trigger a download.

website/docs/usage/models.md
```diff
@@ -515,8 +532,8 @@ Because pipeline packages are valid Python packages, you can add them to your
 application's `requirements.txt`. If you're running your own internal PyPi
 installation, you can upload the pipeline packages there. pip's
 [requirements file format](https://pip.pypa.io/en/latest/reference/pip_install/#requirements-file-format)
```
adrianeboyd, line 534: This link should be updated: https://pip.pypa.io/en/latest/reference/requirements-file-format/

And I think I'd switch to the `package @ URL` format with the wheel URL in the example below because most people are using wheels and the format is easier to parse. (I'd make a suggestion but it's out the range.)

polm: OK, updated. I also removed the paragraph about the `#egg=` fragment since it's no longer in the example.

PR #11193, opened 2022-07-25
Adding alpha support for Maltese (`mt`)

faustusdotbe:
## Description

Following the [discussion](https://github.com/explosion/spaCy/discussions/11183) with @polm , I'm opening a PR to add alpha support for Maltese. I've compiled the spaCy locally and it goes as expected.

### Types of change

This is an enhancement, we add alpha support for Maltese.

## Checklist


polm:
Thanks for this submission! 

I notice that this includes the base language class, example sentences, and stop words, but not lexical attributes or other data. Usually when adding a language it's nice to have extra information like that - would you be able to add it? If not, or if it would take time, it's always fine for us to do it later in another PR. You can see examples of the data by looking at other languages, or check the [`new language`](https://github.com/explosion/spaCy/labels/new%20language) label for other PRs that added language support.

About your stop words, is there any particular reason you chose this list? I haven't gone over all the project documentation yet but it's not clear to me how they were generated. Also there seem to be some terms that translate to words like "environment" or "center" that wouldn't usually be considered stop words, and there may be license issues - it's not clear to me the EuroVoc license allows open-source style redistribution. So we may want to consider a different list, or generate one from Maltese Wikipedia or something.

faustusdotbe:
Thanks for your comments! To answer:

- extra info (lexical attributes, etc.)
 
I do not currently have this information. Given I do not read Maltese I did not feel confident pushing rules on a package as widely used as spaCy. I suppose at some point in the future this info could be retrieved from https://universaldependencies.org/mt/index.html 

- stop words

I took those because these are the ones I work with and they've been satisfactory, but my data is indeed in the same domain. I am not aware of other lists for Maltese. "Cross-lingual part-of-speech tagging for Maltese" by Keith Lia & Héctor Martínez Alonso (not putting the link to the PDF here because Firefox throws a security issue) mentions stop words for Maltese, but a rather rough approach: all Adp, Conj, Det, Part and Sconj in their corpus were considered stop words. 

Regarding licensing issues: you're right, I hadn't thorougly considered that. The JEX tagger containing the stop words was created by public servants financed by public money so it made sense to me it was public domain, but the license seems to state otherwise. Whether the license is actually correct/legal I do not know. 
Do you have suggestions/best practices on how to extract a new list from Wiki? Perhaps considering all words with a tf-idf that's lower than 0.2 as a stop word?

Some more context: the reason I did the PR was because I made the change locally and thought it could benefit the community. In this current project I use [spacy-udpipe](https://github.com/TakeLab/spacy-udpipe) to process texts in languages without a trained pipeline in spaCy (but for which a UDpipe model exists). Given Maltese does not have alpha support in spaCy, spacy-udpipe fails (`spacy.lang.mt does not exist` or similar issue), but adding alpha support solves that.

polm:
Thanks for the clarification about your use case and the stop words. 

> The JEX tagger containing the stop words was created by public servants financed by public money so it made sense to me it was public domain, but the license seems to state otherwise. Whether the license is actually correct/legal I do not know.

As far as I am aware, no copyright for government works is a uniquely American phenomenon, unfortunately.

> Do you have suggestions/best practices on how to extract a new list from Wiki? Perhaps considering all words with a tf-idf that's lower than 0.2 as a stop word?

Normally I would just use most common tokens, though the list would need to be checked by a language speaker.

> Some more context: the reason I did the PR was because I made the change locally and thought it could benefit the community. In this current project I use [spacy-udpipe](https://github.com/TakeLab/spacy-udpipe) to process texts in languages without a trained pipeline in spaCy (but for which a UDpipe model exists). Given Maltese does not have alpha support in spaCy, spacy-udpipe fails (spacy.lang.mt does not exist or similar issue), but adding alpha support solves that.

Thanks, that clarifies a lot. In that case you should probably engineer a way to fall back to the `xx` language - that's what [`spacy-stanza` does](https://github.com/explosion/spacy-stanza#stanza-pipeline-options). 

Since it looks like we can't use these stop words or easily create our own set, I'm going to close this PR for now. But if you (or someone else) have a fluent speaker involved we'd be happy to add Maltese later, and thanks for proposing this in the first place.

PR #11194, opened 2022-07-25
Fix compatibility with CuPy 9.x

danieldk:
## Description

After the precomputable affine table of shape `[nB, nF, nO, nP]` is computed, padding with shape `[1, nF, nO, nP]` is assigned to the first row of the precomputed affine table. However, when we are indexing the precomputed table, we get a row of shape `[nF, nO, nP]`. CuPy versions before 10.0 cannot paper over this shape difference.

This change fixes compatibility with CuPy < 10.0 by squeezing the first dimension of the padding before assignment.

### Types of change

Compatibility fix.

## Checklist


adrianeboyd:

PR #11210, opened 2022-07-26
Adjust mkdir for pretrain command

thomashacker:
## Description

This `PR` adds the `parents=True` argument to the `mkdir` command when initializing the directories. It's the same argument configuration as in the [train command](https://github.com/explosion/spaCy/blob/master/spacy/cli/train.py#L64).

### Types of change

Enhancement

## Checklist


svlandeg (approved):
Nice catch! Definitely makes sense and it's more consistent :-)

PR #11214, opened 2022-07-26
Add ConsoleLogger.v2

thomashacker:
## Description

This `PR` adds the next version of the`ConsoleLogger` that saves the training logs to a `JSONL` file. It saves every row printed in the console as a `JSON` object. The logger has an additional argument `output_file` which sets the path to which the `log.jsonl` is being saved. 

Here's an example of a `log.jsonl` when training an `NER` component. It provides the **full numbers** of every score.

```
{"epoch":0,"step":0,"losses":{"ner":101.4694003456},"scores":{"ents_f":1.8903328932,"ents_p":1.0051736881,"ents_r":15.832363213},"score":0.0189033289}
{"epoch":0,"step":200,"losses":{"ner":10539.4847255106},"scores":{"ents_f":0.2028397566,"ents_p":0.7874015748,"ents_r":0.1164144354},"score":0.0020283976}
{"epoch":0,"step":400,"losses":{"ner":1990.8501044126},"scores":{"ents_f":18.636755824,"ents_p":36.0,"ents_r":12.5727590221},"score":0.1863675582}
{"epoch":1,"step":600,"losses":{"ner":2018.4897573819},"scores":{"ents_f":36.0300250209,"ents_p":63.5294117647,"ents_r":25.1455180442},"score":0.3603002502}
```

### Types of change

New feature and change to documentation

## Checklist


thomashacker:
The documentation will be adjusted as soon as the code is approved 😄

shadeMe:
Don't we want new loggers to be added to the [spacy-loggers](https://github.com/explosion/spacy-loggers) package instead?

thomashacker:
This is a very **very** good question!

adrianeboyd:
Not necessarily, especially if it's something short and general-purpose like the console logger.

svlandeg:
Good point/question, but I agree with Adriane that this is generic enough to be in the main library. Like I wouldn't want to add 3 more loggers here, but one for the console and one for file logging makes sense to me. Anything else can go into `spacy-loggers`.

adrianeboyd:
Is there a way to implement this without duplicating so much code?

thomashacker:
Yes, I think so
I'll try to optimize this

spacy/training/loggers.py
```diff
@@ -98,3 +101,105 @@ def finalize() -> None:
...
 
     return setup_printer
+
+
+@registry.loggers("spacy.FileLogger.v1")
+def file_logger(progress_bar: bool = False, output_path: Union[str, Path] = "./"):
+    """The FileLogger.v1 is a modified ConsoleLogger.v1 that saves the training logs to a jsonl file"""
+
+    output_path = util.ensure_path(output_path)
+    output_file = output_path / "log.jsonl"
```
adrianeboyd, line 111: I would have this setting be the full path to the file instead of a directory.

thomashacker:
I've renamed the `FileLogger.v1` to `ConsoleLogger.v2` and moved the `ConsoleLogger.v1` to `spacy-legacy` in this [PR](https://github.com/explosion/spacy-legacy/pull/22). I changed the `output_file` argument so that it sets the full path to the file now instead of the directory. 

I also added the `print_progress` argument which makes printing to the console an optional feature. For now, I've only ensured that the logger writes `JSONL` files only.

Is there a good way to write unit tests for these kinds of things? I've recreated all possible cases locally, but I think it might be better if we'd have some predefined unit tests.

spacy/training/loggers.py
```diff
@@ -23,13 +26,35 @@ def setup_table(
...
 
-@registry.loggers("spacy.ConsoleLogger.v1")
-def console_logger(progress_bar: bool = False):
+@registry.loggers("spacy.ConsoleLogger.v2")
+def console_logger(
+    progress_bar: bool = False,
+    print_progress: bool = True,
+    output_file: Union[str, Path] = None,
+):
+    """The ConsoleLogger.v2 prints out training logs in the console and saves them into a jsonl file, with both options being optional.
```
adrianeboyd, line 35:
```suggestion
    """The ConsoleLogger.v2 prints out training logs in the console and/or saves them to a jsonl file.
```

spacy/training/loggers.py
```diff
@@ -23,13 +26,35 @@ def setup_table(
...
+    print_progress (bool): Whether the logger should print the logs.
+    progress_bar (bool): Whether the logger should print the progress bar.
+    output_file (Union[str, Path]): Where the logger should save the training logs to.
```
adrianeboyd, line 38:
```suggestion
    output_file (Optional[Union[str, Path]]): The file to save the training logs to.
```

spacy/training/loggers.py
```diff
@@ -23,13 +26,35 @@ def setup_table(
     return final_cols, final_widths, ["r" for _ in final_widths]
 
 
-@registry.loggers("spacy.ConsoleLogger.v1")
-def console_logger(progress_bar: bool = False):
+@registry.loggers("spacy.ConsoleLogger.v2")
+def console_logger(
+    progress_bar: bool = False,
+    print_progress: bool = True,
```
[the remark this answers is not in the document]
svlandeg: Agreed, I also like it better with `console` specifically in the option name

svlandeg:
> Is there a good way to write unit tests for these kinds of things? I've recreated all possible cases locally, but I think it might be better if we'd have some predefined unit tests.

That's going to be pretty involved either way, it's difficult to test these kind of console/output/file things in full detail. Some basic tests that the functionality runs and the different options can coexist together might be sufficient at this point, and it's good to hear that you did more thorough tests locally as well.

spacy/training/loggers.py
svlandeg: "in" the console? Shouldn't it be "on" the console?

spacy/training/loggers.py
```diff
@@ -23,13 +26,35 @@ def setup_table(
...
+        if not output_file.parents[0].exists():
+            output_file.parents[0].mkdir(parents=True)
+
     def setup_printer(
         nlp: "Language", stdout: IO = sys.stdout, stderr: IO = sys.stderr
     ) -> Tuple[Callable[[Optional[Dict[str, Any]]], None], Callable[[], None]]:
         write = lambda text: print(text, file=stdout, flush=True)
         msg = Printer(no_print=True)
+        if not print_progress:
+            write(msg.info(f"Printing to console is disabled"))
```
svlandeg, line 55: I don't think we should write anything here, if users turn console output off then we shouldn't print further warnings on the console IMO

spacy/training/loggers.py
```diff
@@ -45,8 +70,9 @@ def setup_printer(
             cols=["E", "#"] + loss_cols + score_cols + ["Score"],
             widths=[3, 6] + [8 for _ in loss_cols] + [6 for _ in score_cols] + [6],
         )
-        write(msg.row(table_header, widths=table_widths, spacing=spacing))
-        write(msg.row(["-" * width for width in table_widths], spacing=spacing))
+        if print_progress:
```
svlandeg, line 73: is there more code we can have as part of the new `if` here? Like we don't need to format the table if we'll never use it.

svlandeg (commented):
Looking good, apart from some minor review comments :-)

spacy/training/loggers.py
```diff
@@ -73,24 +102,41 @@ def log_step(info: Optional[Dict[str, Any]]) -> None:
...
+            # Write to log file per log_step
+            log_data = {
+                "epoch": info["epoch"],
+                "step": info["step"],
+                "losses": log_losses,
+                "scores": log_scores,
+                "score": float(info["score"]),
+            }
+            if output_file:
+                with (output_file).open("a") as file_:
```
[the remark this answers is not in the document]
thomashacker: I haven't figured out a good way to do that yet...

thomashacker:
I'll try to write some basic unit tests

spacy/training/loggers.py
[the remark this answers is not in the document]
thomashacker: I don't think it's possible to open a stream outside the `log_step` function for saving the log. I think, we might be stuck with this solution?

thomashacker:
The tests are currently failing because `ConsoleLogger.v1` doesn't exist in `spacy-legacy` yet

spacy/training/loggers.py
```diff
@@ -23,13 +26,39 @@ def setup_table(
     return final_cols, final_widths, ["r" for _ in final_widths]
 
 
-@registry.loggers("spacy.ConsoleLogger.v1")
-def console_logger(progress_bar: bool = False):
+@registry.loggers("spacy.ConsoleLogger.v2")
+def console_logger(
+    progress_bar: bool = False,
+    console_output: bool = True,
+    output_file: Optional[Union[str, Path]] = None,  # type: ignore
```
adrianeboyd, line 33:
```suggestion
    output_file: Optional[Union[str, Path]] = None,
```

spacy/training/loggers.py
```diff
@@ -23,13 +26,39 @@ def setup_table(
...
         write = lambda text: print(text, file=stdout, flush=True)
         msg = Printer(no_print=True)
+
+        nonlocal output_file
+
+        if _log_exist:
+            write(msg.warn(f"The log file {output_file} already exists."))
+            output_file = None
+        elif output_file:
+            write(msg.info(f"Saving logs to {output_file}"))
```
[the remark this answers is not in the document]
adrianeboyd: No `finally` needed because this is already handled in the training loop, this is what I would expect to work (aside from the types, which continue to be a hassle for optional paths):

```diff
diff --git a/spacy/training/loggers.py b/spacy/training/loggers.py
index 13999ceca..e4d81d257 100644
--- a/spacy/training/loggers.py
+++ b/spacy/training/loggers.py
@@ -52,12 +52,13 @@ def console_logger(
         msg = Printer(no_print=True)

         nonlocal output_file
-
+        output_fileh = None
         if _log_exist:
             write(msg.warn(f"The log file {output_file} already exists."))
             output_file = None
         elif output_file:
             write(msg.info(f"Saving logs to {output_file}"))
+            output_fileh = open(output_file, "w")

         # ensure that only trainable components are logged
         logged_pipes = [
@@ -124,9 +125,8 @@ def console_logger(
                 "scores": log_scores,
                 "score": float(info["score"]),
             }
-            if output_file:
-                with (output_file).open("a") as file_:  # type: ignore
-                    file_.write(srsly.json_dumps(log_data) + "\n")
+            if output_fileh:
+                output_fileh.write(srsly.json_dumps(log_data) + "\n")

             if progress is not None:
                 progress.close()
@@ -144,7 +144,8 @@ def console_logger(
                     progress.set_description(f"Epoch {info['epoch']+1}")

         def finalize() -> None:
-            pass
+            output_fileh.close()
+

         return log_step, finalize

```

spacy/training/loggers.py
```diff
@@ -73,27 +112,44 @@ def log_step(info: Optional[Dict[str, Any]]) -> None:
...
+                if progress_bar:
+                    # Set disable=None, so that it disables on non-TTY
+                    progress = tqdm.tqdm(
+                        total=eval_frequency, disable=None, leave=False, file=stderr
+                    )
+                    progress.set_description(f"Epoch {info['epoch']+1}")
 
         def finalize() -> None:
-            pass
+            output_stream.close()
```
adrianeboyd, line 152: This may be `None`.

spacy/training/loggers.py
```diff
@@ -23,13 +26,44 @@ def setup_table(
...
+        if _log_exist:
+            write(
+                msg.warn(
+                    f"Saving logs is disabled because {output_file} already exists."
+                )
+            )
+            output_file = None
+        elif output_file:
+            write(msg.info(f"Saving logs to {output_file}"))
+            output_stream = open(output_file, "w")
```
adrianeboyd, line 65: I think because of windows we want to specify the encoding here.

spacy/training/loggers.py
```diff
@@ -73,27 +112,45 @@ def log_step(info: Optional[Dict[str, Any]]) -> None:
...
+            log_data = {
+                "epoch": info["epoch"],
+                "step": info["step"],
+                "losses": log_losses,
+                "scores": log_scores,
+                "score": float(info["score"]),
+            }
+
+            if output_stream:
+                output_stream.write(srsly.json_dumps(log_data) + "\n")
```
adrianeboyd, line 134:
```suggestion
            # Write to log file per log_step
            if output_stream:
                log_data = {
                    "epoch": info["epoch"],
                    "step": info["step"],
                    "losses": log_losses,
                    "scores": log_scores,
                    "score": float(info["score"]),
                }
                output_stream.write(srsly.json_dumps(log_data) + "\n")
```

website/docs/api/legacy.md
```diff
@@ -248,6 +248,59 @@ added to an existing vectors table. See more details in
...
+  1      600          92.13      22794.91     43.80     0.44
+  1      800         183.62      21541.39     56.05     0.56
+  1     1000         352.49      25461.82     65.15     0.65
+  1     1200         422.87      23708.82     71.84     0.72
+  1     1400         601.92      24994.79     76.57     0.77
+  1     1600         662.57      22268.02     80.20     0.80
+  1     1800        1101.50      28413.77     82.56     0.83
+  1     2000        1253.43      28736.36     85.00     0.85
+  1     2200        1411.02      28237.53     87.42     0.87
+  1     2400        1605.35      28439.95     88.70     0.89
```
adrianeboyd, line 292: While we're here, can we update this to look like the typical output, which starts at epoch 0 instead of 1 (in both places)?

adrianeboyd:
Looks good!

Issue #10165, opened 2022-01-28
I cannot initialise EntityRecognizer object with no examples, like in the docs


dataqa:
## How to reproduce the behaviour

```
cfg = {"model": DEFAULT_NER_MODEL}
model = registry.resolve(cfg, validate=True)["model"]
ner = EntityRecognizer(nlp.vocab, model)
ner.initialize(lambda: [], nlp=nlp)
```

The error I get:

```
TypeError: [E930] Received invalid get_examples callback in `Parser.initialize`. Expected function that returns an iterable of Example objects but got: []
```

In the docs, there is an example with a lambda function returning an empty list: https://spacy.io/api/entityrecognizer#initialize.

## Your Environment

- **spaCy version:** 3.2.1
- **Platform:** Linux-5.10.0-1057-oem-x86_64-with-glibc2.31
- **Python version:** 3.9.0
- **Pipelines:** en_core_web_sm (3.2.0), en_core_web_trf (3.2.0)

adrianeboyd:
Thanks for the note! This looks like an error in the API docs for a number of the pipeline components. Many of them do require at least one example in order to initialize the internal models correctly. We'll get this updated!

What are you trying to accomplish with this code, though? This kind of initialization is useful for a unit test but not really for typical use in a pipeline. Usually you'd use `spacy train` with a config (strongly recommended), or you'd call `nlp.initialize` with representative training examples for the whole pipeline instead.

PR #11221, opened 2022-07-27
Update docs for pipeline initialize() methods

ljvmiranda921:
Some pipeline methods have inconsistent API documentation as shown in #10165 . As it turns out, many of them do require at least one example in order to initialize the internal models correctly.

I created a sample change in `website/docs/api/dependencyparser.md` where...

1. I updated the example script to:

```python
parser = nlp.add_pipe("parser")
parser.initialize(get_examples(), nlp=nlp)
```

2. And updated the parameter description table of `get_examples` to explicitly say that it needs "at least one example" to work.

Let me know if these changes are sufficient so I can apply it to the remaining pipelines.

3. Additional note, the Sentencizer documentation is incomplete. It doesn't show the initialize methods.

## Description

Here's a table that outlines the differences between spaCy pipeline components

| Component            | No Examples needed | At least one Example needed | Examples should be annotated | Comments                                       |
|----------------------|--------------------|-----------------------------|------------------------------|------------------------------------------------|
| attribute_ruler      | ✔️                  |                             |                              |                                                |
| parser               |                    | ✔️                           |                              |                                                |
| trainable_lemmatizer |                    | ✔️                           | ✔️                            |                                                |
| entity_linker        |                    | ✔️                           |                              | Need to pass kb                                |
| ner                  |                    | ✔️                           |                              |                                                |
| entity_ruler         | ✔️                  |                             |                              |                                                |
| lemmatizer           | ✔️                  |                             |                              |                                                |
| morphologizer        |                    | ✔️                           | ✔️                            |                                                |
| senter               |                    | ✔️                           |                              |                                                |
| sentencizer          |  ✔️                  |                             |                              | No initialize docs (probably because it doesn't have to be trained |
| spancat              |                    | ✔️                           | ✔️                            |                                                |
| span_ruler           |  ✔️                  |                             |                              |                                                |
| tagger               |                    | ✔️                           | ✔️                            |                                                |
| textcat              |                    | ✔️                           | ✔️                            | Requires at least two labels                   |
| tok2vec              |                    | ✔️                           |                              |                                                |
| transformer          |                    | ✔️                           |                              | Has different init scheme                      |

Here's my way of checking which pipeline components have inconsistent docs:

### Reproducible check script

```python
import spacy

ALL_PIPES = [
    "attribute_ruler",
    "parser",
    "trainable_lemmatizer",
    "entity_linker",
    "ner",
    "entity_ruler",
    "lemmatizer",
    "morphologizer",
    "senter",
    "sentencizer",
    "spancat",
    "span_ruler",
    "tagger",
    "textcat",
    "tok2vec",
    "transformer",  # transformers are initialized differently, but I'm including it here for completeness
]

def try_initialize(pipe):
    try:
        nlp = spacy.blank("en")
        p = nlp.add_pipe(pipe)
        p.initialize(lambda: [], nlp=nlp)
    except Exception as e:
        print(f"Fails at `{pipe}.initialize()`")
        print(f"Error: {e}")

def main():
    for pipe in ALL_PIPES:
        try_initialize(pipe)


if __name__ == "__main__":
    main()
```

<details>
<summary>Output of the script</summary>

```
Fails at `parser.initialize()`
Error: [E930] Received invalid get_examples callback in `Parser.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `trainable_lemmatizer.initialize()`
Error: [E930] Received invalid get_examples callback in `EditTreeLemmatizer.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `entity_linker.initialize()`
Error: [E930] Received invalid get_examples callback in `EntityLinker.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `ner.initialize()`
Error: [E930] Received invalid get_examples callback in `Parser.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `morphologizer.initialize()`
Error: [E930] Received invalid get_examples callback in `Morphologizer.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `senter.initialize()`
Error: [E930] Received invalid get_examples callback in `SentenceRecognizer.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `spancat.initialize()`
Error: [E143] Labels for component 'spancat' not initialized. This can be fixed by calling add_label, or by providing a representative batch of examples to the component's `initialize` method.

Fails at `tagger.initialize()`
Error: [E930] Received invalid get_examples callback in `Tagger.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `textcat.initialize()`
Error: [E930] Received invalid get_examples callback in `TextCategorizer.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `tok2vec.initialize()`
Error: [E930] Received invalid get_examples callback in `Tok2Vec.initialize`. Expected function that returns an iterable of Example objects but got: []

Fails at `transformer.initialize()`
Error: [E930] Received invalid get_examples callback in `Transformer.initialize`. Expected function that returns an iterable of Example objects but got: []
```

</details>

### Types of change

Documentation change

## Checklist


polm (approved):
Thanks for taking this on, I think this is a good issue to address! Your changes look fine to me. 

One thing I would check is whether one example is actually enough. I think it should be, but it's possible some components made assumptions about having more than one.

adrianeboyd:
It should be `get_examples` and not `get_examples()` if we do it this way.

I'd also recommend testing all the initialize methods with one example and your exact changes to the docs to make sure you don't run into any other errors, like I'm not sure whether some of the methods have errors if the expected annotation is missing.

I'd initially try all of them with a completely unannotated example reference (just `Example(nlp.make_doc(text), nlp.make_doc(text))`) and only one example to see what happens.

Technically I think one example is sufficient for initialize to run, which is all we need for the API docs, but you typically wouldn't end up with the right labels from one example (except sometimes for textcat) and maybe this should be clarified better somewhere.

One of the reasons I kept putting this off was that I couldn't figure out how to do this concisely for these snippets. I think switching to `get_examples` doesn't really clarify this for users, though. We could additionally define `get_examples` in the snippet, or use `lambda: examples` maybe?

ljvmiranda921:
Thanks for the reviews, Paul and Adriane!

> One of the reasons I kept putting this off was that I couldn't figure out how to do this concisely for these snippets. I think switching to get_examples doesn't really clarify this for users, though. We could additionally define get_examples in the snippet, or use lambda: examples maybe?

Sure, I'll define a `get_examples() -> Iterable[Example]` function in the snippet so that it's clearer for users what's typically accepted by the API. I'll make sure to keep that function as concise as possible.

ljvmiranda921:
> I'd also recommend testing all the initialize methods with one example and your exact changes to the docs to make sure you don't run into any other errors, like I'm not sure whether some of the methods have errors if the expected annotation is missing.

Ah yeah, as it turns out there are pipeline components that raise E143 even when provided with an `Example` (unannotated). 

```
ValueError: [E143] Labels for component 'trainable_lemmatizer' not initialized. 
This can be fixed by calling add_label, or by providing a representative batch of examples 
to the component's `initialize` method.
```

I updated the test suite to check them. Later on, I'll create separate tests for pipelines that expect annotations to ensure that the updated documentation works correctly (unless there are already existing tests). You can check the test by running:

```sh
python3 -m pytest spacy/tests/pipeline/test_pipe_methods.py --issue 10165
```

adrianeboyd:
Okay, it was good to check so that the details are clear and we know what we can add to the examples without being incorrect, but I don't think it makes sense to have this as a new test (and also the test suite itself never uses any pipelines like `en_core_web_sm`).

I'd remove the test and suggest using `lambda: examples` with a bit more explanation about needing a representative set of examples to get all the labels somewhere. I think that adding a definition for `get_examples` makes this too long without adding a lot of benefit for users, since the types are in the table.

ljvmiranda921:
Hi reviewers!

I updated the PR to reflect the needed documentation changes. You can also check this PR's description to see a **table of differences** regarding the component's `initialize` methods. Some notable things:
- The `sentencizer` documentation doesn't have an `initialize()` section. I reckon it's because this component doesn't need to be trained, so we didn't have to write a doc for `initialize`. I left it the way it is. Let me know if I should update the file.
- I didn't touch the `transformer` method because the documentation works the way it is. Let me know if I should still add some clarifying text, etc.

website/docs/api/entityrecognizer.md
```diff
@@ -189,7 +189,7 @@ This method was previously called `begin_training`.
...
-| `get_examples` | Function that returns gold-standard annotations in the form of [`Example`](/api/example) objects. ~~Callable[[], Iterable[Example]]~~                                                                                                                                                                                                                                                                                  |
+| `get_examples` | Function that returns gold-standard annotations in the form of [`Example`](/api/example) objects. Must contain at least one `Example`. ~~Callable[[], Iterable[Example]]~~                                                                                                                                                                                                                                                                                  |
```
polm, line 192: I guess this works with a non-annotated example if you also provide labels separately, or in the (nonsense) case where you have no labels?

website/docs/api/textcategorizer.md
```diff
@@ -175,7 +175,7 @@ applied to the `Doc` in order. Both [`__call__`](/api/textcategorizer#call) and
 ## TextCategorizer.initialize {#initialize tag="method" new="3"}
 
 Initialize the component for training. `get_examples` should be a function that
-returns an iterable of [`Example`](/api/example) objects. The data examples are
+returns an iterable of [`Example`](/api/example) objects. **At least one annotated example should be supplied.** The data examples are
```
polm, line 178: You should run `prettier` to fix where the newlines are. (Fine to do it after review, since it makes diffs less clear.)

adrianeboyd:
The `Transformer.initialize()` docs are a bad hack to avoid the `validate_examples` check that we should have removed.

It would be nice if we could switch to `lambda: []`, but this is a huge deal, let's just have `lambda: examples` here anyway and then remove the useless check without much fanfare (which is in `spacy-transformers`, not in `spacy`).

website/docs/api/morphologizer.md
```diff
@@ -176,7 +176,7 @@ config.
...
-| `get_examples` | Function that returns gold-standard annotations in the form of [`Example`](/api/example) objects. ~~Callable[[], Iterable[Example]]~~                                                                                                                                                                                                                                                             |
+| `get_examples` | Function that returns gold-standard annotations in the form of [`Example`](/api/example) objects. Must contain at least one annotated `Example`. ~~Callable[[], Iterable[Example]]~~                                                                                                                                                                                                              |
```
adrianeboyd, line 179: The bit about labels vs. annotated examples is also true here, so I'd make a similar change to "must contain at least one example" everywhere.

ljvmiranda921:
Updated the PR! If I understood correctly, I should replace "Must contain at least one **annotated** example" with "Must contain at least one example" because it's a given that `Example` objects should already contain gold-standard annotations. I also updated the transformer docs to reflect the same changes as with the others.

adrianeboyd (approved):

svlandeg:
Nice! Good to have this more consistent/correct/clear throughout the docs.

Issue #11218, opened 2022-07-26
Installation issue on Python 3.9 with option [ko] due to natto-py bug


loctimize:
## How to reproduce the behaviour
When running `pip install -U 'spacy[transformers,lookups,apple,ja,ko,th]'` the installation exits with the error:

````
Collecting natto-py==0.9.0
  Using cached natto-py-0.9.0.tar.gz (35 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      error in natto-py setup command: use_2to3 is invalid.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
````

It seems to be a known issue that has been fixed in natty-py v. 1.0.0, see: https://github.com/buruzaemon/natto-py/issues/119.
Maybe it can be fixed by changing the required version of natto-py in setup.cfg in line 117 to:

````
ko =
    natto-py==1.0.0
````

I tested it on the current master branch and it worked when installing wit ` pip install -U .'[transformers,lookups,apple,ja,ko,th]'`

## Environment

* Operating System: MacOS 12.5
* Python Version Used: 3.9.13
* spaCy Version Used: 3.4.0
* Environment Information: MacBook Pro, Apple M1 Max, 64 GB RAM

polm:
Thanks for the report, that does seem to be an issue - we'll work on fixing it.

In the meantime, for anyone else who runs into this error, you can fix it by manually installing `natto-py` with `pip install -U natto-py`.

PR #11222, opened 2022-07-27
Update natto-py version spec

polm:
## Description

As outlined in #11218, version 0.9.0 of natto-py, which we had pinned, has issues with recent versions of setuptools. This increases the version pin `>=1.0.0,<2.0.0`. 1.0.0 is the current release and the package is feature complete so updates aren't common. 

We could make the version range tighter, but I think in this case it's better for it to be open. I would also be fine with removing the high limit like we do for pythainlp. 

### Types of change

Dependencies change

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -114,7 +114,7 @@ ja =
     sudachipy>=0.5.2,!=0.6.1
     sudachidict_core>=20211220
 ko =
-    natto-py>=1.0.0,<2.0.0
+    natto-py>=0.9.0
 th =
     pythainlp>=2.0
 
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -114,7 +114,7 @@ ja =
     sudachipy>=0.5.2,!=0.6.1
     sudachidict_core>=20211220
 ko =
-    natto-py==0.9.0
+    natto-py>=1.0.0,<2.0.0
```
adrianeboyd, line 117: I don't think there's any particular need to forbid the older versions, since most people will get the most recent one when they install anyway.

```suggestion
    natto-py>=0.9.0
```

None of this is enforced or checked except on the initial install and you can override it without formal conflicts, so open ranges are less of a concern here.

PR #11226, opened 2022-07-27
Add dependency check to project step runs

rmitsch:
## Goal 

Ensure dependencies specified in `requirements.txt` are installed. Notes:
- No tests have been added yet. Dedicated tests would be nice, but we'd need to mess with the test venv for that - not sure that's worth it?
- Documentation will be added once the code review is through.

## Description

Uses `pkg_resources` to check whether all dependencies can be imported without issues. Note that this doesn't check the complete dependency tree - theoretically a sub-dependency might be missing or mismatched without this check noticing.

### Types of change

New feature.

## Checklist


setup.cfg
```diff
@@ -63,6 +63,7 @@ install_requires =
     packaging>=20.0
     typing_extensions>=3.7.4,<4.2.0; python_version < "3.8"
     langcodes>=3.2.0,<4.0.0
+    types-setuptools>=57.0.0
```
adrianeboyd, line 66: Types packages only go in `requirements.txt` for use with mypy, they're not required to run spacy.

```suggestion
```

spacy/cli/project/run.py
```diff
@@ -308,3 +330,26 @@ def get_fileinfo(project_dir: Path, paths: List[str]) -> List[Dict[str, Optional
         md5 = get_checksum(file_path) if file_path.exists() else None
         data.append({"path": path, "md5": md5})
     return data
+
+
+def _validate_requirements(projects_path: Path) -> Tuple[List[str], List[str]]:
```
adrianeboyd, line 335: I think it would make sense to make this a little more general by working from a list of requirement lines instead of a path + hard-coded filename. `project_run` can handle reading the file for a standard project setup.

I think the messages should only be warnings and I don't think we should ever exit as a result. Someone might want to run an individual step from a project (say a corpus preprocessing step) without having everything installed, and we shouldn't prevent them from doing this.

I think I'd suggest moving the warnings into this method. Maybe it could return a status code instead of lists of packages?

adrianeboyd:
Have you looked at the timing for this yet?

setup.cfg
rmitsch: I thought so, but this [fails a check for compatibility between `setup.cfg` and `requirements.txt`](https://github.com/explosion/spaCy/runs/7541263540). Is there an alternative way around this?

adrianeboyd: You have to add it as an exception in `spacy/tests/package/test_requirements.py`.

rmitsch:
> Have you looked at the timing for this yet?

Not yet, wanted to confirm this approach is acceptable first.

adrianeboyd:
If the basic `pkg_resources` checks are extremely slow we can immediately stop working on it. (I don't think they will be, though.)

rmitsch:
> If the basic `pkg_resources` checks are extremely slow we can immediately stop working on it. (I don't think they will be, though.)

So far they felt pretty instantaneous. Will benchmark with more extensive requirement files tomorrow.

rmitsch:
Running this with this cobbled together list of requirements
```
scipy
umap
transformers>=3.4.0,<4.3.0
spacy-transformers>=1.0.1,<1.1.0
stanza>=1.3.0,<1.5.0
flair>=0.6.0,<1.0.0
ufal.udpipe
spacy-lookups-data>=1.0.0,<1.1.0
ml_datasets==0.2.0a0
spacy-experimental==0.2.0
typer
numpy
tqdm
wasabi
pydantic>=1.0.0,<2.0.0
fastapi>=0.61.1,<0.62.0
aiofiles
uvicorn>=0.11.6,<0.12.0
spacy-huggingface_hub
plotext
spacy_ray
spacy-streamlit>=1.0.0a0
streamlit
wandb>=0.12.4,<0.20.0
floret>=0.10.1,<0.11.0
natto-py
spacy-streamlit>=1.0.0a0
spacy-lookups-data>=1.0.0,<1.1.0
google-cloud-storage
streamlit
presidio-analyzer==2.2.1
presidio-anonymizer==2.2.1
skweak==0.2.13
hmmlearn==0.2.6
nlpaug
```
takes - on my local machine - 0.33 seconds in an environment which has all of these packages installed and 0.21 seconds in one that's missing most of these dependencies. I think that's reasonable.

spacy/cli/project/run.py
```diff
@@ -308,3 +314,38 @@ def get_fileinfo(project_dir: Path, paths: List[str]) -> List[Dict[str, Optional
...
+        try:
+            pkg_resources.require(req)
+        except pkg_resources.DistributionNotFound:
+            failed_pkgs.append(req)
+        except pkg_resources.VersionConflict:
+            conflicting_pkgs.append(req)
+
+    if len(failed_pkgs):
+        msg.warn(
+            f"The following requirements could not be validated:\n{failed_pkgs}.",
```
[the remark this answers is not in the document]
rmitsch: `wasabi` seems to return the newlines anyway, so I removed them at both places now.

adrianeboyd (commented):
I think the conflicting dependencies message needs more info. If the conflict is further down in the tree, it's really unclear what's going on. I think we could either skip the conflicting packages check or maybe add more in the message to run `pip check` or something, but right now it says there are conflicts for a package that seems to be installed correctly.

Like this is the output:

```
⚠ The following depencency conflicts were detected:
['spacy-streamlit>=1.0.0a0']
```

The conflict isn't obvious from this output.

And if I put in a non-existent future version, I still get the conflict instead of the requirement message:

```
⚠ The following depencency conflicts were detected:
['spacy-streamlit>=2.0.0a0']
```

If you want to reproduce this, the actual conflict is:

```
streamlit 1.9.2 has requirement click<8.1,>=7.0, but you have click 8.1.3.
```

If I put my whole working environment into the requirements (350 packages) it adds about 1 s to the running time. With a shorter/realistic list it looks like more like 0.1 s. Compared to the running time of actual steps, I think 0.1 s is acceptable.

rmitsch:
> I think the conflicting dependencies message needs more info.

Displaying the full message from `VersionConflict.report()` now. This should show `streamlit 1.9.2 has requirement click<8.1,>=7.0, but you have click 8.1.3.` in the warning message now. Can you verify that?

adrianeboyd:
The merge undid some other edits?

rmitsch:
> The merge undid some other edits?

Weird. But only https://github.com/explosion/spaCy/pull/11226/commits/f58195569c93249d5a1ccbb5fa000706322498e4, right? Re-added the change.

spacy/cli/project/run.py
```diff
@@ -1,5 +1,8 @@
-from typing import Optional, List, Dict, Sequence, Any, Iterable
+import os.path
+from typing import Optional, List, Dict, Sequence, Any, Iterable, Tuple
```
adrianeboyd, line 2:
```suggestion
from typing import Optional, List, Dict, Sequence, Any, Iterable, Tuple
import os.path
```

spacy/cli/project/run.py
```diff
@@ -308,3 +317,37 @@ def get_fileinfo(project_dir: Path, paths: List[str]) -> List[Dict[str, Optional
...
+        try:
+            pkg_resources.require(req)
+        except pkg_resources.DistributionNotFound:
+            failed_pkgs.append(req)
+        except pkg_resources.VersionConflict as vc:
+            conflicting_pkgs_msgs.append(vc.report())
+
+    if len(failed_pkgs):
+        msg.warn(
+            f"The following requirements could not be validated:\n{failed_pkgs}.",
```
adrianeboyd, line 342:
```suggestion
            f"The following requirements were not found:\n{failed_pkgs}.",
```

spacy/cli/project/run.py
```diff
@@ -308,3 +317,37 @@ def get_fileinfo(project_dir: Path, paths: List[str]) -> List[Dict[str, Optional
...
+            "Make sure your Python environment is set up correctly and you installed all requirements specified in "
+            "your project's requirements.txt.",
+        )
+    if len(conflicting_pkgs_msgs):
+        msg.warn(
+            f"The following dependency conflicts were detected:\n{conflicting_pkgs_msgs}",
+            "Make sure your Python environment is set up correctly. Ensure that your project's "
+            "requirements.txt specifies the correct versions for all dependencies.",
```
adrianeboyd, line 350: This is also confusing when the "conflicting" things are the actual requirements vs. conflicts further down in the dependency tree. For example:

```
⚠ The following requirements could not be validated:
['spacy-streamlit==1.0.2'].
Make sure your Python environment is set up correctly and you installed all
requirements specified in your project's requirements.txt.
⚠ The following dependency conflicts were detected: ['spacy-loggers
1.0.3 is installed but spacy-loggers==1.0.2 is required']
Make sure your Python environment is set up correctly. Ensure that your
project's requirements.txt specifies the correct versions for all dependencies.
```

This message is kind of repeated twice and the incorrect version isn't exactly the same thing as other kinds of conflicts? Maybe there's no way to distinguish this, but I think at least the message is confusing.

rmitsch: Should we ditch the message completely? Re-reading it now makes me think it should be obvious to check your env setup/`requirements.txt` if warnings about requirements/dependencies show up.

Regarding conflicts with actual requirements vs. further down the tree: it's easy enough to rephrase the `pkg_resources` output, but I'd keep the original phrasing for consistency.

adrianeboyd: I think the first message is clearer ("Make sure your Python environment is set up correctly and you installed all requirements specified in your project's requirements.txt.") and would like to just see it once even if there are both types of errors.

rmitsch: Combined both into a slightly modified error message that's only shown once. Example output:
```
⚠ Missing requirements or requirement conflicts detected. Make sure
your Python environment is set up correctly and you installed all requirements
specified in your project's requirements.txt. Affected:
["The 'scipy' distribution was not found and is required by the application",
"The 'transformers<4.3.0,>=3.4.0' distribution was not found and is required by
the application", "The 'spacy-transformers<1.1.0,>=1.0.1' distribution was not
found and is required by the application", "The 'stanza<1.5.0,>=1.3.0'
distribution was not found and is required by the application", "The
'flair<1.0.0,>=0.6.0' distribution was not found and is required by the
application", "The 'ufal.udpipe' distribution was not found and is required by
the application", "The 'spacy-experimental==0.2.0' distribution was not found
and is required by the application", "The 'fastapi<0.62.0,>=0.61.1' distribution
was not found and is required by the application", "The 'aiofiles' distribution
was not found and is required by the application", "The
'uvicorn<0.12.0,>=0.11.6' distribution was not found and is required by the
application", "The 'spacy-huggingface_hub' distribution was not found and is
required by the application", "The 'plotext' distribution was not found and is
required by the application", "The 'spacy_ray' distribution was not found and is
required by the application", "The 'spacy-streamlit>=1.0.0a0' distribution was
not found and is required by the application", "The 'streamlit' distribution was
not found and is required by the application", "The 'wandb<0.20.0,>=0.12.4'
distribution was not found and is required by the application", "The
'floret<0.11.0,>=0.10.1' distribution was not found and is required by the
application", "The 'natto-py' distribution was not found and is required by the
application", "The 'spacy-streamlit>=1.0.0a0' distribution was not found and is
required by the application", "The 'google-cloud-storage' distribution was not
found and is required by the application", "The 'streamlit' distribution was not
found and is required by the application", "The 'presidio-analyzer==2.2.1'
distribution was not found and is required by the application", "The
'presidio-anonymizer==2.2.1' distribution was not found and is required by the
application", "The 'skweak==0.2.13' distribution was not found and is required
by the application", "The 'hmmlearn==0.2.6' distribution was not found and is
required by the application", "The 'nlpaug' distribution was not found and is
required by the application", 'ml-datasets 0.2.0 is installed but
ml_datasets==0.2.0a0 is required']
```

adrianeboyd: This looks really hard to read. Can it be split into separate paragraphs?

spacy/cli/project/run.py
```diff
@@ -308,3 +317,32 @@ def get_fileinfo(project_dir: Path, paths: List[str]) -> List[Dict[str, Optional
...
+        except pkg_resources.DistributionNotFound as dnf:
+            failed_pkgs_msgs.append(dnf.report())
+        except pkg_resources.VersionConflict as vc:
+            conflicting_pkgs_msgs.append(vc.report())
+
+    if len(failed_pkgs_msgs) or len(conflicting_pkgs_msgs):
+        msg.warn(
+            title="Missing requirements or requirement conflicts detected. Make sure your Python environment is set up "
+            "correctly and you installed all requirements specified in your project's requirements.txt. "
+            "Affected:",
```
adrianeboyd, line 344:
```suggestion
            "correctly and you installed all requirements specified in your project's requirements.txt: "
```

spacy/cli/project/run.py
rmitsch: `wasabi` uses `textwrap.fill()` internally when preparing the message for printing, which by default seems to convert all space characters into whitespaces. Don't think there's a way around this?

adrianeboyd: Maybe with each split into a separate message with `msg.text` or similar? I really don't think it needs to be the `repr` of a list of strings?

spacy/cli/project/run.py
```diff
@@ -71,6 +74,12 @@ def project_run(
     commands = {cmd["name"]: cmd for cmd in config.get("commands", [])}
     workflows = config.get("workflows", {})
     validate_subcommand(list(commands.keys()), list(workflows.keys()), subcommand)
+
+    req_path = project_dir / "requirements.txt"
+    if os.path.exists(req_path):
+        with req_path.open() as requirements_file:
+            _validate_requirements([req.replace("\n", "") for req in requirements_file])
```
svlandeg, line 81: We need some kind of mechanism to make this optional, or at least provide some option to a user to avoid these warnings. Imagine a project with multiple workflows where a `torch` dependency is only needed for `workflowB` and not `workflowA`- then you don't want to keep seeing these warnings over and over again if you're only running `workflowA`, and you don't want to "forced" into installing a dependency you don't really need.

rmitsch: I hadn't considered that. Perhaps a `--no-validate` (assuming we want to validate dependencies by default) flag for `spacy project run`?

adrianeboyd: I think a `project.yml` setting sounds better?

rmitsch: So if `validate: False` is in `project.yml`, we don't validate; if `validate` is not in the config or `validate: True` is, we validate?

svlandeg: That sounds like a good plan, but perhaps make the setting even more specific so it doesn't clash with other usages of "validating stuff" in a project. Something like `validate_requirements`?

adrianeboyd: I think `validate` is overloaded here and might suggest something like `check_requirements` instead, which also sounds more like `pip check`.

rmitsch: Both's good with me. I'll go with `check_requirements` then, if @svlandeg doesn't mind?

rmitsch: Added.

rmitsch: @adrianeboyd @svlandeg Ping for feedback on `check_requirements`.

website/docs/usage/projects.md
```diff
@@ -148,6 +148,13 @@ skipped. You can also set `--force` to force re-running a command, or `--dry` to
 perform a "dry run" and see what would happen (without actually running the
 script).
 
+By default `spacy projects run` checks your installed dependencies to verify
+that your environment is properly set up and aligns with the project's
+`requirement.txt`, if there is one. If missing or conflicting dependencies are
+detected, a corresponding warning is displayed. If you'd like to disable the
+dependency check, set `check_requirements: False` in your project's
```
adrianeboyd, line 155: Aren't the standard yaml values lowercase?

```suggestion
dependency check, set `check_requirements: false` in your project's
```

website/docs/usage/projects.md
```diff
@@ -148,6 +148,13 @@ skipped. You can also set `--force` to force re-running a command, or `--dry` to
 perform a "dry run" and see what would happen (without actually running the
 script).
 
+By default `spacy projects run` checks your installed dependencies to verify
+that your environment is properly set up and aligns with the project's
+`requirement.txt`, if there is one. If missing or conflicting dependencies are
```
adrianeboyd, line 153:
```suggestion
`requirements.txt`, if there is one. If missing or conflicting dependencies are
```

website/docs/usage/projects.md
rmitsch: I think [both should be fine](https://yaml.org/type/bool.html).

website/docs/usage/projects.md
```diff
@@ -148,6 +148,13 @@ skipped. You can also set `--force` to force re-running a command, or `--dry` to
 perform a "dry run" and see what would happen (without actually running the
 script).
 
+By default `spacy projects run` checks your installed dependencies to verify
+that your environment is properly set up and aligns with the project's
+`requirements.txt`, if there is one. If missing or conflicting dependencies are
+detected, a corresponding warning is displayed. If you'd like to disable the
+dependency check, set `check_requirements: false` in your project's
+`project.yml`.
```
svlandeg, line 156:
```suggestion
Since spaCy v3.4.2, `spacy projects run` checks your installed dependencies to verify
that your environment is properly set up and aligns with the project's
`requirements.txt`, if there is one. If missing or conflicting dependencies are
detected, a corresponding warning is displayed. If you'd like to disable the
dependency check, set `check_requirements: false` in your project's
`project.yml`.
```

spacy/cli/project/run.py
```diff
@@ -71,6 +74,12 @@ def project_run(
     commands = {cmd["name"]: cmd for cmd in config.get("commands", [])}
     workflows = config.get("workflows", {})
     validate_subcommand(list(commands.keys()), list(workflows.keys()), subcommand)
+
+    req_path = project_dir / "requirements.txt"
+    if config.get("check_requirements", True) and os.path.exists(req_path):
```
svlandeg, line 79: Do we want to have this setting in the root of the yml file - i.e. right below `title` and `description`? I guess it's a better spot than in the `vars` section, which is more for user-defined variables?

website/docs/usage/projects.md
svlandeg: This comment is good, but we should also update https://spacy.io/usage/projects#project-yml: the example project should have 
```
check_requirements: true
```
in it's `project.yml`, and the option needs to be added to the table with a note that it's new since 3.4.2.

svlandeg (commented):
Some final questions & comments, then this should be good to merge!

spacy/cli/project/run.py
adrianeboyd: I don't think there's anywhere else that makes sense other than the top level.

website/docs/usage/projects.md
rmitsch: > ...but we should also update https://spacy.io/usage/projects#project-yml: the example project should have...

https://github.com/explosion/projects/pull/131

svlandeg (approved):

adrianeboyd (approved):
It looks like the docs need one more round of prettier, but otherwise this looks good!

PR #11228, opened 2022-07-27
Support full prerelease versions in the compat table

adrianeboyd:
## Description

Switching to minor version based compatibility in the compatibility table has simplified making patch releases and keeps the compatibility table smaller, however it's difficult to publish prerelease model versions for testing because there's no way to specify a prerelease version based on minor version compatibility, and as soon as there's a newer stable-looking minor version in the compatibility table, the website uses it by default for the model entries.

This isn't the only way to solve this problem, but it seemed like a relatively straightforward alternative for dev releases as in #10624, where the models could only be installed through external URLs rather than `spacy download` for v3.3.0.dev0.

### Types of change

?

## Checklist


svlandeg (approved):
Looks fine by me, and I trust your judgement :-)

Issue #10846, opened 2022-05-25
ValueError: [E102] Can't merge non-disjoint spans. - Dutch


kilzone:
## Error message
ValueError: [E102] Can't merge non-disjoint spans. 'opvlamming' is already part of tokens to merge. If you want to find the longest non-overlapping spans, you can use the util.filter_spans helper:
https://spacy.io/api/top-level#util.filter_spans

## Code (use case)
Currently we're trying to use spaCy 3.3 to parse *Dutch* (nl_core_news_lg) texts - I would like the nouns to be merged again, but the error message above comes up. I've noticed several issues about this issue with German language, but not sure if Dutch has also been fixed.

Code:
```python
parser = spacy.load('nl_core_news_lg')
parser.add_pipe("merge_noun_chunks")
parser('okt opvlamming panuveitiss Oculo sinistra > Oculo dextra en alghele malaise moe, tattoos opgezet. 17-03-2020 Ozurdex Oculo dextra et sinistra , opvlamming onder humir luchtweg ondanks wekelijks ada vele corpus vitreum (glasvocht) troebelingen beschreven')
```

## Info about spaCy

- **spaCy version:** 3.3.0
- **Platform:** Windows-10-10.0.25120-SP0
- **Python version:** 3.7.9
- **Pipelines:** nl_core_news_lg (3.3.0)

abdulrahimq:
I'm not sure if you are still having this problem, but if you use the nl_core_news_sm, it will work. I'm unsure how to solve this issue because I'm unfamiliar with spaCy's code base. 

The issue is here:
spacy/pipeline/functions.py

```
    with doc.retokenize() as retokenizer:
        for np in doc.noun_chunks:
            attrs = {"tag": np.root.tag, "dep": np.root.dep}
            retokenizer.merge(np, attrs=attrs)  # type: ignore[arg-type]
    return doc
```

The np in doc.noun_chuncs returns a Span of 9 words. "Oculo dextra en alghele malaise moe, tattoos opgezet." and this breaks the retokenizer.merge(np, attrss=attrs) because it is expecting one Span. If someone has suggestions on how to solve it or where to investigate further, I'd be happy to take a look at it.

adrianeboyd:
Most of the noun chunk iterators check for overlapping spans, but this seems to be missing for Dutch. A PR that adds this would be welcome, in general the check(s) could look similar to this check for English:

[...]

PR #11258, opened 2022-08-02
Fix Dutch [E102] Can't merge non-disjoint spans that happens for nl_core_news_lg 

abdulrahimq:
Fix Dutch [E102] Can't merge non-disjoint spans that happens for nl_core_news_lg
fix!: add if statement to check span

## Description

In file syntax_iterator for nl language during the noun chunking, add a test to make sure the position of the left_edge is less or equal to the span. 

### Types of change

It is a bug fix for the issue reported here #[10846](https://github.com/explosion/spaCy/issues/10846#issuecomment-1203160759)

## Checklist

### DIFF
--- a/spacy/lang/nl/syntax_iterators.py
+++ b/spacy/lang/nl/syntax_iterators.py
@@ -45,9 +45,6 @@ def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Tuple[int, int, int]]:
         # For NOUNS
         # Pick children from syntactic parse (only those with certain dependencies)
         if word.pos == NOUN:
-            # Prevent nested chunks from being produced
-            if word.left_edge.i <= end_span:
-                continue
             # Some debugging. It happens that VERBS are POS-TAGGED as NOUNS
             # We check if the word has a "nsubj", if it's the case, we eliminate it
             nsubjs = filter(
@@ -68,6 +65,8 @@ def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Tuple[int, int, int]]:
         # PRONOUNS only if it is the subject of a verb
         elif word.pos == PRON:
             # Prevent nested chunks from being produced
+            if word.left_edge.i <= end_span:
+                continue
             if word.dep in pronoun_deps:
                 start_span = word.i
                 end_span = word.i + 1
--- a/spacy/lang/nl/syntax_iterators.py
+++ b/spacy/lang/nl/syntax_iterators.py
@@ -43,12 +43,11 @@ def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Tuple[int, int, int]]:
     # Only NOUNS and PRONOUNS matter
     for i, word in enumerate(filter(lambda x: x.pos in [PRON, NOUN], doclike)):
         # For NOUNS
-        # Prevent nested chunks from being produced
-        if word.left_edge.i <= start_span:
-            continue
-
         # Pick children from syntactic parse (only those with certain dependencies)
         if word.pos == NOUN:
+            # Prevent nested chunks from being produced
+            if word.left_edge.i <= start_span:
+                continue
             # Some debugging. It happens that VERBS are POS-TAGGED as NOUNS
             # We check if the word has a "nsubj", if it's the case, we eliminate it
             nsubjs = filter(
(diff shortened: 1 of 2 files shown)


spacy/lang/nl/syntax_iterators.py
```diff
@@ -39,9 +39,14 @@ def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Tuple[int, int, int]]:
...
     span_label = doc.vocab.strings.add("NP")
 
+    start_span = -1
     # Only NOUNS and PRONOUNS matter
     for i, word in enumerate(filter(lambda x: x.pos in [PRON, NOUN], doclike)):
         # For NOUNS
+        # Prevent nested chunks from being produced
+        if word.left_edge.i <= start_span:
+            continue
+
```
adrianeboyd, line 49: This is the right idea, but the `start_span` values that are used below aren't `word.left_edge.i`, so I think this check needs to be moved into each if block below with separate checks to compare to the `start_span` values that will be used by each alternative.

abdulrahimq: I have tried this and don't think this actually works as it is still failing the test for noun_chunking. I believe that the noun_chunking code's logic actually has a problem. I'm trying to see if I can fix that.

adrianeboyd: This should be possible without modifying anything else in the chunking algorithm. Can you commit what you've tried so far, I can take a look...

abdulrahimq: I have pushed the commits. I basically tried all the combinations of moving the if statement in NOUN, PRON, and BOTH, and also tried to use end_span. 

Adding the if `word.left_edge.i` fixes the bug if it was added in NOUN portion but breaks the unit test `def test_chunking(nl_sample, nl_reference_chunking)` if you add it to PRON then the unit test keeps working, but it doesn't fix the bug.

abdulrahimq: If you have pseudocode for Dutch noun chunking or with what labels does it happen because I think the usage of filtering is a bit off in this code but can't confirm it as I don't know Dutch.

adrianeboyd: `word.left_edge.i` isn't the right thing to check because this code uses a different algorithm to choose the start token for the noun chunk spans.

You need to keep track of the end of the most recently yielded span and then check that the following start doesn't overlap with it, using the `start_span` in each if-block.

Let me add a test with the example from issue that checks whether overlapping spans have been produced.

spacy/tests/lang/nl/test_noun_chunks.py
```diff
@@ -207,3 +208,18 @@ def test_chunking(nl_sample, nl_reference_chunking):
     """
     chunks = [s.text.lower() for s in nl_sample.noun_chunks]
     assert chunks == nl_reference_chunking
+
+
+@pytest.mark.issue(10846)
+def test_no_overlapping_chunks(nl_vocab):
+    # fmt: off
```
abdulrahimq, line 215: What do you mean with fmt?

spacy/lang/nl/syntax_iterators.py
abdulrahimq: But isn't this what we are already doing? 
1- The `word.left_edge.i` gives us the following position of the first token of a phrase, which is the following start. 
2- We are keeping track of the end of the previous phrase in `end_span`. 
Why would we compare the first token with the start_span?

adrianeboyd: No, in one case `start_span = min(children_i)` and in the other `start_span = word.i`. Neither of these are `word.left_edge.i`. See the proposal in #11275.

spacy/tests/lang/nl/test_noun_chunks.py
svlandeg: "fmt" is short for "formatting" - it's a meta code signaling to the `black` formatting tool that the following lines can be kept as-is.

spacy/lang/nl/syntax_iterators.py
abdulrahimq: It makes more sense after seeing your solution!

abdulrahimq: Just to confirm I'll close this PR since your PR will be merged?

adrianeboyd: Yes, let's close this in favor of the other PR. Thanks for bringing this to our attention again!

Issue #11260, opened 2022-08-03
Example.to_dict doesn't include spans


stefawolf:
I'm working with data augmentation for examples that include spans. I think there is a bug in the Example.to_dict method where data about spans gets lost.

```
def to_dict(self):
        return {
            "doc_annotation": {
                "cats": dict(self.reference.cats),
                "entities": doc_to_biluo_tags(self.reference),
                "links": self._links_to_dict()
            },
           ....
      }
```
Is it correct, that something like `spans: dict(self.reference.spans)` should be added? I'll try to open a PR with this solution.

adrianeboyd:
Since `Example.from_dict` supports this, I think it makes sense to include the spans in the `Example.to_dict` output. The spans format in the output needs to be supported by this method, which is expecting a format with the span information in tuples rather than as `Span` objects:

https://github.com/explosion/spaCy/blob/2d89dd9db898e66058bf965e1b483b0019ce1b35/spacy/training/example.pyx#L456-L476

I initially thought `doc.spans` wasn't supported at all here, since it's not in the docs and this is more of a transition format from the simple v2 training dict format than something that is intended to support all possible doc annotation.

stefawolf:
Ok, thanks! I've updated my PR #11261 to use tuples instead of spans now.

PR #11261, opened 2022-08-03
adding spans to doc_annotation in Example.to_dict

stefawolf:
Added spans to doc_annotation in the Example.to_dict method.

## Description
I've added a test for the issue in test_new_example.py. I'm not sure, if it's the right place and if it should be tested differently.

### Types of change
Fix to #11260 

## Checklist

### DIFF
--- a/spacy/tests/training/test_new_example.py
+++ b/spacy/tests/training/test_new_example.py
@@ -455,8 +455,8 @@ def test_issue11260(annots):
 
     output_dict = example.to_dict()
     assert "spans" in output_dict["doc_annotation"]
-    assert output_dict["doc_annotation"]["spans"]["cities"] == [(7, 15, "LOC")]
-    assert output_dict["doc_annotation"]["spans"]["people"] == [(0, 1, "PERSON")]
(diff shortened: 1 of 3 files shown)


stefawolf:
I can see that I was too quick yesterday afternoon, as I missed that my UnitTest was failing.

I'm expecting the labels in the tuples to be the string values, but am adding the hash values of the labels.  I can't see from `_add_spans_to_doc`, that it should be one way or the other. How should it look like? 

I'd forgotten to add the kb_id as well, I guess that should be handled similarly?

adrianeboyd:
For `_add_spans_to_doc` it doesn't actually matter, but to be more consistent with other uses and to make it easier to inspect the dict, I'd use strings for `label` and `kb_id`.

adrianeboyd:
Let me simplify the test a little and update the docs, and I think this should be ready to merge, thanks! (Also, I think ... Grüße aus Tübingen!)

stefawolf:
Great, thanks! Und viele Grüße aus Odense nach Tübingen :)

PR #11264, opened 2022-08-03
Add WindowEncoder, which support various activations

danieldk:
## Description

Window encoder that supports multiple element-wise activations.

Currently very drafty, mostly for some experiments with @kadarakos.

### Types of change

Feature

## Checklist


danieldk:
@kadarakos What do we want to do with this? Do we still want to do experiments or is this already useful as-is?

danieldk:
Closing this for the time being, since we don't currently have the time to do the experiments.

PR #11268, opened 2022-08-04
Refactor KB for easier customization

rmitsch:
## Goal

Refactor the existing `KnowledgeBase` class to facilitate easier customization and do away with the tight coupling between KB interface and its internal implementation.

_Note_: Documentation updates and additional tests are still missing and will be added post-review.

## Description

- An (abstract) base class for `KnowledgeBase` derivatives.
- Backwards-compatible adjustments to the existing `KnowledgeBase` implementation to adhere to the new interface,
- Consideration of batched candidate generation/lookup routines.

## Checklist


spacy/kb.pxd
```diff
@@ -1,36 +1,20 @@
...
 from preshed.maps cimport PreshMap
 from libcpp.vector cimport vector
 from libc.stdint cimport int32_t, int64_t
 from libc.stdio cimport FILE
 
-from .vocab cimport Vocab
 from .typedefs cimport hash_t
 from .structs cimport KBEntryC, AliasC
-
+from .kb_base cimport BaseKnowledgeBase
```
rmitsch, line 9: `BaseKnowledgeBase` and `KnowledgeBase` are placeholders - we can find better names, I think.

spacy/kb_base.pyx
```diff
@@ -0,0 +1,168 @@
...
+
+
+def get_candidates_batch(kb: BaseKnowledgeBase, mentions: Iterable[Union[Span, str]]) -> Iterable[Iterator[Candidate]]:
+    """
+    Return candidate entities for the given mentions and fetching appropriate entries from the index.
+    kb (BaseKnowledgeBase): Knowledge base to query.
+    mention (Iterable[Union[Span, str]]): Entity mentions for which to identify candidates.
+    RETURNS (Iterable[Iterator[Candidate]]): Identified candidates.
+    """
+    return kb.get_candidates_batch(mentions)
```
rmitsch, line 72: I went back and forth about whether it makes sense to add a swappable method like `get_candidates()` for batches. In the end I went for it, mostly for consistency. I don't like the current situation that much because we have an mostly useless layer of abstraction here. Ideally we want users to work with `BaseKnowledgeBase.get_candidates()` and `BaseKnowledgeBase.get_candidates_batch()` directly. Actually just the batched version, but getting rid of the non-batched version would obviously break stuff.

So we are redirecting calls once more than necessary or useful. This might be confusing for users, but this seems like the least bad/confusing, most consistent alternative to me that is also non-breaking. Open to other ideas. 

In v4 I'd toss out the non-batched version and the `get_candidates`/`get_candidates_batch` config options.

spacy/kb_base.pyx
```diff
@@ -0,0 +1,168 @@
...
+        """Create a KnowledgeBase."""
+        self.vocab = vocab
+        self._entity_vector_length = entity_vector_length
+        self.mem = Pool()
+
+    property entity_vector_length:
+        def __get__(self):
+            return self._entity_vector_length
+        def __set__(self, int value):
+            self._entity_vector_length = value
```
rmitsch, line 103: The previous double-declaration/usage of `entity_vector_length` bugged me. If necessary, we can revert this - should be backwards-compatible though.

spacy/ml/models/entity_linker.py
```diff
@@ -77,24 +83,46 @@ def span_maker_forward(model, docs: List[Doc], is_train) -> Tuple[Ragged, Callab
     return outputs, lambda x: []
 
 
+@registry.misc("spacy.KBType.v1")
+def get_kb_type() -> Type[BaseKnowledgeBase]:
+    """Returns KB type."""
+    return KnowledgeBase
```
rmitsch, line 89: This is IMO useful because even though we have the KB loader in the `initialize` section, the default loader function `load_kb()` doesn't know which type of `BaseKnowledgeBase` to initialize. 
So either we require the user to provide custom implementations of KB loader and `empty_kb`, or we make them specify the KB type and automatically load/initialize the correct KB type.

Since the loading and empty initialization is boilerplate as long as no additional arguments are needed, this seems more intuitive to me. We could also drop this and require custom loaders and initializers for custom KBs though.

spacy/ml/models/entity_linker.py
```diff
@@ -77,24 +83,46 @@ def span_maker_forward(model, docs: List[Doc], is_train) -> Tuple[Ragged, Callab
...
+
+
 @registry.misc("spacy.KBFromFile.v1")
-def load_kb(kb_path: Path) -> Callable[[Vocab], KnowledgeBase]:
-    def kb_from_file(vocab):
-        kb = KnowledgeBase(vocab, entity_vector_length=1)
+def load_kb(
+    kb_path: Path,
+) -> Callable[[Vocab, Type[BaseKnowledgeBase]], BaseKnowledgeBase]:
+    def kb_from_file(vocab: Vocab, kb_type: Type[BaseKnowledgeBase] = KnowledgeBase):
```
rmitsch, line 96: Default of `KnowledgeBase` is for backwards compatibility.

spacy/pipeline/entity_linker.py
```diff
@@ -74,11 +85,18 @@ def make_entity_linker(
     incl_prior: bool,
     incl_context: bool,
     entity_vector_length: int,
-    get_candidates: Callable[[KnowledgeBase, Span], Iterable[Candidate]],
+    get_candidates: Callable[
+        [BaseKnowledgeBase, Union[Span, str]], Iterator[Candidate]
```
rmitsch, line 89: Strictly speaking we expect `get_candidates()` to return `Union[Iterator[Candidate], Iterable[Candidate]]`, right? As long as it's convertable to a list, it should be fine.

spacy/pipeline/entity_linker.py
```diff
@@ -292,6 +345,8 @@ def batch_has_learnable_example(self, examples):
 
         for eg in examples:
             for ent in eg.predicted.ents:
+                # todo Do we want batching here too? I'm assuming that in most cases this finds a learnable example soon
+                #  anyway - and stops afterwards - so fetching candidates for the entire batch is maybe not faster.
```
rmitsch, line 349: See comment in code. I think batching here might not be that useful, but I'm also not sure about the performance implications of `has_learnable_example()`.

spacy/pipeline/entity_linker.py
```diff
@@ -419,66 +474,92 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
+                candidates = (
+                    self.get_candidates_batch(
+                        self.kb, [ent_batch[idx] for idx in valid_ent_idx]
+                    )
+                    if self.candidates_batch_size > 1
+                    else [
+                        self.get_candidates(self.kb, ent_batch[idx])
+                        for idx in valid_ent_idx
+                    ]
+                )
```
rmitsch, line 497: The reason why I'm not using the batched version regardless of the batch size is that `get_candidates` and `get_candidates_batch` are two independent config entries. In this draft `get_candidates` is run if the batch size is 1, otherwise `get_candidates_batch` is executed. 
This seems to be mostly aligned with what I'd expect when configuring my pipeline.

spacy/pipeline/entity_linker.py
```diff
@@ -419,66 +474,92 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
+                batch_candidates = (
+                    self.get_candidates_batch(
+                        self.kb, [ent_batch[idx] for idx in valid_ent_idx]
+                    )
+                    if self.candidates_batch_size > 1
+                    else [
+                        self.get_candidates(self.kb, ent_batch[idx])
+                        for idx in valid_ent_idx
+                    ]
+                )
```
rmitsch, line 497: The reason why I'm not batching by default is that the existence of both `get_candidates` and `get_candidates_batch` in the config implies that both are being used. It seemed mostly aligned with what I'd expect as a user that `get_candidates` gets called with a `batch_size == 1` and `get_candidates_batch` otherwise.

spacy/kb_base.pxd
```diff
@@ -0,0 +1,25 @@
...
+cdef class Candidate:
+    cdef readonly BaseKnowledgeBase kb
+    cdef hash_t entity_hash
+    cdef float entity_freq
+    cdef vector[float] entity_vector
+    cdef hash_t alias_hash
+    cdef float prior_prob
+
+
+cdef class BaseKnowledgeBase:
```
rmitsch, line 22: It irks me that I couldn't figure out how to mark this class as abstract. Is there really no way to do so in Cython?

svlandeg: I don't think so. We have a similar situation with `cdef class Pipe` - basically it's "abstract" by raising `NotImplementedError`'s if you'd use it directly...

spacy/kb_base.pyx
svlandeg: So basically the conflict we've got here is being able to overwrite some of the functionality in the config, versus providing inheritance as a means to overwrite that same functionality.

I guess our initial reasoning was to allow users to only overwrite the candidate generation without touching the KB class itself. But as we've seen, the two are often so tightly coupled that maybe that didn't make so much sense, or was covering only a subpart of the problem efficiently.

At this point (v3) as you've said we can't remove the config option, so adding a batched one feels like the most consistent to me as well. I also assume that most users won't actually see the internal code structure here so hopefully it shouldn't be too confusing.

spacy/kb_base.pyx
```diff
@@ -0,0 +1,168 @@
...
+cdef class Candidate:
+    """A `Candidate` object refers to a textual mention (`alias`) that may or may not be resolved
+    to a specific `entity` from a Knowledge Base. This will be used as input for the entity linking
+    algorithm which will disambiguate the various candidates to the correct one.
+    Each candidate (alias, entity) pair is assigned to a certain prior probability.
+
+    DOCS: https://spacy.io/api/kb/#candidate_init
+    """
+
+    def __init__(self, BaseKnowledgeBase kb, entity_hash, entity_freq, entity_vector, alias_hash, prior_prob):
```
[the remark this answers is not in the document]
svlandeg: I'd be in favour of leaving the name of `KnowledgeBase` for the "base" concept, and using that name in most places. Then our default implementation could be in `kb_memory.pyx` or something, and we could call it something like `InMemoryKnowledgeBase` or SomethingLessTerrible and that would still be of type `KnowledgeBase`, which would then make the bwd-compat hopefully even more straightforward.

spacy/ml/models/entity_linker.py
```diff
@@ -1,11 +1,17 @@
...
 
 from ...util import registry
-from ...kb import KnowledgeBase, Candidate, get_candidates
+from ...kb_base import (
+    BaseKnowledgeBase,
+    Candidate,
+    get_candidates,
+    get_candidates_batch,
+)
+from ...kb import KnowledgeBase
```
svlandeg, line 14: When a line gets too long like this, we repeat the `from ... import` bit so it becomes 2 lines. You already have the second line, so you can shift one or more two imports onto that one and get rid of the `()` again :-)

spacy/ml/models/entity_linker.py
svlandeg: Hmmm this looks odd to me, but perhaps I'm not fully understanding the issue. Ideally each KB should be able to load/initialize itself, no? The `load_kb` option is currently tied to our specific default Cython implementation, and that should probably stay as such. Agreed it's pretty generic but it's also very little effort to add if you're already reimplementing a KB?

spacy/ml/models/entity_linker.py
```diff
@@ -77,24 +83,45 @@ def span_maker_forward(model, docs: List[Doc], is_train) -> Tuple[Ragged, Callab
...
+) -> Callable[[Vocab, Type[BaseKnowledgeBase]], BaseKnowledgeBase]:
+    def empty_kb_factory(
+        vocab: Vocab, kb_type: Type[BaseKnowledgeBase] = KnowledgeBase
+    ):
+        return kb_type(vocab=vocab, entity_vector_length=entity_vector_length)
 
     return empty_kb_factory
 
 
 @registry.misc("spacy.CandidateGenerator.v1")
```
svlandeg, line 116: If the typing changes, we probably need to bump versions?

spacy/pipeline/entity_linker.py
svlandeg: I think the typing was wrong before? It should have been `Iterable`, no?

svlandeg: The docs actually have it correct, `Iterable`! So we can change this as a "bug fix".

spacy/pipeline/entity_linker.py
svlandeg: Ah difficult to say. Maybe we can leave this as is for now.

svlandeg (commented):
Good first draft, left some first comments.

spacy/kb.pyx
```diff
@@ -286,11 +227,15 @@ cdef class KnowledgeBase:
...
 
+    def get_candidates(self, mention: Union[Span, str]) -> Iterator[Candidate]:
+        return self.get_alias_candidates(mention.text if isinstance(mention, Span) else mention)  # type: ignore
+
     def get_alias_candidates(self, str alias) -> Iterator[Candidate]:
         """
         Return candidate entities for an alias. Each candidate defines the entity, the original alias,
         and the prior probability of that alias resolving to that entity.
         If the alias is not known in the KB, and empty list is returned.
+        This particular function is optimized to work with the built-in KB functionality.
```
polm, line 238: Should we use the class name here so that this stays valid if we add other implementations?

spacy/kb_base.pyx
polm: I think this is similar with what we went through adding patterns with the Matcher, and this does look like the best solution. In v4 I think it'd be good to go back to the simple name (`get_candidates`) but make it clear the input type should be a list.

spacy/kb_base.pyx
```diff
@@ -0,0 +1,168 @@
...
+        mention (Union[Span, str]): Mention for which to get candidates.
+        RETURNS (Iterator[Candidate]): Identified candidates.
+        """
+        raise NotImplementedError
+
+    def get_vectors(self, entities: Iterable[str]) -> Iterable[Iterable[float]]:
+        """
+        Return vectors for entities.
+        entity (str): Entity name/ID.
+        RETURNS (Iterable[List[float]]): Vectors for specified entities.
```
polm, line 129: The Docstring and type annotation disagree about the return type. The Docstring is more specific and not wrong, but I guess you want the type annotation one here for future flexibility.

spacy/ml/models/entity_linker.py
polm: I agree this is pretty weird looking. If the only intended use of this is related to initialization (which isn't obvious to me from the registry declaration / function here) then I think it makes sense to just register the initialization stuff directly, even if it's a little boilerplate.

spacy/pipeline/entity_linker.py
```diff
@@ -90,17 +108,22 @@ def make_entity_linker(
...
         component must provide entity annotations.
+    candidates_batch_size (int): Size of batches for entity candidate generation.
     threshold (Optional[float]): Confidence threshold for entity predictions. If confidence is below the threshold,
         prediction is discarded. If None, predictions are not filtered by any threshold.
+    kb_type (Type[BaseKnowledgeBase]): KB type.
     """
 
     if not model.attrs.get("include_span_maker", False):
-        # The only difference in arguments here is that use_gold_ents is not available
+        # The only differences in arguments here are that use_gold_ents is not available and threshold isn't available.
```
polm, line 126:
```suggestion
        # The only difference in arguments here is that use_gold_ents and threshold aren't available.
```

spacy/pipeline/entity_linker.py
```diff
@@ -41,6 +41,14 @@
 DEFAULT_NEL_MODEL = Config().from_str(default_model_config)["model"]
 
 
+"""
+    [initialize.components.entity_linker.kb_loader]
+    @misc = "spacy.CustomKB.v1"
+    entity_vector_length = 342
+    custom_field = 666
+"""
```
polm, line 49:
```suggestion
"""
[initialize.components.entity_linker.kb_loader]
@misc = "spacy.CustomKB.v1"
entity_vector_length = 342
custom_field = 666
"""
```

Is there a reason this is indented?

spacy/kb_base.pyx
```diff
@@ -0,0 +1,168 @@
...
+        def __get__(self):
+            return self._entity_vector_length
+        def __set__(self, int value):
+            self._entity_vector_length = value
+
+    def get_candidates_batch(self, mentions: Union[Iterable[Span], Iterable[str]]) -> Iterable[Iterator[Candidate]]:
+        """
+        Return candidate entities for specified texts. Each candidate defines the entity, the original alias,
+        and the prior probability of that alias resolving to that entity.
+        If the no candidate is found for a given text, an empty list is returned.
```
polm, line 109:
```suggestion
        If no candidate is found for a given text, an empty list is returned.
```

spacy/kb_base.pyx
```diff
@@ -0,0 +1,168 @@
...
+
+from .tokens import Span
+from .util import SimpleFrozenList
+
+
+cdef class Candidate:
+    """A `Candidate` object refers to a textual mention (`alias`) that may or may not be resolved
+    to a specific `entity` from a Knowledge Base. This will be used as input for the entity linking
+    algorithm which will disambiguate the various candidates to the correct one.
+    Each candidate (alias, entity) pair is assigned to a certain prior probability.
```
polm, line 19:
```suggestion
    Each candidate (alias, entity) pair is assigned a certain prior probability.
```

spacy/pipeline/entity_linker.py
```diff
@@ -233,16 +286,16 @@ def initialize(
...
         """Initialize the pipe for training, using a representative set
         of data examples.
 
         get_examples (Callable[[], Iterable[Example]]): Function that
             returns a representative sample of gold-standard Example objects.
         nlp (Language): The current nlp object the component is part of.
-        kb_loader (Callable[[Vocab], KnowledgeBase]): A function that creates a KnowledgeBase from a Vocab instance.
-            Note that providing this argument, will overwrite all data accumulated in the current KB.
+        kb_loader (Callable[[Vocab], BaseKnowledgeBase]): A function that creates a BaseKnowledgeBase from a Vocab
+            instance. Note that providing this argument, will overwrite all data accumulated in the current KB.
```
polm, line 298:
```suggestion
            instance. Note that providing this argument will overwrite all data accumulated in the current KB.
```

polm (commented):
I had several small nitpicks with text but it looks good in general, thank you for pulling this together!

spacy/ml/models/entity_linker.py
rmitsch: Fair enough, it is very little effort to write those. Maybe saving that little amount of work doesn't justify introducing a new config option. Will revert.

spacy/kb_base.pyx
rmitsch: Agree to returning to `get_candidates` in v4.

spacy/pipeline/entity_linker.py
rmitsch: No, just a typo.

spacy/kb.pyx
rmitsch: Actually, on reading it again it seems kinda redundant. This is part of the class that is the "built-in KB functionality". I think we could remove this line?

spacy/kb_base.pyx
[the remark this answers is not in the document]
rmitsch: Renaming the current `KnowledgeBase` would be breaking too though, no? E.g. if users inherited from that class.

spacy/kb.pyx
polm: Sure, that sounds fine.

spacy/kb.pxd
svlandeg: Internal proposal: `InMemoryLookupKB`

setup.py
```diff
@@ -31,6 +31,7 @@
     "spacy.vocab",
     "spacy.attrs",
     "spacy.kb",
+    "spacy.kb_inmemorylookup",
```
svlandeg, line 34: Maybe a bit too long to parse, how about `kb_inmemory_lookup`?

spacy/errors.py
```diff
@@ -707,9 +707,9 @@ class Errors(metaclass=ErrorsWithCodes):
             "`nlp.enable_pipe` instead.")
     E927 = ("Can't write to frozen list Maybe you're trying to modify a computed "
             "property or default function argument?")
-    E928 = ("A KnowledgeBase can only be serialized to/from from a directory, "
+    E928 = ("A InMemoryLookupKB can only be serialized to/from from a directory, "
```
svlandeg, line 710:
```suggestion
    E928 = ("An InMemoryLookupKB can only be serialized to/from from a directory, "
```

svlandeg (commented):
I realise this is annoying, but I wonder whether there's a nicer way to keep track of the history here. Basically if we would have renamed the old `kb.pyx` to `kb_inmemory_lookup.pyx` first, before adding the new, minimal `kb.pyx` I think it would support tracking changes across history on Github better? 

Not sure this will actually feasibile within this same PR though...

adrianeboyd:
I'd suggest a `kb/` subdirectory so this isn't cluttering up the top-level submodules?

setup.py
polm: Why not `kb_in_memory_lookup`?

svlandeg: Yea, that's better yet.

svlandeg:
> I'd suggest a `kb/` subdirectory so this isn't cluttering up the top-level submodules?

I was actually thinking the same thing, just that it would be nice to still be able to import `KnowledgeBase` from `spacy.kb` for v3 - and then we can clean up further for v4?

adrianeboyd:
You can move it into a subdirectory without affecting this as long as it's imported in `__init__.py`.

rmitsch:
> I realise this is annoying, but I wonder whether there's a nicer way to keep track of the history here. ...

Hm, that's what I did, actually. I renamed `kb.pyx` to `kb_inmemorylookup.pyx` first, then `kb_base.pyx` to `kb.pyx`. Thought that would be sufficient for tracking :neutral_face:

adrianeboyd:
I think git uses heuristics based on the modifications to decide whether new files are related to old files, so it's not always what you did with renaming (which is just add+rm underneath).

setup.py
rmitsch: Since we'll have a `kb` subdir, should we rename the modules to smth. like `base.pyx` and `in_memory_lookup.pyx`? Or do we want to stick with `kb.pyx` and `kb_in_memory_lookup.pyx`?

svlandeg: I'd lean to `kb/kb.pyx` and `kb/in_memory.pyx`, maybe the "lookup" bit is not needed for the module name? We're not going to create another in-memory implementation any time soon, right?

rmitsch: Don't think so, no. Adjusted accordingly.

spacy/cli/debug_config.py
```diff
@@ -1,7 +1,7 @@
 from typing import Optional, Dict, Any, Union, List
 from pathlib import Path
 from wasabi import msg, table
-from thinc.api import Config
+from thinc.config import Config
```
svlandeg, line 4: Why did this need to change? In general we prefer importing from `thinc.api`.

adrianeboyd, line 4: Undo this so future merges are easier.

spacy/ml/models/entity_linker.py
```diff
@@ -1,11 +1,13 @@
...
-from thinc.api import chain, clone, list2ragged, reduce_mean, residual
-from thinc.api import Model, Maxout, Linear, noop, tuplify, Ragged
+from thinc.api import chain, list2ragged, reduce_mean, residual
+from thinc.api import Model, Maxout, Linear, tuplify, Ragged
 
 from ...util import registry
-from ...kb import KnowledgeBase, Candidate, get_candidates
+from ...kb import KnowledgeBase
+from ...kb import Candidate, get_candidates, get_candidates_batch
+from ...kb import InMemoryLookupKB
```
svlandeg, line 10:
```suggestion
from ...kb import KnowledgeBase, InMemoryLookupKB
from ...kb import Candidate, get_candidates, get_candidates_batch
```

spacy/ml/models/entity_linker.py
```diff
@@ -78,23 +80,36 @@ def span_maker_forward(model, docs: List[Doc], is_train) -> Tuple[Ragged, Callab
 
 
 @registry.misc("spacy.KBFromFile.v1")
-def load_kb(kb_path: Path) -> Callable[[Vocab], KnowledgeBase]:
-    def kb_from_file(vocab):
-        kb = KnowledgeBase(vocab, entity_vector_length=1)
+def load_kb(
+    kb_path: Path,
+) -> Callable[[Vocab], InMemoryLookupKB]:
```
svlandeg, line 85: I think we shouldn't change the type of this existing registry function, to avoid any kind of issue with type checking in user's applications. We could in time retire this function for a new version with a naming that makes it more clear that it's the `InMemoryLookupKB` that is returned.

spacy/ml/models/entity_linker.py
```diff
@@ -78,23 +80,36 @@ def span_maker_forward(model, docs: List[Doc], is_train) -> Tuple[Ragged, Callab
...
     return kb_from_file
 
 
 @registry.misc("spacy.EmptyKB.v1")
-def empty_kb(entity_vector_length: int) -> Callable[[Vocab], KnowledgeBase]:
-    def empty_kb_factory(vocab):
-        return KnowledgeBase(vocab=vocab, entity_vector_length=entity_vector_length)
+def empty_kb(
+    entity_vector_length: int,
+) -> Callable[[Vocab], InMemoryLookupKB]:
```
svlandeg, line 97: Same comment about the type.

spacy/kb/kb_in_memory.pyx
```diff
@@ -12,85 +11,32 @@ from libcpp.vector cimport vector
...
+from .kb import get_candidates as _get_candidates
 
-cdef class Candidate:
-    """A `Candidate` object refers to a textual mention (`alias`) that may or may not be resolved
-    to a specific `entity` from a Knowledge Base. This will be used as input for the entity linking
-    algorithm which will disambiguate the various candidates to the correct one.
-    Each candidate (alias, entity) pair is assigned to a certain prior probability.
+# For backwards compatibility.
+Candidate = _Candidate
+get_candidates = _get_candidates
```
adrianeboyd, line 25: I don't understand why you're doing this rather than just importing them under their normal names?

spacy/ml/models/entity_linker.py
```diff
@@ -78,23 +80,36 @@ def span_maker_forward(model, docs: List[Doc], is_train) -> Tuple[Ragged, Callab
...
+        return InMemoryLookupKB(vocab=vocab, entity_vector_length=entity_vector_length)
 
     return empty_kb_factory
 
 
 @registry.misc("spacy.CandidateGenerator.v1")
-def create_candidates() -> Callable[[KnowledgeBase, Span], Iterable[Candidate]]:
+def create_candidates() -> Callable[
+    [KnowledgeBase, Union[Span, str]], Iterator[Candidate]
+]:
```
svlandeg, line 107: Why do we need to change the types for this one?

spacy/cli/debug_config.py
rmitsch: We don't, that was a mistake on my part. Reverted in next commit.

spacy/ml/models/entity_linker.py
svlandeg: If we really need a change of types, I'd prefer a version bump.

spacy/pipeline/entity_linker.py
```diff
@@ -204,22 +232,33 @@ def __init__(
...
         self.threshold = threshold
 
-    def set_kb(self, kb_loader: Callable[[Vocab], KnowledgeBase]):
+        if candidates_batch_size < 1:
+            raise ValueError(Errors.E1044)
+
+    def set_kb(
+        self,
+        kb_loader: Union[
+            Callable[[Vocab, Type[KnowledgeBase]], KnowledgeBase],
```
svlandeg, line 252: What's this additional type for? (it's also not in the doc string)

svlandeg (commented):
Overall looks good! Commented on a few more minor details.

spacy/kb/candidate.pyx
```diff
@@ -0,0 +1,74 @@
...
+    @property
+    def entity_vector(self) -> Iterable[float]:
+        return self.entity_vector
+
+    @property
+    def prior_prob(self) -> float:
+        return self.prior_prob
+
+
+def get_candidates(kb: KnowledgeBase, mention) -> Iterable[Candidate]:
```
rmitsch, line 57: I remove the typing `Union[Span, str]` for `mention` (and for `mentions` further down for `get_candidates_batch()`) here, since importing `Span` here raises an `ImportError` further down the road in `tokens/serialize` due to an `attempted relative import beyond top-level package`. Not sure how to get rid of this, any ideas?

This is not preventing us from merging, I'd say - but it'd be nice to have the proper typing here as soon as possible.

rmitsch: Ah, please dismiss this. Importing it directly from `tokens` works.

adrianeboyd: Are you sure the import was correct? That sounds weird.

spacy/ml/models/entity_linker.py
rmitsch: We don't _need_ one, but the `KnowledgeBase` interface currently supports `Union[Span, str]` and I considered this a non-breaking change (guess type checks might fail though?). 
Should we go with `Span` instead? The motivation was that it might be more convenient to pass only strings instead of `Span`s and that it should be up to the `KnowledgeBase` implementation. I don't mind requiring `Span`s though either.

svlandeg: (there's not just the `Span` but also the `Iterator` change)

spacy/kb/candidate.pyx
rmitsch: The original `from tokens import Span` was auto-generated by PyCharm - I didn't notice at first that it wasn't `from ..tokens import Span`. The latter works, the former causes the mentioned import error.

spacy/pipeline/entity_linker.py
rmitsch: Leftover, removed in next commit.

spacy/ml/models/entity_linker.py
rmitsch: `Iterator` change has been reverted.

svlandeg: I think requiring `Span`'s is fine - you'd expect there is always some text-bound mention which has offsets, making it a `Span` and defining the context in case the candidate generator wants to use it.

rmitsch:
Docs have been updated. Moved this to @polm's board and will move it to @svlandeg's afterwards.

polm (commented):
Looks good! I found one typo but otherwise I think this is good to go. Nice work!

spacy/kb/candidate.pxd
```diff
@@ -0,0 +1,12 @@
+from .kb cimport KnowledgeBase
+from libcpp.vector cimport vector
+from ..typedefs cimport hash_t
+
+# Object used by the Entity Linker that summarizes one entity-alias candidate combination.
```
richardpaulhudson, line 5: I find the term `alias` very misleading. If we're doing major refactoring, perhaps we could consider changing it to `mention`?

spacy/kb/kb.pxd
```diff
@@ -0,0 +1,10 @@
+"""Knowledge-base for entity or concept linking."""
+
+from cymem.cymem cimport Pool
+from libc.stdint cimport int64_t
+from ..vocab cimport Vocab
```
richardpaulhudson, line 5: Isn't using a `Vocab` an implementation detail of the current implementation that may well not be shared by other implementations?

spacy/kb/candidate.pxd
rmitsch: That's a good point, but it would be breaking. Other attributes in `Candidate` should IMO also be renamed or dropped, since they aren't independent from the implementation of `KnowledgeBase`. E.g. `entity_freq`, or the fact that depending on the implementation there is no need for hashes - the minimal set of attributes would IMO be just an `entity_id` and an abstract property for returning some identifiable version of the mention.

We could rename `alias_hash` to `mention_hash` and add a property `alias_hash` returning the value of `mention_hash` to maintain backwards compatibility. At the same time it might be better to leave things as they are - i.e. slightly misleading - and do a bigger refactoring only with v4 (which is what I'd opt for).

What do you think, @svlandeg?

svlandeg: IIRC, in the original code `alias` and `mention` are used for two slightly different things. A mention is an actual `Span` in an actual `Doc`, and an alias is a synonym or term associated to a certain identifier in the knowledge base / ontology.

The confusion then was introduced because the built-in look-up function only uses the textual string of the mention to retrieve alias-ID pairs from the KB that could be relevant for that mention, at which point "alias" and "mention" became almost interchangeable because both refered to the textual string only and even verbatim (no fuzzy matching). But this is only true for the built-in look-up implementation. In the more general case, I agree that a `Candidate` should in fact link a `mention` to a likely entity, and whether this matching happens through an alias or an embedding similarity or whatever should be irrelevant for the `entity_linker` (because it's the candidate generator's job).

As Raphael points out, all of this refactoring/renaming should be kept for `v4` though. (are you keeping track of these things for the next PR? ;-))

spacy/kb/kb.pxd
rmitsch: Hm, that's true. There isn't really a need for it in the base class. @svlandeg Any objections to moving `vocab` to `InMemoryLookupKB`?

spacy/kb/candidate.pxd
rmitsch: I do now :wink: @richardpaulhudson Closing this conversation since we decided not to rename. If you feel there's a need for further discussion, feel free to reopen.

spacy/kb/kb.pxd
svlandeg: We should do it at some point (latest by v4) but I'm not sure we want to do it now, because it'll introduce additional complexity in `Candidate` not being able to rely on `kb.vocab` and then we'll have to end up changing more public APIs ... so maybe better to do this later?

rmitsch: Right, I forgot `Candidate` uses `vocab`. No, let's leave it in. Added it to the list of v4 changes.

spacy/errors.py
```diff
@@ -939,6 +939,11 @@ class Errors(metaclass=ErrorsWithCodes):
              "`{arg2}`={arg2_values} but these arguments are conflicting.")
     E1043 = ("Expected None or a value in range [{range_start}, {range_end}] for entity linker threshold, but got "
              "{value}.")
+    E1044 = ("Expected `candidates_batch_size` to be >= 1, but got: {value}")
+    E1045 = ("Encountered {parent} subclass without `{parent}.{method}` "
+             "method in '{name}'. If you want to use this method, make "
+             "sure it's overwritten on the subclass.")
+    E1046 = ("{cls_name} is an abstract class and cannot be instantiated.")
```
svlandeg, line 946: Lets expand this error message and say that if they want the default KB implementation, they can instantiate `InMemoryLookupKB` instead.

website/docs/api/kb.md
```diff
@@ -222,55 +145,14 @@ Restore the state of the knowledge base from a given directory. Note that the
...
 > kb.from_disk("/path/to/kb")
 > ```
 
 | Name        | Description                                                                                     |
 | ----------- | ----------------------------------------------------------------------------------------------- |
 | `loc`       | A path to a directory. Paths may be either strings or `Path`-like objects. ~~Union[str, Path]~~ |
+| `exclude`   | List of components to exclude. ~~Iterable[str]~~                                                |
 | **RETURNS** | The modified `KnowledgeBase` object. ~~KnowledgeBase~~                                          |
-
-## Candidate {#candidate tag="class"}
```
svlandeg, line 237: By removing some sections from this page, we might be creating broken links on the discussion forum or Stackoverflow. Can we try to prevent this? Maybe have a note here about the refactor and a referral to the in memory KB docs?

rmitsch: Good point. We'd only keep the title (`##`) - and underneath the referral - or all headers (`###`)?

rmitsch: Added a suggestion. We could also leave the `Candidate` section in `kb.md` - moving it to a separate page merely felt more consistent, as it's in its own module now.

svlandeg: It's not strictly necessary. We don't always follow the structure of the codebase so closely in the documentation. I'll have a look at your recent edits!

spacy/pipeline/legacy/entity_linker.py
```diff
@@ -66,10 +66,9 @@ def __init__(
         incl_prior (bool): Whether or not to include prior probabilities from the KB in the model.
         incl_context (bool): Whether or not to include the local context in the model.
         entity_vector_length (int): Size of encoding vectors in the KB.
-        get_candidates (Callable[[KnowledgeBase, Span], Iterable[Candidate]]): Function that
+        get_candidates (Callable[[InMemoryLookupKB, Span], Iterable[Candidate]]): Function that
```
svlandeg, line 69: I don't think we need to change this type?

spacy/pipeline/legacy/entity_linker.py
```diff
@@ -115,7 +114,7 @@ def initialize(
         get_examples (Callable[[], Iterable[Example]]): Function that
             returns a representative sample of gold-standard Example objects.
         nlp (Language): The current nlp object the component is part of.
-        kb_loader (Callable[[Vocab], KnowledgeBase]): A function that creates a KnowledgeBase from a Vocab instance.
+        kb_loader (Callable[[Vocab], InMemoryLookupKB]): A function that creates a InMemoryLookupKB from a Vocab instance.
```
svlandeg, line 117: Same question - do we need to change this type?

spacy/kb/kb.pyx
```diff
@@ -0,0 +1,108 @@
...
+    DOCS: https://spacy.io/api/kb
+    """
+
+    def __init__(self, vocab: Vocab, entity_vector_length: int):
+        """Create a KnowledgeBase."""
+        # Make sure abstract KB is not instantiated.
+        if self.__class__ == KnowledgeBase:
+            raise TypeError(
+                Errors.E1046.format(cls_name=self.__class__.__name__)
+            )
```
svlandeg, line 27: This will break the `nel_emerson` [project](https://github.com/explosion/projects/tree/v3/tutorials/nel_emerson) :(

website/docs/api/candidate.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Candidate
```
svlandeg, line 2: I'd be slightly more in favour of keeping `Candidate` with `KnowledgeBase`. Then if at some point we want a hierarchy of candidates, specific to certain knowledge bases, we can also document the specific KB + the specific Candidate together on their own page.

website/docs/api/kb.md
```diff
@@ -4,169 +4,88 @@ teaser:
...
-information from the relevant KB entities, such as its frequency in text and
-possible aliases. Each entity in the knowledge base also has a pretrained entity
-vector of a fixed size.
+The `KnowledgeBase` object is an abstract class providing a method to generate
+[`Candidate`](/api/candidate) objects, which are plausible external identifiers
+given a certain textual mention. Each such `Candidate` holds information from
+the relevant KB entities, such as its frequency in text and possible aliases.
+Each entity in the knowledge base also has a pretrained entity vector of a fixed
+size.
+
```
svlandeg, line 17: We should have an infobox here that explains that this same class was not abstract up until spaCy version X

website/docs/api/kb.md
```diff
@@ -4,169 +4,88 @@ teaser:
...
-> kb = KnowledgeBase(vocab=vocab, entity_vector_length=64)
+> kb = FullyImplementedKB(vocab=vocab, entity_vector_length=64)
 > ```
 
 | Name                   | Description                                      |
 | ---------------------- | ------------------------------------------------ |
 | `vocab`                | The shared vocabulary. ~~Vocab~~                 |
 | `entity_vector_length` | Length of the fixed-size entity vectors. ~~int~~ |
 
-## KnowledgeBase.entity_vector_length {#entity_vector_length tag="property"}
```
svlandeg, line 35: This property is still available, no?

website/docs/api/kb.md
```diff
@@ -4,169 +4,88 @@ teaser:
...
+| ----------- | -------------------------------------------------------------------- |
+| `mention`   | The textual mention or alias. ~~Span~~                               |
+| **RETURNS** | An iterable of relevant `Candidate` objects. ~~Iterable[Candidate]~~ |
 
-| Name        | Description                                              |
-| ----------- | -------------------------------------------------------- |
-| **RETURNS** | The list of aliases in the knowledge base. ~~List[str]~~ |
+## KnowledgeBase.get_candidates_batch {#get_candidates_batch tag="method"}
 
-## KnowledgeBase.get_alias_candidates {#get_alias_candidates tag="method"}
```
svlandeg, line 155: `kb#get_alias_candidates` is referred to from the `v3.md` file, so we probably want to keep the section and refer to the update of functionality & explain from which version it happened.

website/docs/api/kb_in_memory.md
```diff
@@ -0,0 +1,302 @@
+---
+title: InMemoryLookupKB
+teaser:
+  An implementation of the KnowledgeBase interface. Stores all information
+  in-memory.
+tag: class
+source: spacy/kb/kb_in_memory.pyx
+new: 3.4
```
svlandeg, line 8: It looks like this will be 3.5 instead

website/docs/api/kb_in_memory.md
```diff
@@ -0,0 +1,302 @@
+---
+title: InMemoryLookupKB
+teaser:
+  An implementation of the KnowledgeBase interface. Stores all information
```
svlandeg, line 4: I would probably say "The default implementation"

svlandeg (commented):
Had some minor comments about the docs.

spacy/pipeline/legacy/entity_linker.py
rmitsch: Was a refactoring mistake. Changed back to `KnowledgeBase`.

spacy/pipeline/legacy/entity_linker.py
rmitsch: Was a refactoring mistake. Changed back to `KnowledgeBase`.

spacy/kb/kb.pyx
rmitsch: We could drop the check, but `nel_emerson` - and anything else instantiating `KnowledgeBase` directly - will break anyway at a later point due to non-implemented functions.

website/docs/api/candidate.md
rmitsch: Added back in.

website/docs/api/kb.md
rmitsch: True, deleted it by mistake. Added back in.

website/docs/api/kb.md
rmitsch: Added a short disclaimer. Feel free to resolve this thread if you think it's sufficient.

website/docs/api/kb.md
rmitsch: Added a note, let me know what you think. One concern: it might look like this method is still available, since it's in the docs. I added an infobox, but still. I'm ok with leaving it in until v4 though.

spacy/kb/kb.pyx
svlandeg: Oh yea, I know. I wasn't suggesting to drop the check. Just mentioning this that we need to be aware that this is still a breaking change, that we want to move this to `develop` and ensure we update the impacted projects accordingly when this gets released.

rmitsch: I'll change the target branch to `develop` then?

svlandeg: Yep, please do. But before you do, you might want to make sure `develop` is sync'ed with the latest from `master` (if not, create a PR) because otherwise we'll likely see some git history mess ;p

spacy/kb/candidate.pyx
```diff
@@ -0,0 +1,74 @@
...
+from .kb cimport KnowledgeBase
+from ..tokens import Span
+
+cdef class Candidate:
+    """A `Candidate` object refers to a textual mention (`alias`) that may or may not be resolved
+    to a specific `entity` from a Knowledge Base. This will be used as input for the entity linking
+    algorithm which will disambiguate the various candidates to the correct one.
+    Each candidate (alias, entity) pair is assigned a certain prior probability.
+
+    DOCS: https://spacy.io/api/kb/#candidate_init
```
svlandeg, line 13:
```suggestion
    DOCS: https://spacy.io/api/kb/#candidate-init
```

spacy/pipeline/legacy/entity_linker.py
```diff
@@ -115,7 +114,7 @@ def initialize(
         get_examples (Callable[[], Iterable[Example]]): Function that
             returns a representative sample of gold-standard Example objects.
         nlp (Language): The current nlp object the component is part of.
-        kb_loader (Callable[[Vocab], KnowledgeBase]): A function that creates a KnowledgeBase from a Vocab instance.
+        kb_loader (Callable[[Vocab], KnowledgeBase]): A function that creates a InMemoryLookupKB from a Vocab instance.
```
svlandeg, line 117:
```suggestion
        kb_loader (Callable[[Vocab], KnowledgeBase]): A function that creates an InMemoryLookupKB from a Vocab instance.
```

spacy/pipeline/entity_linker.py
```diff
@@ -41,6 +41,14 @@
 DEFAULT_NEL_MODEL = Config().from_str(default_model_config)["model"]
 
 
+"""
+[initialize.components.entity_linker.kb_loader]
+@misc = "spacy.CustomKB.v1"
+entity_vector_length = 342
+custom_field = 666
+"""
```
svlandeg, line 49: I don't understand what this bit is doing here? It feels like a c/p from the test suite?

website/docs/api/candidate.md
svlandeg, line 2: Looks like we've now got this information duplicated in the separate file `candidate.md` as well as in `kb.md`

svlandeg: We just want to remove this one, right?

svlandeg (approved):
Nice work Raphael, and thanks for following through on the many reviewing comments & discussions so dilligently! Let's merge 🎉

PR #11276, opened 2022-08-08
Fix regex invalid escape sequences

adrianeboyd:
## Description

Fix regex invalid escape sequences

### Types of change

Kind of a bug?

## Checklist


adrianeboyd:
It's only a warning, but we should really try to catch these better. We could add W605 to the errors detected by `flake8` in the CI?

svlandeg:
You're right that we should have paid more attention. It does look like W605 will [become](https://www.flake8rules.com/rules/W605.html) a SyntaxError sometime in the future, but yea why not flag them more prominently as these will typically be bugs.

PR #11279, opened 2022-08-08
Update cupy extras

adrianeboyd:
## Description

* Extend to v11
* Add `cupy-cuda11x` and `cupy-wheel`
* Update quickstart to use `cupy-wheel` for CUDA 10.2+

## Checklist


adrianeboyd:
I'm not sure about the extra name `cuda-wheel` or the way the quickstart looks.

If you use `cupy-wheel` for all the existing CUDA versions in the quickstart (leaving them in the list), then it defaults to the first one (10.2), which works because all the commands are the same, but then it looks a bit weird as a default. But I think saying "10.2+" is going to confuse people. ???

adrianeboyd:
I don't really like `cuda-wheel` as the name, but we can't use plain `cuda` either because that's already used for source cupy. Maybe we can switch to maintaining the extras:

```
# keep cuda as source extra
cuda =
    cupy>=5.0.0b4,<12.0.0
# cupy as source extra
cupy =
    cupy>=5.0.0b4,<12.0.0
# cupy-wheel as wheel extra
cupy-wheel =
    cupy-wheel>=11.0.0,<12.0.0
```

We can keep the existing `cuda*` ones as is, but then just maintain / recommend these in the future?

website/src/widgets/quickstart-install.js
```diff
@@ -9,23 +9,15 @@ const DEFAULT_PLATFORM = 'x86'
...
-    '10.2': 'cuda102',
-    '11.0': 'cuda110',
-    '11.1': 'cuda111',
-    '11.2': 'cuda112',
-    '11.3': 'cuda113',
-    '11.4': 'cuda114',
-    '11.5': 'cuda115',
-    '11.6': 'cuda116',
-    '11.7': 'cuda117',
+    '10.2, CUDA 11.0+': 'cuda-wheel',
```
shadeMe, line 20:
```suggestion
    '10.2-11.x"': 'cuda-wheel',
```
So that `CUDA` doesn't get repeated in the displayed string?

shadeMe (approved):
Just one minor comment.

website/src/widgets/quickstart-install.js
adrianeboyd: Repeating it was intentional since the range 10.2-11.x is so weird and unexpected in the context of the rest of this list. I think a user is going to be looking for "CUDA 11.".

adrianeboyd:
Looking at it again, I'm not sure `cuda-wheel` is stable enough that we want to recommend it yet across the board. It's not used in the main cupy install docs yet.

website/src/widgets/quickstart-install.js
shadeMe: Can't we then just have an extra entry for CUDA 11.0+ that also maps to `cuda-wheel`?

polm:
For the record, this is the thread for status of `cupy-wheel`, which does seem to still be experimental:

https://github.com/cupy/cupy/issues/6688

It's not clearly stated, but it looks like it's supposed to work for CUDA 10.2 and not just 11+ - see the "supported packages" in the PR [here](https://github.com/cupy/cupy/pull/6012).

It looks like there are some environments (like Docker) where there may be issues with it still, though it seems to be working fine for normal use.

kmaehashi:
Hi, CuPy maintainer here.
Technical issues discussed in https://github.com/cupy/cupy/issues/6688 has been resolved in v11 and I believe it's ready to use by downstream packages. Currently, it lacks documentation but we are going to do that in https://github.com/cupy/cupy/issues/7005.

I'd suggest naming the extras_require as `cuda-autodetect`. The primary benefit of the `cupy-wheel` package is that it automatically finds the CUDA version installed in the users' environment and install the correct `cupy-cudaXXX` wheel. For users having multiple CUDA versions installed on the same machine, however, it could be ambiguous whether CUDA picked by `cupy-wheel` is the one user actually want to use (`cupy-wheel` always try to find the latest CUDA version installed), so I think it's better to keep providing `cuda11{01234567x}` extras as well to cover such case.

PR #11280, opened 2022-08-08
find-threshold: CLI command for multi-label classifier threshold tuning

rmitsch:
## Goal

Add a `find-threshold` CLI command investigating different threshold values for multi-label classification model and returning the one maximizing the F-score.

## Description

- New CLI command `find-threshold`; API call is `find_threshold()`.
- New tests in `spacy.tests.test_cli`.
- Docs will be added once the code has been reviewed.

Supported options are:
- `pipe_name`: Which pipe to evaluate (with pipelines with multiple `MultiLabel_TextCategorizer` components the name has to be specified, otherwise it's optional).
- `average`: Whether to use `micro` or `macro` to compute F-score over all labels.
- `n_trials`: Number of sample points in threshold space between 0 and 1.
- `beta`: Beta coefficient for F-score calculation.
### Types of change

New feature.

## Checklist


adrianeboyd:
I realize this is a draft, but some general concerns:

* there are multiple components with thresholds, can this be implemented more generally?
* I'd rather see the scoring code extended (e.g., for `beta` in `PRFScore` and `Scorer.score_cats`) than for this to be completely reimplemented here
* the args should look similar to `spacy evaluate` (note that it's a known problem in `spacy evaluate` that the `DocBin` is loaded generically rather than following `[corpora.dev]` from the config, which can cause issues sometimes)

The `textcat` component is a bit weird because the threshold only affects the scoring, so the annotation doesn't change with the threshold even though for most other components modifying the thresholds would affect the annotation.

rmitsch:
>     * there are multiple components with thresholds, can this be implemented more generally?

Should be possible. Would it be acceptable if we ditch the automatic component recognition then and always require naming the component to be evaluated?

>     * I'd rather see the scoring code extended (e.g., for `beta` in `PRFScore` and `Scorer.score_cats`) than for this to be completely reimplemented here

Wasn't aware of this. `Scorer.score_cats` should be a good fit.

>     * the args should look similar to `spacy evaluate` (note that it's a known problem in `spacy evaluate` that the `DocBin` is loaded generically rather than following `[corpora.dev]` from the config, which can cause issues sometimes)

I'll look into harmonizing the arguments.

> The `textcat` component is a bit weird because the threshold only affects the scoring, so the annotation doesn't change with the threshold even though for most other components modifying the thresholds would affect the annotation.

Can you elaborate on how modifying thresholds would affect annotations for other components?

adrianeboyd:
In general the situation is that you have a component that has:

* some config setting that's a threshold that's used in `set_annotations` or the scorer
* some score that's returned by its scorer that you want to maximize

Examples:

* `textcat_multilabel`: `textcat_multilabel.threshold` and the score `cats_macro_auc`
* `spancat`: `spancat.threshold` and the score `spans_sc_f`
* `span_finder`:`span_finder.threshold` and the score `span_finder_sc_f`

I would initially think that `beta` could be added as a scorer setting in updated versions of the default scorers so it can be customized in the configs. For the exploration here it might be useful if `beta` was easier to override than by modifying the configs, but I think I'd start from the point that there's an existing score in the scorer results that can be used directly and just look at the threshold here?

rmitsch:
Are there smart(-ish) ways to...
1. ...dynamically identify whether a component is suited for the find-threshold command?
2. ...map the correct scorer method to the component? Like `score_spans()` for `span_finder` (?) or `score_cats()` for `spancat` and `textcat_multilabel`?

We can hardcode this ofc, but I was wondering whether there's a better way to do this. 1. could be done by checking for the existence of a `threshold` attribute, but this is not quite consistent - e.g. `SpanFinder` has one, but `TextCategorizer` stores its `threshold` in `self.cfg` (I could check both, but this makes me question whether there are other variations).

adrianeboyd:
No, I think we need to rely on the user to provide a path to the threshold in the config and the scores key to optimize.

In v4, I'm planning to move all these settings out of `self.cfg` so they're only stored in the config, but this kind of threshold could have an arbitrary name in the config.

rmitsch:
> No, I think we need to rely on the user to provide a path to the threshold in the config and the scores key to optimize.

It's not just the scores key, I think. The scoring method in `Scorer` also has be specified or chosen if we want to this to be generic. E.g. `span_finder` wouldn't work with `Scorer.score_cats()`. Or am I misunderstanding something here?

adrianeboyd:
The component already has a registered scorer, so what I mean by "scores key" is the entry in the output of `Language.evaluate` that you want to optimize, like `cats_macro_f`. You don't care how/where this was calculated by the scorer, just that it ends up in `scores`.

rmitsch:
The latest commit should be closer to a generic solution. Two remarks:
- `beta` hasn't been introduced yet. I'd add it as optional argument to `PRFScore` to pass it forward from `Scorer`, if it's available in the latter's config. Are there any potential pitfalls to consider when doing this?
- The part of the test running a `spancat` component fails so far because `nlp.evaluate()` returns `None` for the relevant scores. It's probably related to how I set up the component, but I haven't spotted the issue yet - if there's anything obvious, I'd be thankful for a pointer.

spacy/tests/test_cli.py
```diff
@@ -855,3 +857,126 @@ def test_span_length_freq_dist_output_must_be_correct():
...
+def test_cli_find_threshold(capsys):
+    def make_examples(_nlp: Language) -> List[Example]:
+        docs: List[Example] = []
+
+        for t in [
+            (
+                "I'm angry and confused in the Bank of America.",
+                {
+                    "cats": {"ANGRY": 1.0, "CONFUSED": 1.0, "HAPPY": 0.0},
+                    "spans": {"sc": [(7, 10, "ORG")]},
```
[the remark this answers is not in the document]
rmitsch: :roll_eyes: Ah, right. Thanks, updated.

rmitsch:
Added a draft for integrating `beta`. Feedback much appreciated.

adrianeboyd:
I don't think beta makes sense as a `Scorer`-level setting but rather as an individual component scorer setting that would be set in the config, e.g.:

```
@registry.scorers("spacy.textcat_scorer.v1")
def make_textcat_scorer(beta: float = 1.0):
    return partial(textcat_score, beta=beta)
```

The existing textcat scorer is kind of a bad example because `threshold` should also already be a scorer setting but isn't.

For example, you could have two spancat components with different `beta` settings.

rmitsch:
Let me know if changes for `textcat_multilabel`/`score_cats()` and `spancat`/`score_spans()` match what you meant. If so, I'll update the other components and scoring functions.

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,173 @@
...
+    nlp = util.load_model(model)
+
+    pipe: Optional[Pipe] = None
+    try:
+        pipe = nlp.get_pipe(pipe_name)
+    except KeyError as err:
+        wasabi.msg.fail(title=str(err), exits=1)
+
+    if not isinstance(pipe, TrainablePipe):
+        raise TypeError(Errors.E1044)
```
adrianeboyd, line 121: I don't see any reason for this restriction?

adrianeboyd:
I meant that `beta` is not a parameter for the CLI command, but that `beta` is added as a scorer parameter in the `Scorer` methods and the component scorers. `beta` would already be set in the component config by the user based on their. Adding `beta` is probably better suited as a separate PR, though, and it would increment all the registered scorer versions.

For `find-threshold` you would still just care that there is some score in `scores`.

spacy/cli/find_threshold.py
rmitsch: `Pipe` doesn't necessarily have a `.cfg` attribute, which I assume to be available later on. I thought `mypy` complained about this otherwise, but apparently this isn't the case. Should I drop this?

adrianeboyd: In an ideal world, settings would only be changed in `nlp.config` to modify the thresholds.

In the current actual v3 world, I think you would have to try to set this in both `nlp.config` and `pipe.cfg`.

rmitsch: Setting `threshold` via `pipe.cfg` seems to work. What's the reason we'd do this in `nlp.config` as well?

adrianeboyd: It's not that it doesn't work, it's that in an ideal world these settings would all be only in `nlp.config`. It's a carry-over from v2 that they're stored in `pipe.cfg` and it was a workaround in the scorer refactoring in v3.1 that the `pipe.cfg` settings are passed internally to the scorer.

Actually, let me think a second, you might have to reload the model for every pass to get this work with `nlp.config`, but this is much cleaner and easier to specify for the user. But it is currently bad that the settings are duplicated in `nlp.config` and `pipe.cfg`, and then saved and reloaded from `pipe.cfg` instead of `nlp.config`, since they can get out-of-sync.

adrianeboyd: This is a better design (`pipe.cfg` is empty since everything is just in the config):

https://github.com/explosion/spacy-experimental/blob/7ca435d01aa87b2e61760a266fe8cdb748043ee1/spacy_experimental/span_finder/span_finder_component.py#L45-L144

rmitsch: What's the recommended way to update `nlp.config` programmatically? Took me a while to realize that `config` is rebuild in the getter. And `components` is rebuilt completely, so I'd have to change the content of `_pipe_configs`. I don't see how to though, other than deleting and adding the pipe.

adrianeboyd: I think it will have to involve reloading the pipeline for each iteration, and you'd modify it with overrides in `spacy.load("model", config={...overrides...})`.

adrianeboyd: I think this would be much easier if left until v4. If you do it now you'd have to do both the config and hack things into `pipe.cfg`.

rmitsch: Pipeline is reloaded with overrides for the threshold on every threshold change now.

> I don't see any reason for this restriction?
Coming back to this: this is for proper typing and type hinting. `nlp.get_pipe()` has a signature specifying the return of a `Pipe` object, yet we access its `.cfg` attribute to override the threshold setting. And I stand corrected, `mypy` actually issues an error about this if this type check is removed.

adrianeboyd: Then this should be refactored to work for both components without `.cfg` and non-`Pipe` components.

I think part of this is that the return type for `get_pipe` is incorrect. We don't actually require any particular type for pipeline components. I could also imagine non-trainable pipe components or other kinds of custom components that have thresholds.

rmitsch: What would be our approach be to changing the threshold value for components without `.cfg`? Try to manipulate `.threshold` directly and if that fails, abort?

adrianeboyd: Modifying it in `.cfg` is basically a bad hack. The case that should always work is modifying it in the config.

rmitsch: Not assuming `.cfg` or the component being a `Pipe` in the latest commit.

adrianeboyd:
Can you remove all the `beta`-related code from this PR (as a CLI arg and in the scorer) and also remove the `universe.json` in the tests that was added accidentally?

rmitsch:
@adrianeboyd Are we good with this or do we want another review round?

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,190 @@
...
+import numpy
+import wasabi.tables
+
+from ..errors import Errors
+from ..training import Corpus
+from ._util import app, Arg, Opt, import_code, setup_gpu
+from .. import util
+
+_DEFAULTS = {
+    "average": "micro",
```
adrianeboyd, line 16: This is unused now?

spacy/scorer.py
```diff
@@ -21,13 +21,7 @@
 class PRFScore:
     """A precision / recall / F score."""
 
-    def __init__(
```
adrianeboyd, line 24: I think you can revert all changes to this file?

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,190 @@
...
+            setattr(
+                nlp.get_pipe(pipe_name),
+                "cfg",
+                set_nested_item(getattr(pipe, "cfg"), config_keys, threshold),
+            )
+
+        scores[threshold] = nlp.evaluate(dev_dataset)[scores_key]
+        if not isinstance(scores[threshold], float) and not isinstance(
+            scores[threshold], int
+        ):
```
adrianeboyd, line 173:
```suggestion
        if not isinstance(scores[threshold], (float, int)):
```

spacy/tests/test_cli.py
```diff
@@ -855,3 +857,126 @@ def test_span_length_freq_dist_output_must_be_correct():
...
+        ]:
+            doc = _nlp.make_doc(t[0])
+            docs.append(Example.from_dict(doc, t[1]))
+
+        return docs
+
+    def init_nlp(
+        components: Tuple[Tuple[str, Dict[str, Any]], ...] = ()
+    ) -> Tuple[Language, List[Example]]:
+        _nlp = English()
```
adrianeboyd, line 892: I don't think there's any reason for private-looking names in this method with both `_nlp` and `_examples`:

```suggestion
        nlp = English()
```

spacy/tests/test_cli.py
```diff
@@ -855,3 +857,126 @@ def test_span_length_freq_dist_output_must_be_correct():
...
+        textcat_labels = ("ANGRY", "CONFUSED", "HAPPY")
+        for label in textcat_labels:
+            textcat.add_label(label)
+
+        # Append additional components to pipeline.
+        for cfn, comp_config in components:
+            comp = _nlp.add_pipe(cfn, config=comp_config)
+            if isinstance(comp, TextCategorizer):
+                for label in textcat_labels:
+                    comp.add_label(label)
```
adrianeboyd, line 907: I don't think you need to initialize the labels here if the component is initialized from the examples?

spacy/tests/test_cli.py
```diff
@@ -855,3 +857,126 @@ def test_span_length_freq_dist_output_must_be_correct():
...
+    ) -> Tuple[Language, List[Example]]:
+        _nlp = English()
+        textcat: TextCategorizer = _nlp.add_pipe(  # type: ignore
+            factory_name="textcat_multilabel",
+            name="tc_multi",
+            config={"threshold": 0.9},
+        )
+        textcat_labels = ("ANGRY", "CONFUSED", "HAPPY")
+        for label in textcat_labels:
+            textcat.add_label(label)
```
adrianeboyd, line 900: Same comment: can't this be initialized from the examples? I think that both `textcat` components just end up doing the same thing and the second one clobbers the first?

spacy/tests/test_cli.py
```diff
@@ -855,3 +857,126 @@ def test_span_length_freq_dist_output_must_be_correct():
     span_freqs = _get_spans_length_freq_dist(sample_span_lengths, threshold)
     assert sum(span_freqs.values()) >= threshold
     assert list(span_freqs.keys()) == [3, 1, 4, 5, 2]
+
+
+def test_cli_find_threshold(capsys):
+    thresholds = numpy.linspace(0, 1, 10)
+
+    def make_examples(_nlp: Language) -> List[Example]:
```
adrianeboyd, line 865:
```suggestion
    def make_examples(nlp: Language) -> List[Example]:
```

spacy/tests/test_cli.py
rmitsch: Not strictly, but it would shadow the variable names `nlp` and `examples` in the outer scope. Renamed it to `new_nlp` and `new_examples` - does that make more sense?

rmitsch:
@adrianeboyd Does the current state of this PR address your concerns?

adrianeboyd:
For `textcat` I think it's a bug in the scorer that the threshold is not set to `0.0` for `multi_label=False`. It doesn't make sense for there to be a threshold setting at all in the component, but this is probably a carry-over from before the defaults for exclusive classes were modified?

rmitsch:
>  From a UX perspective I think it would be better to display the table rows as available (so it doesn't look like it's hanging) and then the result at the end...

Done.

> I don't know whether a default of 11 or a default of +1 trials would be better?

I changed the default to 11 (agree it looks unexpected otherwise). I wouldn't run with a number of trials differing from what the user specified.

> There's also the UX question of what to do when the user provides a case where looking for a threshold doesn't make sense.

Hm, that's tricky. I don't think there's a way to detect that, right? Hard-coding some warning for specific use cases is possible, but would be be annoying to maintain. So maybe just emphasize in the docs that threshold optimization doesn't make sense in all scenarios and list examples?

rmitsch:
> For textcat I think it's a bug in the scorer that the threshold is not set to `0.0` for `multi_label=False`.

Do we want to address this in this PR?

adrianeboyd:
No, not in this PR. It should look the same as the `cats_macro_auc` with the same values for all thresholds. I'm trying to think about how you could detect the problematic cases without knowing anything about the scorer. You probably can't, but I think we can add more info/warnings for the built-in components like `textcat` about when this doesn't make sense.

rmitsch:
> You probably can't, but I think we can add more info/warnings for the built-in components like textcat about when this doesn't make sense.

So emitting a warning in those components directly whenever `threshold` has been set but wouldn't change anything about the outcome?

adrianeboyd:
I meant omitting a warning in `find-threshold` if you were doing something that didn't make sense, to avoid users trying to do this for cases where we know it's not sensible.

adrianeboyd:
Otherwise they're going to be scratching their heads for `textcat_multilabel`, like they're using `cats_score` as the provided reference score and they know the threshold can matter, but then the results are weird.

adrianeboyd:
I'm also not getting the results I'd expect for precision with a threshold of `1.0`, which looks like an issue with `PRFScore` as opposed to an individual scoring method:

```
Threshold    cats_macro_p
0.0          0.132     
0.25         0.666     
0.5          0.774     
0.75         0.838     
1.0          0.0

Threshold    cats_macro_r
0.0          1.0       
0.25         0.597     
0.5          0.477     
0.75         0.37      
1.0          0.0
```

I think the precision of nothing is 1.0 rather than 0.0, which is what you currently get...

adrianeboyd:
Let's see, to catch some common problems:

* check for `textcat` since the threshold doesn't make sense for exclusive classes
* check for all returned values being the same since this indicates something wrong, and then don't recommend 0.0 as a threshold for this case
  * for `textcat_multilabel` suggest using `cats_macro_f` or `cats_micro_f` instead

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+
+    pipe: Optional[Any] = None
+    try:
+        pipe = nlp.get_pipe(pipe_name)
+    except KeyError as err:
+        wasabi.msg.fail(title=str(err), exits=1)
+    if not hasattr(pipe, "scorer"):
+        raise AttributeError(Errors.E1045)
+
+    if isinstance(pipe, TextCategorizer):
```
adrianeboyd, line 115: `MultiLabel_TextCategorizer` will also be an instance of `TextCategorizer`. I think you could check the `type` instead?

spacy/cli/find_threshold.py
```diff
@@ -183,9 +191,23 @@ def filter_config(config: Dict[str, Any], keys: List[str]) -> Dict[str, Any]:
         )
 
     best_threshold = max(scores.keys(), key=(lambda key: scores[key]))
-    if not silent:
-        print(
-            f"\nBest threshold: {round(best_threshold, ndigits=4)} with value of {scores[best_threshold]}."
+
+    # If all scores are identical, emit warning.
+    if all([score == scores[thresholds[0]] for score in scores.values()]):
```
adrianeboyd, line 196: Very minor, but I think you could just check that there's only one unique value in `scores.values()` without referring to `thresholds`? (I'm trying to imagine cases where you have weird settings and `len(thresholds) == 0`.)

adrianeboyd:
Can you have a look at the conflicts?

rmitsch:
This is quite weird. Apparently my `master` had diverged from `explosion:master` (I guess due to the new approach to the release prep?). Synchronizing it eradicated all commits in this PR by an automatic force-pushed initiated by GitHub. I'm working on re-adding the changes to this branch.

rmitsch:
Should be fine now. Are we ok with this? Then I'd update the docs.

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+            else config[keys[0]]
+        }
+
+    # Evaluate with varying threshold values.
+    scores: Dict[float, float] = {}
+    config_keys_full = ["components", pipe_name, *config_keys]
+    table_col_widths = (10, 10)
+    thresholds = numpy.linspace(0, 1, n_trials)
+    print(wasabi.tables.row(["Threshold", f"{scores_key}"], widths=table_col_widths))
+    for threshold in thresholds:
```
kadarakos, line 162: I think linear search is probably good, but maybe we should try golden section search. I think that the `F1` is a unimodal function of the `threshold` right? There is a sweet spot and overshooting it or undershooting it makes the  results worse? If so then golden-section search can be used to find the best `threshold` without setting `n_trials` and setting a `tolerance` rather. 

Doesn't matter for now, I think it can be merged and I can try it after and see if it makes a difference, I think its just the semantics of the hyperparameter of the search is nicer.
https://en.wikipedia.org/wiki/Golden-section_search

rmitsch: Seems like an interesting idea. Agree that we don't need to do that know - I'd add an issue for this.

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+    def filter_config(config: Dict[str, Any], keys: List[str]) -> Dict[str, Any]:
+        """Filters provided config dictionary so that only the specified keys path remains.
+        config (Dict[str, Any]): Configuration dictionary.
+        keys (List[Any]): Path to value to set.
+        RETURNS (Dict[str, Any]): Filtered dictionary.
+        """
+        return {
+            keys[0]: filter_config(config[keys[0]], keys[1:])
+            if len(keys) > 1
+            else config[keys[0]]
```
svlandeg, line 153: This crashes ungracefully with a raw `KeyError` when the user specifies a non-existing `threshold_key`. Instead, let's catch it when this happens and provide the list of possible keys at the given level by printing `config.keys()` in a custom error message.

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+    "find-threshold",
+    context_settings={"allow_extra_args": False, "ignore_unknown_options": True},
+)
+def find_threshold_cli(
+    # fmt: off
+    model: str = Arg(..., help="Model name or path"),
+    data_path: Path = Arg(..., help="Location of binary evaluation data in .spacy format", exists=True),
+    pipe_name: str = Arg(..., help="Name of pipe to examine thresholds for"),
+    threshold_key: str = Arg(..., help="Key of threshold attribute in component's configuration"),
+    scores_key: str = Arg(..., help="Name of score to metric to optimize"),
```
svlandeg, line 33: "Name of score to metric" doesn't parse for me. Why not just "Metric name"?

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+            ),
+        )
+        if hasattr(pipe, "cfg"):
+            setattr(
+                nlp.get_pipe(pipe_name),
+                "cfg",
+                set_nested_item(getattr(pipe, "cfg"), config_keys, threshold),
+            )
+
+        scores[threshold] = nlp.evaluate(dev_dataset)[scores_key]
```
svlandeg, line 179: This also crashes ungracefully if the `scores_key` is rubbish. Again, we can print a custom error message with the availble keys from the result of the `nlp.evaluate`

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+                not isinstance(pipe, MultiLabel_TextCategorizer)
+                or scores_key in ("cats_macro_f", "cats_micro_f")
+            )
+            else "Use `cats_macro_f` or `cats_micro_f` when optimizing the threshold for `textcat_multilabel`.",
+        )
+
+    else:
+        if not silent:
+            print(
+                f"\nBest threshold: {round(best_threshold, ndigits=4)} with value of {scores[best_threshold]}."
```
svlandeg, line 210:
```suggestion
                f"\nBest threshold: {round(best_threshold, ndigits=4)} with {scores_key} value of {scores[best_threshold]}."
```

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+    data_path = util.ensure_path(data_path)
+    if not data_path.exists():
+        wasabi.msg.fail("Evaluation data not found", data_path, exits=1)
+    nlp = util.load_model(model)
+
+    pipe: Optional[Any] = None
+    try:
+        pipe = nlp.get_pipe(pipe_name)
+    except KeyError as err:
+        wasabi.msg.fail(title=str(err), exits=1)
```
svlandeg, line 111: We [typically](https://github.com/explosion/spaCy/blob/master/extra/DEVELOPER_DOCS/Code%20Conventions.md#dont-use-tryexcept-for-control-flow) avoid relying on `try`-`except` logic for control flow. In this case I think it makes the code more readible to simply check `nlp.component_names` up front and throw `E001` if need be.

svlandeg (commented):
This will be a useful CL tool to have, nice work!

I mainly had some comments around UX and documentation. It would be a good idea to provide some standard settings for this command 
(like `spacy find-threshold my_nlp data.spacy textcat_multilabel threshold cats_macro_f`) that users can just copy-paste if they're working with standard pipelines/configs.

spacy/tests/test_cli.py
svlandeg: @rmitsch : can you apply this suggestion + alter the var name in the body as well?

rmitsch:
> It would be a good idea to provide some standard settings for this command

I'll include some in the docs. Do you have any suggestions other than this one you'd like to have included?

svlandeg:
> > It would be a good idea to provide some standard settings for this command
> 
> I'll include some in the docs. Do you have any suggestions other than this one you'd like to have included?

It'd be nice to include one for each of the main pipeline components we see as relevant - currently mainly multilabel textcat & spancat, no?

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+    threshold_key: str,
+    scores_key: str,
+    *,
+    n_trials: int = _DEFAULTS["n_trials"],  # type: ignore
+    use_gpu: int = _DEFAULTS["use_gpu"],  # type: ignore
+    gold_preproc: bool = _DEFAULTS["gold_preproc"],  # type: ignore
+    silent: bool = True,
+) -> Tuple[float, float, Dict[float, float]]:
+    """
+    Runs prediction trials for `textcat` models with varying tresholds to maximize the specified metric.
```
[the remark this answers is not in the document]
rmitsch: Fixed in next commit.

spacy/cli/find_threshold.py
rmitsch: Yeah, fumbled that up. Fixed in next commit (to "Metric to optimize").

spacy/cli/find_threshold.py
rmitsch: Added in https://github.com/explosion/spaCy/pull/11280/commits/d080808bf72bb886f8eedb48faef57eaeabe13d5. Feel free to resolve if you think this has been addressed.

spacy/cli/find_threshold.py
rmitsch: Added in https://github.com/explosion/spaCy/pull/11280/commits/d080808bf72bb886f8eedb48faef57eaeabe13d5. Feel free to resolve if you think this has been addressed.

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,213 @@
...
+    pipe_name (str): Name of pipe to examine thresholds for.
+    threshold_key (str): Key of threshold attribute in component's configuration.
+    scores_key (str): Name of score to metric to optimize.
+    n_trials (int): Number of trials to determine optimal thresholds
+    code_path (Optional[Path]): Path to Python file with additional code (registered functions) to be imported.
+    use_gpu (int): GPU ID or -1 for CPU.
+    gold_preproc (bool): Whether to use gold preprocessing. Gold preprocessing helps the annotations align to the
+        tokenization, and may result in sequences of more consistent length. However, it may reduce runtime accuracy due
+        to train/test skew.
+    silent (bool): Display more information for debugging purposes
```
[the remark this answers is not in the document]
rmitsch: Updated in https://github.com/explosion/spaCy/pull/11280/commits/7b4da3f36d3650f6f35d56657fb6c71425ec3e8f. Feel free to resolve if you think this is sufficient.

spacy/cli/find_threshold.py
```diff
@@ -0,0 +1,233 @@
...
+                "cfg",
+                set_nested_item(getattr(pipe, "cfg"), config_keys, threshold),
+            )
+
+        eval_scores = nlp.evaluate(dev_dataset)
+        if scores_key not in eval_scores:
+            wasabi.msg.fail(
+                title=f"Failed to look up score `{scores_key}` in evaluation results.",
+                text=f"Make sure you specified the correct value for `scores_key`. The following scores are "
+                f"available: {eval_scores.keys()}",
```
svlandeg, line 195:
```suggestion
                f"available: {list(eval_scores.keys())}",
```

svlandeg (approved):
Looks good to me! I'll leave it open for one more day in case anyone else wanted to do a final review.

Issue #9588, opened 2021-11-02
RuntimeWarning: invalid value encountered in true_divide in EntityLinker


kinghuang:
I'm training and predicting using the EntityLinker in a pipeline for the first time. The component is outputting a `RuntimeWarning` when it's called.

```
/usr/local/lib/python3.9/site-packages/spacy/pipeline/entity_linker.py:358: RuntimeWarning: invalid value encountered in true_divide
  sims = xp.dot(entity_encodings, sentence_encoding_t) / (
```

This happens both during training and when predicting in a pipeline.

## How to reproduce the behaviour

Not sure yet. Will attempt to narrow down later.

## Info about spaCy

- **spaCy version:** 3.1.3
- **Platform:** Linux-5.10.47-linuxkit-x86_64-with-glibc2.31
- **Python version:** 3.9.7
- **Pipelines:** en_core_web_lg (3.1.0), en_enverus_oi_ner (0.0.0), en_enverus_oi_nlp (0.0.0)

polm:
Did you ever find a reproducible instance for this? I don't think we've seen it before, and without more details I don't think there's much we can do at present.

kinghuang:
Sorry, I haven't been working on this lately. I'll open a new issue if I come across this again!

PR #11284, opened 2022-08-09
Clean up automated label-based issue handling

polm:
## Description

Issues tagged with `more-info-needed` should be automatically closed after two weeks with no reply, but that doesn't seem to be happening - the most recent example I could find ([query](https://github.com/explosion/spaCy/search?o=desc&q=%22no+response+to+a+request%22&s=updated&type=issues)) is #9588 from March this year. The config file for that label refers to [a repo that has been archived](https://github.com/probot/no-response) in favor of another one. Since the behavior is basically the same as the `resolved` label, I just added the settings to the config for the action that does that.

The process that formerly managed this was a bot rather than a Github Action, and I'm not sure if there's some config outside the repo contents that needs to be changed related to that, or if it's already been disabled by the maintainer.

There are also a few other changes here, like upgrading the `issue-manager` version and changing the `more-info-needed` time to be seven days, like `resolved`. I think these are useful changes but this is also fine without them.

1. upgrade tiangolo/issue-manager to latest
2. move more-info-needed to tiangolo
3. change more-info-needed close time to 7 days
4. delete old more-info-needed config

### Types of change

chore?

## Checklist


polm:
It looks like the bot that did this previously was a "Github App", which was deactivated by the person running it at some point, though I can't find any announcement about this.

svlandeg (approved):
Thanks, looks good!

PR #11286, opened 2022-08-09
Rename modules for consistency

svlandeg:
## Description
The `entity_ruler` and `attribute_ruler` components remain unchanged in their naming, but this PR updates their Python modules to the same, consistent name. In `v4`, this will break imports from the old `entityruler.py` and `attributeruler.py` that need to be rewritten to import from `entity_ruler.py` and `attribute_ruler.py` instead.

This change makes the code base more consistent with `entity_linker`, `span_ruler` etc, and makes the naming of the Python modules consistent with the camel casing of the class names.

### Types of change
Python module refactoring

## Checklist


adrianeboyd:
This is pleasingly minimal.

PR #11292, opened 2022-08-10
Replace references to `thinc.backends.linalg` in the parser

shadeMe:
## Description

This PR replaces references to the vestigial `thinc.backend.linalg` module with `CBlas` calls and a local function. The aforementioned module will be removed once [this PR](https://github.com/explosion/thinc/pull/742) is merged.

### Types of change

Refactoring

## Checklist

### DIFF
--- a/spacy/ml/tb_framework.pyx
+++ b/spacy/ml/tb_framework.pyx
@@ -8,7 +8,6 @@ cimport numpy as np
 from thinc.api import Model, normal_init, chain, list2array, Linear
 from thinc.api import uniform_init, glorot_uniform_init, zero_init
 from thinc.api import NumpyOps
-from thinc.backends.linalg cimport Vec, VecVec
 from thinc.backends.cblas cimport CBlas, saxpy, sgemm
 from thinc.types import Floats1d, Floats2d, Floats3d, Floats4d
 from thinc.types import Ints1d, Ints2d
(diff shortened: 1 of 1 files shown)


shadeMe:
Quick note about the failing tests: They continue to fail locally even without my changes 🤔 I'll see if I can fix it.

spacy/ml/tb_framework.pyx
```diff
@@ -561,3 +559,15 @@ cdef void _sum_state_features(CBlas cblas, float* output,
                 feature = &cached[idx]
             saxpy(cblas)(O, one, <const float*>feature, 1, &output[b*O], 1)
         token_ids += F
+
+cdef inline int _arg_max(const float* scores, const int n_classes) nogil:
```
danieldk, line 563: Probably not for this PR, but we have other argmax implementations in `spacy/ml/parser_model.pyx`. It would be nice if they were somewhere together.

danieldk (approved):
Nice to have this replace by C BLAS calls. Did you run an evaluation to ensure that there is no regression parsing accuracy? (It all looks correct, but never bad to do a sanity check.)

spacy/ml/tb_framework.pyx
shadeMe: `spacy/ml/parser_model.pyx` is no longer part of the parser codebase post-refactor, right? I can't find it in this branch anymore.

shadeMe:
> Did you run an evaluation to ensure that there is no regression parsing accuracy? (It all looks correct, but never bad to do a sanity check.)

Re-tested - can confirm that there is no regression in parsing accuracy.

spacy/ml/tb_framework.pyx
danieldk: Ah, right, I was looking at the wrong branch. There is another one in `spacy/pipeline/_parser_internals/transition_system.pyx`.

PR #11312, opened 2022-08-15
Better handling of unexpected types in `SetPredicate`

shadeMe:
## Description

The `IN` and `NOT_IN` set predicates in `Matcher` would raise an unhandled exception when matching against attribute values that are non-hashable (like `list`s). This is now resolved by performing some defensive type checking of the attribute values prior to calling the Python `in` operator.

Going forward, if _any_ set predicate encounters a value of an unexpected type, it will fail early, i.e., return `False`, instead of crashing.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -885,10 +885,10 @@ class _SetPredicate:
                     value = set((get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value)))
                 else:
                     unexpected_value_type = True
-            elif isinstance(value, Iterable):
-                value = set(get_string_id(v) for v in value)
             elif isinstance(value, (str, int)):
                 value = set((get_string_id(value),))
+            elif isinstance(value, Iterable) and all(isinstance(v, (str, int)) for v in value):
+                value = set(get_string_id(v) for v in value)
             else:
                 unexpected_value_type = True
 
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -879,11 +879,8 @@ class _SetPredicate:
         elif self.predicate in ("IS_SUBSET", "IS_SUPERSET", "INTERSECTS"):
             # ensure that all values are enclosed in a set
             if self.attr == MORPH:
-                if isinstance(value, int):
-                    # break up MORPH into individual Feat=Val values
-                    value = set((get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value)))
-                else:
-                    return False
+                # break up MORPH into individual Feat=Val values
+                value = set(get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value))
             elif isinstance(value, (str, int)):
                 value = set((get_string_id(value),))
             elif isinstance(value, Iterable) and all(isinstance(v, (str, int)) for v in value):
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -1,5 +1,5 @@
 # cython: infer_types=True, cython: profile=True
-from typing import List
+from typing import List, Iterable
 
 from libcpp.vector cimport vector
 from libc.stdint cimport int32_t, int8_t
@@ -867,21 +867,34 @@ class _SetPredicate:
 
     def __call__(self, Token token):
         if self.is_extension:
-            value = get_string_id(token._.get(self.attr))
+            value = token._.get(self.attr)
         else:
             value = get_token_attr_for_matcher(token.c, self.attr)
 
-        if self.predicate in ("IS_SUBSET", "IS_SUPERSET", "INTERSECTS"):
-            if self.attr == MORPH:
-                # break up MORPH into individual Feat=Val values
-                value = set(get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value))
+        value_type_mismatch = False
+        if self.predicate in ("IN", "NOT_IN"):
+            if isinstance(value, (str, int)):
+                value = get_string_id(value)
             else:
-                # treat a single value as a list
-                if isinstance(value, (str, int)):
-                    value = set([get_string_id(value)])
+                value_type_mismatch = True
+        elif self.predicate in ("IS_SUBSET", "IS_SUPERSET", "INTERSECTS"):
+            # ensure that all values are enclosed in a set
+            if self.attr == MORPH:
+                if isinstance(value, int):
+                    # break up MORPH into individual Feat=Val values
+                    value = set((get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value)))
                 else:
-                    value = set(get_string_id(v) for v in value)
-        if self.predicate == "IN":
+                    value_type_mismatch = True
+            elif isinstance(value, Iterable):
+                value = set(get_string_id(v) for v in value)
+            elif isinstance(value, (str, int)):
+                value = set((get_string_id(value),))
+            else:
+                value_type_mismatch = True
+
+        if value_type_mismatch:
+            return self._emit_value_type_warning(value)
+        elif self.predicate == "IN":
             return value in self.value
         elif self.predicate == "NOT_IN":
             return value not in self.value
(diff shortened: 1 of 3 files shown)


adrianeboyd:
Thinking about this some more, I wonder if a warning is overkill and it would be better to just not match. If I remember correctly, it's somewhat difficult to filter this kind of warning that includes the values because `filterwarnings` treats each one as a separate warning message.

I do think it's definitely overkill to add a separate method just for the warning here.

shadeMe:
Ah, good catch - That's an oversight from an older iteration of the implementation where the warning needed to be emitted at multiple locations. I can inline the call.

> Thinking about this some more, I wonder if a warning is overkill and it would be better to just not match. If I remember correctly, it's somewhat difficult to filter this kind of warning that includes the values because filterwarnings treats each one as a separate warning message.
According to [the docs](https://docs.python.org/3/library/warnings.html#the-warnings-filter), it seems like the `message` parameter passed to the filter is a regular expression. This ought to allow for filtering based on the warning code, which would be the same for permutations of the warning string.

svlandeg:
> According to [the docs](https://docs.python.org/3/library/warnings.html#the-warnings-filter), it seems like the `message` parameter passed to the filter is a regular expression. This ought to allow for filtering based on the warning code, which would be the same for permutations of the warning string.

The regular expression is used by `warnings` to define the match, but if you filter with action `once` it will still show each and every single variation of that error code on the console once. (we battled with this [before](https://github.com/explosion/spaCy/pull/9438))

shadeMe:
> The regular expression is used by warnings to define the match, but if you filter with action once it will still show each and every single variation of that error code on the console once. (we battled with this https://github.com/explosion/spaCy/pull/9438)

Ah, I see. I'll just remove the warning then - seems like it's more trouble than it's worth.

spacy/matcher/matcher.pyx
```diff
@@ -867,21 +867,34 @@ class _SetPredicate:
...
+            if self.attr == MORPH:
+                if isinstance(value, int):
+                    # break up MORPH into individual Feat=Val values
+                    value = set((get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value)))
                 else:
-                    value = set(get_string_id(v) for v in value)
-        if self.predicate == "IN":
+                    unexpected_value_type = True
+            elif isinstance(value, Iterable):
+                value = set(get_string_id(v) for v in value)
```
adrianeboyd, line 889: Couldn't this `Iterable` contain something that you can't convert?

In general I think that this is getting too convoluted and we can shortcut to `False` much earlier when getting the value instead, something like:

```
        if self.is_extension:
            value = token._.get(self.attr)
            if isinstance(value, (str, int)):
                value = get_string_id(value)
            else:
                if not all(isinstance(v, (str, int)) for v in value):
                    return False
                value = set(get_string_id(v) for v in value)
```

and then add tests like this if there aren't any already:

```
matcher = Matcher(en_vocab)
Token.set_extension("ext", default=[])
pattern = [{"_": {"ext": {"INTERSECTS": ["A", "C"]}}}]
matcher.add("M", [pattern])
doc = Doc(en_vocab, words=["a", "b", "c"])
doc[0]._.ext = ["A", "B"]
assert len(matcher(doc)) == 1
```

and

```
matcher = Matcher(en_vocab)
Token.set_extension("ext", default=[])
pattern = [{"_": {"ext": {"INTERSECTS": ["A", "C"]}}}]
matcher.add("M", [pattern])
doc = Doc(en_vocab, words=["a", "b", "c"])
doc[0]._.ext = [["A"], "B"]
assert len(matcher(doc)) == 0
```

I think the tests can go into the existing operator tests?

And it's still possible there's some edge case for `value` vs. `set(values)` that I've missed. The main edge case I'd be concerned about is iterating over a string value into a set of individual chars by accident, but I think that's covered?

shadeMe: Hmm, the above suggestion won't work as we still need to gate the conversion of `Iterable` extension values for the IN and NOT IN predicates (otherwise the unhashable resultant set will raise an error downstream):
```
if self.is_extension:
            value = token._.get(self.attr)
            if isinstance(value, (str, int)):
                value = get_string_id(value)
            else:
                if not all(isinstance(v, (str, int)) for v in value):
                    return False
                value = set(get_string_id(v) for v in value) # This cannot be done for IN and NOT_IN
```
We could expand the `if not all...` statement to include a check for the above predicates, but that seems less readable to me than the current proposal, more so given that we check for the remaining predicates immediately afterwards to do more special handling of `MORPH` attributes. Also, if we add new attributes in the future whose values aren't `str` or `int`, we'll need to move the type-checking to the outer scope all the same.

> The main edge case I'd be concerned about is iterating over a string value into a set of individual chars by accident, but I think that's covered?

Yeah, that shouldn't happen as in the current implementation.

> Couldn't this Iterable contain something that you can't convert?

Good point. The `get_string_id` call will raise an error in that case, but we'll want to simply return `False` there too. I can expand the `if` statement to short-circuit on a failed convertibility check.

adrianeboyd: Ah, that makes sense. In general, I still find this a bit hard to follow, and I think it would be simpler to just `return False` rather than adding the variable `unexpected_value_type`?

And would it be possible to refactor how `self.value` is stored based on the operator so that there's no difference in the `value` conversion here? (Sets of `frozenset` or something like that?)

shadeMe: > Ah, that makes sense. In general, I still find this a bit hard to follow, and I think it would be simpler to just return False rather than adding the variable unexpected_value_type?

Sure.

> And would it be possible to refactor how self.value is stored based on the operator so that there's no difference in the value conversion here? (Sets of frozenset or something like that?)

I could be misunderstanding you, but do you mean flipping the conversion logic so that we do the `self.predicate` checks in the ctor instead (and use a single statement to normalize `value` in `__call__`)? 

I don't see how we can do that as we need to normalize the per-`Token` attribute values to the same form as those we're matching against. So, we'll need to keep the conversions in sync with each other. Also, the type of `value` is less constrained than that of `self.value`, which has to be an `Iterable` - this means we'll always need more logic to handle all the cases.

spacy/matcher/matcher.pyx
```diff
@@ -867,20 +867,30 @@ class _SetPredicate:
...
+                value = get_string_id(value)
             else:
-                # treat a single value as a list
-                if isinstance(value, (str, int)):
-                    value = set([get_string_id(value)])
+                return False
+        elif self.predicate in ("IS_SUBSET", "IS_SUPERSET", "INTERSECTS"):
+            # ensure that all values are enclosed in a set
+            if self.attr == MORPH:
+                if isinstance(value, int):
```
adrianeboyd, line 882: I don't think you'd ever have a non-`int`/`attr_t` value here for `MORPH`?

spacy/matcher/matcher.pyx
```diff
@@ -867,20 +867,30 @@ class _SetPredicate:
...
+                    # break up MORPH into individual Feat=Val values
+                    value = set((get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value)))
```
adrianeboyd, line 884:
```suggestion
                    value = set(get_string_id(v) for v in MorphAnalysis.from_id(self.vocab, value))
```

spacy/tests/matcher/test_matcher_api.py
```diff
@@ -476,14 +486,22 @@ def test_matcher_extension_set_membership(en_vocab):
...
 def test_matcher_extension_in_set_predicate(en_vocab):
     matcher = Matcher(en_vocab)
     Token.set_extension("ext", default=[])
     pattern = [{"_": {"ext": {"IN": ["A", "C"]}}}]
     matcher.add("M", [pattern])
     doc = Doc(en_vocab, words=["a", "b", "c"])
     doc[0]._.ext = ["A", "B"]
+
+    # The IN predicate expects an exact match between the
+    # extension value and one of the pattern's.
```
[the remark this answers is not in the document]
adrianeboyd: I still can't really parse this. Do you mean "between the extension value and one of the pattern's values"?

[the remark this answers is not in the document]
shadeMe: Correct, but you're right - it isn't very clear, and there's no reason to be so parsimonious with words in comments. I'll add it :sweat_smile:

PR #11317, opened 2022-08-16
Move `thinc.extra.search` to `spacy.pipeline._parser_internals`

shadeMe:
## Description

Moves the `thinc.extra.search` module and its related tests from thinc and fixes a null-pointer exception in `Beam.__dealloc__`. 

Additionally, this PR also adds support for executing unit tests written in Cython using the `pytest` framework (required for the moved test suite, which was actually never executed until now).

### Types of change

Refactoring, bug fix, new tests

## Checklist


davispuh:
Currently this change is only present in `v4` branch, but it might be useful to also merge it in `master` branch.

PR #11318, opened 2022-08-16
Remove default stop words

adrianeboyd:
## Description

Remove default stop words

Stop words are task-specific and attempting to maintain "general-purpose" stop word lists for many different languages is not feasible.

None of the underlying functionality has been modified, the only change is that the default stop word lists are empty. Users can provide their own stop words lists for their own tasks. 

(Admittedly it should be easier to modify the default stop words and it should be a setting that can be serialized with the model, but there are a number of complications related the language defaults and `lex_attr_getters` in general.)

### Types of change

I will personally take the viewpoint: enhancement.

## Checklist


svlandeg:
When we mention this in the release notes / blog post for v4, it would be nice to explain to users how to find the previous default lists (from a `3.x` branch on Github) so they can easily continue working with the same lists they had before.

adrianeboyd:
In terms of usability, I think one option would be to add a `before_creation` callback for setting stop words from a JSON list.

svlandeg (commented):
I found a few more spots in the docs that need updating too, they refer to the current English `stop_words.py`:
* website/docs/usage/101/_language-data.md
* website/docs/api/language.md

There's some references to setting these `stop_words` specifically when subclassing `Language`, not sure what to do with those, probably fine to keep as is?
* `stop_words = set(["custom", "stop"])` in website/docs/usage/linguistic-features.md
* Something similar for `SnekDdefaults` in website/docs/usage/saving-loading.md

Finally, should we add a comment in the docs somewhere to explain how to set custom stop words

svlandeg:
> In terms of usability, I think one option would be to add a `before_creation` callback for setting stop words from a JSON list.

Ah yes, that sounds good.

adrianeboyd:
I looked at all the examples in the docs for setting custom stop words and I think they should all continue to be fine.

Nothing prevents you from providing custom stop words with your custom language.

adrianeboyd:
The `before_creation` callback is kind of problematic because you don't want to reference any external data. But we can think about what to do...

svlandeg:
Could we have a default "stop word list reader" that users can plug in at the `before_init` callback?

adrianeboyd:
No, because the stop words are part of the defaults that aren't serialized with the pipelines. (Because it's effectively an `is_stop` method rather than a plain list/set.)

I'll have to look at it a bit more, but I've already tried to serialize them in various ways and it's basically a disaster, along with other various `lex_attr_getters` hacks.

adrianeboyd:
I would wait until the dust settles around proposals like #12244 before addressing conflicts in this PR.

svlandeg:

PR #11320, opened 2022-08-16
Replace EntityRuler with SpanRuler implementation

adrianeboyd:
## Description

Remove `EntityRuler` and rename the `SpanRuler`-based `future_entity_ruler` to `entity_ruler`.

Main changes:

* It is no longer possible to load patterns on init as with `EntityRuler(patterns=)`.
* The older serialization formats (`patterns.jsonl`) are no longer supported and the related tests are removed.
* The config settings are only stored in the config, not in the serialized component (in particular the `phrase_matcher_attr` and overwrite settings).

### Types of change

?

## Checklist


adrianeboyd:
I'm not really sure how to handle the docs. You could have `/api/entityruler` forward to `/api/spanruler`, but it's hard to cover both configs and options in a non-confusing on the span ruler page.

Instead I removed the `class` tag and source and added the warning at the top.

svlandeg:
> I'm not really sure how to handle the docs. You could have `/api/entityruler` forward to `/api/spanruler`, but it's hard to cover both configs and options in a non-confusing way on the span ruler page.
> 
> Instead I removed the `class` tag and source and added the warning at the top.

I was also going to suggest to just keep both pages separate, highlighting the reduced options on `entityruler` and then refer to `spanruler` to emphasize that that actually has more options/features.

website/docs/api/entityruler.md
```diff
@@ -1,13 +1,19 @@
...
 ---
 
+<Infobox title="New in v4" variant="warning">
+
+As of spaCy v4, there is no separate `EntityRuler` class. The entity ruler is
+implemented as a special case of the `SpanRuler` component. See the
+[`SpanRuler`](/api/spanruler) API docs for the full API.
+
+</Infobox>
+
```
svlandeg, line 16: Perhaps also add a note about `remove()` vs `remove_by_id()` ?

svlandeg (commented):
This is pretty satisfying :-)

adrianeboyd:
A remaining question is how to deal with references to `EntityRuler` (with the class name) in the docs since the class no longer exists.

svlandeg:
> A remaining question is how to deal with references to `EntityRuler` (with the class name) in the docs since the class no longer exists.

I did a quick search and it looks like pretty much all references on other pages in the docs ( `_architecture.md`, `processing-pipelines.md` etc) can be updated to refer to the new `SpanRuler` / `api/spanruler` instead. Did you have any specific problematic cases in mind?

svlandeg:
Adriane, is there anything I can do to help on this PR?

adrianeboyd:
I think it's confusing to switch the docs entirely to `SpanRuler` in these cases because you do have both `entity_ruler` and `span_ruler` and many people will be expecting to find docs about the entity ruler in general. I also don't think switching to "entity ruler" or `entity_ruler` makes sense because it doesn't line up with the rest of the docs, so in general I'm just still not really sure exactly what to do.

PR #11326, opened 2022-08-17
Refactor pipe(as_tuples) into a separate method

adrianeboyd:
## Description

Refactor pipe(as_tuples) into a separate method

### Types of change

?

## Checklist


adrianeboyd:
The main advantage here is that the types are a lot simpler. Otherwise it's not really much benefit for the code or for users.

spacy/errors.py
```diff
@@ -539,6 +539,10 @@ class Errors(metaclass=ErrorsWithCodes):
             "issue tracker: http://github.com/explosion/spaCy/issues")
     E202 = ("Unsupported {name} mode '{mode}'. Supported modes: {modes}.")
 
+    # New errors added in v4.x
+    E300 = ("nlp.pipe(text_tuples, as_tuples=True) has been replaced with:\n"
+            "nlp.pipe_as_tuples(text_tuples)")
+
```
shadeMe, line 545: I think that it'd be less confusing/more consistent to append new errors instead of inserting them inbetween existing ones. That way, we keep the general monotonic relationship b'ween the error numbers and spaCy builds.

adrianeboyd: I don't think the error codes need to be monotonic. Our original plan was to renumber all the v3 errors, but then we didn't, and having four-digit error codes when we never filled in the others is also annoying, just for formatting reasons here.

adrianeboyd: Or we do something more systematic than what happened with v3 starting at E900 for no particularly good reason, like starting at E4000 or something.

svlandeg: cf also https://github.com/explosion/spaCy/pull/11002/files

adrianeboyd: It's a little weird if it's not the same system. If we're already at 4 digits, then maybe W4000 and E4000?

svlandeg: That feels drastic, but if there's a 4 I think that's nice :-)

svlandeg (commented):
Otherwise looks good!

spacy/errors.py
polm: If we're already at four digits, I think starting at 4000 would make it very easy to understand.

svlandeg: Either way is fine by me - let's just pick something and leave a note at #11002 after this gets merged that that PR fixes the merge conflicts.

adrianeboyd:
On second thought, we should implement this first in parallel in v3, raise a deprecation warning for `as_tuples`, and then remove it in v4.

PR #11328, opened 2022-08-17
Make Span/Doc.ents more consistent for ent_kb_id and ent_id

adrianeboyd:
## Description

* Map `Span.id` to `Token.ent_id` in all cases when setting `Doc.ents`
* Reset `Token.ent_id` and `Token.ent_kb_id` when setting `Doc.ents`
* Make `Span.ent_id` an alias of `Span.id` rather than a read-only view
of the root token's `ent_id` annotation

### Types of change

?

## Checklist


adrianeboyd:
Since `Span` is no longer effectively read-only as of v3 (for easier use with `SpanGroup`), there were various inconsistencies in how the token<->span attributes are handled.

This PR makes the mapping for `Span` / `Token` attributes when you are working with entities as:

* `Span.label` <-> `Token.ent_type`
* `Span.kb_id` <-> `Token.ent_kb_id`
* `Span.id` <-> `Token.ent_id`

The span attributes are set from the token attributes at the point when you call `doc.ents`, so they do not reflect any changes to the token annotation after this point. When you set `doc.ents`, all the span properties override any token properties. When you reset entity annotation, all the related token properties are reset.

I think that keeping `Span.ent_id` as a view of the token annotation rather than having the same behavior as `Span.label` and `Span.kb_id` is confusing. Most uses I could find look like this:

```
for ent in doc.ents:
    this_ent_id = ent.ent_id
```

Where having this as an alias of `Span.id` would not cause problems. The v3 API has this as a read-only property, so there aren't cases where people are writing to `Span.ent_id` and would expect this to be separate from `Span.id`.

What would break is setting `Token.ent_id` and expecting this to be reflected in `Span.ent_id`.

spacy/tokens/span.pyx
```diff
@@ -802,28 +802,18 @@ cdef class Span:
...
-            cdef SpanC* span_c = self.span_c()
-            return span_c.id
+            return self.span_c().id
 
         def __set__(self, attr_t id):
-            cdef SpanC* span_c = self.span_c()
-            span_c.id = id
+            self.span_c().id = id
 
     property ent_id:
```
svlandeg, line 810: Should we mark `ent_id` as deprecated and refer to `span.id` instead so we can remove the former at some point in the future?

adrianeboyd: It's used in so much code that I am not sure it makes sense to deprecate it when it's not really a problem to maintain this.

polm: It feels pretty weird to have both of these around forever, but you're right that it's not hard to maintain right now and users would expect both of them. It's not great but it's probably better to have both of them for now.

svlandeg (approved):
I think this one is good to merge?

PR #11334, opened 2022-08-18
dev docs: numeric comparators

svlandeg:
## Description
After internal discussion, add a section on numeric comparisons.

### Types of change
dev docs update

## Checklist

### DIFF
--- a/extra/DEVELOPER_DOCS/Code Conventions.md
+++ b/extra/DEVELOPER_DOCS/Code Conventions.md
@@ -275,6 +275,24 @@ If you have to use `try`/`except`, make sure to only include what's **absolutely
 +     return [v.strip() for v in value.split(",")]
 ```
 
+### Numeric comparisons
+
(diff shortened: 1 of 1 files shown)


extra/DEVELOPER_DOCS/Code Conventions.md
```diff
@@ -275,6 +275,27 @@ If you have to use `try`/`except`, make sure to only include what's **absolutely
 +     return [v.strip() for v in value.split(",")]
 ```
 
+### Numeric comparisons
+
+For numeric comparisons, as a general rule we always use `<` and `>=` and avoid the usage of `<=` and `>`. This is to ensure we consistely
```
adrianeboyd, line 280:
```suggestion
For numeric comparisons, as a general rule we always use `<` and `>=` and avoid the usage of `<=` and `>`. This is to ensure we consistently
```

Issue #10954, opened 2022-06-13
Factory names with a dot can produce invalid thinc config


percevalw:
## Description

Currently, spacy defines the name of a component after its factory name when no name is specified. In the case of a factory whose name contains dots (like `org.component.v1`) a section containing dots is then added to the `nlp.config` object, making it impossible to export/import/interpolate the configuration since thinc expects no dot in section names. This, in turn, prevents calling spacy's training methods on such pipelines because of the operations performed on the `nlp.config`.

One solution could be to replace the dots with dashes in the factory name to provide a valid config, but I'm not entirely sure that this wouldn't affect existing pipelines.

## How to reproduce the behaviour

```python
factory_name = "custom.component.v1"


@Language.factory(factory_name)
class MyComponent:
    def __init__(self, nlp, name):
        self.nlp = nlp
        self.name = name

    def __call__(self, doc):
        pass


nlp = English()
nlp.add_pipe(factory_name)

# Export / import
config = nlp.config.to_str()
new_nlp = English.from_config(Config().from_str(config))
```

produces an error when calling `from_str` 

```
Error parsing config section. Perhaps a section name is wrong?
components -> custom -> component -> v1   Section 'custom' is not defined
```

## Your Environment

- **spaCy version:** 3.3.0
- **Platform:** macOS-10.16-x86_64-i386-64bit
- **Python version:** 3.9.7
- **Pipelines:**

svlandeg:
Just as a note, and as Adriane already mentioned in https://github.com/explosion/thinc/pull/721, we'll want to restrict factory names to not allow dots to prevent this from happening.

percevalw:
Hi, sorry for the late reply and thank you for your feedback. I understand that restricting factory names would be an effective, albeit a tad radical, way to prevent this problem.

However, I would note that the TOML specification [allows dots in section keys](https://toml.io/en/v1.0.0#keys:~:text=%5B-,dog.%22tater.man%22,-%5D%0Atype.name), so conforming to this specification (as in https://github.com/explosion/thinc/pull/721) would be a straightforward way to solve this issue, while allowing some freedom in naming the factories.

adrianeboyd:
Just a note that although it looks similar, the config format isn't TOML. Underneath it's using `configparser` with modifications.

PR #11336, opened 2022-08-18
Check for . in factory names

adrianeboyd:
## Description

Check for `.` in factory names

### Types of change

Bug fix.

## Checklist


polm (approved):

polm:
It's good to catch this. Do we want to check for other invalid characters here too? 

Also let me link in #10954.

adrianeboyd:
We really just split on `.` and I couldn't find any other invalid characters, at least not for the the config parsing itself.

svlandeg (approved):

PR #11337, opened 2022-08-18
Cleanup Cython structs

svlandeg:
Following up on https://github.com/explosion/spaCy/pull/9150

## Description
Removing some attributes that we were keeping around but are no longer necessary.

### Types of change
cleanup

## Checklist


adrianeboyd:
I was just looking a bit more at the symbols, and I think that renumbering the symbols has the potential to break saved `.spacy` files, so I think we're kind of stuck with the current symbol numbering. We can rename `IS_OOV_DEPRECATED` to `FLAG##`?

svlandeg:
You're right that it makes no sense to break existing `.spacy` files if we can help it - I'll update.

adrianeboyd:
Actually, we can't change any of the symbols without potentially breaking things, ugh. I think I'll try to add them to the next version of `.spacy` so there would be more freedom to modify them in the future without breaking saved files.

spacy/tokenizer.pxd
```diff
@@ -23,11 +24,7 @@ cdef class Tokenizer:
     cdef object _infix_finditer
     cdef object _rules
     cdef PhraseMatcher _special_matcher
-    # TODO convert to bool in v4
-    cdef int _faster_heuristics
-    # TODO next one is unused and should be removed in v4
-    # https://github.com/explosion/spaCy/pull/9150
-    cdef int _unused_int2
+    cdef bool _faster_heuristics
```
adrianeboyd, line 27:
```suggestion
    cdef bint _faster_heuristics
```

(and no import)

PR #11344, opened 2022-08-19
`StringStore` refactoring

shadeMe:
## Description

This PR introduces the following changes:

- Add type checks to ensure unexpected input raises a `TypeError`.
- Add `items()` method to iterate over hashes and strings in a pair-wise fashion. 
- Add type hints.
- Clean up vestigial code from the Python 2 era and remove support for `bytes` as inputs.
- Reorganize functions.
- Update usage docs and tests.

At the moment, this PR breaks a couple of `Matcher` tests which were passing unexpected input to the API ([related](https://github.com/explosion/spaCy/pull/11312)).

### Types of change

Refactoring

## Checklist


shadeMe:
The `Morphology` related change could/should probably be cherry-picked into its own PR (after this one gets merged). But we can leave it in if no one minds it.

spacy/strings.pyx
```diff
@@ -293,25 +174,121 @@ cdef class StringStore:
...
+
+cpdef hash_t hash_string(object string) except -1:
+    if not isinstance(string, str):
+        raise TypeError(Errors.E4000.format(expected_types="'str'", received_type=type(string)))
+
+    # Handle reserved symbols and empty strings correctly.
+    if len(string) == 0:
+        return 0
+
+    symbol = SYMBOLS_BY_STR.get(string, None)
```
adrianeboyd, line 262: You can leave out the `, None` because this is the default for `.get`.

spacy/lookups.py
```diff
@@ -6,7 +6,7 @@
 
 from .errors import Errors
 from .util import SimpleFrozenDict, ensure_path, registry, load_language_data
-from .strings import get_string_id
+from .strings import get_string_id  # type: ignore
```
adrianeboyd, line 9: Can you fix the types rather than ignoring this?

spacy/tests/vocab_vectors/test_stringstore.py
```diff
@@ -31,7 +31,7 @@ def test_stringstore_from_api_docs(stringstore):
     assert stringstore["banana"] == banana_hash
 
 
-@pytest.mark.parametrize("text1,text2,text3", [(b"Hello", b"goodbye", b"hello")])
+@pytest.mark.parametrize("text1,text2,text3", [("Hello", "goodbye", "hello")])
```
adrianeboyd, line 34: Does it make sense to remove these kinds of tests, or check for errors instead?

adrianeboyd (commented):
Just a few initial comments, I haven't gone through the `StringStore` code completely yet.

spacy/strings.pyi
```diff
@@ -1,21 +1,15 @@
...
-def get_string_id(key: Union[str, int]) -> int: ...
-
 class StringStore:
-    def __init__(
-        self, strings: Optional[Iterable[str]] = ..., freeze: bool = ...
-    ) -> None: ...
-    @overload
-    def __getitem__(self, string_or_id: Union[bytes, str]) -> int: ...
-    @overload
-    def __getitem__(self, string_or_id: int) -> str: ...
```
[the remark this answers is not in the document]
shadeMe: It's correct in that we still take `str` and `int` inputs, but I thought it would be better to simply have a single declaration with a `Union` as the type hint.

spacy/strings.pyx
```diff
@@ -14,236 +15,116 @@ from .symbols import NAMES as SYMBOLS_BY_INT
...
-
-    def as_string(self, key):
-        """If key is a string, return it; otherwise, get the string value."""
-        if isinstance(key, str):
-            return key
+        string (str, int): The string/hash to check.
+        RETURNS (bool): Whether the store contains the string.
+        """
+        cdef hash_t str_hash = get_string_id(string_or_hash)
+        if str_hash < len(SYMBOLS_BY_INT):
```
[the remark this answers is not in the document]
shadeMe: I see. Then, we should keep this PR open until the above is merged. I'll add the WIP label.

spacy/tests/vocab_vectors/test_stringstore.py
shadeMe: Good idea - I'll rewrite that test to check for errors instead.

spacy/lookups.py
shadeMe: My initial attempts to get that import to type check seemed to point to a `mypy` limitation (`get_string_id` is a `cpdef` function, which apparently isn't recognized by the former), but I'll rummage around a little more to see if there's another workaround.

adrianeboyd: I think you should just add it to `strings.pyi`.

spacy/strings.pyi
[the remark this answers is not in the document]
adrianeboyd: I think the original overloads are clearer / more correct and can probably make typing for downstream code easier.

spacy/strings.pyi
```diff
@@ -1,21 +1,42 @@
...
+    @overload
+    def as_string(self, string_or_hash: int) -> str: ...
     def add(self, string: str) -> int: ...
     def items(self) -> Tuple[int, str]: ...
     def __len__(self) -> int: ...
-    def __contains__(self, string_or_hash: Union[str, int]) -> bool: ...
+    @overload
+    def __contains__(self, string_or_hash: str) -> bool: ...
+    @overload
+    def __contains__(self, string_or_hash: int) -> bool: ...
```
adrianeboyd, line 24: This could stay a `Union`

spacy/strings.pyi
```diff
@@ -1,21 +1,42 @@
...
-    def to_disk(self, path: Union[str, Path]) -> None: ...
-    def from_disk(self, path: Union[str, Path]) -> StringStore: ...
+    @overload
+    def to_disk(self, path: str) -> None: ...
+    @overload
+    def to_disk(self, path: Path) -> None: ...
+    @overload
+    def from_disk(self, path: str) -> StringStore: ...
+    @overload
+    def from_disk(self, path: Path) -> StringStore: ...
```
adrianeboyd, line 34: this, too

spacy/strings.pyi
```diff
@@ -1,21 +1,42 @@
...
+    @overload
+    def __getitem__(self, string_or_hash: int) -> str: ...
+    @overload
+    def as_int(self, string_or_hash: str) -> int: ...
+    @overload
+    def as_int(self, string_or_hash: int) -> int: ...
+    @overload
+    def as_string(self, string_or_hash: str) -> str: ...
+    @overload
+    def as_string(self, string_or_hash: int) -> str: ...
```
adrianeboyd, line 17: this, too

spacy/strings.pyi
shadeMe: For future reference, what criteria do we use to select which functions need `@overload` annotations? Do we only wnant to overload when the return type is dependent on the input type?

adrianeboyd: It's if the input type determines the output type, not just if there's a `Union`.

adrianeboyd: There's no benefit to not using a `Union` above.

spacy/morphology.pyx
```diff
@@ -27,7 +28,7 @@ cdef class Morphology:
         self.strings = strings
 
     def __reduce__(self):
-        tags = set([self.get(self.strings[s]) for s in self.strings])
+        tags = set([self.get(key) for key, _ in self.strings.items()])
```
adrianeboyd, line 31: Both versions of this look incorrect in general? It's supposed to be a list of morphology tags not a list of strings, which is already in the pickled data as `self.strings`?

It can stay buggy here (temporarily), but I don't understand what this particular change is intended to accomplish as part of this PR?

spacy/strings.pxd
```diff
@@ -7,23 +7,20 @@ from murmurhash.mrmr cimport hash64
...
     unsigned char[8] s
     unsigned char* p
 
 
 cdef class StringStore:
     cdef Pool mem
-
     cdef vector[hash_t] keys
-    cdef public PreshMap _map
+    cdef PreshMap key_map
```
adrianeboyd, line 18: What do you mean by the name `key_map`? (It sounds like it's opposed to a non-key map or a keyless map?)

spacy/strings.pyx
```diff
@@ -14,236 +15,116 @@ from .symbols import NAMES as SYMBOLS_BY_INT
...
-        YIELDS (str): A string in the store.
+        string_or_hash (str, int): The hash value to lookup or a string.
+        RETURNS (str): The stored string or the input string.
+        """
+        if isinstance(string_or_hash, str):
+            return string_or_hash
+        else:
+            return self._get_interned_str(string_or_hash)
+
+    def items(self) -> Tuple[int, str]:
```
adrianeboyd, line 117: Looking at the odd `Morphology` example above, do we maybe want to reverse this so that `for key in strings` and `for key, value in strings.items()` aren't swapped in terms of what looks like the key and what looks like the value?

spacy/tests/morphology/test_morph_features.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 from spacy.morphology import Morphology
-from spacy.strings import StringStore, get_string_id
+from spacy.strings import StringStore, get_string_id  # type: ignore
```
adrianeboyd, line 3: Why do you need this ignore?

spacy/morphology.pyx
shadeMe: Oh, this one should probably go to another PR like I mentioned above; will remove it from this one.

The intention was to avoid calculating the hash of the mophological feature string (which is what the `self.strings[s]` call was doing), since it's already present in the `StringStore` instance.

As for picking `tags`, it does seem unnecessary - we can just iterate over the `StringStore`'s values and pass them to the `Morphology.add()` method.

spacy/strings.pxd
shadeMe: Don't quite remember why I went for that particular name, but it is indeed confusing. We could call it `values`, given that we store the keys in `keys`. Or we can rename `keys` to `hashes` and `_map` to `hash_to_string_map`.

spacy/strings.pyx
shadeMe: Yeah, fair point. Will swap the order.

spacy/tests/morphology/test_morph_features.py
shadeMe: Originally, it was due to the function being a `cpdef` declaration and `mypy` failing to detect it, but I believe we now include it in `strings.pyi` which fixes the issue. I'll remove the ignore hint.

spacy/morphology.pyx
adrianeboyd: Ah, I didn't enough at the details of `self.get`, which are weird.

shadeMe:
I'll reintroduce the `Morphology`-related changes (and the fix for `__reduce__`) in another PR.

spacy/strings.pxd
adrianeboyd: Naming: difficult.

Since it's basically private I don't really mind keeping `_map`.

Actually... do we even really need `keys` at all? Not the functionality, but storing it twice in this `vector`? Is is that much faster than iterating over `_map`?

shadeMe: I believe we keep it around to iterate over the strings in insertion order, which isn't guaranteed with the map.

adrianeboyd: Ah, hmm. So in general if you want to add `items()` I think it would make sense to also add `keys()` and `values()`, and then I might rename `keys -> _keys` and then just keep `_map`?

(My other naming ideas are currently all worse.)

shadeMe: Sounds good, will do.

shadeMe: It just occurred to me that returning the hashes with `keys()` and the strings with `values()` inconsistent with the semantics of `items()` since we return the inverted tuple with the latter.

Since we can't change `__iter__` without breaking code elsewhere, I think that the least confounding option would be to ensure that `keys()`, `values()` and `items()` behave like they would in a `dict`. What do you think?

adrianeboyd: My suggestion was not to return the hashes with `keys()` but to implement `keys()/values()/items()` in a way that's consistent with `__iter__`. This is basically the reason for calling it `_keys`.

We want the python API with `__iter__/keys()/values()/items()` to look like a dict where the strings are the keys and the hashes are the values. We should add some comments explaining how this is backwards from the internal representation/naming, but I think the main thing that matters is that it's consistent for the python API.

adrianeboyd: The types look incorrect for `keys/values/items` and I don't think we want them to be generators. The standard dict things are `MappingView` and I think we want whatever we return to be at least `Iterable` + `Sized`? I don't actually know if we want to use `KeysView`, etc., but I wouldn't expect a generator here.

spacy/strings.pyx
```diff
@@ -123,127 +32,120 @@ cdef class StringStore:
...
         else:
-            # TODO: Raise an error instead
-            return self._map.get(string_or_id) is not NULL
+            return self._get_interned_str(string_or_hash)
 
-        if str_hash in SYMBOLS_BY_INT:
-            return True
-        else:
-            return self._map.get(str_hash) is not NULL
+    def items(self) -> Collection[Tuple[str, int]]:
```
adrianeboyd, line 116: I don't think there's any reason to be vague about the return type here. If it's a list it should just say `List[Tuple...]`.

spacy/strings.pyx
```diff
@@ -123,127 +32,120 @@ cdef class StringStore:
...
+
+    def values(self) -> Collection[int]:
+        """Iterate over the stored strings hashes in insertion order.
+
+        RETURNS: An iterable collection of string hashs.
+        """
+        def as_hashes(string: str, str_hash: int) -> int:
+            return str_hash
+
+        return self._transform_to_list(as_hashes)
```
adrianeboyd, line 148: Is there any reason you can't just iterate over `_keys` here instead?

spacy/strings.pyx
```diff
@@ -291,27 +193,141 @@ cdef class StringStore:
             self.add(word)
         return self
 
+    def _transform_to_list(self, function: Callable[[str, int], Any]) -> Collection[Any]:
```
adrianeboyd, line 196: This seems a bit overengineered with the `Callable` for my taste? I think it's okay to just implement the individual `keys` and `values` methods without quite as much generalization?

spacy/strings.pyx
shadeMe: Good point, I overlooked that.

shadeMe:

spacy/strings.pyx
```diff
@@ -123,127 +32,126 @@ cdef class StringStore:
...
-        if isinstance(key, str):
-            return key
-        else:
-            return self[key]
+    def __iter__(self) -> Iterator[str]:
+        """Iterate over the strings in the store in insertion order.
+
+        RETURNS: An iterable collection of strings.
+        """
+        return iter(self.keys())
```
adrianeboyd, line 65: My only remaining concern is have `__iter__` be generating the full list upfront, but I also can't think of anywhere that this actually matters.

shadeMe: As far as I could tell (from the last time I checked), all calls to `__iter__` in the codebase are in a context where the entire `StringStore` needs to be traversed anyway.

spacy/strings.pyx
```diff
@@ -14,105 +15,13 @@ from .symbols import NAMES as SYMBOLS_BY_INT
...
-
 
 cdef class StringStore:
-    """Look up strings by 64-bit hashes.
+    """Look up strings by 64-bit hashes. Implicitly handles reserved symbols.
 
     DOCS: https://spacy.io/api/stringstore
     """
-    def __init__(self, strings=None, freeze=False):
+    def __init__(self, strings: Optional[Iterable[str]]=None):
```
svlandeg, line 24:
```suggestion
    def __init__(self, strings: Optional[Iterable[str]] = None):
```
manual `black` for Cython formatting :trollface:

spacy/strings.pyx
```diff
@@ -123,127 +32,126 @@ cdef class StringStore:
...
-        """Retrieve a string from a given hash, or vice versa.
+    def __getitem__(self, string_or_hash: Union[str, int]) -> Union[str, int]:
+        """Retrieve a string from a given hash. If a string
+        is passed as the input, add it to the store and return
+        its hash.
 
-        string_or_id (bytes, str or uint64): The value to encode.
-        Returns (str / uint64): The value to be retrieved.
+        string_or_hash (str, int): The hash value to lookup or the string to store.
+        RETURNS (str, int): The stored string or the hash of the newly added string.
```
svlandeg, line 41:
```suggestion
        string_or_hash (int / str): The hash value to lookup or the string to store.
        RETURNS (str / int): The stored string or the hash of the newly added string.
```
(keeping the order aligned with the docstring for hopefully a bit more readability)

spacy/strings.pyx
```diff
@@ -294,24 +202,122 @@ cdef class StringStore:
...
             self.add(string)
 
-    cdef const Utf8Str* intern_unicode(self, str py_string):
-        # 0 means missing, but we don't bother offsetting the index.
-        cdef bytes byte_string = py_string.encode("utf8")
-        return self._intern_utf8(byte_string, len(byte_string), NULL)
+    def _get_interned_str(self, hash_value: int) -> str:
+        cdef hash_t str_hash
+        if not _try_coerce_to_hash(hash_value, &str_hash):
+            raise TypeError(Errors.E4001.format(expected_types="'str','int'", received_type=type(hash_value)))
```
svlandeg, line 212: So the input `hash_value` is allowed to be a `str`?

spacy/strings.pyx
```diff
@@ -123,127 +32,126 @@ cdef class StringStore:
...
-        else:
-            return decode_Utf8Str(utf8str)
+    def __contains__(self, string_or_hash: Union[str, int]) -> bool:
+        """Check whether a string or a hash is in the store.
 
-    def as_int(self, key):
-        """If key is an int, return it; otherwise, get the int value."""
-        if not isinstance(key, str):
-            return key
+        string (str, int): The string/hash to check.
```
svlandeg, line 51: While reviewing this I started noticing that we have several ways currently in the codebase to document two possible types in these docstrings. We have `(a or b)`, `(a / b)` and `(a, b)`. It seems like this PR is unifying on the `or` variant, but the variant with `/` (eg `str / Path`) seems to be more prevalent throughout the current codebase, so perhaps we should standardize on that?

website/docs/api/stringstore.md
```diff
@@ -68,15 +69,14 @@ Check whether a string is in the store.
 > assert not "cherry" in stringstore
 > ```
 
-| Name        | Description                                     |
-| ----------- | ----------------------------------------------- |
-| `string`    | The string to check. ~~str~~                    |
-| **RETURNS** | Whether the store contains the string. ~~bool~~ |
+| Name        | Description                                             |
+| ----------- | ------------------------------------------------------- |
+| `string`    | The string or hash to check. ~~Union[str, int]~~        |
```
svlandeg, line 74:
```suggestion
| `string_or_hash`    | The string or hash to check. ~~Union[str, int]~~        |
```
(will need reformatting)

svlandeg (commented):
Looking good! I only had a few more minor nitpicks and then this is probably good to merge?

spacy/strings.pyx
shadeMe: Will do.

svlandeg (approved):
If Adriane is happy as well, this seems good to merge by me!

PR #11359, opened 2022-08-22
Enable fuzzy text matching in Matcher

kwhumphreys:
## Description

work in progress:
- use Levenshtein distance (polyleven implementation) to allow fuzzy matching of specific tokens in Matcher.
- FUZZY1, FUZZY2, ... FUZZY9 operators specify max string edit distance to allow a match.
- operators can apply to single tokens or sets, e.g. `{"LOWER": {"FUZZY3: "string"}}` or `{"LOWER": {"FUZZY2": {"IN": ["string1", "string2"]}}}`

### Types of change

enhancement

## Checklist


kwhumphreys:
any Cython advice on how to avoid `with gil` at https://github.com/explosion/spaCy/pull/11359/files#diff-63f1a25e13e4c412a4decf97e30bf2157f1a0d800efb1d0b25679e1991a8b804R697 ?

spacy/matcher/matcher.pyx
```diff
@@ -51,6 +53,7 @@ cdef class Matcher:
         self.vocab = vocab
         self.mem = Pool()
         self.validate = validate
+        self.fuzzy = fuzzy if fuzzy is not None else 0
```
maxbachmann, line 56: I am not really familiar with the way spacy usually handles normalization, but it is probably worth thinking about whether to use a range `0-100` or `0-1`.

adrianeboyd:
Just a few initial notes:

* I think there are a lot of users would like to have something like this, so this is a nice idea!
* we're pretty conservative about adding new dependencies to `spacy`, so we'll have to seriously consider whether we want to add `rapidfuzz`
* this feels like it ought to be an extra operator (similar to `REGEX`) rather than being a `Matcher` setting that's applied to a hard-coded subset of attributes

   ```python
   pattern = [{"ORTH": {"FUZZY": "word"}}]
   ```

   Then you could apply it to any attributes that you want including custom extensions. What's trickier with the operator is a parameter for fuzziness.

kwhumphreys:
Thanks @adrianeboyd. This plugin uses the notation you suggest https://spacy.io/universe/project/spaczz but in my evaluation it is much too slow to be practical, even with minimal use of the operator. Therefore I was trying for a more direct integration with the underlying rapidfuzz package.

Agreed that a user-specified list of attributes would be better than a hard-coded list, but I'd like to be able to specify this on Matcher init rather than on individual tokens. My use case is a large existing set of patterns which I'd like to optionally apply with fuzzy matching, without having to modify every pattern (but perhaps this could be done in a pattern pre-processing stage).

adrianeboyd:
Have you tried implementing `FUZZY` within spacy's `Matcher` similar to `REGEX`? I think that `spazz` has some of its own python-only matcher implementation, so it's not just the extra operator that's causing the difference.

If you want to get a general idea of the speed difference just for the operator, you can try comparing `{"ORTH": "x"}` to `{"ORTH": {"REGEX": "^the$"}}`.

setup.cfg
```diff
@@ -58,6 +58,8 @@ install_requires =
     requests>=2.13.0,<3.0.0
     pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0
     jinja2
+    rapidfuzz>=2.4.0,<3.0.0
+    rapidfuzz_capi>=1.0.5,<2.0.0
```
maxbachmann, line 62: `rapidfuzz_capi` needs to be placed in `setup_requires` similar to numpy when compiling against the C-API

adrianeboyd: (But note that while we maintain `setup_requires` here for historical/unclear reasons, it doesn't actually work in `setup.cfg` and the place that actually matters is `pyproject.toml`.)

maxbachmann: Oh I did not see the existence of the pyproject.toml file. Yes that is definetly the correct place. Then it is simply not required in setup.cfg

kwhumphreys: tests fail if setup.cfg is not identical to requirements.txt: https://github.com/explosion/spaCy/blob/master/spacy/tests/package/test_requirements.py#L66-L68

build instructions specify `pip install -r requirements.txt`, not `poetry install`: https://github.com/explosion/spaCy/blob/master/spacy/tests/package/test_requirements.py#L66-L68

maxbachmann: Ok. This should be added to the `pyproject.toml` anyways

kwhumphreys:
added a `FUZZY` attribute, as suggested (and as compatible with `spaczz`), but using a fixed fuzzy threshold value specified on init.

also added a parameter to specify a list of attributes on init, to be treated as fuzzy in all patterns, e.g. `Matcher(en_vocab, fuzzy=85, fuzzy_attrs=["ORTH", "LOWER"])`.

TODO: handle extension attributes
TODO: handle values in sets

pyproject.toml
```diff
@@ -8,5 +8,7 @@ requires = [
     "thinc>=8.1.0,<8.2.0",
     "pathy",
     "numpy>=1.15.0",
+    "rapidfuzz>=2.4.0,<3.0.0",
+    "rapidfuzz_capi>=1.0.5,<2.0.0",
```
maxbachmann, line 12: only the c-api is needed here

adrianeboyd:
We think this is a good idea and we discussed a bit internally how we'd like to support it:

* we would prefer not to add `rapidfuzz` as a dependency, but would instead prefer to directly include one edit distance / similarity implementation in spacy (possibly from `rapidfuzz`, or reimplemented ourselves if that makes more sense)
* we don't think that a top-level `fuzzy` setting for `Matcher` or `EntityRuler` makes sense because the appropriate fuzziness often depends so much on the exact strings / pattern / task and different patterns in the same `Matcher` would need different levels of fuzziness
* we're still not quite sure how to implement a fuzziness parameter for the operator; my initial clunky idea would be to support a small range of hard-coded operators like `FUZZY75`, `FUZZY80`, etc. but we're still thinking about the options/details here (ideas welcome!)
* it's possible there would be no way to support this in combination with set operators

maxbachmann:
> we would prefer not to add rapidfuzz as a dependency, but would instead prefer to directly include one edit distance / similarity implementation in spacy (possibly from rapidfuzz, or reimplemented ourselves if that makes more sense)

You can directly build against the C++ library https://github.com/maxbachmann/rapidfuzz-cpp.

kwhumphreys:
> * it's possible there would be no way to support this in combination with set operators

@adrianeboyd I just spent some time adding support for set operators. see https://github.com/explosion/spaCy/pull/11359/files#diff-d7740e9e8c9929c0b0e8e962c46a22e09263ef4891faceaa15415b2be94a5837R221-R263 but there probably isn't any way to use different fuzziness parameters for different set members.

kwhumphreys:
> * we're still not quite sure how to implement a fuzziness parameter for the operator; my initial clunky idea would be to support a small range of hard-coded operators like `FUZZY75`, `FUZZY80`, etc. but we're still thinking about the options/details here (ideas welcome!)

`FUZZY1, FUZZY2, ... FUZZYN` where `N` is the Levenshtein string edit distance might be simpler than a 0-100 scale?

adrianeboyd:
> `FUZZY1, FUZZY2, ... FUZZYN` where `N` is the allowed Levenshtein string edit distance might be simpler than a 0-100 scale?

I like that idea! It is definitely concrete and easy for users to understand.

kwhumphreys:
@adrianeboyd a small, fast, MIT-licensed implementation of Levenshtein distance here https://github.com/roy-ht/editdistance
Would this be acceptable to either import or copy into spaCy?

maxbachmann:
> a small, fast, MIT-licensed implementation of Levenshtein distance here https://github.com/roy-ht/editdistance
Would this be acceptable to either import or copy into spaCy?

What are the expected text lengths for the compared strings? editdistance is a pretty small implementation, but not very fast for longer sequences.
For short strings it is slower, but not by a dramatic margin:
Figure_1

For sequence lenghts > 640 characters editdistance becomes unusable:

Figure_2

RapidFuzz provides a single include file as well https://github.com/maxbachmann/rapidfuzz-cpp/blob/main/extras/rapidfuzz_amalgamated.hpp. It is a lot faster, but a lot more code as well (since it has more specialised implementations and supports more string metrics)

kwhumphreys:
> What are the expected text lengths for the compared strings?

comparisons in the current Matcher implementation are limited to single words. my use case is to apply sets of manually authored token patterns to inputs which may have spelling errors/variants, ideally with minimal performance impact. 

spaCy also has a PhraseMatcher which is typically used for things like lists of names, which can be multiword but rarely more than 2 or 3 words, but I haven't looking into fuzzy matching for this.

maxbachmann:
My general recommendation is:
- for short sequences (< 64 chars) rapidfuzz / polyleven
- mid long sequences rapidfuzz
- very long strings (> 50k characters) rapidfuzz / edlib

I am not aware of any case in which editdistance performs well. 

Since you have mainly short strings + want a simple implementation I would personally take the implementation from polyleven and just throw out all of the Cpython parts: https://github.com/fujimotos/polyleven/blob/master/polyleven.c which would be around 300 lines of code.

adrianeboyd:
We just merged a vendored version of `polyleven` into `master` (#11418), so if you merge in the changes from `master` you can do this without modifying the dependencies.

I think the basic `FUZZY` options look fine. I am not sure we want to add the changes related to set operators (we also don't do this for `REGEX`, which is similar).

kwhumphreys:
patterns using the positive set operator `{'IN': [set]}` can be split out into separate patterns for each set member to add FUZZY operators, but this can cause an explosion of patterns when expanding multiple sets.
however, patterns using the negative set operator `{'NOT_IN': [set]}` cannot be written as separate patterns with FUZZY operators because the intending matching logic applies within the set, not across patterns.

kwhumphreys:
the changes I made to set operators could be extended to also apply to `REGEX` if desired.
this would allow patterns like `{'LOWER': {'REGEX': {'IN': [regex1, regex2, regex3]}}}` which might be useful for pattern readability.

kwhumphreys:
another option to consider is a single FUZZY operator with a distance restriction implemented as a Matcher function which can be customized by users.
a default distance based on token length (`min(len(token), 5)`)has been working well for my purposes, and restrictions like `is_oov` could also be implemented there rather than being hardcoded.

kwhumphreys:
@adrianeboyd the simple FUZZY operator with a default distance limit based on token length implemented at https://github.com/explosion/spaCy/pull/11359/files#diff-63f1a25e13e4c412a4decf97e30bf2157f1a0d800efb1d0b25679e1991a8b804R208-R218 has been working well for me. do you have any suggestions on the best way to allow users to customize this function?

do you think that the FUZZYN operators are still also needed?

using fuzzy matching inside sets is essential for my uses. I have extended this for the regex operator for consistency.

adrianeboyd:
This is coming along nicely! Do you mind if I push directly to this branch?

kwhumphreys:
> Do you mind if I push directly to this branch?

of course not, please go ahead!

adrianeboyd:
I think for speed reasons it would be better to keep one sensible general-purpose `fuzzy_match` method in `matcher.pyx`. I haven't done any careful benchmarking, but I think that a `cpdef` function will be faster to call from within the matcher? (And I'm not really sure we want monkey patching as the recommended way to modify something like this?)

I also noticed that `fuzzy_match` doesn't return the expected results for short strings like  `"a"/"a"` or `"a"/"ab"` because it ends up with thresholds of `0` or below. I think that it might make sense to use a default value of `-1` for `distance` and then also pass `distance` to `levenshtein` so it can optimize its calculations? I think that `levenshtein` already checks the string lengths so you don't need to modify `distance` if it's provided?

kwhumphreys:
Thanks @adrianeboyd. I moved the function back, clarified the logic, and added some unit tests.

I had been experimenting with function overrides to add `is_oov` tests or to call a spellchecker and only attempt fuzzy matching on unrecognized words. Any suggestions on how to allow customization of the function without monkey patching?

adrianeboyd:
`fuzzy_match` as a kwarg is kind of clunky, but better than monkey-patching. The kwarg name and the internal API (especially for the predicates) is still not 100% decided.

adrianeboyd:
I originally thought it made sense to have a symmetric `fuzzy_match` method and we've discussed a number of more complicated options internally, but for the provided default I'd mainly like something that's relatively easy to explain to users in the docs. My current idea is something like this, what do you think?

```
cpdef bint _default_fuzzy_match(input_text: str, pattern_text: str, fuzzy: int):
    if fuzzy >= 0:
        max_edits = fuzzy
    else:
        # allow at least one edit and up to 20% of the pattern string length
        max_edits = ceil(0.2 * len(pattern_text))
    return levenshtein(input_text, pattern_text, max_edits) <= max_edits
```

You could have the exception for no edits to single-char strings, but I feel like it's not worth having a special case.

The reason for not having it be symmetric is primarily long tokens the text (usually URLs) vs. shorter tokens in the patterns.

The following test cases would change:

```
FAILED spacy/tests/matcher/test_levenshtein.py::test_default_fuzzy_match[a-ab-0-False] - AssertionError: assert True == False
FAILED spacy/tests/matcher/test_levenshtein.py::test_default_fuzzy_match[abc-cde-4-False] - AssertionError: assert True == False
FAILED spacy/tests/matcher/test_levenshtein.py::test_default_fuzzy_match[abcdef-cdefgh--1-True] - AssertionError: assert False == True
FAILED spacy/tests/matcher/test_levenshtein.py::test_default_fuzzy_match[abcdefgh-cdefghijk--1-True] - AssertionError: assert False == True
```

adrianeboyd:
I'm not sure I like `fuzzy_compare`, but I think using "fuzzy" and "match" for too many different things gets confusing. In effect it's more like `__eq__`, but I can't think of any names with `eq` that don't seem more confusing to me.

kwhumphreys:
Hi @adrianeboyd . Thanks for your updates. Sorry for not coordinating on the EntityRuler changes, but I needed to make changes to try out your proposed simpler algorithm in my application.

I have a few issues with the simpler algorithm, and I think it makes things less clear.

The explicit fuzzyN case still needs to be dynamically restricted by min token length, otherwise fuzzyN will allow completely different tokens if len(token) <= N, e.g. `fuzzy_compare('and', 'the', 3) == True`. Some token overlap should be required (possibly apart from single character tokens) and so max_edits always needs to consider token length.

And I think the default case (with implicit N) is now too restrictive. It doesn't allow transpositions if len(token) <= 5, so cases like `htis-this, iwth-with, htem-them` now all fail.

adrianeboyd:
I strongly think `FUZZY3` should be exactly 3 edits or it is way too confusing to explain to users.

The fact that the default case is too restrictive on short strings is a good point, that can definitely be adjusted.

kwhumphreys:
> I strongly think `FUZZY3` should be exactly 3 edits or it is way too confusing to explain to users.

even when 3 >= len(input_text)? allowing fuzzy_compare to match anything seems hard to explain and not very useful.

with sets, I have cases like `{'LOWER': {'FUZZY3': {'IN': ['wonderful', 'awesome', 'wow']}` and having N dynamically restrict for inputs <=N avoids an explosion of rules.

adrianeboyd:
> > I strongly think `FUZZY3` should be exactly 3 edits or it is way too confusing to explain to users.
> 
> even when 3 >= len(input_text)? allowing fuzzy_compare to match anything seems hard to explain and not very useful.

Yes, even when `len(input_text) <= 3`. I think "allows up to 3 edits" is much easier to explain than anything conditional on the lengths or requiring a minimum amount of overlap.

> with sets, I have cases like `{'LOWER': {'FUZZY3': {'IN': ['wonderful', 'awesome', 'wow']}` and having N dynamically restrict for inputs <=N avoids an explosion of rules.

What is the argument against using `FUZZY` instead of `FUZZY3` for cases like this?

kwhumphreys:
> What is the argument against using `FUZZY` instead of `FUZZY3` for cases like this?

ok, your more restrictive default case would allow FUZZY here. I'd used FUZZY3 because the original function was allowing too much on the longer words, but didn't want the 3 to apply to short words.

adrianeboyd:
From looking at examples for the docs, I think my proposed default of 20% (floor) is way too low. It means that it allows 2 edits for any words from 1-14 characters, which isn't much of a useful range.

adrianeboyd:
It also seemed kind of odd to just have 1..5, so I added 6..9, which gives you a total of ten buckets total to use for some very custom `fuzzy_compare` method.

kwhumphreys:
> From looking at examples for the docs, I think my proposed default of 20% (floor) is way too low. It means that it allows 2 edits for any words from 1-14 characters, which isn't much of a useful range.

My original function was trying to implement word length - 2 up to a maximum of 5. I don't think I've found any English word pairs where >5 edits is a useful match.

spacy/matcher/matcher.pyx
```diff
@@ -949,24 +951,24 @@ class _SetPredicate:
...
                 value = self.vocab.strings[value]
                 return any(bool(v.search(value)) for v in self.value)
-            elif value in self.value:
-                return True
             elif self.fuzzy is not None:
                 value = self.vocab.strings[value]
                 return any(self.fuzzy_compare(value, self.vocab.strings[v], self.fuzzy)
                            for v in self.value)
+            elif value in self.value:
+                return True
```
kwhumphreys, line 959: @adrianeboyd isn't checking for exact match before trying fuzzy match much cheaper?

adrianeboyd: It would be cheaper, but it kind of violates what we're saying that `FUZZY` does because it doesn't use the `fuzzy_compare` function.

I could imagine custom `fuzzy_compare` methods that explicitly don't match on exact matches or require a minimum edit distance? In the end it's more of a "custom comparison predicate that can be used for fuzzy matching". (I kind of wonder whether we should rename it and add the equivalent of `FUZZY0`. On the other hand I think it's a lot clearer for typical users if it's called `FUZZY`.)

kwhumphreys: ok, I hadn't considered any cases where `fuzzy_compare` would be used to block exact matches. using that in combination with `NOT_IN` and `OP: !` would get really confusing!

I'd viewed `fuzzy_compare` as a (more expensive) fallback when exact match fails.

adrianeboyd: I'm imagining that you can use `1-9` as a "mode" rather than a distance and have all kinds of weird custom matching.

adrianeboyd:
Having `spacy.matcher.matcher.fuzzy_compare` in the docs is kind of ugly, but I'm not sure we really want to intentionally export this higher. Unlike `levenshtein`, there's no particular reason for people to use this independently of the matcher.

spacy/tests/matcher/test_matcher_api.py
```diff
@@ -118,6 +118,144 @@ def test_matcher_match_multi(matcher):
...
+        matcher.add(key, patterns, greedy="LONGEST")
+
+    words = ["They", "like", "Goggle", "Noo"]
+    doc = Doc(matcher.vocab, words=words)
+    assert matcher(doc) == [
+        (doc.vocab.strings["GoogleNow"], 3, 4),
+    ]
+
+
+@pytest.mark.parametrize("fuzzyn", range(1, 6))
```
svlandeg, line 201:
```suggestion
@pytest.mark.parametrize("fuzzyn", range(1, 10))
```

svlandeg (commented):
This is looking great - thanks @kwhumphreys and @adrianeboyd for your work on this!

I had a few comments & suggestions, but overall it'll be great to have this in spaCy 3.5 🙂

spacy/tests/matcher/test_matcher_api.py
```diff
@@ -118,6 +118,144 @@ def test_matcher_match_multi(matcher):
...
+    rules = {
+        "GoogleNow": [[{"ORTH": {"FUZZY2": {set_op: ["Google", "Now"]}}, "OP": "+"}]]
+    }
+    matcher = Matcher(en_vocab)
+    for key, patterns in rules.items():
+        matcher.add(key, patterns, greedy="LONGEST")
+
+    words = ["They", "like", "Goggle", "Noo"]
+    doc = Doc(matcher.vocab, words=words)
+    assert len(matcher(doc)) == 1
```
svlandeg, line 234:
```suggestion
@pytest.mark.parametrize("greedy", ["FIRST", "LONGEST"])
@pytest.mark.parametrize("set_op", ["IN", "NOT_IN"])
def test_matcher_match_fuzzyn_set_op_longest(en_vocab, greedy, set_op):
    rules = {
        "GoogleNow": [[{"ORTH": {"FUZZY2": {set_op: ["Google", "Now"]}}, "OP": "+"}]]
    }
    matcher = Matcher(en_vocab)
    for key, patterns in rules.items():
        matcher.add(key, patterns, greedy=greedy)

    words = ["They", "like", "Goggle", "Noo"]
    doc = Doc(matcher.vocab, words=words)
    spans = matcher(doc, as_spans=True)
    assert len(spans) == 1
    if set_op == "IN":
        assert spans[0].text == "Goggle Noo"
    else:
        assert spans[0].text == "They like"
```

spacy/matcher/matcher.pyx
```diff
@@ -326,8 +347,8 @@ cdef class Matcher:
             return key
 
 
-def unpickle_matcher(vocab, patterns, callbacks):
-    matcher = Matcher(vocab)
+def unpickle_matcher(vocab, patterns, callbacks, validate, fuzzy_compare):
+    matcher = Matcher(vocab, validate=validate, fuzzy_compare=fuzzy_compare)
```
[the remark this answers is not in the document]
adrianeboyd: Now that it's renamed and moved, I don't think this is confusing anymore.

kwhumphreys:
@adrianeboyd note this usage, which I just came across today: https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness
it uses low/high token lengths, similarly to the min/max in the original fuzzy_compare function.

PR #11364, opened 2022-08-23
Fix menu order in linguistic-features.md

tzussman:
## Description
Swap 'Vectors & Similarity' and 'Mappings & Exceptions' in menu to match order in body. The original order leads to incorrect highlighting in the menu when scrolling.

### Types of change
Documentation

## Checklist


polm:
Thanks, nice catch!

PR #11365, opened 2022-08-24
Add spancat_singlelabel pipeline for multiclass and non-overlapping span labelling tasks

ljvmiranda921:
## Context

The current `spancat` implementation always treats the span labelling task as a multilabel problem. It uses the `Logistic` layer to output class probabilities independently for each class. However, when presented with a muticlass problem (exclusive class) it might be a disadvantage to not use the correct modeling assumptions. The `spancat-singlelabel` component uses the `Softmax` layer instead. The PR adds an additional inference parameter `allow_overlap: bool` to constrain the output of `spancat` when it is known apriori that the labeling is non-overlapping. The goal of these features is to make `spancat` more **precise** when it is known ahead of time that the task is multi-class and optionally non-overlapping.

## Description

This PR adds another pipeline, `spancat_singlelabel`, to account for exclusive classes in span categorization tasks. It does this **by introducing the concept of a "negative label" or "no label."** In `spancat`, the number of span labels is exactly the same as what's found in a dataset's annotation. Here, we add another column to account for the negative label. The presence of the negative label in the `SpanCategorizer` can be set with `add_negative_label: bool` so it can be to the multi-label `spancat` too. 

- We didn't touch the `add_label` implementation. It is the same as `spancat`'s. Instead, **we implemented two additional properties:** `_negative_label_i` (returns the index of the negative label) and `_n_labels` returns the length of `label_data` + 1 and changed `initialize` to create a `Softmax` layer with the extra negative label.

- This in turn affects how the annotations are created during inference (cf. `set_annotations`). We modified `_make_span_group` and renamed it to `_make_span_group_multilabel` to include `add_negative_label` and created `_make_span_group_singlelabel` tailored for multi-class inference including the `allow_overlap: bool` feature.

- When `add_negative_label` is `True` the training takes an optional argument `negative_weight` to be able to downweigh the importance of the negative spans in learning due how dominant they can be with naive suggesters. 

### ⏯️  Training: how is the loss computed this time? (also a note about the `negative_weight` param)

During initialization, we pass the number of labels (`n_labels + 1`) so that the score matrix has a shape `(n_samples, n_labels + 1)`, where `+1` accounts for the negative example. At train time, the score matrix should already be accounting for the negative example. In this implementation, the negative example is always at the **last** column.

**Figure:** Simple example using ConLL labels (ORG, MISC, PER, LOC)

image

In the `get_loss()` function, we then assign the value `1.0` to the last column whenever a particular span is a negative example. 
```python
# spancat.py::SpanCategorizer.get_loss()
target = self.model.ops.asarray(target, dtype="f")  # type: ignore
negative_samples = numpy.nonzero(negative_spans)[0]
target[negative_samples, self._negative_label] = 1.0
```

We then compute the scores and loss for backprop as usual (i.e., `d_scores = scores - target`). We also added an option of specifying a `negative_weight` to "control" the effect of the negative class (a form of class weighing). Higher values (>1) increases the effect of the negative class, while lower values minimizes it (<1).

### ⏯️ Inference: how are the annotations predicted? (also a note about the `allow_overlap` param) 

During inference, we remove the samples where the negative label *is* the prediction. In addition, if the `allow_overlap` parameter is set to `False`, then  overlapping spans are not stored (only the span with the highest predict probability). This is then tracked by the `Intervals` data structure. 

### ⏯️ Testing on other datasets  

I tested both `spancat` and `spancat_singlelabel` on some datasets, the spaCy project can be seen [here](https://github.com/ljvmiranda921/projects/pull/7) (I'm thinking that if this PR gets merged, I can make another PR to `explosion/projects` to add this benchmark). I ran them for three trials and reporting their mean and stdev:

| Dataset  | F-score (spancat) | F-score (spancat_singlelabel) |
|----------|-------------------|-----------------------------|
| AnEM     | 72.62 (0.22)      | **78.12 (0.43)**            |
| ConLL nl | 69.67 (0.69)      | **72.03 (0.33)**            |
| ConLL es | 72.46 (0.44)      | **74.10 (0.47)**            |
| WNUT 17  | 21.25 (0.97)      | **23.60 (3.88)**            |
| Archaeo  | 80.45 (0.18)      | **82.85 (0.25)**            |

## Types of change

- Feature implementation
- New pipeline for singlelabel spancat
- Tests and documentation

## Checklist

- [x]  I confirm that I have the right to submit this contribution under the project's MIT license.
- [x]  My changes don't require a change to the documentation, or if they do, I've added all required information.


spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,481 @@
...
+        spans_scores: Scores representing the model's predictions.
+        RETURNS (Tuple[float, float]): The loss and the gradient.
+
+        DOCS: https://spacy.io/api/spancategorizerexclusive#get_loss
+        """
+        spans, scores = spans_scores
+        spans = Ragged(
+            self.model.ops.to_numpy(spans.data), self.model.ops.to_numpy(spans.lengths)
+        )
+        label_map = {label: i for i, label in enumerate(self.labels)}
```
kadarakos, line 351: I know its not much work, but couldn't this `label_map` be  cached?

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,481 @@
...
+                    k = label_map[gold_span.label_]
+                    target[row, k] = 1.0
+                    # delete negative label target.
+                    negative_spans[row] = 0.0
+            # The target is a flat array for all docs. Track the position
+            # we're at within the flat array.
+            offset += spans.lengths[i]
+        target = self.model.ops.asarray(target, dtype="f")  # type: ignore
+        negative_samples = numpy.nonzero(negative_spans)[0]
+        breakpoint()
```
kadarakos, line 379:
```suggestion

```

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,481 @@
...
+from ..tokens import Doc, Span, SpanGroup
+from ..training import Example, validate_examples
+from ..util import registry
+from ..vocab import Vocab
+from .spancat import spancat_score, build_ngram_suggester
+from .trainable_pipe import TrainablePipe
+
+
+@registry.layers("spacy.Softmax.v1")
+def build_linear_logistic(nO=None, nI=None) -> Model[Floats2d, Floats2d]:
```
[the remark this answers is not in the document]
adrianeboyd: This remains unnecessary, I think? Just use the already registered thinc layer.

spacy/pipeline/spancat_exclusive.py
ljvmiranda921: Ok noted on this! I might still keep the breakpoint a bit for testing and debugging. Will keep in mind to remove this when the implementation's a bit final

spacy/pipeline/spancat_exclusive.py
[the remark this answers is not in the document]
ljvmiranda921: Will do that and use the thinc layer instead

spacy/pipeline/spancat_exclusive.py
ljvmiranda921: Just to clarify: do you mean creating this outside of the loop since this won't change often?

ljvmiranda921:
I think implementation-wise this PR can be reviewed. We'd definitely still want to run a few experiments comparing `exclusive_spancat` and `spancat` on a number of datasets. I'm not sure if I should do that one first, or perhaps you want to take a look on the current code before we run experiments. cc: @kadarakos @adrianeboyd

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,471 @@
...
+                    negative_spans[row] = 0.0
+            # The target is a flat array for all docs. Track the position
+            # we're at within the flat array.
+            offset += spans.lengths[i]
+        target = self.model.ops.asarray(target, dtype="f")  # type: ignore
+        negative_samples = numpy.nonzero(negative_spans)[0]
+        target[negative_samples, self._negative_label] = 1.0  # type: ignore
+        d_scores = scores - target
+        neg_weight = cast(float, self.cfg["negative_weight"])
+        d_scores[negative_samples] *= neg_weight
```
kadarakos, line 373:
```suggestion
        if neg_weight != 1.0:
            d_scores[negative_samples] *= neg_weight
```

kadarakos: To save time when the `negative_weight` is just the default `1.0`

svlandeg (commented):
This approach makes sense to me. I left some smaller reviewer comments for now.

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def __contains__(self, rang):
+        i, j = rang
+        for e in range(i, j):
+            if e in self.ranges:
+                return True
+        return False
+
+
+class SpanCategorizerExclusive(TrainablePipe):
```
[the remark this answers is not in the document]
ljvmiranda921: I think that makes sense. Will do that

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,472 @@
...
+    """Pipeline component to label spans of text.
+
+    DOCS: https://spacy.io/api/spancategorizerexclusive
```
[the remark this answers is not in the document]
ljvmiranda921: Ok. I will update the spancat docs instead

ljvmiranda921:
Hi! I've added a benchmark that tests `spancat_exclusive` and `spancat` on a number of NER datasets. Overall, the former seems to work well (across three trials, reporting their avg. and stdev). I'd like another round of review again to check in with the implementation. 🙇 

You can find the benchmarking project [here](https://github.com/ljvmiranda921/projects/pull/7). It's my fork of `explosion/projects`. Once `exclusive_spancat`'s been merged, I can also make another PR to `explosion/projects` to include this benchmark.

adrianeboyd:
Can you add tests for this similar to the existing `spancat` tests?

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,311 @@
...
+
+DEFAULT_SPANCAT_MODEL = Config().from_str(spancat_exclusive_default_config)["model"]
+
+
+@runtime_checkable
+class Suggester(Protocol):
+    def __call__(self, docs: Iterable[Doc], *, ops: Optional[Ops] = None) -> Ragged:
+        ...
+
+
```
adrianeboyd, line 53: This doesn't need to be defined again, does it?

```suggestion
```

ljvmiranda921:
Hi! I extended the tests from `spancat` to also include `spancat_exclusive`. I also updated the website documentation for `spancat` to mention additional parameters from `spancat_exclusive`.

spacy/pipeline/spancat.py
```diff
@@ -227,6 +227,19 @@ def key(self) -> str:
...
+        nO = None
+        # fmt: off
+        if self.model.has_dim("nO"):
+            nO = self.model.get_dim("nO")
+        elif self.model.has_ref("output_layer") and self.model.get_ref("output_layer").has_dim("nO"):
+            nO = self.model.get_ref("output_layer").get_dim("nO")
+        if nO is not None and nO == self._n_labels:
+            if not self.is_resizable:
+                raise ValueError(Errors.E922.format(name=self.name, nO=self.model.get_dim("nO")))
+        # fmt: on
```
adrianeboyd, line 241: I don't think it makes sense to turn off formatting for this?

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+[model]
+@architectures = "spacy.SpanCategorizer.v1"
+scorer = {"@layers": "Softmax.v2"}
+
+[model.scorer.init_W]
+@initializers = "zero_init.v1"
+
+[model.scorer.init_b]
+@initializers = "zero_init.v1"
+
```
adrianeboyd, line 26: This is already the default, so does it need to be in the default config?

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+[model.tok2vec.encode]
+@architectures = "spacy.MaxoutWindowEncoder.v2"
+width = ${model.tok2vec.embed.width}
+window_size = 1
+maxout_pieces = 3
+depth = 4
+"""
+
+DEFAULT_SPANCAT_MODEL = Config().from_str(spancat_exclusive_default_config)["model"]
```
adrianeboyd, line 48: I think it'll be less confusing if this has a unique name:

```suggestion
DEFAULT_EXCL_SPANCAT_MODEL = Config().from_str(spancat_exclusive_default_config)["model"]
```

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+        candidate span offsets. The model predicts a probability for each category
+        for each span.
+    spans_key (str): Key of the doc.spans dict to save the spans under. During
+        initialization and training, the component will look for spans on the
+        reference document under the same key.
+    scorer (Optional[Callable]): The scoring method. Defaults to
+        Scorer.score_spans for the Doc.spans[spans_key] with overlapping
+        spans allowed.
+    negative_weight (float): Multiplier for the loss terms.
+        Can be used to down weigh the negative samples if there are too many.
```
adrianeboyd, line 90:
```suggestion
        Can be used to downweight the negative samples if there are too many.
```

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+        name=name,
+        scorer=scorer,
+        allow_overlap=allow_overlap,
+    )
+
+
+@dataclass
+class Ranges:
+    """
+    Helper class help avoid storing overlapping span.
```
adrianeboyd, line 109:
```suggestion
    Helper class to avoid storing overlapping spans.
```

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+    name: str,
+    suggester: Suggester,
+    model: Model[Tuple[List[Doc], Ragged], Floats2d],
+    spans_key: str,
+    scorer: Optional[Callable],
+    negative_weight: float = 1.0,
+    allow_overlap: bool = True,
+) -> "Exclusive_SpanCategorizer":
+    """Create a SpanCategorizerExclusive component. The span categorizer consists of two
+    parts: a suggester function that proposes candidate spans, and a labeller
```
adrianeboyd, line 73:
```suggestion
    parts: a suggester function that proposes candidate spans, and a labeler
```

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+    def __contains__(self, rang):
+        i, j = rang
+        for e in range(i, j):
+            if e in self.ranges:
+                return True
+        return False
+
+
+class Exclusive_SpanCategorizer(SpanCategorizer):
+    """Pipeline component to label non-overlapping spans of text.
```
adrianeboyd, line 128: Is "non-overlapping" a correct description? (This doesn't make sense in combination with the `allow_overlap` option in particular.)

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+        labels = self.labels
+        indices, scores = indices_scores
+        offset = 0
+        for i, doc in enumerate(docs):
+            indices_i = indices[i].dataXd
+            doc.spans[self.key] = self._make_span_group(
+                doc,
+                indices_i,
+                scores[offset : offset + indices.lengths[i]],
+                labels,  # type: ignore[arg-type]
```
adrianeboyd, line 209: Check whether all the ignores are still necessary with the most recent `mypy`?

spacy/tests/pipeline/test_spancat.py
```diff
@@ -41,38 +41,42 @@ def make_examples(nlp, data=TRAIN_DATA):
     return train_examples
 
 
-def test_no_label():
+@pytest.mark.parametrize("name", ["spancat", "spancat_exclusive"])
```
adrianeboyd, line 44: I don't care a whole lot about this, but it might make sense to have `["spancat", "spancat_exclusive"]` as a global variable instead of repeating it?

website/docs/api/spancategorizer.md
```diff
@@ -52,6 +60,25 @@ architectures and their arguments and hyperparameters.
...
 
+
+> #### Example (spancat_exclusive)
+>
+> ```python
+> from spacy.pipeline.spancat import DEFAULT_SPANCAT_MODEL
+> config = {
+>     "threshold": 0.5,
+>     "spans_key": "labeled_spans",
+>     "max_positive": None,
```
adrianeboyd, line 71: `max_positive` isn't used here, is it?

website/docs/api/spancategorizer.md
```diff
@@ -60,17 +87,26 @@ architectures and their arguments and hyperparameters.
...
 | `max_positive` | Maximum number of labels to consider positive per span. Defaults to `None`, indicating no limit. ~~Optional[int]~~                                                                                                                                                                                      |
 | `scorer`       | The scoring method. Defaults to [`Scorer.score_spans`](/api/scorer#score_spans) for `Doc.spans[spans_key]` with overlapping spans allowed. ~~Optional[Callable]~~                                                                                                                                       |
+| `negative_weight`    | Multiplier for the loss terms. It can be used to down weigh the negative samples if there are too many. It is only available when using the `spancat_exclusive` component. ~~float~~                                                  |
```
adrianeboyd, line 90:
```suggestion
| `negative_weight`    | Multiplier for the loss terms. It can be used to downweight the negative samples if there are too many. It is only available for the `spancat_exclusive` component. ~~float~~                                                  |
```

website/docs/api/spancategorizer.md
```diff
@@ -60,17 +87,26 @@ architectures and their arguments and hyperparameters.
...
+| `allow_overlap`        | If `True`, the data is assumed to contain overlapping spans. It is only available when using the `spancat_exclusive` component. ~~bool~~ |
```
adrianeboyd, line 91:
```suggestion
| `allow_overlap`        | If `True`, the data is assumed to contain overlapping spans. It is only available for the `spancat_exclusive` component. ~~bool~~ |
```

website/docs/api/spancategorizer.md
```diff
@@ -60,17 +87,26 @@ architectures and their arguments and hyperparameters.
...
+%%GITHUB_SPACY/spacy/pipeline/spancat_exclusive.py
+```
+
 ## SpanCategorizer.\_\_init\_\_ {#init tag="method"}
 
 > #### Example
 >
 > ```python
 > # Construction via add_pipe with default model
+> # Use 'spancat_exclusive' for exclusive clases
```
adrianeboyd, line 109:
```suggestion
> # Use 'spancat_exclusive' for exclusive classes
```

(And I'm not sure users will understand what "for exclusive classes" means here.)

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+@layers = spacy.mean_max_reducer.v1
+hidden_size = 128
+
+[model.tok2vec]
+@architectures = "spacy.Tok2Vec.v2"
+[model.tok2vec.embed]
+@architectures = "spacy.MultiHashEmbed.v1"
+width = 96
+rows = [5000, 2000, 1000, 1000]
+attrs = ["ORTH", "PREFIX", "SUFFIX", "SHAPE"]
```
adrianeboyd, line 37:
```suggestion
rows = [5000, 1000, 2500, 1000]
attrs = ["NORM", "PREFIX", "SUFFIX", "SHAPE"]
```

Or possibly use `HashEmbedCNN` so that the defaults come from a central spot rather than specifying them multiple times in each component. (`HashEmbedCNN` could still be adjusted to have fewer `PREFIX` and `SHAPE` rows. It's still on the old defaults. But that would be a separate PR and involve incrementing its version number.)

adrianeboyd:

website/docs/api/spancategorizer.md
```diff
@@ -60,17 +87,26 @@ architectures and their arguments and hyperparameters.
 | `threshold`    | Minimum probability to consider a prediction positive. Spans with a positive prediction will be saved on the Doc. Defaults to `0.5`. ~~float~~                                                                                                                                                          |
 | `max_positive` | Maximum number of labels to consider positive per span. Defaults to `None`, indicating no limit. ~~Optional[int]~~                                                                                                                                                                                      |
```
adrianeboyd, line 88: This would also need a note that it's only for `spancat`.

spacy/pipeline/spancat_exclusive.py
ljvmiranda921: Updating to: "Pipeline component to label spans with exclusive classes"

spacy/pipeline/spancat_exclusive.py
```diff
@@ -18,6 +18,12 @@
 @architectures = "spacy.SpanCategorizer.v1"
 scorer = {"@layers": "Softmax.v2"}
 
+[model.scorer.init_W]
+@initializers = "zero_init.v1"
+
+[model.scorer.init_b]
+@initializers = "zero_init.v1"
+
```
ljvmiranda921, line 26: @adrianeboyd It seems that using `Softmax.v2` directly causes an error in the `@initializers`, hence I put back these lines of code. 

When I inspected the output of `self._config` in [this line of code](https://github.com/explosion/spaCy/blob/master/spacy/language.py#L165), the `init_W` and `init_b` keys have functions within them (e.g. `<function zero_init at 0x7f46ffa68d30>,`), not their string names.

adrianeboyd: Ah, now I remember some related discussions with users about this. Let me look at the details again...

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,308 @@
...
+    suggester: Suggester,
+    model: Model[Tuple[List[Doc], Ragged], Floats2d],
+    spans_key: str,
+    scorer: Optional[Callable],
+    negative_weight: float = 1.0,
+    allow_overlap: bool = True,
+) -> "Exclusive_SpanCategorizer":
+    """Create a SpanCategorizerExclusive component. The span categorizer consists of two
+    parts: a suggester function that proposes candidate spans, and a labeler
+    model that predicts a single label for each span.
```
adrianeboyd, line 74: "a maximum of one label for each span"?

ljvmiranda921: Sounds better, will update

spacy/pipeline/spancat_exclusive.py
adrianeboyd: This should be fixed by https://github.com/explosion/thinc/pull/808, but would (obviously) require a new release of thinc and updated thinc requirements in spacy.

ljvmiranda921: Ok sure, I'll remove the `init_*` parameters for now and wait for the new thinc release

svlandeg:
@ljvmiranda921 : could you have a look at the conflicts? 🙏

spacy/tests/pipeline/test_spancat.py
```diff
@@ -400,15 +407,15 @@ def mixed_zero_suggester(docs, *, ops=None):
     fix_random_seed(0)
     nlp = English()
     spancat = nlp.add_pipe(
-        "spancat",
+        name,
```
ljvmiranda921, line 410: Just noting that the conflict emerged here. Instead of using `test_zero_suggester`, it's now using the `test_mixed_zero_suggester`.

svlandeg: Looks like the test suite is failing now though?

adrianeboyd: It is waiting on the next release of thinc.

spacy/pipeline/spancat_exclusive.py
adrianeboyd: Let's see if this works...

adrianeboyd: This part works now, at least.

adrianeboyd:
It looks like it needs to updated to handle empty docs.

ljvmiranda921:
Ok will do that!

spacy/pipeline/spancat_exclusive.py
```diff
@@ -266,23 +266,26 @@ def _make_span_group(
...
-        if not allow_overlap:
-            # Get the probabilities
-            argmax_probs = numpy.take_along_axis(
-                scores[positive], numpy.expand_dims(predicted, 1), axis=1
-            )
-            argmax_probs = argmax_probs.squeeze()
-            sort_idx = (argmax_probs * -1).argsort()
-            predicted = predicted[sort_idx]
-            indices = indices[sort_idx]
+        if scores.size != 0:
```
ljvmiranda921, line 269: I'm not sure if this is the most elegant solution. The goal is that `_make_span_group` should return an empty `SpanGroup` when there zero suggestions. The problem happens when the `scores` variable is empty, and we cannot get all the other variables from that. I enclosed them in an if statement so that another branch happens.

kadarakos: maybe its more elegant (?) to have this on the very top:

```python
if scores.size == 0:
    return SpanGroup(doc, name=self.key)
scores = self.model.ops.to_numpy(scores)
indices = self.model.ops.to_numpy(indices)
...
```

ljvmiranda921: Ah that's nice! Ok will update in a few

website/docs/api/spancategorizer.md
```diff
@@ -52,25 +60,52 @@ architectures and their arguments and hyperparameters.
...
+%%GITHUB_SPACY/spacy/pipeline/spancat_exclusive.py
+```
+
 ## SpanCategorizer.\_\_init\_\_ {#init tag="method"}
 
 > #### Example
 >
 > ```python
 > # Construction via add_pipe with default model
+> # Use 'spancat_exclusive' for exclusive classes
```
adrianeboyd, line 108: At first glance this looks like the description is incorrect, maybe something like?

```suggestion
> # Replace 'spancat' with 'spancat_exclusive' for exclusive classes
```

ljvmiranda921:
Saw the updates and the merge conflict. I'll solve it first thing tomorrow!

ljvmiranda921:
Fixed the conflict!

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,306 @@
...
+    @property
+    def _n_labels(self) -> int:
+        """RETURNS (int): Number of labels including the negative label."""
+        return len(self.label_data) + 1
+
+    def set_annotations(self, docs: Iterable[Doc], indices_scores) -> None:
+        """Modify a batch of Doc objects, using pre-computed scores.
+
+        docs (Iterable[Doc]): The documents to modify.
+        scores: The scores to set, produced by SpanCategorizerExclusive.predict.
```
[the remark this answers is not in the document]
svlandeg: This also needs to be fixed in `spancat.py` and in the documentation, where it also (wrongly) reads `scores` (from before this PR)

website/docs/api/spancategorizer.mdx
```diff
@@ -52,14 +59,36 @@ architectures and their arguments and hyperparameters.
...
+| `max_positive` | Maximum number of labels to consider positive per span. Defaults to `None`, indicating no limit. It is only available for the `spancat` component. ~~Optional[int]~~                                                                                                                                                                                      |
 | `scorer`       | The scoring method. Defaults to [`Scorer.score_spans`](/api/scorer#score_spans) for `Doc.spans[spans_key]` with overlapping spans allowed. ~~Optional[Callable]~~                                                                                                                                       |
+| `negative_weight`    | Multiplier for the loss terms. It can be used to downweight the negative samples if there are too many. It is only available for the `spancat_exclusive` component. ~~float~~                                                  |
```
svlandeg, line 88: Add that it's `1.0` by default

website/docs/api/spancategorizer.mdx
```diff
@@ -52,14 +59,36 @@ architectures and their arguments and hyperparameters.
...
+| `allow_overlap`        | If `True`, the data is assumed to contain overlapping spans. It is only available for the `spancat_exclusive` component. ~~bool~~ |
```
svlandeg, line 89: Needs `prettier` formatting

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+@Language.factory(
+    "spancat_exclusive",
+    assigns=["doc.spans"],
+    default_config={
+        "spans_key": "sc",
+        "model": DEFAULT_EXCL_SPANCAT_MODEL,
+        "suggester": {"@misc": "spacy.ngram_suggester.v1", "sizes": [1, 2, 3]},
+        "scorer": {"@scorers": "spacy.spancat_scorer.v1"},
```
svlandeg, line 52: The default values for `negative_weight` and `allow_overlap` should be in the `default_config` here instead of in the function definition - cf. `make_spancat` in `spancat.py`.

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,306 @@
...
+    },
+    default_score_weights={"spans_sc_f": 1.0, "spans_sc_p": 0.0, "spans_sc_r": 0.0},
+)
+def make_spancat(
```
svlandeg, line 56: I would rename this to `make_excl_spancat` or `make_exclusive_spancat`

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,306 @@
...
+        vocab (Vocab): The shared vocabulary.
+        model (thinc.api.Model): The Thinc Model powering the pipeline component.
+        name (str): The component instance name, used to add entries to the
+            losses during training.
+        spans_key (str): Key of the Doc.spans dict to save the spans under.
+            During initialization and training, the component will look for
+            spans on the reference document under the same key. Defaults to
+            `"spans"`.
+        negative_weight (float): Multiplier for the loss terms.
+            Can be used to down weigh the negative samples if there are too many.
```
svlandeg, line 149:
```suggestion
            Can be used to down weigh the negative samples if there are too many.
            Defaults to 1.0.
```

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,306 @@
...
+        allow_overlap (bool): If True the data is assumed to
+            contain overlapping spans.
```
svlandeg, line 151:
```suggestion
        allow_overlap (bool): If True the data is assumed to
            contain overlapping spans. Defaults to `True`.
```

spacy/pipeline/spancat_exclusive.py
```diff
@@ -0,0 +1,306 @@
...
+    """Pipeline component to label spans with exclusive classes.
+
+    DOCS: https://spacy.io/api/spancatcategorizer
+    """
+
+    def __init__(
+        self,
+        vocab: Vocab,
+        model: Model[Tuple[List[Doc], Ragged], Floats2d],
+        suggester: Suggester,
```
svlandeg, line 131: This is not documented in the docstring (nor is it in `spancat.py`)

website/docs/api/spancategorizer.mdx
```diff
@@ -11,7 +11,14 @@ api_trainable: true
...
 A span categorizer consists of two parts: a [suggester function](#suggesters)
 that proposes candidate spans, which may or may not overlap, and a labeler model
-that predicts zero or more labels for each candidate.
+that predicts zero or more labels for each candidate. 
+
+This component comes in two forms: `spancat` and `spancat_exclusive`.  When you
+need to perform multi-label classification on your spans, use `spancat`.  The
+`spancat` component uses a `Logistic` layer where the output class probabilities
+are independent for each class.  However, if you need to predict exactly one true
+class for a span, then use `spancat_exclusive`. It uses a `Softmax` layer and treats
```
svlandeg, line 20: exactly one? Or "at most one"?

svlandeg (commented):
Nice work and nice results on the benchmark datasets! It would definitely be great to get that benchmark project into the `projects` repo so we can point users to it :-)

I had some smaller nitpicking comments, mostly about style & documentation.

website/docs/api/spancategorizer.mdx
svlandeg: And add that it's `True` by default

spacy/pipeline/spancat_exclusive.py
ljvmiranda921: I'll use `make_exclusive_spancat`

website/docs/api/spancategorizer.mdx
ljvmiranda921: I think "at most one" is more explicit. I'll update!

spacy/pipeline/spancat_exclusive.py
ljvmiranda921: I copied the docstring from `make_exclusive_spancat` into the `Exclusive_SpanCategorizer` class. 
Should I do the same for spancat in this PR?

ljvmiranda921:
Done! I saw the comments regarding the docstrings for `spancat.py`. I can include the changes here unless we want it in a different PR!

svlandeg:
> Done! I saw the comments regarding the docstrings for `spancat.py`. I can include the changes here unless we want it in a different PR!

These are pretty minimal fixes, I don't think you need to open a new PR for them, it's basically making sure things are consistent across the two components.

adrianeboyd:
A few things related to the overall config/design vs. `spancat`:

* this feels like it should be what happens for `max_positive = 1` in `spancat`
* wouldn't it make sense for `allow_overlap = false` to also be an option for `spancat`?

In general it feels like this functionality should be possible with options for one `SpanCategorizer` class and the only difference for the `spancat_exclusive` factory (or whatever it's called) is that there is a different model in the default config. (I'm not entirely sure, but I do think it's going to be tricky to do all this with one `spancat` config because you can't parameterize the default model.)

kadarakos:
I agree that `max_positive = 1` implies `spancat_exclusive` basically. Also agree that `allow_overlap` is not specific to `spancat_exclusive`. 

But the difference is not only the `Logistic` vs. `Softmax` in the output layer, but the `get_loss` is also different due to the negative class. Also, `spancat_exclusive` has the `negative_weight` parameter, whereas `spancat` has the `threshold`. So if the `SpanCategorizer` was implementing both it would have `threshold` which would be out of use for `max_positive = 1` and `negative_weights` should only be used for `max_positive = 1`. 

Should we merge the two implementations together into a single `SpanCategorizer` first and see how that looks like?

kadarakos:
The `spancat_exclusive.py` will be deleted if we agree on the design. I quite like it that there is one `SpanCategorizer` and two factories.

spacy/pipeline/spancat.py
```diff
@@ -154,6 +186,67 @@ def make_spancat(
     )
```
svlandeg, line 186: For the "old" factory, I would also fill in the "new" variables with their correct values so that it's robust to changing defaults in the constructor.

svlandeg: Basically avoid relying on the default values for both constructor calls.

kadarakos: ok that makes more sense!

spacy/pipeline/spancat.py
```diff
@@ -183,27 +302,52 @@ def __init__(
         vocab: Vocab,
         model: Model[Tuple[List[Doc], Ragged], Floats2d],
         suggester: Suggester,
+        # XXX Not sure what's the best default name when it can both be spancat
+        # and spancat_singlelabel
         name: str = "spancat",
         *,
+        single_label: bool = False,
```
adrianeboyd, line 309: I don't think you need a separate setting for `single_label`, it could just be `max_positive = 1`, right?

kadarakos: ok your right!

kadarakos: I just thought its nice to make it like very explicit, because it does something different when its `single_label`, but I'll change it.

kadarakos: Oh actually! I didn't do that because you might want to use a `Logistic` classifier with a `threshold` even when `max_positive=1` right? Or for simplicity we should not allow this and `max_positive=1` unambiguously means `Softmax` classifier?

kadarakos: If you use the `spancat` factory with `max_positive=1` then we have a `Logistic` classifier and no negative-label. If `single_label=True` we run the `Softmax` classifier with the negative label. So currently its not the same model. If `max_positive=1` should imply `Softmax` classifier, then when you use the `spancat` factory and say `max_positive=1` we should actually exit with an error?

adrianeboyd (commented):
Just a couple small notes, not a thorough review yet

spacy/pipeline/spancat.py
```diff
@@ -215,12 +357,24 @@ def __init__(
...
         self.scorer = scorer
+        self.add_negative_label = add_negative_label
+        if allow_overlap is None:
+            self.cfg["allow_overlap"] = True
+        elif not allow_overlap and max_positive is not None and max_positive > 1:
+            self.cfg["allow_overlap"] = True
+            msg.warn(
+                "'allow_overlap' can only be False when max_positive=1, "
+                f"but found 'max_positive': {max_positive} "
+                "SpanCategorizer is automatically configured with allow_overlap=True."
```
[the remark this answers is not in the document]
kadarakos: Should we `raise` and quit instead?

kadarakos:
> Just a couple small notes, not a thorough review yet

I still need to add a `make_span_group_singlelable` and `make_span_group_multilabel` test to see if they work in the `add_negative_label` case.

kadarakos:
Encountered a bit of problem. When I have the `add_negative_label = True` I need to access the `self._negative_label_index`. But `_make_span_group` takes `labels: List[str]` and so I don't know which one is supposed to be the negative label. In general, not sure why the method ended up like this originally like why allow the `_make_span_group` take any arbitrary list as labels that actually doesn't have to match the scores and indices coming from the model?

I can probably work around it maybe somehow, but I feel like what is already there is kinda clunky to begin with. I don't think we should use this arbitrary `List[str]` and it would be better to use the component's own `self.labels` or `self.label_map`? But maybe I'm missing some use case where this flexibility is nice? But if so, why not have `make_span_group` in utils that takes all the different kinds of arguments, but make sure that the `SpanCategorizer` calls it with something meaningful?

adrianeboyd:
I think it makes sense to refactor `_make_span_group` to remove the `labels` argument.

kadarakos:
> I think it makes sense to refactor `_make_span_group` to remove the `labels` argument.

Okay that's nice! I'm getting somewhere with it, but in the meantime I also realized that I do not necessarily understand the reason for having the `max_positive` argument actually in the `multilabel` case either. We already have the `threshold`. So why would you say this: "keep only the top-3 labels that pass the `threshold`"? Why not all labels that pass the `threshold`?

spacy/pipeline/spancat.py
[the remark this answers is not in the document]
adrianeboyd: Maybe an error would be better for now. We can always start with an error and if someone can make a good case for why they need it, then we can consider changing it.

Since this is in a component, it shouldn't be with `msg` but with python warnings or errors.

spacy/pipeline/spancat.py
```diff
@@ -304,14 +492,25 @@ def set_annotations(self, docs: Iterable[Doc], indices_scores) -> None:
...
-        labels = self.labels
         indices, scores = indices_scores
         offset = 0
         for i, doc in enumerate(docs):
             indices_i = indices[i].dataXd
-            doc.spans[self.key] = self._make_span_group(
-                doc, indices_i, scores[offset : offset + indices.lengths[i]], labels  # type: ignore[arg-type]
-            )
+            allow_overlap = self.cfg["allow_overlap"]
+            allow_overlap = cast(bool, self.cfg["allow_overlap"])
```
adrianeboyd, line 500:
```suggestion
            allow_overlap = cast(bool, self.cfg["allow_overlap"])
```

adrianeboyd (commented):
Just a few small code-related comments. I think the single `spancat` class is looking good. Can you go ahead and update all the related docs so we can do a final review?

spacy/pipeline/spancat.py
```diff
@@ -172,6 +271,27 @@ def make_spancat_scorer():
...
+
+    def add(self, i, j):
+        for e in range(i, j):
+            self.ranges.add(e)
+
+    def __contains__(self, rang):
+        i, j = rang
+        for e in range(i, j):
+            if e in self.ranges:
+                return True
```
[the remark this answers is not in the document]
kadarakos: I'm not exactly sure how exactly to do it with the `intersection`. I wrote a little script to do the speed tests, which includes a bit more coherent `Interval` API. 

```python
import time
import random

from dataclasses import dataclass
from typing import Tuple


class time_context:
    """Register the running time of a context."""

    def __enter__(self):
        self.start = time.perf_counter()
        return self

    def __exit__(self, type, value, traceback):
        self.elapsed = time.perf_counter() - self.start


def make_intervals(n: int, low: int, delta: int):
    out = []
    for i in range(n):
        low = random.randint(0, low)
        delta = random.randint(1, delta)
        out.append((low, low + delta))
    return out


@dataclass
class IntervalsLoop:

    def __init__(self):
        self.ranges = set()

    def add(self, interval: Tuple[int, int]):
        for e in range(interval[0], interval[1]):
            self.ranges.add(e)

    def __contains__(self, interval: Tuple[int, int]):
        for e in range(interval[0], interval[1]):
            if e in self.ranges:
                return True
        return False


@dataclass
class IntervalsIntersection:

    def __init__(self):
        self.ranges = set()

    def add(self, interval: Tuple[int, int]):
        for e in range(interval[0], interval[1]):
            self.ranges.add(e)

    def __contains__(self, interval: Tuple[int, int]):
        return bool(self.ranges.intersection(range(interval[0], interval[1])))


def speed_loop():
    intervals = make_intervals(100000, low=50, delta=4)
    seen = IntervalsLoop()
    with time_context() as elapsed:
        for i in range(100):
            for interval in intervals:
                if interval in seen:
                    continue
                else:
                    seen.add(interval)
    print(f"Loop: {elapsed.elapsed}")


def speed_intersection():
    intervals = make_intervals(100000, low=50, delta=4)
    seen = IntervalsIntersection()
    with time_context() as elapsed:
        for i in range(100):
            for interval in intervals:
                if interval in seen:
                    continue
                else:
                    seen.add(interval)
    print(f"Intersection: {elapsed.elapsed}")


speed_loop()
speed_intersection()
```

In this particular test the `intersection` implementation on my machine is slower:

```
Loop: 3.3513707518577576
Intersection: 4.87794885225594
```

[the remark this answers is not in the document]
adrianeboyd: Okay, I think this okay for now.

website/docs/api/spancategorizer.mdx
```diff
@@ -13,6 +13,13 @@ A span categorizer consists of two parts: a [suggester function](#suggesters)
 that proposes candidate spans, which may or may not overlap, and a labeler model
 that predicts zero or more labels for each candidate.
 
+This component comes in two forms: `spancat` and `spancat_singlelabel`. When you
+need to perform multi-label classification on your spans, use `spancat`. The
+`spancat` component uses a `Logistic` layer where the output class probabilities
+are independent for each class. However, if you need to predict at most one true
+class for a span, then use `spancat_singlelabel`. It uses a `Softmax` layer and
+treats the entities as a multi-class problem.
```
adrianeboyd, line 21:
```suggestion
treats the spans as a multi-class problem.
```

website/docs/api/spancategorizer.mdx
```diff
@@ -52,14 +59,33 @@ architectures and their arguments and hyperparameters.
...
-| `max_positive` | Maximum number of labels to consider positive per span. Defaults to `None`, indicating no limit. ~~Optional[int]~~                                                                                                                                                                                      |
-| `scorer`       | The scoring method. Defaults to [`Scorer.score_spans`](/api/scorer#score_spans) for `Doc.spans[spans_key]` with overlapping spans allowed. ~~Optional[Callable]~~                                                                                                                                       |
+> #### Example (spancat_singlelabel)
+>
+> ```python
+> from spacy.pipeline.spancat import DEFAULT_SPANCAT_SINGLELABEL_MODEL
+> config = {
+>     "threshold": 0.5,
+>     "spans_key": "labeled_spans",
+>     "model": DEFAULT_EXCL_SPANCAT_MODEL,
```
adrianeboyd, line 69: Is this correct given the import above?

spacy/pipeline/spancat.py
```diff
@@ -119,7 +149,9 @@ def make_spancat(
     threshold: float,
     max_positive: Optional[int],
 ) -> "SpanCategorizer":
-    """Create a SpanCategorizer component. The span categorizer consists of two
+    """Create a SpanCategorizer component and configure it for multilabel
```
adrianeboyd, line 152:
```suggestion
    """Create a SpanCategorizer component and configure it for multi-label
```

adrianeboyd (commented):
It looks like the existing usage is mostly "multi-label" and "multi-class" (e.g. for textcat). I think I'd personally prefer it without the dash, but mainly I'd like it to be consistent.

A number of docstring paragraphs need to be justified again, too. (I'm pretty sure this formatting can show up directly in people's editors?)

spacy/pipeline/spancat.py
```diff
@@ -183,27 +306,49 @@ def __init__(
         vocab: Vocab,
         model: Model[Tuple[List[Doc], Ragged], Floats2d],
         suggester: Suggester,
+        # TODO Not sure what's the best default name when it can both be spancat
+        # and spancat_singlelabel
```
adrianeboyd, line 310:
```suggestion
```

spacy/pipeline/spancat.py
```diff
@@ -183,27 +306,49 @@ def __init__(
...
         scorer: Optional[Callable] = spancat_score,
     ) -> None:
-        """Initialize the span categorizer.
+        """Initialize the multi-label or multi-class span categorizer.
+        The 'single_label' argument configures whether the component
+        should only produce one label per span (multi-class) or if it
+        can produce multiple labels per span (multi-label). In the
+        multi-label case the classification layer is expected to be
+        Logistic and Softmax in the multi-class case.
         vocab (Vocab): The shared vocabulary.
```
adrianeboyd, line 327:
```suggestion
        Logistic and Softmax in the multi-class case.

        vocab (Vocab): The shared vocabulary.
```

adrianeboyd:
I think the last thing for the docs is to figure out how to mark all the new things clearly as new.

kadarakos:
> I think the last thing for the docs is to figure out how to mark all the new things clearly as new.

How do we do that usually? We put this indication like from which version a feature is available right?

adrianeboyd:
Yes, there would be new tags from `3.5.1`. What I'm not sure is exactly how best to mark `spancat_singlelabel` on the combined API docs page.

kadarakos:
I added the `<Tag variant="new">3.5.1</Tag>` in in some places where I've felt like it communicates it more or less unambiguously what is new, but I wasn't exactly sure.

adrianeboyd:
The markers should go next to the setting name in the settings column rather than in the descriptions.

adrianeboyd:
Could you also revert all the formatting-only changes to the other `.mdx` files?

website/docs/api/spancategorizer.mdx
```diff
@@ -13,6 +13,13 @@ A span categorizer consists of two parts: a [suggester function](#suggesters)
 that proposes candidate spans, which may or may not overlap, and a labeler model
 that predicts zero or more labels for each candidate.
 
+This component comes in two forms: `spancat` and `spancat_singlelabel`. When you
+need to perform multi-label classification on your spans, use `spancat`. The
+`spancat` component uses a `Logistic` layer where the output class probabilities
+are independent for each class. However, if you need to predict at most one true
+class for a span, then use `spancat_singlelabel` <Tag variant="new">3.5.1</Tag>.
```
adrianeboyd, line 20: This is a bit tricky, we don't usually use the new tag in running text. Something like this (and then formatted)?

```suggestion
This component comes in two forms: `spancat` and `spancat_singlelabel` (added in spaCy v3.5.1). When you
need to perform multi-label classification on your spans, use `spancat`. The
`spancat` component uses a `Logistic` layer where the output class probabilities
are independent for each class. However, if you need to predict at most one true
class for a span, then use `spancat_singlelabel`.
```

website/docs/api/spancategorizer.mdx
```diff
@@ -52,14 +59,33 @@ architectures and their arguments and hyperparameters.
...
+> config = {
+>     "threshold": 0.5,
+>     "spans_key": "labeled_spans",
+>     "model": DEFAULT_SPANCAT_SINGLELABEL_MODEL,
+>     "suggester": {"@misc": "spacy.ngram_suggester.v1", "sizes": [1, 2, 3]},
+>     # Additional spancat_exclusive parameters
+>     "negative_weight": 0.8,
+>     "allow_overlap": True,
+> }
+> nlp.add_pipe("spancat_exclusive", config=config)
```
adrianeboyd, line 75:
```suggestion
>     # Additional spancat_singlelabel parameters
>     "negative_weight": 0.8,
>     "allow_overlap": True,
> }
> nlp.add_pipe("spancat_singlelabel", config=config)
```

spacy/pipeline/spancat.py
```diff
@@ -185,25 +308,46 @@ def __init__(
...
-        threshold: float = 0.5,
+        negative_weight: Optional[float] = 1.0,
+        allow_overlap: Optional[bool] = True,
         max_positive: Optional[int] = None,
+        threshold: Optional[float] = 0.5,
         scorer: Optional[Callable] = spancat_score,
     ) -> None:
-        """Initialize the span categorizer.
+        """Initialize the multi-label or multi-class span categorizer.
+        The 'single_label' argument configures whether the component
```
adrianeboyd, line 320: There are a number of out-of-date references to `single_label` in the docstring.

kadarakos: ugh right, sorry, let me fix it!

spacy/pipeline/spancat.py
```diff
@@ -185,25 +308,49 @@ def __init__(
...
+        negative_weight: Optional[float] = 1.0,
+        allow_overlap: Optional[bool] = True,
         max_positive: Optional[int] = None,
+        threshold: Optional[float] = 0.5,
         scorer: Optional[Callable] = spancat_score,
     ) -> None:
-        """Initialize the span categorizer.
+        """Initialize the multi-label or multi-class span categorizer.
+
+        argument configures whether the component
```
adrianeboyd, line 321: It looks like something went wrong here?

adrianeboyd:
I think this is good to go!

PR #11370, opened 2022-08-24
Update/remove old Matcher syntax

polm:
## Description

In v2 `Matcher.add` was called with `(key, on_match, *patterns)`. In v3 this
was changed to `(key, patterns, *, on_match=None)`, but there were various
points where the old call syntax was documented or handled specially.
This removes all those.

The Matcher itself didn't need any code changes, as it just gives a
generic type error. However the PhraseMatcher required some changes
because it would automatically "fix" the old call style.

Surprisingly, the tokenizer was still using the old call style in one
place.

After these changes tests failed in two places:

1. one test for the "new" call style, including the "old" call style. I removed this test.
2. deserializing the PhraseMatcher fails because the input Docs are a set.

I am not sure why 2 is happening - I guess it's a quirk of the
serialization format? - so for now I just convert the set to a list when
deserializing. The check that the input Docs are a List in the
PhraseMatcher is a new check, but makes it parallel with the other
Matchers, which seemed like the right thing to do.

I am marking this as a draft because I want to check other places where `Matcher.add` is called internally that may not have come up in tests directly.

### Types of change

remove deprecated APIs

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -110,9 +110,9 @@ cdef class Matcher:
         """
         errors = {}
         if on_match is not None and not hasattr(on_match, "__call__"):
-            raise ValueError(Errors.E171.format(arg_type=type(on_match)))
-        if patterns is None or not isinstance(patterns, List):  # old API
-            raise ValueError(Errors.E948.format(arg_type=type(patterns)))
+            raise ValueError(Errors.E171.format(name="Matcher", arg_type=type(on_match)))
+        if patterns is None or not isinstance(patterns, List):
+            raise ValueError(Errors.E948.format(name="Matcher", arg_type=type(patterns)))
         if greedy is not None and greedy not in ["FIRST", "LONGEST"]:
             raise ValueError(Errors.E947.format(expected=["FIRST", "LONGEST"], arg=greedy))
         for i, pattern in enumerate(patterns):
--- a/spacy/matcher/phrasematcher.pyx
+++ b/spacy/matcher/phrasematcher.pyx
@@ -1,4 +1,5 @@
 # cython: infer_types=True, profile=True
+from typing import List
 from libc.stdint cimport uintptr_t
 from preshed.maps cimport map_init, map_set, map_get, map_clear, map_iter
 
@@ -155,26 +156,24 @@ cdef class PhraseMatcher:
         del self._callbacks[key]
         del self._docs[key]
 
-    def add(self, key, docs, *_docs, on_match=None):
+    def add(self, key, docs, *, on_match=None):
         """Add a match-rule to the phrase-matcher. A match-rule consists of: an ID
-        key, an on_match callback, and one or more patterns.
-
-        Since spaCy v2.2.2, PhraseMatcher.add takes a list of patterns as the
-        second argument, with the on_match callback as an optional keyword
-        argument.
+        key, a list of one or more patterns, and (optionally) an on_match callback.
 
         key (str): The match ID.
         docs (list): List of `Doc` objects representing match patterns.
         on_match (callable): Callback executed on match.
-        *_docs (Doc): For backwards compatibility: list of patterns to add
-            as variable arguments. Will be ignored if a list of patterns is
-            provided as the second argument.
 
         DOCS: https://spacy.io/api/phrasematcher#add
         """
-        if docs is None or hasattr(docs, "__call__"):  # old API
-            on_match = docs
-            docs = _docs
+        # TODO we should need only one of the first two checks here
+        if docs and isinstance(docs, Doc):
+            raise ValueError(Errors.E179.format(key=key))
+        # XXX This fails in deserialization because docs are a set
+        if docs is None or not isinstance(docs, List):
+            raise ValueError(Errors.E948.format(name="PhraseMatcher", arg_type=type(docs)))
+        if on_match is not None and not hasattr(on_match, "__call__"):
+            raise ValueError(Errors.E171.format(name="PhraseMatcher", arg_type=type(on_match)))
 
         _ = self.vocab[key]
         self._callbacks[key] = on_match
@@ -184,8 +183,6 @@ cdef class PhraseMatcher:
         cdef MapStruct* internal_node
         cdef void* result
 
-        if isinstance(docs, Doc):
-            raise ValueError(Errors.E179.format(key=key))
         for doc in docs:
             if len(doc) == 0:
                 continue
@@ -345,6 +342,8 @@ def unpickle_matcher(vocab, docs, callbacks, attr):
     matcher = PhraseMatcher(vocab, attr=attr)
     for key, specs in docs.items():
         callback = callbacks.get(key, None)
+        # TODO is this correct?
+        specs = list(specs)
         matcher.add(key, specs, on_match=callback)
     return matcher
 
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -487,7 +487,7 @@ class Errors(metaclass=ErrorsWithCodes):
             "Current DocBin: {current}\nOther DocBin: {other}")
     E169 = ("Can't find module: {module}")
     E170 = ("Cannot apply transition {name}: invalid for the current state.")
-    E171 = ("Matcher.add received invalid 'on_match' callback argument: expected "
+    E171 = ("{name}.add received invalid 'on_match' callback argument: expected "
             "callable or None, but got: {arg_type}")
     E175 = ("Can't remove rule for unknown match pattern ID: {key}")
     E176 = ("Alias '{alias}' is not defined in the Knowledge Base.")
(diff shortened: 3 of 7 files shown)


spacy/matcher/matcher.pyx
```diff
@@ -110,9 +110,9 @@ cdef class Matcher:
         """
         errors = {}
         if on_match is not None and not hasattr(on_match, "__call__"):
-            raise ValueError(Errors.E171.format(arg_type=type(on_match)))
-        if patterns is None or not isinstance(patterns, List):  # old API
-            raise ValueError(Errors.E948.format(arg_type=type(patterns)))
+            raise ValueError(Errors.E171.format(name="Matcher", arg_type=type(on_match)))
+        if patterns is None or not isinstance(patterns, List):
+            raise ValueError(Errors.E948.format(name="Matcher", arg_type=type(patterns)))
```
polm, line 115: This didn't require significant change because the error is generic and doesn't mention the old call style except in the comment.

spacy/matcher/phrasematcher.pyx
```diff
@@ -155,26 +156,24 @@ cdef class PhraseMatcher:
         del self._callbacks[key]
         del self._docs[key]
 
-    def add(self, key, docs, *_docs, on_match=None):
+    def add(self, key, docs, *, on_match=None):
```
polm, line 159: This is the breaking change.

spacy/matcher/phrasematcher.pyx
```diff
@@ -155,26 +156,24 @@ cdef class PhraseMatcher:
...
         """
-        if docs is None or hasattr(docs, "__call__"):  # old API
-            on_match = docs
-            docs = _docs
+        # TODO we should need only one of the first two checks here
+        if docs and isinstance(docs, Doc):
+            raise ValueError(Errors.E179.format(key=key))
+        # XXX This fails in deserialization because docs are a set
+        if docs is None or not isinstance(docs, List):
+            raise ValueError(Errors.E948.format(name="PhraseMatcher", arg_type=type(docs)))
```
polm, line 174: The first test here was a little further down originally. The second test wasn't here, but is consistent with the other Matchers. 

I think the specificity of the first error is good, and it would be reasonable to leave both of these or either of them. I would lean towards leaving in both.

There's also the question of whether we want to require a List here. The other Matchers do. I think that's reasonable since we're going to store a list internally.

polm (commented):
Some notes on places that need particular attention.

spacy/matcher/phrasematcher.pyx
```diff
@@ -345,6 +342,8 @@ def unpickle_matcher(vocab, docs, callbacks, attr):
     matcher = PhraseMatcher(vocab, attr=attr)
     for key, specs in docs.items():
         callback = callbacks.get(key, None)
+        # TODO is this correct?
+        specs = list(specs)
```
polm, line 346: I haven't dug down to the root of this, but I assume this is just a quirk of serialization and it's OK to do this. I can look into it more if it's surprising.

adrianeboyd: Fix this starting at `__reduce__` instead to work with `sorted(dict.items())`.

adrianeboyd: (I'm not entirely sure what the error in the tests was, but there aren't any sets here.)

polm: As part of this PR I added a check that the list of Docs to `PhraseMatcher.add` is a `List` (`isinstance(docs, List)`), and that fails with an error that the variable is a `set`. Let me look at what's going on more...

polm: OK, I figured out the error. 

The PhraseMatcher stores things to match in `self._docs`, which is of type `Dict[str, Set[Tuple[attr_t]]]`. The input `docs` can be a list of Docs, which would be then be converted for internal storage, or the pre-processed values stored in this internal dict. The input Doc objects are not saved anywhere, so serialization is using these sets.

I will loosen the type check for now, though the multi-type stuff here is a little confusing.

adrianeboyd: That is weirder than I remembered.

polm: What we could do, if we wanted the types/interface to be more uniform, is have a separate method that just sets these processed values. It could be called something like `_set_internal_doc_representations` or `_add_processed_patterns` and would normally only be used in deserialization. It would just be an assignment. 

I don't think either way is cleaner so my inclination would be just to leave things as they are.

polm:
I checked all calls to `.add(` in the repo and didn't find anywhere else using the old call syntax, except for one spot in the docs I'd missed. So, unless further changes are desired (like the one for cleaner typing mentioned upthread), I think this is good.

spacy/matcher/phrasematcher.pyx
```diff
@@ -155,26 +155,24 @@ cdef class PhraseMatcher:
...
         DOCS: https://spacy.io/api/phrasematcher#add
         """
-        if docs is None or hasattr(docs, "__call__"):  # old API
-            on_match = docs
-            docs = _docs
+        # NOTE: in normal usage, `docs` will be a List of Docs. However, it can
+        # also be of type Set[Tuple[attr_t]], which is the type used for
+        # internal storage of patterns. This is used in deserialization, for
+        # example.
+        if docs and isinstance(docs, Doc):
```
adrianeboyd, line 172:
```suggestion
        if isinstance(docs, Doc):
```

spacy/matcher/phrasematcher.pyx
```diff
@@ -155,26 +155,24 @@ cdef class PhraseMatcher:
...
-            provided as the second argument.
 
         DOCS: https://spacy.io/api/phrasematcher#add
         """
-        if docs is None or hasattr(docs, "__call__"):  # old API
-            on_match = docs
-            docs = _docs
+        # NOTE: in normal usage, `docs` will be a List of Docs. However, it can
+        # also be of type Set[Tuple[attr_t]], which is the type used for
+        # internal storage of patterns. This is used in deserialization, for
```
adrianeboyd, line 170:
```suggestion
        # internal storage of patterns. This is used in unpickling, for
```

spacy/matcher/phrasematcher.pyx
```diff
@@ -210,6 +206,7 @@ cdef class PhraseMatcher:
                     warnings.warn(Warnings.W012.format(key=key, attr=string_attr))
                 keyword = self._convert_to_array(doc)
             else:
+                # The doc here should be a Sequence of attr_ts. This is used when deserializing.
```
adrianeboyd, line 209:
```suggestion
                # The doc here should be a Sequence of attr_ts. This is used when unpickling.
```

spacy/matcher/phrasematcher.pyx
adrianeboyd: It is kind of messy and since we were both confused, I think it probably would be worth it to add a function to do this. Maybe `_add_from_arrays` or something like that? I guess you'd have to move the actual adding to another private method, too, hmm...

polm: OK, I've changed this to have an `_add_from_array` function, which takes a key and the pattern list. This directly replaces the last bit of `add`. For unpickling I added `_add_from_arrays`, which is not quite just a for loop over the single function - it's more like `add` without preprocessing/safety features, in that it takes a callback. 

One change that isn't strictly necessary here is that I made `self._docs` a `defaultdict` - since `setdefault` was being called for every key, this shouldn't change the behaviour at all, and it means one less thing to initialize when adding.

While working on this, one thought I had was that it would be nice to replace part of `add` with a call to the batch method, which would remove the need for the single add method. However I realized that doesn't match the behaviour of `add` at the moment - if you have multiple patterns, and one of them is invalid, everything up until the invalid pattern will be added (because the items are verified and then added immediately, one at a time). That seems kind of weird, should we change it to verify all the patterns before adding any of them?

adrianeboyd: I think it would be okay to change the behavior to do all the validation first.

polm: OK, I have changed this to do the validation first.

spacy/matcher/phrasematcher.pyx
```diff
@@ -210,28 +249,10 @@ cdef class PhraseMatcher:
                     warnings.warn(Warnings.W012.format(key=key, attr=string_attr))
                 keyword = self._convert_to_array(doc)
             else:
+                # The doc here should be a Sequence of attr_ts. This is used when unpickling.
                 keyword = doc
```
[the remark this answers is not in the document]
polm: Right, good point, I removed it. 

It turns out the types in `pyi` were already specific and didn't cover the unpickle usage types, so they didn't require change. I went ahead and added the `_add_from_arrays` there, though I'm not sure it was necessary.

spacy/matcher/phrasematcher.pyx
```diff
@@ -155,66 +157,24 @@ cdef class PhraseMatcher:
...
 
-        Since spaCy v2.2.2, PhraseMatcher.add takes a list of patterns as the
-        second argument, with the on_match callback as an optional keyword
-        argument.
+    def _add_from_arrays(self, key, specs, *, on_match=None):
+        """Add a preprocessed list of specs, with an optional callback.
 
         key (str): The match ID.
-        docs (list): List of `Doc` objects representing match patterns.
+        specs (sequence): A sequence of sequences of hashes to match.
```
adrianeboyd, line 165:
```suggestion
        specs (List[List[int]]): A list of lists of hashes to match.
```

PR #11371, opened 2022-08-24
testing pydantic v1.10.0a1

samuelcolvin:
### TO NOT MERGE

This is pull request is just to test pydantic v1.10, see https://github.com/pydantic/pydantic/discussions/4359

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ install_requires =
     tqdm>=4.38.0,<5.0.0
     numpy>=1.15.0
     requests>=2.13.0,<3.0.0
-    pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0
+    pydantic==1.10.0a1
     jinja2
     # Official Python utilities
     setuptools


adrianeboyd:
I think that `thinc` might be a better candidate for this test usage than `spacy`. Most of the interesting pydantic usage is in `thinc` anyway and `spacy` would require an updated pydantic version in `thinc` first, or there will be version conflicts.

It will get slightly more complicated in `thinc` at some point because we're moving the interesting config usage into a separate package (`confection`, to be released very soon), but if you start from the `v8.0.x` branch I think it should cover the cases you're interested in.

samuelcolvin:
Thanks so much will do.

adrianeboyd:
Wait, sorry, no not `v8.0.x` because that's not updated for pydantic v1.9. The current `master` (currently `01eb6b78`) should be okay, just be aware that it might change with `confection`.

samuelcolvin:
will do, thanks.

PR #11376, opened 2022-08-24
Add apply CLI

kadarakos:
New spacy cli command to make annotating new documents with the a trained pipeline convenient by just running a single command.

## Description
Adds a new `spacy cli` command called `apply` that takes a model and a file or directory as input.
The file or the files in the directory should be of the following formats:

1. `.spacy` file.
2. `.jsonl` file where the provided `--text-key` is used to find the documents.  
3. Any other file found is considered a plain text file containing a single document.

Given the input file or directory `apply` generates a `.spacy` file containing the `Doc` objects produced by the model.

### Types of change
New feature

## Checklist


kadarakos:
The way the data loading is done is with a bunch of nested `try`/`except` blocks, which I feel like is something we would like to avoid in general but it felt valid here. 

Also this `cli` command is really short and simple compared to the rest so I feel like I must have missed a lot of things.

kadarakos:
> I think it would make more sense to go by file ending here (maybe `.spacy` vs. not-`.spacy`) and take it from there?

okok and then only the decode error remains in the tryexcept

spacy/cli/apply.py
```diff
@@ -0,0 +1,124 @@
...
+    vocab: Vocab,
+    suffix: Optional[str] = None
+) -> Generator[Union[str, Doc], None, None]:
+    """
+    Load data which is either in a single file
+    in .spacy or plain text format or multiple
+    text files in a directory. If a directory
+    is provided skip subdirectories and undecodeable
+    files.
+    """
```
adrianeboyd, line 37: It should also be able to handle recursively searching a directory of `.spacy` files. Look at how `spacy convert` handles this?

kadarakos: Ok I checked `walk_directory` thanks! I don't even accept multiple `.spacy` files at the moment, because I was wondering why would one have multiple of them. If we do the `walk_directory` then if the user provides a directory as input I will also require a `suffix`.

kadarakos: Changed the behavior to either take a single file which can be plain text or `.spacy` file or a directory which then gets walked but then a `suffix` needs to be provided.

adrianeboyd: I think that the existing `walk_directory` function may have too much convert-specific logic in it for use here. I'm not sure we've clearly defined what the accepted non-`.spacy` formats are? The PR description says one text per line but the code looks like it's treated each non-`.spacy` file as a single text? I also though there was some discussion about supporting `"text"` in JSONL dicts?

The pipeline can potentially take a mix of texts and `Doc`s, so it may not be necessary to provide a suffix.

kadarakos: That's true that we could support a mix, but then we should say that okay, you have to have a directory where:

1. text files are called `.txt` and we treat each file as one document.
2. `jsonl` files are called `.jsonl` and we will read the "text" field.
3. you can have serialized `DocBin`s and they must be called `.spacy` also.

So I can change then the behavior by not having this option that one text file can have multiple documents, rather, we are always walking the directory that the user provided and we take the extensions at face value. How does this sound?

adrianeboyd: To avoid too many restrictions in terms of filenames, I'd recommend:

1. `.spacy`
2. `.jsonl`
3. anything else is treated as text with one text per file

spacy/cli/apply.py
```diff
@@ -0,0 +1,133 @@
...
+    paths = walk_directory(data_path)
+    streams: List[Union[Iterable[str], Iterable[Doc]]] = []
+    text_files = []
+    for path in paths:
+        if path.suffix == ".spacy":
+            streams.append(_stream_docbin(path, vocab))
+        elif path.suffix == ".jsonl":
+            streams.append(_stream_jsonl(path))
+        else:
+            text_files.append(path)
```
adrianeboyd, line 125: I was expecting to see `_stream_text` implemented in the same way as the other two. Is there a reason for the difference?

spacy/cli/_util.py
```diff
@@ -404,7 +404,10 @@ def git_checkout(
             if not is_subpath_of(tmp_dir, source_path):
                 err = f"'{subpath}' is a path outside of the cloned repository."
                 msg.fail(err, repo, exits=1)
-            shutil.copytree(str(source_path), str(dest))
+            if source_path.is_dir():
+                shutil.copytree(str(source_path), str(dest))
+            else:
+                shutil.copy(str(source_path), str(dest))
```
adrianeboyd, line 410: This change in unrelated.

spacy/cli/convert.py
```diff
@@ -189,33 +189,6 @@ def autodetect_ner_format(input_data: str) -> Optional[str]:
     return None
 
 
-def walk_directory(path: Path, converter: str) -> List[Path]:
```
adrianeboyd, line 192: I think the changes in this file should be reverted in order not to affect the `spacy convert` behavior.

spacy/cli/apply.py
kadarakos: The other two stream from a single file, but the `_stream_text` takes a list of paths as input. The
`for path in paths` loop create the `_stream_docbin` and `_stream_jsonl` generators and collects the rest of the paths into the `text_files` list. 

Then if we found any text files the generator gets created:

```
if len(text_files) > 0:
        streams.append(filter(None, _stream_texts(text_files)))
```

Its also different from the other in that sometimes it yields `None` if when reading some files we encounter `UnicodeDecodeError`. That's why its also wrapped in `filter(None, ...)`.

spacy/cli/convert.py
kadarakos: I'm not sure how is the behavior affected? The only (intended) change to `walk_directory` is that providing the `suffix` is now `Optional`.

spacy/cli/apply.py
adrianeboyd: It's not that I don't see how this implementation works in general, but that I think it would be simpler and easier to maintain if it were implemented the same way as `_stream_jsonl`, where the method would just handle a single file and yield a single text from that file if it can open it, and deal with potential warnings/errors in the same way. So that we don't have one method that raises an error and one that warns and returns `None`s that have be filtered out.

spacy/cli/convert.py
adrianeboyd: Ah, I thought the converter/suffix behavior was more complicated than it is (and we just don't have good tests here that would catch unintended changes).

spacy/cli/apply.py
kadarakos: I didn't know what you mean by it not being "the same". I'll change it to raise error instead of return None.

adrianeboyd: The current version still warns in one place and raises an error in another similar place. For this functionality, I think I would lean toward warning, but if you think errors make more sense that is also fine.

I really want to see a `_stream_text` function that works the exact same way as `_stream_jsonl` for a file with one line instead of this mixture of behaviors and warnings/errors.

kadarakos: ok they both `raise` now

spacy/tests/test_cli.py
```diff
@@ -855,3 +856,69 @@ def test_span_length_freq_dist_output_must_be_correct():
     span_freqs = _get_spans_length_freq_dist(sample_span_lengths, threshold)
     assert sum(span_freqs.values()) >= threshold
     assert list(span_freqs.keys()) == [3, 1, 4, 5, 2]
+
+
+def test_applycli_empty_dir():
+    with make_tempdir() as data_path:
+        output = os.path.join(data_path / "test.spacy")
```
adrianeboyd, line 863:
```suggestion
        output = data_path / "test.spacy"
```

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+def _stream_jsonl(path: Path) -> Iterable[str]:
+    """
+    Stream "text" field from JSONL. If the field "text" is
+    not found it raises error.
+    """
+    for entry in srsly.read_jsonl(path):
+        if "text" not in entry:
+            raise ValueError(
+                f"{path} does not contain the required 'text' field."
+            )
```
adrianeboyd, line 49: In the CLI commands we'd use `msg.fail(message, exits=1)` instead of raising a python error.

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+        else:
+            text_files.append(path)
+    if len(text_files) > 0:
+        streams.append(_stream_texts(text_files))
+    datagen = cast(DocOrStrStream, chain(*streams))
+    for doc in tqdm.tqdm(nlp.pipe(datagen, batch_size=batch_size, n_process=n_process)):
+        docbin.add(doc)
+    if output_path.is_dir():
+        output_path = output_path / "predictions.spacy"
+    docbin.to_disk(output_path)
```
adrianeboyd, line 123: I'm not sure about the name `predictions.spacy`.

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+        yield doc
+
+
+def _stream_jsonl(path: Path) -> Iterable[str]:
+    """
+    Stream "text" field from JSONL. If the field "text" is
+    not found it raises error.
+    """
+    for entry in srsly.read_jsonl(path):
+        if "text" not in entry:
```
svlandeg, line 46: I would make the name of this JSON field configurable.

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+            text = fin.read()
+            yield text
+
+
+@app.command("apply")
+def apply_cli(
+    # fmt: off
+    model: str = Arg(..., help="Model name or path"),
+    data_path: Path = Arg(..., help=path_help, exists=True),
+    output: Path = Arg(..., help=out_help, dir_okay=False),
```
svlandeg, line 69: I would rename this `output_file` to make it clear it's always a file and never a directory

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+            streams.append(_stream_jsonl(path))
+        else:
+            text_files.append(path)
+    if len(text_files) > 0:
+        streams.append(_stream_texts(text_files))
+    datagen = cast(DocOrStrStream, chain(*streams))
+    for doc in tqdm.tqdm(nlp.pipe(datagen, batch_size=batch_size, n_process=n_process)):
+        docbin.add(doc)
+    if output_path.is_dir():
+        output_path = output_path / "predictions.spacy"
```
svlandeg, line 122: This code path can't actually happen in the current implementation as `dir_okay` is set to `False` for `output`. I would keep it as such, as a single file in a directory seems a bit silly?

svlandeg (commented):
I think this will be a good additional CLI tool to have!

Left some comments & suggestions on the current implementation - nothing major ;-)

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+DocOrStrStream = Union[Iterable[str], Iterable[Doc]]
+
+
+def _stream_docbin(path: Path, vocab: Vocab) -> Iterable[Doc]:
+    """
+    Stream Doc objects from DocBin.
+    """
+    docbin = DocBin().from_disk(path)
+    for doc in docbin.get_docs(vocab):
+        yield doc
```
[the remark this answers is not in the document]
adrianeboyd: They can't control this in the default `Corpus` either? I see this usage for things like adding sentence boundaries or entities automatically that are needed by other components for training.

It looks like user data is always loaded by `DocBin.from_disk`, but it probably does make sense to be sure that user data is always included in the new saved `DocBin`?

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+):
+    data_path = ensure_path(data_path)
+    output_path = ensure_path(output)
+    if not data_path.exists():
+        msg.fail("Couldn't find data path.", data_path, exits=1)
+    nlp = load_model(model)
+    msg.good(f"Loaded model {model}")
+    vocab = nlp.vocab
+    docbin = DocBin()
+    paths = walk_directory(data_path)
```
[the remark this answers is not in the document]
kadarakos: I'll change the docstring instead I think, because we agreed with @adrianeboyd so far to work on:

1. `.jsonl` with a `"text"` field or some other specified field as you mention in a later comment.
2. `.spacy` files.
3. any other file is assumed to be a text file containing a single document.

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+from ..vocab import Vocab
+from ..util import ensure_path, load_model
+
+
+path_help = ("Location of the documents to predict on."
+             "Can be a single file in .spacy format or "
+             "a text file with one document per line."
+             "If a directory is provided each "
+             "text file in the directory will be treated "
+             "as a single document.")
```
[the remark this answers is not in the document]
kadarakos: Your right, this is not true anymore.

spacy/cli/apply.py
kadarakos: Makes sense. How do you test this with `pytest`? Is it possible?

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+out_help = "Path where to save the result .spacy file"
+code_help = ("Path to Python file with additional "
```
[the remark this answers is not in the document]
kadarakos: I think that makes sense, now if the user's path does not contain an extension we append `.spacy`.

spacy/cli/apply.py
```diff
@@ -0,0 +1,137 @@
...
+    output_file: Path,
+    model: str,
+    json_field: str,
+    batch_size: int,
+    n_process: int,
+):
+    nlp = load_model(model)
+    msg.good(f"Loaded model {model}")
+    vocab = nlp.vocab
+    docbin = DocBin()
```
adrianeboyd, line 119: This is where you need `store_user_data=True` in order not to lose the user data from any existing annotation in `.spacy` files or from custom extensions used in the pipeline.

spacy/tests/test_cli.py
```diff
@@ -859,6 +861,66 @@ def test_span_length_freq_dist_output_must_be_correct():
...
+            ftest.write("Testing apply cli.")
+        apply(data_path, output, "blank:en", "text", 1, 1)
+
+
+def test_applycli_mixed():
+    with make_tempdir() as data_path:
+        output = data_path / "testout.spacy"
+        text = "Testing apply cli"
+        nlp = spacy.blank("en")
+        doc = nlp(text)
```
adrianeboyd, line 908: Can you include some examples with custom extensions in this test?

spacy/cli/apply.py
```diff
@@ -0,0 +1,137 @@
...
+path_help = ("Location of the documents to predict on."
+             "Can be a single file in .spacy format or "
+             ".jsonl file and files with other extensions"
+             "are treated as single plain text documents."
+             "If a directory is provided "
+             "it is traversed recursively to grab all files to "
+             "be processed. The files can be a mixture of .spacy"
+             ".jsonl and text files. If .jsonl is provided the "
+             "specified field is going to be grabbed ('text') "
+             "by default.")
```
svlandeg, line 26:
```suggestion
path_help = ("Location of the documents to predict on. "
             "Can be a single file in .spacy format or a "
             ".jsonl file. Files with other extensions "
             "are treated as single plain text documents. "
             "If a directory is provided "
             "it is traversed recursively to grab all files to "
             "be processed. The files can be a mixture of .spacy, "
             ".jsonl and text files. If .jsonl is provided the "
             "specified field is going to be grabbed ('text' "
             "by default).")
```

spacy/cli/apply.py
```diff
@@ -0,0 +1,137 @@
...
+def _stream_docbin(path: Path, vocab: Vocab) -> Iterable[Doc]:
+    """
+    Stream Doc objects from DocBin.
+    """
+    docbin = DocBin().from_disk(path)
+    for doc in docbin.get_docs(vocab):
+        yield doc
+
+
+def _stream_jsonl(path: Path, field) -> Iterable[str]:
```
svlandeg, line 47:
```suggestion
def _stream_jsonl(path: Path, field: str) -> Iterable[str]:
```

spacy/cli/apply.py
```diff
@@ -66,39 +74,45 @@ def apply_cli(
     # fmt: off
     model: str = Arg(..., help="Model name or path"),
     data_path: Path = Arg(..., help=path_help, exists=True),
-    output: Path = Arg(..., help=out_help, dir_okay=False),
+    output_file: Path = Arg(..., help=out_help, dir_okay=False),
     code_path: Optional[Path] = Opt(None, "--code", "-c", help=code_help),
+    json_field: str = Opt("text", "--field", "-f", help="Field to grab from .jsonl"),
```
svlandeg, line 79: I think we should call this `jsonl_field` to be consistent, right?

spacy/cli/apply.py
```diff
@@ -66,39 +74,45 @@ def apply_cli(
...
-    Expects a loadable spaCy pipeline and some data as input.
-    The data can be provided multiple formats. It can be a single
-    .spacy file or a single text file with one document per line.
-    A directory can also be provided in which case the 'suffix'
-    argument is required. All paths pointing to files with the
-    provided suffix will be recursively collected and processed.
-
+    Expects a loadable spaCy pipeline and path to the data, which
+    can be a directory or a file.
+    The data files can be provided multiple formats:
```
svlandeg, line 89:
```suggestion
    The data files can be provided in multiple formats:
```

spacy/cli/apply.py
kadarakos: sorry forgot to push it yesterday

spacy/tests/test_cli.py
kadarakos: Included a specific test for it, let me know if we'd need more examples.

spacy/cli/apply.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@app.command("apply")
+def apply_cli(
+    # fmt: off
+    model: str = Arg(..., help="Model name or path"),
+    data_path: Path = Arg(..., help=path_help, exists=True),
+    output_file: Path = Arg(..., help=out_help, dir_okay=False),
+    code_path: Optional[Path] = Opt(None, "--code", "-c", help=code_help),
+    json_field: str = Opt("text", "--json-field", "-jf", help="Field to grab from .jsonl"),
```
svlandeg, line 79: What I meant was `jsonl_field` instead of `json_field` because we're processing JSONL and not JSON, right?

adrianeboyd: I think JSONL is more confusing than JSON in this context, and I'm not sure "field" is the right name for a dictionary key? Do we use "field" for this in other places?

kadarakos: i thought of `field` in the sense of a schema and could be a `key` in the sense of map. I have no idea what's better.

svlandeg: `text_key`?

adrianeboyd: I think something like `text_key` and "Key containing text string for JSONL"?

spacy/cli/apply.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    Apply a trained pipeline to documents to get predictions.
+    Expects a loadable spaCy pipeline and path to the data, which
+    can be a directory or a file.
+    The data files can be provided in multiple formats:
+        1. .spacy files
+        2. .jsonl files with a specified "field" to read the text from.
+        3. Files with any other extension are assumed to be containing
+           a single document.
+    DOCS: https://spacy.io/api/cli#tba
```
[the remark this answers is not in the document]
svlandeg: It's much more pleasant to review if the docs are already formatted, and it's just best practice to always push changes that are already black/prettier formatted 🙏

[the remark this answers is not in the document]
kadarakos: okok ran prettier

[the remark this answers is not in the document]
svlandeg: It still has `tba` here

spacy/cli/_util.py
```diff
@@ -582,6 +582,29 @@ def setup_gpu(use_gpu: int, silent=None) -> None:
             local_msg.info("To switch to GPU 0, use the option: --gpu-id 0")
 
 
+def walk_directory(path: Path, suffix: Optional[str] = None) -> List[Path]:
```
polm, line 585: This recursively gets all non-hidden files with a certain suffix in a directory? It would be good to have a docstring, as the function name definitely doesn't make that clear to me. 

Additionally I think this can be simplified a lot, something like:

```
paths = path.rglob(f"*/*.{suffix}")
out = []
for path in paths:
    hidden = False
    for part in path.parts:
        if part.startswith("."):
            hidden = True
    if not hidden:
        out.append(path)
return out
```

And also... do we care about hidden files? I'm not sure there's a need for them in data directories, and this isn't how hidden files work on Windows, for example.

spacy/cli/apply.py
```diff
@@ -0,0 +1,123 @@
...
+from ..util import ensure_path, load_model
+
+
+path_help = ("Location of the documents to predict on."
+             "Can be a single file in .spacy format or "
+             "a text file with one document per line."
+             "If a directory is provided each "
+             "text file in the directory will be treated "
+             "as a single document.")
+out_help = "Path where to save the result .spacy file"
```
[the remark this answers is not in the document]
polm: The suggestion here doesn't seem to have been applied?

```suggestion
out_help = "Path to save the resulting .spacy file"
```

spacy/cli/apply.py
```diff
@@ -0,0 +1,141 @@
...
+    "Can be a single file in .spacy format or a "
+    ".jsonl file. Files with other extensions "
+    "are treated as single plain text documents. "
+    "If a directory is provided "
+    "it is traversed recursively to grab all files to "
+    "be processed. The files can be a mixture of .spacy, "
+    ".jsonl and text files. If .jsonl is provided the "
+    "specified field is going to be grabbed ('text' "
+    "by default)."
+)
```
polm, line 28: I know we do this sometime with error strings, but it would probably be easier to make this one big string with triple quotes?

spacy/cli/_util.py
polm: OK, I see now that this was moved from the `convert` code, in which case my comments don't apply to this PR. But we may want to revisit this.

spacy/cli/apply.py
```diff
@@ -0,0 +1,141 @@
...
+def _stream_jsonl(path: Path, field: str) -> Iterable[str]:
+    """
+    Stream "text" field from JSONL. If the field "text" is
+    not found it raises error.
+    """
+    for entry in srsly.read_jsonl(path):
+        if field not in entry:
+            raise msg.fail(
+                f"{path} does not contain the required '{field}' field.", exits=1
+            )
```
polm, line 61: `msg.fail` doesn't create an exception, so it doesn't make sense to raise it. Also with `exits=1` it'll exit the program, so you can't raise anyway.

website/docs/api/cli.md
```diff
@@ -1162,6 +1163,37 @@ $ python -m spacy evaluate [model] [data_path] [--output] [--code] [--gold-prepr
...
+When a directory is provided it is traversed recursively to collect all files.
+
+```cli
+$ python -m spacy apply [model] [data-path] [output-file] [--code] [--text-key] [--force-overwrite] [--gpu-id] [--batch-size] [--n-process]
+```
+
+| Name                                      | Description                                                                                                                                                                          |
+| ----------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `model`                                   | Pipeline to apply to the data. Can be a package or a path to a data directory. ~~str (positional)~~                                                                                  |
+| `data_path`                               | Location of evaluation data in spaCy's [binary format](/api/data-formats#training). ~~Path (positional)~~                                                                            |
```
polm, line 1186:
```suggestion
| `data_path`                               | Location of data to be evaluated in spaCy's [binary format](/api/data-formats#training), jsonl, or plain text. ~~Path (positional)~~                                                                            |
```

website/docs/api/cli.md
```diff
@@ -1162,6 +1163,37 @@ $ python -m spacy evaluate [model] [data_path] [--output] [--code] [--gold-prepr
...
+| `output-file`, `-o`                       | Output `DocBin` path. ~~str (positional)~~                                                                                                                                           |
+| `--code`, `-c` <Tag variant="new">3</Tag> | Path to Python file with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~ |
+| `--text-key`, `-tk`                       | The key for `.jsonl` files to use to grab the texts from. ~~Optional[str] \(option)~~                                                                                                |
```
polm, line 1189:
```suggestion
| `--text-key`, `-tk`                       | The key for `.jsonl` files to use to grab the texts from. Defaults to `text`. ~~Optional[str] \(option)~~                                                                                                |
```

website/docs/api/cli.md
```diff
@@ -1162,6 +1163,37 @@ $ python -m spacy evaluate [model] [data_path] [--output] [--code] [--gold-prepr
...
+| `--force-overwrite`, `-F`                 | If the provided `output-file` already exists, then force `apply` to overwrite it. If this is `False` (default) then quits with a warning instead. ~~bool (flag)~~                    |
+| `--gpu-id`, `-g`                          | GPU to use, if any. Defaults to `-1` for CPU. ~~int (option)~~                                                                                                                       |
+| `--batch-size`, `-g`                      | Batch size to use for prediction. Defaults to `1`. ~~int (option)~~                                                                                                                  |
+| `--n-process`, `-g`                       | Number of processes to use for prediction. Defaults to `1`. ~~int (option)~~                                                                                                         |
```
polm, line 1193:
```suggestion
| `--batch-size`, `-b`                      | Batch size to use for prediction. Defaults to `1`. ~~int (option)~~                                                                                                                  |
| `--n-process`, `-n`                       | Number of processes to use for prediction. Defaults to `1`. ~~int (option)~~                                                                                                         |
```

polm (commented):
Thanks for working on this, I think it's a really important feature to have! I'm sorry I didn't look at it early. 

I have comments and docs fixes, but they're mostly pretty minor formatting stuff.

website/docs/api/cli.md
```diff
@@ -1162,6 +1163,37 @@ $ python -m spacy evaluate [model] [data_path] [--output] [--code] [--gold-prepr
...
+| `output-file`, `-o`                       | Output `DocBin` path. ~~str (positional)~~                                                                                                                                           |
+| `--code`, `-c` <Tag variant="new">3</Tag> | Path to Python file with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~ |
+| `--text-key`, `-tk`                       | The key for `.jsonl` files to use to grab the texts from. Defaults to `text`. ~~Optional[str] \(option)~~                                                                                                |
+| `--force-overwrite`, `-F`                 | If the provided `output-file` already exists, then force `apply` to overwrite it. If this is `False` (default) then quits with a warning instead. ~~bool (flag)~~                    |
+| `--gpu-id`, `-g`                          | GPU to use, if any. Defaults to `-1` for CPU. ~~int (option)~~                                                                                                                       |
```
svlandeg, line 1191: WIll need one more run of `prettier` ;-)

svlandeg (approved):
Thanks again for all your work on this Akos, this will be a great new command to have!

PR #11380, opened 2022-08-25
`SpanGroup(s)`-related optimizations

shadeMe:
## Description

We were needlessly performing a serialization-deserialization round-trip just to copy a string `Dict` of `SpanGroup`s. This PR replaces that with a straight-forward copy of the `dict`'s constituents. To facilitate this, a new keyword argument was added to `SpanGroup.copy()` that allows the user to pass a new reference document to which the copy is bound.

Improvements in performance can be seen in [this PR](https://github.com/explosion/spacy-experimental/pull/20).

### Types of change

Refactor, enhancement

## Checklist


shadeMe:

adrianeboyd:
Where is this kind of copying happening? I don't really understand the need for `new_doc`?

shadeMe:
[In `SpanGroups.copy()`](https://github.com/explosion/spaCy/pull/11380/files#diff-81ded29954f496c3a87231ea2ddcd2072b7d8adb92ace446893eaba46025d50dR45) > `Doc.copy()` > `_copy_examples` > `Language.update`. 

Without the new param, the copied `SpanGroup` instances will point to their original `Doc` objects. This was not the case previously as the `from_/to_bytes` code path initializes the former differently.

adrianeboyd:
Ah, I understand. Can you switch this have the same kwargs format as `SpanGroups.copy()`?

spacy/tokens/span_group.pyx
```diff
@@ -241,15 +241,17 @@ cdef class SpanGroup:
     cdef void push_back(self, SpanC span) nogil:
         self.c.push_back(span)
 
-    def copy(self)  -> SpanGroup:
+    def copy(self, new_doc: Optional["Doc"] = None)  -> SpanGroup:
```
adrianeboyd, line 244: Also the name:

```suggestion
    def copy(self, doc: Optional["Doc"] = None)  -> SpanGroup:
```

adrianeboyd:
I'm trying to think if there's something I've missed in terms of the refcounts, but nothing jumps out at me.

Can you also update `span_groups.pyi`, the docstrings, and the API docs?

spacy/tokens/span_group.pyx
```diff
@@ -241,15 +241,18 @@ cdef class SpanGroup:
     cdef void push_back(self, SpanC span) nogil:
         self.c.push_back(span)
 
-    def copy(self)  -> SpanGroup:
+    def copy(self, doc: Optional["Doc"] = None)  -> SpanGroup:
```
adrianeboyd, line 244:
```suggestion
    def copy(self, doc: Optional["Doc"] = None) -> SpanGroup:
```

website/docs/api/spangroup.md
```diff
@@ -255,9 +255,10 @@ Return a copy of the span group.
 > new_group = doc.spans["errors"].copy()
 > ```
 
-| Name        | Description                                     |
-| ----------- | ----------------------------------------------- |
-| **RETURNS** | A copy of the `SpanGroup` object. ~~SpanGroup~~ |
+| Name        | Description                                                                                                     |
+| ----------- | --------------------------------------------------------------------------------------------------------------- |
+| `doc`       | New reference document to which the copy is bound. Defaults to `None`. ~~Optional[Doc]~~                        |
+| **RETURNS** | A copy of the `SpanGroup` object, bound to either its original reference document or the new one. ~~SpanGroup~~ |
```
adrianeboyd, line 261: Better to avoid "new" because it's not necessarily new and we're not creating any docs here. I think it's also better to avoid "reference doc" because that may be confusing vs. `Example`.

```suggestion
| `doc`       | The document to which the copy is bound. Defaults to `None` for the current doc. ~~Optional[Doc]~~                        |
| **RETURNS** | A copy of the `SpanGroup` object. ~~SpanGroup~~ |
```

(and then reformat)

PR #11382, opened 2022-08-26
Fix lookup usage in French/Catalan (fix #11347)

### LINKED ISSUE #11347: French lemmatizer/Pos tagger issues, lemmatizer suggestion
Hello, 

As a follow up to #11298, I would like to share some examples for which the french tagger and lemmatizer fail to return the correct results (spaCy v3.4.1 and fr_core_news_lg v3.4.0). 

I'm using the "rule" mode for the french lemmatizer, but it fails to give the right lemma for some "simple" examples (eg., _je vous confirme mon numéro_, it returns "_confirme_" instead of "_confirmer_".
So I tried to figure out how to solve this by debugging its behavior on `rule_lemmatize`, and I came across this:

https://github.com/explosion/spaCy/blob/3e4cf1bbe1745a55ede0dece31353aebc3f82729/spacy/lang/fr/lemmatizer.py#L74-L77

If we don't find a rule for the token we check in the lookup table.
However, while debugging, I think that `string in lookup_table.keys()` will never be satisfied:

- `string = token.text, `and
- `lookup_table.keys()` is a list of odict_keys` odict_keys([11901859001352538922, 11991225512756778597,...)`

If I'm not mistaken, I think that's what's meant to be tested is: `token.norm in lookup_table.keys()`.

If that's the case, I think we can safely refactor the code as: 

```
        if not forms:
            forms.extend(oov_forms)        
            forms.append(self.lookup_lemmatize(token)[0])
        
```

[UPDATE]: after some more checking, I found that `string in lookup_table.keys()` will invoke` __getitem__` of `Table`. But I still don't get why don't we simply use `self.lookup_lemmatize(token)[0]` as I suggested.

Even if the token is not in the lookup table, `self.lookup_lemmatize(token)[0]` will do the same thing in line 77.

This will solve the lemmatization problems for examples like: "je vous confirme mon numéro..." (confirme will be lemmatized to "confirmer" instead of "confirme"). 

Here are some other examples for which we have tagging and lemmatization issues:

| **Text**                                                                                              | **Word analysed** | **Lemma detected** | **Lemma expected** | **Pos detected** | **Pos expected** | **Comments**                                                                          |
|-------------------------------------------------------------------------------------------------------|-------------------|--------------------|--------------------|------------------|------------------|---------------------------------------------------------------------------------------|
| je remarque qu'elle n'est toujours pas parti....                                                      | parti             | parti              | partir             | ADJ              | VERB             | "Parti" is a past participle it should be analyzed as a verb and not as an adjective  |
| alors que je l'avais annulé via le formulaire sur Internet.                                           | Internet          | Internet           | internet           | PROPN            | NOUN             |                                                                                       |
| Je vous confirme que mon numéro de téléphone est bien le [phone_number].                              | confirme          | confirme           | confirmer          | VERB             | VERB             | Problem"pos tagging" + LEMMA => same problem with the VERB "informe"                  |
| Comme vous me l’avez demandé par mail je vous communique mon numéro                                   | communique        | communique         | communiquer        | VERB             | VERB             | Problem"pos tagging" + LEMMA                                                          |
| Je vous redonne mon numéro                                                                            | redonne           | redonne            | redonner           | VERB             | VERB             | Problem"pos tagging" + LEMMA                                                          |
| le numéro de téléphone de votre service client communiqué en bas de vos mails                         | communiqué        | communiquer        | communiquer        | ADJ              | VERB             | Problem"pos tagging"                                                                  |
| Avez-vous un service client téléphone?                                                                | téléphone         | téléphon           | téléphone          | ADJ              | NOUN             | Problem"pos tagging" + LEMMA                                                          |
| MERCI DE SUPPRIMER DEFINITIVEMENT MON COMPTE ET TOUTES LES DONNEES PERSONNELLES ASSOCIEES A CELUI-CI. | SUPPRIMER         | supprimer          | supprimer          | NOUN             | VERB             | problem "pos tag"                                                                     |
| MERCI DE SUPPRIMER DEFINITIVEMENT MON COMPTE ET TOUTES LES DONNEES PERSONNELLES ASSOCIEES A CELUI-CI. | DEFINITIVEMENT    | definitivemer      | définitivement     | VERB             | ADV              | problem "pos tag" + lemma                                                             |
| MERCI DE SUPPRIMER DEFINITIVEMENT MON COMPTE ET TOUTES LES DONNEES PERSONNELLES ASSOCIEES A CELUI-CI. | DONNEES           | donnee             | donnée             | NOUN             | NOUN             | problem "lemma" + lemma                                                               |
| MERCI DE SUPPRIMER DEFINITIVEMENT MON COMPTE ET TOUTES LES DONNEES PERSONNELLES ASSOCIEES A CELUI-CI. | ASSOCIEES         | associee           | associer           | NOUN             | VERB             | problem "pos tag" + lemma                                                             |
| JE SOUHAITE IMPERATIVEMENT ETRE SUPPRIMEE DE VOS LISTES DE DIFFUSION.                                 | SOUHAITE          | souhait            | souhaiter          | ADJ              | VERB             | problem "pos tag" + lemma                                                             |
| JE SOUHAITE IMPERATIVEMENT ETRE SUPPRIMEE DE VOS LISTES DE DIFFUSION.                                 | IMPERATIVEMENT    | imperativement     | impérativement     | NOUN             | ADV              | problem "pos tag" + lemma                                                             |
| JE SOUHAITE IMPERATIVEMENT ETRE SUPPRIMEE DE VOS LISTES DE DIFFUSION.                                 | SUPPRIMEE         | supprime           | supprimer          | ADJ              | VERB             | problem "pos tag" + lemma                                                             |
| JE SOUHAITE IMPERATIVEMENT ETRE SUPPRIMEE DE VOS LISTES DE DIFFUSION.                                 | DIFFUSION         | DIFFUSION          | diffusion          | PROPN            | NOUN             | problem "pos tag" + lemma                                                             |

## Your Environment

* Operating System:
* Python Version Used: 3.8
* spaCy Version Used: 3.4

polm:
## Description

Before using the lookups table in the French (and Catalan) lemmatizers, there's a check to see if the current term is in the table. But it's checking a string against hashes, so it's always false. Also the lookup table is designed so that you don't have to check if something is already present in advance.

The lookup lemmatizer always falls back to the input string, so I've also modified these lemmatizers to not have their own separate fallback. However I'm not sure if that's something we should rely on or an implementation detail. 

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/lang/ca/lemmatizer.py
+++ b/spacy/lang/ca/lemmatizer.py
@@ -73,9 +73,9 @@ class CatalanLemmatizer(Lemmatizer):
         if not forms:
             forms.extend(oov_forms)
 
-        # use lookups, which fall back to the token itself
+        # use lookups, and fall back to the token itself
         if not forms:
-            forms.append(self.lookup_lemmatize(token)[0])
+            forms.append(lookup_table.get(token, [token])[0])
         forms = list(dict.fromkeys(forms))
         self.cache[cache_key] = forms
         return forms
--- a/spacy/lang/ca/lemmatizer.py
+++ b/spacy/lang/ca/lemmatizer.py
@@ -72,10 +72,10 @@ class CatalanLemmatizer(Lemmatizer):
                         oov_forms.append(form)
         if not forms:
             forms.extend(oov_forms)
-        if not forms and string in lookup_table.keys():
-            forms.append(self.lookup_lemmatize(token)[0])
+
+        # use lookups, which fall back to the token itself
         if not forms:
-            forms.append(string)
+            forms.append(self.lookup_lemmatize(token)[0])
         forms = list(dict.fromkeys(forms))
         self.cache[cache_key] = forms
         return forms
--- a/spacy/lang/ca/lemmatizer.py
+++ b/spacy/lang/ca/lemmatizer.py
@@ -75,7 +75,7 @@ class CatalanLemmatizer(Lemmatizer):
 
         # use lookups, and fall back to the token itself
         if not forms:
-            forms.append(lookup_table.get(token, [token])[0])
+            forms.append(lookup_table.get(string, [string])[0])
         forms = list(dict.fromkeys(forms))
         self.cache[cache_key] = forms
         return forms
--- a/spacy/lang/fr/lemmatizer.py
+++ b/spacy/lang/fr/lemmatizer.py
@@ -81,7 +81,7 @@ class FrenchLemmatizer(Lemmatizer):
 
         # use lookups, which fall back to the token itself
         if not forms:
-            forms.append(self.lookup_lemmatize(token)[0])
+            forms.append(lookup_table.get(token, [token])[0])
         forms = list(dict.fromkeys(forms))
         self.cache[cache_key] = forms
         return forms
--- a/spacy/lang/fr/lemmatizer.py
+++ b/spacy/lang/fr/lemmatizer.py
@@ -53,11 +53,16 @@ class FrenchLemmatizer(Lemmatizer):
         rules = rules_table.get(univ_pos, [])
         string = string.lower()
         forms = []
+        # first try lookup in table based on upos
         if string in index:
             forms.append(string)
             self.cache[cache_key] = forms
             return forms
+
+        # then add anything in the exceptions table
         forms.extend(exceptions.get(string, []))
+
+        # if nothing found yet, use the rules
         oov_forms = []
         if not forms:
             for old, new in rules:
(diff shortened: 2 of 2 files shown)


adrianeboyd:
The lookup tables are set up to do automatically do the str->int mapping, so I think `token in lookup_table` would have worked instead of `token in lookup_table.keys()`.

I think it would nicer if this did not use `lookup_lemmatize`, which isn't guaranteed to have the same tables. (I mean, we can implement this here so that it works, but in general the idea is that it's independent.)

So instead something like:

```python
if not forms:
    forms.append(lookup_table.get(token, token)[0])
```

polm:
Got it, that makes perfect sense - have changed it to work that way.

adrianeboyd:
I think this should probably be `develop` because it's going to change model output?

PR #11384, opened 2022-08-26
Remove setup_requires from setup.cfg

adrianeboyd:
## Description

Remove `setup_requires` from `setup.cfg`.

`setup_requires` worked in the original `setup.py`-only version but I have never found a version of `setuptools` where it was supported in `setup.cfg`. But since it was possible that there was some old version of `setuptools` that did indeed support this, I kept it around for v3. However I really do not think it is needed at this point, and this reduces the duplication of build requirements specifications. Now they should only go in `pyproject.toml` and `build-constraints.txt`. Install requirements remain in `install_requires` and `requirements.txt` as before.

### Types of change

Chore.

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,14 +31,6 @@ project_urls =
 zip_safe = false
 include_package_data = true
 python_requires = >=3.6
-setup_requires =
-    cython>=0.25,<3.0
-    numpy>=1.15.0
-    # We also need our Cython packages here to compile against
-    cymem>=2.0.2,<2.1.0
-    preshed>=3.0.2,<3.1.0
-    murmurhash>=0.28.0,<1.1.0
-    thinc>=8.1.0,<8.2.0
 install_requires =
     # Our libraries
     spacy-legacy>=3.0.9,<3.1.0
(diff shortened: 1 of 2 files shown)


svlandeg:
Will this disable some use-cases or users installing/compiling via `setup.py`?

adrianeboyd:
No, not really. Since I redid all the packages in Nov. 2020, users calling `python setup.py` have needed to install the build requirements separately first because `setup.py` imports from cython directly. If someone did still happen to be reading in the requirements from `setup.cfg`, they can switch to reading in the exact same requirements from `pyproject.toml` now.

The intention was for `setup_requires` to be deprecated and it got very close to being actually deprecated last year, although I see that they reverted this decision to some degree for `setup.py` (related to dynamic build-time requirements, which doesn't affect us).

The history of `setuptools` is pretty hard to follow, but when I worked on this in 2020 I never found a recent version of `setuptools` that supported `setup_requires` from `setup.cfg` at all. The exact same setup in `setup.py` worked as intended (with our pre-`pyproject.toml` setups), but not from `setup.cfg`. Since we were switching to `pyproject.toml`, I didn't spend a lot of time trying to track it down further.

PR #11385, opened 2022-08-26
Prevent tok2vec to broadcast to listeners when predicting

svlandeg:
## Description
Addressing the same problem as laid out in https://github.com/explosion/spacy-transformers/pull/342, i.e. the fact that `tok2vec.predict` broadcasts the outputs to the listeners, breaks the usual contract in which the training mechanism should rely on that broadcasting, but in prediction we should use `doc.tensor`.

Because we still want to be able to have a frozen (but annotating!) embedding layer, have a fallback in the listener's `forward` function that will access the `doc.tensor` even when training, if there is otherwise no valid data available.

### Types of change
bug fix 

## Checklist


svlandeg:
We might want to move this to `develop`, but I wanted to open this PR first to instigate some internal discussion.

spacy/pipeline/tok2vec.py
```diff
@@ -286,8 +283,20 @@ def verify_inputs(self, inputs) -> bool:
...
-        model.verify_inputs(inputs)
-        return model._outputs, model._backprop
+        # This might occur during training when the tok2vec layer is frozen / hasn't been updated.
+        # In that case, it should be set to "annotating" so we can retrieve the embeddings from the doc.
+        if model._batch_id is None:
+            outputs = []
+            for doc in inputs:
+                if doc.tensor.size == 0:
+                    # TODO move to errors.py
+                    raise ValueError("If the embedding layer is not updated during training, include it in 'annotating components'")
```
kadarakos, line 293: `embedding layer` meaning the `tok2vec` ?

svlandeg: Yea. We can be more specific and say `tok2vec/transformer layer` instead (I want to reuse the error msg for transformers, too).

PR #11390, opened 2022-08-29
Fix `test_{prefer,require}_gpu`

danieldk:
## Description

These tests assumed that GPUs are only supported with CuPy, but since Thinc 8.1 we also support Metal Performance Shaders.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/test_misc.py
+++ b/spacy/tests/test_misc.py
@@ -11,6 +11,7 @@ from spacy.util import dot_to_object, SimpleFrozenList, import_file
 from spacy.util import to_ternary_int
 from thinc.api import Config, Optimizer, ConfigValidationError
 from thinc.api import set_current_ops
+from thinc.compat import has_cupy_gpu, has_torch_mps_gpu
 from spacy.training.batchers import minibatch_by_words
 from spacy.lang.en import English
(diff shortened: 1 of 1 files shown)


danieldk:

spacy/tests/test_misc.py
```diff
@@ -11,14 +11,15 @@
...
+from thinc.compat import has_cupy_gpu, has_torch_mps_gpu
 from spacy.training.batchers import minibatch_by_words
 from spacy.lang.en import English
 from spacy.lang.nl import Dutch
 from spacy.language import DEFAULT_CONFIG_PATH
 from spacy.schemas import ConfigSchemaTraining, TokenPattern, TokenPatternSchema
 from pydantic import ValidationError
 
-from thinc.api import get_current_ops, NumpyOps, CupyOps
+from thinc.api import get_current_ops, NumpyOps, CupyOps, MPSOps
```
svlandeg, line 22: This is not this PR's fault, but can you rearrange the `thinc` imports to be together? We can probably move `set_current_ops` to this list so we cut down one line.

danieldk: Fixed.

PR #11410, opened 2022-08-30
New `_parser_utils` Cython module

shadeMe:
## Description

Move a couple of `argmax` functions used in the parser to a new `_parser_utils` Cython module.

### Types of change

Refactor

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ MOD_NAMES = [
     "spacy.pipeline._parser_internals.stateclass",
     "spacy.pipeline._parser_internals.transition_system",
     "spacy.pipeline._parser_internals._beam_utils",
+    "spacy.pipeline._parser_internals._parser_utils",
     "spacy.tokenizer",
     "spacy.training.align",
     "spacy.training.gold_io",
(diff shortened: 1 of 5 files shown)


danieldk (approved):

danieldk:
@honnibal we were talking about that there were now argmax functions in different parser Cython modules and that'd make sense to have them together (perhaps later other utility functions as well). Does that seems sensible to you?

svlandeg (commented):
These aren't even specific to the `parser`, right? So maybe they could be in some kind of even more generic `spacy.ml._util`?

shadeMe:
> These aren't even specific to the parser, right? So maybe they could be in some kind of even more generic spacy.ml._util?

One of the two functions - `arg_max_if_valid` - takes a `bool` adjacent array parameter that pertains to parser action/move validity. In fact, the main reason we have these two functions is because we need to use `argmax` in the parser's `nogil` context. I'd expect other components that are based on thinc `Model`s to simply use the `Ops` API instead (which is what we ideally want to happen). 

If we moved it something like `spacy.ml._util`, I think we might inadvertently open the door to adding more functions to it than we need to/should (IMO).

PR #11419, opened 2022-08-31
`trainable_lemmatizer` in `debug data`

pmbaumgartner:
This PR adds checks for `trainable_lemmatizer` to `debug data`.

For a test case, I made a branch of the project we used in the [Neural edit-tree lemmatization for spaCy blog post](https://explosion.ai/blog/edit-tree-lemmatizer) that works with spaCy 3.4. After a spaCy 3.4 install, you can get up and running with that example project like this.

```
python -m spacy project clone projects/edit_tree_lemmatizer \
  --repo https://github.com/pmbaumgartner/spacy-experimental \
  --branch edit-tree-lemmatizer-3.4
cd edit_tree_lemmatizer
spacy project assets
spacy project run all

# for only this step
spacy project run debug
```

### Types of change
enhancement

## Checklist


spacy/cli/debug_data.py
```diff
@@ -861,6 +898,20 @@ def _compile_gold(
...
+                    tree_id = trees.add(token.text, token.lemma_)
+                    tree_str = trees.tree_to_str(tree_id)
+                    data["lemmatizer_trees"].add(tree_str)
+    if "trainable_lemmatizer" in factory_names:
+        # After the edittree is built, let's try and capture all the
+        # string trees in that structure
+        all_trees = set()
+        for tree_id in range(len(trees)):
+            all_trees.add(trees.tree_to_str(tree_id))
+        data["lemmatizer_trees_full"] = all_trees
```
pmbaumgartner, line 914: `n=1120`

spacy/cli/debug_data.py
```diff
@@ -861,6 +898,20 @@ def _compile_gold(
                 data["n_nonproj"] += 1
             if nonproj.contains_cycle(aligned_heads):
                 data["n_cycles"] += 1
+        if "trainable_lemmatizer" in factory_names:
+            # NOTE: From EditTreeLemmatizer._labels_from_data
+            for token in eg.reference:
+                if token.lemma != 0:
+                    tree_id = trees.add(token.text, token.lemma_)
+                    tree_str = trees.tree_to_str(tree_id)
+                    data["lemmatizer_trees"].add(tree_str)
```
pmbaumgartner, line 907: `n=565`

spacy/cli/debug_data.py
```diff
@@ -954,6 +1005,20 @@ def _get_labels_from_spancat(nlp: Language) -> Dict[str, Set[str]]:
...
+    pipe_names = [
+        pipe_name
+        for pipe_name in nlp.pipe_names
+        if nlp.get_pipe_meta(pipe_name).factory == "trainable_lemmatizer"
+    ]
+    labels: Set[str] = set()
+    for pipe_name in pipe_names:
+        pipe = nlp.get_pipe(pipe_name)
+        for tree_id in range(len(pipe.trees)):
+            labels.add(pipe.trees.tree_to_str(tree_id))
```
pmbaumgartner, line 1018: `n=510`

(this should also maybe not be called `labels` to avoid confusion since these arent the actual labels, but we'll get to that later`

spacy/cli/debug_data.py
```diff
@@ -670,6 +671,38 @@ def debug_data(
...
+
+        labels = set(label_list)
+        missing_labels = model_labels - labels
+
+        if missing_labels:
+            msg.warn(
+                f"{len(missing_labels)} label(s) in model (component.tree)"
+                " and not present in the train data (trees.add and trees.tree_to_str)."
+            )
+
```
pmbaumgartner, line 705: Current output run on example project in PR description:

```
Output from run on 

```
ℹ 565 label(s) in train data (created from trees.add and
trees.tree_to_str)
ℹ 1120 label(s) in train data (from iterating through tree_id)
ℹ 510 label(s) in component (from iterating through tree_id on
component.trees
⚠ 275 label(s) in model (component.tree) and not present in the train
data (trees.add and trees.tree_to_str).
```

Comment (adrianeboyd) [reply spacy/cli/debug_data.py:705]: These messages/numbers aren't going to meaningful to users, who don't know anything about the internals (and shouldn't need to know anything about the internals).

In general there are a couple different kinds of missing labels to consider:

* you have labels in the dev data that aren't in the train data
* you're updating a sourced model with existing labels and your current training data only includes a subset of the labels, so you'll run into catastrophic forgetting

I think the last line here is looking at yet another kind of missing label, which is edit trees that are omitted for a newly initialized component due to a threshold in the component config, which might not matter at all for coverage of the dev data. It's not that they're not present in the train data (there's nowhere else for these edit trees to have come from), but that they're not included in the initialized component even though they're in the train data. Does that make sense?

Numbers related to the edit tree counts probably aren't very informative for users, but what would be interesting is the number of dev lemmas that can be predicted at all from the current edit trees in the component, which is the first kind of missing label above, similar to a tagger that has `ADJ` in the dev data but not in the train data.

And looking at `debug data` again, I think it's not handling the initialization correctly for sourced components, so anything related to the second kind of missing label isn't working at all right now.

Comment (pmbaumgartner) [reply spacy/cli/debug_data.py:705]: > These messages/numbers aren't going to meaningful to users, who don't know anything about the internals (and shouldn't need to know anything about the internals).

I agree these aren't helpful, I was putting these and other explanatory comments in here as a draft for myself to better understand how these numbers lined up, and would remove them before review.

> It's not that they're not present in the train data (there's nowhere else for these edit trees to have come from), but that they're not included in the initialized component even though they're in the train data. Does that make sense?

I think so. If we've got component labels `CL` and train data labels `TL`, this questions is about `TL - CL`, which might be not empty because of some component config threshold. The set of `CL - TL` with a sourced component results in catastrophic forgetting.

> Numbers related to the edit tree counts probably aren't very informative for users, but what would be interesting is the number of dev lemmas that can be predicted at all from the current edit trees in the component, which is the first kind of missing label above, similar to a tagger that has ADJ in the dev data but not in the train data.
>
> And looking at debug data again, I think it's not handling the initialization correctly for sourced components, so anything related to the second kind of missing label isn't working at all right now.

It seems like we should prioritize the first kind then, and ignore the second kind for now?

Comment (pmbaumgartner) [reply spacy/cli/debug_data.py:705]: > which is the first kind of missing label above, similar to a tagger that has `ADJ` in the dev data but not in the train data.

We do this with `deps`, but not for `tags`, right? https://github.com/explosion/spaCy/pull/11419/files#diff-237bbf8610a0a5f63114db5e097db1325b2f4c11c1a8f79f9e20b41e6ec0e8cbR631-R645

File: spacy/cli/debug_data.py (hunk 1 above)
Comment (svlandeg) [reply spacy/cli/debug_data.py]: What do these review comments refer to?

Comment (pmbaumgartner) [reply spacy/cli/debug_data.py]: I was noting the sizes of the various trees that were being generated while this was in draft, so it was easier to refer to them when reviewing. They can be closed.

File: spacy/cli/debug_data.py [hunk 5]
```diff
@@ -670,6 +671,40 @@ def debug_data(
...
+            # Would we ever expect this case? It seems like it would be pretty rare,
+            # and we might actually want a warning?
+            msg.good("All trees in dev data present in training data.")
+
+        if gold_train_data["n_low_cardinality_lemmas"] > 0:
+            msg.warn(
+                f"{gold_train_data['n_low_cardinality_lemmas']} docs with 1 or 0 unique lemmas."
+            )
+        else:
+            msg.good("All training docs meet lemma uniqueness requirements.")
```
adrianeboyd, line 699: I don't think users are going to understand "meet lemma uniqueness requirements". I might just leave this part out for the good case?

spacy/cli/debug_data.py
```diff
@@ -670,6 +671,40 @@ def debug_data(
...
+
+        # This is necessary context when someone is attempting to interpret whether the
+        # number of trees exclusively in the dev set is meaningful.
+        msg.info(f"{len(trees_train)} lemmatizer trees generated from training data.")
+        msg.info(f"{len(trees_dev)} lemmatizer trees generated from dev data.")
+
+        dev_not_train = trees_dev - trees_train
+        if len(dev_not_train) != 0:
+            msg.warn(
+                f"{len(dev_not_train)} lemmatizer trees were found exclusively in the dev data."
```
adrianeboyd, line 687: This is pretty much always going to be the case, so I'm not sure this warning is going to be helpful. Is there a threshold of cases that might make more sense, like 5% of edit trees or something?

spacy/cli/debug_data.py
pmbaumgartner: I like the idea to leave it out, I'll do that. I had a hard time wording it in a succinct but descriptive way anyway.

spacy/cli/debug_data.py
pmbaumgartner: Maybe @danieldk has an opinion on a threshold. Based on my understanding, a threshold without any context is difficult to make sense of. Here's an example he gave from our discussion:

> As an example, suppose that 100 of the trees in the dev set are not in the train set, it doesn’t tell much yet. It could be that those 100 trees are unique because they are all derived from misspelled words and consequently only 100 words could not be lemmatized. Or it could be that those 100 trees are fairly frequent and 1000 words could not be lemmatized.

Maybe the thing to do here is to switch this to an `info` rather than `warn`, I do think it's information a user would want but you're right in that it's almost always going to be triggered.

adrianeboyd: I think at the very least the numbers need more context, because a count alone isn't going to mean much, while something like a percentage (even if it's not a perfect measure) might put it more in perspective. Or something like the percentage of words in the dev set that are affected by missing trees?

danieldk: I agree that it makes most sense to make this an info message. Hard to say what would be a good threshold without knowing more about the underlying properties of the data.

spacy/cli/debug_data.py
```diff
@@ -670,6 +671,40 @@ def debug_data(
...
+            msg.good("All trees in dev data present in training data.")
+
+        if gold_train_data["n_low_cardinality_lemmas"] > 0:
+            msg.warn(
+                f"{gold_train_data['n_low_cardinality_lemmas']} docs with 1 or 0 unique lemmas."
+            )
+        else:
+            msg.good("All training docs meet lemma uniqueness requirements.")
+
+        if gold_train_data["n_low_cardinality_lemmas"] > 0:
```
danieldk, line 701:
```suggestion
        if gold_dev_data["n_low_cardinality_lemmas"] > 0:
```

I think?

spacy/cli/debug_data.py
```diff
@@ -670,6 +671,40 @@ def debug_data(
...
+                f"{len(dev_not_train)} lemmatizer trees were found exclusively in the dev data."
+            )
+        else:
+            # Would we ever expect this case? It seems like it would be pretty rare,
+            # and we might actually want a warning?
+            msg.good("All trees in dev data present in training data.")
+
+        if gold_train_data["n_low_cardinality_lemmas"] > 0:
+            msg.warn(
+                f"{gold_train_data['n_low_cardinality_lemmas']} docs with 1 or 0 unique lemmas."
```
danieldk, line 696: This reports if a doc is not annotated at all. But I guess another possibility is that only a subset of tokens are annotated, which wouldn't be reported now, as long as there are enough (unique) annotations in the document. Maybe it's worthwhile to report the number of tokens without a lemma annotation?

adrianeboyd: Usually our goal is to support training from partial annotation, so this isn't a check that we do for other components or that we necessarily want to do here. I think this intended to be parallel to "we found 47 tags in the training data" as a check for training data that's not going to work for this component at all. (Does the edit tree lemmatizer skip unlabeled examples?)

danieldk: Ah, right.

> Does the edit tree lemmatizer skip unlabeled examples?

It does:

https://github.com/explosion/spaCy/blob/6b78135b9e158e5bc02e39c1a73ef28bb360a44f/spacy/pipeline/edit_tree_lemmatizer.py#L131

adrianeboyd: Okay, then I think checking for absolutely no annotation (usually a data conversion issue) is what we want here.

spacy/cli/debug_data.py
adrianeboyd: I think some form of relative numbers would be good here (percentage of trees if that's all we have, or better percentage of dev tokens).

spacy/cli/debug_data.py
pmbaumgartner: Just to clarify: Do we want to change this to check how many documents have one or more `None` values as the `lemma` attribute? Or to show how many lemmas across all documents are without an annotation?

spacy/cli/debug_data.py
pmbaumgartner: Here's what this looks like as % of trees:

```
ℹ 105 lemmatizer trees (18.6% of dev trees) were found exclusively in
the dev data.
```

In implementing a percentage of words measure, I'm encountering an discrepancy calculating the unique trees and I want to confirm the correct way. (The number of trees above I believe is incorrect)

Right now I'm calculating the string representation of each tree and adding that to a `set` as well as storing the tree strings into a dictionary with the `token.text`. The idea is that if we want to know what % of dev tokens don't have trees, we can map back to `data["words"]`, which is a Counter of words, and sum those values for the keys that are in `dev_not_train`

It looks something like this -

```python
def debug_data(
        ...
        tree_id = trees.add(token.text, token.lemma_)
        tree_str = trees.tree_to_str(tree_id)
        data["lemmatizer_trees_dict"][token.text] = tree_str
        data["lemmatizer_trees_set"].add(tree_str)
```

However, when I do this, I get different values for `len(set(data["lemmatizer_trees_dict"].values()))` and `len(data["lemmatizer_trees_set"])`. Specifically on the demo data I mention in the parent comment:

```
ℹ 2298 lemmatizer trees generated from training data. (dict)
ℹ 528 lemmatizer trees generated from dev data. (dict)
ℹ 2709 lemmatizer trees generated from training data. (set)
ℹ 565 lemmatizer trees generated from dev data. (set)
```

So the `set` always ends up with more. I'm thinking this is because the trees & subtress are recursively updating over time so the string representations change, so at `t1` the tree for `word` is going to be different than at `t2`, but in the case of using `set.add` the tree at `t1` and `t2` are incorrectly added when all we really care about is the value at `t_final` (`t2` in this dummy example). 

Is the `dict` approach right @danieldk?

spacy/cli/debug_data.py
svlandeg: @pmbaumgartner : is this done? If so - can you resolve?

spacy/cli/debug_data.py
danieldk: The dictionary approach is incorrect because the same surface form may occur with different parts of speech and thus different lemmas. E.g.:

_Hij **dicht** de de scheur._ (He fixes the crack) -> POS: verb, lemma: dichten
_De deur is **dicht**._ (The door is closed) -> POS: adjective, lemma: dicht

If you'd key the dictionary by surface form and part-of-speech, the count should be much closer to the set count. But I am not sure if that is ok for `debug data`, since the corpus may not actually have POS annotations.

spacy/cli/debug_data.py
```diff
@@ -861,6 +900,20 @@ def _compile_gold(
                 data["n_nonproj"] += 1
             if nonproj.contains_cycle(aligned_heads):
                 data["n_cycles"] += 1
+        if "trainable_lemmatizer" in factory_names:
+            # from EditTreeLemmatizer._labels_from_data
+            lemma_set = set()
+            for token in gold:
+                lemma_set.add(token.lemma)
```
danieldk, line 907: Seems more straightforward to add to the set after checking that `lemma != 0` and then check:

```
if len(lemma_set) == 0 and len(gold) > 1:
```

below?

pmbaumgartner: I agree that is more straightforward. We'd lose the debug case where every token in a document has the same (non-`0`) lemma, but I'm not sure how often that would actually happen in practice? I think the first part about not adding to the set if it is `0` is still a good improvement though.

danieldk: Ah right. Maybe the condition can be left as is (just moving the `add`), then both cases are covered.

spacy/cli/debug_data.py
pmbaumgartner: I think we should just take the `set` approach then, as I don't think we want to depend on POS annotations, and we can compare sets between training and dev and display the percent.

spacy/cli/debug_data.py
pmbaumgartner: I've updated it to check if all lemmas are `0` for a doc and if they are, increase a counter by 1 (and do not loop through the tokens and collect the string tree representations). 

@danieldk A question on this: if a doc doesn't have an annotation is the value `None` or `0`? In the lemmatizer code linked above, it checks whether the attribute is `None`, but if I create a doc with without lemmas, they're all `0`. Is it something different with annotations that they have a value of `None` when they don't exist? Either way, I'm not clear what we need to check on here (`None` or `0`).

```python
import spacy

nlp = spacy.blank("en")
doc = nlp("This is a document to serialize.")
print([t.lemma for t in doc])
# [0, 0, 0, 0, 0, 0, 0]
```

danieldk: > @danieldk A question on this: if a doc doesn't have an annotation is the value `None` or `0`? In the lemmatizer code linked above, it checks whether the attribute is `None`, but if I create a doc with without lemmas, they're all `0`. Is it something different with annotations that they have a value of `None` when they don't exist? Either way, I'm not clear what we need to check on here (`None` or `0`).

We should use `Token.lemma` for checking absence and when a lemma is absent, the value is `0`. It cannot be `None`, since `Token.lemma` just passes through the C representation, which is integral:

https://github.com/explosion/spaCy/blob/8e558095a1223fcb81e0abeda46e1e6975729254/spacy/structs.pxd#L44

(`attr_t` is an `uint64_t`).

I think that the fragment that I linked for the edit tree lemmatizer is incomplete, since `get_aligned` returns `None` for failed alignments, but empty strings for absent lemma annotations (like `Token.lemma_` would). I'll open a PR for that.

spacy/cli/debug_data.py
```diff
@@ -670,6 +671,51 @@ def debug_data(
...
+            n = gold_dev_data["n_low_cardinality_lemmas"]
+            msg.warn(f"{n} docs with 1 or 0 unique lemmas.")
+        else:
+            msg.good("All dev docs meet lemma uniqueness requirements.")
+
+        if gold_train_data["no_lemma_annotations"] > 0:
+            n = gold_train_data["no_lemma_annotations"]
+            msg.warn(f"{n} docs with no lemma annotations.")
+        else:
+            msg.good("All training docs have complete lemma annotations.")
```
danieldk, line 711:
```suggestion
            msg.good("All training docs have lemma annotations.")
```

_complete_ suggests that every token is annotated. We'd still allow partial annotations.

spacy/cli/debug_data.py
```diff
@@ -670,6 +671,51 @@ def debug_data(
...
+
+        if gold_dev_data["no_lemma_annotations"] > 0:
+            n = gold_dev_data["no_lemma_annotations"]
+            msg.warn(f"{n} docs with no lemma annotations.")
+        else:
+            msg.good("All dev docs have complete lemma annotations.")
```
danieldk, line 717:
```suggestion
            msg.good("All dev docs have lemma annotations.")
```

spacy/cli/debug_data.py
pmbaumgartner: Ah, the check I currently have is `all(token.lemma == 0 for token in gold)` for this, so I think the current message makes sense. But I'll also add some lines for identifying the number of docs with partial annotations. That would be the similar logic but switching `all` to `any`, then I think changing the message when `count > 0` from a `warn` to an `info`.

pmbaumgartner: Updated in latest commit.

spacy/cli/debug_data.py
pmbaumgartner: Resolved in latest commit, which distinguishes the complete and partial annotation cases and now has a separate counter for both. The message is a warn if there are a non-zero amount of docs with absolutely no annotations, and an info for the number of docs with partial annotations (and a good if both are 0).

svlandeg:
Does this PR still require changes to the documentation?

svlandeg:
> And on the topic of testing, I don't think we have tests for some specific components in debug data (spans) but not others (textcat), but that would be a good next step in in the initiative to add more tests for CLI commands.

I'm missing the context of this remark, but in general I want to clarify that I do think we should add tests for new functionality whenever possible. The outstanding internal task of adding more tests to CLI commands concerns fixing technical debt that we've let build up too long - instead we need to get into the habit of writing tests, also for CLI commands, together with the actual code (rather than letting the technical debt grow larger).

pmbaumgartner:
> Does this PR still require changes to the documentation?

I don't think we mention what components are supported in `debug data` and I'm not sure what would need to be changed. Is there somewhere you think we should add documentation about this?

pmbaumgartner:
> > And on the topic of testing, I don't think we have tests for some specific components in debug data (spans) but not others (textcat), but that would be a good next step in in the initiative to add more tests for CLI commands.
> 
> I'm missing the context of this remark, but in general I want to clarify that I do think we should add tests for new functionality whenever possible. The outstanding internal task of adding more tests to CLI commands concerns fixing technical debt that we've let build up too long - instead we need to get into the habit of writing tests, also for CLI commands, together with the actual code (rather than letting the technical debt grow larger).

I agree, I'll add tests. There should be some examples I can use in the existing trainable_lemmatizer tests in order to generate a dataset for testing this out.

svlandeg:
> > Does this PR still require changes to the documentation?
> 
> I don't think we mention what components are supported in `debug data` and I'm not sure what would need to be changed. Is there somewhere you think we should add documentation about this?

That's mainly what I wanted to check. If it doesn't need documentation - if there is nothing relevant to update - it'd be good to back to the original PR description and get that box checked :-)

svlandeg:
> > > And on the topic of testing, I don't think we have tests for some specific components in debug data (spans) but not others (textcat), but that would be a good next step in in the initiative to add more tests for CLI commands.
> > 
> > 
> > I'm missing the context of this remark, but in general I want to clarify that I do think we should add tests for new functionality whenever possible. The outstanding internal task of adding more tests to CLI commands concerns fixing technical debt that we've let build up too long - instead we need to get into the habit of writing tests, also for CLI commands, together with the actual code (rather than letting the technical debt grow larger).
> 
> I agree, I'll add tests. There should be some examples I can use in the existing trainable_lemmatizer tests in order to generate a dataset for testing this out.

Awesome, thanks! I'll review these early next week :-)

spacy/cli/debug_data.py
```diff
@@ -671,6 +672,63 @@ def debug_data(
...
+        if len(dev_not_train) != 0:
+            pct = len(dev_not_train) / len(trees_dev)
+            msg.info(
+                f"{len(dev_not_train)} lemmatizer trees ({pct*100:.1f}% of dev trees)"
+                " were found exclusively in the dev data."
+            )
+        else:
+            # Would we ever expect this case? It seems like it would be pretty rare,
+            # and we might actually want a warning?
+            msg.good("All trees in dev data present in training data.")
```
svlandeg, line 694: Maybe switch this to `info`?

spacy/tests/test_cli_app.py
```diff
@@ -40,3 +42,52 @@ def test_benchmark_accuracy_alias():
...
+        train_bin = DocBin(docs=train_docs)
+        train_bin.to_disk(d_in / "train.spacy")
+        dev_bin = DocBin(docs=dev_docs)
+        dev_bin.to_disk(d_in / "dev.spacy")
+        # `debug data` requires an input pipeline config
+        CliRunner().invoke(
+            app,
+            [
+                "init",
+                "config",
```
svlandeg, line 71: I really like this test and the fact that it calls into `spacy init config` with `trainable_lemmatizer`, ensuring that workflow works well :-)

svlandeg (commented):
I only had a few small comments. If Adriane and Daniël agree, I think this should otherwise be in a good shape to merge soon :-)

spacy/cli/debug_data.py
pmbaumgartner: Yeah, that's a good call, I don't think this is actually "good" if this happens.

spacy/cli/debug_data.py
pmbaumgartner: Whoops, sorry - I totally glanced over this when revisiting this. I still agree it's a bit confusing and a check for a rare case that would require quite a bit of explanation to a user to be useful. I removed this functionality and the corresponding test.

svlandeg: I could be mistaken, but I think @adrianeboyd was actually proposing to leave the warning checks in but not print anything for the `good` case?

pmbaumgartner: `:me_slapping_forehead:` Let me undo that commit.

spacy/tests/test_cli.py
```diff
@@ -1207,3 +1207,65 @@ def test_walk_directory():
         assert (len(walk_directory(d, suffix="iob"))) == 2
         assert (len(walk_directory(d, suffix="conll"))) == 3
         assert (len(walk_directory(d, suffix="pdf"))) == 0
+
```
adrianeboyd, line 1210: Please reformat this!

spacy/tests/test_cli_app.py
```diff
@@ -40,3 +42,52 @@ def test_benchmark_accuracy_alias():
     assert result_benchmark.stdout == result_evaluate.stdout.replace(
         "spacy evaluate", "spacy benchmark accuracy"
     )
+
+
+def test_debug_data_trainable_lemmatizer_cli():
+    nlp = English()
```
adrianeboyd, line 48: Minor, but you can use the fixture `en_vocab` instead for cases like this and it's slightly faster.

adrianeboyd (commented):
All the recent adjustments look good, just a few remaining details...

spacy/tests/test_cli_app.py
pmbaumgartner: Thanks for the suggestion, I wasn't aware we had a fixture for that. I've updated to use that instead.

spacy/cli/debug_data.py
```diff
@@ -671,6 +672,59 @@ def debug_data(
...
+                " were found exclusively in the dev data."
+            )
+        else:
+            # Would we ever expect this case? It seems like it would be pretty rare,
+            # and we might actually want a warning?
+            msg.info("All trees in dev data present in training data.")
+
+        if gold_train_data["n_low_cardinality_lemmas"] > 0:
+            n = gold_train_data["n_low_cardinality_lemmas"]
+            msg.warn(f"{n} docs with 1 or 0 unique lemmas.")
```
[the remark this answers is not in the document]
pmbaumgartner: I've updated the messages to clarify which dataset the warning is about.

spacy/tests/test_cli.py
pmbaumgartner: Done! Something's been off with my "format on save" I have to debug.

svlandeg (approved):
Nice PR! I'll get it merged when the tests go green :-)

PR #11424, opened 2022-09-01
Check for any non-Doc returned value for components

adrianeboyd:
## Description

Check for any non-`Doc` returned value for components.

### Types of change

?

## Checklist


svlandeg (approved):
Makes sense!

PR #11429, opened 2022-09-05
Custom extensions for spans with equal boundaries

thomashacker:
## Description

This `PR` is the continuation of `PR` #9708 and aims to fix `Issue` #9706.
Big thanks to @nrodnova for working on the previous `PR` and creating the `Issue` ✨ 

At the moment, the keys of the underscore class only contain the `start_char` and `end_char` which makes it impossible to save custom attributes to spans with the same boundaries but with different `labels`/`kb_id`/`id` because they're overwriting each other. This `PR` adds three new keys to the underscore class for `Span` objects to make spans with different `labels`/`kb_id`/`id` unique.

### Types of change

Bug fix

## Checklist


spacy/tokens/underscore.py
```diff
@@ -25,6 +25,9 @@ def __init__(
         obj: Union["Doc", "Span", "Token"],
         start: Optional[int] = None,
         end: Optional[int] = None,
+        label: Optional[Union[str, int]] = None,
+        kb_id: Optional[Union[str, int]] = None,
+        span_id: Optional[Union[str, int]] = None,
```
adrianeboyd, line 30: I think this should only support `int` so that you don't have the situation where equivalent `str` and `int` representations look like different tuples. (I'm also on the fence about `Optional`. It might be easier/faster to have it be `0` in cases where it doesn't matter.)

spacy/tokens/underscore.py
```diff
@@ -36,6 +39,13 @@ def __init__(
         object.__setattr__(self, "_doc", obj.doc)
         object.__setattr__(self, "_start", start)
         object.__setattr__(self, "_end", end)
+        if label is not None or kb_id is not None or span_id is not None:
```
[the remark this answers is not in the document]
thomashacker: Yes, definitely! It also improves the readability of the code

spacy/tokens/underscore.py
thomashacker: I've change it to `int = 0`

adrianeboyd:
There are a number of places where this data is accessed directly in various places in the library that haven't been updated accordingly, see: https://github.com/explosion/spaCy/pull/9708#issuecomment-1018276846

spacy/tests/doc/test_underscore.py
```diff
@@ -171,3 +171,343 @@ def test_method(doc, arg1=1, arg2=2):
...
+                "span_extension",
+                span_1.start_char,
+                span_1.end_char,
+                span_1.label,
+                span_1.kb_id,
+                span_1.id,
+            )
+        ]
+        == "span_1 extension"
+    )
```
svlandeg, line 208: To make these tests a bit more compact & readible, I suggest we introduce a helper function in this file:
```
def _get_tuple(s: Span):
    return "._.", "span_extension", s.start_char, s.end_char, s.label, s.kb_id, s.id
```
after which all `assert` statements can be just one line:
```
assert doc.user_data[_get_tuple(span_1)] == "span_1 extension"
```
This will also be useful for the other unit tests.

spacy/tokens/underscore.py
```diff
@@ -88,8 +95,46 @@ def get(self, name: str) -> Any:
...
+                "._.",
+                name,
+                self._start,
+                self._end,
+                self._label,
+                self._kb_id,
+                self._span_id,
+            )
+        else:
+            return ("._.", name, self._start, self._end)
```
svlandeg, line 123:
```suggestion
            return "._.", name, self._start, self._end
```

spacy/tokens/underscore.py
```diff
@@ -88,8 +95,46 @@ def get(self, name: str) -> Any:
...
+
+    @staticmethod
+    def _replace_keys(old_underscore, new_underscore):
```
svlandeg, line 126:
```suggestion
    def _replace_keys(old_underscore: "Underscore", new_underscore: "Underscore"):
```

spacy/tokens/underscore.py
```diff
@@ -88,8 +95,46 @@ def get(self, name: str) -> Any:
...
+        """
+        This function is called by Span when its kb_id or label are re-assigned.
+        It checks if any user_data is stored for this span and replaces the keys
+        """
+        for name in old_underscore._extensions:
+            old_key = old_underscore._get_key(name)
+            new_key = new_underscore._get_key(name)
```
svlandeg, line 133: To shorten the following lines:
```
old_doc = old_underscore._doc
```

spacy/tokens/underscore.py
```diff
@@ -88,8 +95,46 @@ def get(self, name: str) -> Any:
...
+            if old_key in old_underscore._doc.user_data and old_key != new_key:
```
svlandeg, line 134: Do the `old_key != new_key` comparison first in this `if` statement, as it's going to be quicker than the `in` statement, and as soon as the first condition is `False`, the second one won't be evaluated anymore.

spacy/tokens/span.pyx
```diff
@@ -275,12 +274,20 @@ cdef class Span:
...
                         start -= char_offset
                         if end is not None:
                             end -= char_offset
-                        user_data[(data_type, name, start, end)] = copy.copy(value)
+                            _label = key[4]
+                            _kb_id = key[5]
+                            _span_id = key[6]
+                            user_data[(data_type, name, start, end, _label, _kb_id, _span_id)] = copy.copy(value)
+                        else:
+                            user_data[(data_type, name, start, end)] = copy.copy(value)
```
svlandeg, line 290: I'm slightly confused by these nested `if`- `else` statements. Why are we reading the label/ID data conditional on the `end` not being `None`? At the very least, we should add some comments here.

spacy/tokens/span.pyx
```diff
@@ -791,21 +798,42 @@ cdef class Span:
...
         def __set__(self, attr_t label):
-            self.span_c().label = label
+            if label != self.span_c().label :
+                old_label = self.span_c().label
+                self.span_c().label = label
+                new = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label = self.label, kb_id = self.kb_id, span_id=self.id)
+                old = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label=old_label, kb_id=self.kb_id, span_id=self.id)
+                Underscore._replace_keys(old, new)
+            else:
+                self.span_c().label = label
```
svlandeg, line 808: What's the use of this `else` statement if they were already equal?

spacy/tokens/span.pyx
```diff
@@ -791,21 +798,42 @@ cdef class Span:
...
         def __set__(self, attr_t kb_id):
-            self.span_c().kb_id = kb_id
+            if kb_id != self.span_c().kb_id :
+                old_kb_id = self.span_c().kb_id
+                self.span_c().kb_id = kb_id
+                new = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label=self.label, kb_id=self.kb_id, span_id=self.id)
+                old = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label=self.label, kb_id=old_kb_id, span_id=self.id)
+                Underscore._replace_keys(old, new)
+            else:
+                self.span_c().kb_id = kb_id
```
svlandeg, line 822: Same question: is this `else` necessary?

spacy/tokens/span.pyx
```diff
@@ -791,21 +798,42 @@ cdef class Span:
...
         def __set__(self, attr_t id):
-            self.span_c().id = id
+            if id != self.span_c().id :
+                old_id = self.span_c().id
+                self.span_c().id = id
+                new = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label=self.label, kb_id=self.kb_id, span_id=self.id)
+                old = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label=self.label, kb_id=self.kb_id, span_id=old_id)
+                Underscore._replace_keys(old, new)
+            else:
+                self.span_c().id = id
```
svlandeg, line 836: Same question: is this `else` necessary?

spacy/tokens/span.pyx
```diff
@@ -791,21 +798,42 @@ cdef class Span:
             return self.span_c().label
 
         def __set__(self, attr_t label):
-            self.span_c().label = label
+            if label != self.span_c().label :
+                old_label = self.span_c().label
+                self.span_c().label = label
+                new = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label = self.label, kb_id = self.kb_id, span_id=self.id)
```
svlandeg, line 804:
```suggestion
                new = Underscore(Underscore.span_extensions, self, start=self.span_c().start_char, end=self.span_c().end_char, label=self.label, kb_id=self.kb_id, span_id=self.id)
```

spacy/tokens/underscore.py
```diff
@@ -88,8 +95,46 @@ def get(self, name: str) -> Any:
...
+        Tuple[
+            str,
+            str,
+            Optional[int],
+            Optional[int],
+            Optional[Union[str, int]],
+            Optional[Union[str, int]],
+            Optional[Union[str, int]],
+        ],
+    ]:
```
svlandeg, line 111: As the types have changed to just `int`, this can be (luckily) shortened quite a bit:
```suggestion
    ) -> Union[
        Tuple[str, str, Optional[int], Optional[int]],
        Tuple[str, str, Optional[int], Optional[int], int, int, int],
    ]:
```

svlandeg (commented):
Nice work Edi! I had some comments, mainly around readibility of the code.
The functionality itself seems to work fine 👍

spacy/tokens/span.pyx
thomashacker: With `end not None` we ensure that it's a `Span` object because it has a `start` and an `end` position. If `end == None` then we know it's a Token. I'll add comments! 😄

spacy/tokens/span.pyx
thomashacker: Yes, you're right it doesn't make sense! I'll remove them

thomashacker:
Thanks for the great review @svlandeg ! 😇 I hopefully managed to implement all things you mentioned, assured that the formatting is correct, and fixed the test issues.

svlandeg (approved):
This looks good to me, nice work!

PR #11432, opened 2022-09-05
Replace Pipe with Callable in Language

adrianeboyd:
## Description

Replace `Pipe` with `Callable` in `Language`.

### Types of change

?

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1,6 +1,6 @@
 from typing import Iterator, Optional, Any, Dict, Callable, Iterable, Collection
 from typing import Union, Tuple, List, Set, Pattern, Sequence
-from typing import NoReturn, TYPE_CHECKING, TypeVar, cast, overload
+from typing import NoReturn, TypeVar, cast, overload
 
 from dataclasses import dataclass
 import random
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Well, this reminded me why we're using `Pipe` here. It looks like mypy may support `hasattr` soon, so I think once that is supported, it would make sense to switch to the correct type `Callable` instead of `Pipe` in `Language.

spacy/language.py
```diff
@@ -180,7 +176,7 @@ def __init__(
         self.vocab: Vocab = vocab
         if self.lang is None:
             self.lang = self.vocab.lang
-        self._components: List[Tuple[str, "Pipe"]] = []
+        self._components: List[Tuple[str, Callable]] = []
```
rmitsch, line 179: The `Callable` would have to return a `Doc`, right? Should we type this as `Callable[..., Doc]` then?

adrianeboyd: That's a good point. Then it would be more accurate as: `Callable[[Doc], Doc]`

robookwus:
Hey I'd like to help with the docs, or create my own version, how do I proceed. I emailed the developers and it was OK'd. Still a GitHub greenhorn but enjoy technical writing. Thx

adrianeboyd:
@robookwus: Welcome! This particular thread is for a pull request that's not related to the docs.

Instead, please open a new issue (https://github.com/explosion/spaCy/issues/new?assignees=&labels=&template=02_docs.md) with more details about what you're interested in working on?

adrianeboyd:
Replaced by #11803.

PR #11437, opened 2022-09-06
Sync `develop` with `master`

shadeMe:
## Description

Sync branches to aid #11344, which depends on #11312 (to fix test breakage).

A follow-up PR will merge `develop` into `v4`.

### Types of change

Chore

## Checklist


adrianeboyd:
I think `master` was just synced to `develop`, so it would be better to do `master` -> `develop` + `develop`->`v4` (otherwise you end up having to do 3 PRs instead of 2, as I did unintentionally a few weeks ago).

svlandeg:
Ah, good point Adriane, but it looks like `develop` is again one commit behind `master` 😭

shadeMe:
Oh, I missed the suggestion to open new PRs. Closing this.

adrianeboyd:
Usually we do this by merging into branches in forks instead of working directly with the main repo branches. It spares some of the weird interactions with main repo branches once you get into situations where the merge is not trivial.

PR #11453, opened 2022-09-07
Add way to lower in rule_lemmatizer

jademlc:
The special lemmatization will lowercase the whole text, even proper nouns

## Description

Added a function to define a special_lemmatization that returns `text.lower()` even when the token is detected as a proper nouns. This change is useful when using spaCy Matchers with noisy texts containing a lot of uppercases. This was discussed in this issue: [https://github.com/explosion/spaCy/discussions/11051](https://github.com/explosion/spaCy/discussions/11051)
Using the `special` lemmatizer, the lemma of 'CAT' will be 'cat' instead of 'CAT'.

### Types of change

This change is a enhancement of the lemmatizer.

## Checklist

### DIFF
--- a/spacy/pipeline/lemmatizer.py
+++ b/spacy/pipeline/lemmatizer.py
@@ -106,6 +106,8 @@ class Lemmatizer(Pipe):
             self.lemmatize = self.lookup_lemmatize
         elif self.mode == "rule":
             self.lemmatize = self.rule_lemmatize
+        elif self.mode == "rule_lower":
+            self.lemmatize = self.rule_lower_lemmatize
         else:
             mode_attr = f"{self.mode}_lemmatize"
             if not hasattr(self, mode_attr):
@@ -193,14 +195,7 @@ class Lemmatizer(Pipe):
             result = [result]
         return result
 
-    def rule_lemmatize(self, token: Token) -> List[str]:
-        """Lemmatize using a rule-based approach.
-
-        token (Token): The token to lemmatize.
-        RETURNS (list): The available lemmas for the string.
-
-        DOCS: https://spacy.io/api/lemmatizer#rule_lemmatize
-        """
+    def __rule_lemmatize__(self, token: Token, tolower: bool = False):
         cache_key = (token.orth, token.pos, token.morph.key)  # type: ignore[attr-defined]
         if cache_key in self.cache:
             return self.cache[cache_key]
@@ -223,7 +218,7 @@ class Lemmatizer(Pipe):
                 rules_table.get(univ_pos),
             )
         ):
-            if univ_pos == "propn":
+            if not tolower and univ_pos == "propn":
                 return [string]
             else:
                 return [string.lower()]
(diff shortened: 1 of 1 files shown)


databill86:
This could be very useful! thank you. I hope it will get merged soon.

hetpin:
Very interesting, and should be merged after passing all tests.

adrianeboyd:
There are a number of ways to approach this, so it's useful to hear from everyone who is interested in this!

Can I ask some questions about what kind of output you're looking for?

Currently, you have situations like this where tagging inconsistencies between common and proper nouns make it hard to match on lemmas:

`cat`/`NOUN`/`Number=Sing` -> `cat`
`cats`/`NOUN`/`Number=Plur` -> `cat`
`Cat`/`NOUN`/`Number=Sing` -> `cat`
`Cat`/`PROPN`/`Number=Sing` -> `Cat`
`Cats`/`NOUN`/`Number=Plur` -> `cat`
`Cats`/`PROPN`/`Number=Plur` -> `Cats`
`CAT`/`NOUN`/`Number=Sing` -> `cat`
`CAT`/`PROPN`/`Number=Sing` -> `CAT`
`CATS`/`NOUN`/`Number=Plur` -> `cat`
`CATS`/`PROPN`/`Number=Plur` -> `CATS`

This proposal would lowercase the uppercase lemmas for proper nouns but not make any other changes, so you'd end up with:

`Cat`/`PROPN`/`Number=Sing` -> `cat`
`Cats`/`PROPN`/`Number=Plur` -> `cats`
`CAT`/`PROPN`/`Number=Sing` -> `cat`
`CATS`/`PROPN`/`Number=Plur` -> `cats`

Which would still mean you'd be looking at matching both `cat` and `cats` as lemmas in some cases.

I could imagine options that both lowercase and then treat proper nouns as common nouns to get `cat` in all cases, but it could be that this isn't what anyone is actually looking for? There are obviously tricky cases like "Cats are soft. Cats is a musical." since the tagger is unlikely to be very accurate.

I'm interested to hear the details about your hands-on practical problems with matching on lemmas?

jademlc:
Thanks for your answer.

We've offered this solution to decrease the number of potential lemmas we would need to put on our Matchers, so now instead multiple forms we would need to add two.
We are dealing with a lot of noisy data containing a lot of uppercases even when they are not needed and need to have Matchers that work on them with minimum patterns.
That's why for us treating proper nouns as nouns would be particularly useful. If you tell us that it is something possible, we would love to hear the solution for that.
We are aware that this could cause issues like the one you described but we would rather have these issues than not matching something we should.

adrianeboyd:
If you want to treat proper nouns as nouns and you don't care whether you preserve the original POS tags in the doc, you can modify the POS before lemmatization by adding rules in the attribute ruler.

```python
import spacy
from spacy.tokens import Doc

ruler = nlp.get_pipe("attribute_ruler")
patterns = [[{"TAG": {"IN": ["NNP", "NNPS"]}}]]
attrs = {"POS": "NOUN"}
ruler.add(patterns=patterns, attrs=attrs)
doc = Doc(nlp.vocab, words=["cats"], tags=["NNPS"])
print([(t.tag_, t.morph, t.pos_, t.lemma_) for t in nlp(doc)])

To be honest the singular/plural distinction in the trained models for all caps tokens is not great, so I can't come up with a good short example here that is actually tagged by the model as `NNPS` to show that this works, so I created a doc by hand above. But by converting `PROPN` to `NOUN` you should at least get lowercase lemmas everywhere even if it does include both `cat` and `cats`.

A kind of brute force plural treatment for `PROPN` tokens that end in `s` could look like this:

```python
ruler = nlp.get_pipe("attribute_ruler")
# mark all proper nouns that end in s as plural common nouns
patterns = [[{"LOWER": {"REGEX": "s$"}, "TAG": {"IN": ["NNP", "NNPS"]}}]]
attrs = {"POS": "NOUN", "MORPH": "Number=Plur"}
ruler.add(patterns=patterns, attrs=attrs)
# mark any remaining proper nouns as common nouns
patterns = [[{"TAG": {"IN": ["NNP", "NNPS"]}}]]
attrs = {"POS": "NOUN"}
ruler.add(patterns=patterns, attrs=attrs)
doc = Doc(nlp.vocab, words=["cats"], tags=["NNPS"])
print([(t.tag_, t.morph, t.pos_, t.lemma_) for t in nlp(doc)])
```

jademlc:
Thanks for the different options. 

We tested them and unfortunately it won't be the best option for our usecase as we need the POS-tag information of proper nouns. We thought it would be possible to have something in the code to treat proper nouns the same as nouns without changing the POS-tag. 

In conclusion, the best solution for us is to change the lemmatizer so we can lowercase the proper nouns to allow us to catch them with our Matchers patterns.

polm:
The lemmatizer FAQ we mentioned before is now available: https://github.com/explosion/spaCy/discussions/11685

We think the options described there should be a good starting point for most customizations, but if you still don't understand how something works feel free to open a discussion about how we can improve things.

Issue #11443, opened 2022-09-06
ValueError: [E1042] Function was called with `enable`=['tagger', 'lemmatizer'] and `disable`=['senter'] but these arguments are conflicting.


rbracco:
I am trying to load "en_core_web_sm" and only enable the tagger and lemmatizer following [the docs](https://spacy.io/usage/processing-pipelines#disabling) but I get the stack trace included below. I am not disabling 'senter' and adding 'senter' as an argument to enable, and/or passing `disable=[]` both fail to fix the problem. 

```
File "/home/rob/code/text-corpus/parsers/tagger.py", line 14, in __init__
    self.tagger = spacy.load("en_core_web_sm", enable=["tagger", "lemmatizer"])
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/__init__.py", line 54, in load
    return util.load_model(
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/util.py", line 429, in load_model
    return load_model_from_package(name, **kwargs)  # type: ignore[arg-type]
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/util.py", line 465, in load_model_from_package
    return cls.load(vocab=vocab, disable=disable, enable=enable, exclude=exclude, config=config)  # type: ignore[attr-defined]
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/en_core_web_sm/__init__.py", line 10, in load
    return load_model_from_init_py(__file__, **overrides)
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/util.py", line 646, in load_model_from_init_py
    return load_model_from_path(
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/util.py", line 503, in load_model_from_path
    nlp = load_model_from_config(
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/util.py", line 551, in load_model_from_config
    nlp = lang_cls.from_config(
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/language.py", line 1869, in from_config
    disabled_pipes = cls._resolve_component_status(
  File "/home/rob/code/text-corpus/.venv/lib/python3.8/site-packages/spacy/language.py", line 2052, in _resolve_component_status
    raise ValueError(
[...]

PR #11459, opened 2022-09-08
Simplify and clarify enable/disable behavior of spacy.load()

rmitsch:
## Description

As brought up in https://github.com/explosion/spaCy/issues/11443, the current behavior of `spacy.load()` w.r.t. `enable`, `disable` and the `enabled`/`disabled` options in the config is opaque and inconvenient to work around in case of a conflict. This PR 
- extends the error message to be informative,
- lets `enable`/`disable` values passed to `spacy.load()` take precedence over config values, i.e. config values are overwritten if function arguments are passed, and
- adds a warning message in case this (discarding config values because function arguments were passed) happens, since this can have unforeseen consequences for other pipeline components.

### Types of change

Enhancement (maybe fix?).

## Checklist


spacy/language.py
```diff
@@ -1727,12 +1729,6 @@ def from_config(
 
         DOCS: https://spacy.io/api/language#from_config
         """
-        if isinstance(disable, str):
-            disable = [disable]
-        if isinstance(enable, str):
-            enable = [enable]
-        if isinstance(exclude, str):
-            exclude = [exclude]
```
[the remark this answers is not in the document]
rmitsch: The checks need to be modified in order to be able to know whether an argument was passed to `spacy.load()`. This and the new approach to handling conflicting values from the config and passed arguments are implemented in `fetch_pipes_status()`.

spacy/language.py
```diff
@@ -1876,10 +1872,36 @@ def from_config(
...
+        ) -> Iterable[str]:
+            """Fetch value for `enable` or `disable` w.r.t. the specified config and passed arguments passed to
+            .load(). If both arguments and config specified values for this field, the passed arguments take precedence
+            and a warning is printed.
+            value (Optional[Union[str, Iterable[str]]]): Passed value for `enable` or `disable`.
+            key (str): Key for field in config (either "enabled" or "disabled").
+            RETURN (Iterable[str]):
+            """
+            # We assume that no argument was passed if the value is the default value of an empty SimpleFrozenList.
+            if isinstance(value, SimpleFrozenList) and len(value) == 0:
```
rmitsch, line 1886: I'm not really happy about this way of inferring that no value was passed for this argument. An alternative would be setting the default values for `enable` and `disable` to `None`, but this would have to be done for multiple functions.

spacy/errors.py
```diff
@@ -212,6 +212,8 @@ class Warnings(metaclass=ErrorsWithCodes):
     W121 = ("Attempting to trace non-existent method '{method}' in pipe '{pipe}'")
     W122 = ("Couldn't trace method '{method}' in pipe '{pipe}'. This can happen if the pipe class "
             "is a Cython extension type.")
+    W123 = ("Argument {arg} with value {arg_value} is used instead of the in the config defined value of "
```
adrianeboyd, line 215: "instead of the in the config defined value" looks like a mistake?

spacy/language.py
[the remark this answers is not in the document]
adrianeboyd: I don't understand how modifying this here makes a difference in the later checks?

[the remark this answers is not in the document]
rmitsch: `fetch_pipes_status()` changes the behavior w.r.t. `disable`/`enable`/`disabled`/`enabled` values. It only uses `disabled`/`enable` from the config if `disable`/`enable` are not set at all (hence the [check for an empty `SimpleFrozenList`](https://github.com/explosion/spaCy/pull/11459#discussion_r965884449)). Otherwise only `disable`/`enable` are used. Before this PR we merged `disable` with `disabled, `enable` with `enabled` and passed all values to `_resolve_component_status()`, which raised the issue that motivated this change.

I integrated the list conversion you commented on into `fetch_pipes_status()`. I could also pull it out from there and reinstate
```python
        if isinstance(disable, str):
            disable = [disable]
        if isinstance(enable, str):
            enable = [enable]
        if isinstance(exclude, str):
            exclude = [exclude]
```
\- it shouldn't make a difference though, as far as I can tell. 

Does this answer your question?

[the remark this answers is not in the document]
adrianeboyd: I think it would be easier to follow what's going on with the types if you handle the `str` type initially as it was before and then simplify the types for `fetch_pipes_status`. To me this change makes it seem like you need to know whether it was originally `str` later when you don't.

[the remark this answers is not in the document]
rmitsch: Restored the original type checks/list conversions and simplified types for `fetch_pipes_status()`. Feel free to reopen if you think something's still amiss.

spacy/language.py
svlandeg: Hm, it's not great, especially because of the indirect dependency across almost 200 lines of code.
What if we store one of these frozen lists in a private global var, specify that as default in the `from_config` signature and check for identity here instead of `isinstance`? Then at least this check becomes more meaningful and correct, and because of the way `SimpleFrozenList` is implemented I don't think there could be any harm in reusing the var for all three input vars here? We could also have 3 global vars - one for `disable`/`enable`/`exclude` each.

spacy/tests/pipeline/test_pipe_methods.py
```diff
@@ -605,10 +605,35 @@ def component(doc):
             assert results[component] == ""
 
 
-def test_load_disable_enable() -> None:
-    """
-    Tests spacy.load() with dis-/enabling components.
-    """
+@pytest.mark.issue(11443)
+def test_issue11443():
```
svlandeg, line 609: Let's make the name of this test descriptive (we've moved away from `test_issueXXX` as we've got the pytest markers now)

svlandeg (commented):
This looks like the best solution to me, and at least provides some helpful warnings to guide users to the correct usage of this functionality.

rmitsch:
> Hm, it's not great, especially...

I [agree](https://github.com/explosion/spaCy/pull/11459#discussion_r965884449), but is it really more correct than checking if it's an empty `SimpleFrozenList`, as all empty instances should be identical? I don't mind doing it this way, I'm just wondering. The only alternative I can think of is changing the default param to `None`, which is messy.

svlandeg:
> > Hm, it's not great, especially...
> 
> I [agree](https://github.com/explosion/spaCy/pull/11459#discussion_r965884449), but is it really more correct than checking if it's an empty `SimpleFrozenList`, as all empty instances should be identical? I don't mind doing it this way, I'm just wondering. The only alternative I can think of is changing the default param to `None`, which is messy.

Are all empty instances identical? I mean, what if a user specifically calls the function with an empty `SimpleFrozenList` - couldn't we tell the difference?

Apart from that argument, there's also a worry of maintainability - if we ever do change the default to `None` (which I don't advocate for right now), then we might not notice that 200 lines further down the line, an implicit dependency has broken. But this will be obvious when we store this default in a var.

rmitsch:
> Apart from that argument, there's also a worry of maintainability...

Yeah, I agree. 

> Are all empty instances identical? I mean, what if a user specifically calls the function with an empty SimpleFrozenList - couldn't we tell the difference?

Not via equality, but via identity. The latest commit includes a suggestion for how we could do this, based on your idea of a private global variable. Since this default value `DEFAULT_PIPES_STATUS` is used in several modules, it's currently not private though, which I'm not fond of. Let me know what you think (name is also up for debate).

spacy/util.py
```diff
@@ -394,13 +393,17 @@ def get_module_path(module: ModuleType) -> Path:
     return file_path.parent
 
 
+# Default value for passed enable/disable values.
+DEFAULT_PIPES_STATUS = SimpleFrozenList()
```
svlandeg, line 397: I know we import it in the `language` module, but nevertheless I think this can be "private", i.e. `_DEFAULT_PIPES_STATUS`

And actually, maybe singular? `_DEFAULT_PIPE_STATUS`

svlandeg (commented):
This looks good to me. Still trying to think of a decent name though. Maybe `_DEFAULT_EMPTY_PIPES`? We'd never have one default which contained actual pipe names, because then we couldn't reuse this var across `disable`, `enable` and `exclude`.

spacy/util.py
rmitsch: Renamed to `_DEFAULT_EMPTY_PIPES`. Not super happy with any of the names so far (`DEFAULT_PIPE_STATUS` included), but couldn't come up with a better one yet. Since this is private now and limited in its use anyway, I think it's alright if we go ahead and change the name later if we feel the need to?

PR #11464, opened 2022-09-08
Make numpy install requirement dynamic

adrianeboyd:
## Description

Use the current numpy version as the minimum numpy version for the install requirements when compiling the package from source.

* Remove unused `setup_requires`
* Move `install_requires` to `setup.py` and add the dynamic `numpy` requirement during setup
* Use proper requirements parsing in package tests as much as possible (but without additional libraries, so we still can't parse toml properly)

To do in the future when it's supported: mark `install_requires` explicitly as dynamic.

### Types of change

Enhancement?

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ keys/
 
 # Tests
 spacy/tests/package/setup.cfg
+spacy/tests/package/setup.py
 spacy/tests/package/pyproject.toml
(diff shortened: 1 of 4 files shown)


adrianeboyd:
This will probably not work as intended due to interactions with the pip cache.

PR #11465, opened 2022-09-09
Remove has_letters in config template

adrianeboyd:
## Description

Due to problems with the javascript conversion in the website quickstart, remove the `has_letters` setting to simplify generating `attrs` for the default `tok2vec`.

Additionally reduce `PREFIX` as in the trained pipelines.

### Types of change

Primarily docs.

## Checklist


adrianeboyd:
I'd like to go ahead and merge this to fix the website. It makes it slightly out-of-sync with the current release in terms the quickstart vs. `init config` for Chinese, but I think it's better to go ahead and fix the quickstart.

PR #11489, opened 2022-09-12
more explicit Example constructor example

svlandeg:
## Description
The API docs for `Example` refer to a dummy function `parse_gold_doc` that implies "parse whatever data to create your `Doc` object", but this seems to create some confusion among users. Instead, proposing to make this example code much more explicit by calling the `Doc` constructor for both `predicted` and `reference`.

### Types of change
doc fix

## Checklist

### DIFF
--- a/website/docs/api/example.md
+++ b/website/docs/api/example.md
@@ -23,11 +23,11 @@ both documents.
 > ```python
 > from spacy.tokens import Doc
 > from spacy.training import Example
->
-> words = ["hello", "world", "!"]
-> spaces = [True, False, False]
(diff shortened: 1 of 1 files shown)


adrianeboyd:
I wonder a bit about leaving out `spaces` here? I worry that users might end up confused by the mismatched texts in the end when they thought they were doing something simple?

svlandeg:
I thought it would be useful to show that `reference` and `predicted` can have a different tokenization, especially because the docs page talk about the alignment between the two early on? I can put the `spaces` back, sure.

adrianeboyd:
The difference in tokenization is fine, I'm only concerned about the resulting whitespace in the texts.

Issue #11499, opened 2022-09-14
Doc.init changes input to is_sent_start to List[int]


stefawolf:
Hey,

I just stumbled accross the following behaviour, when creating a Doc manually. Is this intended behaviour?

## How to reproduce the behaviour
    import spacy
    from spacy.tokens import Doc
    nlp = spacy.blank("xx")
    words_1 = ["This","is","the","first","sentence","."]
    spaces_1 =[True,True,True,True,False,True]
    sent_starts_1 = [True,False,False,False,False,False]
    test_doc = Doc( vocab=nlp.vocab,words=words_1, spaces=spaces_1, sent_starts=sent_starts_1)
    print(sent_starts_1)
    #which prints [1, -1, -1, -1, -1, -1]

## Info about spaCy

- **spaCy version:** 3.4.1
- **Platform:** Linux-5.10.16.3-microsoft-standard-WSL2-x86_64-with-glibc2.29
- **Python version:** 3.8.10

richardpaulhudson:
Thanks for letting us know about this, I'll have a look at fixing it.

PR #11506, opened 2022-09-15
Remove side effects from Doc.__init__()

richardpaulhudson:
## Description
The existing version of `Doc.__init__()` changes `sent_starts` and `ents` lists passed as parameters. This PR removes these side effects.

Closes [#11499.](https://github.com/explosion/spaCy/issues/11499)

### Types of change
Bugfix

## Checklist


spacy/tokens/doc.pyx
```diff
@@ -285,27 +285,27 @@ cdef class Doc:
             heads = [0] * len(deps)
         if heads and not deps:
             raise ValueError(Errors.E1017)
+        _sent_starts: Optional[List[int]] = [] if sent_starts is not None else None
```
adrianeboyd, line 288: Based on the processing below, it looks like `int` values are supported, so I think the original type should be corrected to `Optional[List[Union[bool, int, None]]]` and then I think you can just make a copy and not need to modify the other processing steps:

```python
sent_starts = list(sent_starts) if sent_starts is not None else None
```

richardpaulhudson: It would certainly be good to keep `sent_starts` rather than `_sent_starts` lower down in the method, but is performance not a concern at all? That way we make an additional copy of the list.

adrianeboyd: I don't think there's going to be a big difference between creating a copy vs. creating a new list variable. I think it's much simpler to just create a copy to avoid this side effect.

adrianeboyd:
Also update the types in `doc.pyi`?

richardpaulhudson:
Oh yes, sorry, hadn't seen that!

spacy/tests/doc/test_doc_api.py
```diff
@@ -82,6 +82,28 @@ def test_issue2396(en_vocab):
...
+        vocab=en_vocab,
+        words=words,
+        spaces=spaces,
+        heads=heads,
+        deps=deps,
+        ents=ents,
+        sent_starts=sent_starts,
+    )
+    assert ents == ["B-TYPE1", "I-TYPE2"]
+    assert sent_starts == [True, False]
```
adrianeboyd, line 104:
```suggestion
def test_init_args_unmodified(en_vocab):
    words = ["A", "sentence"]
    ents = ["B-TYPE1", ""]
    sent_starts = [True, False]
    Doc(
        vocab=en_vocab,
        words=words,
        ents=ents,
        sent_starts=sent_starts,
    )
    assert ents == ["B-TYPE1", ""]
    assert sent_starts == [True, False]
```

adrianeboyd:
Looking good, I think the last thing that's missing is the corresponding update for the `Doc` API docs.

PR #11551, opened 2022-09-28
Fix multiple entries per custom extension in doc json

thomashacker:
## Description

Currently, the `Doc` to `JSON` conversion only saves one `Token`/`Span` object per custom extension. This gets problematic when multiple `Token`/`Span` objects use the same extension. This `PR` ensures that the objects are saved in a list per custom extension.

Additionally, this `PR` fixes the problem that the `JSON -> Doc` conversion treats the indices as `character_offsets` instead of `token_offsets`.

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -1610,22 +1610,20 @@ cdef class Doc:
 
         if doc_json.get("underscore_token", {}):
             for token_attr in doc_json["underscore_token"]:
-                token_start = doc_json["underscore_token"][token_attr]["token_start"]
-                value = doc_json["underscore_token"][token_attr]["value"]
-
                 if not Token.has_extension(token_attr):
                     Token.set_extension(token_attr)
-                self[token_start]._.set(token_attr, value)
+                for token in doc_json["underscore_token"][token_attr]:
+                    token_start = token_by_char(self.c, self.length, token["token_start"])
+                    value = token["value"]
+                    self[token_start]._.set(token_attr, value)
                 
         if doc_json.get("underscore_span", {}):
             for span_attr in doc_json["underscore_span"]:
-                token_start = doc_json["underscore_span"][span_attr]["token_start"]
-                token_end = doc_json["underscore_span"][span_attr]["token_end"]
-                value = doc_json["underscore_span"][span_attr]["value"]
-
                 if not Span.has_extension(span_attr):
                     Span.set_extension(span_attr)
-                self[token_start:token_end]._.set(span_attr, value)
+                for span in doc_json["underscore_span"][span_attr]:
+                    value = span["value"]
+                    self.char_span(span["token_start"],span["token_end"])._.set(span_attr, value)
         return self
 
     def to_json(self, underscore=None):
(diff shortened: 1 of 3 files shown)


spacy/tokens/doc.pyx
```diff
@@ -1692,11 +1690,13 @@ cdef class Doc:
                             # Check if token attribute
                             elif end is None:
                                 if attr not in data["underscore_token"]:
-                                    data["underscore_token"][attr] = {"token_start": start, "value": value}
+                                    data["underscore_token"][attr] = []
+                                data["underscore_token"][attr].append({"token_start": start, "value": value})
```
adrianeboyd, line 1694: Since these are character offsets, I would use `start` and `end` like the other character span entries in `Doc.to_json`.

spacy/tokens/doc.pyx
```diff
@@ -1610,22 +1610,20 @@ cdef class Doc:
 
         if doc_json.get("underscore_token", {}):
             for token_attr in doc_json["underscore_token"]:
-                token_start = doc_json["underscore_token"][token_attr]["token_start"]
-                value = doc_json["underscore_token"][token_attr]["value"]
-
                 if not Token.has_extension(token_attr):
                     Token.set_extension(token_attr)
-                self[token_start]._.set(token_attr, value)
+                for token in doc_json["underscore_token"][token_attr]:
```
adrianeboyd, line 1615: In the code base in general calling a non-`Token` variable `token` can get kind of confusing. Maybe `token_entry`? `token_dict`? `token_ext_data`? ???

```suggestion
                for token_ext_data in doc_json["underscore_token"][token_attr]:
```

spacy/tokens/doc.pyx
```diff
@@ -1610,22 +1610,20 @@ cdef class Doc:
...
         if doc_json.get("underscore_span", {}):
             for span_attr in doc_json["underscore_span"]:
-                token_start = doc_json["underscore_span"][span_attr]["token_start"]
-                token_end = doc_json["underscore_span"][span_attr]["token_end"]
-                value = doc_json["underscore_span"][span_attr]["value"]
-
                 if not Span.has_extension(span_attr):
                     Span.set_extension(span_attr)
-                self[token_start:token_end]._.set(span_attr, value)
+                for span in doc_json["underscore_span"][span_attr]:
```
adrianeboyd, line 1624: Ditto for non-`Span` `span`.

spacy/tokens/doc.pyx
```diff
@@ -1612,18 +1612,18 @@ cdef class Doc:
...
-                    token_start = token_by_char(self.c, self.length, token["token_start"])
-                    value = token["value"]
-                    self[token_start]._.set(token_attr, value)
+                for token_data in doc_json["underscore_token"][token_attr]:
+                    start = token_by_char(self.c, self.length, token_data["start"])
+                    value = token_data["value"]
+                    self[start]._.set(token_attr, value)
                 
         if doc_json.get("underscore_span", {}):
             for span_attr in doc_json["underscore_span"]:
```
polm, line 1621: I realize this isn't in the current commit, but can't this be replaced with a line like this?

```
for span_attr in doc_json.get("underscore_span", {}):
```

Same for the doc one further up. 

The current structure seems weird unless we are concerned about the case where `underscore_span` has been set to an unusuable / non-iterable value, like `null`, which I don't think is possible with data we generate?

polm (approved):
I found one thing that looks like it can be improved, though it's not actually in the changes in this PR, but in the surrounding code. The changes look good to me. 

Having not worked with the user data internal structure before, I realized it was a little confusing. Is the four-field structure documented somewhere? It wasn't that hard to figure out but it would have helped to have a comment.

polm:
Oh yeah, one other concern - do we want to handle data generated with the previous iteration of this? In that case, when we try to read in the data, we'll find single values where we expect lists given these changes, and it'll blow up. Data generated in that format isn't very useful, so maybe that's acceptable.

adrianeboyd:
The buggy version wasn't in a release yet, so I think it's okay to skip error checking or backwards compat for that.

spacy/tokens/doc.pyx
thomashacker: `underscore_span` and `underscore_token` are only set when `underscore` is.

thomashacker:
Thanks @adrianeboyd! Sorry for blocking the release

PR #11555, opened 2022-09-29
Update docs to reflect Doc input to Language

polm:
## Description

In #9069 `__call__`/`pipe` were changed so they could take Docs as input, but the API docs weren't updated. This just updates the documentation and explains what happens when passing a Doc.

### Types of change

docs update

## Checklist


svlandeg (approved):
Good catch!

PR #11583, opened 2022-10-05
Add new features and options to tok2vec to improve accuracy

richardpaulhudson:
## Description
This PR presents a new way of generating `tok2vec` features to improve the accuracy of the edit tree lemmatizer and morphologizer.

### Types of change
Enhancement / new feature

## Checklist

## Introduction

For a description of the functionality introduced by this PR, please read the documentation in [architectures.mdx](https://github.com/explosion/spaCy/pull/11583/files#diff-2514925cc24ac1e94f41e2a3d3bd3b8936563667ba04873cc843e753a32f5cfd).

This PR is one of a series of three (#11583; #11959; #12017) that increase the accuracy of the edit-tree lemmatizer: the cumulative accuracy improvement is shown in the final section below for each language. The mean morphologizer accuracy increased from 95.7% to 96.2% (+0.5%); the mean lemmatizer accuracy increased from 94.3% to 96.0% (+1.7%). The figures are particularly encouraging for the weaker models: while the pre-existing code yields models with accuracies in the mid-80s% for a few languages, with the cumulative changes at least 91% is achieved for all languages.

#### General comments

- The new functionality was tested with the whole range of pipeline components. It was observed to improve accuracy for the edit tree lemmatizer and the morphologizer but to slightly decrease accuracy for other components like the dependency parser. If it is incorporated into the standard model pipelines, this means it will make sense to define a second upstream `tok2vec` component using the new functionality to feed into the lemmatizer and morphologizer and to retain the existing `tok2vec` definition for the other components like the dependency parser. Although I haven't tested this in detail, initial results seemed to suggest that the accuracy of these other components may very slightly increase by virtue of them being trained without the morphologizer and lemmatizer; this impression may however not have been statistically significant.
- Throughout, the standard features `LOWER`, `SHAPE` and `SPACY` were embedded with row widths 5000, 2500 and 50 respectively in addition to the new affixes and search results. Using `LOWER` (or `NORM`) rather than `ORTH` was found to give better or equal results, even for German where capitalisation is a part-of-speech marker. This will be doubly the case if this PR is used in conjunction with https://github.com/explosion/spaCy/pull/11959.
- Morphologizer and edit-tree-lemmatizer models were trained for the 17 languages with published spaCy models and appropriate configurations and linguistic typologies (e.g. English and Chinese were not included as they do not use trained lemmatizers and Chinese displays so little morphological variation). For each language, a best-guess configuration was used based on the salient features of the language's grammar. The configurations are described in the table below using the following notation:

`P(x, y, ...)` :             prefixes of lengths `x` and `y`
`S(x, y, ...)` :             suffixes of lengths `x` and `y`

- The best row width for the new features appeared to be 10000, although this was not a particularly strong/significant finding.

#### Accuracy

The lemmatizer accuracy increased for every language except Portuguese, where it remained the same; the mean increase was 0.8%; as expected, there were large rises for languages like Finnish and Lithuanian that typically feature long morphemes and that tended to have low accuracy with the pre-existing solution. The morphologizer accuracy increased for every language except Croatian, where it decreased slightly; the mean increase was 0.6%.

The global dropout probability used for the tests was 0.2, which had observed to give somewhat better results than 0.1 during development.

| Model | Morph. acc. old (%) | Morph. acc. new (%) | Morph. acc. diff (%) | Lemm. acc. old (%) | Lemm. acc. new (%) | Lemm. acc. diff (%) | Config | 
| --- | --- | --- | --- | --- | --- | --- | --- |
| ca_core_news_lg | 98.2 | 98.3 | 0.1 | 98.7 | 99.1 | 0.4 | S(2,3,4,5) |
| da_core_news_lg | 95.3 | 96.2 | 0.9 | 95.2 | 95.4 | 0.2 | S(2,3,4) |
| de_core_news_lg | 92.2 | 92.5 | 0.3 | 97.9 | 98.5 | 0.6 | P(2,3,5); S(2,3,4,5) |
| el_core_news_lg | 91.0 | 92.0 | **1.0** | 89.9 | 91.2 | **1.3** | S(2,3,4,5,6) |
| es_core_news_lg | 98.2 | 98.3 | 0.1 | 98.4 | 99.0 | 0.6 | S(2,3,4,5) |
| fi_core_news_lg | 92.2 | 93.5 | **1.3** | 86.4 | 88.9 | **2.5** | S(2,3,4,5,6) |
| fr_core_news_lg | 96.8 | 97.1 | 0.3 | 94.9 | 96.1 | **1.2** | S(2,3,4,5,6) |
| hr_core_news_lg | 92.8 | 92.5 | *-0.3* | 92.9 | 93.3 | 0.4 | S(2,3,4,5) |
| it_core_news_lg | 97.4 | 97.8 | 0.4 | 97.6 | 97.7 | 0.1 | S(2,3,4,5) |
| ko_core_news_lg | - | - | - | 90.2 | 91.7 | **1.5** | S(1,2,3) |
| lt_core_news_lg | 89.0 | 90.1 | **1.1** | 85.8 | 88.6 | **2.8** | S(2,3,4,5,6) |
| nb_core_news_lg | 96.3 | 96.7 | 0.4 | 97.2 | 97.3 | 0.1 | S(2,3,4) |
| nl_core_news_lg | 96.4 | 96.7 | 0.3 | 95.7 | 96.1 | 0.4 | P(2,3,5); S(2,3,4,5) |
| pl_core_news_lg | 90.8 | 91.2 | 0.4 | 94.3 | 94.7 | 0.4 | S(2,3,4,5) |
| pt_core_news_lg | 95.8 | 96.0 | 0.2 | 97.3 | 97.3 | *0.0* | S(2,3,4,5,6) |
| ro_core_news_lg | 95.1 | 97.0 | **1.9** | 95.8 | 95.9 | 0.1 | S(2,3,4,5) | 
| sv_core_news_lg | 95.7 | 96.2 | 0.5 | 95.4 | 96.1 | 0.7 | S(2,3,4) |

#### Speed

- Inference speed for `pl_core_news_lg`, using only tok2vec, morphologizer and lemmatizer, was compared using this PR versus using the pre-existing hash embedding functionality with fixed-length prefixes and suffixes. The speed penalty was measured at 6.5% on CPU and 3.8% on GPU.

#### Changes

- Considering multiple predictions for each lemma, i.e. setting `top_k = 5` rather than `top_k = 1`, which was made feasible by #12017;
- Feeding suffix and prefix features of various lengths through to `tok2vec` (this PR)
- Adding a dropout layer between `tok2vec` and the components it feeds into, and setting a global dropout probability of 0.2 (this PR)
- Learning lowercasing as a distinct task (#11959)
- Using the glorot uniform initializer for the ETL neural network (##11959)

#### Approaches that were investigated and abandoned

- Modelling regular morphological alternations: many languages have morphological alternations that are applied to a number of different letters or groups of letters in similar grammatical situations. Examples include replacing single letters with double letters or vice versa in Dutch and replacing voiceless consonants with voiced consonants in Croatian. The hope was that supplying the edit tree lemmatizer with the details of any regular alternations in a given language would allow it to apply them as "abstract substitutions" and lead to more parsimonious tree structures and more efficient and accurate learning. However, this did not work as expected.
- Assessing predictions using n-gram fitness: The n-gram frequencies observed in the training set can be used to determine the acceptability of a lemma prediction. The hope was that, in conjunction with setting `top_k = 5`, this information could be used to filter out incorrect predictions; however, the positive effect on precision was accompanied by a more or less equal negative effect on recall.

#### Accuracy

The mean morphologizer accuracy increased from 95.7% to 96.2% (+0.5%); the mean lemmatizer accuracy increased from 94.3% to 96.0% (+1.7%):

| Model | Morph. acc. old (%) | Morph. acc. new (%) | Morph. acc. diff (%) | Lemm. acc. old (%) | Lemm. acc. new (%) | Lemm. acc. diff (%) |
| --- | --- | --- | --- | --- | --- | --- | 
| ca_core_news_lg | 98.2 | 98.2 | *0.0* | 98.7 | 99.1 | 0.4 |
| da_core_news_lg | 95.3 | 96.0 | 0.7 | 95.2 | 96.5 | **1.3** |
| de_core_news_lg | 92.2 | 92.6 | 0.4 | 97.9 | 98.6 | 0.7 |
| el_core_news_lg | 91.0 | 92.0 | **1.0** | 89.9 | 92.9 | **3.0** |
| es_core_news_lg | 98.2 | 98.3 | 0.1 | 98.4 | 99.1 | 0.7 |
| fi_core_news_lg | 92.2 | 93.6 | **1.4** | 86.4 | 91.0 | **4.6** |
| fr_core_news_lg | 96.8 | 97.1 | 0.3 | 94.9 | 96.9 | **2.0** |
| hr_core_news_lg | 92.8 | 92.6 | *-0.2* | 92.9 | 94.7 | **1.8** |
| it_core_news_lg | 97.4 | 97.8 | 0.4 | 97.6 | 98.2 | 0.6 |
| ko_core_news_lg | - | - | - | 90.2 | 91.8 | **1.6** |
| lt_core_news_lg | 89.0 | 90.9 | **1.9** | 85.8 | 91.2 | **5.4** |
| nb_core_news_lg | 96.3 | 96.8 | 0.5 | 97.2 | 98.0 | 0.8 |
| nl_core_news_lg | 96.4 | 96.7 | 0.3 | 95.7 | 96.5 | 0.8 |
| pl_core_news_lg | 90.8 | 91.2 | 0.4 | 94.3 | 96.4 | **2.1** |
| pt_core_news_lg | 95.8 | 96.1 | 0.3 | 97.3 | 97.8 | 0.5 |
| ro_core_news_lg | 95.1 | 97.1 | **2.0** | 95.8 | 97.0 | **1.2** |
| sv_core_news_lg | 95.7 | 96.2 | 0.5 | 95.4 | 97.2 | **1.8** |

There is one transformer-based model, `de_dep_news_trf`, that uses the edit-tree lemmatizer. Those three of the five changes listed above that are **not** part of this PR are relevant to transformer models; applying them increased the accuracy from 98.7% to 98.9% (+0.2%).

#### Speed

Lemmatizer and morphologizer inference with `pl_core_news_lg` were measured as being 12.2% slower on CPU and 42.4% slower on GPU with the five changes listed above included than without them.

With `de_dep_news_trf`, which was only run on GPU, the speed penalty was 7.3%.


richardpaulhudson:

spacy/tokens/doc.pyx
```diff
@@ -1734,6 +1735,89 @@ cdef class Doc:
...
+        cdef unsigned int num_toks = len(self), num_norm_hashes = len(affix_lengths), num_search_hashes = len(search_lengths)
+        cdef np.ndarray[np.int64_t, ndim=2] hashes = numpy.empty((num_toks, num_norm_hashes + num_search_hashes), dtype="int64")
+
+        cdef const unsigned char[:] tok_str_v
+        cdef unsigned int tok_idx, tok_str_v_len, hash_idx, affix_start, char_comb_len
+        cdef attr_t num_tok_attr
+        cdef str str_tok_attr
+        
+        for tok_idx in range(num_toks):
+            num_tok_attr = self.c[tok_idx].lex.orth if case_sensitive else self.c[tok_idx].lex.lower
```
richardpaulhudson, line 1793: I tried various ways of retrieving the string for each token. Perhaps somewhat counterintuitively this is the fastest I came up with, although there may well be a better one I haven't tried. It's certainly faster than retrieving the character index and length of each token and slicing `doc.text`. Declaring a memory view on `doc.text` is unfortunately not an option because there is no guarantee that all characters will be the same width.

spacy/tokens/doc.pyx
```diff
@@ -1915,6 +1999,70 @@ cdef int [:,:] _get_lca_matrix(Doc doc, int start, int end):
...
 
 
+cdef const unsigned char[:] _get_utf16_memoryview(str unicode_string, const bint check_2_bytes):
+    """
+    Return a memory view of the UTF-16 representation of a string with the default endianness of the platform.
+    Throw a ValueError if *check_2_bytes == True* and one or more characters in the UTF-16 representation
+    occupies four bytes rather than two.
+    """
+    cdef const unsigned char[:] view = unicode_string.encode("UTF-16")
+    view = view[2:] # first two bytes express endianness
```
richardpaulhudson, line 2009: We can ignore endianness, because each two-byte pair is passed with the default endianness of whatever platform we are on to murmurhash, which takes endianness into account. The tests confirm that the same hash values are returned on all three platforms.

spacy/tokens/doc.pyx
```diff
@@ -1734,6 +1735,89 @@ cdef class Doc:
...
+            str_tok_attr = self.vocab.strings[num_tok_attr]
+            tok_str_v = _get_utf16_memoryview(str_tok_attr, False)
+            tok_str_v_len = len(tok_str_v)
+
+            for hash_idx in range(num_norm_hashes):
+                char_comb_len = affix_lengths[hash_idx] * 2
+                if char_comb_len > tok_str_v_len:
+                    char_comb_len = tok_str_v_len
+                affix_start = tok_str_v_len - char_comb_len if suffs_not_prefs else 0
+                hashes[tok_idx, hash_idx] = hash32(<void*> &tok_str_v[affix_start], char_comb_len, 0)
```
richardpaulhudson, line 1803: Calling `murmurhash` directly in C is faster than hashing via `self.vocab`.

spacy/ml/richfeatureextractor.py
```diff
@@ -0,0 +1,72 @@
...
+    ops = model.ops
+    case_sensitive: bool = model.attrs["case_sensitive"]
+    pref_lengths: List[int] = model.attrs["pref_lengths"]
+    pref_search_chars: str = model.attrs["pref_search_chars"]
+    pref_search_lengths: List[int] = model.attrs["pref_search_lengths"]
+    suff_lengths: List[int] = model.attrs["suff_lengths"]
+    suff_search_chars: str = model.attrs["suff_search_chars"]
+    suff_search_lengths: List[int] = model.attrs["suff_search_lengths"]
+    features: List[Ints2d] = []
+    for doc in docs:
```
richardpaulhudson, line 54: TODO: change to call a single Cython method that returns both prefix and suffix hashes. At present each token is being encoded twice, once for prefixes and once for suffixes; and calling a single Cython method will allow us to get rid of the `hstack`.

spacy/ml/models/tok2vec.py
```diff
@@ -186,6 +189,197 @@ def make_hash_embed(index):
...
+            raise ValueError(Errors.E1047.format(label=label))
+    elif search_chars is not None:
+        raise ValueError(Errors.E1047.format(label=label))
+    if lengths is not None:
+        if lengths[-1] > 63:
+            raise ValueError(Errors.E1048.format(label=label))
+        if len(lengths) != len(set(lengths)) or lengths != sorted(lengths):
+            raise ValueError(Errors.E1049.format(label=label))
+
+
```
richardpaulhudson, line 218: This function contains all fields from `MultiHashEmbed.v2`. The only change required to make it backwards compatible with respect to `MultiHashEmbed.v2` would be to make `case_sensitive` optional. If so desired, then, we could replace `MultiHashEmbed.v2` and name this function `MultiHashEmbed.v3` instead of `RichMultiHashEmbed.v1`.

spacy/strings.pyx
```diff
@@ -315,3 +315,25 @@ cdef class StringStore:
...
+
+    @cython.boundscheck(False)  # Deactivate bounds checking
+    cdef const unsigned char* utf8_ptr(self, const attr_t hash_val):
+        if hash_val == 0:
+            return b""
+        elif hash_val < len(SYMBOLS_BY_INT):
+            return SYMBOLS_BY_INT[hash_val].encode("utf-8")
+        cdef Utf8Str* string = <Utf8Str*>self._map.get(hash_val)
+        if string.s[0] < sizeof(string.s) and string.s[0] != 0:
+            return string.s[1:string.s[0]+1]
```
danieldk, line 327: I think this also creates a wild pointer. Cython slices the C character data, creating a `bytes`, then the `return` will return a pointer to the data. However, since there is no Python reference anymore it will be a wild pointer (from my reading of the Cython docs). The returned data needs to be allocated on the C heap and then deallocated at the call site. Or return a Python `bytes` object or a C++ `std::string` with a copy of the data.

spacy/strings.pyx
```diff
@@ -315,3 +315,25 @@ cdef class StringStore:
...
+        elif string.p[0] < 255:
+            return string.p[1:string.p[0]+1]
```
danieldk, line 329: Same here and below.

spacy/strings.pyx
richardpaulhudson: OK, thanks, I get the problem. I tried returning a `bytes` object instead of a `char*` but that slows down everything quite significantly. Am I right in thinking that I'd be fine to return `string` itself because it's a pointer to a value in the hash map and thus won't be garbage collected? If so maybe I should do that, deal with extracting the bytes from `string` in the calling code, and come up with some other solution(s) for the special cases (which don't need to be fast because they don't happen very often).

shadeMe: That should be fine as I don't see any method on `StringStore` that removes strings from the backing store. But then the lifetime of the `Utff8Str` is bound to the store itself, so you'll need to ensure that it stays alive at least as long as the string pointer needs to.

spacy/strings.pyx
```diff
@@ -315,3 +315,25 @@ cdef class StringStore:
         self._map.set(key, value)
         self.keys.push_back(key)
         return value
+
+    @cython.boundscheck(False)  # Deactivate bounds checking
+    cdef const unsigned char* utf8_ptr(self, const attr_t hash_val):
+        if hash_val == 0:
+            return b""
+        elif hash_val < len(SYMBOLS_BY_INT):
+            return SYMBOLS_BY_INT[hash_val].encode("utf-8")
```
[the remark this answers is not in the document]
richardpaulhudson: Thanks, should be fixed now.

spacy/strings.pyx
```diff
@@ -315,3 +315,25 @@ cdef class StringStore:
         self._map.set(key, value)
         self.keys.push_back(key)
         return value
+
+    @cython.boundscheck(False)  # Deactivate bounds checking
+    cdef const unsigned char* utf8_ptr(self, const attr_t hash_val):
+        if hash_val == 0:
+            return b""
```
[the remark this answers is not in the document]
richardpaulhudson: Thanks, should be fixed now. Btw because not all the representations within the string store are 0-terminated, I've now moved away from using 0-termination and use the declared lengths of each string instead.

spacy/strings.pyx
richardpaulhudson: Thanks, the bug should be fixed now. Because the `StringStore` instance is necessary to process the `Doc` object, I don't think there's any situation in which the `Doc` object could outlive it.

spacy/strings.pyx
richardpaulhudson: Thanks, should be fixed now.

spacy/strings.pyx
richardpaulhudson: Changing it like this should theoretically have a small positive impact on speed, although I couldn't measure one.

spacy/tokens/doc.pyx
```diff
@@ -1743,6 +1743,127 @@ cdef class Doc:
...
+        # Define / allocate buffers
+        cdef Pool mem = Pool()
+        cdef unsigned char* pref_l_buf = <unsigned char*> mem.alloc(p_max_l, 1)
+        cdef unsigned char* suff_l_buf = <unsigned char*> mem.alloc(s_max_l, 1)
+        cdef unsigned char* ps_res_buf = <unsigned char*> mem.alloc(ps_max_l, 4)
+        cdef unsigned char* ps_l_buf = <unsigned char*> mem.alloc(ps_max_l, 1)
+        cdef unsigned char* ss_res_buf = <unsigned char*> mem.alloc(ss_max_l, 4)
+        cdef unsigned char* ss_l_buf = <unsigned char*> mem.alloc(ss_max_l, 1)
+        cdef int doc_l = self.length, total_hashes = doc_l * hashes_per_tok 
+        cdef np.uint64_t* hashes_ptr = <np.uint64_t*> mem.alloc(
```
shadeMe, line 1819: Any reason this buffer can't be the `numpy` array itself? We could call `numpy.zeros` with `total_hashes` and use its `data` attribute to perform the operations in-place, and we save a copy in the process.

richardpaulhudson: Good point! I've changed it accordingly (using `numpy.empty` rather than `numpy.zeros` as all the positions in the buffer are guaranteed to be written to).

spacy/ml/models/tok2vec.py
richardpaulhudson: Good point!

spacy/tokens/doc.pyx
```diff
@@ -1743,6 +1743,124 @@ cdef class Doc:
                     j += 1
         return output
 
+    @cython.boundscheck(False)  # Deactivate bounds checking
+    def get_character_combination_hashes(self,
```
danieldk, line 1747: Does this need to be added to `Doc`? This seems functionality that is relatively tightly coupled to the rich feature extractor and `Doc.c` is accessible to Cython code outside `Doc`.

spacy/tokens/doc.pyx
```diff
@@ -1743,6 +1743,124 @@ cdef class Doc:
...
+        the fact that we are hashing short affixes and searching for small groups of characters. The calling code is responsible
+        for ensuring that lengths being passed in cannot exceed 63 and hence that resulting values with maximally four-byte 
+        character widths can never exceed 255.
+
+        Note that this method performs no data validation itself as it expects the calling code will already have done so, and
+        that the behaviour of the code may be erratic if the supplied parameters do not conform to expectations.
+        """
+        
+        # Work out lengths
+        cdef int p_lengths_l = strlen(<char*> p_lengths)
```
danieldk, line 1800: We already have the lengths at the call site. Also, character pointers are not great. Can we use a typed memoryview for these parameters?

spacy/tokens/doc.pyx
```diff
@@ -1743,6 +1743,124 @@ cdef class Doc:
...
+        cdef int ss_lengths_l = strlen(<char*> ss_lengths)
+        cdef int hashes_per_tok = p_lengths_l + s_lengths_l + ps_lengths_l + ss_lengths_l
+        cdef int p_max_l = p_lengths[p_lengths_l - 1] if p_lengths_l > 0 else 0
+        cdef int s_max_l = s_lengths[s_lengths_l - 1] if s_lengths_l > 0 else 0
+        cdef int ps_max_l = ps_lengths[ps_lengths_l - 1] if ps_lengths_l > 0 else 0
+        cdef int ss_max_l = ss_lengths[ss_lengths_l - 1] if ss_lengths_l > 0 else 0
+        
+        # Define / allocate buffers
+        cdef Pool mem = Pool()
+        cdef unsigned char* pref_l_buf = <unsigned char*> mem.alloc(p_max_l, 1)
```
danieldk, line 1812: I'd use `std::vector` for all of these, they give you the determenistic destruction. (Or perhaps `std::string` which is just a container for raw byte data.)

(Even though `char` is always 1 byte on modern platforms, I'd use `sizeof` rather than raw sizes, conveys the semantics better.)

spacy/tokens/doc.pyx
```diff
@@ -1743,6 +1743,124 @@ cdef class Doc:
...
+        cdef unsigned char* suff_l_buf = <unsigned char*> mem.alloc(s_max_l, 1)
+        cdef unsigned char* ps_res_buf = <unsigned char*> mem.alloc(ps_max_l, 4)
```
danieldk, line 1814: Why 4? something like `SOME_CONSTANT * sizeof(char)` would make it clearer what is going on.

spacy/tokens/doc.pyx
```diff
@@ -1743,6 +1743,124 @@ cdef class Doc:
                     j += 1
         return output
 
+    @cython.boundscheck(False)  # Deactivate bounds checking
+    def get_character_combination_hashes(self,
+        *,
+        const bint cs,
```
danieldk, line 1749: Rename to `case_sensitive`?

spacy/tokens/doc.pyx
```diff
@@ -1743,6 +1743,124 @@ cdef class Doc:
                     j += 1
         return output
 
+    @cython.boundscheck(False)  # Deactivate bounds checking
```
danieldk, line 1746: Did you test benchmark the overhead of bounds checks?

spacy/tokens/doc.pyx
```diff
@@ -1924,6 +2042,180 @@ cdef int [:,:] _get_lca_matrix(Doc doc, int start, int end):
...
 
 
+@cython.boundscheck(False)  # Deactivate bounds checking
+cdef void _set_prefix_lengths(
+    const unsigned char* tok_str,
+    const int tok_str_l,
+    const int p_max_l, 
+    unsigned char* pref_l_buf,
+) nogil:
+    """ Populate *pref_l_buf*, which has length *pref_l*, with the byte lengths of the first *pref_l* characters within *tok_str*.
```
danieldk, line 2052: Where is `pref_l`?

spacy/tokens/doc.pyx
```diff
@@ -1924,6 +2042,180 @@ cdef int [:,:] _get_lca_matrix(Doc doc, int start, int end):
...
+    unsigned char* pref_l_buf,
+) nogil:
+    """ Populate *pref_l_buf*, which has length *pref_l*, with the byte lengths of the first *pref_l* characters within *tok_str*. 
+        Lengths that are greater than the character length of the whole word are populated with the byte length of the whole word.
+
+        tok_str: a UTF-8 representation of a string.
+        tok_str_l: the length of *tok_str*.
+        p_max_l: the number of characters to process at the beginning of the word.
+        pref_l_buf: a buffer of length *p_max_l* in which to store the lengths. The calling code ensures that lengths
+            greater than 255 cannot occur.
```
danieldk, line 2059: Where? And what is special about the length 255?

spacy/tokens/doc.pyx
```diff
@@ -1924,6 +2042,180 @@ cdef int [:,:] _get_lca_matrix(Doc doc, int start, int end):
     return lca_matrix
 
 
+@cython.boundscheck(False)  # Deactivate bounds checking
+cdef void _set_prefix_lengths(
+    const unsigned char* tok_str,
```
danieldk, line 2047: See above, if these arrays could just be `vector` or `strings`, then they could be passed as (const) references.

spacy/tokens/doc.pyx
```diff
@@ -1924,6 +2042,180 @@ cdef int [:,:] _get_lca_matrix(Doc doc, int start, int end):
...
+    """
+    cdef int tok_str_idx = 1, pref_l_buf_idx = 0
+
+    while pref_l_buf_idx < p_max_l:
+        if (tok_str_idx >= tok_str_l 
+            or 
+            ((tok_str[tok_str_idx] & 0xc0) != 0x80)  # not a continuation character
+        ):
+            pref_l_buf[pref_l_buf_idx] = tok_str_idx
+            pref_l_buf_idx += 1
```
danieldk, line 2069: If you'd end up using pointers here, you don't need the `pref_l_buf_idx` variable, you could increment `pref_l_buf`, so that the pointer points to the next element.

spacy/tokens/doc.pyx
```diff
@@ -1924,6 +2042,180 @@ cdef int [:,:] _get_lca_matrix(Doc doc, int start, int end):
     return lca_matrix
 
 
+@cython.boundscheck(False)  # Deactivate bounds checking
+cdef void _set_prefix_lengths(
```
danieldk, line 2046: This seems like a nice, reusable utility function to look for offsets of UTF-8 characters, that is not specific to prefix lengths. I think the only non-reusable part is that the output buffer size is predetermined. Does it need to be? What is the overhead of getting all offsets and then slicing the result to get the offsets for the prefix? Especially because you can also reuse the output for the suffix char offsets.

danieldk: Another question (I must admit that at this point I do not have a good-enough overview of the PR yet): couldn't you slice the array in the original Python codepoint representation (or `Py_UCS4` in Cython-land) to the prefix/suffix length and then recode that as UTF-8?

spacy/tokens/doc.pyx
richardpaulhudson: I already tried both using a memoryview and passing in the lengths as parameters, and both slowed things down:

- the memoryview slowed things down massively (when I profiled the code, around 50% of the runtime was memoryview processing, and when I removed the memoryview the run time halved)
- passing in the lengths slowed things down slightly. Presumably passing them in from Python, and converting them from Python to C, takes longer than redetermining them within the C code?

I know a character pointer is somewhat counterintuitive here, but the calling code guarantees that each length occupies exactly one byte.

spacy/tokens/doc.pyx
richardpaulhudson: I think `Doc` is probably the wrong place, but wasn't sure where else to put it: do you have any suggestions?

spacy/tokens/doc.pyx
richardpaulhudson: No, I thought it was best practice to remove them once I'd established that all the tests passed and everything worked as expected?

spacy/tokens/doc.pyx
richardpaulhudson: The aim here was to prevent commands from having to be distributed over too many lines and so becoming hard to read. Which do you think is more important: the readability of the variables or the readability of the commands? I don't really have a strong opinion either way.

spacy/tokens/doc.pyx
richardpaulhudson: The second suggestion was what I started off with, and it was unacceptably slow.

As regards the utility function, I think it's important that the typical configuration doesn't call `set_prefix_lengths()` at all (I'm don't know off the top of my head but I think there are only two languages that use it). Most configurations only use `set_suffix_lengths()` and only need the last few characters in each word.

spacy/tokens/doc.pyx
danieldk: > the memoryview slowed things down massively (when I profiled the code, around 50% of the runtime was memoryview processing, and when I removed the memoryview the run time halved)

Did you try to use the data pointer of the typed memory view? (It could be that the construction overhead is too large.)

spacy/tokens/doc.pyx
danieldk: I had to read the code to understand what `cs` is, so I think it's better to be explicit.

spacy/tokens/doc.pyx
danieldk: I don't think that it is enough. We have had code before that corrupts memory, but passed tests with flying colors 😅 .

spacy/tokens/doc.pyx
danieldk: > The second suggestion was what I started off with, and it was unacceptably slow.

Do you know why? Slicing `Py_UCS4` arrays in C-land is cheap? Was it the recoding as UTF-8? (Which should also be plenty fast.)

spacy/tokens/doc.pyx
shadeMe: Can't it stay in the `richfeatureextrator` module (after converting it to a Cython module)?

richardpaulhudson: @adrianeboyd @svlandeg what do you think of @shadeMe's suggestion? Would putting everything in `richfeatureextractor` conform to all explicit and implicit standards around where things go within spaCy?

spacy/tokens/doc.pyx
richardpaulhudson: Bounds checks have no measurable impact on speed, so I've removed the annotations.

spacy/tokens/doc.pyx
richardpaulhudson: I looked into replacing these with C++ types, but came across the problem (presuming that it makes sense to use the same type for all buffers, which stylistically I would say is the case) that:

- the `_write_hashes()` method has to be able to accept both the raw token string returned from the `StringStore` and the contents of one of these buffers as values of the same variable. This essentially precludes using `std::vector`.
- it would be possible to change the new `StringStore` method to return `std::string` rather than `char*`, but if we want to use `std::string` throughout we have the problem of how to allocate the buffers in the code marked here. While with `char*` we can simply allocate the memory, retaining whatever already happens to be there because the relevant part of it will be overwritten anyway, it seems that with `std:string` we have to initialize the memory with something.

Doesn't `cymem.Pool` give you effective memory management as well?

spacy/tokens/doc.pyx
richardpaulhudson: I had a [solution](https://github.com/richardpaulhudson/spaCy/blob/7d8258bec851c3b6dabfeef874f16fb1847d63dc) based completely around `Py_UCS4` and [PEP 393](https://peps.python.org/pep-0393/), which seemed the most sensible way to approach the problem because it meant fixed-width characters. However, on several occasions I found that the PEP 393 code didn't behave as documented (to mention just a trivial example, the `kind` variable is supposed to have value `4` for four-byte Unicode characters, but actually has the value `3`). I normally look for alternatives to such interfaces, because I know from past experience that you can never know what other problems might be lurking unnoticed and come to bite you once your code is live.

I think it was you who mentioned (internal communication) that it was easy to recognize the boundaries of UTF-8 characters using bitmaps, and basically once I ditched `Py_UCS4` and went straight to the UTF-8 representation of each string I never looked back. I think the important point is that with a typical configuration you are only ever interested in the last few characters of each token. This means that a solution that only looks at the relevant characters at the end (and, where relevant, at the beginning) of each token will always be faster than one that handles the whole string of each token (or of the whole document, which is how I did it in the version linked above.

spacy/tokens/doc.pyx
richardpaulhudson: Creating a Cython memoryview from a Python `bytes` object seems to have a much greater overhead than simply using a pointer to it. I did a quick test passing a two-byte buffer `b"\2\3"` a million times from Python to the methods

```
def test_p(const unsigned char* char_p)
def test_v(const unsigned char[:] char_v)
```

both of which add the values of the two bytes together and return the results: the overall runtime was 3-4x longer using the memoryview than using the pointer. 

Eight separate buffers are passed into the method `get_character_combination_hashes()` in this PR; because the code is generally trimmed for effiency, it makes a big difference whether eight memoryviews have to be created or not.

spacy/tokens/doc.pyx
svlandeg: I can definitely see Daniël's & Madeesh's reasoning to keep this closer to its related functionality - i.e. in the `richfeatureextractor`.

spacy/tokens/doc.pyx
richardpaulhudson: I have now looked into merging `get_prefix_lengths()` and `get_suffix_lengths()` into a single method `get_affix_lengths()` with the aim of copying information common to the prefixes and the suffixes to avoid calculating it twice. However, because the methods calculate the number of bytes for the first/last _n_ characters for each of a list of values of _n_, it turns out that the only time a value is generated that is common to the prefixes and suffixes is when each contains the length of the whole word, and it does not seem likely that testing for this (relatively rare) condition would lead to an overall speed benefit.

spacy/tokens/doc.pyx
richardpaulhudson: I tried moving all the functionality to `richfeatureextractor`, but came across the problem that the `get_character_combination_hashes()` method directly accesses the C/Cython member variables of the `Doc` object it is analyzing, specifically `TokenC* c` and `StringStore vocab.strings`, and needs to do so for speed. Is there a way of accessing these from another Cython module that has been called from Python with a `Doc` object as a parameter? There may well be one I am not aware of as I am far from a Cython expert! If there is none, though, it looks as though the `get_character_combination_hashes()` method may need to stay within `Doc`, although there would obviously be scope for moving the private methods like `_write_hashes()` into a separate module (probably not `richfeatureextractor` in that case though, because that would lead to a dependency loop).

spacy/tokens/doc.pyx
richardpaulhudson: See my comment above.

spacy/tokens/doc.pyx
danieldk: I am not sure what gcc/C++ version we have access to in the wheel builds, but newer C++ versions have `string_view`, which can be used to provide views for both borrowed strings (like strings from the string strore) and owned strings (`std::string`). This can be used to provide a generic API for functions that need to deal with owned and borrowed strings.

richardpaulhudson: `string_view` isn't available via `libcpp`, so using it would require wrapping the relevant library from C++17. I'm not sure whether the additional safety is worth the considerably more complex program architecture (which would be a potential source of bugs in itself). I'd say no, but feel free to go ahead and make the change if you think it makes sense!

svlandeg:
Adding this to our internal backlog for now - though we definitely want to pick this up and finish this!

PR #11592, opened 2022-10-07
Fix issues with DVC commands

polm:
## Description

This fixes a few issues with DVC commands - flags in the wrong place, lack of documentation, weird errors in an edge case. These were originally in #11378 and are partly entangled because they depend a bit on the command representation, but this PR is independent. 

### Types of change

bug fix

## Checklist


spacy/cli/project/dvc.py
```diff
@@ -105,6 +108,14 @@ def update_dvc_config(
         dvc_config_path.unlink()
     dvc_commands = []
     config_commands = {cmd["name"]: cmd for cmd in config.get("commands", [])}
+
+    # some flags that apply to every command
```
svlandeg, line 112: Is there a reason why you removed the check for when both `quiet` and `verbose` were set? I felt it made sense to point out that contradiction when it happened? One thing we could do is make it a warning instead of a failure and opt for quiet as default?

polm: That was an oversight, thanks for catching it! However, it did make me wonder what happens if we pass both of these to DVC. It turns out it will fail with an error anyway:

```
ERROR: argument -q/--quiet: not allowed with argument -v/--verbose
```

So maybe there's no reason for us to catch this? I don't have strong feelings either way.

rmitsch: Do those flags affect anything else than the DVC output?

polm: No, they're just passed through to DVC.

rmitsch: My vote is for passing it through then and not adding another check. I'd let contradictory arguments fail with DVC.

polm: I'll just leave this check out then, so this should be ready to go.

rmitsch (approved):
LGTM.

PR #11593, opened 2022-10-07
Fix example code for spacy-wordnet

polm:
## Description

It looks like in the most recent version of this library, 0.1.0, it's no longer possible to pass the lang parameter to the component separately. Doing so will raise an error.

This change just removes the config argument so the example code works.

### Types of change

docs update

## Checklist


svlandeg (approved):
Nice catch!

website/meta/universe.json
```diff
@@ -2463,7 +2463,7 @@
                 "# Load an spacy model (supported models are \"es\" and \"en\") ",
                 "nlp = spacy.load('en')",
                 "# Spacy 3.x",
-                "nlp.add_pipe(\"spacy_wordnet\", after='tagger', config={'lang': nlp.lang})",
+                "nlp.add_pipe(\"spacy_wordnet\", after='tagger')",
                 "# Spacy 2.x",
```
svlandeg, line 2467:
```suggestion
                "# spaCy 2.x",
```

website/meta/universe.json
```diff
@@ -2463,7 +2463,7 @@
                 "# Load an spacy model (supported models are \"es\" and \"en\") ",
```
svlandeg, line 2463:
```suggestion
                "# Load a spacy model (supported models are \"es\" and \"en\") ",
```

website/meta/universe.json
```diff
@@ -2463,7 +2463,7 @@
...
                 "nlp = spacy.load('en')",
```
svlandeg, line 2464: We should probably update this, too?

polm: It would be good to update it, but then the code wouldn't work with v2, so I guess we can just leave it for now?

polm: Eh, I guess it's pseudocode anyway, I'll add a note.

svlandeg:
Looks good to merge!

PR #11611, opened 2022-10-11
update default configs to recent versions

svlandeg:
## Description
Some default configurations of our components were still using [legacy](https://spacy.io/api/legacy) version for the tok2vec layer.

### Types of change
fix

## Checklist


svlandeg:
Not sure whether we should have this in 3.5 or `master` is fine?

adrianeboyd:
I think v3.5 is better, similar to the language defaults?

PR #11612, opened 2022-10-11
Fix mypy error in edittree lemmatizer

svlandeg:
## Description
Since Thinc 8.1.3,  `mypy` raises an error in the edittree lemmatizer code.

### Types of change
bug fix

## Checklist


spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -150,7 +149,7 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
             # Handle cases where there are no tokens in any docs.
             n_labels = len(self.cfg["labels"])
             guesses: List[Ints2d] = [
-                self.model.ops.alloc((0, n_labels), dtype="i") for doc in docs
+                self.model.ops.alloc2i(0, n_labels, dtype="i") for _ in docs
```
adrianeboyd, line 152: If these two lines aren't equivalent I feel like the error might still be in `thinc`?

```
self.model.ops.alloc((0, n_labels), dtype="i")
self.model.ops.alloc2i(0, n_labels)
```

svlandeg: The issue was the before 8.1.3, `ops.alloc` wasn't properly typed. Now it is properly typed and the error that is being thrown is actually a reasonable one by `mypy`. If you want to ensure you get a specific type back (like here, expecting a `List[2d]`) you can't expect `op.alloc` to work that out based on the tuple format and the `dtype` - that's why we have the more specific methods, right?

svlandeg: So these lines are equivalent in functionality, but not in typing behaviour.

adrianeboyd: For the future, I think we should try to make them equivalent in typing behavior, too, or this will be confusing for users.

adrianeboyd, line 152:
```suggestion
                self.model.ops.alloc2i(0, n_labels) for _ in docs
```

adrianeboyd: There is no harm in changing this here, but I think we should revert the source of this error in the next patch release in thinc and postpone it to v8.2 or v9.

svlandeg: Sure - let's look at that as a separate task!

adrianeboyd: But I wanted this edit in this PR. There's no need to have the type twice.

svlandeg: Sorry, the thread got confusing with the two different topics and I forgot to actually merge the recommendation.

svlandeg: https://github.com/explosion/spaCy/pull/11615

Issue #11626, opened 2022-10-12
handling empty POS and multiple morph annotations in pymorphy2 wrapper


DSLituiev:
## How to reproduce the behaviour
Current pymorphy2 wrapper returns the original form when POS is empty (in `mode="pymorphy2"`) on [line 48](https://github.com/explosion/spaCy/blob/master/spacy/lang/ru/lemmatizer.py#L48), even though there are better fallback solutions

```python
from spacy.lang.uk import Ukrainian
nlp = Ukrainian()
lemmatizer = nlp.add_pipe("lemmatizer", config=
{ #'factory': 'lemmatizer',
 'mode': 'pymorphy2',
 'model': None,
 'overwrite': True,
  # 'scorer': {'@scorers': 'spacy.lemmatizer_scorer.v1'}
 })

nlp("якусь")[0].lemma_
# Output: 'якусь'
```

Similarly with `mode = "pymorphy2_lookup"`, when multiple "analyses" point to the same `normal_form`, the normal form is discarded [on L103](https://github.com/explosion/spaCy/blob/master/spacy/lang/ru/lemmatizer.py#L103), e.g.:

```python

from pymorphy2 import MorphAnalyzer
morph_ = MorphAnalyzer(lang="ru")

morph_.parse("хорошим")
#[
# Parse(word='хорошим', tag=OpencorporaTag('ADJF,Qual masc,sing,ablt'), normal_form='хороший', score=0.636363, methods_stack=((DictionaryAnalyzer(), 'хорошим', 3135, 5),)),
# Parse(word='хорошим', tag=OpencorporaTag('ADJF,Qual neut,sing,ablt'), normal_form='хороший', score=0.181818, methods_stack=((DictionaryAnalyzer(), 'хорошим', 3135, 18),)),
# Parse(word='хорошим', tag=OpencorporaTag('ADJF,Qual plur,datv'), normal_form='хороший', score=0.181818, methods_stack=((DictionaryAnalyzer(), 'хорошим', 3135, 22),))
# ]
```

## Proposed solution

in `pymorphy2_lemmatize`:
```
-            return [string.lower()]
+            return self.pymorphy2_lookup_lemmatize(token)
```

in `pymorphy2_lookup_lemmatize`:

replace

```python
        if len(analyses) == 1:
            return [analyses[0].normal_form]
```

with 

```python
        normal_forms = set([an.normal_form for an in analyses])
        if len(normal_forms) == 1:
            return [next(iter(normal_forms))]
[...]

Issue #11620, opened 2022-10-11
setting `pymorphy2_lemmatize` vs `pymorphy2_lookup_lemmatize` in UkrainianLemmatizer


adrianeboyd:
### Discussed in https://github.com/explosion/spaCy/discussions/11609

<div type='discussions-op-text'>

<sup>Originally posted by **DSLituiev** October 11, 2022</sup>
I would like to run `UkrainianLemmatizer().pymorphy2_lookup_lemmatize()` instead of `UkrainianLemmatizer().pymorphy2_lemmatize()` , as the latter mis-handles lemmatization of certain pronouns.  

I can't quite find where it is called. I cannot see `pymorphy2_lemmatize` mentioned in the config (in `"uk_core_news_lg"`  model), or being called in either `UkrainianLemmatizer` or `RussianLemmatizer`, but assume it can be changed by re-setting the config `"lemmatizer": {'mode': 'pymorphy2', ...}` to  `"lemmatizer": {'mode': 'pymorphy2_lookup',}`.

Here is my attempt that fails:

```python
nlp = spacy.load("uk_core_news_lg")

nlp("якусь")[0].lemma_
# "якусь"

dict(nlp.pipeline)["lemmatizer"].pymorphy2_lemmatize(nlp("якусь")[0])
#  ['якусь']

dict(nlp.pipeline)["lemmatizer"].pymorphy2_lookup_lemmatize(nlp("якусь")[0])
# ['якийсь']

# remove un-used pipeline
nlp.remove_pipe("ner")

old_lemmatizer = nlp.remove_pipe("lemmatizer")

lemmatizer = nlp.add_pipe("lemmatizer", config=
{ #'factory': 'lemmatizer',
 'mode': 'pymorphy2_lookup',
 'model': None,
 'overwrite': True,
  # 'scorer': {'@scorers': 'spacy.lemmatizer_scorer.v1'}
 })


nlp("якусь")[0].lemma_
```

```
  File "/opt/anaconda3/envs/asr/lib/python3.10/site-packages/spacy/lang/ru/lemmatizer.py", line 102, in pymorphy2_lookup_lemmatize
    analyses = self._morph.parse(string)

AttributeError: 'UkrainianLemmatizer' object has no attribute '_morph'

```

I tried adding `nlp.initialize()` but that does not seem to help.

This also does not seem to help `lemmatizer._morph = old_lemmatizer[1]._morph` as it returns empty lemmas: `nlp("якусь")[0].lemma_`

[...]

Issue #11625, opened 2022-10-12
bug due to inconsistent treatment of "DET" vs "PRON" POS in Slavic languages (pymorphy2)


DSLituiev:
## Conceptual issue

It may seem like a naming convention question, but this have some practical implications in implementation of lemmatizers.

In Slavic languages (at least Russian, Ukrainian, [Czech](https://universaldependencies.org/docs/u/pos/DET.html)), *determiner* is not a thing, and it is normally treated as pronoun. However, (at least some) spacy Slavic models treat them as DET. When such POS-tagged document is matched against an external tokenizer, we get a trouble:

## How to reproduce the behaviour
```python
import spacy
nlp = spacy.load("uk_core_news_lg")

doc = nlp("якась жінка") # "some woman"

doc[0].pos_
# DET

doc[0].lemma_
# output: "якась" ("some", f)
# expected "якийсь" ("some", m)
```

This is because during the comparison of the POS assigned by the spacy model and the external morphologizer [`analysis_pos == univ_pos`](https://github.com/explosion/spaCy/blob/master/spacy/lang/ru/lemmatizer.py#L56), `"PRON" != "DET"`.

## Environment

- **spaCy version:** 3.4.1
- **Platform:** macOS-10.16-x86_64-i386-64bit
- **Python version:** 3.10.4
- **Pipelines:** uk_core_news_lg (3.4.0)

DSLituiev:
suggested fix: add extra logic in `pymorphy2_lemmatize`:

before
```
            if analysis_pos == univ_pos or (
                analysis_pos in ("NOUN", "PROPN") and univ_pos in ("NOUN", "PROPN")
            ):
```

after:
```
            if analysis_pos == univ_pos or (
                analysis_pos in ("NOUN", "PROPN") and univ_pos in ("NOUN", "PROPN")
            ) or ((analysis_pos=="PRON") and (univ_pos=="DET")):
```

PR #11627, opened 2022-10-12
refactoring of pymorphy2 wrapper and handling of repeated normal forms with tests

DSLituiev:
## Description
This addresses #11620 and #11626 by:
- refactoring pymorph2 wrapper out of Russian Lemmatizer into the common `spacy/pipeline/lemmatizer.py`. This ensures that initialization code is not replicated between Russian and Ukrainian (and any other languages pymorphy2 may support, like pre-revolution Russian spelling). 
- in the refactored modified `PyMorhpy2Lemmatizer(Lemmatizer)`, the allowed modes are `in {"pymorphy2", "pymorphy2_lookup"}` to address #11620 
- `pymorphy2_lookup` is modified to better handle unambiguous cases where a text form maps to the same normal form, e.g. `хорошим [(plural, dat), (sing, masc, instr), ...] --> хороший`, see #11626
- `pymorphy2` falls back to `pymorphy2_lookup` if POS of the token is empty, instead of returning the text form

### Types of change
bug fix

## Checklist

- [x] My changes don't require a change to the documentation

### DIFF
--- a/spacy/lang/ru/lemmatizer.py
+++ b/spacy/lang/ru/lemmatizer.py
@@ -22,5 +22,5 @@ class RussianLemmatizer(PyMorhpy2Lemmatizer):
         scorer: Optional[Callable] = lemmatizer_score,
     ) -> None:
         super().__init__(
-            vocab, model, lang="ru", name=name, mode=mode, overwrite=overwrite, scorer=scorer,
+            vocab, model, "ru", name=name, mode=mode, overwrite=overwrite, scorer=scorer,
         )
--- a/spacy/lang/ru/lemmatizer.py
+++ b/spacy/lang/ru/lemmatizer.py
@@ -2,8 +2,7 @@ from typing import Optional, List, Dict, Tuple, Callable
 
 from thinc.api import Model
 
-from ...pipeline import Lemmatizer
-from ...pipeline.lemmatizer import lemmatizer_score
+from ...pipeline.lemmatizer import lemmatizer_score, PyMorhpy2Lemmatizer
 from ...symbols import POS
 from ...tokens import Token
 from ...vocab import Vocab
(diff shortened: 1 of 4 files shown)


adrianeboyd:
Thanks for looking at this! Let's separate this into two separate PRs:

1. Minimal fix for `pymorphy2_lookup` mode that will go into v3.4.2 (the current `master` branch). I can do this PR.
2. Changes that modify the lemmatizer behavior, which should go into spacy v3.5 (the current `develop` branch) instead of v3.4.2 because it changes the output. I think we have also already switched `develop` to the default as `pymorphy3`, so we should test any changes with the updated version.

Other notes:

* I'd rather keep `spacy.pipeline.lemmatizer` as only the lookups-table-based lemmatizer that uses `spacy-lookups-data`.
* I think it's okay to keep the core implementation in `RussianLemmatizer` and have `UkrainianLemmatizer` extend it, since this is also similar to the defaults in pymorphy2/3.
* You can use `Literal` as `from spacy.compat import Literal`.
* The class variable `RussianLemmatizer._morph` was intended to save memory if you have multiple Russian pipelines loaded in the same script, but memory is not as limited as it was when this was originally implemented, so we could also simplify this by making it an instance variable.

We're planning on releasing v3.4.2 soon so I'll go ahead and do the first PR today.

adrianeboyd:
I wasn't sure whether you were planning to come back to this PR? If you'd prefer, we can put this on our todo list, but it will be relatively low priority.

Given #11631 and the conflicts in this branch, I think it would be easier to start with a new PR (currently against `master` for v3.5), and we'll continue to track the issues under #11625 and #11626.

I'll close this for now, but feel free to reopen or submit a new PR!

DSLituiev:
@adrianeboyd from your comment, I understood that you had plans to modify it. I am open to help, but I am not sure how I can be helpful giving you have multiple versions being baked. Please let me know.

DSLituiev:
Are you suggesting pushing it to `develop` branch? I can do that. I would still prefer to keep inheritance from a common pymorph2 lemmatizer class, rather than Ru>Ukr. If you prefer to keep `spacy.pipeline.lemmatizer` with only one class, I can place it in a separate file within the same folder.

PR #11640, opened 2022-10-13
Handle Docs with no entities in EntityLinker

polm:
## Description

If a whole batch contains no entities it won't make it to the model, but it's possible for individual Docs to have no entities. Before this commit, those Docs would cause an error when attempting to concatenate arrays because the dimensions didn't match.

It turns out the process of preparing the Ragged at the end of the span maker forward was a little different from list2ragged, which just uses the flatten function directly. Letting list2ragged do the conversion avoids the dimension issue.

This did not come up before because in NEL demo projects it's typical for data with no entities to be discarded before it reaches the NEL component.

This includes a simple direct test that shows the issue and checks it's resolved. It doesn't check if there are any downstream changes, so a more complete test could be added. A full run was tested by adding an example with no entities to the Emerson sample project.

Leaving as a draft until I figure out what to do about a fuller test.

### Types of change

bug fix

## Checklist


polm:
About full testing - since this is basic functionality, rather than adding an end-to-end test just for this, I added it to the default training data set so that existing and future tests cover it. This is in addition to the unit test that should make the problem clear if there's a regression for some reason.

spacy/ml/models/entity_linker.py
```diff
@@ -20,7 +20,7 @@ def build_nel_encoder(
         token_width = tok2vec.maybe_get_dim("nO")
         output_layer = Linear(nO=nO, nI=token_width)
         model = (
-            ((tok2vec >> list2ragged()) & build_span_maker())
+            ((tok2vec >> list2ragged()) & (build_span_maker() >> list2ragged()))
```
adrianeboyd, line 23: I think you'd need to increase the architecture version for this kind of change?

polm: Ah, that's right, this will break serialization. This is cleaner, but let me see if I can use `list2ragged` internally to avoid a version bump, since this is mainly a bugfix.

polm: OK, I have made some simple changes so that the architecture is unchanged and this does not require a version increase.

polm:
If the tests pass, which it looks like they will, this should be fine and no longer require a version increase in the architecture.

rmitsch (approved):
This looks fine to me. Let's wait with merging until @adrianeboyd gives her go-ahead as well though?

Issue #11641, opened 2022-10-13
spacy[apple] cannot load en_core_web_trf


pudo:
As soon as `spacy.prefer_gpu()` is enabled, loading the `en_core_web_trf` model will fail with the stack trace below. 

## How to reproduce the behaviour

```python
import spacy

spacy.prefer_gpu()
# disable everything but NER:
nlp = spacy.load(
  "en_core_web_trf",
   disable=["tok2vec", "tagger", "parser", "attribute_ruler", "lemmatizer"]
)
nlp.add_pipe("sentencizer")
```

## Your Environment
* Operating System: macOS-12.5.1-arm64-arm-64bit
* Python Version Used: Python 3.10.6 (homebrew)
* spaCy Version Used: spacy[apple]==3.4.1
* Pipelines: en_core_web_sm (3.4.0), en_core_web_trf (3.4.0)

## Exception

```
  File "/Users/pudo/Code/storyweb/storyweb/pipeline.py", line 26, in load_nlp
    nlp = spacy.load(
  File "/Users/pudo/.virtualenvs/wrangle/lib/python3.10/site-packages/spacy/__init__.py", line 54, in load
    return util.load_model(
  File "/Users/pudo/.virtualenvs/wrangle/lib/python3.10/site-packages/spacy/util.py", line 429, in load_model
    return load_model_from_package(name, **kwargs)  # type: ignore[arg-type]
  File "/Users/pudo/.virtualenvs/wrangle/lib/python3.10/site-packages/spacy/util.py", line 465, in load_model_from_package
    return cls.load(vocab=vocab, disable=disable, enable=enable, exclude=exclude, config=config)  # type: ignore[attr-defined]
  File "/Users/pudo/.virtualenvs/wrangle/lib/python3.10/site-packages/en_core_web_trf/__init__.py", line 10, in load
    return load_model_from_init_py(__file__, **overrides)
  File "/Users/pudo/.virtualenvs/wrangle/lib/python3.10/site-packages/spacy/util.py", line 646, in load_model_from_init_py
    return load_model_from_path(
  File "/Users/pudo/.virtualenvs/wrangle/lib/python3.10/site-packages/spacy/util.py", line 511, in load_model_from_path
    return nlp.from_disk(model_path, exclude=exclude, overrides=overrides)
  File "/Users/pudo/.virtualenvs/wrangle/lib/python3.10/site-packages/spacy/language.py", line 2115, in from_disk
[...]

PR #11666, opened 2022-10-18
Add warning to install widget for M1 GPUs

polm:
## Description

Currently support for M1 GPUs is experimental and limited, but the install widget allows you to select the M1/GPU combination as though everything will work. This adds a warning about the experimental status and links to the FAQ on the subject. This should prevent issues like #11641.

The way this displays is not ideal - for example, there's a link but it's not obvious because it's not underlined or anything. I will see if there's a way to modify the CSS to make it clearer, but any suggestions on other places to display this would also be welcome.

## Checklist

### DIFF
--- a/website/src/widgets/quickstart-install.js
+++ b/website/src/widgets/quickstart-install.js
@@ -201,7 +201,13 @@ const QuickstartInstall = ({ id, title }) => {
                             {nightly ? ' --pre' : ''}
                         </QS>
                         <QS package="conda">conda install -c conda-forge spacy</QS>
-                        <QS package="conda" hardware="gpu">
+                        <QS package="conda" hardware="gpu" os="windows">
+                            conda install -c conda-forge cupy
+                        </QS>
+                        <QS package="conda" hardware="gpu" os="linux">
+                            conda install -c conda-forge cupy
+                        </QS>
+                        <QS package="conda" hardware="gpu" os="mac" platform="x86">
                             conda install -c conda-forge cupy
                         </QS>
                         <QS package="conda" config="train">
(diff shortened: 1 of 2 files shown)


website/src/widgets/quickstart-install.js
```diff
@@ -160,7 +160,7 @@ const QuickstartInstall = ({ id, title }) => {
                         showDropdown={showDropdown}
                     >
                         <QS os="mac" hardware="gpu" platform="arm">
-                            # Note M1 GPUs support is experimental, see <a href="https://github.com/explosion/spaCy/discussions/11436">#11436</a>
+                            # Note M1 GPUs support is experimental, see <a href="https://github.com/explosion/thinc/issues/792">this issue</a>
```
adrianeboyd, line 163:
```suggestion
                            # Note M1 GPU support is experimental, see <a href="https://github.com/explosion/thinc/issues/792">this issue</a>
```

polm:
I figured out how to modify the CSS to underline the URL. I don't think this affects any other documentation, or if it does I can't imagine it would be an issue, so I think this is OK?

adrianeboyd:
While we're looking at this, can you take `conda install -c conda-forge cupy` out of the M1+GPU options?

website/src/widgets/quickstart-install.js
```diff
@@ -159,6 +159,9 @@ const QuickstartInstall = ({ id, title }) => {
                         setters={setters}
                         showDropdown={showDropdown}
                     >
+                        <QS os="mac" hardware="gpu" platform="arm">
+                            # Note M1 GPU support is experimental, see <a href="https://github.com/explosion/thinc/issues/792">this issue</a>
```
adrianeboyd, line 163: I'd like link text looks more informative than basically "here", but I'm not particularly attached to this particular suggestion:

```suggestion
                            # Note M1 GPU support is experimental, see <a href="https://github.com/explosion/thinc/issues/792">Thinc issue #792</a>
```

polm: Looks good to me!

polm:
Removed `cupy` from the M1+GPU options.

PR #11668, opened 2022-10-18
Updates for python 3.10 and 3.11, set version to v2.3.8

adrianeboyd:
## Checklist


adrianeboyd:
This removes the generated cython files from the sdist, which has been a longer-standing problem for people trying to use spacy v2 with python 3.10+. (Downgrading python was better advice as long as wheels were lacking for many dependencies, but now there should be wheels.)

PR #11669, opened 2022-10-18
Prototype: Entity linking with custom KBs

rmitsch:
## Description

The current implementation of the candidate generation (CG) step in `develop` includes a batched version of CG. Following up on some internal experiments, it seems likely that information on the entire context of a document is useful for both CG and disambiguation. The batched version is insufficient to this end, since it doesn't pass on information on which entities/spans are part of which document to downstream steps of the NEL process.

This PR hence introduces a mode for CG lookups per document(s), with which entities are passed to the CG function as a `Generator[Iterable[Span], None, None]` - i.e. it is clear to the CG lookup which entities/spans belong to the same document. This mode can be enabled via the new config parameter `candidates_doc_mode` (alternative naming suggestions welcome). 

The NEL config still defaults to its previous values, so this is backwards-compatible with the current state of `master`, but not entirely for `develop` (which in itself is already slightly breaking for `KnowledgeBase`-related changes): 
- The NEL config parameter `candidates_batch_size` has been dropped, since batching is entirely up to the `KnowledgeBase` implementation now.
- Error `E1044` has been removed, since it was exclusively related to `candidates_batch_size`.
- `get_candidates_batch()` has been removed to `get_candidates_all()`.

This is a necessary precursor to changes in the CG and disambiguation logic and model architecture to incorporate additional information about the document context.

Note: the documentation will be updated once the code has been informally approved.

### Types of change

Enhancement.

## Checklist


polm:
This is an interesting change. There are a couple of things about it that aren't clear to me.

The candidate generator isn't the only source of information about the input to the EntityLinker architecture, right? The component gets the Docs and passes them to the model, while making use of the KB. If the original Docs are present, is it necessary to do this grouping in the KB, rather than in the component or the model? If it could be done either way, why is this better? I don't have a strong opinion for one or the other, but normally I would expect the issue of building representations, including inferring global Doc features, should be left up to the model (/component) internals. 

Also, for the CG internally, if the input to it is Spans, it's already possible to look up the Doc from a Span. So it's not clear to me why that needs to be in the API structure either.

Looking at the implementation so far, it looks like this Doc information is in fact available where these methods are called anyway. 

Separately, if I understand correctly, the batch mode was designed to make the candidate generation more efficient. But this change seems primarily motivated by accuracy / information concerns, which makes it a bit confusing to me that it's a replacement. I can see how grouping by Doc might be more efficient in candidate generation too (by re-using Doc representations), but again that seems like something we could be more agnostic about from the outside. 

Basically, I understand that grouping by Doc can be useful for building features, but it's not clear to me why that should be a built-in assumption in candidate generation at the API level. As a default implementation, size-based batching seems more general. (That said, since the batching itself is brand-new and unreleased we don't have to worry too much about changing from it at this point.)

rmitsch:
Thanks for your feedback!

> The candidate generator isn't the only source of information about the input to the EntityLinker architecture, right?

Correct. Currently the EL architecture uses the pre-generated entity embedding and the mention embeddings from trained encoder.

>  If the original Docs are present, is it necessary to do this grouping in the KB, rather than in the component or the model? If it could be done either way, why is this better? I don't have a strong opinion for one or the other, but normally I would expect the issue of building representations, including inferring global Doc features, should be left up to the model (/component) internals.

The primary motivation of this PR is that we want to able to leverage the document context, i.e. the other entities in the doc, also in the candidate generation (CG). I agree that building representations for the model itself should be left for the model, but in this case we want to improve the CG heuristics by exploiting information about the candidates for other entities in this document.

> Also, for the CG internally, if the input to it is Spans, it's already possible to look up the Doc from a Span. So it's not clear to me why that needs to be in the API structure either.

It's not necessary to obtain information about the other entities, that is correct. However, we are (at least for experiments) also interested in the _candidates_ for these other entities, since ideally we identify those combinations of candidates with high similarity to the entity mentions as well as high coherence with each other. 

> Separately, if I understand correctly, the batch mode was designed to make the candidate generation more efficient. But this change seems primarily motivated by accuracy / information concerns, which makes it a bit confusing to me that it's a replacement. I can see how grouping by Doc might be more efficient in candidate generation too (by re-using Doc representations), but again that seems like something we could be more agnostic about from the outside.

I can see how this is confusing. The performance aspect is still considered in this PR - in "doc mode", the candidate generator is called only once for all entities in all documents. All batching is hence up to the CG implementation, which still allows for the efficiency gains possible in the "batch mode". Depending on the requirements, the implementation can choose to process one entity after the other, everything at once or anything in between. 

> Basically, I understand that grouping by Doc can be useful for building features, but it's not clear to me why that should be a built-in assumption in candidate generation at the API level. As a default implementation, size-based batching seems more general. (That said, since the batching itself is brand-new and unreleased we don't have to worry too much about changing from it at this point.)

The key point is that batching in arbitrary sizes, i.e. regardless of document association, means that calling the CG with a batch could entail that some documents are split - the CG won't have _direct_ access to all entities. Of course, as you mentioned, we could obtain those missing entities via the `doc` reference in the span object. Since our goal is to consider _all possible viable candidates for all entities_ in a document when selecting candidates and returning them to the entity linker, this makes handling non-doc-aware batching a bit awkward: we would have to do the work of fetching candidates for some particular mentions twice or "pre-fetch" entities from the next batch.

So in a nutshell: IMO grouping entities by doc and passing them to the CG only once allows the same efficiency increase as repeatedly calling CG with arbitrarily sized batches, but has the advantage of always having all entities for a doc in your batch (and hence making stuff easier if you are aiming to leverage information from other entity (candidates) in generating candidates).

Does that make sense to you?

spacy/kb/candidate.pyx
```diff
@@ -64,11 +64,13 @@ def get_candidates(kb: KnowledgeBase, mention: Span) -> Iterable[Candidate]:
     return kb.get_candidates(mention)
 
 
-def get_candidates_batch(kb: KnowledgeBase, mentions: Iterable[Span]) -> Iterable[Iterable[Candidate]]:
+def get_candidates_all(
+    kb: KnowledgeBase, mentions: Generator[Iterable[Span], None, None]
+) -> Generator[Iterable[Iterable[Candidate]], None, None]:
```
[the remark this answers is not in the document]
rmitsch: Sure. I think `Generator[Something, None, None]` is more informative, but it's also a bit unwieldy.

polm:
Thanks for the extra explanation, that does make sense. The part I was missing was that in generating candidates you would consider not just general Doc information, but also previously generated candidate lists. I assumed "document context" was like a document vector.

> this makes handling non-doc-aware batching a bit awkward: we would have to do the work of fetching candidates for some particular mentions twice or "pre-fetch" entities from the next batch.

This totally makes sense and makes it clear why this API design has an advantage. (I think the way I would handle it would be a Doc-keyed LRU cache, but then you have to worry about memory usage there, which is also awkward and has problems.)

rmitsch:
> (I think the way I would handle it would be a Doc-keyed LRU cache, but then you have to worry about memory usage there, which is also awkward and has problems.)

Ah, hadn't considered that. That would work too, I think. But yeah, I think grouping by docs is less of a hassle.

rmitsch:
@polm Any more comments? Otherwise I'd consider the interface as stable and start updating the docs.

polm:
Ah, sorry for not being more clear - interface should be fine, and you can go ahead with working on it. 

One thing I would like to look at more that doesn't affect the API is the logic Docs without entities or Docs without tokens - I'm not sure all the iterators line up quite right. But since this is a draft I guess we'll go over that and more tests can be added.

rmitsch:
> One thing I would like to look at more that doesn't affect the API is the logic Docs without entities or Docs without tokens - I'm not sure all the iterators line up quite right. But since this is a draft I guess we'll go over that and more tests can be added.

Ah, good point. I'll add tests for that.

spacy/pipeline/entity_linker.py
```diff
@@ -440,96 +458,104 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
-                        sentence_encoding_t = sentence_encoding.T
-                        sentence_norm = xp.linalg.norm(sentence_encoding_t)
-                    entity_count += 1
-                    if ent.label_ in self.labels_discard:
-                        # ignoring this entity - setting to NIL
+        for doc_idx, doc in enumerate(docs):
+            if len(doc) == 0:
+                continue
+            sentences = [s for s in doc.sents]
+            doc_ent_cands = list(next(all_ent_cands)) if len(doc.ents) else []
```
polm, line 497: On the topic of empty Docs, is the iteration here OK?

If I understand correctly, `all_ent_cands` has one element per Doc. Those elements (lists) can be empty if a Doc has no entities, but they are always present.

Here the input Docs and `all_ent_cands` are iterated separately. Because of the early `continue` for Docs with no tokens, wouldn't that result in a misalignment, because the `all_ent_cands` generator wouldn't be advanced?

rmitsch: So I'll added a [test to check for this](https://github.com/explosion/spaCy/pull/11669/files#diff-60c79e322819366d1dcdbec89602b0c008dc4b8f15832da6ca6ffe8ebf666fb6R1214), but I'm running into the problem that motivated https://github.com/explosion/spaCy/pull/11640 - perhaps we can merge that that?

polm: #11640 has been merged, so I guess this should be OK now once that's incorporated?

rmitsch: Just merged - tests succeed locally.

svlandeg:
> The NEL config still defaults to its previous values, so this is backwards-compatible with the current state of `master`, but not entirely for `develop`

I think that's fine - because no version of `develop` has already been released?

Also, actually, the current `master` equals the old `develop` - i.e. the branch from which we'll release 3.5.

rmitsch:
> I think that's fine - because no version of develop has already been released?

That's my understanding as well.

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -496,43 +496,59 @@ def create_kb(vocab):
     doc = nlp(text)
     assert doc[0].ent_kb_id_ == "NIL"
     assert doc[1].ent_kb_id_ == ""
-    assert doc[2].ent_kb_id_ == "Q2"
+    # todo It's unclear why EL doesn't learn properly for this test anymore (scores are 0). Seemed to work before, but
+    #  no relevant changes in EL code were made since these tests were added AFAIK (CG seems to work fine).
+    assert doc[2].ent_kb_id_ in ("Q2", "Q3")
```
rmitsch, line 501: It's unclear why EL doesn't link Q2 reliably anymore here (scores are 0). No relevant changes in EL code were made since these tests were added AFAIK (CG seems to work fine).

spacy/kb/candidate.py
```diff
@@ -0,0 +1,114 @@
...
+    """`Candidate` for InMemoryLookupKBCandidate."""
+
+    # todo
+    #  - glue together
+    #  - is candidate definition necessary for EL? as long as interface fulfills requirements, this shouldn't matter.
+    #    otherwise incorporate new argument.
+    #  - fix test failures (100% backwards-compatible should be possible after changing EntityLinker)
+    def __init__(
+        self,
+        retrieve_string_from_hash: Callable[[int], str],
```
rmitsch, line 57: Breaking change, done to avoid circular dependencies (`candidate.py` <-> `kb.py`).

spacy/kb/kb.pyx
```diff
@@ -84,25 +88,45 @@ cdef class KnowledgeBase:
...
+        path (Union[str, Path]): Target file path.
+        exclude (Iterable[str]): List of components to exclude.
         """
-        Load KnowledgeBase content from disk.
+        raise NotImplementedError(
+            Errors.E1044.format(parent="KnowledgeBase", method="from_disk", name=self.__name__)
+        )
+
+    @classmethod
+    def generate_from_disk(
```
rmitsch, line 113: For convenience: factory method. Originally I thought this would be a necessary part of our new EL workflow - turns out it's not. I think this is useful anyway though, and since it's not an abstract method per se, it's not mandated for sub-classes.

spacy/kb/kb_in_memory.pyx
```diff
@@ -238,7 +246,7 @@ cdef class InMemoryLookupKB(KnowledgeBase):
         alias_index = <int64_t>self._alias_index.get(alias_hash)
         alias_entry = self._aliases_table[alias_index]
 
-        return [Candidate(kb=self,
+        return [Candidate(retrieve_string_from_hash=self.vocab.strings.__getitem__,
```
rmitsch, line 249: Mentioned above, but: this is a breaking change (although I doubt this would be noticed) and I don't particularly like this solution. It'd be cleaner to just pass the alias, since I don't think it'd occur a significant memory overhead to have `Candidate` instances store their mentions/aliases instead of looking them up. This would be even more (?) breaking, as it would eliminate the need for hashes. 

As mentioned I don't think many devs are writing their own KBs (deriving from the published spaCy KB) anyway, so the more breaking version would be preferrable to me.

spacy/ml/models/entity_linker.py
```diff
@@ -99,13 +99,44 @@ def empty_kb_factory(vocab: Vocab):
...
 
+@registry.misc("spacy.EmptyKB.v2")
+def empty_kb_for_config() -> Callable[[Vocab, int], KnowledgeBase]:
+    def empty_kb_factory(vocab: Vocab, entity_vector_length: int):
+        return InMemoryLookupKB(vocab=vocab, entity_vector_length=entity_vector_length)
+
+    return empty_kb_factory
+
+
+def get_candidates(kb: KnowledgeBase, mention: Span) -> Iterable[Candidate]:
```
rmitsch, line 110: Moved this and `get_candidates_all()` here to avoid circular dependency (`candidate.py` <-> `kb.py`) - I don't think this matters?

svlandeg:
@rmitsch : I protected this branch by locking it.

PR #11696, opened 2022-10-24
Update textcat scorer threshold behavior

adrianeboyd:
## Description

For `textcat` (with exclusive classes) the scorer should always use a threshold of 0.0 because there should be one predicted label per doc and the numeric score for that particular label should not matter.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/pipeline/textcat.py
+++ b/spacy/pipeline/textcat.py
@@ -72,7 +72,7 @@ subword_features = true
     "textcat",
     assigns=["doc.cats"],
     default_config={
-        "threshold": 0.5,
+        "threshold": 0.0,
         "model": DEFAULT_SINGLE_TEXTCAT_MODEL,
         "scorer": {"@scorers": "spacy.textcat_scorer.v1"},
     },
--- a/spacy/tests/pipeline/test_textcat.py
+++ b/spacy/tests/pipeline/test_textcat.py
@@ -826,7 +826,7 @@ def test_textcat_loss(multi_label: bool, expected_loss: float):
 def test_textcat_threshold():
     # Ensure the scorer can be called with a different threshold
     nlp = English()
-    nlp.add_pipe("textcat")
+    nlp.add_pipe("textcat_multilabel")
 
     train_examples = []
     for text, annotations in TRAIN_DATA_SINGLE_LABEL:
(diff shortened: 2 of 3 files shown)


adrianeboyd:
There's no practical way to remove the `threshold` setting from the `textcat` config at this point. One option could be to make the threshold optional and set the default to `null` so that it's clearer to users that it's unused rather than that `0.0` matters.

spacy/pipeline/textcat.py
```diff
@@ -72,7 +72,7 @@
     "textcat",
     assigns=["doc.cats"],
     default_config={
-        "threshold": 0.5,
+        "threshold": 0.0,
```
rmitsch, line 75: I think `0.0` as default is fine - either way can be somewhat misleading, depending on whether this is `multi_label`or not.

spacy/tests/pipeline/test_textcat.py
```diff
@@ -826,7 +826,7 @@ def test_textcat_loss(multi_label: bool, expected_loss: float):
 def test_textcat_threshold():
     # Ensure the scorer can be called with a different threshold
     nlp = English()
-    nlp.add_pipe("textcat")
+    nlp.add_pipe("textcat_multilabel")
```
rmitsch, line 829: Should rename the function to `test_textcat_multilabel_threshold()`?

rmitsch (approved):

spacy/pipeline/textcat.py
adrianeboyd: `textcat` should never be `multi_label` (well, unless you've hacked a bunch of things into meaninglessness). `multi_label=False` is hard-coded in the default scorer.

rmitsch: Ah, right. Then I'm in favor of `None` as default `threshold` value for `textcat`, but I'm ok with `0` as well.

adrianeboyd: I changed it to `Optional[float]` for `textcat` but it makes the joint API docs tricky, so I'm still not sure if we want to do this in the end.

rmitsch: Let's stick with `0` if it's too much of a bother. The docstring documents this sufficiently well, I think.

rmitsch:
@adrianeboyd Changes since the last review look good to me. Anything else you want to add? Otherwise we're good to go.

PR #11698, opened 2022-10-24
Adjust default attrs for textcat configs

adrianeboyd:
## Description

* Use `NORM` instead of `ORTH`
* Remove `ID`
* Reduce rows for `PREFIX` and `SHAPE`

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/pipeline/textcat.py
+++ b/spacy/pipeline/textcat.py
@@ -24,8 +24,8 @@ single_label_default_config = """
 [model.tok2vec.embed]
 @architectures = "spacy.MultiHashEmbed.v2"
 width = 64
-rows = [2000, 2000, 1000, 1000, 1000, 1000]
-attrs = ["ORTH", "LOWER", "PREFIX", "SUFFIX", "SHAPE", "ID"]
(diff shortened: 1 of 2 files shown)


adrianeboyd:
`ID` is an unexpected attribute. If there are no static vectors it's the same for all tokens; if there are static vectors it corresponds to the row in the vector table, which does have an inverse correlation with frequency, but I suspect that something-resembling-frequency or something-resembling-OOV is not a particular useful feature for textcat models.

Anyone is welcome to contradict me here?

adrianeboyd:
Reducing the rows for `PREFIX` vs. `SUFFIX` makes sense for all langauges given the default settings. Reducing the number of rows for `PREFIX` to this degree makes more sense for English than for Chinese.

polm:
That all sounds reasonable to me. I am a little surprised ID was in there as well, but I think it was in a default list of attributes somewhere at some point - maybe it got copy/pasted? Should be fine to remove in any case.

polm (approved):

PR #11705, opened 2022-10-26
More version updates for github action deprecation warnings

ryndaniels:
A couple more fixes for the deprecation warnings. 

There will be one more after this - `lock.yml` uses an action that hasn't pushed a fix for this yet. Once they do I'll make that last fix.

### DIFF
--- a/.github/workflows/autoblack.yml
+++ b/.github/workflows/autoblack.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
             ref: ${{ github.head_ref }}
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v4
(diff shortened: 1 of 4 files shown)


adrianeboyd:
Should this PR stay open until there's a fix for the last issue?

ryndaniels:
@adrianeboyd whichever you'd prefer - we have I think at least 6 months until these deprecation warnings will become problems so there's not a huge rush - if you'd prefer that we wait and do just one PR with everything, I'm happy to do that.

ryndaniels:
Looking at [this issue](https://github.com/dessant/lock-threads/issues/36) on `lock-threads` it might make make sense to merge this PR now, since that repo hasn't had any commits in the past year so it might be some time before it gets fixed.

PR #11711, opened 2022-10-27
Require annotation decorator

kadarakos:
Decorator to require annotations on a `Doc` or `Span` for a function.

## Description
The `syntax_iterators.py` for a couple of languages uses the `if not Doc.has_annotation("ATTR")` then `raise` a `ValueError` saying that we need this attribute for the function to work. Since the `noun_chunk` implementations in `syntax_iterators.py` forget to check `Doc.has_annotation("POS")` and only check for `"DEP"` they do not exit gracefully with an informative error message. To make it more ergonomic to describe annotation requirements of a function this PR introduces a `require_annotation` decorator that can be used like:

```python
@require_annotation(
    ["TAG", "DEP", "ENT_IOB"],
    require_complete=[True, True, False]
)
def do_stuff(doc: Union[Doc, Span]) -> Any:
       do something
```

In this example the `"TAG", "DEP", "ENT_IOB"` is expected by the function and its expected that all tokens are annotated with `"TAG"` and `"DEP"` information, but `"ENT_IOB"` can be partial. The hope is that this pattern makes it easy to get the requirements of a function correct right at definition.

### Types of change
Small improvement for error handling.

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -952,6 +952,7 @@ class Errors(metaclass=ErrorsWithCodes):
              "sure it's overwritten on the subclass.")
     E1046 = ("{cls_name} is an abstract class and cannot be instantiated. If you are looking for spaCy's default "
              "knowledge base, use `InMemoryLookupKB`.")
+    E1047 = ("The function `{name}` requires annotations {annotations} to be set, but did not find {missing}.")
 
 
 # Deprecated model shortcuts, only used in errors and warnings
--- a/spacy/tests/test_misc.py
+++ b/spacy/tests/test_misc.py
@@ -1,14 +1,17 @@
 import pytest
 import os
 import ctypes
+import spacy
+from typing import Union
 from pathlib import Path
 from spacy.about import __version__ as spacy_version
 from spacy import util
 from spacy import prefer_gpu, require_gpu, require_cpu
+from spacy.tokens import Doc, Span
 from spacy.ml._precomputable_affine import PrecomputableAffine
 from spacy.ml._precomputable_affine import _backprop_precomputable_affine_padding
 from spacy.util import dot_to_object, SimpleFrozenList, import_file
-from spacy.util import to_ternary_int
+from spacy.util import to_ternary_int, require_annotation
 from thinc.api import Config, Optimizer, ConfigValidationError
 from thinc.api import get_current_ops, set_current_ops, NumpyOps, CupyOps, MPSOps
 from thinc.compat import has_cupy_gpu, has_torch_mps_gpu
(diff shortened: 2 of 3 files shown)


adrianeboyd:
My initial thought is that it might make sense to have two separate decorators, one for complete annotation and one for partial annotation, which would simplify the API a lot.

We would still need to carefully compare the speed of using the decorator vs. a direct check added to `noun_chunks`.

kadarakos:
> My initial thought is that it might make sense to have two separate decorators, one for complete annotation and one for partial annotation, which would simplify the API a lot.

Do you mean something like `@require_annotation(["DEP", "ENT_IOB"])` and then another `@require_complete_annotation(["DEP", "ENT_IOB"])`? I think that makes sense, but also then you cannot say that you want the `"DEP"` to be complete, but its fine if the `ENT_IOB` is partial.

adrianeboyd:
Can't you use two decorators as:

```
@require_annotation(["ENT_IOB"])
@require_complete_annotation(["DEP"])
def whatever():
```

I don't know about the speed, and the speed may be enough of a concern even for a single decorator that we don't want to do this at all.

kadarakos:
> Can't you use two decorators as:
> 
> ```
> @require_annotation(["ENT_IOB"])
> @require_complete_annotation(["DEP"])
> def whatever():
> ```
> 
> I don't know about the speed, and the speed may be enough of a concern even for a single decorator that we don't want to do this at all.

ah yeah, you are right you could do this. hmm okay, I'll implement a comparison for the `noun_chunk` use case then. Are decorators known to be slow?

kadarakos:
I did a test with replacing the English `noun_chunks` with a version that uses the `requires_annotation` instead of the `if` statements. It did make it 10x slower unfortunately. I tried on two test documents running both function 1000 times. This is the timing results:

```
old time 2.952050417661667e-07
new time 2.375354990363121e-06
old time 2.3197499103844167e-07
new time 2.309122122824192e-06
```

kadarakos:
Created a gist that runs a speed test to compare the English `noun_chunks` iterator with the `if` statements as it is currently implemented and with the proposed `require_annotation` decorator: https://gist.github.com/kadarakos/92f2f1335068c31ddd7672805eb2e1f0

The output of the script on my machine:

```
old time 3.335881046950817e-07
new time 1.5152569394558669e-06
slowdown 4.542299075205025
old time 2.7916417457163333e-07
new time 1.4653911348432302e-06
slowdown 5.249209133270114
```

If I wrote the test correctly, it does seem like the decorated version is 4-5x slower.

kadarakos:
Closing this PR due to the the decorator solution being slower than writing out the `if` statements explicitly.

kadarakos:
Closing this PR due to the the decorator solution being slower than writing out the `if` statements explicitly.

PR #11713, opened 2022-10-28
Fix default parameters for load functions (fix #11706)

### LINKED ISSUE #11706: UserWarning W123 printed on console after upgrading to 3.4.2
After upgrading to `3.4.2` I see this warning on the console when a model is loaded. How to fix or silence this?

> spacy/language.py:1895: UserWarning: [W123] Argument disable with value [] is used instead of ['senter'] as specified in the config. Be aware that this might affect other components in your pipeline.

```python
import en_core_web_sm
lang_model = en_core_web_sm.load()
```

## Your Environment

- **spaCy version:** 3.4.2
- **Platform:** macOS-13.0-arm64-arm-64bit
- **Python version:** 3.10.6
- **Pipelines:** en_core_web_md (3.4.1), en_core_web_sm (3.4.1), en_core_web_lg (3.4.1)

polm:
## Description

Some load functions used SimpleFrozenList() directly instead of the _DEFAULT_EMPTY_PIPES parameter. That mostly worked as intended, but the changes in #11459 check for equality using identity, not value, so a warning is incorrectly raised sometimes, as in #11706.

This change just has all the load functions use the singleton value instead.

### Types of change

minor bugfix

## Checklist

### DIFF
--- a/.github/azure-steps.yml
+++ b/.github/azure-steps.yml
@@ -73,6 +73,11 @@ steps:
     condition: eq(variables['python_version'], '3.8')
 
   - script: |
+      python -W error -c "import ca_core_news_sm; nlp = ca_core_news_sm.load(); doc=nlp('test')"
+    displayName: 'Test no warnings on load (#11713)'
+    condition: eq(variables['python_version'], '3.8')
+
+  - script: |
       python -m spacy convert extra/example_data/ner_example_data/ner-token-per-line-conll2003.json .
     displayName: 'Test convert CLI'
     condition: eq(variables['python_version'], '3.8')
--- a/spacy/util.py
+++ b/spacy/util.py
@@ -443,9 +443,9 @@ def load_model_from_package(
     name: str,
     *,
     vocab: Union["Vocab", bool] = True,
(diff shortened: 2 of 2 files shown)


rmitsch (approved):
Thanks for fixing this! This can be merged, right?

adrianeboyd:
Can you add a test for this? I think the easiest place would be the extra `ca_core_news_sm` tests that are run for one CI instance at the end of the CI steps. You probably have to check a bit to be sure you're detecting (the absence of) the intended warnings, but I think there's at least one test that does something similar.

polm:
The new test fails on OSX with Python 3.8, and this isn't transient. There's an error about trying to load spaCy twice. Maybe this is related to some of the vendored libraries? It probably is a pre-existing error and not tied to this PR.

adrianeboyd:
It's not specific to macos or python 3.8 (probably), it's that the extra tests that involve downloading models are only run for python 3.8.

adrianeboyd:
#11738 cleans up the working directory to address the unrelated warnings that were causing the new CI test to fail.

PR #11714, opened 2022-10-28
Corrected typos

gremur:


adrianeboyd:
Thanks for the PR! I merged in the current `master` and it looks like we ended up reviewing PRs a bit of out of order and this same change was also in #11753. It's nice to see that everyone is in agreement, in any case!

Issue #11708, opened 2022-10-27
Is it possible to loosen the upper-bound constraint on Typer?


amitkparekh:
Hi! I'd like to use spaCy in a project where I already used Typer v0.6.2. 

I cannot install spaCy because spaCy forces all Typer versions to be the following: 
https://github.com/explosion/spaCy/blob/865691d169c3be413007f0d7324e03a7aac3b3cb/requirements.txt#L12

I use Poetry as my dependency manager, which means that it will not allow the inclusion of spaCy as a dependency with the constraint, even if I do not want to use the CLI aspect of spaCy. 

Is there a reason for limiting Typer to <0.5.0, and is it possible to loosen this constraint? 

Thank you!

polm:
Thanks for the note! I am not aware of a specific reason for this, but in general we try to avoid overly open version specifications, and we've had some version-related issues with Typer in the past. 

We'll look into increasing this, but it might take a bit.

PR #11720, opened 2022-10-31
Raise Typer limit

polm:
## Description

This raises the maximum version limit for Typer, as requested in #11708, in order to run the tests. Note that even if the tests succeed there may be other things not check or other reasons to not raise the limit. In particular, we need to check if anything was deprecated or changed significantly in 0.5.0 or later. Looking at the [release notes](https://typer.tiangolo.com/release-notes/) this doesn't seem to be an issue.

### Types of change

dependency update

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ murmurhash>=0.28.0,<1.1.0
 wasabi>=0.9.1,<1.1.0
 srsly>=2.4.3,<3.0.0
 catalogue>=2.0.6,<2.1.0
-typer>=0.3.0,<0.7.0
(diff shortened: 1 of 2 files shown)


adrianeboyd:
From the release notes it doesn't look like there are any breaking changes and the extra CI CLI tests cover most of the CLI commands.

BLKSerene:
Hi, `Typer` 0.7.0 has been released, is it possible to further raise the verison limit?

polm:
Went ahead and bumped requirements. Based on the release notes for the latest one, it also looks like there's nothing to worry about here.

PR #11722, opened 2022-10-31
Remove sentiment extension

thomashacker:
## Description

This `PR` removes the `sentiment` extension attribute from the `Token`, `Span`, `Doc` and `Lexeme` class.

The `sentiment` extension is tied to one particular model of sentiment analysis that is kind of old and limited in applications. In general it makes sense to make custom field support better where there's not a performance impact rather than having more special-case fields built in.

### Types of change

Remove feature and change to documentation

## Checklist


polm (approved):
Looks good to me! I'm not sure if we should get other people to weigh in on this or not?

adrianeboyd:
Additional files aren't included in the package by default, so you can copy the bytestring directly into the test file instead. Try to make the example as short as you can, maybe just one token + `.sentiment` and no additional annotation. Alternatively, you could adjust the settings in `MANIFEST.in` to include this file, but I think this can be made short enough to include directly.

Separate from this PR, we probably need to think about how to extend the test suite with more examples like this without bloating the final `spacy` package too much. We were thinking about having two test suites, one that's packaged with `spacy.tests` and a separate one with larger/slower tests that run in the CI but not with `pytest --pyargs spacy`.

website/docs/usage/rule-based-matching.md
```diff
@@ -777,6 +777,8 @@ whitespace, making them easy to match as well.
 from spacy.lang.en import English
 from spacy.matcher import Matcher
 
+Doc.set_extension("sentiment", default=0.0)
```
adrianeboyd, line 780: Both the examples here are missing the `Doc` import.

spacy/tests/doc/test_doc_api.py
```diff
@@ -990,3 +990,11 @@ def test_doc_spans_setdefault(en_tokenizer):
     assert len(doc.spans["key2"]) == 1
     doc.spans.setdefault("key3", default=SpanGroup(doc, spans=[doc[0:1], doc[1:2]]))
     assert len(doc.spans["key3"]) == 2
+
+def test_doc_sentiment_from_bytes_v3_to_v4():
```
adrianeboyd, line 994: Can you reformat this file?

spacy/tests/README.md
```diff
@@ -10,12 +10,16 @@ Tests for spaCy modules and classes live in their own directories of the same na
 
 ## Table of contents
 
-1. [Running the tests](#running-the-tests)
-2. [Dos and don'ts](#dos-and-donts)
-3. [Parameters](#parameters)
-4. [Fixtures](#fixtures)
-5. [Helpers and utilities](#helpers-and-utilities)
-6. [Contributing to the tests](#contributing-to-the-tests)
+- [spaCy tests](#spacy-tests)
```
adrianeboyd, line 13: What happened to the formatting here? Was this intentional?

spacy/tests/doc/test_span.py
```diff
@@ -304,32 +304,6 @@ def test_span_similarity_match():
         assert span1.similarity(doc) == 0.0
         assert span1[:1].similarity(doc.vocab["a"]) == 1.0
 
-
```
adrianeboyd, line 307: Can you reformat this file, I think there should be one more blank line.

spacy/tests/README.md
thomashacker: Oh, I didn't realize, `prettier` formatted the table of contents automatically to this. I've replaced the old table and saved without formatting.

svlandeg (approved):
Looks good to me! I'll leave the final decision to merge or not with Adriane so that she can check that all her reviewer comments have been addressed.

adrianeboyd:
Looks good!

PR #11735, opened 2022-11-02
Add fallback in requirements check, only check once

adrianeboyd:
## Description

The requirements check should only warn and never raise an error, so add a fallback for any requirements that can't be handled otherwise. In addition, add a flag to `project_run` so that the requirements check is only run once for the top-level `run` command rather than also for each subcommand in workflows.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/cli/project/run.py
+++ b/spacy/cli/project/run.py
@@ -53,6 +53,7 @@ def project_run(
     force: bool = False,
     dry: bool = False,
     capture: bool = False,
+    skip_check_requirements: bool = False,
 ) -> None:
     """Run a named script defined in the project.yml. If the script is part
     of the default pipeline (defined in the "run" section), DVC is used to
@@ -69,6 +70,7 @@ def project_run(
         sys.exit will be called with the return code. You should use capture=False
         when you want to turn over execution to the command, and capture=True
         when you want to run the command more like a function.
+    skip_check_requirements (bool): Whether to skip the requirements check.
     """
     config = load_project_config(project_dir, overrides=overrides)
     commands = {cmd["name"]: cmd for cmd in config.get("commands", [])}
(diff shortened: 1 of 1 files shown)


adrianeboyd:
It seems like it would be nicer to override `check_requirements` in the config, but this isn't a supported override.

I'm also not attached to the `skip_check_requirements` naming. `skip_requirements_check` might be better?

spacy/cli/project/run.py
```diff
@@ -53,6 +53,7 @@ def project_run(
     force: bool = False,
     dry: bool = False,
     capture: bool = False,
+    skip_check_requirements: bool = False,
```
rmitsch, line 56: I favor `skip_requirements_check` over `skip_check_requirements`.

rmitsch (approved):

adrianeboyd:
You get:

```
check_requirements	not a section value that can be overwritten
```

It seems to look for `.` in any overrides.

spacy/cli/project/run.py
```diff
@@ -69,16 +70,18 @@ def project_run(
...
     validate_subcommand(list(commands.keys()), list(workflows.keys()), subcommand)
 
     req_path = project_dir / "requirements.txt"
-    if config.get("check_requirements", True) and os.path.exists(req_path):
-        with req_path.open() as requirements_file:
-            _check_requirements([req.replace("\n", "") for req in requirements_file])
+    if not skip_requirements_check:
+        if config.get("check_requirements", True) and os.path.exists(req_path):
+            with req_path.open() as requirements_file:
+                _check_requirements([req.replace("\n", "") for req in requirements_file])
```
polm, line 84:
```suggestion
                _check_requirements([req.strip() for req in requirements_file])
```

I had to check, but this is more like what `pip install -r` does. In general this is not important, but the current version won't work with DOS newlines, for example.

polm (commented):
There's one minor thing I think should be fixed but looks good.

polm:
Doesn't this require a docs update?

adrianeboyd:
I think that v3.4.2 should have had the relevant doc updates (#11226)? Unless you mean the `@` part?

polm:
Ah, yes, that should be fine - I forgot this was a followup to that PR.

thomashacker (approved):

PR #11739, opened 2022-11-02
Add `training.before_update` callback

shadeMe:
## Description

This callback can be used to implement training paradigms like gradual (un)freezing of components (e.g: the Transformer) after a certain number of training steps to mitigate catastrophic forgetting during fine-tuning.

### Types of change

New feature

## Checklist


spacy/schemas.py
```diff
@@ -329,6 +329,7 @@ class ConfigSchemaTraining(BaseModel):
     frozen_components: List[str] = Field(..., title="Pipeline components that shouldn't be updated during training")
     annotating_components: List[str] = Field(..., title="Pipeline components that should set annotations during training")
     before_to_disk: Optional[Callable[["Language"], "Language"]] = Field(..., title="Optional callback to modify nlp object after training, before it's saved to disk")
+    before_update: Optional[Callable[["Language", int], None]] = Field(..., title="Optional callback that is invoked at the start of each training step")
```
shadeMe, line 332: We might want to consider making this signature forward-compatible, i.e., use a `Dict[str, Any]` as the second parameter to allow for passing more values to the callback in the future.

adrianeboyd: It's extremely hard to change the config setting types in the future (other than widening), so it would make sense to go ahead and implement the more general version, I think?

spacy/training/loop.py
```diff
@@ -198,6 +201,9 @@ def train_while_improving(
     words_seen = 0
     start_time = timer()
     for step, (epoch, batch) in enumerate(train_data):
+        if before_update:
+            before_update_args = {"current_step": step}
```
danieldk, line 205: Maybe pass through the epoch as well?

danieldk: Also, why `current_stop` over simply `step`?

danieldk (commented):
Did you find out if there is an easy way to test this?

spacy/training/loop.py
shadeMe: Re. `current_step`: No particular reason; I think I copy-pasted that from elsewhere. Will keep it as `step` for consistency.

shadeMe:
> Did you find out if there is an easy way to test this?

Nothing short of bundling a small, toy corpus file and invoking the `train` and `init_nlp` functions like I do in the [gradual transformer unfreezing PR](https://github.com/danieldk/curated-transformers/pull/49/files#diff-ada5e46fc0199cafae7d5b28a18fbd3e9e9363eb4afa1ecb6f4b7f3390d08e6e).

danieldk (approved):

adrianeboyd:
Can you add some minimal tests for this? It doesn't look like there are any training-specific callback tests yet, but there are similar tests for the `[nlp]` callbacks.

adrianeboyd:
I think you should be able to test this with `train_while_improving` without a separate `.spacy` file? (I'm pretty sure the tests are going to fail because it's not automatically included in the package, and we'd want to be careful in general about adding a lot of extra data to the test suite.)

shadeMe:
Ah, didn't realize the extra data wouldn't be included anyway. I'll try out your suggestion.

adrianeboyd:
It's not that we can't add extra data with some additional settings, and it might make sense to have one toy `.spacy` file that's shared between different tests.

As I've mentioned a few times, the extra CLI tests in the CI should probably be moved into a proper test suite that's separate from the package tests and we can have the CI run both, and that would be an easier place for longer/larger tests.

spacy/tests/training/test_training.py
```diff
@@ -1112,3 +1115,39 @@ def test_retokenized_docs(doc):
...
+
+        # Raise an error here as the rest of the loop
+        # will not run to completion due to uninitialized
+        # models.
+        raise ValueError("ran_before_update")
+
+    def generate_batch():
+        yield 1, [Example(doc, doc)]
+
+    nlp = spacy.blank("en", config={"training": {}})
```
adrianeboyd, line 1133: Ah, I mean putting the `training.before_update` setting here? I guess it doesn't matter if you call `train_before_improving` directly?

shadeMe: Yeah, it shouldn't matter.

spacy/tests/conftest.py
```diff
@@ -1,3 +1,4 @@
+from pathlib import Path
```
adrianeboyd, line 1:
```suggestion
```

spacy/tests/training/test_training.py
adrianeboyd, line 1133:
```suggestion
    nlp = spacy.blank("en")
```

spacy/tests/training/test_training.py
```diff
@@ -1,7 +1,9 @@
 import random
+from confection import Config
```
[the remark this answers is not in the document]
shadeMe: Done. No other changes (none pertinent to this PR, anyway) required apart from this one.

PR #11744, opened 2022-11-03
Fix types for Span.id and Span.id_

adrianeboyd:
## Description

Make types for `Span.id` and `Span.id_` the same as for other properties in `Span` with getters and setters.

It seems like the property types are not interpreted as writable by pyright (see #11113). mypy doesn't seem to have an issue and I'm not sure what's intended/correct, but it seems sensible that they are all the same.

### Types of change

Types.

## Checklist


polm (approved):
Looks fine.

PR #11745, opened 2022-11-03
Revert disable/disabled merging behavior

rmitsch:
## Description

Resolves https://github.com/explosion/spaCy/issues/11706 and related issues. This modifies/reverts the merging behavior for `disable`/`disabled`, so that both sets of pipe names are merged instead of the latter being overwritten by the former (which is still the active behavior for `enable`/`enabled`). 

Adjusts tests, warnings and errors.

### Types of change

Bug fix? We revert to pre-v3.4.2 behavior.

## Checklist


spacy/language.py
```diff
@@ -2084,10 +2075,12 @@ def _resolve_component_status(
...
-            to_disable = [
-                pipe_name for pipe_name in pipe_names if pipe_name not in enable
-            ]
-            if disable and disable != to_disable:
+            to_disable = {
+                *[pipe_name for pipe_name in pipe_names if pipe_name not in enable],
+                *disable,
+            }
+            # If any pipe to be enabled is in to_disable, the specification is inconsistent.
+            if any([pipe_to_enable in to_disable for pipe_to_enable in enable]):
```
adrianeboyd, line 2083: I was surprised that this check didn't use a set operation?

rmitsch: No particular reason for this. Replaced with a set overlap check.

adrianeboyd: I was just surprised because I initially thought something more complicated must be going on, but it wasn't, right?

rmitsch: Before this PR we [checked for equivalence](https://github.com/explosion/spaCy/pull/11745/files#diff-fe971afb0a4f4d99848a30cfc2708c359e30703084ea5220629928ac9015f15dL2090) between the sets of pipes defined via `exclude` and those via `include` (and raised an error in case of inequality).

With our current assumptions, I think a conflict should occur if and only if the sets of pipes to be enabled (in `enable(d)`) and pipes to be disabled (pipes not in `enable(d)` + pipes in `disable(d)`). Do you agree?

adrianeboyd: The current check is what I expected.

adrianeboyd:
Is there a reason that this is still a draft?

rmitsch:
> Is there a reason that this is still a draft?

Yes, I haven't updated the documentation. I wanted to wait for the behavior to be agreed upon (which it is now IMO) before doing that.

website/docs/api/language.md
```diff
@@ -63,18 +63,18 @@ spaCy loads a model under the hood based on its
...
+| `vocab`                               | A `Vocab` object. If `True`, a vocab is created using the default language data settings. ~~Vocab~~                                                                                                                                                                                                                |
+| `disable`                             | Name(s) of pipeline component(s) to [disable](/usage/processing-pipelines#disabling). Disabled pipes will be loaded but they won't be run unless you explicitly enable them by calling [nlp.enable_pipe](/api/language#enable_pipe). Is merged with the config entry `nlp.disabled`. ~~Union[str, Iterable[str]]~~ |
+| `enable` <Tag variant="new">3.4</Tag> | Name(s) of pipeline component(s) to [enable](/usage/processing-pipelines#disabling). All other pipes will be disabled, but can be enabled again using [nlp.enable_pipe](/api/language#enable_pipe). Overwrites the config entry `nlp.enabled`. ~~Union[str, Iterable[str]]~~                                       |
```
adrianeboyd, line 72: There is no `nlp.enabled` config entry, there's just `nlp.pipeline` and `nlp.disabled`.

rmitsch: Ah, indeed. Why not?

adrianeboyd: It would be redundant with `nlp.pipeline - nlp.disabled`.

rmitsch: Right, but that also applies to the `enable`/`disable` situation? Anyway, not important - thanks for the catching this.

PR #11747, opened 2022-11-03
Remove references to local remotes in docs

adrianeboyd:
## Description

Remove references to local project URLs in docs since it does not seem to be currently supported.

### Types of change

Docs.

## Checklist


adrianeboyd:
I think either this should be modified or all the references to `smart_open` should also probably be modified to refer to buckets supported by `pathy` instead, since it seems all URLs are treated as `Pathy` URLs and other things mentioned like SSH aren't supported either.

PR #11762, opened 2022-11-07
Support local filesystem remotes for projects

adrianeboyd:
## Description

* Fix support for local filesystem remotes for projects
  * Use `FluidPath` instead of `Pathy` to support both filesystem and remote paths
  * Create missing parent directories if required for local filesystem
* Update `pathy` dependency to exclude older versions that aren't compatible with required `smart_open` version
* Update docs to refer to `Pathy` instead of `smart_open` for project remotes
* Add tests for local filesystem remotes

### Types of change

Bug.

## Checklist


adrianeboyd:
Reopening for testing because `pathy` has recently been updated.

spacy/cli/project/remote_storage.py
```diff
@@ -119,16 +118,19 @@ def find(
...
         elif command_hash is not None:
             urls = list((self.url / name / command_hash).iterdir())
         else:
-            urls = list((self.url / name).iterdir())
+            urls = []
+            for command_hash_dir in (self.url / name).iterdir():
+                urls.extend(command_hash_dir.iterdir())
             if content_hash is not None:
                 urls = [url for url in urls if url.parts[-1] == content_hash]
+        # TODO: URLs should be sorted by last modified
```
adrianeboyd, line 130: `pathy` doesn't return file stats in the same format for local paths and remote paths, so this would still need an additional helper method to sort the files by last modified.

justindujardin: Can you elaborate on what the problem is? I'd be happy to look at making it more consistent across local/remote paths.

adrianeboyd: It looks like you get `BlobStat` for remote paths and `os.stat_result` for local paths?

justindujardin: You're right that calling `.stat()` on a local file system path was returning `os.stat_result`. 

I released Pathy `0.10.0` today which returns `BlobStat` for local and remote paths.

adrianeboyd: Thanks, we'll try it out!

spacy/tests/test_cli.py
```diff
@@ -3,6 +3,8 @@
 from collections import Counter
 from typing import Tuple, List, Dict, Any
 import pkg_resources
+from typing import Counter
```
svlandeg, line 6:
```suggestion
from typing import Tuple, List, Dict, Any, Counter
import pkg_resources
```

svlandeg (approved):
Looks good!

Issue #11757, opened 2022-11-07
OverflowError: Invalid Nan value when encoding double


IllyShaieb:
## How to reproduce the behaviour

An extract of my python code:

```python
    ...

    def train(self) -> None:
        """
        Train the model on the given dataset for the given number of epochs.
        """
        if self.config.rebuild_data:
            self._prepare_data()

        if self.config.rebuild_config:
            os.system(
                "poetry run python -m spacy init fill-config .\\data\\intents\\base_config.cfg .\\data\\intents\\config.cfg"
            )

        os.system(
            "poetry run python -m spacy train .\\data\\intents\\config.cfg --output .\\models\\intents"
        )

    def _make_spacy_docs(
        self, data: list[tuple[str, str]], labels_to_idx: dict[str, int]
    ) -> list:
        """
        Helper function to create take a list of texts and labels and
        create a list of spaCy docs.

        returns: A list of spaCy docs.
        """
        docs = []
        for doc, label in tqdm(self.nlp.pipe(data, as_tuples=True), total=len(data)):
            doc.cats[label] = labels_to_idx[label]
            docs.append(doc)

        return docs

    def _prepare_data(self) -> None:
        """
        Helper function to prepare and save the data for training and validation.
        """
        dataset = data.IntentClassifierDataset(
            Path(self.config.data_path), shuffle=True
        )
        train_data, test_data = dataset.split(self.config.train_percentage)

        labels_to_idx = {label: idx for idx, label in enumerate(dataset.intents)}

        train_docs = self._make_spacy_docs(train_data.values.tolist(), labels_to_idx)
        test_docs = self._make_spacy_docs(test_data.values.tolist(), labels_to_idx)

        train_bin = DocBin(docs=train_docs)
        test_bin = DocBin(docs=test_docs)

        train_bin.to_disk(self.config.train_data_save_path)
        test_bin.to_disk(self.config.valid_data_save_path)

    ...
``` 

After creating the data and running the spaCy train loop I get an overflow error:

```bash
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████| 59/59 [00:00<00:00, 629.33it/s]
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████| 15/15 [00:00<00:00, 556.84it/s]
✔ Auto-filled config with all values
✔ Saved config
data\intents\config.cfg
You can now add your data and train your pipeline:
[...]

PR #11763, opened 2022-11-07
Check textcat values for validity

polm:
## Description

The textcat components expect all category values to be either 1 or 0 if present. When this expectation is not met the errors generated are surprising and don't make the issue clear, like in #11757. 

This change checks sample data during initialization and throws an error if invalid values are found. That's not perfect, since it might not catch things if only a few examples have bad values, but in the typical case it should prevent confusion by quickly identifying the issue.

### Types of change

data validation

## Checklist

### DIFF
--- a/spacy/pipeline/textcat.py
+++ b/spacy/pipeline/textcat.py
@@ -405,4 +405,4 @@ class TextCategorizer(TrainablePipe):
                 raise ValueError(Errors.E895.format(value=ex.reference.cats))
             for val in ex.reference.cats.values():
                 if not (val == 1.0 or val == 0.0):
-                    raise ValueError(Errors.E852.format(val=val))
+                    raise ValueError(Errors.E851.format(val=val))
--- a/spacy/pipeline/textcat.py
+++ b/spacy/pipeline/textcat.py
@@ -403,3 +403,6 @@ class TextCategorizer(TrainablePipe):
         for ex in examples:
             if list(ex.reference.cats.values()).count(1.0) > 1:
                 raise ValueError(Errors.E895.format(value=ex.reference.cats))
+            for val in ex.reference.cats.values():
+                if not (val == 1.0 or val == 0.0):
+                    raise ValueError(Errors.E852.format(val=val))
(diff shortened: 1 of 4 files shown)


spacy/pipeline/textcat.py
```diff
@@ -403,3 +403,6 @@ def _validate_categories(self, examples: Iterable[Example]):
         for ex in examples:
             if list(ex.reference.cats.values()).count(1.0) > 1:
                 raise ValueError(Errors.E895.format(value=ex.reference.cats))
+            for val in ex.reference.cats.values():
+                if not (val == 1.0 or val == 0.0):
+                    raise ValueError(Errors.E852.format(val=val))
```
adrianeboyd, line 408: The error number is incorrect, also below.

adrianeboyd:
This seems like a check that belongs in `debug data` rather than `initialize`. Is there another method that would handle this more cleanly during training, without being too slow? In `get_loss` maybe?

polm:
I wouldn't want to put this just in `debug data`, as then we'd still be stuck with unclear error messages at training time. It would make sense to also add it there, though. The idea is to check all data, right?

Looking at `get_loss`, I overlooked this, but the current code in `textcat` is in `_validate_categories`, which is called in both initialization and `get_loss`. I haven't timed it but since we're already checking all the values I don't think it's that slow? 

In `textcat_multilabel` I put the code directly in the initialize function, since `_validate_categories` isn't called there and this isn't category validation, but I could put it there so it works the same way. In this case the values are not all checked so it might make a difference, but if we want to check every item I'm not sure there's a better way. I can check with one of the sample projects.

adrianeboyd:
I guess if it's not in `debug data` it would make sense to add. I think it makes sense to add this check in place that is relevant for all training data and not just in `initialize`. We've also had cases where people had values like `0.75` where the training and/or eval didn't work either.

polm:
Turns out this is already in `debug data`:

https://github.com/explosion/spaCy/blob/3478ff1eb0fd57c48a332e7787efe6ea47492e13/spacy/cli/debug_data.py#L449-L456

I'll go ahead and modify this to check all data while training, rather than just running in initialize, but then I guess it'll be ready?

polm:
Tested this with `textcat_goemotions` CNN config, which I would expect to be more affected by speed changes, but there was no discernable change in the speed of training.

PR #11774, opened 2022-11-08
Add greCy to Universe

jmyerston:
## Description

Add greCy (ancient Greek models to spaCy universe)
### Types of change

Project description with links. 
## Checklist

- [X ] I confirm that I have the right to submit this contribution under the project's MIT license.

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -1,5 +1,31 @@
 {
     "resources": [
+         {
+            "id": "grecy",
+            "title": "greCy",
+            "slogan": "Ancient Greek pipelines for spaCy",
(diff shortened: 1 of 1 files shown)


polm:
Thanks for the PR! Since this is a draft, is there something else you wanted to add before it gets merged?

The current contents look OK, except that the Github value should be just of the format `jmyerston/grecy`, and not the whole Github URL - that's why the tests are failing.

jmyerston:
Thanks for the feedback! I fixed the GitHub value. I will extend the description in the future as the project develops. Next release should contain improved sentence segmentation, and after that I will add the NER pipeline.

jmyerston:
I think you can merge it if you agree.

polm:
Looks good, thanks!

adrianeboyd:
Just a note: usually universe updates like this are pushed to the website as soon as they're merged, but we're working on some website upgrades in the background so this will be delayed a bit.

PR #11777, opened 2022-11-09
Update warning, add tests for project requirements check

adrianeboyd:
## Description

* Make warning more general for differences between PEP 508 and pip
* Add tests for _check_requirements

### Types of change

?

## Checklist

### DIFF
--- a/spacy/tests/test_cli.py
+++ b/spacy/tests/test_cli.py
@@ -859,33 +859,40 @@ def test_span_length_freq_dist_output_must_be_correct():
     assert list(span_freqs.keys()) == [3, 1, 4, 5, 2]
 
 
-def test_project_check_requirements():
-    reqs = """
-    spacy
+@pytest.mark.parametrize(
+    "reqs,output",
+    [
+        [
+            """
+            spacy
 
-    # comment
+            # comment
 
-    thinc"""
-    output = _check_requirements([req.strip() for req in reqs.split("\n")])
-    assert output == (False, False)
-
-    reqs = """# comment
-    --some-flag
-    spacy"""
-    output = _check_requirements([req.strip() for req in reqs.split("\n")])
-    assert output == (False, False)
-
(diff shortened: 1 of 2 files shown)


spacy/tests/test_cli.py
```diff
@@ -855,3 +857,35 @@ def test_span_length_freq_dist_output_must_be_correct():
...
+    assert output == (False, False)
+
+    # excessive guard against unlikely package name
+    try:
+        pkg_resources.require("spacyunknowndoesnotexist12345")
+    except pkg_resources.DistributionNotFound:
+        reqs = """# comment
+        spacyunknowndoesnotexist12345"""
+        output = _check_requirements([req.strip() for req in reqs.split("\n")])
+        assert output == (True, False)
```
rmitsch, line 891: The point of the `try`/`except` is to verify first that the unlikely package is not installed, right? Alternative suggestion, feel free to ignore:
```suggestion
    pkg_name = "spacyunknowndoesnotexist12345"
    with pytest.raises(pkg_resources.DistributionNotFound):
        pkg_resources.require(pkg_name)
    reqs = f"# comment\n{pkg_name}"
    output = _check_requirements([req.strip() for req in reqs.split("\n")])
    assert output == (True, False)
```

rmitsch (approved):

spacy/tests/test_cli.py
adrianeboyd: I might be confused about what you're trying to propose, but the point of this was to completely skip the test in the (rare, unlikely) case that someone intentionally created an venv with this package in order to break the test suite.

rmitsch: Ah, I thought we wanted to guarantee that this package is not installed. Skipping makes more sense.

adrianeboyd:
The test is now skipped entirely if you have this unlikely package installed, but I think I can live with this.

PR #11779, opened 2022-11-09
Rename icelandic and multi-language lang codes

thomashacker:
## Description

This `PR` renames the Icelandic language code from `is` to `isl` and the multi-language code `xx` to `mul`

### Types of change

Renaming

## Checklist


adrianeboyd:
What happens when someone runs `spacy.blank("xx")` for the current state of this PR?

thomashacker:
Here's the output. 
The `util.find_matching_language()` function replaces `xx` with `mul`
```
import spacy
print(spacy.blank("xx"))

> mul
```

adrianeboyd:
I'm wondering whether we need to warn people when the internal name has changed because if they're checking `nlp.lang` or looking at their config it won't be the same as before even their code hasn't changed.

What happens for `spacy init config -l xx`?

thomashacker:
It throws an error when running the command. Should I implement it that way so that it uses the `util.find_matching_language()` functionality when creating the config? (Replacing `xx` with `mul`)

adrianeboyd:
I am kind of undecided myself. In general it would be nice for v3 configs to continue working without modifications in v4, but the `[nlp.lang]` setting is tricky and it's also going to be confusing if you specify `-l whatever` and it's not what you get in the end. Also it's a little unexpected if `spacy init config --lang whatever` doesn't have the same behavior as `spacy train --nlp.lang whatever`.

Since there's no good way to support multiple codes for the same language (as we discovered trying to support both `is` and `isl`), we may just want to live with this break in the config settings, since we have good reasons for modifying both `is` and `xx`.

I think if there are helpful warnings that specifically address these exact language codes, it's okay to keep `spacy.blank` supporting the wider range of codes with `langcodes` and the other cases require the exact spacy-supported code.

What do you think?

thomashacker:
I agree, I think it's a good trade-off 😄

thomashacker:
I removed the lang check from `cli.init_config()` and added more detail to the new warning so that every time `util.find_matching_language()` finds a match, that is not `None`, it returns:

`[W124] Language code 'is' was not found and replaced with 'isl'. Language codes like 'xx' and 'is' were renamed to 'mul' and 'isl' in v4.` (It should be run whenever `get_lang_class()` is used.)

I thought about adding more detail to the error that gets thrown from the `spacy init config` but I don't think I can add any more value to the existing one.

`[E958] Language code defined in config (xx) does not match language code of current Language subclass MultiLanguage (mul). If you want to create an nlp object from a config, make sure to use the matching subclass with the language-specific settings and data.`

adrianeboyd:
> I thought about adding more detail to the error that gets thrown from the `spacy init config` but I don't think I can add any more value to the existing one.
> 
> `[E958] Language code defined in config (xx) does not match language code of current Language subclass MultiLanguage (mul). If you want to create an nlp object from a config, make sure to use the matching subclass with the language-specific settings and data.`

I actually think this is really confusing. The user doesn't know that `xx` was converted to `mul` internally so they don't know why they're seeing this in the first place. I think there should be a special error case that is checked before `E958` just for `is` and `xx` that tells users exactly what they need to change.

adrianeboyd:
I don't think it's a good idea to ignore this warning in the tests. From looking at the CI logs it's showing up for a lot of cases where it didn't in v3, even for cases like `("en", "en")` where I think it definitely shouldn't. Can you take another look at this?

thomashacker:
The ignored warning is the newly added one, it gets thrown every time `util.find_matching_language()` finds a match. I found it a bit noisy since the test [`test_languages::test_blank_languages`](https://github.com/thomashacker/spaCy/blob/rename-islandic-and-multi-lang-code/spacy/tests/test_language.py#L589) uses the function ten times because the `registry` check fails. Weird, I tested it locally and it didn't throw a warning for `("en", "en")`, were they thrown in the azure checks before?

thomashacker:
Ah, okay I see now. The tests in the CLI fail, that's weird. Are `UserWarnings` regarded as errors? And it's interesting that `("en", "en")` shows up

adrianeboyd:
The CI runs `pytest -W error` so that unexpected warnings (in particular deprecation warnings) are treated like errors and we can catch problems sooner. This is a relatively recent change (within the last few months?) and some of the slow and GPU tests might still have warnings, so it's only the core non-slow test suite where we went through and make sure it ran without warnings, or they were intended and caught by pytest.

thomashacker:
Ah, I missed that there was also a test using `find_matching_language()` directly. This is why `("en", "en")` warnings showed up in the CLI.

thomashacker:
Should I let pytest catch these warning instead of ignoring them?

spacy/util.py
```diff
@@ -300,11 +302,9 @@ def find_matching_language(lang: str) -> Optional[str]:
...
     # shouldn't automatically match.
     match = langcodes.closest_supported_match(lang, possible_languages, max_distance=9)
-    if match == "mul":
-        # Convert 'mul' back to spaCy's 'xx'
-        return "xx"
-    else:
-        return match
+    if match is not None:
+        warnings.warn(Warnings.W124.format(lang=lang, renamed_lang=match))
+    return match
```
adrianeboyd, line 307: To make this utility function easier to use (as in the tests), I think it would be better if it doesn't add any warnings internally and then the function that calls it can handle any unexpected values with warnings or errors, based on how they want to use the returned value.

Wherever the warning is, it shouldn't be the case that it warns on `("en", "en")` because the code wasn't modified.

thomashacker: I've moved the warning to the `get_lang_class()` function which uses the `find_matching_language()`. It should only throw an error when `lang != match` is `True`, so that cases like `("en", "en")` don't throw an error.

adrianeboyd: I'm still concerned that this way more warnings that we had before. We added `langcodes` adjustments without warnings for sensible cases and either no one has noticed or no one has minded the behavior, so re-introducing warnings seems odd in a lot of cases.

Is it possible to provide this error message for `xx` and `is` for all the CLI `-l` options (there are 4-5 cases total?) and then not for `get_lang_class` and `spacy.blank` so that they aren't any different than v3 for things like `en-US`?

thomashacker: Yes, will implement is this way 😄

thomashacker: I've removed the warnings from `get_lang_class()` and `spacy.blank`. The warning only gets thrown when `spacy convert` is used with either `-l xx` or `-l is`. When using `spacy init config -l xx` (or `is`) an error gets thrown that gives more information about that `xx` and `is` were renamed to `mul` and `isl` and that the user should change it accordingly.

I haven't found any other CLI commands that have the `-l` parameter except for `spacy init config` and `spacy convert`.

`spacy convert` only throws a warning instead of an error because most of the converters don't use the `lang` parameter anyway and would still work.

spacy/cli/convert.py
```diff
@@ -110,6 +112,14 @@ def convert(
     input_path = Path(input_path)
     if not msg:
         msg = Printer(no_print=silent)
+
+    # Add warnings for renamed language code in v4
+    renamed_lang_codes = {"xx": "mul", "is": "isl"}
+    if lang in renamed_lang_codes:
+        warnings.warn(
```
adrianeboyd, line 119: In the CLI commands, use `msg.warn` instead of `warnings.warn` and `msg.fail` instead of raising errors.

I think that a warning isn't enough here (since you're not automatically remapping) and it should fail instead? I think we'd want `msg.fail(exits=1)` so that people switch to the updated codes and aren't surprised to have different codes in their codes than they specified. (I could potentially be convinced otherwise, but I think we don't want to deal with the confusion once you get to overrides like `--nlp.lang xx` and it's cleaner to do it right at the beginning.)

We usually use hard-coded messages in the CLI commands instead of `Errors`/`Warnings`, but repeating this in so many places also seems like a bad idea.

I think the language code map could go into `spacy/errors.py` just like `OLD_MODEL_SHORTCUTS`.

I think `lang` is used by: `convert`, `init config`, and `init vectors`

spacy/util.py
adrianeboyd: Hmm, I may be confused about exactly what happens here, and it's something that's not really covered by the tests. Let me try it out more directly...

adrianeboyd: The `--lang` setting is not actually used by any of the converters as far as I can tell.

adrianeboyd: The error for `init config` shows up as a relatively ugly traceback, and I think it might be nicer if it were caught sooner and reported with `msg.fail` instead.

thomashacker: Would it make sense to remove the warning from `spacy convert` then?

thomashacker: Should the error for `init config` be specifically for `xx` and `is` cases?

spacy/cli/convert.py
thomashacker: I created `RENAMED_LANGUAGE_CODES` in `errors.py` and replaced all errors/warnings with `msg.fail(exits=1)` for `convert`, `init config`, and `init vectors`. The tracebacks should be much more clearer now, we could maybe adjust the error text message a bit.

spacy/cli/convert.py
```diff
@@ -110,6 +111,15 @@ def convert(
...
     if not msg:
         msg = Printer(no_print=silent)
+
+    # Throw error for renamed language codes in v4
+    if lang in RENAMED_LANGUAGE_CODES:
+        msg.fail(
+            title="Renamed language code",
+            text=f"Language code '{lang}' was replaced with '{RENAMED_LANGUAGE_CODES[lang]}' in v4. Please change your current defined language code from '{lang}' to '{RENAMED_LANGUAGE_CODES[lang]}'.",
+            exits=1,
+        )
```
adrianeboyd, line 121: Since this is repeated, let's move it to a method in `spacy.cli._util`. It could be a private method like `_handle_renamed_language_codes` or something clunky.

We can remove `lang` from `convert` in a follow-up PR.

For the error message, maybe:

```
            text=f"Language code '{lang}' was replaced with '{RENAMED_LANGUAGE_CODES[lang]}' in spaCy v4. Update the language code from '{lang}' to '{RENAMED_LANGUAGE_CODES[lang]}'.",
```

If all of these are `--lang lg`, then it could make sense to provide the full option in the error message.

spacy/errors.py
```diff
@@ -965,6 +964,11 @@ class Errors(metaclass=ErrorsWithCodes):
     "lt": "lt_core_news_sm", "xx": "xx_ent_wiki_sm"
 }
 
+# Renamed language codes in v4
```
adrianeboyd, line 967: I think this needs to be reformatted.

spacy/tests/lang/test_initialize.py
```diff
@@ -7,10 +7,10 @@
 # excluded: ja, ko, th, vi, zh
 LANGUAGES = ["af", "am", "ar", "az", "bg", "bn", "ca", "cs", "da", "de", "el",
              "en", "es", "et", "eu", "fa", "fi", "fr", "ga", "gu", "he", "hi",
-             "hr", "hu", "hy", "id", "is", "it", "kn", "ky", "lb", "lt", "lv",
+             "hr", "hu", "hy", "id", "isl", "it", "kn", "ky", "lb", "lt", "lv",
              "mk", "ml", "mr", "nb", "ne", "nl", "pl", "pt", "ro", "ru", "sa",
              "si", "sk", "sl", "sq", "sr", "sv", "ta", "te", "ti", "tl", "tn",
-             "tr", "tt", "uk", "ur", "xx", "yo"]
+             "tr", "tt", "uk", "ur", "mul", "yo"]
```
adrianeboyd, line 13: A picky request, but this list should be alphabetized.

spacy/cli/convert.py
thomashacker: Created `_handle_renamed_language_codes` in `cli._util.py`, replaced all code blocks, and adjusted the error message.
I think, I don't fully understand what you mean by:
> If all of these are --lang lg, then it could make sense to provide the full option in the error message.

spacy/errors.py
```diff
@@ -965,6 +964,8 @@ class Errors(metaclass=ErrorsWithCodes):
...
+RENAMED_LANGUAGE_CODES = {"xx":"mul", "is":"isl"}
 
 # fmt: on
```
adrianeboyd, line 970: Oh, I didn't see that this was in the format off block. Now with formatting:

```suggestion
# fmt: on

# Renamed language codes in v4
RENAMED_LANGUAGE_CODES = {"xx": "mul", "is": "isl"}

```

thomashacker:
I accidentally merged with `master` instead of `v4` 🤯

thomashacker:
I'll create a new clean `PR` and close this

thomashacker:
New `PR` https://github.com/explosion/spaCy/pull/12149

PR #11780, opened 2022-11-09
Remove new v2.x tags

thomashacker:
## Description

This `PR` removes all `new` tags for v2.x

### Types of change

Change to documentation

## Checklist


polm (approved):
Looks good!

polm:
Note that while I have gone ahead and merged this, we're working on the site build pipeline, so it won't be live for a bit.

Also note that while this removes `<Tag>` buttons for v2, there's still references to features being new in v2 in headings, for example. But those appear to be stripped out by the build process so they don't show up on the site, so that's fine as-is.

PR #11781, opened 2022-11-10
Remove spacy-ray from docs

polm:
## Description

This removes docs related to spacy-ray since it's kind of out of date, as came up in https://github.com/explosion/spacy-ray/issues/15#issuecomment-1305613681. Specifically it removes:

- cli docs
- parallel training section
- projects integration
- Universe entry

I am not sure about the last one - maybe we want to keep it? There's also mention of ray in previous release docs, so maybe it would be better to put a warning on some of these than to remove them? 

### Types of change

docs cleanup

## Checklist


adrianeboyd:
I think it makes sense to remove it from the main docs. I'm less sure about the universe entry, although I think I would tend toward removing it because it really just doesn't work. If we do keep it, maybe we can add a clear warning that it's experimental and under development? (Would we accept a new universe entry that was in a similar state or that hadn't been updated for this long?)

polm:
> Would we accept a new universe entry that was in a similar state or that hadn't been updated for this long?

I think we wouldn't, though it hasn't come up. If a package wasn't actively maintained I don't think someone would be submitting it to the universe. 

Given it just doesn't work I guess it's better to remove it.

adrianeboyd:
I think we should probably just remove it everywhere.

PR #11793, opened 2022-11-11
change displacy default port to 5002

marioangst:
# Description
This changes the default port of displacy.serve from 5000 to 5002.

This is suggested because Apple now seems to occupy localhost:5000 for their AirPlay devices on macOs Monterey.
Related: https://stackoverflow.com/questions/70913242/access-to-localhost-was-denied-you-dont-have-authorisation-to-view-this-page-h 

Given that many users are likely to serve displacy to their localhost, it might be nice to change the default port to something like 5002. Otherwise, users on macOs Monterey will encounter a permission denied error (happened in my team).

## Types of change

Precaution to avoid bad user experience. Not essential.

## Checklist

- [x]  I confirm that I have the right to submit this contribution under the project's MIT license.
- [ ]  I ran the tests, and all new and existing tests passed.
- [x]  My changes don't require a change to the documentation, or if they do, I've added all required information.

### DIFF
--- a/spacy/displacy/__init__.py
+++ b/spacy/displacy/__init__.py
@@ -80,7 +80,7 @@ def serve(
     minify: bool = False,
     options: Dict[str, Any] = {},
     manual: bool = False,
-    port: int = 5000,
+    port: int = 5002,
     host: str = "0.0.0.0",
 ) -> None:
     """Serve displaCy visualisation.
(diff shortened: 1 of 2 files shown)


marioangst:
Hi, this sounds like a very sensible way to go forward. Also in terms of principle, it does not really make sense to try to play catch up with what manufacturers do with their default ports.
Should I open an issue, reference it and close this PR? Or have you already opened an issue (did not find one)?

svlandeg:
> Should I open an issue, reference it and close this PR?

Yes, that would be perfect 🙏

PR #11801, opened 2022-11-14
Update to mypy 0.990

adrianeboyd:
## Description

Update to mypy 0.990.

### Types of change

Types.

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,7 @@ pytest-timeout>=1.3.0,<2.0.0
 mock>=2.0.0,<3.0.0
 flake8>=3.8.0,<6.0.0
 hypothesis>=3.27.0,<7.0.0
-mypy>=0.980,<0.990; platform_machine != "aarch64" and python_version >= "3.7"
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -30,7 +30,7 @@ pytest-timeout>=1.3.0,<2.0.0
 mock>=2.0.0,<3.0.0
 flake8>=3.8.0,<6.0.0
 hypothesis>=3.27.0,<7.0.0
-mypy>=0.980,<0.990; platform_machine != "aarch64" and python_version >= "3.7"
+mypy>=0.990,<0.1000; platform_machine != "aarch64" and python_version >= "3.7"
```
svlandeg, line 33: We'll need to coordinate with Thinc, right? Or the two repos will be having conflicting requirements...

adrianeboyd: Since the requirements are optional and the required changes don't really interact, I don't think it will require much coordination.

svlandeg: Resolving as the Thinc [PR](https://github.com/explosion/thinc/pull/801) is open

PR #11803, opened 2022-11-14
Replace Pipe type with Callable in Language

adrianeboyd:
## Description

Replace `Pipe` type with `Callable[[Doc], Doc]` in `Language`.

Requires mypy 0.990 with support for `hasattr`.

### Types of change

Types.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -326,7 +326,7 @@ class Language:
         components are called on the Doc in order as it passes through the
         pipeline.
 
-        RETURNS (List[Tuple[str, Pipe]]): The pipeline.
+        RETURNS (List[Tuple[str, Callable[[Doc], Doc]]]): The pipeline.
         """
         pipes = [(n, p) for n, p in self._components if n not in self._disabled]
         return SimpleFrozenList(pipes, error=Errors.E926.format(attr="pipeline"))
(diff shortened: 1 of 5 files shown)


polm:
Technically a Pipe has a few more features than a Callable, like having a factory associated with it, right? I tried to see if that would cause any issues in Language, but I couldn't come up with any examples. 

Separately from that, docstrings still refer to `Pipe`. Do we want to keep them that way?

spacy/language.py
```diff
@@ -1363,7 +1362,7 @@ def resume_training(self, *, sgd: Optional[Optimizer] = None) -> Optimizer:
 
     def set_error_handler(
         self,
-        error_handler: Callable[[str, "Pipe", List[Doc], Exception], NoReturn],
+        error_handler: Callable[[str, PipeCallable, List[Doc], Exception], NoReturn],
```
polm, line 1365: The docstring here is an example of one where the function signature in the docstring and the actual one no longer match. Do we want to update this?

adrianeboyd:
I intentionally left the docstrings alone initially.

I don't think the docstrings should have `PipeCallable`. I think they could have `Callable` or `Callable[[Doc], Doc]`. Or we can just replace it all with `Callable[[Doc], Doc]` maybe and drop `PipeCallable`?

I don't think they should have `Pipe`, which isn't accurate.

adrianeboyd:
Using `Callable[[Doc], Doc]`, which also matches the API docs.

polm (approved):
Looks good.

PR #11806, opened 2022-11-15
Add equality definition for vectors

polm:
## Description

This uses the check for vector equality from sourcing components as a defintion on the Vectors class. At the moment it's only used in the one place, but I recently had need of it when working on the [merging script](https://github.com/explosion/projects/pull/147). There are probably other contexts where it could be useful too.

### Types of change

Minor enhancement.

## Checklist

### DIFF
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -243,6 +243,16 @@ cdef class Vectors:
         else:
             return key in self.key2row
 
+
+    def __eq__(self, other):
+        return (
+                self.shape == other.shape
+                and self.key2row == other.key2row
+                and self.to_bytes(exclude=["strings"])
+                  == other.to_bytes(exclude=["strings"])
+               )
+
+
     def resize(self, shape, inplace=False):
         """Resize the underlying vectors array. If inplace=True, the memory
         is reallocated. This may cause other references to the data to become
--- a/spacy/vectors.pyx
+++ b/spacy/vectors.pyx
@@ -243,7 +243,6 @@ cdef class Vectors:
         else:
             return key in self.key2row
 
-
     def __eq__(self, other):
         # Check for equality, with faster checks first
         return (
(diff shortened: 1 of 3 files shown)


adrianeboyd:
This precedence/ordering of checks was designed for `__ne__`, not for `__eq__`, where I think it doesn't really make as much sense.

polm:
Would it make sense to implement a separate `__ne__` and `__eq__` then? We may not need it in many places yet, but it does feel like this is logically a definition of equivalence, so we might as well put something in the implementation.

adrianeboyd:
Eh, this is probably getting too complicated for no particularly great reason. I guess the question is whether we want the `==` or `!=` case to be prioritized in terms of speed?

polm:
I think it makes sense to fail fast / prioritize non-equal speed. In the typical case of combining pipelines, that would raise an error and indicate edits are needed, while in the equal case you're done (so waiting doesn't matter). In either case I don't think it's run very often so it probably doesn't make that big a difference.

spacy/vectors.pyx
```diff
@@ -243,6 +243,17 @@ cdef class Vectors:
...
+    def __eq__(self, other):
+        # Check for equality, with faster checks first
+        return (
+                self.shape == other.shape
+                and self.key2row == other.key2row
+                and self.to_bytes(exclude=["strings"])
+                  == other.to_bytes(exclude=["strings"])
+               )
+
+
```
adrianeboyd, line 256:
```suggestion
    def __eq__(self, other):
        # Check for equality, with faster checks first
        return (
                self.shape == other.shape
                and self.key2row == other.key2row
                and self.to_bytes(exclude=["strings"])
                  == other.to_bytes(exclude=["strings"])
               )

```

adrianeboyd:
It's a little unexpected that `__eq__` is not the fast version for `==`, but I don't think it's a big deal, and I think maintaining separate/different `__eq__` and `__ne__` doesn't make sense here.

PR #11820, opened 2022-11-17
Improve score_cats for use with multiple textcat components

svlandeg:
## Description
<s>`nlp.evaluate` is having a hard time when multiple components in a pipeline are writing to the same type of attribute. While there will always be a ray of complex cases we might not be able to solve, this PR caters for one specifically that we can accomodate: 2 components (e.g. 2 textcat's) with a distinct set of labels each. It should be possible to get a score for each distinct label across the two components.</s>

Further, I've removed code that extended the given `labels` in `score_cats` with labels found in the `Example` object. I might be missing a use-case where this is required (in which case we need more unit tests) but it feels like we should be evaluating on the actual labels from the component. If we do extend the `labels` based on the gold data in the example, we end up confusing one textcat component with the labels for another, ultimately resulting in `KeyError` crashes.

### Types of change
bug fix

## Checklist


adrianeboyd:
I'm not sure this is a bug exactly. It's intentional that only the scores from the final component's scorer for a certain set of score keys are present in `scores` since there's no general-purpose way to merge scores dicts like this. I don't think it makes sense to merge only some of the entries. You end up with `cats_score` and `cats_macro_f` etc. that don't reflect the stored fine-grained scores.

If you want to separate the two sets of scores, you can write a custom scorer for the one of the textcat components that uses custom `attr` + `getter` to run a completely separate evaluation.

I'd have to look in more detail about whether it makes sense to change how labels beyond the current component's known labels are handled.

adrianeboyd:
The labels stuff is definitely buggy. I think the intention was:

```python
        labels = set(labels)
        if not labels:
            for eg in examples:
                labels.update(getter(eg.predicted, attr).keys())
                labels.update(getter(eg.reference, attr).keys())
        f_per_type = {label: PRFScore() for label in labels}
        auc_per_type = {label: ROCAUCScore() for label in labels}
```

adrianeboyd:
Let's see if I can update this cleanly from #11892...

svlandeg:
> I'm not sure this is a bug exactly. It's intentional that only the scores from the final component's scorer for a certain set of score keys are present in scores since there's no general-purpose way to merge scores dicts like this. I don't think it makes sense to merge only some of the entries. You end up with cats_score and cats_macro_f etc. that don't reflect the stored fine-grained scores.

> If you want to separate the two sets of scores, you can write a custom scorer for the one of the textcat components that uses custom attr + getter to run a completely separate evaluation.

I'm fine with taking this part out of the current PR as it allows us to focus on one problem at a time. However I do feel like the current default overwriting behaviour is pretty unexpected and we should look into this. It's not that weird to be wanting/expecting aggregated scores across similar components, especially if they have distinct label sets. Anyway I added a task to our internal backlog so we can come back to this later.

adrianeboyd:
It's awkward that all the annotations get thrown into `doc.cats` and you can't really evaluate it at all without referring to the pipeline components. The scorers were written with the idea that their goal was only to compare two docs on some attribute without reference to the pipeline configuration. (Which is why it added the labels from the predicted and reference doc, so if all your reference docs had some label that all your predicted docs were missing, or you predicted some label that wasn't ever in the reference docs, this should show in the evaluation.)

Do you think it would be a good idea to add the example from the tests with the custom scorer to the docs in some form?

adrianeboyd:
Thanks for the doing the merge! I think all the outstanding steps have been addressed at this point?

svlandeg:
Yep - should be good to merge if you agree.

PR #11837, opened 2022-11-21
Remove unused error object in add_pipe

polm:
## Description

In `Language.add_pipe`, an error is created if a factory that doesn't exist is specified, but the error isn't used for anything. This has never been an issue because `create_pipe` is called immediately after and raises the same kind of error early in its definition.

Because `create_pipe` seems to be doing the right thing I just removed the error check and object creation here, but we could alternately raise here instead. 

### Types of change

Cleanup?

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -784,14 +784,6 @@ class Language:
                 factory_name, source, name=name
             )
         else:
-            if not self.has_factory(factory_name):
-                err = Errors.E002.format(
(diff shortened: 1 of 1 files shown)


rmitsch (approved):

svlandeg (approved):
Good catch - this is indeed redundant code and the `create_pipe` function call already takes care of this.

PR #11845, opened 2022-11-22
Don't throw an error if using displacy on an unset span key

polm:
## Description

If you use displaCy to render spans but the span key is unset (as opposed to merely empty), it will throw a generic error. This change makes it so no error is thrown, but instead the (already in place) warning about no spans being found is shown. 

The case where the key is unset is a stronger sign that a mistake has been made than the key being empty, but it still feels off to throw an error here.

### Types of change

cleanup?

## Checklist

### DIFF
--- a/spacy/displacy/__init__.py
+++ b/spacy/displacy/__init__.py
@@ -233,7 +233,8 @@ def parse_spans(doc: Doc, options: Dict[str, Any] = {}) -> Dict[str, Any]:
     tokens = [token.text for token in doc]
 
     if not spans:
-        warnings.warn(Warnings.W117.format(spans_key=spans_key))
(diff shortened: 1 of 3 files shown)


rmitsch (approved):
Looks reasonable to me. Haven't worked with `displacy` yet, so up to you if you think a second set of eyes would be a good idea.

svlandeg (commented):
Agreed that the custom warning makes more sense. I suggest to extend `W117` though and provide the keys that are in fact available by including `list(doc.spans.keys())` in the warning message, rather than it only focusing on the use-case where there are no spans at all.

polm:
Good idea, I added a list of the available keys to the output.

svlandeg (approved):

PR #11860, opened 2022-11-23
Fix spancat for zero suggestions

adrianeboyd:
## Description

`spancat.predict` crashes on docs with zero suggestions.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/ml/extract_spans.py
+++ b/spacy/ml/extract_spans.py
@@ -32,8 +32,8 @@ def forward(
         Y = Ragged(X.dataXd[indices], spans.dataXd[:, 1] - spans.dataXd[:, 0])  # type: ignore[arg-type, index]
     else:
         Y = Ragged(
-            ops.xp.zeros(X.dataXd.shape, dtype=X.dataXd.dtype),
-            ops.xp.zeros((len(X.lengths),), dtype="i"),
+            ops.xp.zeros((0, 0), dtype=X.dataXd.dtype),
+            ops.xp.zeros((0,), dtype="i"),
         )
     x_shape = X.dataXd.shape
     x_lengths = X.lengths
--- a/spacy/ml/extract_spans.py
+++ b/spacy/ml/extract_spans.py
@@ -32,8 +32,8 @@ def forward(
         Y = Ragged(X.dataXd[indices], spans.dataXd[:, 1] - spans.dataXd[:, 0])  # type: ignore[arg-type, index]
     else:
         Y = Ragged(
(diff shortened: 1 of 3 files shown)


adrianeboyd:
To clarify: I only created this PR to show the failing test. I don't have concrete plans to work on this immediately. Anyone else is welcome to work it.

spacy/ml/extract_spans.py
```diff
@@ -32,8 +32,8 @@ def forward(
         Y = Ragged(X.dataXd[indices], spans.dataXd[:, 1] - spans.dataXd[:, 0])  # type: ignore[arg-type, index]
     else:
         Y = Ragged(
-            ops.xp.zeros(X.dataXd.shape, dtype=X.dataXd.dtype),
-            ops.xp.zeros((len(X.lengths),), dtype="i"),
+            ops.xp.zeros((0, 0), dtype=X.dataXd.dtype),
+            ops.xp.zeros((0,), dtype="i"),
```
adrianeboyd, line 36: This is now never reached in `spancat`. The previous placeholder was clearly incorrect. This shouldn't crash the model now, but it's also not really necessary. You could just require that `len(indices) > 0` instead.

adrianeboyd: Nevermind about this, this was wrong.

spacy/tests/pipeline/test_spancat.py
```diff
@@ -372,31 +372,56 @@ def test_overfitting_IO_overlapping():
...
-
-        return zero_suggester
+            spans = []
+            lengths = []
+            for doc in docs:
+                if len(doc) > 0 and len(doc) % 2 == 0:
+                    spans.append((0, 1))
+                    lengths.append(1)
+                else:
+                    lengths.append(0)
```
polm, line 389: The test here is passing, but I'm having trouble understanding why having the `spans` and `lengths` being different lengths works. Is it intentional that they end up different lengths? Should a Ragged where the input array and the lengths are different lengths be valid?

adrianeboyd: `lengths` is the number of spans per doc. Here, some docs have 1 span and some docs have 0 spans.

polm: Got it, that makes sense. I had some trouble following the difference between span lengths, number of spans per doc, and indices in used in extracting spans.

polm (approved):
Looks good.

PR #11879, opened 2022-11-27
Add example sentence for Chinese on website.

zrpxx:
# Description
Added example sentence for Chinese in website/meta/languages.json so that it can be displayed instead of "No text available yet" when the "Show text example" option is checked.

# Types of change
Docs

# Checklist

- [ ]  I ran the tests, and all new and existing tests passed.


polm:
Thanks, that's great to have!

PR #11885, opened 2022-11-29
Lock current node version for `/website`

essenmitsosse:
## Description
This adds information to make sure developers use the correct version of Node, and also enables a simpler workflow via NVM to switch to the correct version of Node. Additionally switching to a newer version of Node will be more clear to users and PR reviewers in the future.

### Types of change
- adds `.nvmrc` file to support switching to the intended Node versions via NVM. ([Docs for more details](https://github.com/nvm-sh/nvm#nvmrc)). This also makes sure Netlify automatically uses the desired Node version without the need to specify an env (`NODE_VERSION`) through their UI, which might get out of sync with the actual desired version.
- sets the ["engines" property in `package.json`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#engines), which will give the user a warning when using the wrong Node version
- Update `README.md` for the now supported NVM workflow

## Checklist


svlandeg (approved):
Looks good to me. The CI was failing due to a flaky unrelated test. I restarted it and assume the CI will go green soon, after which we can merge this.

PR #11888, opened 2022-11-29
Cleanup/remove backwards compat

polm:
## Description

This removes the backwards compatibility default `overwrite` settings added in #9050. All of these are false, except for `extend` of the Morphologizer, which is true. We might need to think about whether these are the values we want.

Note this specifically doesn't modify the code related to the legacy EntityLinker architecture at all - that should be removed, and will be done in another PR. 

Marked as draft because I still need to check the docs.

### Types of change

cleanup

## Checklist

### DIFF
--- a/spacy/pipeline/morphologizer.pyx
+++ b/spacy/pipeline/morphologizer.pyx
@@ -20,10 +20,6 @@ from ..scorer import Scorer
 from ..training import validate_examples, validate_get_examples
 from ..util import registry
 
-# See #9050
-BACKWARD_OVERWRITE = True
-BACKWARD_EXTEND = False
-
 default_model_config = """
 [model]
 @architectures = "spacy.Tagger.v2"
@@ -101,8 +97,8 @@ class Morphologizer(Tagger):
         model: Model,
         name: str = "morphologizer",
         *,
-        overwrite: bool = BACKWARD_OVERWRITE,
-        extend: bool = BACKWARD_EXTEND,
+        overwrite: bool = False,
+        extend: bool = True,
         scorer: Optional[Callable] = morphologizer_score,
         save_activations: bool = False,
     ):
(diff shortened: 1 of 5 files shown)


spacy/pipeline/morphologizer.pyx
```diff
@@ -101,8 +97,8 @@ class Morphologizer(Tagger):
         model: Model,
         name: str = "morphologizer",
         *,
-        overwrite: bool = BACKWARD_OVERWRITE,
-        extend: bool = BACKWARD_EXTEND,
+        overwrite: bool = False,
+        extend: bool = True,
```
adrianeboyd, line 101: I don't understand the motivation for changing this from the current default?

```suggestion
        extend: bool = False,
```

polm: I don't really have strong thoughts about it, but it seemed reasonable (like a mode I would be likely to use) because it only adds. It's fine to make it false as well.

polm:
The docs have been updated and this should be ready.

spacy/pipeline/morphologizer.pyx
adrianeboyd: Maybe there could be good reasons to have it be true?

adrianeboyd: I think it's okay to leave it as proposed.

adrianeboyd: Wait, this was changed, so it's not as proposed.

adrianeboyd: But I think I like `False` here so it doesn't change from v3.

polm: To be clear it is currently `False` and not what I argued for above. But I'm OK with either way - while I do think the argument above makes sense, I don't feel very strongly about it, and it's not like it's that hard to change.

adrianeboyd: Let's keep `False` for now then, sorry for the mixed-up comments above.

PR #11889, opened 2022-11-29
Remove legacy Entity Linker component

polm:
## Description

The Entity Linker was unusual in that the component (rather than the model) had to be modified due to issues, creating a legacy version. This PR removes the legacy version and one test that covered it.

### Types of change

Cleanup.

## Checklist


svlandeg (commented):
So that would mean that people won't be able to use the v1 anymore as soon as they update to spaCy v4, right? 

How much effort is it to update a config file with v1 to v2?

polm:
It's correct this means that v1 won't work with spaCy v4 any longer. 

Updating the config shouldn't require any changes besides changing the version - only one option was added, it has a reasonable default, and none of the other options changed. Changes from v1 to v2 were primarily motivated by a bugfix related to listeners, and though it ended up changing the internals a lot, it shouldn't affect the accuracy or anything. However, if I remember correctly, because the structure of the Thinc models changed, it does require retraining.

For reference, v2 was added and v1 was made legacy in #9669.

polm:
Here's a thread where I confirmed the need to retrain. https://github.com/explosion/spaCy/discussions/10887#discussioncomment-2885821

polm:
I reviewed this again and it should be good to go.

azure-pipelines.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       matrix:
         # We're only running one platform per Python version to speed up builds
         Python36Linux:
-          imageName: "ubuntu-latest"
+          imageName: "ubuntu-20.04"
```
rmitsch, line 44: Out of curiosity, why this change?

rmitsch (approved):
I think this it's alright to remove this for v4 (and that retraining is an acceptable compromise). Approving, but let's double-check with @svlandeg before merging.

azure-pipelines.yml
polm: See #11928 - `ubuntu-latest` changed a little earlier than we expected, and it drops 3.6.

svlandeg, line 44: I just merged the updates from `master` to `v4` which contain this commit: https://github.com/explosion/spaCy/commit/8267aa1b65783ea29e42f470eaf7553ff7508206
So it's probably best if you merge in the latest `v4` again so we don't end up with merge conflicts :-)

svlandeg (changes requested):
As discussed internally, can we first explore the option to move this code to `spacy-legacy` instead?

polm:
I'll mark this as draft while we figure out how to handle this.

polm:
Since we'll continue to support this in `spacy-legacy`, let me make a new PR rather than reverting the only change here.

PR #11892, opened 2022-11-29
Restrict cats_score to provided labels

adrianeboyd:
## Description

Only score the provided labels with `score_cats` because any other combination of labels from the reference or predicted doc doesn't make sense when you have multiple `textcat` or `textcat_multilabel` components. Always use the provided `getter` rather than accessing `.cats`.

Related to #11820.

### Types of change

Bug fix.

## Checklist


spacy/scorer.py
```diff
@@ -473,17 +473,17 @@ def score_cats(
             threshold = 0.5 if multi_label else 0.0
         if not multi_label:
             threshold = 0.0
+        labels = set(labels)
         f_per_type = {label: PRFScore() for label in labels}
         auc_per_type = {label: ROCAUCScore() for label in labels}
-        labels = set(labels)
-        if labels:
```
adrianeboyd, line 479: I think this was intended to be `if not labels` so it could infer the labels from the examples. I think the idea was that your reference docs might have some label that your component has completely missed and you want to include that in the evaluation, but that idea doesn't scale beyond one textcat component.

adrianeboyd:
The custom scorer with the custom getter is extremely clunky, but the only workaround with the current `cats_score` implementation.

adrianeboyd:
I think if we do this then we'd want to do https://github.com/explosion/spacy-legacy/pull/28 with moving copying scorers to `spacy-legacy`.

spacy/scorer.py
```diff
@@ -473,17 +473,17 @@ def score_cats(
...
         for example in examples:
             # Through this loop, None in the gold_cats indicates missing label.
-            pred_cats = getter(example.predicted, attr)
-            gold_cats = getter(example.reference, attr)
+            pred_cats = {
+                k: v for k, v in getter(example.predicted, attr).items() if k in labels
+            }
+            gold_cats = {
+                k: v for k, v in getter(example.reference, attr).items() if k in labels
+            }
```
svlandeg, line 486: As these are all the same edits as in https://github.com/explosion/spaCy/pull/11820/files, I would have preferred to continue working from that PR rather than open a parallel one. We can remove the merging of the dict and keep that in a separate PR / for a separate discussion / more thorough solution, but this part is pretty much the same.

adrianeboyd: Sure, I can move these changes over there.

PR #11899, opened 2022-11-30
Config generation fails for GPU without transformers

polm:
## Description

If you don't have spacy-transformers installed, but try to use `init config` with the GPU flag, you'll get an error. The issue is that the `use_transformers` flag in the config is conflated with the GPU flag, and then there's an attempt to access transformers config info that may not exist.

There may be a better way to do this, but this stops the error.

### Types of change

cli/docs fix

## Checklist


adrianeboyd:
Looks good!

PR #11900, opened 2022-11-30
Change GPU efficient textcat to use CNN, not BOW in generated configs

polm:
## Description

If you generate a config with a textcat component using GPU (transformers), the defaut option (efficiency) uses a BOW architecture, which does not use tok2vec features. While that can make sense as part of a larger pipeline, in the case of just a transformer and a textcat, that means the transformer is doing a lot of work for no purpose.

This changes it so that the CNN architecture is used instead. It could also be changed to be the same as the accuracy config, which uses the ensemble architecture.

I haven't tested this in the browser yet because I'm having difficulty with the site's build env, but it works fine in the CLI, and imagine it's fine - there are no actual logic changes. 

## Checklist


adrianeboyd:
This doesn't add the `transformer` to the pipeline for the GPU case (otherwise there are conditions that remove the tok2vec parts for the BOW case):

```
[nlp]
lang = "en"
pipeline = ["textcat"]
batch_size = 128

[components]

[components.transformer]
factory = "transformer"

[components.transformer.model]
@architectures = "spacy-transformers.TransformerModel.v3"
name = "roberta-base"
tokenizer_config = {"use_fast": true}
```

polm:
Thanks for pointing that out, I fixed it. 

Going to mark this as a draft while the issue with #11925 is investigated, since it looks like affects the CNN textcat with transformers.

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -24,8 +24,10 @@ gpu_allocator = null
 lang = "{{ lang }}"
 {%- set has_textcat = ("textcat" in components or "textcat_multilabel" in components) -%}
 {%- set with_accuracy = optimize == "accuracy" -%}
-{%- set has_accurate_textcat = has_textcat and with_accuracy -%}
-{%- if ("tagger" in components or "morphologizer" in components or "parser" in components or "ner" in components or "spancat" in components or "trainable_lemmatizer" in components or "entity_linker" in components or has_accurate_textcat) -%}
+{# The BOW textcat doesn't need a source of features, so it can omit the
+tok2vec/transformer. #}
+{%- set textcat_needs_features = has_textcat and (with_accuracy or use_transformer) -%}
```
adrianeboyd, line 29: This needs to be checked extremely carefully in the website version because my experience is that there are bugs in how longer expressions like this are converted to javascript. (cf. #9610)

polm: I checked this after merging in the rebuilt site and it works fine as far as I can tell. (This PR is still waiting on the CNN bugfix though.)

polm:
Brought this out of draft since https://github.com/explosion/spaCy/issues/11968 is resolved, though this probably shouldn't be merged until it's released.

adrianeboyd:
thinc v8.1.8 hasn't been released yet, so this should fail for now.

adrianeboyd:
This test looks good: https://github.com/explosion/spacy-transformers/pull/357

So this change can be merged for v3.5.1.

adrianeboyd:
Putting this back in draft because it appears to break the training quickstart in the deploy preview.

adrianeboyd:
Locally it's not completely broken, but it doesn't appear to do what it's intended to do. This may take a big longer to debug because of the buggy/frustrating jinja2js conversion.

adrianeboyd:
I think the older website preview in the PR had gotten too old or out-of-sync with more recent website changes, and regenerated with the minor change above this now looks like it's working as intended to me.

Locally the issue was that `npm run dev` doesn't automatically regenerate the js version of the config template like it used to, so my local preview was also out-of-sync.

PR #11902, opened 2022-11-30
Add a `spacy benchmark speed` subcommand

danieldk:
## Description

This subcommand reports the mean batch performance of a model on a data set with a 95% confidence interval. For reliability, it first performs some warmup rounds. Then it will measure performance on batches with randomly shuffled documents.

To avoid having too many spaCy commands, `speed` is a subcommand of `benchmark` and `benchmark accuracy` is now an alias of `evaluate`.

Out of draft:

- [x] We need to decide how to handle that `accuracy` is now a subcommand.
- [x] I need to check/update the documentation and document the new subcommand.
- [x] ~Should probably target spaCy 4, since the module `evaluate` was renamed to `evaluate_accuracy`.~

### Types of change

New feature.

## Checklist


adrianeboyd:
Is there no way to keep the current `spacy evaluate` command?

danieldk:
> Hmm, `spacy evaluate` already includes a minimal `SPEED`. Could the extended version somehow be a flag instead? I realize it's tricky with the additional options.

It really works differently with the eval loop and stuff, it seems a bit bolted-on?

> But I dislike breaking all existing uses of `spacy evaluate` if we don't have to. I might propose:
> 
> * `spacy evaluate` adds a flag for this that uses the defaults
> * a separate command lets you run it with the extended options

So rather than adding a flag, what about:

* Keep `evaluate` as-is.
* Put the speed test under `benchmark speed`.
* Make `benchmark accuracy` an alias for `evaluate`.

That way we can move to subcommands without breaking the current interface? Not sure if `benchmark` is the best name, but you get the drift (maybe `analyze`?).

shadeMe:
> That way we can move to subcommands without breaking the current interface? Not sure if benchmark is the best name (maybe analyze?).

IMO, the `speed` subcommand for `benchmark` seems fairly superfluous (in this context, at the least). Can't we simply stick to an additional `benchmark` command and call it a day? That way, the default associations of the words `evaluate` -> `accuracy` and `benchmark` -> `speed` will still hold.

adrianeboyd:
`evaluate` already includes the simple `SPEED`, and "benchmark" doesn't have an "only speed" association for me?

danieldk:
> IMO, the speed subcommand for benchmark seems fairly superfluous (in this context, at the least). Can't we simply stick to an additional benchmark command and call it a day?

We discussed this a bit the other day, and one of the ideas was to have a suite of evaluation-related subcommands like we have `debug` now. `benchmark {accuracy,speed}` could be the first subcommands. The alias would be to preserve backwards compatibility, so that `spacy evaluate` still works. I think the downside of adding a top-level command (regardless of the name) is that we clutter the interface too much.

danieldk:
> Daniël's suggestion sounds okay to me. I'm not sure that it's easy to have a 100% alias, but I'm sure there's some way to get close.

Changed in this way, also made `benchmark accuracy` an alias for `evaluate`. I think this can now target `master`, since there are no changes to existing interfaces anymore. I still need to update the docs.

danieldk:
Added docs and moved this out of draft.

website/docs/api/cli.md
```diff
@@ -1162,6 +1163,38 @@ $ python -m spacy evaluate [model] [data_path] [--output] [--code] [--gold-prepr
...
+| `model`              | Pipeline to benchmark the speed of. Can be a package or a path to a data directory. ~~str (positional)~~ |
+| `data_path`          | Location of benchmark data in spaCy's [binary format](/api/data-formats#training). ~~Path (positional)~~ |
+| `--batch-size`, `-b` | Set the batch size. If not set, the pipeline's batch size is used. ~~Optional[int] \(option)~~           |
+| `--no-shuffle`       | Do not shuffle documents in the benchmark data. ~~bool (flag)~~                                          |
+| `--gpu-id`, `-g`     | GPU to use, if any. Defaults to `-1` for CPU. ~~int (option)~~                                           |
+| `--batches`          | Number of batches to benchmark on. Defaults to `50`. ~~Optional[int] \(option)~~                         |
+| `--warmup`, `-w`     | Iterations over the benchmark data for warmup. Defaults to `3` ~~Optional[int] \(option)~~               |
+| `--help`, `-h`       | Show help message and available arguments. ~~bool (flag)~~                                               |
+| **CREATES**          | Pipeline speed with 95% confidence interval.                                                             |
```
shadeMe, line 1197: `**REPORTS** ` might be a better verb here, no?

danieldk: Good catch. Changed to `**PRINTS**`, since that verb is used for some other CLI commands.

svlandeg (commented):
We should add some tests that ensure `spacy evaluate` and `spacy benchmark accuracy` both work as expected.

Cf https://github.com/explosion/spaCy/blob/master/spacy/tests/test_cli_app.py -> We can use `CliRunner` for this :-)

spacy/cli/evaluate.py
```diff
@@ -7,12 +7,14 @@
...
-from ._util import app, Arg, Opt, setup_gpu, import_code
+from ._util import app, Arg, Opt, setup_gpu, import_code, benchmark_cli
 from ..scorer import Scorer
 from .. import util
 from .. import displacy
 
-
+@benchmark_cli.command(
+    "accuracy",
+)
```
svlandeg, line 17: Oh, that's nice and clean :-)

website/docs/api/cli.md
```diff
@@ -1163,6 +1164,38 @@ $ python -m spacy evaluate [model] [data_path] [--output] [--code] [--gold-prepr
...
 | **CREATES**                               | Training results and optional metrics and visualizations.                                                                                                                            |
 
+## benchmark {#benchmark new="3.5"}
+
+The `spacy benchmark` CLI includes commands for benchmarking the accuracy and
+speed of your spaCy pipelines.
+
+### accuracy {#benchmark-accuracy new="3.5" tag="command"}
+
+This subcommand is an alias for [`spacy evaluate`](#evaluate).
```
svlandeg, line 1174: Maybe we should also have a note over at the `evaluate` documentation? Realistically we won't be deprecating `evaluate` anytime soon, but it's always good to warn long up-front.

website/docs/api/cli.md
```diff
@@ -1163,6 +1164,38 @@ $ python -m spacy evaluate [model] [data_path] [--output] [--code] [--gold-prepr
...
+| `model`              | Pipeline to benchmark the speed of. Can be a package or a path to a data directory. ~~str (positional)~~ |
+| `data_path`          | Location of benchmark data in spaCy's [binary format](/api/data-formats#training). ~~Path (positional)~~ |
+| `--batch-size`, `-b` | Set the batch size. If not set, the pipeline's batch size is used. ~~Optional[int] \(option)~~           |
+| `--no-shuffle`       | Do not shuffle documents in the benchmark data. ~~bool (flag)~~                                          |
+| `--gpu-id`, `-g`     | GPU to use, if any. Defaults to `-1` for CPU. ~~int (option)~~                                           |
+| `--batches`          | Number of batches to benchmark on. Defaults to `50`. ~~Optional[int] \(option)~~                         |
+| `--warmup`, `-w`     | Iterations over the benchmark data for warmup. Defaults to `3` ~~Optional[int] \(option)~~               |
+| `--help`, `-h`       | Show help message and available arguments. ~~bool (flag)~~                                               |
+| **PRINTS**           | Pipeline speed with 95% confidence interval.                                                             |
```
svlandeg, line 1197:
```suggestion
| **PRINTS**           | Pipeline speed in words per second (WPS) with 95% confidence interval.                                                             |
```
(will require `prettier`, too)

spacy/cli/benchmark_speed.py
```diff
@@ -0,0 +1,179 @@
...
+def print_mean_with_ci(sample: numpy.ndarray):
+    mean = numpy.mean(sample)
+    bootstrap_means = bootstrap(sample)
+    bootstrap_means.sort()
+
+    # 95% confidence interval
+    low = bootstrap_means[int(len(bootstrap_means) * 0.025)]
+    high = bootstrap_means[int(len(bootstrap_means) * 0.975)]
+
+    print(f"Mean: {mean:.1f} WPS (95% CI: {low-mean:.1f} +{high-mean:.1f})")
```
svlandeg, line 156: I wouldn't assume all users know what "WPS" stands for

svlandeg (commented):
I really like this feature, and I think having `evaluate` as an alias for `benchmark accuracy` makes sense and allows us to extend `benchmark` more in the future.

Had a few more comments/requests though!

spacy/cli/benchmark_speed.py
```diff
@@ -0,0 +1,179 @@
...
+from ..training import Corpus
+from ._util import Arg, Opt, benchmark_cli, setup_gpu
+
+
+@benchmark_cli.command(
+    "speed",
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True},
+)
+def benchmark_speed_cli(
+    ctx: typer.Context,
```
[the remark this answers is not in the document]
danieldk: Fixed.

spacy/cli/benchmark_speed.py
```diff
@@ -0,0 +1,179 @@
...
+def benchmark(
+    nlp: Language,
+    docs: List[Doc],
+    n_batches: int,
+    batch_size: int,
+    shuffle: bool,
+) -> numpy.ndarray:
+    if shuffle:
+        bench_docs = [
+            nlp.make_doc(random.choice(docs).text)
```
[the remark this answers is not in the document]
danieldk: An error is now emitted if the corpus is empty.

website/docs/api/cli.md
danieldk: It's a bit weird anyway to have all the documentation under `evaluate` if `benchmark speed` is supposed to be the replacement. So I moved the docs to `benchmark speed` and changed the docs for `evaluate` to state that it is superseded and that it is an alias for compatibility.

website/docs/api/cli.md
danieldk: Fixed, but I removed WPS completely (see below).

spacy/cli/benchmark_speed.py
danieldk: Changed to `words/s`.

danieldk:
> We should add some tests that ensure `spacy evaluate` and `spacy benchmark` both work as expected.
> 
> Cf https://github.com/explosion/spaCy/blob/master/spacy/tests/test_cli_app.py -> We can use `CliRunner` for this :-)

Added a test that compares the output of `evaluate` and `benchmark accuracy` to check that the alias works correctly.

PR #11903, opened 2022-11-30
Update custom solutions links

koaning:
Making changes to the docs so that we point to the `/custom-solutions` page.


koaning:
The links on the readme and landing page still need to updated but those require assets I currently don't have access to as well as a discussion around design.

svlandeg (commented):
Thanks for getting these corrected!

> The links on the readme and landing page still need to updated but those require assets I currently don't have access to as well as a discussion around design.

Let's have that internal discussion first, so we can add those edits to this PR 🙏

koaning:
I will now be adding this image to the README. 

spaCy tailored analysis logo

I'm adding it as a message here now, because it seems like we're using `user-images.githubusercontent.com` for the other image as well?

ines (approved):

PR #11910, opened 2022-12-01
Add failing test for span-sentence mapping

rmitsch:
## Description

There seems to be an issue with the span-to-sentence mapping when multiple spaces occur at the beginning of a sentence. This PR adds a minimal failing test (marked as `xfail`).

## Checklist

### DIFF
--- a/spacy/tests/doc/test_span.py
+++ b/spacy/tests/doc/test_span.py
@@ -709,7 +709,7 @@ def test_span_group_copy(doc):
     assert len(doc_copy.spans["test"]) == 2
 
 
-# @pytest.mark.xfail
+@pytest.mark.xfail
 @pytest.mark.parametrize("use_double_space", [True, False])
 def test_span_sentence_mapping(en_tokenizer, use_double_space: bool):
     """Tests correct mapping of spans to sentences. This is currently failing due to some issue with the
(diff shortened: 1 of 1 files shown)


spacy/tests/doc/test_span.py
```diff
@@ -680,3 +681,41 @@ def test_span_group_copy(doc):
...
+
+
+@pytest.mark.xfail
+@pytest.mark.parametrize("use_double_space", [False, True])
+def test_span_sentence_mapping(use_double_space: bool):
+    """Tests correct mapping of spans to sentences. This is currently failing due to some issue with the
+    span-to-sentence mapping.
+    use_double_space (bool): Whether to use double space after end of first sentence.
+    """
+    nlp = load("en_core_web_sm")
```
adrianeboyd, line 693: The unit tests don't have access to any trained pipelines. You can export the output of `en_core_web_sm` and recreate the relevant docs by hand with `Doc` for the tests?

rmitsch: Ah, right. Fixed in https://github.com/explosion/spaCy/pull/11910/commits/5349c1811c036b8740122ac96ac434fb83f667b4.

adrianeboyd:
I wouldn't be extremely surprised if there are some bugs here, but I'm confused by the current example. The second sentence starts on the second `This`, so it's expected that you get two sentences from the span for `space + "This"`.

Space tokens don't have any special status in terms of sentence boundaries. You could also mark the inserted space token as the sentence boundary if you wanted. `Span.text` does strip the single trailing space from the result that it returns if the last token has one, but that space isn't a separate token in the text.

rmitsch:
Closed after discussion: this is expected behavior.

Issue #11235, opened 2022-07-27
Spacy Projects: Directories are created before vars interpolation


polkaYK:
This may not be a bug per se, but I was a bit surprised by the behavior.

I discovered that directories are created before variable interpolation in a `spacy project`, so you can't have something like 
```
directories: [
  "cfg",
  "${vars.lang}_model",
]
```
I searched the code and found this place:
https://github.com/explosion/spaCy/blob/8ffff18ac4e6a1d4fdae76dd7a9ecdf251b149fa/spacy/cli/_util.py#L161-L170
I swapped these two code blocks and it still worked, at least in my small project. But maybe there are concerns that prevent you from doing this.
So, are you just not supposed to create directories with variable interpolation because of some reason or the order of these code blocks can be swapped?

adrianeboyd:
Thanks for the report, I think we just didn't consider interpolation in the directory names. I don't think there's any particular reason these blocks couldn't be swapped.

AhmedIssa11:
I would like to take care of this issue

adrianeboyd:
@AhmedIssa11 Sure, we'd be happy to review a PR for this!

PR #11914, opened 2022-12-02
Fix the interpolation in directory names when loading config files using cli.

zrpxx:
## Description
Solving #11235.
When loading config files, do the interpolation before checking the directories to handle variables in the directory names correctly. Related tests are added to test_cli.py.

Testing result:
- OS: `Windows 11`
- Python: `3.10`
- Result: 
```bash
$ py.test spacy
3298 passed, 1211 skipped, 32 xfailed, 1 xpassed in 354.31s (0:05:54)
```

### Types of change
Bugfix.

## Checklist

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -158,15 +158,15 @@ def load_project_config(
         sys.exit(1)
     validate_project_version(config)
     validate_project_commands(config)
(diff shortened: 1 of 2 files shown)


polm (approved):
Thanks for the PR - looks good to me!

PR #11928, opened 2022-12-05
Switch ubuntu-latest to ubuntu-20.04 in main tests

polm:
## Description

Azure seems to unexpectedly switched `ubuntu-latest` to be `ubuntu-22.04`, which drops Python 3.6. We're not ready for that since we're still testing 3.6, so this changes the ubuntu spec to the older version.

Not sure if we also want to do this for slow tests.

## Checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,7 +25,7 @@ jobs:
   # defined in .flake8 and overwrites the selected codes.
   - job: "Validate"
     pool:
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Only change 3.6 to `ubuntu-20.04` and leave the others for now. (I used to think that specifying a particular release made it more stable, but it doesn't because they change the software for all the images regularly.)

PR #11930, opened 2022-12-05
update lock-threads to v4

ryndaniels:
The `lock-threads` action has been updated upstream to fix the node deprecation warnings, so now we can bump it here.


svlandeg (approved):
Thanks!

PR #11934, opened 2022-12-05
EditTreeLemmatizer: correctly add strings when initializing from labels

danieldk:
## Description

Strings in replacement nodes where not added to the `StringStore` when `EditTreeLemmatizer` was initialized from a set of labels. The corresponding test did not capture this because it added the strings through the examples that were passed to the initialization.

This change fixes both this bug in the initialization as the 'shadowing' of the bug in the test.

Bug found while implementing distillation, since it relies purely on this functionality to initialize the teacher.

### Types of change

Bug fix.

## Checklist


polm (approved):
Looks good to me.

PR #11935, opened 2022-12-06
Add in errors used in the beam code that were removed at some point

polm:
## Description

There are a few errors in the beam search code that were removed (it looks like in #5664). I don't think there's any way to use the beam code at the moment, but as long as it's around the errors it refers to should also be present. 

### Types of change

minor fix?

## Checklist


svlandeg:
Good catch!

> There are a few errors in the beam search code that were removed (it looks like in https://github.com/explosion/spaCy/pull/5664).

https://github.com/explosion/spaCy/pull/5664 was fine, because at the time (if you look at the history) the beam parsing was in fact removed, it was only added back in with https://github.com/explosion/spaCy/pull/6369, where we didn't catch that the error messages were not pointing to anything anymore...

svlandeg (approved):

PR #11943, opened 2022-12-07
Remove all references to "begin_training"

polm:
## Description

When v3 was released, `begin_training` was renamed to `initialize`. There were warnings in the code and docs about that. This PR removes them.

### Types of change

cleanup

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -131,13 +131,6 @@ class Warnings(metaclass=ErrorsWithCodes):
             "and make it independent. For example, `replace_listeners = "
             "[\"model.tok2vec\"]` See the documentation for details: "
             "https://spacy.io/usage/training#config-components-listeners")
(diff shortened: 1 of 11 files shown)


azure-pipelines.yml
```diff
@@ -41,7 +41,7 @@ jobs:
       matrix:
         # We're only running one platform per Python version to speed up builds
         Python36Linux:
-          imageName: "ubuntu-latest"
+          imageName: "ubuntu-20.04"
```
svlandeg, line 44: Same comment about this edit now being in `v4` ;-)

svlandeg (approved):
Makes sense to clean this up!

I was wondering a little whether we should keep some of the warnings in the documentation, but if you search for "begin_training" you'll get to https://spacy.io/usage/v3/#incompat-api easily, so that should be sufficient.

Issue #11920, opened 2022-12-02
Improve UX around changing displacy's default port


svlandeg:
The default port of `displacy.serve` is 5000, which sometimes causes conflicts as described in https://github.com/explosion/spaCy/pull/11793.

`displacy.serve` already has a port keyword arg, but perhaps adding a global setting/env var would be good, too. Additionally then, it would be nice to try and catch port conflicts and throw a custom spaCy error that tells users how to change the port.

zhiiw:
done

polm:
Thanks for the reminder, I'll close this.

PR #11948, opened 2022-12-08
improve ux for displacy when the serve port is in use

zrpxx:
## Description
Solving #11920.
If the default / specified port for `displacy.serve()` is in use, it will now:
1. try to find the nearest available port and serve on it instead.
2. raise an exception if no available ports were found.

### Types of change
Enhancement

## Checklist

### Existing tests
Test env: `Python 3.9` on `macOS 12`
```
> py.test spacy
===== 3298 passed, 1211 skipped, 32 xfailed, 1 xpassed in 296.98s (0:04:56) ======
```

### Functional test
Since it's a bit tricky to design a test case that can:
- Test the behavior of the underlying WSGI server without modifying the current code structure;
- Capable of running on any port usage situation so everyone can pass it; 

I've only tested the code manually, using a script like:
```python3
import socket
import spacy
from spacy import displacy

doc = nlp("This is a test.")

s1 = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s2 = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s1.bind(("localhost", 65534))
s2.bind(("localhost", 65535))
s1.listen(1)
s2.listen(1)

# This should raise an exception
try:
    displacy.serve(doc, port=65534)
except ValueError as e:
    print("ValueError raised: ", e)

# This should raise a warning
s2.close()
displacy.serve(doc, port=65534)

# This should serve normally
displacy.serve(doc, port=65535)
```
and here is the corresponding output:
```
ValueError raised:  [E1048] No port available found for displacy on host 0.0.0.0. Please specify an available port by `displacy.serve(doc, host, port)`.

======================================

UserWarning: [W124] 0.0.0.0:65534 is already in use, using the nearest available port 65535 as an alternative.
Using the 'dep' visualizer
Serving on http://0.0.0.0:65535 ...

======================================

Using the 'dep' visualizer
Serving on http://0.0.0.0:65535 ...
```


polm:
Thanks for working on this, this would be a great feature to have!

We're a little hesitant to make auto-selecting a port the default behavior. Besides being a change from the current behavior, it can cause unexpected things to happen depending on the user's environment, such as disrupting other servers. 

Could you change it so that the port auto-selection behavior requires a special argument, like a port value of `auto`, before it's used?

zrpxx:
Maybe adding a new argument like `auto_detect_port: bool = True` is a better idea? Cuz passing a string to `port` will introduce some cumbersome checking code.

polm:
Adding a new argument is fine. `auto_select_port` would make more sense, since the selection is the automatic part, and it should be False by default, though we can add a note about turning it on to error messages about busy ports.

zrpxx:
> Adding a new argument is fine. `auto_select_port` would make more sense, since the selection is the automatic part, and it should be False by default, though we can add a note about turning it on to error messages about busy ports.

Absolutely.

spacy/errors.py
```diff
@@ -957,6 +958,10 @@ class Errors(metaclass=ErrorsWithCodes):
     E1046 = ("{cls_name} is an abstract class and cannot be instantiated. If you are looking for spaCy's default "
              "knowledge base, use `InMemoryLookupKB`.")
     E1047 = ("`find_threshold()` only supports components with a `scorer` attribute.")
+    E1048 = ("No port available found for displacy on host {host}. Please specify an available port "
+             "by `displacy.serve(doc, port)`")
+    E1049 = ("Port {port} is already in use. Please specify an available port by `displacy.serve(doc, port) "
+             "or let displacy automatically find a port by `displacy.serve(doc, auto_switch_port=True)`.")
```
polm, line 964:
```suggestion
    E1049 = ("Port {port} is already in use. Please specify an available port with `displacy.serve(doc, port)` "
             "or use `auto_switch_port=True` to pick an available port automatically.")
```

spacy/errors.py
```diff
@@ -957,6 +958,10 @@ class Errors(metaclass=ErrorsWithCodes):
     E1046 = ("{cls_name} is an abstract class and cannot be instantiated. If you are looking for spaCy's default "
              "knowledge base, use `InMemoryLookupKB`.")
     E1047 = ("`find_threshold()` only supports components with a `scorer` attribute.")
+    E1048 = ("No port available found for displacy on host {host}. Please specify an available port "
+             "by `displacy.serve(doc, port)`")
```
polm, line 962:
```suggestion
    E1048 = ("No available port found for displaCy on host {host}. Please specify an available port "
             "with `displacy.serve(doc, port)`")
```

website/docs/api/top-level.md
```diff
@@ -237,16 +237,17 @@ browser. Will run a simple web server.
...
+| `minify`           | Minify HTML markup. Defaults to `False`. ~~bool~~                                                                                                                 |
+| `options`          | [Visualizer-specific options](#displacy_options), e.g. colors. ~~Dict[str, Any]~~                                                                                 |
+| `manual`           | Don't parse `Doc` and instead expect a dict or list of dicts. [See here](/usage/visualizers#manual-usage) for formats and examples. Defaults to `False`. ~~bool~~ |
+| `port`             | Port to serve visualization. Defaults to `5000`. ~~int~~                                                                                                          |
+| `host`             | Host to serve visualization. Defaults to `"0.0.0.0"`. ~~str~~                                                                                                     |
+| `auto_switch_port` | If `True`, automatically switch to a different port if the default port is already in use. Defaults to `False`. ~~bool~~                                          |
```
polm, line 250:
```suggestion
| `auto_switch_port` | If `True`, automatically switch to a different port if the specified port is already in use. Defaults to `False`. ~~bool~~                                          |
```

polm (commented):
Thanks for the updates, I think this design looks pretty good. There are some small fixes needed in the docs. 

We'll also have to add tests for this somehow, I'll think about the right way to do that. It should be possible to background displaCy without too much trouble.

polm:
I added a test for this using multiprocessing. Works on my machine, though no idea if it'll work on Windows - we'll see how the CI does with that.

polm:
That failed rather more than I expected. Let me do a quick check to see if it's timing related, but it wouldn't be surprising if there's OS-related issues too. I'll have to wrap this up next week.

polm:
Testing `displacy.serve` directly is a little tricky because we don't have direct access to the http server, so I factored the port logic into its own util function for easier testing. This has the side benefit that the code can be used elsewhere and keeps the `displacy.serve` code simpler.

polm:
OK, this is almost working, except that on Windows the port seem to be free even after the server is opened.

polm:
Sorry this took me a while to get to. Before today's pushes, tests were failing on Windows, where it looked like the port used by the server was never detected as busy. So this was really unclear from the errors, but it looks like one issue might be that "0.0.0.0" and "localhost" are treated differently on Windows. The main reason that's a problem here is that the port availability was always checked using "localhost", and not the actual host bound, so I've fixed that.

I had some trouble getting this working on my local Windows, but if this doesn't resolve it then the issue is that the httpd server doesn't make the port busy just by being started. In that case we can bind the port directly to make it busy for the test, though I'd prefer to leave the httpd server here to be clearer about what situation we're checking for.

polm:
OK, with the tests passing I believe this should be good to go. I'll check with the team about it, though with the holidays it probably won't get merged until next year.

spacy/displacy/__init__.py
```diff
@@ -93,15 +95,20 @@ def serve(
     manual (bool): Don't parse `Doc` and instead expect a dict/list of dicts.
     port (int): Port to serve visualisation.
     host (str): Host to serve visualisation.
+    auto_select_port (bool): Automatically select a port if the specified port is in use.
 
     DOCS: https://spacy.io/api/top-level#displacy.serve
     USAGE: https://spacy.io/usage/visualizers
     """
     from wsgiref import simple_server
```
rmitsch, line 103: Why do we import `wsgiref` locally?

spacy/util.py
```diff
@@ -1736,3 +1737,40 @@ def all_equal(iterable):
     (or if the input is an empty sequence), False otherwise."""
     g = itertools.groupby(iterable)
     return next(g, True) and not next(g, False)
+
+
+def is_port_in_use(port, host="localhost"):
```
rmitsch, line 1742: - Let's add a docstring here and specify the argument types.
- Is `is_port_in_use()` called outside of `spacy/util.py`? If not (and we don't intend to), I'd suggest renaming this to `_is_port_in_use()`.

spacy/util.py
```diff
@@ -1736,3 +1737,40 @@ def all_equal(iterable):
...
+    try:
+        s.bind((host, port))
+        return False
+    except socket.error:
+        return True
+    finally:
+        s.close()
+
+
+def find_available_port(start, host, auto_select=False):
```
rmitsch, line 1754: Please add types for arguments and return values and fully specify the docstring.

rmitsch (commented):
Looks good - thanks for your contribution, @zrpxx! Only minor questions and comments from my side.

spacy/displacy/__init__.py
zrpxx: this import was [committed](https://github.com/zrpxx/spaCy/commit/c31792aaec68a67ddaac9ef7ea812345eee0a6bf#diff-ec2e6ce08f100314344ecf042422ffcf98220700235baf283104ace0d79a8068) 6 years ago and kept as is, maybe moving it to the top of the file is a good idea?

spacy/util.py
zrpxx: Done.

spacy/util.py
zrpxx: Done.

spacy/displacy/__init__.py
rmitsch: Ah, sorry - let's ignore that. Thought this was introduced by you.

spacy/util.py
```diff
@@ -1736,3 +1737,52 @@ def all_equal(iterable):
...
     g = itertools.groupby(iterable)
     return next(g, True) and not next(g, False)
+
+
+def _is_port_in_use(port: int, host: str = "localhost") -> bool:
+    """Check if 'host:port' is in use. Return True if it is, False otherwise.
+
+    Keyword arguments:
+    port -- the port to check
+    host -- the host to check (default "localhost")
```
rmitsch, line 1747:
```suggestion
    port (int): the port to check
    host (str): the host to check (default "localhost")
    RETURNS (bool): Whether 'host:port' is in use.
```

rmitsch (approved):
Thanks @zrpxx! LGTM, just updated the docstrings to match the spaCy format.

polm:
Thanks for all your work on this, I think it's good to go! Sorry for the noise with the extra commits - I thought there was an oversight in the docs, but I was looking at the wrong place. I'll let the tests run to completion once more and merge this.

PR #11959, opened 2022-12-09
Learn lowercasing as a distinct task within the edit tree lemmatizer

richardpaulhudson:
## Description

In the pre-existing code, the edit tree lemmatizer (ETL) uses standard edit trees to learn when to convert a token (typically its first letter) to lowercase. This is problematic because the patterns around when words are lowercased are largely independent from the morphology of a language, which is the main thing the ETL is learning. This PR sets up lowercasing as a separately learned task within the ETL neural network, hence allowing the edit trees to generalize between capitalized and non-capitalized instances of the same word or of words with the same affixes. It additionally introduces the use of the glorot uniform initializer rather than the zero initializer for the ETL neural network. 

These changes yield an average increase in accuracy of 0.4%, with increases for 14 of the 17 languages tested and no change for the other three (see below).

This PR is one of a series of three  (#11583; #11959; #12017) that increase the accuracy of the edit-tree lemmatizer: the cumulative accuracy improvement is shown in the final section below for each language. The mean morphologizer accuracy increased from 95.7% to 96.2% (+0.5%); the mean lemmatizer accuracy increased from 94.3% to 96.0% (+1.7%). The figures are particularly encouraging for the weaker models: while the pre-existing code yields models with accuracies in the mid-80s% for a few languages, with the cumulative changes at least 91% is achieved for all languages.

### Types of change
Enhancement

## Checklist

#### Accuracy

This table shows the impact of this PR, i.e. of learning lowercasing as a separate task (average improvement: 0.4%). In general, the effect tends to be greater for languages with free word order because such languages display a greater variety of words (and word types) in the initial position of each sentence. Although this PR only affects the accuracy of the lemmatizer, the lemmatizer was trained together with the morphologizer because training it separately has an additional effect on accuracy that would confound the results.

| Model | Lemmatizer accuracy, trained alone with pre-existing code (%) | Lemmatizer accuracy, trained alone with this PR (%) | Difference (%) | 
| --- | --- | --- | --- |
| ca_core_news_lg | 98.7 | 99.0 | 0.3 |
| da_core_news_lg  | 95.2 | 95.2 | *0.0* |
| de_core_news_lg | 97.9 | 98.1 | 0.2 |
| el_core_news_lg | 89.9 | 90.1 | 0.2 |
| es_core_news_lg | 98.4 | 98.9 | 0.5 |
| fi_core_news_lg | 86.4 | 87.0 | 0.6 |
| fr_core_news_lg | 94.9 | 96.0 | **1.1** |
| hr_core_news_lg | 92.9 | 93.5 | 0.6 |
| it_core_news_lg  | 97.6 | 97.6 | *0.0* |
| ko_core_news_lg | 90.2 | 90.2 | *0.0* |
| lt_core_news_lg | 85.8 | 86.9 | **1.1** |
| nb_core_news_lg | 97.4 | 97.5 | 0.1 |
| nl_core_news_lg | 95.7 | 95.8 | 0.1 |
| pl_core_news_lg | 94.3 | 94.7 | 0.4 |
| pt_core_news_lg | 97.3 | 97.5 | 0.2 |
| ro_core_news_lg | 95.8 | 96.1 | 0.3 |
| sv_core_news_lg | 95.4 | 96.1 | 0.7 |

There is one transformer-based model, `de_dep_news_trf`, that uses the edit-tree lemmatizer. Lowercasing improved the accuracy here from 98.7% to 98.9% (+0.2%).

#### Speed

With the CNN models, lemmatizer inference was measured as being 5.2% slower on CPU and 26.1% slower on GPU. The larger difference on GPU is both because GPU is faster to start with and because the lowercasing flags have to be copied across to CPU for processing.

With `de_dep_news_trf`, which was only run on GPU, the speed penalty was 6.4%.

#### Changes

- Considering multiple predictions for each lemma, i.e. setting `top_k = 5` rather than `top_k = 1`, which was made feasible by #12017;
- Feeding suffix and prefix features of various lengths through to `tok2vec` (#11583)
- Adding a dropout layer between `tok2vec` and the components it feeds into, and setting a global dropout probability of 0.2 (#11583)
- Learning lowercasing as a distinct task (this PR)
- Using the glorot uniform initializer for the ETL neural network (this PR)

#### Approaches that were investigated and abandoned

- Modelling regular morphological alternations: many languages have morphological alternations that are applied to a number of different letters or groups of letters in similar grammatical situations. Examples include replacing single letters with double letters or vice versa in Dutch and replacing voiceless consonants with voiced consonants in Croatian. The hope was that supplying the edit tree lemmatizer with the details of any regular alternations in a given language would allow it to apply them as "abstract substitutions" and lead to more parsimonious tree structures and more efficient and accurate learning. However, this did not work as expected.
- Assessing predictions using n-gram fitness: The n-gram frequencies observed in the training set can be used to determine the acceptability of a lemma prediction. The hope was that, in conjunction with setting `top_k = 5`, this information could be used to filter out incorrect predictions; however, the positive effect on precision was accompanied by a more or less equal negative effect on recall.

#### Accuracy

The mean morphologizer accuracy increased from 95.7% to 96.2% (+0.5%); the mean lemmatizer accuracy increased from 94.3% to 96.0% (+1.7%):

| Model | Morph. acc. old (%) | Morph. acc. new (%) | Morph. acc. diff (%) | Lemm. acc. old (%) | Lemm. acc. new (%) | Lemm. acc. diff (%) |
| --- | --- | --- | --- | --- | --- | --- | 
| ca_core_news_lg | 98.2 | 98.2 | *0.0* | 98.7 | 99.1 | 0.4 |
| da_core_news_lg | 95.3 | 96.0 | 0.7 | 95.2 | 96.5 | **1.3** |
| de_core_news_lg | 92.2 | 92.6 | 0.4 | 97.9 | 98.6 | 0.7 |
| el_core_news_lg | 91.0 | 92.0 | **1.0** | 89.9 | 92.9 | **3.0** |
| es_core_news_lg | 98.2 | 98.3 | 0.1 | 98.4 | 99.1 | 0.7 |
| fi_core_news_lg | 92.2 | 93.6 | **1.4** | 86.4 | 91.0 | **4.6** |
| fr_core_news_lg | 96.8 | 97.1 | 0.3 | 94.9 | 96.9 | **2.0** |
| hr_core_news_lg | 92.8 | 92.6 | *-0.2* | 92.9 | 94.7 | **1.8** |
| it_core_news_lg | 97.4 | 97.8 | 0.4 | 97.6 | 98.2 | 0.6 |
| ko_core_news_lg | - | - | - | 90.2 | 91.8 | **1.6** |
| lt_core_news_lg | 89.0 | 90.9 | **1.9** | 85.8 | 91.2 | **5.4** |
| nb_core_news_lg | 96.3 | 96.8 | 0.5 | 97.2 | 98.0 | 0.8 |
| nl_core_news_lg | 96.4 | 96.7 | 0.3 | 95.7 | 96.5 | 0.8 |
| pl_core_news_lg | 90.8 | 91.2 | 0.4 | 94.3 | 96.4 | **2.1** |
| pt_core_news_lg | 95.8 | 96.1 | 0.3 | 97.3 | 97.8 | 0.5 |
| ro_core_news_lg | 95.1 | 97.1 | **2.0** | 95.8 | 97.0 | **1.2** |
| sv_core_news_lg | 95.7 | 96.2 | 0.5 | 95.4 | 97.2 | **1.8** |

There is one transformer-based model, `de_dep_news_trf`, that uses the edit-tree lemmatizer. Three of the five changes listed above are relevant to transformer models; applying them increased the accuracy from 98.7% to 98.9% (+0.2%).

#### Speed

With the CNN models, lemmatizer and morphologizer inference with `pl_core_news_lg` were measured as being 12.2% slower on CPU and 42.4% slower on GPU with the five changes listed above included than without them.

With `de_dep_news_trf`, which was only run on GPU, the speed penalty was 7.3%.


spacy/ml/models/lemmatizer.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    with Model.define_operators({">>": chain, "|": concatenate}):
+        t2v_width = tok2vec.get_dim("nO") if tok2vec.has_dim("nO") else None
+        softmax = Softmax_v2(
+            nO, t2v_width, init_W=glorot_uniform_init, normalize_outputs=normalize
+        )
+        model = tok2vec >> with_array(softmax)
+        if lowercasing:
+            lowercasing_output = Sigmoid(1)
+            sigmoid_appendage = Relu(50) >> Dropout(0.2) >> lowercasing_output
```
danieldk, line 41: These hyperparameters should probably be arguments of the layer. I think we also usually override dropouts in training anyway?

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -115,29 +117,65 @@ def __init__(
...
         self, examples: Iterable[Example], scores: List[Floats2d]
     ) -> Tuple[float, List[Floats2d]]:
         validate_examples(examples, "EditTreeLemmatizer.get_loss")
-        loss_func = SequenceCategoricalCrossentropy(normalize=False, missing_value=-1)
-
-        truths = []
+        tree_loss_func = SequenceCategoricalCrossentropy(
+            normalize=False, missing_value=-1
+        )
+        lowercasing_loss_func = L2Distance(normalize=False)
```
danieldk, line 129: Since the layer is the logistic function (sigmoid), I'd expect a binary cross entropy loss. Due to how we handle gradients, I think L2Distance will have the same effect in this case, but it's not nice semantically and it will report incorrect losses.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -115,29 +117,65 @@ def __init__(
...
         for eg in examples:
-            eg_truths = []
+            eg_tree_truths = []
+            eg_lowercasing_truths = []
             for (predicted, gold_lemma) in zip(
                 eg.predicted, eg.get_aligned("LEMMA", as_string=True)
             ):
                 if gold_lemma is None:
                     label = -1
+                    eg_lowercasing_truths.append([0])
```
danieldk, line 140: Hmmm, this makes me believe we should perhaps even use the current cross-entropy implementation? We don't want to train absent annotations as a particular class.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -151,15 +189,21 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
...
         assert len(guesses) == n_docs
         return guesses
 
-    def _scores2guesses(self, docs, scores):
+    def _scores2guesses(
+        self, docs, scores, lowercasing_flags: Optional[List[Floats2d]]
+    ):
         guesses = []
-        for doc, doc_scores in zip(docs, scores):
+        for i, (doc, doc_scores) in enumerate(zip(docs, scores)):
```
danieldk, line 206: three-arg `zip`?

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -169,15 +213,19 @@ def _scores2guesses(self, docs, scores):
                 doc_guesses = doc_guesses.get()
 
             doc_compat_guesses = []
-            for token, candidates in zip(doc, doc_guesses):
+            for j, (token, candidates) in enumerate(zip(doc, doc_guesses)):
```
danieldk, line 216: three-arg `zip`?

danieldk (commented):
Cool, this is such a nice extension! 🎉.

spacy/ml/models/lemmatizer.py
shadeMe, line 41: Would it make sense to expose the hidden layer size and dropout probability to the config as hyperparameters?

spacy/pipeline/edit_tree_lemmatizer.py
richardpaulhudson: As we discussed offline, you plan to add a new loss function to Thinc to enable this.

spacy/pipeline/edit_tree_lemmatizer.py
richardpaulhudson: See above.

danieldk:
I am going to unassign this PR from myself, because I don't want to block anyone else from picking this up, but summary of what still needs to be done: we need a binary cross entropy loss in Thinc to support the sigmoid layer. This PR should also be changed to correctly handle missing annotations (we don't want gradients for missing annotations).

svlandeg:
We want to pick up this work and properly finish it, but we can't prioritize this right now. I'm closing this PR but we still have it on our internal backlog, to pick it up sometime in the future.

PR #11972, opened 2022-12-14
New console logger with expanded progress tracking

shadeMe:
## Description

This PR adds `ConsoleLogger.v3` which expands the progress bar feature to count up the training/distillation steps to either the next evaluation pass or the maximum number of steps.

Additionally, `ConsoleLogger.v2` was modified to call `v3` while retaining its current behaviour.

### Types of change

New feature

## Checklist

### DIFF
--- a/spacy/training/loggers.py
+++ b/spacy/training/loggers.py
@@ -40,7 +40,7 @@ def console_logger(
     output_file (Optional[Union[str, Path]]): The file to save the training logs to.
     """
     return console_logger_v3(
-        progress_bar=None if progress_bar is False else "eval_steps",
+        progress_bar=None if progress_bar is False else "eval",
         console_output=console_output,
         output_file=output_file,
     )
@@ -54,8 +54,8 @@ def console_logger_v3(
 ):
     """The ConsoleLogger.v3 prints out training logs in the console and/or saves them to a jsonl file.
     progress_bar (Optional[str]): Type of progress bar to show in the console. Allowed values:
-        all_steps - Tracks the number of steps until `training.max_steps` is reached.
-        eval_steps - Tracks the number of steps until `training.eval_frequency` is reached.
+        train - Tracks the number of steps from the beginning of training until the full training run is complete (training.max_steps is reached).
+        eval - Tracks the number of steps between the previous and next evaluation (training.eval_frequency is reached).
     console_output (bool): Whether the logger should print the logs on the console.
     output_file (Optional[Union[str, Path]]): The file to save the training logs to.
     """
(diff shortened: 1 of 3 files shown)


spacy/training/loggers.py
```diff
@@ -26,14 +26,36 @@ def setup_table(
...
+@registry.loggers("spacy.ConsoleLogger.v3")
+def console_logger_v3(
+    progress_bar: Optional[str] = None,
+    console_output: bool = True,
+    output_file: Optional[Union[str, Path]] = None,
+):
+    """The ConsoleLogger.v3 prints out training logs in the console and/or saves them to a jsonl file.
+    progress_bar (Optional[str]): Type of progress bar to show in the console. Allowed values:
+        all_steps - Tracks the number of steps until `training.max_steps` is reached.
+        eval_steps - Tracks the number of steps until `training.eval_frequency` is reached.
```
shadeMe, line 58: Not too happy with the naming of either type, to be honest. Suggestions are welcome.

kadarakos: Maybe it could be just called `progress` and its might be clearer like `--progress train` or `--progress eval`?

And for the explanation maybe like "train - Track the number of steps from the beginning of training until the full training run is complete (training.max_steps is reached). eval - Track the number of steps between the previous and next evaluation (training.eval_frequency is reached)".

I would just give this kind of explanation there because i personally had to double take a bit on what do these mean exactly.

shadeMe: That sounds better, yeah. Thanks, will change it.

polm (commented):
This looks good, thanks for putting it together! 

One thing that came up is that the docs didn't explain defaults - it looks like they had been like that before, but we should clarify it. 

One thing came to mind looking at the code here was whether the case where `max_steps = 0` would work, but it looks like tqdm will handle that reasonably by showing progress without a bar.

polm (approved):
Thanks, there were a few small punctuation issues, but this should be good now!

spacy/training/loggers.py
```diff
@@ -26,14 +26,36 @@ def setup_table(
     return final_cols, final_widths, ["r" for _ in final_widths]
 
 
+# We cannot rename this method as it's directly imported
+# and used by external packages such as spacy-loggers.
```
svlandeg, line 30: We probably want to change that in `spacy-loggers` and get the function through the registry instead. (clearly that's not a blocker for this PR though)

spacy/training/loggers.py
```diff
@@ -84,6 +107,13 @@ def setup_printer(
             write(msg.row(table_header, widths=table_widths, spacing=spacing))
             write(msg.row(["-" * width for width in table_widths], spacing=spacing))
         progress = None
+        expected_progress_types = ("train", "eval", None)
+        if progress_bar not in expected_progress_types:
+            raise ValueError(
+                Errors.E1048.format(
+                    unexpected=progress_bar, expected=expected_progress_types
+                )
+            )
```
svlandeg, line 116: I find it a bit odd to add `None` to the printed types here, instead I'd do 
```
if progress_bar is not None and progress_bar not in expected_progress_types
    ....
```

svlandeg (commented):
Looking good! We'll want to move the v2 to legacy though, right?

spacy/training/loggers.py
shadeMe: Done.

svlandeg (approved):
Nice feature, thanks!

PR #11974, opened 2022-12-14
Minor website fixes (#11974) [ci skip]

essenmitsosse:
Those are a few minor fixes before the actual migration to Next can happen. If someone wants to take a look at the changes it's best to walk through them commit by commit. I cleaned up all commits to be one atomic change each.

Feel free to squash.


website/docs/api/sentencizer.mdx
```diff
@@ -38,7 +38,7 @@ how the component should be configured. You can override its settings via the
 > ```
 
 | Setting                                  | Description                                                                                                                                            |
-| ---------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| ---------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ | ------ |
 | `punct_chars`                            | Optional custom list of punctuation characters that mark sentence ends. See below for defaults if not set. Defaults to `None`. ~~Optional[List[str]]~~ | `None` |
```
svlandeg, line 42:
```suggestion
| `punct_chars`                            | Optional custom list of punctuation characters that mark sentence ends. See below for defaults if not set. Defaults to `None`. ~~Optional[List[str]]~~ |
```

essenmitsosse: ✅

website/docs/api/morphology.mdx
```diff
@@ -105,11 +105,11 @@ representation.
...
-| ------------- | ------------------------------------------------------------------------------------------------------------------------------ |
-| `FEATURE_SEP` | The [FEATS](https://universaldependencies.org/format.html#morphological-annotation) feature separator. Default is `|`. ~~str~~ |
-| `FIELD_SEP`   | The [FEATS](https://universaldependencies.org/format.html#morphological-annotation) field separator. Default is `=`. ~~str~~   |
-| `VALUE_SEP`   | The [FEATS](https://universaldependencies.org/format.html#morphological-annotation) value separator. Default is `,`. ~~str~~   |
+| Name          | Description                                                                                                                  |
+| ------------- | ---------------------------------------------------------------------------------------------------------------------------- | ---------- |
+| `FEATURE_SEP` | The [FEATS](https://universaldependencies.org/format.html#morphological-annotation) feature separator. Default is `          | `. ~~str~~ |
```
svlandeg, line 110: Oh, this is interesting, we never caught this before: the `|` is not ignored by prettier but actually used for formatting the table, resulting in a stray column.

website/docs/images/displacy-ent1.html
```diff
@@ -1,36 +1,83 @@
 <div
     class="entities"
-    style="line-height: 2.5; font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'; font-size: 16px"
+    style="
+        line-height: 2.5;
+        font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif,
+            'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+        font-size: 16px;
+    "
```
svlandeg, line 8: Where do these formatting edits come from?

website/docs/usage/visualizers.mdx
```diff
@@ -118,12 +118,16 @@ displacy.serve(doc, style="ent")
...
+  html={DisplacyEntHtml}
+  height={180}
+/>
 
 The entity visualizer lets you customize the following `options`:
 
 | Argument | Description                                                                                                   |
-| -------- | ------------------------------------------------------------------------------------------------------------- |
+| -------- | ------------------------------------------------------------------------------------------------------------- | ------ |
 | `ents`   | Entity types to highlight (`None` for all types). Defaults to `None`. ~~Optional[List[str]]~~                 | `None` |
```
svlandeg, line 131:
```suggestion
| `ents`   | Entity types to highlight (`None` for all types). Defaults to `None`. ~~Optional[List[str]]~~                 |
```
(I realise that this PR isn't at fault here!)

website/meta/languages.json
```diff
@@ -36,12 +36,7 @@
             "name": "Catalan",
             "example": "Això és una frase.",
             "has_examples": true,
-            "models": [
-                "ca_core_news_sm",
-                "ca_core_news_md",
-                "ca_core_news_lg",
-                "ca_core_news_trf"
-            ]
+            "models": ["ca_core_news_sm", "ca_core_news_md", "ca_core_news_lg", "ca_core_news_trf"]
```
svlandeg, line 39: Again, wondering why the formatting has to change?

svlandeg (commented):
Thanks for all of your work on this!

While I think it's fine to apply `prettier` to the `mdx` files like we typically do, I'm more unsure about all the formatting in other file types, as it makes this PR more difficult to review and it makes it less clear afterwards which were the actual fixes that needed to happen. I would have preferred keeping all the formatting for a separate PR.

website/UNIVERSE.md
```diff
@@ -2,93 +2,103 @@
...
-    "thumb": "https://example.com/thumb.jpg",
-    "image": "https://example.com/image.jpg",
-    "author": "Your Name",
-    "author_links": {
-        "twitter": "username",
-        "github": "username",
-        "website": "https://example.com"
-    },
-    "category": ["pipeline", "standalone"],
-    "tags": ["some-tag", "etc"]
```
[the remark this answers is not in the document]
essenmitsosse: Oh that's interesting. Prettier for `.mdx` files is setup to indent with two spaces, so this also applies to code within the `.mdx` files. Easiest fix would be to change indentation for `.mdx` to 4 spaces as well, since it's the only file type were we currently make an exception.

website/docs/api/morphology.mdx
essenmitsosse: Fixed

website/docs/images/displacy-ent1.html
essenmitsosse: Everything in 94beed2 is just automatically applying Prettier to all files.

website/meta/languages.json
essenmitsosse: Basically because "Prettier says so". I just ran it to apply to all files that it would apply to. Anyone who has the Prettier extension installed with fix-on-save would get the same result when opening and saving this file.

essenmitsosse:
All the Prettier changes are in a single commit. Ideally this PR should be looked at at a per-commit basis. I can pull the Prettier changes out — but in that case, should the rest stay together or should we split that up as well?

website/docs/usage/visualizers.mdx
essenmitsosse: So revert the changes from [2613dbb](https://github.com/explosion/spaCy/pull/11974/commits/2613dbb45efbc5d1ccd4e402ff09c93a49404f02), is that correct?

svlandeg:
What if you cherrypick the formatting with prettier commit and PR that separately, we get that merged in the migration branch, and then update this PR? Then the diff will only show the non-formatting changes, right? And that should hopefully be not too much work on your end. I know that does make the history a little more unclear, but with the trade-off that it makes this PR easier to review from diff-view (which is what we normally do and I prefer to get a holistic view on the changes introduced by one PR)

essenmitsosse:
Switching this to draft, until [#11996 Apply Prettier](https://github.com/explosion/spaCy/pull/11996) is merged.

essenmitsosse:
@svlandeg if it helps, I can also pull out [fb5f7b0](https://github.com/explosion/spaCy/pull/11974/commits/fb5f7b0dc5906147b9fdca9bca92ee185fc7a2e4), because that's the only larger complex commit in this PR.

essenmitsosse:
Rebase the branch on `origin/website/migration`, so that the "Apply Prettier" commit is removed from this PR. Reopening.

website/src/templates/index.js
```diff
@@ -197,7 +197,7 @@ class Layout extends React.Component {
                     search={<Search settings={siteMetadata.docSearch} />}
                     alert={nightly ? null : navAlert}
                 >
-                    <Progress key={location.href} />
```
ines, line 199: I think the idea her was to ensure the progress bar is reset on site changes, which would otherwise not be the case. Can you think of a better way to solve this? Alternatively, I think we probably need to leave this in.

(Sorry, currently battling some Gatsby errors and haven't managed to run the site locally yet from this PR to check!)

essenmitsosse: I tested it, and it's also gone in the final [deploy preview](https://deploy-preview-11967--spacy.netlify.app/) and it is still working. As per [3f00d7d](https://github.com/explosion/spaCy/pull/11974/commits/3f00d7d9386e3ef3299cf60ad78a6cb870970ea8) this was removed to reduce the reliance of the template on `location`.

PR #12005, opened 2022-12-20
Migrate to Next WEB-17

essenmitsosse:
# [🖥️ Deploy Preview](https://deploy-preview-12005--spacy.netlify.app/)

# Intro 
This is the big one. I apologize in advance for the size (see below for more details)

This PR contains all the necessary work to migrate from Gatsby to Next. Even without `package-lock.json` these are ~4000LOC. The plan was to change as little as possible. There are no unnecessary bug fixes, no code style improvements, and no syntax updates unless absolutely necessary or in cases where it would have made life much harder to stick with the old implementation. This is the minimal amount of changes I could make to get the same site as before on a different framework.

# Main changes (relevant commits linked in headings)
## Routing
This one is pretty straightforward. All the routing is rewritten from scratch, because there is no similarities between Gatsby and Next as to how they handle routing. The site now uses file based routing as is common for Next. New routes can still be created by adding `.mdx` to `/website/docs` and will automatically be picked up by `/website/pages/[...].ts`

## MDX
MDX2 has a lot of breaking changes and due to the heavily customized nature of our MDX syntax (mainly through use of plugins and custom components), some changes to our MDX syntax were unavoidable. The reasoning for all of those changes is, that the old syntax is now invalid and makes MDX fail even before a plugin could handle the syntax. The syntax changes are:

### Heading meta data (94aa3629bb56a85dd66e1aa3fb92b5d30ba869e7)
```diff
- ## HEADING TITLE {#ID tag="method"}
+ ## HEADING TITLE {id="ID",tag="method"}
```

### Code block headings (b305510227776bc5c7af30b081575e4318fc1b47)
```diff
- ```python
- ### THE TITLE
+ ```python {title="THE TITLE"}
# PYTHON CODE
\```
```
New syntax (similar to the new one for heading meta data, now also allows for more meta data on code blocks, which might be useful in the future)

### Headings with Logo (7dad1596f7345b4e023dc8654413bb612cabbdc6)
Old syntax (only used in `usage/projects`)
```diff
- ### HEADING TEXT {id="THIS_ID"} <IntegrationLogo name="NAME" title="TITLE" width={999} height="auto" align="right" />
+ {<H3 id="THIS_ID"> HEADING TEXT
+
+ <IntegrationLogo name="NAME" title="TITLE" width={999} height="auto" align="right" />
+ </H3>}
```

This is unfortunately less elegant, but the only way to get another component inside a heading in MDX2.

### Rename tag `new` to `version` (94aa3629bb56a85dd66e1aa3fb92b5d30ba869e7)

There was a reserved keyword collision with the `new` tag so it is changed to `version`, since this was already used in other places in the code already.

### Remove component imports from MDX (84db6ea20f25b14e62d02c5f5cf7288ea39f1b42)

```diff
- import SOME_COMPONENT from 'src/someComponent'
```

While theoretical still supported in MDX2, this is not compatible with (`next-mdx-remote`)[https://github.com/hashicorp/next-mdx-remote]. There for all imports from MDX files are removed and all components that can be used in MDX have to be defined in `website/src/remark.js`

## Syntax Highlighting in Codeblocks (b305510227776bc5c7af30b081575e4318fc1b47)
While functionally the same the implementation was completely rewritten.

## TypeScript
TypeScript support has been added. No existing code was rewritten in TypeScript thought. So far it is only used for new files (mostly routing). But the option to convert everything to TypeScript is there.

## Linting & Prettier
There should be no more linting errors when running `npm run lint`, and nothing for Prettier to change when running `npm run prettier`.

## Other changes
Most other change just reimplement old features in Next (PWA, `<head lang="en">`, Plausible, Sitemap etc.)

# Why is this such a big PR?
The reason all of these changes are in one PR is, that the alternative would be to start with a default Next site and then do atomic PRs until we got the old site back (which was basically how I worked). But this would also mean removing a lot of content until the system supports it. This way you would be looking at a broken site — which wouldn't be very helpful. 

We can break the PR up, but then we don't have a working state to look at.

# Structure of this PR
Since this is such a huge PR, I recommend looking at the individual commits to not be overwhelmed. All commits are atomic changes. There are a few commits at the beginning that start from a new initial commit (where I setup a basic Next site), before being merged with the existing work in `website/migration`.

The commit structure looks as follows:
Screenshot 2022-12-20 at 22 21 23

Since this PR is going to be squashed, there is not to much point in this approach anymore, but I also saw no reason to flatten those commits for the PR.


website/.prettierrc
```diff
@@ -20,12 +20,11 @@
             }
         },
         {
-            "files": "*.md",
+            "files": ["package.json", "package-lock.json"],
```
svlandeg, line 23: Can you explain this change? It used to be so that `prettier` would only be applied to `md` files? But you've extended that to more file types in https://github.com/explosion/spaCy/pull/11996?

website/.prettierrc
```diff
@@ -20,12 +20,11 @@
...
             "options": {
                 "tabWidth": 2,
                 "printWidth": 80,
-                "proseWrap": "always",
-                "htmlWhitespaceSensitivity": "strict"
+                "proseWrap": "always"
```
svlandeg, line 27: Is this a change that is absolutely necessary?

website/.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+    "recommendations": [
+        "dbaeumer.vscode-eslint",
+        "unifiedjs.vscode-mdx",
+        "esbenp.prettier-vscode",
+        "syler.sass-indented"
+    ]
+}
```
svlandeg, line 8: Sorry if I missed earlier conversations on this, can you provide a bit more background information about these?

website/.prettierrc
ines: Looks like this change is just applies specific formatting rules to those two JSON files. The diff here is potentially a bit misleading because it combined the addition of the new rule and the removal of an old rule (`.md` is not needed anymore, our files are all `.mdx` now).

website/.prettierrc
ines: Since this is now the rule applied to the JSON files (see my previous comment on slightly misleading diff), `htmlWhitespaceSensitivity` is just a setting not needed anymore since it's the formatting rules for JSON, not Markdown.

website/.vscode/extensions.json
svlandeg: Thanks!

svlandeg (approved):
This has been a huge amount of effort, thanks Marcus! The deploy preview is looking really good.

It's all looking really good. Let's leave this open for maybe one more day, in case anyone spots anything off on the deploy preview, and if not this should be good to merge 🎉

website/.vscode/extensions.json
essenmitsosse: Also VSCode will offer you to automatically activate these extensions if you choose so. Especially when switching projects a lot this is quite handy.

website/.prettierrc
essenmitsosse: Sometimes git just finds the most co fusing way to display a diff 😅

website/docs/usage/visualizers.mdx
```diff
@@ -116,11 +114,9 @@ doc = nlp(text)
...
 ```
 
-import DisplacyEntHtml from 'images/displacy-ent2.html'
-
 <Iframe
   title="displaCy visualizer for entities"
-  html={DisplacyEntHtml}
+  src="/images/displacy-ent2.html"
   height={180}
 />
```
svlandeg, line 122: The table below this line is badly formatted (from before), the last column should be removed from the 2nd and 3rd line. It's currently causing the table not to render properly: https://deploy-preview-12005--spacy.netlify.app/usage/visualizers#ent

website/docs/usage/visualizers.mdx
```diff
@@ -227,23 +218,20 @@ specify which one displaCy should use with `spans_key` (`sc` is the default).
 > displacy.serve(doc, style="span", options=options)
 > ```
 
-import DisplacySpanCustomHtml from 'images/displacy-span-custom.html'
-
 <Iframe
   title="displaCy visualizer for spans (custom spans_key)"
-  html={DisplacySpanCustomHtml}
+  src="images/displacy-span-custom.html"
```
svlandeg, line 223:
```suggestion
  src="/images/displacy-span-custom.html"
```

PR #12006, opened 2022-12-21
Move Entity Linker v1 to spacy-legacy

polm:
## Description

This is a follow up to #11889 that moves the component instead of removing it.

In general, we never import from spacy-legacy in spaCy proper. However, to use this component, that kind of import will be necessary. I was able to test this without issues, but is this current import strategy acceptable? Or should we put the component in a registry?

### Types of change

legacy cleanup

## Checklist


polm:
Note tests will fail here unless coordinated with spacy-legacy changes to add the component.

polm:
Closing/reopen and draft conversion was an attempt to get tests working, but the issue is that the recent commit only modified workflows, which are ignored. I'll use a dummy commit we can squash later to trigger this.

polm:
The tests are still failing, but finally after explicitly uninstalling spacy-legacy before exlicitly installing the git version, it seems this is finally installing the branch as intended. 

I'll look into what's causing the other errors - they seem to be unrelated and may just be related to the state of master when this was started.

polm:
Still not sure what was up with `requirements.txt` pulling in v9 of Thinc, but that is resolved now. Unfortunately it looks like unrelated tests are failing, I'll look in to why.

.github/azure-steps.yml
```diff
@@ -99,10 +99,15 @@ steps:
 #    condition: eq(variables['python_version'], '3.8')
 
   - script: |
+      cat requirements.txt
       python -m pip install -U -r requirements.txt
+      python -m pip uninstall -y spacy-legacy
+      python -m pip install "git+https://github.com/polm/spacy-legacy.git@feature/entity-linker-component-v1"
```
polm, line 105: I'm not sure why this is necessary to get it working, but it should be removed before merging anyway. It will need to wait on https://github.com/explosion/spacy-legacy/pull/31

requirements.txt
```diff
@@ -1,5 +1,7 @@
 # Our libraries
-spacy-legacy>=3.0.10,<3.1.0
+#spacy-legacy>=3.0.10,<3.1.0
+# XXX TEMP FOR TESTING again
+git+https://github.com/polm/spacy-legacy.git@feature/entity-linker-component-v1
```
polm, line 4: To be removed / reverted.

spacy/pipeline/entity_linker.py
```diff
@@ -120,6 +119,13 @@ def make_entity_linker(
     """
 
     if not model.attrs.get("include_span_maker", False):
+        try:
+            from spacy_legacy.components.entity_linker import EntityLinker_v1
+        except:
+            raise ImportError(
+                """In order to use v1 of the EntityLinker, you must use
+                    spacy-legacy >=XXX."""
```
polm, line 127: TODO: update with spacy-legacy release version.

spacy/tests/package/test_requirements.py
```diff
@@ -1,7 +1,9 @@
 import re
 from pathlib import Path
+import pytest
 
 
+@pytest.mark.xfail(reason="Using spacy-legacy branch for PR")
```
polm, line 6: Needs to be removed before merging.

polm:
Most of the diffs in this PR ended up being stuff to make sure the right branch of spacy-legacy is installed for CI, but the good news is that the tests run fine.

polm:
Yes, this is ready to be reviewed. It is marked as draft because it cannot be merged as-is, due to all the commented changes required to get CI working. Sorry that wasn't clear.

polm:
I will also note the supposed conflict can also be ignored, since it is also just related to the branch change.

polm:
Thanks to the spacy-experimental release I was able to remove all the temporary changes and the tests have passed, so this should be good to go. (I missed one debugging line, but it shouldn't change the test results.)

svlandeg:
@polm : could you address the merge conflicts? Then I had a few smaller comments, but otherwise this should be good to go :-)

polm:
Took care of the conflicts - it was just the `spacy-legacy` update.

spacy/pipeline/entity_linker.py
```diff
@@ -120,6 +119,13 @@ def make_entity_linker(
     """
 
     if not model.attrs.get("include_span_maker", False):
+        try:
+            from spacy_legacy.components.entity_linker import EntityLinker_v1
+        except:
+            raise ImportError(
+                """In order to use v1 of the EntityLinker, you must use
+                    spacy-legacy>=3.0.12."""
```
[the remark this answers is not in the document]
polm: Changed to single quotes.

polm:
Thanks for the comments, I think they should all be taken care of now, but let me know if there's anything else.

PR #12016, opened 2022-12-22
Add `TrainablePipe.{distill,get_teacher_student_loss}`

danieldk:
## Description

This is the first PR in a range of PRs that add support for distillation. This change adds two methods:

- `TrainablePipe::distill` which performs a training step of a student pipe on a teacher pipe, giving a batch of `Doc`s.
- `TrainablePipe::get_teacher_student_loss` computes the loss of a student relative to the teacher.

The `distill` or `get_teacher_student_loss` methods are also implemented in the tagger, edit tree lemmatizer, and parser pipes, to enable distillation in those pipes and as an example for other pipes.

One question that might come up it how this relates to the rehearsal functionality that is already present in spaCy. The distillation support is very similar to rehearsal. However, the existing `rehearse` methods do not match what we need for distillation:

```
def rehearse(self, examples, *, drop=0., sgd=None, losses=None):
    # ...
   teacher_tag_scores, _ = self._rehearsal_model.begin_update(docs)
```

The issues are:

* In distillation we also need access to the teacher's pipe. For instance, in tok2vec pipes, we need the teacher's pipe to properly add the outputs to the document.
* We usually perform distillation on unannotated data, so the inputs to distillation are `Doc`s, not `Example`s.
* In distillation we need to pass a separate set of `Doc`s to the student and the teacher, since they may add different annotations (cf. `annotating_components`).

Where possible, the gradients/losses of `rehearse` and `distill` are both calculated with the `get_teacher_student_loss` method. The current exception being the parser, since we use a normalized loss in `rehearse` and an unnormalized loss in `distill`. We still have to investigate empirically what the difference is in practice an unify them when we can use one loss for both methods.

Many thanks to @shadeMe for reviewing the first version of this PR in https://github.com/danieldk/spaCy/pull/3

### Types of change

New feature.

## Checklist


svlandeg:
> This is the first PR in a range of PRs that add support for distillation. 

Super excited about this!!!

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -169,6 +219,17 @@ cdef class TrainablePipe(Pipe):
...
+    def get_teacher_student_loss(self, teacher_scores, student_scores):
+        """Calculate the loss and its gradient for a batch of student
+        scores, relative to teacher scores.
+
+        teacher_scores: Scores representing the teacher model's predictions.
+        student_scores: Scores representing the student model's predictions.
+
+        DOCS: https://spacy.io/api/pipe#get_teacher_student_loss
+        """
+        raise NotImplementedError(Errors.E931.format(parent="TrainablePipe", method="get_teacher_student_loss", name=self.name))
```
svlandeg, line 231: It would be nice to have a mechanism that can tell you whether distillation is in fact implemented for a given pipe, something like `is_distillable()`. As we have a default dummy implementation here of `get_teacher_student_loss` we can't implement it checking whether the right functions are available though, so maybe it should be `False` by default and pipes should set it specifically to `True` when they've implemented the behaviour fully?

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -56,6 +56,56 @@ cdef class TrainablePipe(Pipe):
...
+            different annotations.
+        drop (float): dropout rate.
+        sgd (Optional[Optimizer]): An optimizer. Will be created via
+            create_optimizer if not set.
+        losses (Optional[Dict[str, float]]): Optional record of loss during
+            distillation.
+        RETURNS: The updated losses dictionary.
+        """
+        # By default we require a teacher pipe, but there are downstream
+        # implementations that don't require a pipe.
```
svlandeg, line 86: Such as? (I'm curious)

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -56,6 +56,56 @@ cdef class TrainablePipe(Pipe):
...
+        """Train a pipe (the student) on the predictions of another pipe
+        (the teacher). The student is typically trained on the probability
+        distribution of the teacher, but details may differ per pipe.
+
+        teacher_pipe (Optional[TrainablePipe]): The teacher pipe to learn
+            from.
+        teacher_docs (Iterable[Doc]): Documents passed through teacher pipes.
+        student_docs (Iterable[Doc]): Documents passed through student pipes.
+            Must contain the same tokens as `teacher_docs` but may have
+            different annotations.
```
svlandeg, line 77: Can we provide a utility that checks this requirement up-front?

spacy/errors.py
```diff
@@ -949,6 +949,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E4000 = ("Expected a Doc as input, but got: '{type}'")
     E4001 = ("Expected input to be one of the following types: ({expected_types}), "
              "but got '{received_type}'")
+    E4002 = ("Pipe '{name}' requires teacher pipe for distillation.")
```
svlandeg, line 952:
```suggestion
    E4002 = ("Pipe '{name}' requires a teacher pipe for distillation.")
```

svlandeg, line 952:
```suggestion
    E4002 = ("Pipe '{name}' requires a teacher pipe for distillation.")
```

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -56,6 +56,56 @@ cdef class TrainablePipe(Pipe):
...
+        # implementations that don't require a pipe.
+        if teacher_pipe is None:
+            raise ValueError(Errors.E4002.format(name=self.name))
+        if losses is None:
+            losses = {}
+        losses.setdefault(self.name, 0.0)
+        if not any(len(doc) for doc in teacher_docs):
+            return losses
+        if not any(len(doc) for doc in student_docs):
+            return losses
```
svlandeg, line 95: This shouldn't happen - we should ensure that `teacher_docs` and `student_docs` are compatible up-front

website/docs/api/pipe.md
```diff
@@ -234,6 +234,33 @@ predictions and gold-standard annotations, and update the component's model.
 | `losses`       | Optional record of the loss during training. Updated using the component name as the key. ~~Optional[Dict[str, float]]~~ |
 | **RETURNS**    | The updated `losses` dictionary. ~~Dict[str, float]~~                                                                    |
 
+## TrainablePipe.distill {#rehearse tag="method,experimental" new="4"}
```
svlandeg, line 237:
```suggestion
## TrainablePipe.distill {#distill tag="method,experimental" new="4"}
```

website/docs/api/pipe.md
```diff
@@ -281,6 +308,33 @@ This method needs to be overwritten with your own custom `get_loss` method.
 | `scores`    | Scores representing the model's predictions.                                |
 | **RETURNS** | The loss and the gradient, i.e. `(loss, gradient)`. ~~Tuple[float, float]~~ |
 
+## TrainablePipe.get_teacher_student_loss {#get_teacher_student_loss tag="method"}
```
svlandeg, line 311:
```suggestion
## TrainablePipe.get_teacher_student_loss {#get_teacher_student_loss tag="method" new="4"}
```

svlandeg, line 311:
```suggestion
## TrainablePipe.get_teacher_student_loss {#get_teacher_student_loss tag="method" new="4"}
```

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -56,6 +56,56 @@ cdef class TrainablePipe(Pipe):
...
+        teacher_docs (Iterable[Doc]): Documents passed through teacher pipes.
+        student_docs (Iterable[Doc]): Documents passed through student pipes.
+            Must contain the same tokens as `teacher_docs` but may have
+            different annotations.
+        drop (float): dropout rate.
+        sgd (Optional[Optimizer]): An optimizer. Will be created via
+            create_optimizer if not set.
+        losses (Optional[Dict[str, float]]): Optional record of loss during
+            distillation.
+        RETURNS: The updated losses dictionary.
```
svlandeg, line 83:
```suggestion
        RETURNS: The updated losses dictionary.
        
        DOCS: https://spacy.io/api/pipe#distill
```

spacy/ml/callbacks.py
```diff
@@ -23,6 +23,7 @@
     "update",
     "rehearse",
     "get_loss",
+    "get_student_teacher_loss",
```
svlandeg, line 26: While slightly less descriptive/accurate, I wonder whether we could just call this `get_distill_loss`?

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -155,6 +155,23 @@ def get_loss(
...
+    ) -> Tuple[float, List[Floats2d]]:
+        """Calculate the loss and its gradient for a batch of student
+        scores, relative to teacher scores.
+
+        teacher_scores: Scores representing the teacher model's predictions.
+        student_scores: Scores representing the student model's predictions.
+
+        DOCS: https://spacy.io/api/edittreelemmatizer#get_teacher_student_loss
+        """
+        loss_func = LegacySequenceCategoricalCrossentropy(normalize=False)
```
svlandeg, line 169: Is there no non-legacy loss function applicable? (same comment for other pipes)

spacy/pipeline/tagger.pyx
```diff
@@ -259,12 +259,30 @@ class Tagger(TrainablePipe):
...
+        teacher_scores: Scores representing the teacher model's predictions.
+        student_scores: Scores representing the student model's predictions.
+
+        DOCS: https://spacy.io/api/tagger#get_teacher_student_loss
+        """
+        loss_func = LegacySequenceCategoricalCrossentropy(normalize=False)
+        d_scores, loss = loss_func(student_scores, teacher_scores)
+        if self.model.ops.xp.isnan(loss):
+            raise ValueError(Errors.E910.format(name=self.name))
+        return float(loss), d_scores
```
svlandeg, line 284: Ugh, how did we end up with a return order of `grads, loss` that is reversed compared to the losss functions in Thinc 😭

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -56,6 +56,56 @@ cdef class TrainablePipe(Pipe):
         except Exception as e:
             error_handler(self.name, self, [doc], e)
 
+
+    def distill(self,
+               teacher_pipe: Optional["TrainablePipe"],
+               teacher_docs: Iterable["Doc"],
+               student_docs: Iterable["Doc"],
```
svlandeg, line 63: In a sense, the `teacher_docs` and `student_docs` do form a batch of `Example` objects, no? It's just that the "reference" in the example isn't exactly "gold data", it's "teacher predictions". Still, it might all be more straightforward if we pass in a batch of `Example` objects instead? It'd also make the distill loss function look much more like a "normal" loss function - if not in fact entirely the same?

spacy/pipeline/transition_parser.pyx
```diff
@@ -203,6 +207,120 @@ cdef class Parser(TrainablePipe):
...
+        teacher_docs (Iterable[Doc]): Documents passed through teacher pipes.
+        student_docs (Iterable[Doc]): Documents passed through student pipes.
+            Must contain the same tokens as `teacher_docs` but may have
+            different annotations.
+        drop (float): dropout rate.
+        sgd (Optional[Optimizer]): An optimizer. Will be created via
+            create_optimizer if not set.
+        losses (Optional[Dict[str, float]]): Optional record of loss during
+            distillation.
+        RETURNS: The updated losses dictionary.
```
svlandeg, line 233:
```suggestion
        RETURNS: The updated losses dictionary.
        
        DOCS: https://spacy.io/api/dependencyparser#distill
```

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -155,6 +155,23 @@ def get_loss(
...
+    def get_teacher_student_loss(
+        self, teacher_scores: List[Floats2d], student_scores: List[Floats2d]
+    ) -> Tuple[float, List[Floats2d]]:
+        """Calculate the loss and its gradient for a batch of student
+        scores, relative to teacher scores.
+
+        teacher_scores: Scores representing the teacher model's predictions.
+        student_scores: Scores representing the student model's predictions.
+
+        DOCS: https://spacy.io/api/edittreelemmatizer#get_teacher_student_loss
```
svlandeg, line 167:
```suggestion
        RETURNS (Tuple[float, float]): The loss and the gradient.
        
        DOCS: https://spacy.io/api/edittreelemmatizer#get_teacher_student_loss
```

spacy/pipeline/tagger.pyx
```diff
@@ -259,12 +259,30 @@ class Tagger(TrainablePipe):
...
+    def get_teacher_student_loss(
+        self, teacher_scores: List[Floats2d], student_scores: List[Floats2d]
+    ) -> Tuple[float, List[Floats2d]]:
+        """Calculate the loss and its gradient for a batch of student
+        scores, relative to teacher scores.
+
+        teacher_scores: Scores representing the teacher model's predictions.
+        student_scores: Scores representing the student model's predictions.
+
+        DOCS: https://spacy.io/api/tagger#get_teacher_student_loss
```
svlandeg, line 278:
```suggestion
        RETURNS (Tuple[float, float]): The loss and the gradient.
        
        DOCS: https://spacy.io/api/tagger#get_teacher_student_loss
```

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -169,6 +219,17 @@ cdef class TrainablePipe(Pipe):
...
         raise NotImplementedError(Errors.E931.format(parent="TrainablePipe", method="get_loss", name=self.name))
 
+    def get_teacher_student_loss(self, teacher_scores, student_scores):
+        """Calculate the loss and its gradient for a batch of student
+        scores, relative to teacher scores.
+
+        teacher_scores: Scores representing the teacher model's predictions.
+        student_scores: Scores representing the student model's predictions.
+
+        DOCS: https://spacy.io/api/pipe#get_teacher_student_loss
```
svlandeg, line 229:
```suggestion
        RETURNS (Tuple[float, float]): The loss and the gradient.
        
        DOCS: https://spacy.io/api/pipe#get_teacher_student_loss
```

spacy/pipeline/transition_parser.pyx
```diff
@@ -203,6 +207,120 @@ cdef class Parser(TrainablePipe):
...
+    def get_teacher_student_loss(
+        self, teacher_scores: List[Floats2d], student_scores: List[Floats2d]
+    ) -> Tuple[float, List[Floats2d]]:
+        """Calculate the loss and its gradient for a batch of student
+        scores, relative to teacher scores.
+
+        teacher_scores: Scores representing the teacher model's predictions.
+        student_scores: Scores representing the student model's predictions.
+
+        DOCS: https://spacy.io/api/dependencyparser#get_teacher_student_loss
```
svlandeg, line 316:
```suggestion
        RETURNS (Tuple[float, float]): The loss and the gradient.
        
        DOCS: https://spacy.io/api/dependencyparser#get_teacher_student_loss
```

svlandeg, line 316:
```suggestion
        RETURNS (Tuple[float, float]): The loss and the gradient.
        
        DOCS: https://spacy.io/api/dependencyparser#get_teacher_student_loss
```

svlandeg (commented):
This will be a really nice feature to have in v4 :-)

website/docs/api/dependencyparser.md
```diff
@@ -268,6 +268,27 @@ predicted scores.
 | `scores`    | Scores representing the model's predictions. ~~StateClass~~                 |
 | **RETURNS** | The loss and the gradient, i.e. `(loss, gradient)`. ~~Tuple[float, float]~~ |
 
+## DependencyParser.get_teacher_student_loss {#get_teacher_student_loss tag="method" new="4"}
```
svlandeg, line 271: For all these pipes, I think we should also repeat the `distill()` documentation.

spacy/pipeline/trainable_pipe.pyx
shadeMe: At the moment, the [`TransformerDistiller`](https://github.com/explosion/curated-transformers/blob/8e387e5ac4dcde22f1548792b0aada695e68f2f7/curated_transformers/pipeline/distill.py#L108) is the only downstream pipe that doesn't have a teacher.

spacy/ml/callbacks.py
shadeMe: IIRC, the reason we went for this was because we use the same loss in the model rehearsal process as well. So, it made sense not to use `distill` in the name.

spacy/pipeline/edit_tree_lemmatizer.py
danieldk: The non-legacy losses are not tested at all yet. So, I wanted to wait until @kadarakos gets around to implementing and testing them for regular training in spaCy.

spacy/pipeline/trainable_pipe.pyx
danieldk: Hmmm, it would be a bit weird though, since you'd expect `Example.reference` to have annotations, but we don't let the teacher annotate the `Doc` in the usual case (currently only for `tok2vec`). But does clean up things quite a bit.

spacy/pipeline/trainable_pipe.pyx
danieldk: I am a bit worried that downstream pipes forget to set this. Though, I think it would be enough to check in `is_distillable()` whether `distill` and `get_student_teacher_loss` are still the base class (`TrainablePipe`) methods? Then `is_distillable()` would return `True` if a pipe implements either of these methods. Let me implement it in this way and see if you like it...

danieldk: Added.

spacy/pipeline/trainable_pipe.pyx
danieldk: Checked in the new validation function.

website/docs/api/dependencyparser.md
danieldk: Done.

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -71,16 +70,17 @@ cdef class TrainablePipe(Pipe):
 
         teacher_pipe (Optional[TrainablePipe]): The teacher pipe to learn
             from.
-        teacher_docs (Iterable[Doc]): Documents passed through teacher pipes.
-        student_docs (Iterable[Doc]): Documents passed through student pipes.
-            Must contain the same tokens as `teacher_docs` but may have
-            different annotations.
+        examples (Iterable[Example]): Distillation examples. The eference
```
svlandeg, line 73:
```suggestion
        examples (Iterable[Example]): Distillation examples. The reference
```

spacy/pipeline/transition_parser.pyx
```diff
@@ -221,31 +221,31 @@ cdef class Parser(TrainablePipe):
 
         teacher_pipe (Optional[TrainablePipe]): The teacher pipe to learn
             from.
-        teacher_docs (Iterable[Doc]): Documents passed through teacher pipes.
-        student_docs (Iterable[Doc]): Documents passed through student pipes.
-            Must contain the same tokens as `teacher_docs` but may have
-            different annotations.
+        examples (Iterable[Example]): Distillation examples. The eference
```
svlandeg, line 224:
```suggestion
        examples (Iterable[Example]): Distillation examples. The reference
```

spacy/pipeline/trainable_pipe.pyx
```diff
@@ -266,6 +265,10 @@ cdef class TrainablePipe(Pipe):
         """
         raise NotImplementedError(Errors.E931.format(parent="Pipe", method="add_label", name=self.name))
 
+    @property
+    def is_distillable(self) -> bool:
+        return not (self.__class__.distill is TrainablePipe.distill and self.__class__.get_teacher_student_loss is TrainablePipe.get_teacher_student_loss)
+
```
svlandeg, line 271: I think this is a fine solution - maybe we can just add a comment here to explain what's going on, basically that we return `False` if there's no non-default implementation?

spacy/pipeline/transition_parser.pyx
```diff
@@ -313,6 +313,10 @@ cdef class Parser(TrainablePipe):
         teacher_scores: Scores representing the teacher model's predictions.
         student_scores: Scores representing the student model's predictions.
 
+        RETURNS (Tuple[float, float]): The loss and the gradient.
+        
+        RETURNS (Tuple[float, float]): The loss and the gradient.
```
svlandeg, line 318:
```suggestion
        RETURNS (Tuple[float, float]): The loss and the gradient.
```

svlandeg (commented):
Looking good! Few more small comments and then I think this will be good to merge!

spacy/pipeline/trainable_pipe.pyx
danieldk: Done!

svlandeg (approved):
Yay, looking great!

PR #12017, opened 2022-12-22
Fix speed problem with `top_k>1` on CPU in edit tree lemmatizer

richardpaulhudson:
## Description

The edit-tree lemmatizer has a hyperparameter `top_k` that specifies the number of alternative predicted trees to consider for each token: if the first predicted tree is not applicable to the raw token text, the second predicted tree is considered and so on up to the value of `top_k`.

Although accuracies are normally higher if alternative predictions can be considered, the current standard published models all specify `top_k = 1`. This is because the pre-existing code is several times slower when executed on a CPU for any higher value of `top_k` owing to an expensive NumPy sort that is used to order the predictions for each token (the sort is avoided in the pre-existing code if `top_k == 1`).

This PR retains the existing approach if `top_k==1`, but introduces a procedural approach if `top_k>1`. The procedural approach avoids the NumPy sort and also has the important advantage that time is only spent processing alternative predictions if earlier predictions were not applicable. Because with a well-trained model the first predictions are typically applicable to most tokens, this largely removes the performance hit where `top_k > 1` and should allow us to choose values for standard models in the future based purely on accuracy requirements.

When using the edit-tree lemmatizer for its original intended purpose, it is unlikely that values of `top_k` above about 10 would ever be useful. However, it is conceivable that the code might be used for some other purpose where much higher values of `top_k` are required. For values of `top_k` above about 20, the pre-existing approach is more efficient than the new approach. To take the various scenarios into account, the new code checks the value of `top_k` and selects one of three strategies depending on whether it is `1`; `2<n<=20`; or `>20`. 

The following speed figures were measured training `pl_core_news_lg`. The relevant improvement is for `2<n<=20` and especially on CPU, although selecting the optimal strategy once for each batch of documents rather than testing to see whether `top_k` is greater than 1 or not each time an individual document is processed also has a small but consistent positive impact on speed for other scenarios:

| `top_k` | Speed CPU old | Speed CPU new | Speed GPU old | Speed GPU new |
| --------   | --------------------- | --------------------   | --------------------- | ---------------------- |
| 1 | 37040 | 37483 | 184859 | 192005 |
| 5 | **9585** | **36050** | 113997 | 123473 |
| 25 | 9271 | 9569 | 111335 | 112726 |

### Types of change
Speed enhancement

## Checklist

### DIFF
--- a/spacy/pipeline/edit_tree_lemmatizer.py
+++ b/spacy/pipeline/edit_tree_lemmatizer.py
@@ -20,6 +20,10 @@ from ..vocab import Vocab
 from .. import util
 
 
+# The cutoff value of *top_k* above which an alternative method is used to process guesses.
+TOP_K_GUARDRAIL = 20
+
+
 default_model_config = """
 [model]
 @architectures = "spacy.Tagger.v2"
@@ -145,6 +149,18 @@ class EditTreeLemmatizer(TrainablePipe):
         return float(loss), d_scores
 
     def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
+        if self.top_k == 1:
+            scores2guesses = self._scores2guesses_top_k_equals_1
+        elif self.top_k <= TOP_K_GUARDRAIL:
+            scores2guesses = self._scores2guesses_top_k_greater_1
+        else:
+            scores2guesses = self._scores2guesses_top_k_guardrail
+        # The behaviour of *_scores2guesses_top_k_greater_1()* is efficient for values
+        # of *top_k>1* that are likely to be useful when the edit tree lemmatizer is used
+        # for its principal purpose of lemmatizing tokens. However, the code could also
+        # be used for other purposes, and with very large values of *top_k* the method
+        # becomes inefficient. In such cases, *_scores2guesses_top_k_guardrail()* is used
+        # instead.
         n_docs = len(list(docs))
         if not any(len(doc) for doc in docs):
             # Handle cases where there are no tokens in any docs.
@@ -154,11 +170,28 @@ class EditTreeLemmatizer(TrainablePipe):
             return guesses
         scores = self.model.predict(docs)
         assert len(scores) == n_docs
-        guesses = self._scores2guesses(docs, scores)
+        guesses = scores2guesses(docs, scores)
         assert len(guesses) == n_docs
         return guesses
 
-    def _scores2guesses(self, docs, scores):
+    def _scores2guesses_top_k_equals_1(self, docs, scores):
+        guesses = []
+        for doc, doc_scores in zip(docs, scores):
+            doc_guesses = doc_scores.argmax(axis=1)
+            doc_guesses = self.numpy_ops.asarray(doc_guesses)
+
+            doc_compat_guesses = []
+            for i, token in enumerate(doc):
+                tree_id = self.cfg["labels"][doc_guesses[i]]
+                if self.trees.apply(tree_id, token.text) is not None:
+                    doc_compat_guesses.append(tree_id)
+                else:
+                    doc_compat_guesses.append(-1)
+            guesses.append(np.array(doc_compat_guesses))
+
+        return guesses
+
+    def _scores2guesses_top_k_greater_1(self, docs, scores):
         guesses = []
         predictions_to_consider = min(self.top_k, len(self.labels))
         for doc, doc_scores in zip(docs, scores):
@@ -175,6 +208,27 @@ class EditTreeLemmatizer(TrainablePipe):
                 else:
                     doc_compat_guesses.append(-1)
             guesses.append(np.array(doc_compat_guesses))
+        return guesses
+
+    def _scores2guesses_top_k_guardrail(self, docs, scores):
+        guesses = []
+        for doc, doc_scores in zip(docs, scores):
+            doc_guesses = np.argsort(doc_scores)[..., : -self.top_k - 1 : -1]
+            doc_guesses = self.numpy_ops.asarray(doc_guesses)
+
+            doc_compat_guesses = []
+            for token, candidates in zip(doc, doc_guesses):
+                tree_id = -1
+                for candidate in candidates:
+                    candidate_tree_id = self.cfg["labels"][candidate]
+
+                    if self.trees.apply(candidate_tree_id, token.text) is not None:
+                        tree_id = candidate_tree_id
+                        break
+                doc_compat_guesses.append(tree_id)
+
+            guesses.append(np.array(doc_compat_guesses))
+
         return guesses
 
     def set_annotations(self, docs: Iterable[Doc], batch_tree_ids):
--- a/spacy/pipeline/edit_tree_lemmatizer.py
+++ b/spacy/pipeline/edit_tree_lemmatizer.py
@@ -165,7 +165,7 @@ class EditTreeLemmatizer(TrainablePipe):
             doc_compat_guesses = []
             for i, token in enumerate(doc):
                 for _ in range(self.top_k):
-                    candidate = doc_scores[i].argmax()
+                    candidate = int(doc_scores[i].argmax())
                     candidate_tree_id = self.cfg["labels"][candidate]
                     if self.trees.apply(candidate_tree_id, token.text) is not None:
                         doc_compat_guesses.append(candidate_tree_id)
--- a/spacy/pipeline/edit_tree_lemmatizer.py
+++ b/spacy/pipeline/edit_tree_lemmatizer.py
@@ -193,7 +193,7 @@ class EditTreeLemmatizer(TrainablePipe):
 
     def _scores2guesses_top_k_greater_1(self, docs, scores):
         guesses = []
-        predictions_to_consider = min(self.top_k, len(self.labels))
+        top_k = min(self.top_k, len(self.labels))
         for doc, doc_scores in zip(docs, scores):
             doc_scores = self.numpy_ops.asarray(doc_scores)
             doc_compat_guesses = []
--- a/spacy/pipeline/edit_tree_lemmatizer.py
+++ b/spacy/pipeline/edit_tree_lemmatizer.py
@@ -198,7 +198,7 @@ class EditTreeLemmatizer(TrainablePipe):
             doc_scores = self.numpy_ops.asarray(doc_scores)
             doc_compat_guesses = []
             for i, token in enumerate(doc):
-                for _ in range(predictions_to_consider):
+                for _ in range(top_k):
                     candidate = int(doc_scores[i].argmax())
                     candidate_tree_id = self.cfg["labels"][candidate]
                     if self.trees.apply(candidate_tree_id, token.text) is not None:
--- a/spacy/pipeline/edit_tree_lemmatizer.py
+++ b/spacy/pipeline/edit_tree_lemmatizer.py
@@ -160,27 +160,18 @@ class EditTreeLemmatizer(TrainablePipe):
     def _scores2guesses(self, docs, scores):
         guesses = []
         for doc, doc_scores in zip(docs, scores):
-            if self.top_k == 1:
-                doc_guesses = doc_scores.argmax(axis=1).reshape(-1, 1)
-            else:
-                doc_guesses = np.argsort(doc_scores)[..., : -self.top_k - 1 : -1]
-
-            if not isinstance(doc_guesses, np.ndarray):
-                doc_guesses = doc_guesses.get()
-
             doc_compat_guesses = []
-            for token, candidates in zip(doc, doc_guesses):
-                tree_id = -1
-                for candidate in candidates:
+            for i, token in enumerate(doc):
+                for _ in range(self.top_k):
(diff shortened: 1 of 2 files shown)


spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -160,27 +160,18 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
...
-
-            if not isinstance(doc_guesses, np.ndarray):
-                doc_guesses = doc_guesses.get()
-
             doc_compat_guesses = []
-            for token, candidates in zip(doc, doc_guesses):
-                tree_id = -1
-                for candidate in candidates:
+            for i, token in enumerate(doc):
+                for _ in range(self.top_k):
```
richardpaulhudson, line 165: One obvious question here is what happens if the value of `top_k` is greater than the number of trees in the model. It doesn't seem worth testing for this explicitly:

- with genuine models and genuinely useful values of `top_k`, it's not going to happen (a typical models has more than a thousand trees, a high useful value of `top_k` might be 10 or 20).
- Any real model will contain a tree that does not apply any change to the original form, which by definition can be applied to any token, so that trees will never be looped through a second time.
- In the exceedingly unlikely event that for some reason all the trees in a model were not applicable to a given token form, nothing would break: the first tree in the NumPy array would just be tried repeatedly until the value of `top_k` were reached.
- Testing for this explicitly would however have a real-world speed impact.

richardpaulhudson:

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -160,27 +160,20 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
...
-            if self.top_k == 1:
-                doc_guesses = doc_scores.argmax(axis=1).reshape(-1, 1)
-            else:
-                doc_guesses = np.argsort(doc_scores)[..., : -self.top_k - 1 : -1]
-
-            if not isinstance(doc_guesses, np.ndarray):
-                doc_guesses = doc_guesses.get()
-
+            if not isinstance(doc_scores, np.ndarray):
+                doc_scores = doc_scores.get()
```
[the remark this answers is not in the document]
richardpaulhudson: To be honest I can't say much about either question — why we are pinning to the CPU or why we are doing it this way. I just copied the existing code for `doc_guesses` (for which there might conceivably be different factors at play as it's a smaller array). @danieldk can you shed any light on these questions?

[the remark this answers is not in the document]
richardpaulhudson: What I also still have to investigate is whether or not performing the `argmax` in a loop causes a measurable speed penalty when using a GPU. If so, I'll change the method back to use separate strategies for `argmax = 1` and `argmax > 1`.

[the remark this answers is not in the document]
danieldk: We are iterating over the scores here. Doing that on the GPU kills performance.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -160,27 +160,20 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
...
+                    candidate = int(doc_scores[i].argmax())
                     candidate_tree_id = self.cfg["labels"][candidate]
-
                     if self.trees.apply(candidate_tree_id, token.text) is not None:
-                        tree_id = candidate_tree_id
+                        doc_compat_guesses.append(candidate_tree_id)
                         break
-                doc_compat_guesses.append(tree_id)
-
+                    doc_scores[i, candidate] = -1
```
danieldk, line 173: This should definitely be the largest possible negative number. If we are using unnormalized softmax, the scores are logits and `-1` may be a relatively high value.

spacy/pipeline/edit_tree_lemmatizer.py
danieldk: I don't follow. The number of labels is `len(self.labels)`, so this would be a single `range(min(self.top_k, len(self.labels))`. Maybe I am missing something?

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -160,27 +160,20 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
...
-
+            if not isinstance(doc_scores, np.ndarray):
+                doc_scores = doc_scores.get()
             doc_compat_guesses = []
-            for token, candidates in zip(doc, doc_guesses):
-                tree_id = -1
-                for candidate in candidates:
+            for i, token in enumerate(doc):
+                for _ in range(self.top_k):
+                    candidate = int(doc_scores[i].argmax())
```
danieldk, line 168: I think some variation of quickselect has better complexity. This approach is O(n*k), while with quickselect you can do it with O(n) + O(k^2) or O(n) + O(k log k). Not sure how it compares in practice with all the memory movement. It's mostly an empirical question.

danieldk:
> As well as being slightly faster than the pre-existing code in its own right (i.e. where top_k = 1)

I don't completely understand this. For `top_k=1`, the new and old code do the same? (A single argmax)

Getting rid of the sort makes sense for small values of k. Though, it does lead to the degenerate O(n^2) case when someone says 'lets try all edit trees'. So, I am very interested in the empirical results. Maybe it's worth in the end using this approach for small k's and sort for large k's.

spacy/pipeline/edit_tree_lemmatizer.py
richardpaulhudson: OK, no, you're right — not really a real-world speed impact, but still extra processing: do you think this is worth doing given how unlikely it is to ever happen and the fact that nothing would break if it did?

spacy/pipeline/edit_tree_lemmatizer.py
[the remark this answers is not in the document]
richardpaulhudson: > Though, it does lead to the degenerate O(n^2) case when someone says 'lets try all edit trees'. 

An important point is that even if you say "let's try all edit trees" it only iterates through them until it comes across a tree that can be applied to the raw token text. And in any normal model a good proportion of trees are e.g. 'add a `u` to the end' which can be applied to any text. This will be shown in the results I'm collecting: increasing the value of `top_k` beyond about 5 has no significant effect, and increasing it beyond about 10 has no effect, because it's so unlikely that there is no applicable tree in the first handful tried.

richardpaulhudson:
> Though, it does lead to the degenerate O(n^2) case when someone says 'lets try all edit trees'.

An important point is that even if you say "let's try all edit trees" it only iterates through them until it comes across a tree that can be applied to the raw token text. And in any normal model a good proportion of trees are e.g. 'add a u to the end' which can be applied to any text. This will be shown in the results I'm collecting: increasing the value of top_k beyond about 5 has no significant effect, and increasing it beyond about 10 has no effect, because it's so unlikely that there is no applicable tree in the first handful tried.

spacy/pipeline/edit_tree_lemmatizer.py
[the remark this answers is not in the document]
richardpaulhudson: Having collected the first results on GPU (I'd been focussed on CPU until now) I can confirm that:

- the speed problem that this PR fixes relates exclusively to CPU and is not relevant to GPU;
- the performance on GPU with this PR is either equivalent or slightly worse than the performance with the pre-existing code. Depending on what the numbers I'm collecting show, it might make sense to retain the pre-existing code for GPU and use the new code on CPU;
- What is definitely the case: if you don't pin `doc_scores` to the CPU when working on the GPU, inference is thirteen times slower than if you work on the GPU and do copy it across and three times slower than if you work exclusively on the CPU. So pinning is definitely necessary if we end up using the new code for GPU.

spacy/pipeline/edit_tree_lemmatizer.py
danieldk: Yes, it’s the semantics that it should convey and if it is cheap to be correct, why not?

spacy/pipeline/edit_tree_lemmatizer.py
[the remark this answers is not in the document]
danieldk: > What is definitely the case: if you don't pin doc_scores to the CPU when working on the GPU, inference is thirteen times slower than if you work on the GPU and do copy it across and three times slower than if you work exclusively on the CPU. So pinning is definitely necessary if we end up using the new code for GPU.

Yeah, my first week of GPU performance improvements in spaCy/Thinc was finding loops that iterate over GPU arrays by element. These 4-byte transfers are insanely expensive.

danieldk:
> > Though, it does lead to the degenerate O(n^2) case when someone says 'lets try all edit trees'.
> 
> An important point is that even if you say "let's try all edit trees" it only iterates through them until it comes across a tree that can be applied to the raw token text. And in any normal model a good proportion of trees are e.g. 'do nothing' or 'add an `a` to the end' which can be applied to any text.

I think it’s still good to have bounds on the complexity of it is easy to do so. These things don’t happen until they do. (E.g. we had a degenerate case in parser feature extraction that wasn’t noticed until a user used it in a way that triggered quadratic complexity).

It’s only a simple if statement: use sort when k is higher than a certain value. Two lines of code for avoiding quadratic complexity seems like an easily trade-off.

richardpaulhudson:
Here are the figures for `pl_core_news_lg`testing the "new" procedural solution against the "old" pre-existing code:

| `top_k` | Accuracy | Speed CPU old | Speed CPU new | Speed GPU old | Speed GPU new |
| --------   | -----------  | --------------------- | --------------------   | --------------------- | ---------------------- |
| 1 | 94.80 | 36939 | 36249 | 186701 | 122782 |
| 2 | 95.01 | 9431 | 36435 | 114735 | 124300 |
| 3 | 95.08 | 9623 | 36285 | 115295 | 121976 |
| 5 | 95.11 | 9657 | 36085 | 113974 | 124915 |
| 10 | 95.12 | 9635 | 36095 | 113873 | 121860 |
| 20 | 95.12 | 9605 | 36160 | 110383 | 123045 |

To check the patterns were reproducible for a language with very different morphology, I performed a couple of the experiments with `de_core_news_lg`:

| `top_k` | Accuracy | Speed CPU old | Speed CPU new | Speed GPU old | Speed GPU new |
| --------   | -----------  | --------------------- | --------------------   | --------------------- | ---------------------- |
| 1 | 98.14 | 43174 | 41707 | 198944 | 146672 |
| 20 | 98.17 | 16896 | 41985 | 136063 | 144506 |

This shows it makes sense to retain the pre-existing solution for `top_k = 1` and the new code for `1 < top_k <= 20`. As we discussed in personal communication, we should also retain the second pre-existing solution for `top_k > 20`: although with the normal use of the edit-tree lemmatizer there will always be a tree in the first handful that is applicable to the form, meaning that the procedural solution will never iterate through more than a few predictions, it is possible that the edit-tree lemmatizer could be used to predict something other than `lemma_` where this is not the case, and the performance for high values of `top_k` would then degrade heavily using the new code.

danieldk:
Since `top_k=1` will probably be the default for some languages, I think it would be nice if the GPU regression for inference with that value would be addressed.

danieldk:
I am a bit confused, because the tables have different numbers?

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -153,20 +170,52 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
...
+                    doc_compat_guesses.append(tree_id)
+                else:
+                    doc_compat_guesses.append(-1)
+            guesses.append(np.array(doc_compat_guesses))
+
+        return guesses
+
+    def _scores2guesses_top_k_greater_1(self, docs, scores):
         guesses = []
+        predictions_to_consider = min(self.top_k, len(self.labels))
```
danieldk, line 196:
```suggestion
        top_k = min(self.top_k, len(self.labels))
```

Shorter and expresses better below what we are doing.

spacy/pipeline/edit_tree_lemmatizer.py
```diff
@@ -153,20 +170,52 @@ def predict(self, docs: Iterable[Doc]) -> List[Ints2d]:
...
         for doc, doc_scores in zip(docs, scores):
-            if self.top_k == 1:
-                doc_guesses = doc_scores.argmax(axis=1).reshape(-1, 1)
-            else:
-                doc_guesses = np.argsort(doc_scores)[..., : -self.top_k - 1 : -1]
+            doc_scores = self.numpy_ops.asarray(doc_scores)
+            doc_compat_guesses = []
+            for i, token in enumerate(doc):
+                for _ in range(predictions_to_consider):
```
danieldk, line 201:
```suggestion
                for _ in range(top_k):
```

danieldk (commented):
Looks good to me. I also did some experiments this morning, and works very nicely.

One small naming nitpick.

**Note:** merge after 3.5.0 is tagged.

danieldk (approved):

PR #12020, opened 2022-12-23
Add commands for automatically modifying configs

polm:
## Description

This continues work started in
https://github.com/explosion/projects/pull/147,
which provides features for automatically manipulating pipelines and configs. The functions included are:

- merge: combine components from two pipelines and handle listeners
- use_transformer: use transformer as feature source
- use_tok2vec: use CNN tok2vec as feature source
- resume: make a version of a config for resuming training

Currently these are all grouped under a new `spacy configure` command. That may not be the best place for them; in particular, `merge` may belong elsewhere, since it outputs a pipeline rather than a config.

The current state of the PR is that the commands run, but there's only one small test, and docs haven't been written yet. Docs can be started but will depend somewhat on how the naming issues work out.

### Types of change

enhancement

## Checklist

### DIFF
--- a/spacy/cli/configure.py
+++ b/spacy/cli/configure.py
@@ -1,0 +1,330 @@
+from pathlib import Path
+import re
+from wasabi import msg
+import typer
+
+import spacy
+from spacy.language import Language
+
+from ._util import configure_cli, Arg, Opt
+
+# These are the architectures that are recognized as tok2vec/feature sources.
+TOK2VEC_ARCHS = [("spacy", "Tok2Vec"), ("spacy-transformers", "TransformerModel")]
+# These are the listeners.
+LISTENER_ARCHS = [
+    ("spacy", "Tok2VecListener"),
+    ("spacy-transformers", "TransformerListener"),
+]
+
+
+def _deep_get(obj, key, default):
+    """Given a multi-part key, try to get the key. If at any point this isn't
+    possible, return the default.
+    """
+    out = None
+    slot = obj
+    for notch in key:
+        if slot is None or notch not in slot:
+            return default
+        slot = slot[notch]
+    return slot
+
+
+def _get_tok2vecs(config):
+    """Given a pipeline config, return the names of components that are
+    tok2vecs (or Transformers).
+    """
+    out = []
+    for name, comp in config["components"].items():
+        arch = _deep_get(comp, ("model", "@architectures"), False)
+        if not arch:
+            continue
+
+        ns, model, ver = arch.split(".")
+        if (ns, model) in TOK2VEC_ARCHS:
+            out.append(name)
+    return out
+
+
+def _has_listener(nlp, pipe_name):
+    """Given a pipeline and a component name, check if it has a listener."""
+    arch = _deep_get(
+        nlp.config,
+        ("components", pipe_name, "model", "tok2vec", "@architectures"),
+        False,
+    )
+    if not arch:
+        return False
+    ns, model, ver = arch.split(".")
+    return (ns, model) in LISTENER_ARCHS
+
+
+def _get_listeners(nlp):
+    """Get the name of every component that contains a listener.
+
+    Does not check that they listen to the same thing; assumes a pipeline has
+    only one feature source.
+    """
+    out = []
+    for name in nlp.pipe_names:
+        if _has_listener(nlp, name):
+            out.append(name)
+    return out
+
+
+def _increment_suffix(name):
+    """Given a name, return an incremented version.
+
+    If no numeric suffix is found, return the original with "2" appended.
+
+    This is used to avoid name collisions in pipelines.
+    """
+
+    res = re.search(r"\d+$", name)
+    if res is None:
+        return f"{name}2"
+    else:
+        num = res.match
+        prefix = name[0 : -len(num)]
+        return f"{prefix}{int(num) + 1}"
+
+
+def _check_single_tok2vec(name, config):
+    """Check if there is just one tok2vec in a config.
+
+    A very simple check, but used in multiple functions.
+    """
+    tok2vecs = _get_tok2vecs(config)
+    fail_msg = f"""
+        Can't handle pipelines with more than one feature source, 
+        but {name} has {len(tok2vecs)}."""
+    if len(tok2vecs) > 1:
+        msg.fail(fail_msg, exits=1)
+
+
+def _check_pipeline_names(nlp, nlp2):
+    """Given two pipelines, try to rename any collisions in component names.
+
+    If a simple increment of a numeric suffix doesn't work, will give up.
+    """
+
+    fail_msg = """
+        Tried automatically renaming {name} to {new_name}, but still
+        had a collision, so bailing out. Please make your pipe names
+        more unique.
+        """
+
+    # map of components to be renamed
+    rename = {}
+    # check pipeline names
+    names = nlp.pipe_names
+    for name in nlp2.pipe_names:
+        if name in names:
+            inc = _increment_suffix(name)
+            # TODO Would it be better to just keep incrementing?
+            if inc in names or inc in nlp2.pipe_names:
+                msg.fail(fail_msg.format(name=name, new_name=inc), exits=1)
+            rename[name] = inc
+    return rename
+
+
+@configure_cli.command("resume")
+def configure_resume_cli(
+    # fmt: off
+    base_model: Path = Arg(..., help="Path or name of base model to use for config"),
+    output_path: Path = Arg(..., help="File to save the config to or - for stdout (will only output config and no additional logging info)", allow_dash=True),
+    # fmt: on
+):
+    """Create a config for resuming training.
+
+    A config for resuming training is the same as the input config, but with
+    all components sourced.
+    """
+
+    nlp = spacy.load(base_model)
+    conf = nlp.config
+
+    # Paths are not JSON serializable
+    path_str = str(base_model)
+
+    for comp in nlp.pipe_names:
+        conf["components"][comp] = {"source": path_str}
+
+    if str(output_path) == "-":
+        print(conf.to_str())
+    else:
+        conf.to_disk(output_path)
+        msg.good("Saved config", output_path)
+
+    return conf
+
+
+@configure_cli.command("transformer")
+def use_transformer(
+    base_model: str, output_path: Path, transformer_name: str = "roberta-base"
+):
+    """Replace pipeline tok2vec with transformer."""
+
+    # 1. identify tok2vec
+    # 2. replace tok2vec
+    # 3. replace listeners
+    nlp = spacy.load(base_model)
+    _check_single_tok2vec(base_model, nlp.config)
+
+    tok2vecs = _get_tok2vecs(nlp.config)
+    assert len(tok2vecs) > 0, "Must have tok2vec to replace!"
+
+    nlp.remove_pipe(tok2vecs[0])
+    # the rest can be default values
+    trf_config = {
+        "model": {
+            "name": transformer_name,
+        }
+    }
+    trf = nlp.add_pipe("transformer", config=trf_config, first=True)
+
+    # TODO maybe remove vectors?
+
+    # now update the listeners
+    listeners = _get_listeners(nlp)
+    for listener in listeners:
+        listener_config = {
+            "@architectures": "spacy-transformers.TransformerListener.v1",
+            "grad_factor": 1.0,
+            "upstream": "transformer",
+            "pooling": {"@layers": "reduce_mean.v1"},
+        }
+        nlp.config["components"][listener]["model"]["tok2vec"] = listener_config
+
+    if str(output_path) == "-":
+        print(nlp.config.to_str())
+    else:
+        nlp.config.to_disk(output_path)
+        msg.good("Saved config", output_path)
+
+    return nlp.config
+
+
+@configure_cli.command("tok2vec")
+def use_tok2vec(base_model: str, output_path: Path) -> Language:
+    """Replace pipeline tok2vec with CNN tok2vec."""
+    nlp = spacy.load(base_model)
+    _check_single_tok2vec(base_model, nlp.config)
+
+    tok2vecs = _get_tok2vecs(nlp.config)
+    assert len(tok2vecs) > 0, "Must have tok2vec to replace!"
+
+    nlp.remove_pipe(tok2vecs[0])
+
+    tok2vec = nlp.add_pipe("tok2vec", first=True)
+    width = "${components.tok2vec.model.encode:width}"
+
+    listeners = _get_listeners(nlp)
+    for listener in listeners:
+        listener_config = {
+            "@architectures": "spacy.Tok2VecListener.v1",
+            "width": width,
+            "upstream": "tok2vec",
+        }
+        nlp.config["components"][listener]["model"]["tok2vec"] = listener_config
+
+    if str(output_path) == "-":
+        print(nlp.config.to_str())
+    else:
+        nlp.config.to_disk(output_path)
+        msg.good("Saved config", output_path)
+
+    return nlp.config
+
+
+def _inner_merge(nlp, nlp2, replace_listeners=False) -> Language:
+    """Actually do the merge.
+
+    nlp: Base pipeline to add components to.
+    nlp2: Pipeline to add components from.
+    replace_listeners (bool): Whether to replace listeners. Usually only true
+      if there's one listener.
+    returns: assembled pipeline.
+    """
+
+    # we checked earlier, so there's definitely just one
+    tok2vec_name = _get_tok2vecs(nlp2.config)[0]
+    rename = _check_pipeline_names(nlp, nlp2)
+
+    if len(_get_listeners(nlp2)) > 1:
+        if replace_listeners:
+            msg.warn(
+                """
+                Replacing listeners for multiple components. Note this can make
+                your pipeline large and slow. Consider chaining pipelines (like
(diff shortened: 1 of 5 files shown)


polm:
This might require more adjustment - for example, maybe merge should be split out into a separate command - but it is substantially complete and ready for review. Any feedback on how to make the design clearer would be welcome.

spacy/cli/configure.py
```diff
@@ -0,0 +1,337 @@
...
+    ("spacy-transformers", "TransformerModel"),
+]
+# These are the listeners.
+LISTENER_ARCHS = [
+    ("spacy", "Tok2VecListener"),
+    ("spacy-transformers", "TransformerListener"),
+]
+
+
+def _deep_get(obj, key, default):
```
rmitsch, line 25: Could you type the args and return values? Would be easier for me to understand and review.

spacy/cli/configure.py
```diff
@@ -0,0 +1,337 @@
...
+        if replace_listeners:
+            msg.warn(
+                """
+                Replacing listeners for multiple components. Note this can make
+                your pipeline large and slow. Consider chaining pipelines (like
+                nlp2(nlp(text))) instead.
+                """
+            )
+        else:
+            # TODO provide a guide for what to do here
```
rmitsch, line 273: Are you planning to expand this?

spacy/cli/configure.py
```diff
@@ -0,0 +1,337 @@
...
+    if nlp.config["nlp"]["tokenizer"] != nlp2.config["nlp"]["tokenizer"]:
+        msg.fail("Can't merge - tokenizers don't match", exits=1)
+
+    # Check that each pipeline only has one feature source
+    _check_single_tok2vec(base_model, nlp.config)
+    _check_single_tok2vec(added_model, nlp2.config)
+
+    # Check how many listeners there are and replace based on that
+    # TODO: option to recognize frozen tok2vecs
+    # TODO: take list of pipe names to copy, ignore others
```
rmitsch, line 328: Are you planning to do this in this PR?

website/docs/api/cli.mdx
```diff
@@ -249,6 +250,94 @@ $ python -m spacy init labels [config_path] [output_path] [--code] [--verbose] [
 | overrides         | Config parameters to override. Should be options starting with `--` that correspond to the config section and value to override, e.g. `--paths.train ./train.spacy`. ~~Any (option/flag)~~                         |
 | **CREATES**       | The label files.                                                                                                                                                                                                   |
 
+## configure {id="configure", version="TODO"}
+
+Modify or combine existing configs in high-level ways. Can be used to automate
+config changes made as part of the development cycle.
```
rmitsch, line 256: Could you elaborate on this? "high-level ways" and "changes made as part of the development cycle" doesn't seem very informative to me.

rmitsch (commented):
Thanks for this, I think this functionality can be super handy! I just did a superficial pass now and will continue reviewing at a later point.

spacy/cli/configure.py
```diff
@@ -0,0 +1,337 @@
...
+TOK2VEC_ARCHS = [
+    ("spacy", "Tok2Vec"),
+    ("spacy", "HashEmbedCNN"),
+    ("spacy-transformers", "TransformerModel"),
+]
+# These are the listeners.
+LISTENER_ARCHS = [
+    ("spacy", "Tok2VecListener"),
+    ("spacy-transformers", "TransformerListener"),
+]
```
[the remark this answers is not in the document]
polm: Not that I'm aware of - I don't think there's any metadata to mark tok2vec or listener architectures. For listeners we could match on the architecture name, though that seems quite brittle.

spacy/cli/configure.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+@configure_cli.command("tok2vec")
+def use_tok2vec(base_model: str, output_file: Path) -> Config:
+    """Replace pipeline tok2vec with CNN tok2vec."""
+    nlp = spacy.load(base_model)
+    _check_single_tok2vec(base_model, nlp.config)
+
+    tok2vecs = _get_tok2vecs(nlp.config)
+    assert len(tok2vecs) > 0, "Must have tok2vec to replace!"
```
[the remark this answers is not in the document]
polm: Not at this assert. If there is more than one tok2vec, then `_check_single_tok2vec` would fail.

spacy/cli/configure.py
polm: No, these are TODOs for "someday", or extra features that would be nice to have.

spacy/cli/configure.py
polm: Not in this PR, this is a "someday" TODO. I'm honestly not sure exactly what the right way to manage multiple non-embedded tok2vecs would be. I expect that the vast majority of users are fine with this.

website/docs/api/cli.mdx
polm: It'd be great to have any suggestions here, I had trouble summarizing the functionality more specifically since it really is a variety of high-level config modifications. 

The reference to the development cycle is about starting with small models to iron out data issues and switching to larger models (like transformers) only later. Similarly, it's common for a user to train a single specific model, like a textcat, and then want to combine it with one of the pretrained pipelines for components that they want but don't need to re-train, like a lemmatizer.

spacy/cli/configure.py
rmitsch: Hm, how do we handle this usually? I'm not a fan of leaving `# TODO` notes in the code. I'd rather pull them into a dedicated issue.

rmitsch:
> Currently these are all grouped under a new spacy configure command. That may not be the best place for them; in particular, merge may belong elsewhere, since it outputs a pipeline rather than a config.

I agree - having `merge` under `spacy configure` would be confusing. I'd favor it as a separate top-level command.

spacy/cli/configure.py
polm: There are over 50 TODOs in the spaCy source and using them for "notes about potential future improvements" is described as normal use in [the code conventions](https://github.com/explosion/spaCy/blob/master/extra/DEVELOPER_DOCS/Code%20Conventions.md#including-todos). 

I think these ideas need to be more concrete before being made into issues, but if you prefer I could create some.

polm:
Thanks for the feedback, I moved `merge` to a separate top-level command.

website/docs/api/cli.mdx
```diff
@@ -249,6 +251,93 @@ $ python -m spacy init labels [config_path] [output_path] [--code] [--verbose] [
...
+| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `base_model`       | A trained pipeline to resume training (package name or path). ~~str (positional)~~                                                                                                                                                    |
+| `output_file`      | Path to output `.cfg` file or `-` to write the config to stdout (so you can pipe it forward to a file or to the `train` command). Note that if you're writing to stdout, no additional logging info is printed. ~~Path (positional)~~ |
+| `transformer_name` | The name of the base HuggingFace model to use. Defaults to `roberta-base`. ~~str (option)~~                                                                                                                                           |
+
+### configure tok2vec {id="configure-tok2vec", tag="command"}
+
+Modify the base model config to use a CNN tok2vec component. Useful for
```
rmitsch, line 297: Not sure about how we usually format spaCy component names.
```suggestion
Modify the base model config to use a CNN `tok2vec` component. Useful for
```

website/docs/api/cli.mdx
```diff
@@ -249,6 +251,93 @@ $ python -m spacy init labels [config_path] [output_path] [--code] [--verbose] [
...
+```cli
+$ python -m spacy configure tok2vec [base_model] [output_file]
+```
+
+| Name          | Description                                                                                                                                                                                                                           |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `base_model`  | A trained pipeline to resume training (package name or path). ~~str (positional)~~                                                                                                                                                    |
+| `output_file` | Path to output `.cfg` file or `-` to write the config to stdout (so you can pipe it forward to a file or to the `train` command). Note that if you're writing to stdout, no additional logging info is printed. ~~Path (positional)~~ |
+
+## merge {id="merge", tag="command"}
```
rmitsch, line 310: High-level question: what's the advantage of merging over chaining your pipelines?

website/docs/api/cli.mdx
```diff
@@ -249,6 +251,93 @@ $ python -m spacy init labels [config_path] [output_path] [--code] [--verbose] [
...
+
+For components with listeners, the resulting pipeline structure depends on the
+number of listeners. If the second pipeline has only one listener, then
+[`replace_listeners`](https://spacy.io/api/language/#replace_listeners) will be
+used. If there is more than one listener, `replace_listeners` will not be used.
+In the multi-listener case, the resulting pipeline may require more adjustment
+for training to work.
+
+This is useful if you have trained a specialized component, such as NER or
+textcat, and want to provide with one of the official pretrained pipelines or
```
rmitsch, line 328: Same here, dunno about formatting.
```suggestion
`textcat`, and want to provide with one of the official pretrained pipelines or
```

rmitsch (commented):
Apart from the command naming and some smaller stuff this looks good, but I'd definitely get feedback from a third party before moving forward.

spacy/cli/configure.py
rmitsch: Nono, that's fine - thanks for pointing this out! Should have checked myself.

website/docs/api/cli.mdx
polm: Good question! I assume by "chaining" you mean `nlp2(nlp1(text))`? 

The first thing a user notices is convenience - a single pipeline can be packaged as a single file. Things like `nlp.pipe` will also work correctly, while they would be weird with chained pipelines. These are pretty big advantages. 

Another, less obvious advantage but also significant advantage is shared vocabs, which should reduce memory usage.

It's out of scope for this PR, but if duplicated and frozen components can be recognized, duplicates there can be eliminated, further saving on memory usage and processing time.

website/docs/api/cli.mdx
polm: It always confuse me too, and in practice I think we're a bit inconsistent. I would generally use unmarked text (like here) when referring to the task/component in general, rather than to a specific instance of it in code. Looking at existing docs, "tok2vec" in particular seems to be used in unmarked fashion, like [here](https://github.com/explosion/spaCy/blob/03a0c2badcc3e0355d52110fa0c526e8bbb52324/website/docs/api/architectures.mdx?plain=1#L48). 

If we do change it here, we also need to edit the `configure transformer` section to be consistent.

website/docs/api/cli.mdx
polm: I feel like this is also being used in a generic sense, and "NER or `textcat`" is weird, but it does seem that it's less common to use this unmarked than "tok2vec", though there are exceptions like [here](https://github.com/explosion/spaCy/blob/03a0c2badcc3e0355d52110fa0c526e8bbb52324/website/docs/api/language.mdx?plain=1#L260).

spacy/cli/configure.py
```diff
@@ -0,0 +1,249 @@
...
+        print(nlp.config.to_str())
+    else:
+        nlp.config.to_disk(output_file)
+        msg.good("Saved config", output_file)
+
+    return nlp.config
+
+
+@configure_cli.command("tok2vec")
+def use_tok2vec(base_model: str, output_file: Path) -> Config:
```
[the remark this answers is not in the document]
polm: Best thing I can come up with is `configure_x_feature_source`. Maybe `use_x_features`?

spacy/cli/configure.py
```diff
@@ -0,0 +1,342 @@
...
+        msg.good("Saved config", output_file)
+
+    return conf
+
+
+@configure_cli.command("transformer")
+def use_transformer(
+    base_model: str, output_file: Path, transformer_name: str = "roberta-base"
+) -> Config:
+    """Replace pipeline tok2vec with transformer."""
```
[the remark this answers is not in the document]
polm: Looking at other CLI commands in spaCy, none of them have parameter listings. They do link to the full documentation though, so I added links like that.

spacy/cli/configure.py
```diff
@@ -0,0 +1,342 @@
...
+        but {name} has {len(tok2vecs)}."""
+    if len(tok2vecs) > 1:
+        msg.fail(fail_msg, exits=1)
+
+
+def _check_pipeline_names(nlp: Language, nlp2: Language) -> Dict[str, str]:
+    """Given two pipelines, try to rename any collisions in component names.
+
+    If a simple increment of a numeric suffix doesn't work, will give up.
+    """
```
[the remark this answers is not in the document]
polm: Changed to `_make_unique_pipeline_names`, which I think gets the point across?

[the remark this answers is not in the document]
rmitsch: Sounds good - maybe even `_make_unique_pipe_names`? Since we rename pipes and not pipelines. Feel free to stick with `_make_unique_pipeline_names` though.

spacy/cli/configure.py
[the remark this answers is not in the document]
rmitsch: Hm, that's true. Thanks for adding the doc links.

spacy/cli/configure.py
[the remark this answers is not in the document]
rmitsch: I'll think about it some more. If I can't come up with something better, it's fine to leave it as is, I think.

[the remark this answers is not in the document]
polm: OK. For now I went ahead and changed it `configure_x_feature_source`, but I would lean towards making it short and leaning on context / types for clarity.

spacy/cli/configure.py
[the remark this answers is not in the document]
polm: Good point, I think that's better. Updated.

website/docs/api/cli.mdx
rmitsch: Yes, that's what I meant. Thanks for the summary!

spacy/cli/configure.py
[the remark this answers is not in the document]
rmitsch: I agree, that's too verbose.

website/docs/api/cli.mdx
rmitsch: Sounds reasonable. In that case I agree with the lack of formatting here.

website/docs/api/cli.mdx
rmitsch: In line with the previous answer - I think you made the right call.

website/docs/api/cli.mdx
rmitsch: I see what you mean. Maybe it's easier to describe it a less abstract manner? Feel free to adjust in any way.

> Modify or combine existing configs. Can be used for e. g. swapping feature sources, creating configs to resume training or merging several configs into a single one. This may simplify parts of the development cycle. For example, it allows starting off with a config for a faster, less accurate model (using the CNN tok2vec) and conveniently switching to transformers later, without having to manually adjust the config.

spacy/cli/configure.py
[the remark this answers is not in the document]
polm: I have changed this back to the short names, though if there are any other suggestions I would be glad to change it again.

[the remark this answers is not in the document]
rmitsch: Yeah, I think it's fine. I can't come up with something snappy and more accurate.

svlandeg (commented):
There's a lot of features in this PR and ideally we would have split this up into separate PRs to make reviewing easier and the commits in the history more atomic. I guess we can keep it as is for now, but we'll want to review this carefully as there's a lot going on :-)

spacy/cli/configure.py
```diff
@@ -0,0 +1,214 @@
...
+    listeners = _get_listeners(nlp)
+    for listener in listeners:
+        listener_config = {
+            "@architectures": "spacy-transformers.TransformerListener.v1",
+            "grad_factor": 1.0,
+            "upstream": "transformer",
+            "pooling": {"@layers": "reduce_mean.v1"},
+        }
+        nlp.config["components"][listener]["model"]["tok2vec"] = listener_config
+
```
adrianeboyd, line 172: This would also need to update the `[training]` block.

spacy/cli/configure.py
```diff
@@ -0,0 +1,214 @@
...
+    listeners = _get_listeners(nlp)
+    for listener in listeners:
+        listener_config = {
+            "@architectures": "spacy.Tok2VecListener.v1",
+            "width": width,
+            "upstream": "tok2vec",
+        }
+        nlp.config["components"][listener]["model"]["tok2vec"] = listener_config
+
```
adrianeboyd, line 207: This may also need to update the `[training]` block. (I know that `tok2vec`->`transformer` doesn't work. I'm not 100% sure it doesn't work the other way around, but probably the `tok2vec` defaults are better.)

spacy/cli/configure.py
```diff
@@ -0,0 +1,214 @@
...
+TOK2VEC_ARCHS = [
+    ("spacy", "Tok2Vec"),
+    ("spacy", "HashEmbedCNN"),
+    ("spacy-transformers", "TransformerModel"),
+]
+# These are the listeners.
+LISTENER_ARCHS = [
+    ("spacy", "Tok2VecListener"),
+    ("spacy-transformers", "TransformerListener"),
+]
```
adrianeboyd, line 22: I wonder if there's a more general way to determine these lists?

PR #12039, opened 2023-01-01
Delete unused freeze argument from StringStore

tetsuok:
## Description

This PR removes unused `freeze` argument from `StringStrore`. The argument is not documented at https://spacy.io/api/stringstore, I think it's OK to remove it from `__init__`.

## Checklist


svlandeg:
Thanks for the PR! It looks like we forgot to clean up this parameter when we [removed](https://github.com/explosion/spaCy/pull/1424) the related functionality. I double check and it does seem like this is not used anywhere.

To be entirely on the safe side and avoid unnecessary downstream impact, I'll edit this PR to target the `v4` branch so that it becomes part of the larger cleanup & refactoring we're doing for the future spaCy `4.0.0`.

svlandeg:
In fact, PR https://github.com/explosion/spaCy/pull/11344 has already removed the `freeze` argument for the v4 branch, so I'll go ahead and close this PR. Thanks again!

Issue #12044, opened 2023-01-02
Mismatched IDs error when using nlp.rehearse on textcat


thomashacker:
### Discussed in https://github.com/explosion/spaCy/discussions/10861

Using `nlp.rehearse` on a `textcat` pipeline with a `tok2vec` listener results in `ValueError: [E953] Mismatched IDs`. This is not the case when using  `tok2vec` directly within the `textcat` component. The same goes for `textcat_multilabel`.

<div type='discussions-op-text'>

<sup>Originally posted by **nashcaps2255** May 27, 2022</sup>
Have a textcat multilabel model which I am trying to update with nlp.rehearse to alleviate the catastrophic forgetting problem. 

```
nlp = spacy.load('my_model')

examples = []
for line in file_:
   text, label = line.split("|")
   doc = nlp(text)
   gold_dict = {"cats": {label: float(1)}}
   gold_dict = Example.from_dict(doc, gold_dict)
   examples.append(example)


optimizer = nlp.resume_training()
nlp.rehearse(examples, sgd = optimizer) 
```

Results in......

```
ValueError: [E953] Mismatched IDs received by the Tok2Vec listener: 179568814531392983158587824 vs. 2172509679243279887229
```

 </div>

PR #12045, opened 2023-01-02
nlp.rehearse with textcat and tok2vec listener

thomashacker:
## Description

Related to [#12044](https://github.com/explosion/spaCy/issues/12044)

When using `nlp.rehearse` on a `textcat` pipeline with a `tok2vec` listener, it throws `ValueError: [E953] Mismatched IDs received by the Tok2Vec listener`. This is not the case when using an inline `tok2vec` listener. (The same goes for when using `textcat_multilabel`)

This PR aims to fix the issue, however it is still WIP and currently only contains the failing unit tests.

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/tests/training/test_rehearse.py
+++ b/spacy/tests/training/test_rehearse.py
@@ -1,6 +1,8 @@
 import pytest
 import spacy
 
+from thinc.api import Config
+
 from typing import List
 from spacy.training import Example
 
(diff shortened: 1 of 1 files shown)


thomashacker:
Update: `nlp.rehearse` works on a `ner` pipeline with a `tok2vec` listener. This indicates that the problem probably lies within the textcat architecture

thomashacker:
Alright, I figured out what the issue is.

**Issue summary**
Using `nlp.rehearse` with components that use `Tok2VecListeners` will cause a `mismatched ID` error, because the `Tok2Vec` component has no `rehearse` method and thus can't sync with the other components that do have the `rehearse` method implemented.

**Suggestion**
Add the `rehearse` method to the `Tok2Vec` component, so the other components are synced correctly. Additionally, 
make sure that the `_rehearsal_models` are also correctly linked to the `Tok2Vec` component (since they also have the `Tok2VecListeners`)

**ToDo's**
- Currently, the `rehearse` method of the `Tok2Vec` is not using the output of the `rehearsal_model` to influence the results of the actual model.

- The docs need to be updated (if we're going to add the new method to the `Tok2Vec` component)

thomashacker:
The tests currently fail because of this warning 
`DeprecationWarning: Out of bound index found. This was previously ignored when the indexing result contained no elements. In the future the index error will be raised. This error occurs either due to an empty slice, or if an array has zero elements even before indexing.`

It's somehow related to `textcat` + listener, since the fix works for the `NER` pipeline with listeners. I'll look into it, and try to provide a fix for that.

thomashacker:
Created a more clean PR https://github.com/explosion/spaCy/pull/12150 with focus on adding `Tok2Vec.rehearse`

PR #12047, opened 2023-01-03
Fix inconsistency in displaCy docs about page option

polm:
## Description

The `page` option, which wraps the output SVG in HTML, is true by default for `serve` but not for `render`. The `render` docs were wrong though, so this updates them.

### Types of change

Minor docs fix

## Checklist


svlandeg (approved):
Good catch!

polm:
The same language was used in a few internal docstrings so I cleaned that up too.

PR #12050, opened 2023-01-03
fix processing of "auto" in convert

svlandeg:
## Description
The refactoring of `walk_directory` in https://github.com/explosion/spaCy/pull/11376 changed the behaviour of matching on a suffix that is not "iob", "jsonl" or "conll", returning zero hits instead of all files. However, `convert` relies on getting all files when using the option "auto", but it wasn't properly hooked up to the new `walk_directory` implementation. This PR fixes that and introduces new unit tests to ensure the correct behaviour going forward.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -29,6 +29,8 @@ if TYPE_CHECKING:
 SDIST_SUFFIX = ".tar.gz"
 WHEEL_SUFFIX = "-py3-none-any.whl"
 
+AUTO = "auto"
+
 PROJECT_FILE = "project.yml"
 PROJECT_LOCK = "project.lock"
 COMMAND = "python -m spacy"
@@ -596,6 +598,8 @@ def walk_directory(path: Path, suffix: Optional[str] = None) -> List[Path]:
             continue
         elif path.is_dir():
             paths.extend(path.iterdir())
+        elif suffix == AUTO:
+            locs.append(path)
         elif suffix is not None and not path.parts[-1].endswith(suffix):
             continue
         else:
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -29,8 +29,6 @@ if TYPE_CHECKING:
 SDIST_SUFFIX = ".tar.gz"
 WHEEL_SUFFIX = "-py3-none-any.whl"
 
-AUTO = "auto"
-
 PROJECT_FILE = "project.yml"
 PROJECT_LOCK = "project.lock"
 COMMAND = "python -m spacy"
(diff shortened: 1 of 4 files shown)


spacy/cli/_util.py
```diff
@@ -596,6 +598,8 @@ def walk_directory(path: Path, suffix: Optional[str] = None) -> List[Path]:
             continue
         elif path.is_dir():
             paths.extend(path.iterdir())
+        elif suffix == AUTO:
+            locs.append(path)
```
polm, line 602: If I'm reading this correctly, this is the only logic change here. 

I think that the intent when this was written was that `suffix` should be `None` when `converter` is `auto`. While it's not a problem now given the limited ways we use this function, I think it makes sense to reserve special behavior for `None` rather than for a specific keyword, which kind of feels like mixing behavior. 

In that case, we can get the desired behavior by calculating suffix before this is called in conversion like this:

```
suffix = converter if converter != "auto" else None
... = walk_directory(path, suffix)
``` 

If we change this, then this particular `elif` segment can just be removed.

Also, I mentioned it in the last PR, but I don't think that the name of this function makes it clear what it's actually doing, so it would be good to add a docstring. Something like:

> Given a directory and a suffix, recursively find all files matching the suffix. Directories or files with names beginning with a `.` are ignored, but hidden flags on filesystems are not checked.

svlandeg: >I think that the intent when this was written was that `suffix` should be `None` when `converter` is `auto`. While it's not a problem now given the limited ways we use this function, I think it makes sense to reserve special behavior for None rather than for a specific keyword, which kind of feels like mixing behavior.

That makes sense, then we can keep the "auto" stuff limited to the `convert` functionality only.

svlandeg: This required a little more additional work, as the `verify_cli_args` function is impacted too. I moved all functionality of converting "auto" to the right suffix + warnings to the `_get_converter` function which now should be called before `verify_cli_args`. I also added more specific CLI tests.

spacy/cli/_util.py
```diff
@@ -583,6 +583,10 @@ def setup_gpu(use_gpu: int, silent=None) -> None:
 
 
 def walk_directory(path: Path, suffix: Optional[str] = None) -> List[Path]:
+    """Given a directory and a suffix, recursively find all files matching the suffix.
+    Directories or files with names beginning with a . are ignored, but hidden flags on
+    filesystems are not checked.
+    When provided with a suffix `None`, all files are returned without filtering."""
```
svlandeg, line 589:
```suggestion
    When provided with a suffix `None`, there is no suffix-based filtering."""
```

polm (approved):
Thanks for the edit, looks good.

PR #12073, opened 2023-01-09
Restore v2 token_acc score implementation

adrianeboyd:
## Description

In the v3 scorer refactoring, `token_acc` was implemented incorrectly. It should use `precision` instead of `fscore` for the measure of correctly aligned tokens / number of predicted tokens.

Fix the docs to reflect that the measure uses the number of predicted tokens rather than the number of gold tokens.

### Types of change

Bug fix, docs.

## Checklist

### DIFF
--- a/spacy/scorer.py
+++ b/spacy/scorer.py
@@ -174,7 +174,7 @@ class Scorer:
             prf_score.score_set(pred_spans, gold_spans)
         if len(acc_score) > 0:
             return {
-                "token_acc": acc_score.fscore,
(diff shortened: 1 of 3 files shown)


adrianeboyd:
Because the tokenizer scorer isn't versioned in the config, I'm not 100% sure we want to change this for v3 at this point. I do feel like it's probably not a huge deal to change this for v3.5.0, though.

svlandeg (approved):
Seems fine to me to have this as a bug fix in 3.5, yea.

PR #12077, opened 2023-01-09
Moved 'spacy' folder inside 'master' folder

Retinpkumar:
Moved the spacy folder inside the master folder.

## Description
```python
python -m spacy download <model-name>
```
command is not working. Most of the code blocks within the documentation doesn't show the codes.

### Types of change
Bugfix
 It can be seen that previously the spacy folder was inside a master folder which is missing now. Hence moved the spacy folder within a master folder.

## Checklist

- [* ] I confirm that I have the right to submit this contribution under the project's MIT license.
- [*] I ran the tests, and all new and existing tests passed.
- [*] My changes don't require a change to the documentation, or if they do, I've added all required information.


shadeMe:
Hi, thanks for opening this pull request! However, the issue you had was likely due to an incorrect installation of the `spacy` package as opposed to a bug in the codebase. So, the change in directory structure that you're proposing shouldn't be necessary; in fact, it would break the library for our existing users.

> It can be seen that previously the spacy folder was inside a master folder which is missing now. Hence moved the spacy folder within a master folder.

Could you clarify what you mean with _previously_? Did you happen to download the code as a ZIP archive from GitHub? GitHub creates a top-level directory under the name of the active branch when one does that - that could explain where the `master` folder came from.

If you continue to have issues with getting spaCy up and running locally, you can open a new [Discussion thread](https://github.com/explosion/spaCy/discussions/categories/help-installation) with the relevant details.

PR #12086, opened 2023-01-10
Bring back support for `update_with_oracle_cut_size`

danieldk:
## Description

This option was available in the pre-refactor parser, but was never implemented in the refactored parser. This option cuts transition sequences that are longer than `update_with_oracle_cut` size into separate sequences that have at most `update_with_oracle_cut` transitions. The oracle (gold standard) transition sequence is used to determine the cuts and the initial states for the additional sequences.

Applying this cut makes the batches more homogeneous in the transition sequence lengths, making forward passes (and as a consequence training) much faster.

Training time 1000 steps on `de_core_news_lg`:

- Before this change: 149s
- After this change: 68s
- Pre-refactor parser: 81s

Also fix up an unrelated issue where the parser rehearsal tests would fail. Since this PR affects rehearsal as well (in a minor way), I bundled up that fix here.

### Types of change

Performance improvement.

## Checklist

### DIFF
--- a/spacy/ml/tb_framework.pyx
+++ b/spacy/ml/tb_framework.pyx
@@ -148,32 +148,74 @@ def init(
     # model = _lsuv_init(model)
     return model
 
-InWithoutActions = Tuple[List[Doc], TransitionSystem]
-InWithActions = Tuple[List[Doc], TransitionSystem, List[Ints1d]]
-InT = TypeVar("InT", InWithoutActions, InWithActions)
 
-def forward(model, docs_moves: InT, is_train: bool):
-    if len(docs_moves) == 2:
-        docs, moves = docs_moves
-        actions = None
-    else:
-        docs, moves, actions = docs_moves
+class TransitionModelInputs:
+    """
+    Input to transition model.
+    """
+
+    # dataclass annotation is not yet supported in Cython 0.29.x,
+    # so, we'll do something close to it.
+
+    actions: Optional[List[Ints1d]]
+    docs: List[Doc]
+    max_moves: int
+    moves: TransitionSystem
+    states: Optional[List[State]]
+
+    __slots__ = [
+        "actions",
+        "docs",
+        "max_moves",
+        "moves",
+        "states",
+    ]
+
+    def __init__(self, docs: List[Doc], moves: TransitionSystem,
+        actions: Optional[List[Ints1d]]=None, max_moves: int=0,
+        states: Optional[List[State]]=None):
+        """
+        actions (Optional[List[Ints1d]]): actions to apply for each Doc.
+        docs (List[Doc]): Docs to predict transition sequences for.
+        max_moves: (Optional[int]): the maximum number of moves to apply,
+            values less than 1 will apply moves to states until they are
+            final states.
+        moves (TransitionSystem): the transition system to use when predicting
+            the transition sequences.
+        states (Optional[List[States]]): the initial states to predict the
+            transition sequences for. When absent, the initial states are
+            initialized from the provided Docs.
+        """
+        self.actions = actions
+        self.docs = docs
+        self.moves = moves
+        self.max_moves = max_moves
+        self.states = states
+
+
+def forward(model, inputs: TransitionModelInputs, is_train: bool):
+    docs = inputs.docs
+    moves = inputs.moves
+    actions = inputs.actions
 
     beam_width = model.attrs["beam_width"]
     hidden_pad = model.get_param("hidden_pad")
     tok2vec = model.get_ref("tok2vec")
 
-    states = moves.init_batch(docs)
+    states = moves.init_batch(docs) if inputs.states is None else inputs.states
     tokvecs, backprop_tok2vec = tok2vec(docs, is_train)
     tokvecs = model.ops.xp.vstack((tokvecs, hidden_pad))
     feats, backprop_feats = _forward_precomputable_affine(model, tokvecs, is_train)
     seen_mask = _get_seen_mask(model)
 
-    # Fixme: support actions in forward_cpu
     if beam_width == 1 and not is_train and isinstance(model.ops, NumpyOps):
+        # Note: max_moves is only used during training, so we don't need to
+        #       pass it to the greedy inference path.
         return _forward_greedy_cpu(model, moves, states, feats, seen_mask, actions=actions)
     else:
-        return _forward_fallback(model, moves, states, tokvecs, backprop_tok2vec, feats, backprop_feats, seen_mask, is_train, actions=actions)
+        return _forward_fallback(model, moves, states, tokvecs, backprop_tok2vec,
+            feats, backprop_feats, seen_mask, is_train, actions=actions,
+            max_moves=inputs.max_moves)
 
 
 def _forward_greedy_cpu(model: Model, TransitionSystem moves, states: List[StateClass], np.ndarray feats,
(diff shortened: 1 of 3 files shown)


spacy/ml/tb_framework.pyx
```diff
@@ -148,32 +148,74 @@ def init(
...
+
+    def __init__(self, docs: List[Doc], moves: TransitionSystem,
+        actions: Optional[List[Ints1d]]=None, max_moves: int=0,
+        states: Optional[List[State]]=None):
+        """
+        actions (Optional[List[Ints1d]]): actions to apply for each Doc.
+        docs (List[Doc]): Docs to predict transition sequences for.
+        max_moves: (Optional[int]): the maximum number of moves to apply,
+            values less than 1 will apply moves to states until they are
+            final states.
```
shadeMe, line 182: A couple of things:
- `max_moves` isn't an optional by virtue of accepting a `None`, so we should update the description.
- On a similar note, given that the semantics of a zero value and is the same as that of any negative value, we could make this more explicit by restricting the lower-bound to zero and raising an exception when we receive a negative value.

spacy/ml/tb_framework.pyx
```diff
@@ -148,32 +148,74 @@ def init(
...
 
-    states = moves.init_batch(docs)
+    states = moves.init_batch(docs) if inputs.states is None else inputs.states
     tokvecs, backprop_tok2vec = tok2vec(docs, is_train)
     tokvecs = model.ops.xp.vstack((tokvecs, hidden_pad))
     feats, backprop_feats = _forward_precomputable_affine(model, tokvecs, is_train)
     seen_mask = _get_seen_mask(model)
 
-    # Fixme: support actions in forward_cpu
     if beam_width == 1 and not is_train and isinstance(model.ops, NumpyOps):
```
shadeMe, line 211: Minor nitpick:
```suggestion
    if not is_train and beam_width == 1 and isinstance(model.ops, NumpyOps):
```

spacy/pipeline/transition_parser.pyx
```diff
@@ -312,12 +332,15 @@ class Parser(TrainablePipe):
         del backprop_scores
         return losses
 
-    def get_loss(self, states_scores, examples):
-        states, scores = states_scores
+    def get_loss(self, states_scores, examples, max_moves):
```
shadeMe, line 335: Missing type hints.

spacy/pipeline/transition_parser.pyx
```diff
@@ -335,7 +358,7 @@ class Parser(TrainablePipe):
         d_scores -= (costs <= best_costs) * (exp_gscores / gZ)
         return d_scores
 
-    def _get_costs_from_histories(self, examples, histories):
+    def _get_costs_from_histories(self, examples, gold_states, init_states, histories, max_moves):
```
shadeMe, line 361: Missing type hints.

spacy/pipeline/transition_parser.pyx
```diff
@@ -502,6 +527,49 @@ class Parser(TrainablePipe):
                     raise ValueError(Errors.E149) from None
         return self
 
+    def _init_gold_batch(self, examples, max_length):
```
shadeMe, line 530: Missing type hints.

spacy/pipeline/transition_parser.pyx
```diff
@@ -502,6 +527,49 @@ class Parser(TrainablePipe):
...
+        """Make a square batch, of length equal to the shortest transition
+        sequence or a cap. A long
+        doc will get multiple states. Let's say we have a doc of length 2*N,
+        where N is the shortest doc. We'll make two states, one representing
+        long_doc[:N], and another representing long_doc[N:]."""
```
shadeMe, line 535:
```suggestion
        sequence or a cap. A long doc will get multiple states. Let's say we have a doc 
        of length 2*N, where N is the shortest doc. We'll make two states, one representing
        long_doc[:N], and another representing long_doc[N:]."""
```
Whitespace.

spacy/pipeline/transition_parser.pyx
danieldk: I think we should do these in a separate PR, practically nothing in these modules have type hints. So I'd be more in favor of adding them all at once in a separate PR.

spacy/ml/tb_framework.pyx
```diff
@@ -148,32 +148,74 @@ def init(
...
+        "actions",
+        "docs",
+        "max_moves",
+        "moves",
+        "states",
+    ]
+
+    def __init__(self, docs: List[Doc], moves: TransitionSystem,
+        actions: Optional[List[Ints1d]]=None, max_moves: int=0,
+        states: Optional[List[State]]=None):
```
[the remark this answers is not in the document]
danieldk: What I wouldn't give for black for Python 😭 .

spacy/pipeline/transition_parser.pyx
danieldk: I’ll create a ticket for it tomorrow.

spacy/ml/tb_framework.pyx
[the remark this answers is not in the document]
danieldk: Fixed.

spacy/ml/tb_framework.pyx
danieldk: Fixed `max_moves` description.

> On a similar note, given that the semantics of a zero value and is the same as that of any negative value, we could make this more explicit by restricting the lower-bound to zero and raising an exception when we receive a negative value.

The pre-refactor parser does not have this restriction, so adding this restriction now could break existing configurations and it doesn't give us much.

spacy/ml/tb_framework.pyx
danieldk: Fixed.

spacy/pipeline/transition_parser.pyx
danieldk: Created a ticket. I resolved the other type hints comments.

spacy/pipeline/transition_parser.pyx

spacy/ml/tb_framework.pyx
```diff
@@ -230,7 +272,7 @@ cdef list _parse_batch(CBlas cblas, TransitionSystem moves, StateC** states,
 
 
 def _forward_fallback(model: Model, moves: TransitionSystem, states: List[StateClass], tokvecs, backprop_tok2vec, feats, backprop_feats, seen_mask, is_train: bool,
-                  actions: Optional[List[Ints1d]]=None):
+        actions: Optional[List[Ints1d]]=None, max_moves=0):
```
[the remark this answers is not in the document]
danieldk: Fixed for the newly-added argument.

shadeMe (approved):

spacy/ml/tb_framework.pyx
```diff
@@ -281,6 +336,9 @@ def _forward_fallback(model: Model, moves: TransitionSystem, states: List[StateC
             all_ids.append(ids)
             all_statevecs.append(statevecs)
             all_which.append(which)
+        if max_moves >= 1 and n_moves >= max_moves:
```
svlandeg, line 339: I'd personally prefer chaining this: 
```suggestion
        if n_moves >= max_moves >= 1:
```

svlandeg (commented):
Looking good!

Can we add some unit tests here for the general behaviour, but also to test some edge cases like `max_moves` being 0 or 1, etc?

spacy/ml/tb_framework.pyx
danieldk: Fixed.

danieldk:
> Can we add some unit tests here for the general behaviour, but also to test some edge cases like max_moves being 0 or 1, etc?

Extended the parser overfitting test to train with cut sizes 0, 1, 5, and 100. Found (and fixed) a bug with a cut size of 1.

PR #12092, opened 2023-01-11
Allow spacy-transformers v1.2.x in transformers extra

adrianeboyd:
## Description

Allow `spacy-transformers` v1.2.x in transformers extra.

### Types of change

Enhancement?

## Checklist


adrianeboyd:
To clarify the version range: a model package should specify either v1.1.x or v1.2.x due to differences in behavior (model serialization, slow/fast tokenizer alignments), but spacy v3.5.0 itself is compatible with either.

PR #12094, opened 2023-01-11
Fix batching regression

danieldk:
## Description

Some time ago, the spaCy v4 branch switched to the new Thinc v9 schedule. However, this introduced an error in how batching is handed.

In the PR, the batchers were changed to keep track of their step, so that the step can be passed to the schedule. However, the issue is that the training loop repeatedly calls the batching functions (rather than using an infinite generator/iterator). So, the step and therefore the schedule would be reset each epoch. Before the schedule switch we didn't have this issue, because the old schedules were stateful.

This PR fixes this issue by reverting the batching functions to use a (stateful) generator. Their registry functions do accept a `Schedule` and we convert `Schedule`s to generators.

### Types of change

Bugfix

## Checklist


spacy/training/batchers.py
```diff
@@ -112,13 +113,12 @@ def minibatch_by_padded_size(
         The `len` function is used by default.
     """
     if isinstance(size, int):
-        size_ = constant_schedule(size)
+        size_ = itertools.repeat(size)  # type: Iterator[int]
```
shadeMe, line 116: Can't this be a regular type hint attached to `size_`?

spacy/util.py
```diff
@@ -1583,12 +1583,12 @@ def minibatch(items, size):
...
-        size_ = constant_schedule(size)
+        size_ = itertools.repeat(size)
     else:
-        size_ = size
+        size_ = iter(size)
     items = iter(items)
-    for step in itertools.count():
-        batch_size = size_(step)
+    while True:
+        batch_size = next(size_)
```
shadeMe, line 1591: General question: does the `Schedule` API (or the `Optimizer`) guarantee that any generator derived from it will never throw a `StopIteration` error? I see that we call `next` in the batcher code with that implicit expectation.

danieldk: Yes, that's guaranteed. It uses an infinite iterator:

https://github.com/explosion/thinc/blob/e58ca10bf97f8490eb4132208d80f3bb43431186/thinc/schedules.py#L62

Which is used to feed values to the schedule, which does not return `StopIteration` and must always return a value:

https://github.com/explosion/thinc/blob/e58ca10bf97f8490eb4132208d80f3bb43431186/thinc/schedules.py#L32

spacy/training/batchers.py
danieldk: Yep. Fixed.

shadeMe (approved):

PR #12098, opened 2023-01-12
Handle missing annotations in the edit tree lemmatizer

danieldk:
## Description

The losses/gradients of missing annotations were not correctly masked out. Fix this and check the masking in the partial data test.

### Types of change

Bugfix.

## Checklist


svlandeg (approved):
Thanks! LGTM.

Issue #12000, opened 2022-12-19
spacy package CLI command accepts list of code_paths, but the others do not


kinghuang:
I recently started a new spaCy project and decided not to create a separate Python module with all the custom code. While I can pass a comma-separated list of code paths to the [spacy package](https://spacy.io/api/cli#package) command, other CLI commands such as [spacy train](https://spacy.io/api/cli#train) and [spacy assemble](https://spacy.io/api/cli#assemble) only accept a single value for the `--code` option. This makes it impossible to build a project with more than one code file, even though it's possible to assemble one with multiple files.

It would be really helpful if all the spaCy CLI commands accepted a comma-separated list for the `--code` option. Otherwise, all the code has to be stuffed into a single file.

The `--code` option exists in the following commands:

* [debug](https://spacy.io/api/cli#debug)
* [train](https://spacy.io/api/cli#train)
* [pretrain](https://spacy.io/api/cli#pretrain)
* [evaluate](https://spacy.io/api/cli#evaluate)
* [assemble](https://spacy.io/api/cli#assemble)
* [package](https://spacy.io/api/cli#package) (accepts comma-separated list)

## How to reproduce the behaviour

**OK**:

```sh
spacy package --code file_a.py,file_b.py …
```

The comma-separated value to the `--code` option is split and all the code files are loaded ([package.py#L48](https://github.com/explosion/spaCy/blob/18ffe5bbd6a554920107ff48d1387df34c3f872a/spacy/cli/package.py#L48)).

**Not OK**:

```sh
spacy assemble --code file_a.py,file_b.py …

Path to Python code not found
```

The comma-separated value to the `--code` option is used as the literal path, which fails to load ([assemble.py#L41](https://github.com/explosion/spaCy/blob/18ffe5bbd6a554920107ff48d1387df34c3f872a/spacy/cli/assemble.py#L41)).

## Your Environment

* Operating System: macOS Ventura 13.1 (22C65)
* Python Version Used: 3.10.9
* spaCy Version Used: 3.4.3
* Environment Information:

polm:
Thanks for pointing that out! I think that's just an oversight, and it sounds like a nice feature to me.

polm:
As a note, this came up in https://github.com/explosion/spaCy/discussions/10694 before.

PR #12101, opened 2023-01-13
Accept multiple code files in all CLI commands

polm:
## Description

Prior to this PR, only the package command accepted multiple code files. This adds support for multiple files to all relevant commands. This suggestion was raised in #12000. 

This is in draft for now because there are no tests for this yet. (There are also no tests for the existing functionality.)

### Types of change

enhancement

## Checklist


polm:
There are tests for all commands except `pretrain` now. 

This ended up being trickier than expected. The issue is the CliRunner doesn't provide process isolation, and the `--code` argument uses `import` to modify the global environment, so it's necessary to use `subprocess.run` (or similar) for the tests to run properly. 

Most commands could be run using the same structure, so the test is compact and easy to follow: `debug data`, `debug config`, `train`, and `assemble` work that way. `evaluate` can't use the same structure, so it has its own separate test, though it's still pretty compact.

For `pretrain`, it can't be used without an actual trainable component. I could add that, but it would essentially need its own config, and would get pretty verbose. Given there's nothing special about the code in pretrain in particular, I think it's reasonable to omit tests there.

spacy/cli/assemble.py
```diff
@@ -19,7 +19,7 @@ def assemble_cli(
     ctx: typer.Context,  # This is only used to read additional arguments
     config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
     output_path: Path = Arg(..., help="Output directory to store assembled pipeline in"),
-    code_path: Optional[Path] = Opt(None, "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    code_paths: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
adrianeboyd, line 22: I wonder whether we can make this less breaking for people calling into this from python? We can probably keep the name `code_path` without it being a real headache and also still allow `None` and `Path`?

But I'm not sure what others think?

polm: Given that this is for v4, I think it's more important that this argument be consistent between commands than consistent with previous versions.

svlandeg:
@polm : can you resolve the conflicts? 🙏

polm:
Conflicts are resolved now, though it looks like there are other PRs changing the CLI tests too (#12155 is one).

spacy/cli/assemble.py
svlandeg: I agree with Adriane that this feels unnecessarily breaking, even for v4. Users will have their CI scripts and it'll be annoying to change all of these if you're only using a single path. It looks like right now, the only command using `code_paths` is `package` and I agree that it's slightly unfortunate not to have this consistent, but I'm not sure the consistency is worth the disruption.

polm: OK, I'll go ahead and change it, but to be clear this only affects usage from Python - I expect most people using this feature are using command line scripts, which would be unaffected.

About the types, I think one issue here is that there are no Union types for arguments, so it has to be `str` or `Path`. I am not sure if #12258 will fix that or not.

polm: Actually, on thinking about this a little more, I'm not sure it even makes sense to use this argument if you're calling these functions from Python, since you can just import the modules yourself directly in that case. 

That said, I have changed the names back, and I will leave them that way for now.

website/docs/api/cli.mdx
```diff
@@ -175,15 +175,15 @@ validation error with more details.
...
+| Name                   | Description                                                                                                                                                                                            |
+| ---------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `base_path`            | Path to base config to fill, e.g. generated by the [quickstart widget](/usage/training#quickstart). ~~Path (positional)~~                                                                              |
+| `output_file`          | Path to output `.cfg` file or "-" to write to stdout so you can pipe it to a file. Defaults to "-" (stdout). ~~Path (positional)~~                                                                     |
+| `--code`, `-c`         | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~ |
```
rmitsch, line 182: I'm not sure whether we should drop `\(option)` here. Technically it's `str` and not `Optional[str]`, but we use a blank string as default argument - so from a usage viewpoint it's optional. Or we change `code_path` to be `Optional[str]`?
```suggestion
| `--code`, `-c`         | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~ |
```

website/docs/api/cli.mdx
```diff
@@ -447,7 +447,7 @@ File       /path/to/thinc/thinc/schedules.py (line 91)
...
 | ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `config_path`            | Path to [training config](/api/data-formats#config) file containing all settings and hyperparameters. If `-`, the data will be [read from stdin](/usage/training#config-stdin). ~~Union[Path, str] \(positional)~~             |
-| `--code`, `-c`           | Path to Python file with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~                                           |
+| `--code`, `-c`           | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~                         |
```
rmitsch, line 450: See above.
```suggestion
| `--code`, `-c`           | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~                         |
```

website/docs/api/cli.mdx
```diff
@@ -1056,7 +1056,7 @@ $ python -m spacy train [config_path] [--output] [--code] [--verbose] [--gpu-id]
 | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
 | `config_path`     | Path to [training config](/api/data-formats#config) file containing all settings and hyperparameters. If `-`, the data will be [read from stdin](/usage/training#config-stdin). ~~Union[Path, str] \(positional)~~ |
 | `--output`, `-o`  | Directory to store trained pipeline in. Will be created if it doesn't exist. ~~Optional[Path] \(option)~~                                                                                                          |
-| `--code`, `-c`    | Path to Python file with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~                               |
+| `--code`, `-c`    | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~             |
```
rmitsch, line 1059: See above.
```suggestion
| `--code`, `-c`    | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~             |
```

website/docs/api/cli.mdx
```diff
@@ -1126,7 +1126,7 @@ $ python -m spacy pretrain [config_path] [output_dir] [--code] [--resume-path] [
...
 | `config_path`           | Path to [training config](/api/data-formats#config) file containing all settings and hyperparameters. If `-`, the data will be [read from stdin](/usage/training#config-stdin). ~~Union[Path, str] \(positional)~~ |
 | `output_dir`            | Directory to save binary weights to on each epoch. ~~Path (positional)~~                                                                                                                                           |
-| `--code`, `-c`          | Path to Python file with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~                               |
+| `--code`, `-c`          | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~             |
```
rmitsch, line 1129: See above.
```suggestion
| `--code`, `-c`          | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~             |
```

website/docs/api/cli.mdx
```diff
@@ -1162,18 +1162,18 @@ skew. To render a sample of dependency parses in a HTML file using the
...
+| ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `model`                   | Pipeline to evaluate. Can be a package or a path to a data directory. ~~str (positional)~~                                                                                                             |
+| `data_path`               | Location of evaluation data in spaCy's [binary format](/api/data-formats#training). ~~Path (positional)~~                                                                                              |
+| `--output`, `-o`          | Output JSON file for metrics. If not set, no metrics will be exported. ~~Optional[Path] \(option)~~                                                                                                    |
+| `--code`, `-c`            | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~ |
```
rmitsch, line 1170: See above.
```suggestion
| `--code`, `-c`            | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~ |
```

website/docs/api/cli.mdx
```diff
@@ -1215,19 +1215,19 @@ When a directory is provided it is traversed recursively to collect all files.
...
+| ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `model`                   | Pipeline to apply to the data. Can be a package or a path to a data directory. ~~str (positional)~~                                                                                                    |
+| `data_path`               | Location of data to be evaluated in spaCy's [binary format](/api/data-formats#training), jsonl, or plain text. ~~Path (positional)~~                                                                   |
+| `output-file`, `-o`       | Output `DocBin` path. ~~str (positional)~~                                                                                                                                                             |
+| `--code`, `-c`            | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~ |
```
rmitsch, line 1223: See above.
```suggestion
| `--code`, `-c`            | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~ |
```

website/docs/api/cli.mdx
```diff
@@ -1254,19 +1254,19 @@ be provided.
...
+| `pipe_name`             | Name of pipe to examine thresholds for. ~~str (positional)~~                                                                                                                                           |
+| `threshold_key`         | Key of threshold attribute in component's configuration. ~~str (positional)~~                                                                                                                          |
+| `scores_key`            | Name of score to metric to optimize. ~~str (positional)~~                                                                                                                                              |
+| `--n_trials`, `-n`      | Number of trials to determine optimal thresholds. ~~int (option)~~                                                                                                                                     |
+| `--code`, `-c`          | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~ |
```
rmitsch, line 1265: See above.
```suggestion
| `--code`, `-c`          | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~ |
```

website/docs/api/cli.mdx
```diff
@@ -1290,7 +1290,7 @@ $ python -m spacy assemble [config_path] [output_dir] [--code] [--verbose] [over
 | ----------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `config_path`     | Path to the [config](/api/data-formats#config) file containing all settings and hyperparameters. If `-`, the data will be [read from stdin](/usage/training#config-stdin). ~~Union[Path, str] \(positional)~~ |
 | `output_dir`      | Directory to store the final pipeline in. Will be created if it doesn't exist. ~~Optional[Path] \(option)~~                                                                                                   |
-| `--code`, `-c`    | Path to Python file with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions). ~~Optional[Path] \(option)~~                                                |
+| `--code`, `-c`    | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~        |
```
rmitsch, line 1293: See above.
```suggestion
| `--code`, `-c`    | Comma-separated paths to Python files with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~str \(option)~~        |
```

spacy/cli/assemble.py
rmitsch: @svlandeg @adrianeboyd As this has been renamed, are we good to go ahead and close this thread?

website/docs/api/cli.mdx
adrianeboyd: I think "option" is referring to the structure of the CLI args (vs. flag), and is not related to `Optional` here.

spacy/cli/assemble.py
```diff
@@ -19,7 +19,7 @@ def assemble_cli(
     ctx: typer.Context,  # This is only used to read additional arguments
     config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
     output_path: Path = Arg(..., help="Output directory to store assembled pipeline in"),
-    code_path: Optional[Path] = Opt(None, "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
adrianeboyd, line 22: The help messages all need to be updated.

```suggestion
    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be imported"),
```

spacy/cli/assemble.py
adrianeboyd: Looking at this again, since we can't support any of the original types for `code_path` it will break existing code anyway, so I would go back to `code_paths` for the new comma-separated option.

To be less breaking, we could continue to support `code_path` as a hidden option / python argument as it was before?

In general it might be nice to support the more frequent comma-hassle-avoiding kind of option where you provide the option multiple times for multiple files like `-c file1.py -c file2.py` but I'm not sure it's really worth this change in this particular case.

spacy/cli/debug_config.py
```diff
@@ -20,7 +20,7 @@ def debug_config_cli(
     # fmt: off
     ctx: typer.Context,  # This is only used to read additional arguments
     config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
-    code_path: Optional[Path] = Opt(None, "--code-path", "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
rmitsch, line 23:
```suggestion
    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be imported"),
```

spacy/cli/debug_data.py
```diff
@@ -52,7 +52,7 @@ def debug_data_cli(
     # fmt: off
     ctx: typer.Context,  # This is only used to read additional arguments
     config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
-    code_path: Optional[Path] = Opt(None, "--code-path", "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
rmitsch, line 55:
```suggestion
    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be imported"),
```

spacy/cli/evaluate.py
```diff
@@ -22,7 +22,7 @@ def evaluate_cli(
     model: str = Arg(..., help="Model name or path"),
     data_path: Path = Arg(..., help="Location of binary evaluation data in .spacy format", exists=True),
     output: Optional[Path] = Opt(None, "--output", "-o", help="Output JSON file for metrics", dir_okay=False),
-    code_path: Optional[Path] = Opt(None, "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
rmitsch, line 25:
```suggestion
    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be imported"),
```

spacy/cli/package.py
```diff
@@ -20,7 +20,7 @@ def package_cli(
     # fmt: off
     input_dir: Path = Arg(..., help="Directory with pipeline data", exists=True, file_okay=False),
     output_dir: Path = Arg(..., help="Output parent directory", exists=True, file_okay=False),
-    code_paths: str = Opt("", "--code", "-c", help="Comma-separated paths to Python file with additional code (registered functions) to be included in the package"),
+    code_paths: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
rmitsch, line 23:
```suggestion
    code_paths: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be imported"),
```

spacy/cli/pretrain.py
```diff
@@ -19,7 +19,7 @@ def pretrain_cli(
     ctx: typer.Context,  # This is only used to read additional arguments
     config_path: Path = Arg(..., help="Path to config file", exists=True, dir_okay=False, allow_dash=True),
     output_dir: Path = Arg(..., help="Directory to write weights to on each epoch"),
-    code_path: Optional[Path] = Opt(None, "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
rmitsch, line 22:
```suggestion
    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be imported"),
```

spacy/cli/train.py
```diff
@@ -20,7 +20,7 @@ def train_cli(
     ctx: typer.Context,  # This is only used to read additional arguments
     config_path: Path = Arg(..., help="Path to config file", exists=True, allow_dash=True),
     output_path: Optional[Path] = Opt(None, "--output", "--output-path", "-o", help="Output directory to store trained pipeline in"),
-    code_path: Optional[Path] = Opt(None, "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
+    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be included in the package"),
```
rmitsch, line 23:
```suggestion
    code_path: str = Opt("", "--code", "-c", help="Comma-separated paths to Python files with additional code (registered functions) to be imported"),
```

spacy/cli/assemble.py
rmitsch: We use `code_path` for the comma-separated option right now (except in `package`, where we use `code_paths`). Do we want to use `code_paths` for comma-separated values and `code_path` for BW compatibility or are we happy with things as they are?

adrianeboyd:
I think the inconsistency between `code_paths` and `code_path` is not great, but I think the python `_cli` APIs are rarely used (and should stay that way), and we can break the names and the types or just the types, so let's stick with just the types for right now.

Thanks again @polm!

adrianeboyd:

adrianeboyd:

adrianeboyd:

PR #12119, opened 2023-01-18
Update Dockerfile to work with Next.js

danieldk:
## Description

Update Dockerfile to work with Next.js

- Update to Node 18.
- Do not run as root, this also works better with Node privilege-dropping.
- Update README with new run instructions and adding the `--rm` flag to avoid leaving a bunch of unused Docker containers.
- Also change README to recommend building the image locally. Image builds are pretty fast and the uploaded images get outdated pretty quickly.
- Add .dockerignore to avoid sending large build contexts.

### Types of change

Maintenance

## Checklist

### DIFF
--- a/website/.dockerignore
+++ b/website/.dockerignore
@@ -1,0 +1,9 @@
+.cache/
+.next/
+public/
+node_modules
+.npm
+logs
+*.log
+npm-debug.log*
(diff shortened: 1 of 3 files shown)


svlandeg (approved):
Thanks!

PR #12122, opened 2023-01-18
Add `spacy.PlainTextCorpusReader.v1`

danieldk:
## Description

This is a corpus reader that reads plain text corpora with the following format:

- UTF-8 encoding
- One line per document.
- Blank lines are ignored.

It is useful for applications where we deal with very large corpora, such as distillation, and don't want to deal with the space overhead of serialized formats. Additionally, many large corpora already use such a text format, keeping the necessary preprocessing to a minimum.

### Types of change

New feature

## Checklist

### DIFF
--- a/spacy/tests/training/test_corpus.py
+++ b/spacy/tests/training/test_corpus.py
@@ -1,0 +1,75 @@
+from typing import Iterable, List, Tuple
+from contextlib import contextmanager
+from pathlib import Path
+import pytest
+import tempfile
+
+from spacy.lang.en import English
+from spacy.training import Example, PlainTextCorpus
+
+# Intentional newlines to check that they are skipped.
+PLAIN_TEXT_DOC = """
+
+This is a doc. It contains two sentences.
+This is another doc.
+
+A third doc.
+
+"""
+
+PLAIN_TEXT_DOC_TOKENIZED = [
+    [
+        "This",
+        "is",
+        "a",
+        "doc",
+        ".",
+        "It",
+        "contains",
+        "two",
+        "sentences",
+        ".",
+    ],
+    ["This", "is", "another", "doc", "."],
+    ["A", "third", "doc", "."],
+]
+
+
+@pytest.mark.parametrize("min_length, max_length", [(0, 0), (0, 5), (5, 0), (5, 5)])
+def test_plain_text_reader(min_length, max_length):
+    nlp = English()
+    with _string_to_tmp_file(PLAIN_TEXT_DOC) as f:
+        corpus = PlainTextCorpus(
+            Path(f.name), min_length=min_length, max_length=max_length
+        )
+
+        check = [
+            doc
+            for doc in PLAIN_TEXT_DOC_TOKENIZED
+            if len(doc) >= min_length and (max_length == 0 or len(doc) <= max_length)
+        ]
+        reference, predicted = _examples_to_tokens(corpus(nlp))
+
+        assert reference == check
+        assert predicted == check
+
+
+@contextmanager
+def _string_to_tmp_file(s: str):
+    with tempfile.NamedTemporaryFile(suffix=".txt") as f:
+        f.write(s.encode("utf-8"))
+        f.seek(0)
+        yield f
+
+
+def _examples_to_tokens(
+    examples: Iterable[Example],
+) -> Tuple[List[List[str]], List[List[str]]]:
+    reference = []
+    predicted = []
+
+    for eg in examples:
+        reference.append([t.text for t in eg.reference])
+        predicted.append([t.text for t in eg.predicted])
+
+    return reference, predicted
--- a/spacy/tests/training/test_corpus.py
+++ b/spacy/tests/training/test_corpus.py
@@ -38,9 +38,9 @@ PLAIN_TEXT_DOC_TOKENIZED = [
 @pytest.mark.parametrize("min_length, max_length", [(0, 0), (0, 5), (5, 0), (5, 5)])
 def test_plain_text_reader(min_length, max_length):
     nlp = English()
-    with _string_to_tmp_file(PLAIN_TEXT_DOC) as f:
+    with _string_to_tmp_file(PLAIN_TEXT_DOC) as file_path:
         corpus = PlainTextCorpus(
-            Path(f.name), min_length=min_length, max_length=max_length
+            file_path, min_length=min_length, max_length=max_length
         )
 
         check = [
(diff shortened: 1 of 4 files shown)


spacy/training/corpus.py
```diff
@@ -257,3 +268,54 @@ def __call__(self, nlp: "Language") -> Iterator[Example]:
...
+                        doc = nlp.make_doc(text)
+                        if self.min_length >= 1 and len(doc) < self.min_length:
+                            continue
+                        elif self.max_length >= 1 and len(doc) > self.max_length:
+                            continue
+                        words = [w.text for w in doc]
+                        spaces = [bool(w.whitespace_) for w in doc]
+                        # We don't *need* an example here, but it seems nice to
+                        # make it match the Corpus signature.
+                        yield Example(doc, Doc(nlp.vocab, words=words, spaces=spaces))
```
adrianeboyd, line 321: I don't think there's any need to go through `words` and `spaces`?

```suggestion
                        # We don't *need* an example here, but it seems nice to
                        # make it match the Corpus signature.
                        yield Example(doc, doc.copy())
```

You could also consider using `nlp.make_doc()` again for something that looks a bit more duplicate/simple, but I guess with a slower tokenizer that's probably slightly slower.

spacy/training/corpus.py
```diff
@@ -58,6 +58,17 @@ def read_labels(path: Path, *, require: bool = False):
     return srsly.read_json(path)
 
 
+@util.registry.readers("spacy.PlainTextCorpus.v1")
+def create_plain_text_reader(
+    path: Optional[Path],
+    min_length: int = 0,
+    max_length: int = 0,
+) -> Callable[["Language"], Iterable[Doc]]:
+    if path is None:
```
shadeMe, line 67: Can we add a docstring to the entry point as well? I'm not fond of the duplication myself, but it'll allow for our VSCode extension to display the correct string on hover without having to do anything like parse the function body, locate the class being constructed and pulling the docstring from it.

website/docs/api/corpus.mdx
```diff
@@ -175,3 +175,68 @@ Yield examples from the data.
 | ---------- | -------------------------------------- |
 | `nlp`      | The current `nlp` object. ~~Language~~ |
 | **YIELDS** | The examples. ~~Example~~              |
+
+## PlainTextCorpus {id="plaintextcorpus",tag="class"}
```
shadeMe, line 179: We should add `v4` tags to the new additions like this one and the ones below.

spacy/training/corpus.py
danieldk: Done.

website/docs/api/corpus.mdx
danieldk: Added a version attribute to the class.

shadeMe (approved):

danieldk:
Completely forgot to stage the tests that I wrote, added now as well.

spacy/tests/training/test_corpus.py
```diff
@@ -0,0 +1,75 @@
...
+            if len(doc) >= min_length and (max_length == 0 or len(doc) <= max_length)
+        ]
+        reference, predicted = _examples_to_tokens(corpus(nlp))
+
+        assert reference == check
+        assert predicted == check
+
+
+@contextmanager
+def _string_to_tmp_file(s: str):
```
shadeMe, line 58: Return value type hint missing (for the sake of consistency).

shadeMe (commented):
Minor nitpick; LGTM otherwise!

spacy/training/corpus.py
```diff
@@ -257,3 +279,52 @@ def __call__(self, nlp: "Language") -> Iterator[Example]:
...
+
+        nlp (Language): The current nlp object.
+        YIELDS (Example): The example objects.
+
+        DOCS: https://spacy.io/api/corpus#plaintextcorpus-call
+        """
+        for loc in walk_corpus(self.path, ".txt"):
+            with open(loc, encoding="utf-8") as f:
+                for text in f:
+                    text = text.strip()
```
adrianeboyd, line 321: I think I'd just remove the final newline, not strip all whitespace?

spacy/tests/training/test_corpus.py
danieldk: Fixed.

spacy/training/corpus.py
danieldk: Changed to strip trailing carriage return/newline characters.

svlandeg (commented):
Apologies if this is a silly question, but why can't this just go in `master` instead of `v4`?

danieldk:
> Apologies if this is a silly question, but why can't this just go in `master` instead of `v4`?

That's a good question. I didn't need it there, but I could as well rebase it for `master`.

svlandeg:
Then we'd just update the tags to "3.5.1" and merge this after 3.5 is tagged/out

spacy/tests/training/test_corpus.py
```diff
@@ -0,0 +1,76 @@
...
+        ]
+        reference, predicted = _examples_to_tokens(corpus(nlp))
+
+        assert reference == check
+        assert predicted == check
+
+
+@contextmanager
+def _string_to_tmp_file(s: str) -> Generator[Path, None, None]:
+    with tempfile.TemporaryDirectory() as d:
```
adrianeboyd, line 59: Use our helper method `make_tempdir` instead. (I think there used to be more cross-platform compat issues than there are now, but at least for now this is used consistently in the test suite.)

spacy/tests/training/test_corpus.py
```diff
@@ -0,0 +1,76 @@
...
+        file_path = Path(d) / "string.txt"
+        with open(file_path, "wb") as f:
+            f.write(s.encode("utf-8"))
```
adrianeboyd, line 62: I realize the result is the same, but this feels like python 2.

```suggestion
        with open(file_path, "wb", encoding="utf-8") as f:
            f.write(s)
```

spacy/tests/training/test_corpus.py
```diff
@@ -0,0 +1,76 @@
...
+        "two",
+        "sentences",
+        ".",
+    ],
+    ["This", "is", "another", "doc", "."],
+    ["A", "third", "doc", "."],
+]
+
+
+@pytest.mark.parametrize("min_length, max_length", [(0, 0), (0, 5), (5, 0), (5, 5)])
```
adrianeboyd, line 38:
```suggestion
@pytest.mark.parametrize("min_length"), [0, 5])
@pytest.mark.parametrize("max_length"), [0, 5])
```

adrianeboyd (commented):
Just a few nitpicky comments on the tests...

spacy/tests/training/test_corpus.py
danieldk: Fixed (but `wb`, since we can't specify an encoding otherwise).

spacy/tests/training/test_corpus.py
danieldk: Fixed.

spacy/tests/training/test_corpus.py
adrianeboyd: `b` doesn't make sense? I'm sure you can specify an encoding with `w`.

adrianeboyd: Ah, I misunderstood.

PR #12125, opened 2023-01-19
Refactor lexeme mem passing

polm:
## Description

This changes the `_new_lexeme` method in the vocab to not take a memory pool as an argument. The passed value has not been used since #6809. 

~~I am having some local test failures, but they are related to my thinc version, and don't seem to have anything to do with the changes in this PR. (On review, this seems to be normal because Thinc is currently using a dev version.)~~ This is not an issue in CI.

### Types of change

minor refactor

## Checklist

### DIFF
--- a/spacy/vocab.pyx
+++ b/spacy/vocab.pyx
@@ -157,7 +157,7 @@ cdef class Vocab:
                                                   orth=key, orth_id=string))
             return lex
         else:
-            return self._new_lexeme(mem, string)
+            return self._new_lexeme(string)
 
     cdef const LexemeC* get_by_orth(self, Pool mem, attr_t orth) except NULL:
         """Get a pointer to a `LexemeC` from the lexicon, creating a new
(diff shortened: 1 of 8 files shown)


spacy/vocab.pyx
```diff
@@ -157,7 +157,7 @@ cdef class Vocab:
                                                   orth=key, orth_id=string))
             return lex
         else:
-            return self._new_lexeme(mem, string)
+            return self._new_lexeme(string)
```
shadeMe, line 160: Since this PR targets `v4`, I think it'd make sense to also propagate the change upward, i.e., remove the now-unused `mem` parameter from `Vocab.get` and update its callsites?

spacy/vocab.pyx
```diff
@@ -171,21 +171,10 @@ cdef class Vocab:
...
-        # memory use is to periodically reset the vocab, which is an action
-        # that should be up to the user to do (so we don't need to keep track
-        # of the doc ownership).
-        # TODO: Change the C API so that the mem isn't passed in here.
-        mem = self.mem
-        #if len(string) < 3 or self.length < 10000:
-        #    mem = self.mem
-        cdef bint is_oov = mem is not self.mem
-        lex = <LexemeC*>mem.alloc(1, sizeof(LexemeC))
+            return self._new_lexeme(self.strings[orth])
```
shadeMe, line 174: Same question as above.

polm:
Good point about removing the `mem` argument from the calling methods. The methods involved are used in many places, so the changes were pretty extensive, but I think this should be resovled now.

shadeMe (approved):

PR #12131, opened 2023-01-19
Add rehearsal_components config setting

thomashacker:
## Description

- This PR adds the `rehearse` functionality to the `training config` by providing a new variable `rehearse_components`.
- Adds the new setting to the `training schema`
- The new config setting is added to the docs, and the `Language.rehearse` and `TrainablePipe.rehearse` docs now mention that they need to be run together with `update`. 
- The `Language.rehearse` function was updated to only use the optimizer if provided, allowing us to use `finish_update`  later. 
- The `tagger.pyx` received a small fix, a check whether `sgd is None` before running `finish_update`

## Checklist


thomashacker:
I'm open to changes about the general naming and the new doc entries 😇 I've tested it with the current `Textcat.rehearse` experiment, and it provides the same/similar results, which indicates that it works. It will be helpful to have this for the other experiments without writing custom `rehearsal` loops and taking advantage of the existing training loop. 

One concern that I might have is that by using `update` and `rehearse`, we are using `self.model.begin_update(docs)` twice, which might have an impact on speed/memory. Let me know what you think and whether this addition makes sense 😄

website/docs/api/cli.mdx
```diff
@@ -1059,6 +1060,7 @@ $ python -m spacy train [config_path] [--output] [--code] [--verbose] [--gpu-id]
 | `--code`, `-c`    | Path to Python file with additional code to be imported. Allows [registering custom functions](/usage/training#custom-functions) for new architectures. ~~Optional[Path] \(option)~~                               |
 | `--verbose`, `-V` | Show more detailed messages during training. ~~bool (flag)~~                                                                                                                                                       |
 | `--gpu-id`, `-g`  | GPU ID or `-1` for CPU. Defaults to `-1`. ~~int (option)~~                                                                                                                                                         |
+| `--use_rehearse`, `-r`  | Use 'rehearsal' updates on a pre-trained model to address the catastrophic forgetting problem. Defaults to `False`. ~~bool (flag)~~                                                                                                                                                         |
```
svlandeg, line 1063: Let's remember to add a tag here from which spaCy version this option will become available (3.5.1?)

spacy/training/loop.py
```diff
@@ -214,6 +221,13 @@ def train_while_improving(
                 exclude=exclude,
                 annotates=annotating_components,
             )
+            if use_rehearse:
+                nlp.rehearse(
+                    subbatch,
+                    losses=losses,
+                    sgd=False,  # type: ignore[arg-type]
```
shadeMe, line 228:
```suggestion
                    sgd=None,
```

thomashacker: Ah, good catch!

website/docs/api/cli.mdx
thomashacker: Added the tags 3.5.1 to the new docs

spacy/training/loop.py
thomashacker: Fixed it for both update and rehearse

spacy/training/loop.py
```diff
@@ -217,17 +214,17 @@ def train_while_improving(
...
-                nlp.rehearse(
-                    subbatch,
-                    losses=losses,
-                    sgd=False,  # type: ignore[arg-type]
-                    exclude=exclude,
-                )
+            nlp.rehearse(
+                subbatch,
+                losses=losses,
+                sgd=None,  # type: ignore[arg-type]
```
shadeMe, line 224:
```suggestion
                sgd=None,
```
I think we can remove the `mypy` hint now?

shadeMe: I've seen this in a couple of other places too. My best guess: a compatibility hack to support both Python 2 and 3 (from the time we were supporting both).

website/docs/usage/training.mdx
```diff
@@ -575,6 +575,29 @@ now-updated model to the predicted docs.
...
+[components.ner]
+source = "en_core_web_sm"
+
+[training]
+rehearse_components = ["ner"]
+```
+
+<Infobox variant="warning" title="Loss calculation" id="rehearse-components-loss">
+
+Be aware that the loss is calculated by the sum of both the `update` and `rehearse` function.
```
shadeMe, line 595: I'm not sure I follow the "sum of both `update` and `rehearse`" phrase?

website/docs/usage/training.mdx
```diff
@@ -575,6 +575,29 @@ now-updated model to the predicted docs.
...
+source = "en_core_web_sm"
+
+[training]
+rehearse_components = ["ner"]
+```
+
+<Infobox variant="warning" title="Loss calculation" id="rehearse-components-loss">
+
+Be aware that the loss is calculated by the sum of both the `update` and `rehearse` function. 
+If both the loss and accuracy of the component increases over time, it can be caused due to the trained component making more different predictions that the inital model,
```
shadeMe, line 596: Why would an increase in accuracy indicate catastrophic forgetting?

kadarakos:
I know that I'm late to the party, but `rehearse` I feel like is an unlucky choice, because what I understand is that it means something different in the catastrophic forgetting sense. 

It means revisiting samples from a past dataset. Pseudo-rehearsal implies sampling inputs from some proposal distribution and consolidating the response between the old and new models to those samples. 

My understanding in terms of what `spaCy` is doing is kinda like self-distillation: given *only new data* make sure the new model's response is similar to the old one. I wonder if its a good idea to change the name now that we are making it more usable or should we just stick to the old one? Maybe I'm missing something, but this is my understanding and I was surprised when I saw the code.

website/docs/usage/training.mdx
thomashacker: **My understanding is that:**

If the accuracy of the model improves on a test data set, which has a similar domain as the training data, while the loss is also increasing, we can assume that the trained model is learning new data, but at the same time, the loss/difference between the teacher and student predictions is also getting higher, which might indicate that the model is forgetting.
I can be completely wrong here, but that's how I would interpret it 😇

thomashacker:
> I know that I'm late to the party, but the name `rehearse` I feel like is an unlucky choice, because what I understand is that it means something different in the catastrophic forgetting sense.
> 
> It means revisiting samples from a past dataset. Pseudo-rehearsal implies sampling inputs from some proposal distribution and consolidating the response between the old and new models to those samples.
> 
> My understanding in terms of what `spaCy` is doing is kinda like self-distillation: given _only new data_ make sure the new model's response is similar to the old one. I wonder if its a good idea to change the name now that we are making it more usable or should we just stick to the old one? Maybe I'm missing something, but this is my understanding and I was surprised when I saw the code.

What @kadarakos says is very interesting and I would say that now would be the best time to rename it 😇 What do you think @svlandeg @honnibal ?

website/docs/usage/training.mdx
thomashacker: I agree; This might be a bit unclear. 
What I meant was: "The loss is the sum of both the `update` loss and the `rehearse` loss."

website/docs/usage/training.mdx
```diff
@@ -575,6 +575,29 @@ now-updated model to the predicted docs.
 
 </Infobox>
 
+### Using rehearsing to address catastrophic forgetting {id="rehearse-components", tag="experimental", version="3.5.1"}
+
+Perform “rehearsal” updates to pre-trained components. Rehearsal updates teach the current component to make predictions similar to an initial model, to try to address the “catastrophic forgetting” problem. This feature is experimental.
```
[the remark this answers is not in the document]
thomashacker: Thanks! It's much better

spacy/training/initialize.py
```diff
@@ -67,6 +67,10 @@ def init_nlp(config: Config, *, use_gpu: int = -1) -> "Language":
         with nlp.select_pipes(enable=resume_components):
             logger.info(f"Resuming training for: {resume_components}")
             nlp.resume_training(sgd=optimizer)
+    # Components that shouldn't be updated during training
```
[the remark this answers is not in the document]
thomashacker: Nice catch! Thanks

spacy/language.py
```diff
@@ -1211,32 +1213,29 @@ def rehearse(
...
         for name, proc in pipes:
-            if name in exclude or not hasattr(proc, "rehearse"):
+            if (
+                name in exclude
+                or not hasattr(proc, "rehearse")
+                or name not in rehearse_components
+            ):
                 continue
-            grads = {}
             proc.rehearse(  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
thomashacker: I'll check, thanks for the catch

website/docs/api/language.mdx
```diff
@@ -346,13 +346,16 @@ and custom registered functions if needed. See the
 
 Perform a "rehearsal" update from a batch of data. Rehearsal updates teach the
 current model to make predictions similar to an initial model, to try to address
-the "catastrophic forgetting" problem. This feature is experimental.
+the "catastrophic forgetting" problem. Please note that `Language.rehearse` needs to be used together with `Language.update`. This feature is experimental.
```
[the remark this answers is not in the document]
thomashacker: That's a good question, and I like the idea! What do you think @svlandeg ?

spacy/default_config.cfg
```diff
@@ -84,6 +84,8 @@ score_weights = {}
 frozen_components = []
 # Names of pipeline components that should set annotations during training
 annotating_components = []
+# Names of pipeline components that should get rehearsed during training
+rehearse_components = []
```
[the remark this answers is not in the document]
thomashacker: Agree!

spacy/language.py
```diff
@@ -1211,32 +1213,29 @@ def rehearse(
...
-        for key, (W, dW) in grads.items():
-            sgd(key, W, dW)  # type: ignore[call-arg, misc]
+            if isinstance(sgd, Optimizer):
+                if (
+                    name not in exclude
+                    and isinstance(proc, ty.TrainableComponent)
+                    and proc.is_trainable
+                    and proc.model not in (True, False, None)
+                ):
+                    proc.finish_update(sgd)
```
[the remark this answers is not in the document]
thomashacker: Ah, so you mean `Language.rehearse` would never use `finish_update`? And that it shouldn't be triggered outside `TrainablePipe`?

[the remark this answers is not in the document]
shadeMe: Yeah, I'd just pass `sgd` to `proc.rehearse`.

website/docs/usage/training.mdx
kadarakos: Shouldn't we than report `rehearse` and `update` losses separately actually?

thomashacker: Yes, I think that would make more sense

website/docs/api/language.mdx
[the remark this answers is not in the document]
svlandeg: I don't have a strong opinion, but sure emphasizing this more makes sense!

thomashacker:
I'm thinking about the best way to implement the `rehearsal loss`. Right now, I've simply added a new key to the `losses` dict with `losses[self.name+"_rehearse"]` , if you specify the key in the `[training.score_weights]` it will show up but I think the fact that it's not showing automatically is bad usability? I think, best case would be to show everything with`* _rehearse` as suffix automatically, currently figuring out how to implement that. Nevertheless, I'm very open for any ideas and input 😇

svlandeg:
Closing for now, though we'll get back to this as part of the work on v4 later.

PR #12136, opened 2023-01-20
`Language.update`: ensure that tok2vec gets updated

danieldk:
## Description

**Preamble:** This it the first PR in probably 2-3 PRs to improve `Language.update`, but I thought it is best to split these up.

The components in a pipeline can be updated independently. However, tok2vec implementations are an exception to this, since they depend on listeners for their gradients. The update method of a tok2vec implementation computes the tok2vec forward and passes this along with a backprop function to the listeners. This backprop function accumulates gradients for all the listeners. There are two ways in which the accumulated gradients can be used to update the tok2vec weights:

1. Call the `finish_update` method of tok2vec *after* the `update` method is called on all of the pipes that use a tok2vec listener.
2. Pass an optimizer to the `update` method of tok2vec. In this case, tok2vec will give the last listener a special backprop function that calls `finish_update` on the tok2vec.

Unfortunately, `Language.update` did neither of these. Instead, it immediately called `finish_update` on every pipe after `update`. As a result, the tok2vec weights are updated when no gradients have been accumulated from listeners yet. And the gradients of the listeners are only used in the next call to `Language.update` (when `finish_update` is called on tok2vec again).

This change fixes this issue by applying the approach outlined in (1). This is chosen over (2), because (2) would perform sequential learning for the shared parameters and not true multi-task learning. Secondly, it is slower.

The main updating loop in `Language.update` is also simplified by using the `TrainableComponent` protocol consistently.

### Types of change

Bugfix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1150,17 +1150,12 @@ class Language:
             component_cfg[name].setdefault("drop", drop)
             pipe_kwargs[name].setdefault("batch_size", self.batch_size)
         for name, proc in self.pipeline:
-            # ignore statements are used here because mypy ignores hasattr
-            if name not in exclude and hasattr(proc, "update"):
-                proc.update(examples, sgd=None, losses=losses, **component_cfg[name])  # type: ignore
-            if sgd not in (None, False):
-                if (
-                    name not in exclude
-                    and isinstance(proc, ty.TrainableComponent)
-                    and proc.is_trainable
-                    and proc.model not in (True, False, None)
-                ):
-                    proc.finish_update(sgd)
+            if (
+                name not in exclude
+                and isinstance(proc, ty.TrainableComponent)
+                and proc.is_trainable
+            ):
+                proc.update(examples, sgd=sgd, losses=losses, **component_cfg[name])
             if name in annotates:
                 for doc, eg in zip(
                     _pipe(
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1155,7 +1155,7 @@ class Language:
                 and isinstance(proc, ty.TrainableComponent)
                 and proc.is_trainable
             ):
-                proc.update(examples, sgd=sgd, losses=losses, **component_cfg[name])
+                proc.update(examples, sgd=None, losses=losses, **component_cfg[name])
             if name in annotates:
                 for doc, eg in zip(
                     _pipe(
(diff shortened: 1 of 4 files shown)


spacy/language.py
```diff
@@ -1150,17 +1150,12 @@ def update(
...
-                    name not in exclude
-                    and isinstance(proc, ty.TrainableComponent)
-                    and proc.is_trainable
-                    and proc.model not in (True, False, None)
-                ):
-                    proc.finish_update(sgd)
+            if (
+                name not in exclude
+                and isinstance(proc, ty.TrainableComponent)
+                and proc.is_trainable
```
shadeMe, line 1156: Under what circumstances is a `TrainableComponent`/`TrainablePipe` not trainable, i.e, `is_trainable` is `False`? There are no instances of pipes returning `False` in the codebase, as far as I can tell.

shadeMe (commented):
GitHub won't let me post a review if this comment field is empty :confounded:

spacy/language.py
danieldk: Yeah, I was going back and forth on this as well. I was trying to think about cases where something is a `TrainableComponent` but is not trainable. The primary thing I could come up is a component that is used only during rehearsal or distillation, like our layer distiller. Such components could implement `TrainableComponent` by virtue of deriving from `TrainablePipe`, but are not trainable.

shadeMe (approved):

danieldk:
Changed the PR to use approach (1). This is chosen over (2), because (2) would perform sequential learning for the shared parameters and not true multi-task learning. Secondly, it is slower.

shadeMe (approved):

spacy/language.py
```diff
@@ -1173,6 +1168,14 @@ def update(
                     examples,
                 ):
                     eg.predicted = doc
+        for name, proc in self.pipeline:
```
svlandeg, line 1171: Maybe add a comment here to explain why it's a separate loop?

svlandeg (approved):
Nice catch, Daniël. I propose to leave a quick comment in the code, but otherwise this is good to merge.

spacy/language.py
danieldk: Ah, yeah, good idea, helps avoiding that the loops get fused again.

PR #12145, opened 2023-01-23
Add alignment_mode argument to Span.char_span()

itssimon:
## Description
Adds the `alignment_mode` parameter present in `Doc.char_span()` to `Span.char_span()` as well.

### Types of change
Enhancement

## Checklist

### DIFF
--- a/spacy/tokens/span.pyx
+++ b/spacy/tokens/span.pyx
@@ -362,7 +362,7 @@ cdef class Span:
         result = xp.dot(vector, other.vector) / (self.vector_norm * other.vector_norm)
         # ensure we get a scalar back (numpy does this automatically but cupy doesn't)
         return result.item()
-    
+
     cpdef np.ndarray to_array(self, object py_attr_ids):
         """Given a list of M attribute IDs, export the tokens to a numpy
         `ndarray` of shape `(N, M)`, where `N` is the length of the document.
@@ -639,7 +639,7 @@ cdef class Span:
         else:
             return self.doc[root]
 
-    def char_span(self, int start_idx, int end_idx, label=0, kb_id=0, vector=None, id=0):
+    def char_span(self, int start_idx, int end_idx, label=0, kb_id=0, vector=None, alignment_mode="strict", id=0):
         """Create a `Span` object from the slice `span.text[start : end]`.
 
         start (int): The index of the first character of the span.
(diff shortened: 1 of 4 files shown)


spacy/tokens/span.pyx
```diff
@@ -639,7 +639,7 @@ cdef class Span:
         else:
             return self.doc[root]
 
-    def char_span(self, int start_idx, int end_idx, label=0, kb_id=0, vector=None, id=0):
+    def char_span(self, int start_idx, int end_idx, label=0, kb_id=0, vector=None, alignment_mode="strict", id=0):
```
adrianeboyd, line 642: I see that you were trying to match the order in `Doc.char_span`, but if anyone has been using positional args here, adding a new arg in the middle can break existing code:

```suggestion
    def char_span(self, int start_idx, int end_idx, label=0, kb_id=0, vector=None, id=0, alignment_mode="strict"):
```

This result is frustrating for the `Doc` vs. `Span` API (well, for people using positional args), but we can put it on our list of things to clean up in the API for v4.0.

spacy/tokens/span.pyx
```diff
@@ -649,11 +649,16 @@ cdef class Span:
...
+            boundaries. Options: "strict" (character indices must be aligned
+            with token boundaries), "contract" (span of all tokens completely
+            within the character span), "expand" (span of all tokens at least
+            partially covered by the character span). Defaults to "strict".
         RETURNS (Span): The newly constructed object.
         """
         start_idx += self.c.start_char
         end_idx += self.c.start_char
-        return self.doc.char_span(start_idx, end_idx, label=label, kb_id=kb_id, vector=vector)
+        return self.doc.char_span(start_idx, end_idx, label=label, kb_id=kb_id, vector=vector, alignment_mode=alignment_mode)
```
adrianeboyd, line 661: Adding a note for myself to follow up with a PR that adds `span_id` here.

adrianeboyd (commented):
Thanks for the PR!

This also needs to be added to the API docs under `website/docs/api/span.mdx`. Add a tag in the Name column that this is new in 3.5.1 (`<Tag variant="new">3.5.1</Tag>`) and note that the docs are formatted using prettier, since you'll probably need to reformat the whole table.

spacy/tokens/span.pyx
itssimon: The `id` parameter is undocumented and not used in the function body. Wonder why it's there to start with. Still wanna put it before `alignment_mode`?

adrianeboyd: `id` should be passed as `span_id` but isn't. I would still append `alignment_mode`. It's not a huge issue and we can straighten it out for v4.

adrianeboyd:
Thanks again for the PR!

PR #12155, opened 2023-01-24
Add basic tests for project commands

polm:
## Description

This adds basic tests for all the project commands. They are minimal and don't cover all options or usage patterns, and so they still need to be expanded, which is why this is in draft.

The DVC command is considered deprecated and is not included.

### Types of change

adding tests

## Checklist


spacy/tests/test_cli_app.py
```diff
@@ -101,12 +104,23 @@ def test_project_run(project_dir):
 
 
 def test_project_clone():
-    with make_tempdir() as workspace:
+    # Git clones create some readonly/hidden files. On Windows, trying to
+    # delete these gives an error, so we have to handle that specially.
```
adrianeboyd, line 108: Would it make sense to add this to `make_tempdir` instead?

polm: Sure, that makes sense. In the narrow case of git clones it's clear why this is happening. In the more general case it could maybe mask a real issue, though I guess for `make_tempdir` it wouldn't matter.

polm: It turns out the changes here didn't fix the issue because it's in the `clone` command, not the test, anyway, so moving it to the general implementation fixes that.

spacy/util.py
```diff
@@ -1032,9 +1033,18 @@ def make_tempdir() -> Generator[Path, None, None]:
     YIELDS (Path): The path of the temp directory.
     """
     d = Path(tempfile.mkdtemp())
```
adrianeboyd, line 1035: Maybe this is a carry-over from python 2 and we should switch to `tempfile.TemporaryDirectory` or similar?

spacy/util.py
```diff
@@ -1032,9 +1033,18 @@ def make_tempdir() -> Generator[Path, None, None]:
...
+
+    # On Windows, attempts to delete a git directory fail with permissions
+    # errors due to read only / hidden files. This mainly addresses that issue.
+    def _fix_perms(func, path, exc_info):
+        # IWUSR = give user write permission
+        os.chmod(path, stat.S_IWUSR)
```
danieldk, line 1041: This should probably check first that the path is not writable, so that we can immediately re-raise the exception on other issues.

adrianeboyd: Your previous version with `tempfile.TemporaryDirectory` worked correctly in this case, right? Or did you not use that with git at that point?

spacy/util.py
danieldk: But then we'd always leave junk around on Windows since `tempfile.TemporaryDirectory` only seems to allow ignoring errors, not providing a function to override the behavior? Or maybe I missed it.

spacy/util.py
adrianeboyd: Hmm, I had some hope that something here would work correctly in windows?

spacy/util.py
adrianeboyd: Hmm, I had some hope that it might actually work correctly?

spacy/util.py
adrianeboyd: I agree that this looks like it's going to mask other `PermissionError` cases?

spacy/util.py
danieldk: Cool, it seems to do so:

https://github.com/python/cpython/blob/498598e8c2d64232d26c075de87c513415176bbf/Lib/tempfile.py#L867

I'll try it in the other PR.

adrianeboyd: If it works in the other PR, let's move those changes to a separate PR?

polm: It looks like it works in Python 3.8 and higher, and <=3.7 will need a workaround:

https://github.com/python/cpython/issues/70847

adrianeboyd: ugh

polm: The change is small, so I've modified it to use TemporaryDirectory so we can check it works as documented here as well. If we do it in another PR we can just merge it in here.

danieldk: Didn't work in the other PR due to the issue that you pointed out above (<=3.7). So, do we want the uglier version or maybe target spaCy v4, since we are probably going to drop 3.7?

polm: This is already targeting v4, actually, so I suggest we just leave it as it is now. (The tests also confirmed it works except for 3.7 Windows.)

polm:
Assuming we make this rely on dropping <=3.7, this depends on #12187.

spacy/util.py
polm: This has been removed with the latest changes, so marking this resolved.

polm:
This is still waiting on the other PR, but now the tests should cover most options of CLI commands. 

One issue that's come up is that sparse checkout works without issue locally, but fails across the board in CI - for some reason the checkout simply doesn't work and says to make sure access is OK. Not sure why that would happen, so disabling the sparse test only for now.

spacy/util.py
```diff
@@ -32,6 +32,7 @@
 import pkgutil
 import logging
 import socket
+import stat
```
adrianeboyd, line 35:
```suggestion
```

polm:
Since #12187 was merged and the tests passed, this should be ready for review. There is a question of how extensive the tests need to be, and whether this should be more thorough or not, but it is a good start at least.

I will try the sparse checkout test again, but the issue there seems to be with CI configuration rather than a bug or issue with the test as such, so I might need help with that.

spacy/tests/test_cli_app.py
```diff
@@ -89,3 +91,136 @@ def test_debug_data_trainable_lemmatizer_cli(en_vocab):
...
+
+@pytest.fixture
+def project_dir():
+    with make_tempdir() as pdir:
+        (pdir / "project.yml").write_text(SAMPLE_PROJECT_TEXT)
+        yield pdir
+
+
+def test_project_document(project_dir):
+    readme_path = project_dir / "README.md"
```
svlandeg, line 145: Perhaps `assert` here that the file does not exist yet at this point

spacy/tests/test_cli_app.py
```diff
@@ -89,3 +91,136 @@ def test_debug_data_trainable_lemmatizer_cli(en_vocab):
...
+        app, ["project", "document", str(project_dir), "-o", str(readme_path)]
+    )
+    assert result.exit_code == 0
+    assert readme_path.is_file()
+    text = readme_path.read_text("utf-8")
+    assert SAMPLE_PROJECT["description"] in text
+
+
+def test_project_assets(project_dir):
+    asset_dir = project_dir / "assets"
```
svlandeg, line 156: Same here - `assert` that this dir is empty before running `assets`

spacy/util.py
```diff
@@ -1032,12 +1032,12 @@ def make_tempdir() -> Generator[Path, None, None]:
...
-    d = Path(tempfile.mkdtemp())
-    yield d
+
     try:
-        shutil.rmtree(str(d))
+        with tempfile.TemporaryDirectory() as td:
+            yield Path(td)
     except PermissionError as e:
-        warnings.warn(Warnings.W091.format(dir=d, msg=e))
+        warnings.warn(Warnings.W091.format(dir=td, msg=e))
```
svlandeg, line 1040: Can you make a separate PR with this change + explanation?

svlandeg (commented):
This is great to see, thanks Paul! I'm sure that there's other options and variants to test, but this is a great start. I had a few small comments and would like to pull out the change to `util.py` in a separate PR, but otherwise this looks good!

svlandeg (commented):
Why does this target `v4`? We want to be able to move to `weasel` from 3.5.x onwards, and we'd need to have these tests available on the 3.x branch when we make that switch...

polm:
Thanks for the feedback! 

About v4 - I think I may have initially based this on v4 due to a misunderstanding of what "after 3.5" meant, but the temporary directory change makes this a little more complicated. `tempfile.TemporaryDirectory` only works the way we want in Python 3.8+, see [here](https://github.com/explosion/spaCy/pull/12155#discussion_r1086466701). In <=3.7 there will be issues with Git checkouts on Windows that will cause tests to fail.

I will go ahead and split the TemporaryDirectory change into its own PR targeting v4. If we want to make a similar change for v3, we'll need to figure out a different approach if we want to support the older Pythons.

spacy/util.py
polm: Here's the PR 

https://github.com/explosion/spaCy/pull/12285

polm:
Changing the base branch was a mess, so closing in favor of #12303.

PR #12157, opened 2023-01-24
Normalize whitespace in evaluate CLI output test

adrianeboyd:
## Description

Depending on terminal settings, it looks like lines may be padded to the screen width so the comparison is too strict with only the command string replacement.

### Types of change

Tests.

## Checklist

### DIFF
--- a/spacy/tests/test_cli_app.py
+++ b/spacy/tests/test_cli_app.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from typer.testing import CliRunner
 
 from spacy.cli._util import app
-from .util import make_tempdir, assert_strings_equal_except_ws
+from .util import make_tempdir, normalize_whitespace
 
 
 def test_convert_auto():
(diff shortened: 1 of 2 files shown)


spacy/tests/test_cli_app.py
```diff
@@ -37,6 +38,8 @@ def test_benchmark_accuracy_alias():
     # Verify that the `evaluate` alias works correctly.
     result_benchmark = CliRunner().invoke(app, ["benchmark", "accuracy", "--help"])
     result_evaluate = CliRunner().invoke(app, ["evaluate", "--help"])
-    assert result_benchmark.stdout == result_evaluate.stdout.replace(
-        "spacy evaluate", "spacy benchmark accuracy"
+    assert re.sub(r"\s+", " ", result_benchmark.stdout) == re.sub(
+        r"\s+",
+        " ",
+        result_evaluate.stdout.replace("spacy evaluate", "spacy benchmark accuracy"),
```
svlandeg, line 44: Could we have this as a utility function in `tests\util.py` ?

adrianeboyd: I don't see a whole lot of advantage in moving that to a util method?

svlandeg: it's just that the `re.sub` command is copy/pasted in the current implementation, making it brittle to future changes IMO

adrianeboyd: I don't really care...

PR #12182, opened 2023-01-25
CI: Add black formatting check to validation

adrianeboyd:
## Description

Add black formatting check to validation.

### Types of change

CI.

## Checklist


adrianeboyd:
We have avoided having this check in the past because we were worried that it could be off-putting for new users making small contributions (in particular for users making small contributions through the github web editors), but I think we're at the point where it would make sense to require this for all PRs.

svlandeg (approved):

PR #12186, opened 2023-01-26
Fix even more color problems

essenmitsosse:
### Types of change
- Fixes some remaining color problems introduced by #12178
- ~removes some dead code~
 
## Checklist


svlandeg (commented):
Is "removing dead code" as part of a bugfix PR a good idea? (I worry about the potential to introduce unrelated, new bugs)

essenmitsosse:
@svlandeg yeah, probably a good point. Removed it and will create a separate PR

ines:
Yeah, are you sure we don't need the Jupyter classes anymore? Those are for styling the output of interactive code widgets.

PR #12187, opened 2023-01-26
Drop python 3.6/3.7, remove unneeded compat

adrianeboyd:
## Description

Drop python 3.6/3.7, remove unneeded compat.

### Types of change

?

## Checklist


adrianeboyd:
There is one place where `spacy-legacy` uses the compat import, so this requires a v4.0.0.dev0 release of `spacy-legacy`: https://github.com/explosion/spacy-legacy/pull/38, https://github.com/explosion/spacy-legacy/pull/40

spacy/pipeline/spancat.py
```diff
@@ -1,12 +1,11 @@
 from typing import List, Dict, Callable, Tuple, Optional, Iterable, Any, cast
-from typing import Union
+from typing import Union, Literal, Protocol, runtime_checkable
 from thinc.api import Config, Model, get_current_ops, set_dropout_rate, Ops
 from thinc.api import Optimizer
 from thinc.types import Ragged, Ints2d, Floats2d
 
 import numpy
 
-from ..compat import Protocol, runtime_checkable
```
polm, line 9: Looks like you don't need Literal here?

polm (commented):
Looks good to me except for one extra import.

Separately, this reminded me that we can use `shlex.join` now instead of `spacy.util.join_command`. That function itself will go away when the #11378 successor is merged, but might as well update it for now.

adrianeboyd:
This PR isn't intended to include all the updates and refactoring that would be possible for 3.8+, just to drop 3.6/3.7 from the setup/CI and remove code that is only related to <3.8 compatibility.

svlandeg (approved):

PR #12188, opened 2023-01-26
Don't re-download installed models

polm:
## Description

When downloading a model, this checks if the same version of the same model is already installed. If it is then the download is skipped.

This is necessary because pip uses the final download URL for its caching feature, but because of the way models are hosted on Github, their URLs change every few minutes, so pip is unable to cache downloads correctly. This is basically a workaround for that caching issue. 

### Types of change

minor enhancement

## Checklist

### DIFF
--- a/spacy/cli/download.py
+++ b/spacy/cli/download.py
@@ -3,11 +3,12 @@ import requests
 import sys
 from wasabi import msg
 import typer
+import srsly
 
 from ._util import app, Arg, Opt, WHEEL_SUFFIX, SDIST_SUFFIX
 from .. import about
 from ..util import is_package, get_minor_version, run_command
-from ..util import is_prerelease_version
+from ..util import is_prerelease_version, get_installed_models, get_package_path
 
 
 @app.command(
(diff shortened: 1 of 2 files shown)


polm:
https://github.com/pypa/pip/issues/10694 is the related pip issue.

spacy/cli/download.py
```diff
@@ -63,6 +64,16 @@ def download(
         compatibility = get_compatibility()
         version = get_version(model_name, compatibility)
 
+    # If we already have this version installed, skip downloading
```
adrianeboyd, line 67: I think it might make sense to get the version from `importlib` instead?

polm: Good point, that ends up simpler.

adrianeboyd: Sorry for editing the comment, a new comment would have been clearer. I think you can just use `get_package_version` directly?

polm: Ah, I see what you mean - done.

svlandeg (approved):
Looks good to me, but I'll leave this open for Adriane to give a final review!

adrianeboyd:
I'm trying to think if there's a situation where you need a `--force` flag. There shouldn't really be because you should always get these models from us and you can also always use `pip uninstall`, but I'm not sure whether it would occur to users to use `pip uninstall` in this case.

polm:
I had similar thoughts about a `--force` flag. I can add a note when skipping that `pip uninstall` could be used, though I don't think it will be necessary often...?

adrianeboyd:
I think we could try without and only add `--force` or a longer message if people end up complaining about it. I bet it's rare.

adrianeboyd:
This is also currently kind of hard to test. You could add an additional CI test, but they're all disabled for v4 because there aren't published models.

PR #12190, opened 2023-01-26
Remove dead code and rename components for docs

essenmitsosse:
## Description
Removes unused code, after reimplementing the MyBinder integration with JupyterLabs. This also removes everything referencing [Juniper](https://github.com/ines/juniper), since that is no longer used after migrating to Next.

### Types of change
- Remove dead code in docs implementation
- Rename things in docs implementation

## Checklist

### DIFF
--- a/website/docs/styleguide.mdx
+++ b/website/docs/styleguide.mdx
@@ -420,8 +420,8 @@ matches = matcher(doc)
 
 Adding `{executable="true"}` to the title turns the code into an executable
 block, powered by [Binder](https://mybinder.org) and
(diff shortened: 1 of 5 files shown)


ines:
I'm not sure this change is that useful – while it's not using my `juniper` package directly, it's still the same implementation that I wrote for it and that I published a similar version of under that name. So if people are interested in how it's done, this is a better reference than just JupyterLab (which people typically associate with the notebook tool, not the underlying APIs). Plus, it gives people more details, including other implementations it's based on / inspired by.

svlandeg:
Sounds reasonable Ines, so let's just keep it.

PR #12193, opened 2023-01-27
Update to `spacy-legacy` v4, remove Python 3.6/3.7 support, reenable `spacy.TransitionBasedParser.v1` test

danieldk:
## Description

Update to `spacy-legacy` dependency to `4.0.0.dev0`.  Reenable `spacy.TransitionBasedParser.v1` test, which we can do now that this function is operational in `spacy-legacy` again. Remove Python 3.6 and 3.7 support, we can't build against those versions anymore, because we have already removed support in `spacy-legacy`.

### Types of change

Chore

## Checklist

### DIFF
--- a/.github/azure-steps.yml
+++ b/.github/azure-steps.yml
@@ -26,7 +26,6 @@ steps:
   - script: |
       python -m mypy spacy
     displayName: 'Run mypy'
-    condition: ne(variables['python_version'], '3.6')
 
   - task: DeleteFiles@1
     inputs:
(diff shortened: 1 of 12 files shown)


adrianeboyd:
Please just wait on #12187 instead.

danieldk:
> Please just wait on #12187 instead.

Closing this, but maybe you could re-enable the `spacy.TransitionBasedParser.v1` in the other PR (since it's only a one-liner, it probably doesn't make sense to do a separate PR for that).

adrianeboyd:
I think it's fine as a separate PR!

PR #10714, opened 2022-04-27
Allow assets to be optional in spacy project

rmitsch:
## Goals

Allow assets in spaCy projects to be downloaded optionally.

## Description

- Assets can be marked as `optional`. If `optional` is not set, it is assumed to be `False`. Example:
```yaml
assets:
  - dest: "assets/${vars.treebank}"
    optional: True
    git:
      repo: "https://github.com/UniversalDependencies/${vars.treebank}"
      branch: "master"
      path: ""
```
- New argument for `spacy project assets`: `--all` (`-a`). If this is set, all assets - including those marked as `optional` - are downloaded. Otherwise `optional` assets are ignored.

### Types of change

New feature

## Checklist

- [x] I confirm that I have the right to submit this contribution under the project's MIT license.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #12194, opened 2023-01-27
Add extra flag to assets docs

polm:
## Description

The `project assets` command has had an `--extra` flag for a while now, but it isn't in the docs. This adds it.

### Types of change

docs fix

## Checklist


svlandeg (commented):
Thanks! Did we introduce this right at the start, or from a given spaCy version?

polm:
It looks like it was added in #10714, which was released in [3.3.1](https://github.com/explosion/spaCy/releases/tag/v3.3.1).

PR #12201, opened 2023-01-30
Add the configuration schema for distillation

danieldk:
## Description

This change adds the configuration schema for distillation, as well as the default configuration and some tests. The schema will be used by the training loop and `distill` subcommand.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/spacy/cli/init_config.py
+++ b/spacy/cli/init_config.py
@@ -83,7 +83,7 @@ def init_fill_config_cli(
     # fmt: off
     base_path: Path = Arg(..., help="Path to base config to fill", exists=True, dir_okay=False),
     output_file: Path = Arg("-", help="Path to output .cfg file (or - for stdout)", allow_dash=True),
-    distillation: bool = Opt(False, "--distillation", "-d", help="Include config for distillation (with 'spacy distill')"),
+    distillation: bool = Opt(False, "--distillation", "-dt", help="Include config for distillation (with 'spacy distill')"),
     pretraining: bool = Opt(False, "--pretraining", "-pt", help="Include config for pretraining (with 'spacy pretrain')"),
     diff: bool = Opt(False, "--diff", "-D", help="Print a visual diff highlighting the changes"),
     code_path: Optional[Path] = Opt(None, "--code-path", "--code", "-c", help="Path to Python file with additional code (registered functions) to be imported"),
--- a/spacy/cli/init_config.py
+++ b/spacy/cli/init_config.py
@@ -8,7 +8,7 @@ import re
 from jinja2 import Template
 
 from .. import util
-from ..language import DEFAULT_CONFIG_PRETRAIN_PATH
+from ..language import DEFAULT_CONFIG_DISTILL_PATH, DEFAULT_CONFIG_PRETRAIN_PATH
 from ..schemas import RecommendationSchema
 from ..util import SimpleFrozenList
 from ._util import init_cli, Arg, Opt, show_validation_error, COMMAND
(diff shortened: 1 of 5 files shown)


spacy/cli/init_config.py
```diff
@@ -83,6 +83,7 @@ def init_fill_config_cli(
     # fmt: off
     base_path: Path = Arg(..., help="Path to base config to fill", exists=True, dir_okay=False),
     output_file: Path = Arg("-", help="Path to output .cfg file (or - for stdout)", allow_dash=True),
+    distillation: bool = Opt(False, "--distillation", "-pt", help="Include config for distillation (with 'spacy distill')"),
```
svlandeg, line 86: Needs a different shortcut, `-pt` is c/p from `pretraining`

spacy/schemas.py
```diff
@@ -405,13 +405,35 @@ class Config:
...
+    class Config:
+        extra = "forbid"
+
+
+class ConfigSchemaDistill(BaseModel):
+    # fmt: off
+    batcher: Batcher = Field(..., title="Batcher for the training data")
+    corpus: StrictStr = Field(..., title="Path in the config to the distillation data")
+    dropout: StrictFloat = Field(..., title="Dropout rate")
+    max_epochs: StrictInt = Field(..., title="Maximum number of steps to distill for")
```
svlandeg, line 418:
```suggestion
    max_epochs: StrictInt = Field(..., title="Maximum number of epochs to distill for")
```

svlandeg (commented):
Few small comments.

Also, can we add more tests around this to check that the (default) `distillation` config runs correctly?

spacy/cli/init_config.py
danieldk: Not sure how I missed that. Changed to `-d`.

spacy/schemas.py
danieldk: Fixed.

spacy/cli/init_config.py
svlandeg: You don't think `-d` and `-D` are too similar? For only a small set of options it seems unfortunate to reuse the same single letter maybe?

svlandeg: maybe `-dt`? 🤷

svlandeg (approved):

spacy/cli/init_config.py
danieldk: My unix brain can't parse multi-letter single dash option 😄 😓 , but changed...

svlandeg: Yea I get that! Sorry 🙈

spacy/schemas.py
```diff
@@ -405,13 +405,35 @@ class Config:
...
+
+class ConfigSchemaDistill(BaseModel):
+    # fmt: off
+    batcher: Batcher = Field(..., title="Batcher for the training data")
+    corpus: StrictStr = Field(..., title="Path in the config to the distillation data")
+    dropout: StrictFloat = Field(..., title="Dropout rate")
+    max_epochs: StrictInt = Field(..., title="Maximum number of epochs to distill for")
+    max_steps: StrictInt = Field(..., title="Maximum number of steps to distill for")
+    optimizer: Optimizer = Field(..., title="The optimizer to use")
+    pipe_map: Dict[str, str] = Field(..., title="Mapping from teacher to student pipe")
```
shadeMe, line 421: The mapping is reversed in the [code](https://github.com/danieldk/spaCy/blob/e36234d70a0e981ed624bac9e0282847c7468ebc/spacy/language.py#L1048), i.e., student pipe -> teacher pipe. I think it'd be best to keep it the same in both the config and the code so as to avoid subtle bugs.

danieldk: 👍 good catch. Also renamed this to `student_to_teacher` like @svlandeg suggested in the other PR.

shadeMe (approved):

PR #12202, opened 2023-01-30
Add displaCy data structures to docs

thomashacker:
## Description

This PR adds a new section to the `api/top-level#functions` documentation for displaCy: `Visualizer data structures`. The section shows examples of the three data structures (`Dependency Parsing`, `Entity Recognition`, and `Span Classification`) and their types formatted as tables. To reduce cluttering, the nested dictionary types are formatted as a collapsable table.

The idea for this PR initially came from this discussion https://github.com/explosion/spaCy/pull/10950
The types are based on [this gist](https://gist.github.com/pmbaumgartner/85b3a0256131e86fbdd78a8b4060d361) by @pmbaumgartner  

### Types of change

Documentation

## Checklist

### DIFF
--- a/website/docs/api/top-level.mdx
+++ b/website/docs/api/top-level.mdx
@@ -343,7 +343,7 @@ use with the `manual=True` argument in `displacy.render`.
 ### Visualizer data structures {id="displacy_structures"}
 
 You can use displaCy's data format to manually render data. This can be useful
-if you want to visualize output from other libaries. You can find examples of
+if you want to visualize output from other libraries. You can find examples of
 displaCy's different data formats below.
 
 > #### DEP example data structure
--- a/website/docs/api/top-level.mdx
+++ b/website/docs/api/top-level.mdx
@@ -397,11 +397,10 @@ displaCy's different data formats below.
 > #### ENT example data structure
 >
 > ```json
->  {
->    "text": "But Google is starting from behind.",
->    "ents": [{"start": 4, "end": 10, "label": "ORG"}],
->    "title": None
->  }
+> {
+>   "text": "But Google is starting from behind.",
+>   "ents": [{ "start": 4, "end": 10, "label": "ORG" }]
+> }
 > ```
 
 #### Named Entity Recognition data structure {id="structure-ent"}
@@ -410,20 +409,20 @@ displaCy's different data formats below.
 | -------------- | ------------------------------------------------------------------------------------------- |
 | `text`         | String representation of the document text. ~~str~~                                         |
 | `ents`         | List of dictionaries describing entities (see structure below). ~~List[Dict[str, Any]]~~    |
-| `title`        | Title of the visualization. ~~str~~                                                         |
 | _Optional_     |                                                                                             |
+| `title`        | Title of the visualization. ~~Optional[str]~~                                               |
 | `settings`     | Entity Visualizer options (see [here](/api/top-level#displacy_options)). ~~Dict[str, Any]~~ |
 
 <Accordion title="Ents data structure">
 
-| Dictionary Key | Description                                         |
-| -------------- | --------------------------------------------------- |
-| `start`        | The index of the first token of the entity. ~~int~~ |
-| `end`          | The index of the last token of the entity. ~~int~~  |
-| `label`        | Label attached to the entity. ~~str~~               |
-| _Optional_     |                                                     |
-| `kb_id`        | `KnowledgeBase` ID. ~~str~~                         |
-| `kb_url`       | `KnowledgeBase` URL. ~~str~~                        |
+| Dictionary Key | Description                                                            |
+| -------------- | ---------------------------------------------------------------------- |
+| `start`        | The index of the first character of the entity. ~~int~~                |
+| `end`          | The index of the last character of the entity. (not inclusive) ~~int~~ |
+| `label`        | Label attached to the entity. ~~str~~                                  |
+| _Optional_     |                                                                        |
+| `kb_id`        | `KnowledgeBase` ID. ~~str~~                                            |
+| `kb_url`       | `KnowledgeBase` URL. ~~str~~                                           |
 
 </Accordion>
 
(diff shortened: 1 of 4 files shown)


website/docs/api/top-level.mdx
```diff
@@ -340,6 +340,128 @@ use with the `manual=True` argument in `displacy.render`.
 | `options`   | Span-specific visualisation options. ~~Dict[str, Any]~~             |
 | **RETURNS** | Generated entities keyed by text (original text) and ents. ~~dict~~ |
 
+### Visualizer data structures {id="displacy_structures"}
+
+You can also use displaCy's data format to manually render data. This can be
+useful if you want to visualize output from other libaries. You can find
+examples of displaCy's data format on the
+[usage page](/usage/visualizers#manual-usage).
```
pmbaumgartner, line 348: Do you think we should still link to the Usage page when the same examples are now in the sidebar? I think it might be more confusing to a user that they would be reading and click this link to be brought away from the documentation.

website/docs/api/top-level.mdx
```diff
@@ -340,6 +340,128 @@ use with the `manual=True` argument in `displacy.render`.
...
+| `spans`        | List of spans. ~~List[Dict[str, Any]]~~   |
+| `title`        | Title of the visualization. ~~str~~       |
+| `tokens`       | List of tokens. ~~List[str]~~             |
+| `settings`     | Visualization options. ~~Dict[str, Any]~~ |
+
+<Accordion title="Span data structure">
+
+| Dictionary Key | Description                |
+| -------------- | -------------------------- |
+| `start`        | Start index. ~~int~~       |
```
pmbaumgartner, line 455: This is more of a general comment we should check across all keys and descriptions: I think the descriptions should match their equivalent data structure description in the spaCy API. For example, in the `Span` [docs](https://spacy.io/api/span) we describe the `start` attribute as "The index of the first token of the span."

website/docs/api/top-level.mdx
```diff
@@ -340,6 +340,128 @@ use with the `manual=True` argument in `displacy.render`.
...
+| `end`          | End index. ~~int~~         |
+| `start_token`  | Start token. ~~int~~       |
+| `end_token`    | End token. ~~int~~         |
+| `label`        | Label of the span. ~~str~~ |
+| `kb_id`        | Knowledgebase ID. ~~str~~  |
```
pmbaumgartner, line 460:
```suggestion
| `kb_id`        | `KnowledgeBase` ID. ~~str~~  |
```

Should we style this like this and maybe link to the [KnowledgeBase](https://spacy.io/api/kb) docs as well?

pmbaumgartner (changes requested):
My in-line code comments are a little out of order so let me summarize them in a comment:

- I don't think we should link back to the Usage docs since we provide those examples in the sidebar
- Can we find a better way to make it clear the accordions have information about the sub-structures in each data structure?
- We should be consistent with the descriptions for the values/attributes where they exist in the spaCy API.

website/docs/api/top-level.mdx
thomashacker: Ah, yes! I agree

website/docs/api/top-level.mdx
thomashacker: Ah, I didn't realize! Thanks, I agree!

website/docs/api/top-level.mdx
```diff
@@ -340,6 +340,128 @@ use with the `manual=True` argument in `displacy.render`.
...
+
+#### Dependency Visualizer data structure {id="structure-dep"}
+
+| Dictionary Key | Description                               |
+| -------------- | ----------------------------------------- |
+| `words`        | List of words. ~~List[Dict[str, Any]]~~   |
+| `arcs`         | List of arcs. ~~List[Dict[str, Any]]~~    |
+| `settings`     | Visualization options. ~~Dict[str, Any]~~ |
+
+<Accordion title="Word data structure">
```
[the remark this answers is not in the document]
thomashacker: I agree, it might need more clarification and the naming needs to be clearer! I'll change it

website/docs/api/top-level.mdx
thomashacker: I agree! I'll remove the link

thomashacker: Removed the link 😇

website/docs/api/top-level.mdx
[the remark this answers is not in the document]
thomashacker: I've added `see structure below` . I think it makes it more clearer now

website/docs/api/top-level.mdx
thomashacker: I tried to stay consistent with the code base now, but at some attributes, I'm still a bit unsure.

website/docs/api/top-level.mdx
thomashacker: Done!

website/docs/api/top-level.mdx
```diff
@@ -340,6 +340,127 @@ use with the `manual=True` argument in `displacy.render`.
...
+
+<Accordion title="Ents data structure">
+
+| Dictionary Key | Description                                         |
+| -------------- | --------------------------------------------------- |
+| `start`        | The index of the first token of the entity. ~~int~~ |
+| `end`          | The index of the last token of the entity. ~~int~~  |
+| `label`        | Label attached to the entity. ~~str~~               |
+| `kb_id`        | `KnowledgeBase` ID. ~~str~~                         |
+| `kb_url`       | `KnowledgeBase` URL. ~~str~~                        |
```
pmbaumgartner, line 423: I think these are both not required, so we might want to type as `Optional` to be consistent. However, this isn't exactly right because we're using `Optional` here to represent that the key doesn't need to be present, not that the type of the value is `Union[str, None]`.

Maybe we want another way to indicate the key is optional rather than the type. Like placing `(optional)` after the key name in the `Dictionary Key` column?

This is one area it would be good to have feedback on. I'm not sure if there's another place in the docs we have the same issue and what the practice is.  cc @svlandeg

svlandeg: This is basically similar to kwd-args that are optional but not necessarily of type `Optional`. We denote those with a specific divider, cf https://spacy.io/api/language#init for example - so perhaps do something similar to that?

thomashacker: This is great! Thanks, I'll try to add them with a divider

thomashacker: I've added the `Optional` divider for the `KnowledgeBase` parameters and also for the `setting` parameter.

thomashacker:
@pmbaumgartner What was the reason for removing the docs label from this PR? 😢

pmbaumgartner:
> @pmbaumgartner What was the reason for removing the `docs` label from this PR? 😢

I must have misclicked or used a keyboard shortcut accidentally. Sorry!

svlandeg:
@pmbaumgartner: is your "changes requested" review still valid? (as discussed internally we don't usually use this feature as typically we'd go by the open review comments to determine whether something requires changes or not, but because you used it earlier I want to double check with you and not just dismiss the review as that looks weird)

pmbaumgartner:
> @pmbaumgartner: is your "changes requested" review still valid? (as discussed internally we don't usually use this feature as typically we'd go by the open review comments to determine whether something requires changes or not, but because you used it earlier I want to double check with you and not just dismiss the review)

I wasn't aware this was still blocking. I dismissed the review, but haven't had time to review the recent changes yet, so I'll look again today.

pmbaumgartner:
In the `ENT example data structure`, `title` has a value of `None`. Shouldn't its actual type be `Optional[str]` then then? Same for Spans?

And it looks like `title` isn't a field on the `deps` data, only `ents` and `spans`, is that right? I think we might want to open an issue for that, if that's the case.

thomashacker:
- Adjusted the type of the `title` key in the docs
- Added tests that check `displaCy.render` with manual data
- Implemented `title` key to dependency visualizations and adjusted docs

pmbaumgartner:
👌 love the tests. I don't have feedback on anything else at this time.

svlandeg:
Closing in favour of #12875 (PR redone to target `master`)

PR #12209, opened 2023-01-31
Language.distill: copy both reference and predicted

danieldk:
## Description

In distillation we also modify the teacher docs (e.g. in tok2vec components), so we need to copy both the reference and predicted doc.

Problem caught by @shadeMe.

### Types of change

Bugfix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1059,7 +1059,7 @@ class Language:
             return losses
 
         validate_distillation_examples(examples, "Language.distill")
-        examples = _copy_examples(examples)
+        examples = _copy_examples(examples, copy_x=True, copy_y=True)
 
         if sgd is None:
             if self._optimizer is None:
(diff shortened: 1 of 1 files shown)


spacy/language.py
```diff
@@ -2328,13 +2328,18 @@ def restore(self) -> None:
         self[:] = []
 
 
-def _copy_examples(examples: Iterable[Example]) -> List[Example]:
+def _copy_examples(
+    examples: Iterable[Example], copy_x: bool = True, copy_y: bool = False
```
shadeMe, line 2332: Can we make the new parameters keyword-only (if only to eliminate one potential source of confusion)?

spacy/language.py
```diff
@@ -2328,13 +2328,18 @@ def restore(self) -> None:
...
     """Make a copy of a batch of examples, copying the predicted Doc as well.
     This is used in contexts where we need to take ownership of the examples
     so that they can be mutated, for instance during Language.evaluate and
     Language.update.
     """
-    return [Example(eg.x.copy(), eg.y) for eg in examples]
+    return [
+        Example(eg.x.copy() if copy_x else eg.x, eg.y.copy() if copy_y else eg.y)
+        for eg in examples
+    ]
```
shadeMe, line 2342: I wonder if we should emit a warning here if we do a no-op, i.e., both `copy_x` and `copy_y` are `False`. I can't think of a reason why the user would call the function at all in such a case, but it feels weird to have a function that basically goes `def do_something(really_do_something: bool = True):`?

spacy/language.py
danieldk: Dones.

spacy/language.py
danieldk: It's an internal function. Besides that, I don't think it's a no-op since it clears things like cached alignments.

PR #12213, opened 2023-02-01
Backslash fixes in docs

svlandeg:
## Description
Found two cases where `\_` was rendered incorrectly and only `_` will do because it's in backticks.

### Types of change
docs fix

## Checklist


svlandeg:
I double checked the Netlify preview and the fixes seem right, so merging :-)

PR #12214, opened 2023-02-01
Reimplement distillation with oracle cut size

danieldk:
## Description

The code for distillation with an oracle cut size was not reimplemented after the parser refactor. We did not notice, because we did not have tests for this functionality. This change brings back the functionality and adds this to the parser tests.

Also add extra checks to `_parse_batch` to avoid segfaults that I encountered during implementation when feeding the function wrong inputs.

### Types of change

Enhancement or bugfix :)

## Checklist

### DIFF
--- a/spacy/ml/tb_framework.pyx
+++ b/spacy/ml/tb_framework.pyx
@@ -249,9 +249,11 @@ cdef list _parse_batch(CBlas cblas, TransitionSystem moves, StateC** states,
     cdef np.ndarray step_actions
 
     scores = []
-    while sizes.states >= 1:
+    while sizes.states >= 1 and (actions is None or actions):
         step_scores = numpy.empty((sizes.states, sizes.classes), dtype="f")
         step_actions = actions[0] if actions is not None else None
+        assert step_actions is None or step_actions.size == sizes.states, \
+            f"number of step actions ({step_actions.size}) must equal number of states ({sizes.states})"
         with nogil:
             _predict_states(cblas, &activations, <float*>step_scores.data, states, &weights, sizes)
             if actions is None:
--- a/spacy/ml/tb_framework.pyx
+++ b/spacy/ml/tb_framework.pyx
@@ -249,7 +249,7 @@ cdef list _parse_batch(CBlas cblas, TransitionSystem moves, StateC** states,
     cdef np.ndarray step_actions
 
     scores = []
-    while sizes.states >= 1 and (actions is None or actions):
+    while sizes.states >= 1 and (actions is None or len(actions) > 0):
         step_scores = numpy.empty((sizes.states, sizes.classes), dtype="f")
         step_actions = actions[0] if actions is not None else None
         assert step_actions is None or step_actions.size == sizes.states, \
--- a/spacy/pipeline/transition_parser.pyx
+++ b/spacy/pipeline/transition_parser.pyx
@@ -795,7 +795,7 @@ def _states_diff_to_actions(
         c_state_before = before_state.c
         c_state_after = after_state.c
 
-        assert equal(c_state_before.history.begin(), c_state_before.history.begin(),
+        assert equal(c_state_before.history.begin(), c_state_before.history.end(),
             c_state_after.history.begin())
 
     actions = []
--- a/spacy/pipeline/transition_parser.pyx
+++ b/spacy/pipeline/transition_parser.pyx
@@ -782,7 +782,7 @@ def _states_diff_to_actions(
     """
     Return for two sets of states the actions to go from the first set of
     states to the second set of states. The histories of the first set of
-    states must be prefix of the second set of states.
+    states must be a prefix of the second set of states.
     """
     cdef StateClass before_state, after_state
     cdef StateC* c_state_before
--- a/spacy/pipeline/transition_parser.pyx
+++ b/spacy/pipeline/transition_parser.pyx
@@ -795,8 +795,7 @@ def _states_diff_to_actions(
         c_state_before = before_state.c
         c_state_after = after_state.c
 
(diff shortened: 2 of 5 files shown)


spacy/pipeline/transition_parser.pyx
```diff
@@ -757,3 +774,46 @@ def states2actions(states: List[StateClass]) -> List[Ints1d]:
...
     return actions
+
+def _states_diff_to_actions(
+    before_states: List[StateClass],
+    after_states: List[StateClass]
+) -> List[Ints1d]:
+    """
+    Return for two sets of states the actions to go from the first set of
+    states to the second set of states. The histories of the first set of
+    states must be prefix of the second set of states.
```
danieldk, line 785:
```suggestion
    states must be a prefix of the second set of states.
```

spacy/ml/tb_framework.pyx
```diff
@@ -249,9 +249,11 @@ cdef list _parse_batch(CBlas cblas, TransitionSystem moves, StateC** states,
     cdef np.ndarray step_actions
 
     scores = []
-    while sizes.states >= 1:
+    while sizes.states >= 1 and (actions is None or actions):
         step_scores = numpy.empty((sizes.states, sizes.classes), dtype="f")
         step_actions = actions[0] if actions is not None else None
+        assert step_actions is None or step_actions.size == sizes.states, \
```
shadeMe, line 255: Would it be easy to come up with a straight-forward (failing) test to catch this assertion? Given how the violation of this invariant can lead to segfaults, it might not be a bad idea to have a test or two for it.

spacy/pipeline/transition_parser.pyx
```diff
@@ -642,7 +647,7 @@ class Parser(TrainablePipe):
                     raise ValueError(Errors.E149) from None
         return self
 
-    def _init_batch(self, teacher_step_model, docs, max_length):
+    def _init_batch(self, teacher_pipe, docs, max_length):
```
shadeMe, line 650: Given that this method is only used by the `distill` method, could we have a more appropriate name like `_init_batch_from_teacher`?

spacy/pipeline/transition_parser.pyx
```diff
@@ -757,3 +774,46 @@ def states2actions(states: List[StateClass]) -> List[Ints1d]:
...
+    assert len(before_states) == len(after_states)
+
+    # Check invariant: before states histories must be prefixes of after states.
+    for before_state, after_state in zip(before_states, after_states):
+        c_state_before = before_state.c
+        c_state_after = after_state.c
+
+        assert equal(c_state_after.history.begin(),
+            c_state_after.history.begin() + c_state_before.history.size(),
+            c_state_after.history.begin())
```
shadeMe, line 800: That should be `c_state_before.history.begin()`, right? 

I also wonder if we should assert the lengths of the before and after histories before we do any kind of pointer/iterator arithmetic on them. Maybe I'm just being paranoid, but we can't guarantee that `c_state_after.history.begin() + c_state_before.history.size()` won't trigger a segfault when it's dereferenced, e.g: when `c_state_after.history` is smaller than `c_state_before.history` and is weirdly placed at a boundary bordering a page that doesn't have the read permissions.

spacy/ml/tb_framework.pyx
danieldk: Added. I would also like to fix the segfault that occurs when feeding the wrong actions. But that requires changes to `TransitionSystem`, for which I'll make a PR after this one (to avoid that this PR becomes too unfocused).

spacy/tests/parser/test_model.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.fixture
+def nlp_parser():
+    nlp = English()
+    parser = nlp.add_pipe("parser")
+
+    train_examples = []
+    for text, annotations in TRAIN_DATA:
+        train_examples.append(Example.from_dict(nlp.make_doc(text), annotations))
+        for dep in annotations.get("deps", []):
```
shadeMe, line 34:
```suggestion
        for dep in annotations["deps"]:
```
Since `deps` is defined for all training instances.

shadeMe (approved):
Just one more minor comment.

spacy/ml/tb_framework.pyx
```diff
@@ -249,9 +249,11 @@ cdef list _parse_batch(CBlas cblas, TransitionSystem moves, StateC** states,
...
+            f"number of step actions ({step_actions.size}) must equal number of states ({sizes.states})"
```
svlandeg, line 256: Why do you propose an `assert` with a fail message here? If this is user-facing, we'd rather have an exception + custom error message.

spacy/ml/tb_framework.pyx
```diff
@@ -249,9 +249,11 @@ cdef list _parse_batch(CBlas cblas, TransitionSystem moves, StateC** states,
     cdef np.ndarray step_actions
 
     scores = []
-    while sizes.states >= 1:
+    while sizes.states >= 1 and (actions is None or actions):
```
svlandeg, line 252: I'm finding the combined condition hard to read and somewhat implicit. Perhaps 
```
actions is None or len(actions)>0
``` 
to make it more explicit that we're guarding against an empty list?

spacy/ml/tb_framework.pyx
danieldk: Discussed yesterday with @svlandeg , both assertions are to catch programming errors.

spacy/ml/tb_framework.pyx
danieldk: Fixed.

PR #12215, opened 2023-02-01
Have logging calls use string formatting types

svlandeg:
## Description
After internal discussion, we decided to stick to format string templates when using the `logger`. The advantage here is that if calling `str()` on a variable is expensive, it should only be done when the log line is actually getting printed at a given log level.

I'm very sad to lose the readability of our beloved f-strings, but efficiency of the code base is still number one priority.

### Types of change
enhancement + docs update

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -173,6 +173,11 @@ formatting and [`flake8`](http://flake8.pycqa.org/en/latest/) for linting its
 Python modules. If you've built spaCy from source, you'll already have both
 tools installed.
 
+As a general rule of thumb, we use f-strings for any formatting of strings.
+One exception are calls to Python's `logging` functionality.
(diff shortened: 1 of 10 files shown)


shadeMe:
Is the performance hit of f-strings really that noticeable? The locations that are being changed don't seem to be part of a hot path, i.e., they all seem like one-off log statements that happen at program/training initialization.

svlandeg:
> Is the performance hit of f-strings really that noticeable? The locations that are being changed don't seem to be part of a hot path, i.e., they all seem like one-off log statements that happen at program/training initialization.

While going through the changes I didn't encounter any cases where I felt like the performance was taking a significant hit with f-strings, no.

That said this is the policy we want to implement, after internal discussions, and we want to have a clear rule to follow also to avoid future potential performance hits.

PR #12220, opened 2023-02-03
Use black version constraints from requirements.txt

adrianeboyd:
## Description

Use `black` version constraints from `requirements.txt`.

Does not yet handle pre-commit hooks.

### Types of change

Chore.

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,4 +37,4 @@ types-mock>=0.1.1
 types-setuptools>=57.0.0
 types-requests
 types-setuptools>=57.0.0
-black>=22.0,<23.0
+black==22.3.0
--- a/.github/workflows/autoblack.yml
+++ b/.github/workflows/autoblack.yml
@@ -16,7 +16,7 @@ jobs:
         with:
             ref: ${{ github.head_ref }}
       - uses: actions/setup-python@v4
-      - run: pip install black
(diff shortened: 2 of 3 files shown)


adrianeboyd:
We can potentially consider getting rid of the autoblack step at this point, but it doesn't solve the pre-commit hook version problem yet.

requirements.txt
```diff
@@ -37,4 +37,4 @@ types-mock>=0.1.1
 types-setuptools>=57.0.0
 types-requests
 types-setuptools>=57.0.0
-black>=22.0,<23.0
+black==22.3.0
```
rmitsch, line 40: Why do we pin to 22.3.0 now?

adrianeboyd: The point here is more to have the same thing across the board, so this is reflecting the actual CI and pre-commit checks that were currently running.

The actual pin that we want could be up for discussion (also moving to the newest release), but is a separate question from syncing this better for the future.

rmitsch: > The point here is more to have the same thing across the board, so this is reflecting the actual CI and pre-commit checks that were currently running.

Yes, I understood this. I was just curious about why this particular version instead of the previous range, but it's not central.

rmitsch (approved):

PR #12227, opened 2023-02-03
Adjust return type of `registry.find`

thomashacker:
## Description

Adjusts the return type of `registry.find` from `Callable` to `Dict[str, Optional[Union[str, int]]]` and code documentation so that it matches with `catalogue.registry.find`. 

### Types of change

bug fix

## Checklist

### DIFF
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -252,7 +252,7 @@ def get_third_party_dependencies(
                     raise regerr from None
             module_name = func_info.get("module")  # type: ignore[attr-defined]
             if module_name:  # the code is part of a module, not a --code file
-                modules.add(func_info["module"].split(".")[0])  # type: ignore
+                modules.add(func_info["module"].split(".")[0])  # type: ignore[union-attr]
     dependencies = []
     for module_name in modules:
         if module_name in distributions:
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -243,10 +243,10 @@ def get_third_party_dependencies(
         for func_name in func_names:
             # Try the lang-specific version and fall back
             try:
-                func_info = util.registry.find(reg_name, lang + "." + func_name)
+                func_info = util.registry.find(reg_name, lang + "." + func_name)  # type: ignore
             except RegistryError:
                 try:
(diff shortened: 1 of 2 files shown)


spacy/cli/package.py
```diff
@@ -252,7 +252,7 @@ def get_third_party_dependencies(
                     raise regerr from None
             module_name = func_info.get("module")  # type: ignore[attr-defined]
             if module_name:  # the code is part of a module, not a --code file
-                modules.add(func_info["module"].split(".")[0])  # type: ignore[index]
+                modules.add(func_info["module"].split(".")[0])  # type: ignore
```
rmitsch, line 255: Is there a specific `ignore` directive we can use here? Why does `[index]` not apply anymore?

spacy/util.py
```diff
@@ -144,8 +144,17 @@ def get(cls, registry_name: str, func_name: str) -> Callable:
         return func
 
     @classmethod
-    def find(cls, registry_name: str, func_name: str) -> Callable:
-        """Get info about a registered function from the registry."""
+    def find(
+        cls, registry_name: str, func_name: str
+    ) -> Dict[str, Optional[Union[str, int]]]:
+        """Find the information about a registered function, including the
```
rmitsch, line 150:
```suggestion
        """Find information about a registered function, including the
```

rmitsch (commented):
Minor questions/suggestions, but LGTM.

spacy/cli/package.py
thomashacker: It's throwing an `[union-index]` error because of the newly added `Union[str,int,None]`, I've adjusted the directive. 😄

rmitsch: Thanks for checking!

rmitsch: @thomashacker Wait, the code says `[union-attr]`, you mention `[union-index]` in your comment. Just making sure that `[union-attr]` is the correct one - is it?

thomashacker: Oh sorry! I mixed them up somehow. I meant `[union-attr]` 😄

rmitsch (approved):

svlandeg (approved):
Good catch, thanks!

PR #12234, opened 2023-02-06
Add Language.pipe_as_tuples

adrianeboyd:
## Description

As part of the transition to v4, add `Language.pipe_as_tuples()` and deprecate `Language.pipe(as_tuples=True)`.

### Types of change

Enhancement?

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1598,7 +1598,6 @@ class Language:
             context = doc._context
             doc._context = None
             yield (doc, context)
-        return
 
     def _has_gpu_model(self, disable: Iterable[str]):
         for name, proc in self.pipeline:
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1516,6 +1516,7 @@ class Language:
         DOCS: https://spacy.io/api/language#pipe
         """
         if as_tuples:
+            warnings.warn(Warnings.W125, DeprecationWarning)
             texts = cast(Iterable[Tuple[Union[str, Doc], _AnyContext]], texts)
             docs_with_contexts = (
                 self._ensure_doc_with_context(text, context) for text, context in texts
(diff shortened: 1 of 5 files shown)


spacy/language.py
```diff
@@ -1574,6 +1575,31 @@ def pipe(  # noqa: F811
...
+            batch_size=batch_size,
+            disable=disable,
+            n_process=n_process,
+            component_cfg=component_cfg,
+        )
+        for doc in docs:
+            context = doc._context
+            doc._context = None
+            yield (doc, context)
+        return
```
shadeMe, line 1601:
```suggestion

```

website/docs/usage/processing-pipelines.mdx
```diff
@@ -88,11 +88,12 @@ have to call `list()` on it first:
...
 </Infobox>
 
-You can use the `as_tuples` option to pass additional context along with each
-doc when using [`nlp.pipe`](/api/language#pipe). If `as_tuples` is `True`, then
-the input should be a sequence of `(text, context)` tuples and the output will
-be a sequence of `(doc, context)` tuples. For example, you can pass metadata in
-the context and save it in a [custom attribute](#custom-components-attributes):
+You can use the [`nlp.pipe_as_tuples`](/api/language#pipe_as_tuples) method to
+pass additional context along with each doc when using the functionality of
+[`nlp.pipe`](/api/language#pipe). The input should be a sequence of
```
svlandeg, line 93: The "when using the functionality of `nlp.pipe`" bit of the sentence doesn't quite parse anymore to readers who might not be aware that the same function is still called under the hood.

adrianeboyd:
Looking at this again, I still kind of feel like we should just remove `as_tuples` entirely, since you could remove `Doc._context` and just let people use custom extensions instead.

Maybe we can just mark it as deprecated in v3.6 and provide a link to an update guide/FAQ? It has been possible since v3.2.

svlandeg:
> Looking at this again, I still kind of feel like we should just remove `as_tuples` entirely, since you could remove `Doc._context` and just let people use custom extensions instead.
> 
> Maybe we can just mark it as deprecated in v3.6 and provide a link to an update guide/FAQ? It has been possible since v3.2.

I'm a bit on the fence too. On the one hand, I can see how this would be useful to a lot of people. On the other hand, I don't like hardcoding the extension name.

One more alternative we could consider is not having this as a feature or alternative of `pipe`, but rather in the `make_doc` method. It feels more appropriate there, no? And then you run `make_doc` on your texts and you can still process the results with `pipe`. But in this case, I'd almost want to be able to customize the name of the `Doc` extension attribute.

honnibal:
I don't think we can get rid of the `as_tuples` mechanism entirely.

People need to be able to do this without manipulating the state on the `Doc` themselves. It feels pretty bad to have to set attributes just to do this. I also worry about the complexity of the rewrite we ask people to do: it needs to be easy to migrate code that used `as_tuples=True`.

Btw, in terms of the implementation of this, did you try making `Language.pipe` delegate to `Language.pipe_as_tuples`, rather than the other way around? This would make the implementation of `Language.pipe` simpler which would be nice. On the other hand I can see that there would be error handling logic and such that would end up duplicated that way.

svlandeg:
Closing as we're not deprecating `as_tuples` for now.

PR #12242, opened 2023-02-07
Remove Vocab.oov_prob, fix lexeme_ lookups table creation

adrianeboyd:
## Description

Remove `Vocab.oov_prob` and `Vocab.cfg` (both unused) and fix how `lexeme_prob` and `lexeme_cluster` tables are added dynamically when an attribute is set on a vocab where these tables don't already exist.

### Types of change

?

## Checklist


spacy/tests/vocab_vectors/test_lexeme.py
```diff
@@ -22,16 +22,21 @@ def test_issue600():
...
-def test_vocab_lexeme_lt(en_vocab, text1, text2, prob1, prob2):
-    """More frequent is l.t. less frequent"""
-    lex1 = en_vocab[text1]
-    lex1.prob = prob1
-    lex2 = en_vocab[text2]
-    lex2.prob = prob2
-
-    assert lex1 < lex2
-    assert lex2 > lex1
+@pytest.mark.parametrize("word1,prob1,word2,prob2", [("NOUN", -1, "opera", -2)])
```
adrianeboyd, line 25: The original test did not make sense. `lex1 < lex2` is comparing `ORTH` and not `PROB`. Replaced with a test for the lookups table creation.

adrianeboyd:
Actually, I'm not sure this is helpful for `lex_attr_getters` because those are only for the values in the struct. It is a mess at this point!

honnibal:
You'll maybe hate me for saying this because it's indeed been pretty broken / messy. But:

I don't feel great about having the `OOV_PROB` be a constant value. This is probably pedantic because nobody really makes these unigram frequency distributions anymore, but what we're really trying to do here is say if `n=0` in some sample, what probability should we assume that the item has of occurring? The value that it's sensible to say for this will depend on how many items we've sampled. (As a probably irrelevant technical note, in the early versions of the lexeme data these probabilities were estimated with Simple Good Turing estimation, so the probability was a projection based on the probability of items with count 2 and the probability of items with count 1).

Is there a version of this that lets us keep the oov probability as a value in the vocab, but handles it correctly?

svlandeg:

PR #12251, opened 2023-02-08
Use the same tuple in Span cmp and hash

adrianeboyd:
## Description

Use the same tuple in Span cmp and hash.

### Types of change

?

## Checklist


adrianeboyd:
To deal with the current situation where all `Span` attributes are writable, this adds`SpanC.start` and `SpanC.end` to the tuple. I think we need to do something about the writing/syncing between `start`/`start_char` separately because it is not a good situation overall.

PR #12257, opened 2023-02-08
Make Span.char_span optional args keyword-only

adrianeboyd:
## Description

Make `Span.char_span` optional args keyword-only

### Types of change

?

## Checklist


adrianeboyd:
There were inconsistencies in the names `start` vs. `start_idx`. Since `Doc.char_span` also uses `start_idx`, I updated the docs here in addition.

Actually, we'd need to do this in `Doc`, too, wait a sec.

svlandeg:
> Actually, I'm not sure we want to do this. I think many people are using `Doc.char_span(3, 15, "LABEL").

Hmm, true, though as soon as you start getting into the KB ID stuff etc, it really is better and more safe to use kwd-args. So maybe include `label` in the non-kwd-only-args?

adrianeboyd:
I'm still not sure in this particular case that it is worth being this breaking. There was even a case in `Doc`.

svlandeg:
It feels pretty arbitrary to me to have `kb_id` as the fourth positional argument, and by introducing the kwd-args in v4 we'd allow ourself to make changes to the list without being breaking going forward, as well enforcing downstream code to be more specific and thus less error-prone, so I personally think that's a good trade-off.

PR #12258, opened 2023-02-08
Update CLI to use radicli

ines:
## Description
This PR updates the spaCy v4 CLI to use [`radicli`](https://github.com/explosion/radicli). I've also used the opportunity to do some minor tidying up and removing CLI wrapper functions that were not needed anymore (since we can now just define regular reusable Python functions for the CLI), as well as converting `Enum`s to `Literal`s where possible. I've also updated the CLI tests and added a new one to ensure docstrings and help arguments.

~~One outstanding ticket is supporting descriptions for non-existent parent commands with subcommands (e.g. `debug`, `benchmark`). Still need to add that.~~

Also still needs testing and re-running all the commands to make sure they work as expected.

**Edit:** For the type checking, it looks like we need to stub out the `Path`-subtypes properly 😭 Otherwise we'll keep getting errors here.

### Types of change
enhancement, CLI

## Checklist


spacy/cli/package.py
```diff
@@ -54,8 +64,8 @@ def package_cli(
         name=name,
         version=version,
         create_meta=create_meta,
-        create_sdist=create_sdist,
-        create_wheel=create_wheel,
+        create_sdist="sdist" in build and "none" not in build,
+        create_wheel="sdist" in build and "none" not in build,
```
adrianeboyd, line 68: Should these both be the same?

ines: Oh no, sorry, this was a copy-paste mistake!

spacy/cli/init_config.py
```diff
@@ -1,55 +1,64 @@
...
-    lang: str = Opt(InitValues.lang, "--lang", "-l", help="Code of the language to use"),
-    pipeline: str = Opt(",".join(InitValues.pipeline), "--pipeline", "-p", help="Comma-separated names of trainable pipeline components to include (without 'tok2vec' or 'transformer')"),
-    optimize: Optimizations = Opt(InitValues.optimize, "--optimize", "-o", help="Whether to optimize for efficiency (faster inference, smaller model, lower memory consumption) or higher accuracy (potentially larger and slower model). This will impact the choice of architecture, pretrained weights and related hyperparameters."),
-    gpu: bool = Opt(InitValues.gpu, "--gpu", "-G", help="Whether the model can run on GPU. This will impact the choice of architecture, pretrained weights and related hyperparameters."),
-    pretraining: bool = Opt(InitValues.pretraining, "--pretraining", "-pt", help="Include config for pretraining (with 'spacy pretrain')"),
-    force_overwrite: bool = Opt(InitValues.force_overwrite, "--force", "-F", help="Force overwriting the output file"),
+    output_file=Arg(help="File to save the config to or - for stdout (will only output config and no additional logging info)"),
```
svlandeg, line 45: Wouldn't it be nicer to have a convertor function here that returns a `Path` also for the `"-"` option, so the rest of the code can be agnostic to it? In `radicli` there's the function `convert_path_or_dash` but it returns a `Union[Path, str]` instead of converting to Path in both cases, why is that?

ines: It would make the typing more consistent but it kinda felt weird because `Path("-")` is not really a valid path, it's kinda just a `Literal["-"]` symbol to indicate stdin/stdout.

(I also briefly considered having a converter that already opens the file, but this can get problematic, because we often want different handling depending on the file type vs. stdin/stdout. Like, if it's a JSON file, load as JSON, if it's JSONL, load as JSONL, if it's stdin/stdout, treat it as JSONL etc.)

svlandeg:
Adding this to our internal v4 backlog board and closing until we have time to pick this up again :-)

Issue #12024, opened 2022-12-24
Doc span group spans aren't adjusted for retokenization


kinghuang:
When a [Doc](https://spacy.io/api/doc) object is retokenized, the entity spans in [Doc.ents](https://spacy.io/api/doc#ents) reflect the new token alignment, but the spans in [Doc.spans](https://spacy.io/api/doc#spans) retain the original doc's token indexes, leading to unexpected spans.

## How to reproduce the behaviour

Here is a modification of the example for [Retokenizer.split](https://spacy.io/api/doc#spans) that stores `NewYork` as an entity and in a span group before retokenization.

**Split Example Code:**

```python
import spacy
from spacy.tokens import Span

nlp = spacy.blank("en")

doc = nlp("I live in NewYork")
ny = Span(doc, 3, 4, label="CITY")
doc.ents = [ny]
doc.spans["spans"] = [ny]

with doc.retokenize() as retokenizer:
    heads = [(doc[3], 1), doc[2]]
    attrs = {"POS": ["PROPN", "PROPN"],
             "DEP": ["pobj", "compound"]}
    retokenizer.split(doc[3], ["New", "York"], heads=heads, attrs=attrs)
```

**Output:**

```python
>>> print(doc.ents)
(NewYork,)
>>> print(doc.spans["spans"])
[New]
```

Notice how the entity span is still `NewYork`, but the span group span is now just `New`.

In the case of a merge, spans can be lost altogether by exceeding the range of the retokenized doc. Here is a modification of the example of [Retokenizer.merge](https://spacy.io/api/doc#retokenizer.merge).

**Merge Example Code:**

```python
doc = nlp("I like David Bowie")
db = Span(doc, 2, 4, label="PERSON")
doc.ents = [db]
doc.spans["spans"] = [db]

with doc.retokenize() as retokenizer:
    attrs = {"LEMMA": "David Bowie"}
    retokenizer.merge(doc[2:4], attrs=attrs)

print(doc.spans["spans"])
print(doc.ents)
```

**Output:**

```python
>>> print(doc.ents)
(David Bowie,)
>>> print(doc.spans["spans"])
[]
```

The range of the original `David Bowie` span is no longer valid and disappears from the span group.

## Your Environment

* Operating System: macOS Ventura 13.1 (22C65)
* Python Version Used: 3.11.0
* spaCy Version Used: 3.4.4
* Environment Information:

polm:
[...]

PR #12268, opened 2023-02-10
Enforce that Span.start/end(_char) remain valid and in sync

adrianeboyd:
## Description

Allowing span attributes to be writable starting in v3 has made it possible for the internal `Span.start/end/start_char/end_char` to get out-of-sync or have invalid values.

This checks that the values are valid and syncs the token and char offsets if any attributes are modified directly. It does not yet handle the case where the underlying doc is modified.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
This is a step towards addressing #12024.

It is currently strict about the whitespace for an `end_char`, so it doesn't accept the end char that includes in the offset, which is also what `Doc.char_span` does.

svlandeg (approved):
Nice work!

svlandeg:
Can be merged once the conflicts are resolved :-)

PR #12272, opened 2023-02-10
Matcher performance fix for extension predicates: use shared key function

kwhumphreys:
## Description

v3.5 introduced a performance regression for Matcher rules which use predicates with extension attributes.
Cause was found to be that cache key used to search for extension predicates had a different format from the cache key generated by the predicates themselves. Fix is to use a single method to generate all cache keys.

### Types of change

performance bug fix
discussion at https://github.com/explosion/spaCy/discussions/12261

## Checklist

### DIFF
--- a/spacy/matcher/matcher.pyx
+++ b/spacy/matcher/matcher.pyx
@@ -828,7 +828,7 @@ def _get_attr_values(spec, string_store):
     return attr_values
 
 
-def _predicate_cache_key(attr, predicate, value, regex=False, fuzzy=None):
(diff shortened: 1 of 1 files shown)


kwhumphreys:
existing tests for extension predicates seem comprehensive, but do not test performance.
any plans/ideas for testing the predicate cache?

adrianeboyd:
Thanks for the PR! This looks like a good solution. We'll try to have a v3.5.1 release with this fix soon.

We have vague plans for a suite of regular performance tests to catch regressions, but don't have anything concrete set up yet.

PR #12283, opened 2023-02-15
Recommend lookups tables from URLs or other loaders

adrianeboyd:
## Description

Shift away from the `lookups` extra (which isn't removed, just no longer mentioned) and recommend loading data from the `spacy-lookups-data` repo or other sources on initialization rather than the `spacy-lookups-data` package.

If the tables can't be loaded from the `lookups` registry in the lemmatizer, show how to specify the tables in `[initialize]` rather than recommending installing the `spacy-lookups-data` package.

### Types of change

?

## Checklist


adrianeboyd:
I'm not sure if this is exactly how we want to do this or how we want to name anything or load from URLs or specify the `spacy-lookups-data` repo URL within the library.

I was originally thinking of removing loading from the `spacy-lookups-data` package entirely, but it's a pretty generic mechanism to load from the `lookups` registry which seems fine to keep, since it allows users to provide their own data as python packages if they want.

The main thing we want is to stop publishing this as a package to pypi or conda-forge. We can continue having tagged releases in `spacy-lookups-data` without publishing on pypi. It would probably make sense to have the final pypi release include a warning at the top of the README that further updates will only be available at the repo and not through `spacy-lookups-data` as a package or through `spacy[lookups]`.

We could also keep `spacy-lookups-data` structured as a python package that people can install from the source or github URL, since this also isn't a real burden. The main problem is dealing with a large package on pypi, so I would just drop that part without dropping the remaining structure.

adrianeboyd:
The new registered functions could be backported to v3 for an easier transition, but I would keep the lemmatizer changes and related error messages for v4.

website/docs/api/top-level.mdx
```diff
@@ -931,6 +932,54 @@ methods are wrapped: `pipe`, `predict`, `set_annotations`, `update`, `rehearse`,
 | `additional_pipe_functions` | Additional pipeline methods to wrap. Keys are pipeline names and values are lists of method identifiers. Defaults to `None`. ~~Optional[Dict[str, List[str]]]~~ |
 | **CREATES**                 | A function that takes the current `nlp` and wraps pipe models and methods in NVTX ranges. ~~Callable[[Language], Language]~~                                    |
 
+## Miscellaneous {id="misc",version="3"}
+
+### spacy.LookupsDataLoaderFromURL.v1 {id="lookups_data_reader",tag="registered function"}
```
svlandeg, line 937:
```suggestion
## Miscellaneous {id="misc"}

### spacy.LookupsDataLoader.v1 {id="lookups_data_reader",tag="registered function",version="3"}
```

svlandeg (commented):
Generally looks good - only had a few small nitpicks.

honnibal:
The overall change sounds good to me. I agree that the size limits on publishing the lookups data are problematic, and the extensibility is clearer by making the config use the load-from-file or load-from-url style.

PR #12288, opened 2023-02-16
Return Tuple[Span] for all Doc/Span attrs that provide spans

adrianeboyd:
## Description

Return `Tuple[Span]` for all `Doc`/`Span` attrs that provide spans.

This unifies the behavior for `Doc.ents`, `Doc.sents`, and `Doc.noun_chunks` in a way that shouldn't break much existing code. `Doc.ents` already returned `Tuple[Span]`, `Doc.noun_chunks` was already calculating all spans in advance so only the return type not the speed is affected, and `Doc.sents` is the only property that may be slightly slower for very long documents. `Span` objects are very lightweight so I don't think the additional memory usage is a major concern.

### Types of change

Enhancement?

## Checklist


adrianeboyd:
It has always driven me nuts that `Doc.ents` and `Doc.sents` don't have the same signature.

This does not change the `Token`-returning attributes like `lefts` and `rights`. (They are at least internally consistent, so it doesn't bother me as much.)

svlandeg (commented):
I see your point, and it's probably too breaking to change the signature of `doc.ents`...

[update]: after some internal discussion, I do think we should consider making everything generators instead.

adrianeboyd:
I looked into making everything generators and I think it is too breaking.

svlandeg (approved):
Will be good to have this finally unified :-)

PR #12290, opened 2023-02-17
Clean up Vocab constructor

adrianeboyd:
## Description

* Change effective type of `strings` from `Iterable[str]` to `Optional[StringStore]`
  * Don't automatically add strings to vocab
* Change default values to `None`
* Remove `**deprecated_kwargs`

### Types of change

?

## Checklist


adrianeboyd:
It wouldn't be very difficult to use `Optional[Iterable[str], StringStore]]` for `Vocab(strings=)` instead if we think it is important to keep this option, but I suspect it is very rarely used outside of unit tests?

svlandeg (approved):
I think this is fine, yea!

Issue #5471, opened 2020-05-21
Calling `doc.vector` with Spacy models with no vectors produce zero length vectors for empty strings


ned2:
I noticed this in our dev environment with dummy data that used the `en_core_web_sm` rather than our custom model that has a set of vectors. 

```
nlp = spacy.load("en_core_web_sm")
len(nlp("hello").vector)
>>> 96

len(nlp("").vector)
>>> 0
```

Whereas doing the same for the `en_core_web_lg` model produces a doc vector of length 300 for both examples. 

This inconsistent lengths causes us some grief from TensorFlow which took a little while to track down. I think this might happen whenever the vector length is zero. If I drop the vectors from the en_core_web_lg model like below, the new model exhibits the same behaviour:

```python
import spacy
from spacy.lang.en import English

nlp = spacy.load("en_core_web_lg")
nlp.vocab = English.Defaults.create_vocab()
nlp.to_disk("lg_no_vectors")
nlp = spacy.load("lg_no_vectors")
len(nlp("").vector)
>>> 0
```

## Info about spaCy

* **spaCy version:** 2.2.4
* **Platform:** Linux-5.3.0-1017-aws-x86_64-with-glibc2.27
* **Python version:** 3.8.2

adrianeboyd:
I see why that could cause some grief.

The vector property uses the tensor property as a backup if there aren't any vectors, but empty documents don't get a tensor (no real processing is done for empty docs) and the empty doc itself doesn't really have access to what the tensor width would have been used for a non-empty doc, either.

It also turns out (I'm surprised I hadn't noticed before!) that only the `tagger` sets the tensor, not the `parser` or `ner`, so custom pipelines may necessarily not end up with tensors, either. (I'm not even sure whether this is the intended behavior.)

The v2 config is a bit buried (this will all be in an explicit config in v3), but you can get the tensor width from something like `nlp.get_pipe("tagger").model.tok2vec.nO`, or also just from inspecting a non-empty document.

Because it's using the shape of the vectors data array underneath when it returns the tensor-as-vector property, I can suggest a hack to get it to return the same length for empty docs:

```
import spacy
from thinc.neural.util import get_array_module

nlp = spacy.load("en_core_web_sm")

tensor_width = nlp.get_pipe("tagger").model.tok2vec.nO # or tensor_width = len(nlp("a").vector)
assert nlp.vocab.vectors.data.shape == (0, 0) # don't do this to a non-empty vector table!

[...]

PR #12292, opened 2023-02-17
Remove backoff from .vector to .tensor

adrianeboyd:
## Description

Remove backoff from `.vector` to `.tensor`

This backoff is problematic in various ways:

* tensors from `tok2vec` are used to make the `similarity` API easier to demo for `sm` models, but don't actually work for this purpose
* there's no way to get the right vector size for an empty doc (see #5471)

### Types of change

?

## Checklist


adrianeboyd:
This will definitely break various tutorials, but only kind of for things that weren't a great demo in the first place, and the fix to download a `md` model instead is straight-forward.

PR #12293, opened 2023-02-17
Tagger label smoothing

vin-ivar:
## Description
Added label smoothing to the tagger, currently a boolean that defaults to `False`. Setting it to `True` automatically sets label smoothing to `0.05`, which empirically seems like a good value.

### Types of change
Feature

## Checklist


website/docs/api/tagger.mdx
```diff
@@ -40,12 +40,13 @@ architectures and their arguments and hyperparameters.
...
+| `scorer` <Tag variant="new">3.2</Tag>          | The scoring method. Defaults to [`Scorer.score_token_attr`](/api/scorer#score_token_attr) for the attribute `"tag"`. ~~Optional[Callable]~~                                                                                                                                                            |
+| `neg_prefix` <Tag variant="new">3.2.1</Tag>    | The prefix used to specify incorrect tags while training. The tagger will learn not to predict exactly this tag. Defaults to `!`. ~~str~~                                                                                                                                                              |
+| `label_smoothing` <Tag variant="new">3.5</Tag> | Whether or not to use label smoothing. Defaults to `False`. ~~bool~~                                                                                                                                                                                                                                   |
```
adrianeboyd, line 49:
```suggestion
| `label_smoothing` <Tag variant="new">3.6</Tag> | Whether or not to use label smoothing. Defaults to `False`. ~~bool~~                                                                                                                                                                                                                                   |
```

spacy/pipeline/tagger.pyx
```diff
@@ -256,7 +258,8 @@ class Tagger(TrainablePipe):
         DOCS: https://spacy.io/api/tagger#get_loss
         """
         validate_examples(examples, "Tagger.get_loss")
-        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False, neg_prefix=self.cfg["neg_prefix"])
+        self.cfg["label_smoothing"] = 0.05 if self.cfg["label_smoothing"] else 0.0
+        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False, neg_prefix=self.cfg["neg_prefix"], label_smoothing=self.cfg["label_smoothing"])
```
adrianeboyd, line 262:
```suggestion
        label_smoothing = 0.05 if self.cfg["label_smoothing"] else 0.0
        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False, neg_prefix=self.cfg["neg_prefix"], label_smoothing=label_smoothing)
```

adrianeboyd (commented):
Usually we regret having derived hard-coded defaults in the code like this. What is the argument against having `label_smoothing: float = 0.0`?

spacy/pipeline/tagger.pyx
adrianeboyd: I realized that this might not be the best suggestion looking at the larger picture. If this is what you want to do every time, it would probably make sense to do this in `__init__` instead. (I'm not thrilled at having the same internal setting name with a different type in `cfg`, though.)

vin-ivar: > (I'm not thrilled at having the same internal setting name with a different type in cfg, though.)

Yeah this is why I don't set it in `__init__`. I don't think there's any actual problem having it here, at least not yet.

vin-ivar:
> Usually we regret having derived hard-coded defaults in the code like this. What is the argument against having `label_smoothing: float = 0.0`?

We decided to have it be either a boolean or a ternary because it makes sense to deterministically use `0.05` as the actual parameter rather than allow for too much flexibility. Is there some other way to hard-code it without doing it in the code directly, maybe splitting it into two different elements in the config?

spacy/pipeline/tagger.pyx
adrianeboyd: If it's something that only needs to be done once, it's clearer in the future if it's in `__init__` and not buried in another method. Then maybe call the internal value something else like `label_smoothing_value`. (This is still kind of dissatisfying given the `thinc` API, though.)

adrianeboyd:
So I guess one option would be to have the default be `0.05`, which is less likely to people trying out really weird things. But in general as a user looking at this, I would mainly wonder why you're trying to prevent me from setting this myself.

We try to provide good defaults and make it as easy as possible to get a good default config, but we don't try to protect people from themselves with the config settings in general. They can try out whatever ill-advised settings they would like.

spacy/tests/pipeline/test_tagger.py
```diff
@@ -67,6 +67,31 @@ def test_tagger_initialize_tag_map():
...
+    for t in TRAIN_DATA:
+        train_examples.append(Example.from_dict(nlp.make_doc(t[0]), t[1]))
+
+    nlp.initialize(get_examples=lambda: train_examples)
+    tag_scores, bp_tag_scores = tagger_ls.model.begin_update(
+        [eg.predicted for eg in train_examples]
+    )
+    no_ls_probs = tagger_no_ls.get_loss(train_examples, tag_scores)[1][0]
+    ls_probs = tagger_ls.get_loss(train_examples, tag_scores)[1][0]
+    assert_array_almost_equal((ls_probs - no_ls_probs)[0], [0.05, -0.025, -0.025])
```
danieldk, line 92: These are gradients rather than losses, so I'd name the variables like that.

spacy/pipeline/tagger.pyx
danieldk: I agree it's nicer to set this in the constructor. If we want to expose this differently than a bool in the future, it's nice if the stored label smoothing is just a `float` and not some union type.

danieldk:
> But in general as a user looking at this, I would mainly wonder why you're trying to prevent me from setting this myself.

I agree. If we wan to provide a reasonable default when enabling it, then maybe we could have `Union[bool,float]` as the argument type. (Even though I am not a big fan of unions.)

spacy/pipeline/tagger.pyx
vin-ivar: Okay, so I think the consensus is to just store this as a `float` instead of a `bool` or a union - now that we've decided to have it enabled by default, it makes more sense to use `0.05` as a default value, and allow people to set it to `0.0`/whatever other value they had in mind.

vin-ivar:
Using a default `label_smoothing = 0.05` brings some (older) tests out of sync, particularly ones that assert loss values etc. I've updated them to continue to use taggers with `label_smoothing = 0.0`, but it's also an option to update the actual comparison so it doesn't fail with `0.05`.

danieldk (approved):
Great, looks good to me!

spacy/tests/pipeline/test_tagger.py
```diff
@@ -136,7 +160,7 @@ def test_no_data():
 def test_incomplete_data():
     # Test that the tagger works with incomplete information
     nlp = English()
-    nlp.add_pipe("tagger")
+    nlp.add_pipe("tagger", config=dict(label_smoothing=0.0))
```
svlandeg, line 163: Why exactly did you need to add this?

website/docs/api/tagger.mdx
```diff
@@ -40,12 +40,13 @@ architectures and their arguments and hyperparameters.
...
+| `scorer` <Tag variant="new">3.2</Tag>          | The scoring method. Defaults to [`Scorer.score_token_attr`](/api/scorer#score_token_attr) for the attribute `"tag"`. ~~Optional[Callable]~~                                                                                                                                                            |
+| `neg_prefix` <Tag variant="new">3.2.1</Tag>    | The prefix used to specify incorrect tags while training. The tagger will learn not to predict exactly this tag. Defaults to `!`. ~~str~~                                                                                                                                                              |
+| `label_smoothing` <Tag variant="new">3.6</Tag> | Label smoothing factor. Defaults to `0.05`. ~~float~~                                                                                                                                                                                                                                   |
```
svlandeg, line 49: Do we explain what "label smoothing" is somewhere?

kadarakos: Maybe it could be explained here: https://spacy.io/usage/training#api-loop

kadarakos:
> Using a default `label_smoothing = 0.05` brings some (older) tests out of sync, particularly ones that assert loss values etc. I've updated them to continue to use taggers with `label_smoothing = 0.0`, but it's also an option to update the actual comparison so it doesn't fail with `0.05`.

Just bringing it up in the discussion that `thinc` does complain when the `label_smoothing` factor is completely wrong, meaning that when the `argmax` before and after smoothing is not the same https://github.com/explosion/thinc/blob/997731a3e82b84396332993b3fd134b6cd09c8e7/thinc/util.py#L244.

spacy/tests/pipeline/test_tagger.py
vin-ivar: Label smoothing changes the loss a bit because the logits are slightly smoother, so the test fails with `label_smoothing=0.05`. It's just a question of precision in the end, the loss is still the right order of magnitude - eg. `assert 3.28570704368758e-05 < 1e-05`.

vin-ivar:
> meaning that when the `argmax` before and after smoothing is not the same

Yeah, I ran into this thankfully. That isn't the issue with the tests though, it's mostly just precisions that are off, and that is sometimes expected behaviour - for eg. it _should_ be harder to overfit when smoothing.

website/docs/api/tagger.mdx
vin-ivar: Hmm, not sure where an explanation would fit into the API loop because this is more of a tagger parameter, at least for now... maybe a brief explanation in this table, + a link to a paper or something describing it?

adrianeboyd:
Looking at this again, if this is targeting v3.6, I think the default should stay `0.0`, since the test suite edits show how this can affect users' defaults. It is difficult that the components like `tagger` are only really versioned through the spacy versions, but this seems like it might be too much of a change.

Adding a new setting that doesn't change the defaults is something we will do in minor versions (your v3.6 config isn't backwards-compatible with v3.5), but now if you fill in your v3.5 config, you effectively get different settings/results.

This is more of a principle-of-the-thing rather than a concern that this particular change will cause huge differences in performance, but if we're having to make modifications to our own test suite, it's often a sign that it's not the best idea, and it doesn't seem like the benefits of this change outweigh these concerns?

adrianeboyd:
Instead, it would be fine to modify the quickstart to recommend this setting for v3.6, since this is how most people get their default configs anyway, so it would still be used in practice.

spacy/pipeline/tagger.pyx
```diff
@@ -89,6 +90,7 @@ class Tagger(TrainablePipe):
         overwrite=BACKWARD_OVERWRITE,
         scorer=tagger_score,
         neg_prefix="!",
+        label_smoothing=0.05,
```
adrianeboyd, line 93:
```suggestion
        label_smoothing=0.0,
```

adrianeboyd (commented):
I guess my main remaining question is whether we want this PR to go ahead and also add this to `morphologizer` in addition to `tagger` since they're pretty similar internally.

vin-ivar:
It looks like it has fairly consistent (yet minor) improvements for `morphologizer` (and maybe `lemmatizer` as well), so yeah it could make sense to include it for them too.

kadarakos:
> I guess my main remaining question is whether we want this PR to go ahead and also add this to `morphologizer` in addition to `tagger` since they're pretty similar internally.

svlandeg:
@kadarakos : did you close this on accident?

adrianeboyd:
It's intentional that you're only adding it to the config template for `tok2vec`?

vin-ivar:
Yes, I haven't tested smoothing with transformers yet.

spacy/tests/pipeline/test_tagger.py
```diff
@@ -67,6 +67,29 @@ def test_tagger_initialize_tag_map():
 ]
 
 
+def test_label_smoothing():
```
honnibal, line 70: Maybe a backwards compatibility test?

* Create an `nlp` object
* Add a tagger
* Remove the `label_smoothing` entry from its config
* Call `to_bytes()`
* Load back into a new NLP object
* Make sure that succeeds

spacy/pipeline/tagger.pyx
```diff
@@ -256,7 +258,7 @@ class Tagger(TrainablePipe):
         DOCS: https://spacy.io/api/tagger#get_loss
         """
         validate_examples(examples, "Tagger.get_loss")
-        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False, neg_prefix=self.cfg["neg_prefix"])
+        loss_func = SequenceCategoricalCrossentropy(names=self.labels, normalize=False, neg_prefix=self.cfg["neg_prefix"], label_smoothing=self.cfg["label_smoothing"])
```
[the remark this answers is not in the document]
adrianeboyd: I think this shouldn't be needed because `"label_smoothing"` is added by `__init__` and the `cfg` dicts are updated in `from_disk`, not replaced.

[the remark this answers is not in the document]
honnibal: You're right, I misread that

website/docs/api/tagger.mdx
svlandeg: Yea, that sounds good :-)

website/docs/api/morphologizer.mdx
```diff
@@ -42,12 +42,13 @@ architectures and their arguments and hyperparameters.
...
+| `extend` <Tag variant="new">3.2</Tag>          | Whether existing feature types (whose values may or may not be overwritten depending on `overwrite`) are preserved. Defaults to `False`. ~~bool~~                                                                                                                      |
+| `scorer` <Tag variant="new">3.2</Tag>          | The scoring method. Defaults to [`Scorer.score_token_attr`](/api/scorer#score_token_attr) for the attributes `"pos"` and `"morph"` and [`Scorer.score_token_attr_per_feat`](/api/scorer#score_token_attr_per_feat) for the attribute `"morph"`. ~~Optional[Callable]~~ |
+| `label_smoothing` <Tag variant="new">3.6</Tag> | Whether or not to use label smoothing. Defaults to `False`. ~~bool~~                                                                                                                                                                                                   |
```
svlandeg, line 51: This needs to be updated to `float`

danieldk:
Looks good to me, @vin-ivar maybe you can do a follow up PR if we also want to enable label smoothing in the lemmatizer?

PR #12314, opened 2023-02-21
Small `update_with_oracle_cut_size` fixes

danieldk:
## Description

Fix an off-by-one in `TransitionModel.forward`, where we always did one move more than the maximum number of moves.

This explosed another issue: when creating cut states, we skipped states where the (maximum number of) moves from that state only applied transitions that did not modify the buffer.

Replace uses of `random.uniform` by `random.randrange`.

Todo: replace remaining uses of `has_gold`, since this is more correct.

## Checklist


spacy/pipeline/transition_parser.pyx
```diff
@@ -729,9 +729,8 @@ class Parser(TrainablePipe):
                     action.do(state.c, action.label)
                     if state.is_final():
                         break
-                if moves.has_gold(eg, start_state.B(0), state.B(0)):
-                    states.append(start_state)
-                    golds.append(gold)
+                states.append(start_state)
+                golds.append(gold)
```
danieldk, line 733: I am not 100% sure if this is what we want, so this definitely needs extra sanity checking.

PR #12315, opened 2023-02-22
Bugfix/swedish tokenizer

lise-brinck:
## Description
Instead of using punctuation.py from Danish, a punctuation.py for Swedish has been created. It's a copy of the one from Danish, but with ":" only as an infix if the succeeding character is uppercase.
This fixes problem with tokenizer splitting tokens containing ":" which in Swedish works as an abbreviation, genitive and ordinal numbers marker.

### Types of change
This can be seen as a bugfix or enhancement to the Swedish tokenizer.

## Checklist


svlandeg:
Thanks for the PR, we appreciate it! We'll get this reviewed within the next week.

svlandeg (approved):
This is looking good, thanks!

PR #12318, opened 2023-02-23
Fix FUZZY operator definition

andyjessen:
## Description
The default length of the FUZZY operator is 2 and not 3. 

Example:
````
pattern = [{"LOWER": {"FUZZY": "kitten"}}]
matcher.add("KITTEN", [pattern])
doc = nlp("kitten -> sitten -> sittin -> sitting")
matches = matcher(doc)
````

Returns matches on 'kitten' (0 edits),' sitten' (1 edit), and 'sittin' (2 edits).

### Types of change

Bug fix
## Checklist

### DIFF
--- a/website/docs/usage/rule-based-matching.mdx
+++ b/website/docs/usage/rule-based-matching.mdx
@@ -384,10 +384,10 @@ the more specific attributes `FUZZY1`..`FUZZY9` you can specify the maximum
 allowed edit distance directly.
 
 ```python
-# Match lowercase with fuzzy matching (allows 3 edits)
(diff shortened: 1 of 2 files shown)


svlandeg:
Thanks!

svlandeg (approved):

adrianeboyd:
Sorry, this is a misunderstanding not a mistake in the docs. The number of edits for the default `FUZZY` implementation depends on the pattern string length, so since "kitten" is shorter than "definitely", it only allows 2 edits instead of 3 edits. See more details: https://spacy.io/usage/rule-based-matching/#fuzzy

PR #12320, opened 2023-02-23
Make generation of empty `KnowledgeBase` instances configurable in `EntityLinker`

rmitsch:
## Description

https://github.com/explosion/spaCy/discussions/12289 points out that the recent `EntityLinker` refactoring neglected to account for `empty_kb()`, which complicates proper inheritance from `KnowledgeBase`. This PR fixes that.

## Checklist


website/docs/api/entitylinker.mdx
```diff
@@ -53,19 +53,21 @@ architectures and their arguments and hyperparameters.
...
+| `get_candidates`                                    | Function that generates plausible candidates for a given `Span` object. Defaults to [CandidateGenerator](/api/architectures#CandidateGenerator), a function looking up exact, case-dependent aliases in the KB. ~~Callable[[KnowledgeBase, Span], Iterable[Candidate]]~~                                         |
+| `get_candidates_batch` <Tag variant="new">3.5</Tag> | Function that generates plausible candidates for a given batch of `Span` objects. Defaults to [CandidateBatchGenerator](/api/architectures#CandidateBatchGenerator), a function looking up exact, case-dependent aliases in the KB. ~~Callable[[KnowledgeBase, Iterable[Span]], Iterable[Iterable[Candidate]]]~~ |
+| `generate_empty_kb` <Tag variant="new">3.6</Tag>    | Function that generates an empty `KnowledgeBase` object. Defaults to generating an empty [`InMemoryLookupKB`](/api/inmemorylookupkb). ~~Callable[[Vocab, int], KnowledgeBase]~~                                                                                                                                  |
```
rmitsch, line 67: The diff for reformatted markdown tables is terrible. I added these two lines (I seem to have forgotten updating the docs in the last refactoring, so I added info for `get_candidates_batch` now :upside_down_face:).

spacy/pipeline/entity_linker.py
```diff
@@ -220,16 +226,15 @@ def __init__(
...
         self.incl_prior = incl_prior
         self.incl_context = incl_context
         self.get_candidates = get_candidates
         self.get_candidates_batch = get_candidates_batch
         self.cfg: Dict[str, Any] = {"overwrite": overwrite}
         self.distance = CosineDistance(normalize=False)
-        # how many neighbour sentences to take into account
-        # create an empty KB by default
-        self.kb = empty_kb(entity_vector_length)(self.vocab)
+        self.kb = generate_empty_kb(self.vocab, entity_vector_length)
```
rmitsch, line 237: `empty_kb()` couldn't be configured. The default implementation for `generate_empty_kb()` mirrors the behavior of the default implementation `empty_kb()`, the only difference being a slightly different signature. So this should be backwards-compatible IMO.

spacy/tests/serialize/test_serialize_kb.py
```diff
@@ -139,6 +159,6 @@ def custom_kb_factory(vocab):
         nlp2 = util.load_model_from_path(tmp_dir)
         entity_linker2 = nlp2.get_pipe("entity_linker")
         # After IO, the KB is the standard one
-        assert type(entity_linker2.kb) == InMemoryLookupKB
+        assert type(entity_linker2.kb) == SubInMemoryLookupKB
         assert entity_linker2.kb.entity_vector_length == 342
-        assert not hasattr(entity_linker2.kb, "custom_field")
+        assert hasattr(entity_linker2.kb, "custom_field")
```
svlandeg, line 164: I would repeat the same `assert` as before: 
```
assert entity_linker.kb.custom_field == 666
```

rmitsch: Updated in https://github.com/explosion/spaCy/pull/12320/commits/b6d9f4c5ab7ebcf8a3845835de8b142692cba2fe.

spacy/tests/serialize/test_serialize_kb.py
```diff
@@ -86,12 +86,19 @@ def test_serialize_subclassed_kb():
...
 
     [components]
 
     [components.entity_linker]
     factory = "entity_linker"
-
+    
+    [components.entity_linker.generate_empty_kb]
+    @misc = "spacy.CustomEmptyKB.v1"
+    custom_field = ${default_values.custom_field}
```
svlandeg, line 100: As a nitpick, I would actually have the `spacy.CustomEmptyKB.v1` (note that this shouldn't be prefixed with spaCy as it mimicks a user case) not take the `custom_field` parameter and just set it so some random int like 0. Then the test double checks that the correct `custom_field` had actually been loaded, not the default one from the empty factory.

website/docs/api/architectures.mdx
```diff
@@ -924,6 +924,15 @@ plausible [`Candidate`](/api/kb/#candidate) objects. The default
 `CandidateGenerator` uses the text of a mention to find its potential aliases in
 the `KnowledgeBase`. Note that this function is case-dependent.
 
+### spacy.CandidateBatchGenerator.v1 {id="CandidateBatchGenerator"}
```
svlandeg, line 927: This part doesn't seem related to the rest of the PR? Or did you mean to document `spacy.EmptyKB.v2` instead?

rmitsch: Ah, I meant to move that into a separate PR. I noticed that `spacy.CandidateBatchGenerator.v1` hasn't been documented yet.

rmitsch: [PR](https://github.com/explosion/spaCy/pull/12350).

spacy/tests/serialize/test_serialize_kb.py
rmitsch: That was a good catch, we kinda cheated so far on this test - we didn't (de-)serialize `custom_field`. I added custom `.to_disk()` and `.from_disk()` implementations that piggyback on the `InMemoryLookupKB` implementations. It's not super elegant, but I couldn't think of a better way without diving too deep into the `InMemoryLookupKB` serialization logic. If you have a better idea, I'd be happy to hear it.

Done in https://github.com/explosion/spaCy/pull/12320/commits/322b0050c40fcc40b18fccd651442d72d070766b.

website/docs/api/entitylinker.mdx
rmitsch: Like so? https://github.com/explosion/spaCy/pull/12320/commits/f1c48baeb591c60e2c1cd89001c6dc804d5f37c4

spacy/tests/serialize/test_serialize_kb.py
```diff
@@ -112,6 +114,47 @@ def __init__(self, vocab, entity_vector_length, custom_field):
...
+            if not path.exists():
+                path.mkdir(parents=True)
+            if not path.is_dir():
+                raise ValueError(Errors.E928.format(loc=path))
+
+            def serialize_custom_fields(
+                values: Dict[str, Any], file_path: Path
+            ) -> None:
+                with open(file_path, "wb") as file:
+                    pickle.dump(values, file)
```
svlandeg, line 129: We could do something similar to what the pipeline components do - they have an internal `self.cfg` dictionarity that needs to have json-serializable values, and then we can use `srsly.json_loads` and `srsly.write_json`

website/docs/api/entitylinker.mdx
svlandeg: yes!

spacy/tests/serialize/test_serialize_kb.py
rmitsch: `self.cfg` would be `{"custom_field": self.custom_field}` then? Or do we want to drop `self.custom_field` and only keep `self.cfg` (with entry `custom_field`)?

svlandeg: In this particular case we don't really need to define a `self.cfg`, we can just compile the dict on the fly like you did. I wanted to reference this so you could have a look how it's done in the components and how they use `srsly` .

rmitsch: Avoiding `pickle` is sound advice - I tripped over that recently with some `wikid` stuff :upside_down_face: 
Updated in https://github.com/explosion/spaCy/pull/12320/commits/3a58c19cc1419f1b15dd213461750061da846539.

PR #12328, opened 2023-02-24
Make Korean tokenizer easier to use

MycroftKang:
## Description

Currently to use the Korean tokenizer, [mecab-ko](https://bitbucket.org/eunjeon/mecab-ko/src/master/README.md) and [mecab-ko-dic](https://bitbucket.org/eunjeon/mecab-ko-dic) must be installed separately outside of python, which causes considerable inconvenience in many cases.

This change replaces [mecab-ko](https://bitbucket.org/eunjeon/mecab-ko/src/master/README.md), [mecab-ko-dic](https://bitbucket.org/eunjeon/mecab-ko-dic) and [natto-py](https://github.com/buruzaemon/natto-py) with [python-mecab-ko](https://github.com/jonghwanhyeon/python-mecab-ko), a python binding for mecab-ko, which doesn't require installing additional dependencies.

### Types of change

* enhancement

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -120,7 +120,7 @@ ja =
     sudachipy>=0.5.2,!=0.6.1
     sudachidict_core>=20211220
 ko =
-    natto-py>=0.9.0
+    python-mecab-ko>=1.3.3
 th =
     pythainlp>=2.0
 
--- a/spacy/lang/ko/__init__.py
+++ b/spacy/lang/ko/__init__.py
@@ -38,46 +38,38 @@ class KoreanTokenizer(DummyTokenizer):
     @property
     def mecab_tokenizer(self):
         # This is a property so that initializing a pipeline with blank:ko is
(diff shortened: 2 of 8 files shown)


adrianeboyd:
Ah, sorry this wasn't easier for you to find, but we've already implemented this for the upcoming spaCy v4 in #11294. We decided that due to subtle differences in the default settings we'd wait for v4 to make this change. We definitely agree that this is much easier to use and we're happy to avoid the separate mecab installs.

If you want to test the Korean tokenizer part, you can try out a dev version (currently `spacy==4.0.0.dev0`), but just keep in mind that it is an unstable release. Feedback is very welcome if you run into any problems using the new version!

MycroftKang:
@adrianeboyd Thanks for letting me know that this feature is already being prepared for release. Accordingly, since this feature has already been implemented, I will close this PR. :)

PR #12336, opened 2023-02-27
Revert "Fix FUZZY operator definition (#12318)"

adrianeboyd:
## Description

This reverts commit daedc45d050b15be8c5422aadff7b652439a562d.

The default length depends on the length of the pattern string and the original description was correct for these examples.

### Types of change

Docs.

## Checklist


svlandeg (approved):
Right! Maybe we could change 1 of the 2 examples to make it a shorter string with default 2, then have the other longer one and explain it's 3 because the length is 10.

Either way I'll merge in the revert cleanly first (and also cherry-pick to `spacy-io`)

PR #12341, opened 2023-02-28
Introduce hierarchy for EL `Candidate` objects

rmitsch:
## Description

Introduces a hierarchy for `Candidate` objects used by `KnowledgeBase` classes. This is useful for customized KBs as we need a minimal set of attributes for `EntityLinker` to work, but `Candidate` objects can hold other useful information (that may at some point also be leveraged by the EL model).

Split off of https://github.com/explosion/spaCy/pull/11669.

### Types of change

Refactoring.

## Checklist


rmitsch:
@danieldk See https://github.com/explosion/spaCy/pull/12341/commits/a97ef65b33befd107f88a89716b3d4236799c67d - I assume that `.entity_` was just a typo and you meant `.entity`. Can you confirm?

danieldk:
> @danieldk See [a97ef65](https://github.com/explosion/spaCy/commit/a97ef65b33befd107f88a89716b3d4236799c67d) - I assume that `.entity_` was a typo and you meant `.entity`. Can you confirm?

Yeah, I think it should indeed be `entity`.

spacy/pipeline/entity_linker.py
```diff
@@ -540,7 +540,7 @@ def predict(self, docs: Iterable[Doc]) -> ActivationsT:
                                 doc_scores=doc_scores,
                                 doc_ents=doc_ents,
                                 scores=[1.0],
-                                ents=[candidates[0].entity_],
+                                ents=[candidates[0].entity],
```
svlandeg, line 543: This type of fix could have been a separate PR, especially considering that it's a bug fix that we could get into `master` without worrying about whether the rest of this PR is to `master` or `v4`

spacy/kb/candidate.py
```diff
@@ -0,0 +1,109 @@
...
+    """A `BaseCandidate` object refers to a textual mention (`alias`) that may or may not be resolved
+    to a specific `entity_id` from a Knowledge Base. This will be used as input for the entity_id linking
+    algorithm which will disambiguate the various candidates to the correct one.
+    Each candidate (alias, entity_id) pair is assigned a certain prior probability.
+
+    DOCS: https://spacy.io/api/kb/#candidate-init
+    """
+
+    def __init__(
+        self, mention: str, entity_id: Union[int, str], entity_vector: List[float]
```
svlandeg, line 15: The `entity_vector` feels already pretty specific to the in-memory approach, so we can probably leave that field out of the `BaseCandidate` implementation?

spacy/kb/candidate.py
```diff
@@ -0,0 +1,109 @@
...
+    ):
+        """Create new instance of `Candidate`. Note: has to be a sub-class, otherwise error will be raised.
+        mention (str): Mention text for this candidate.
+        entity_id (Union[int, str]): Unique entity ID.
+        entity_vector (List[float]): Entity embedding.
+        """
```
svlandeg, line 21: TODO: update doc string

spacy/kb/candidate.py
```diff
@@ -0,0 +1,109 @@
...
+        self,
+        retrieve_string_from_hash: Callable[[int], str],
+        entity_hash: int,
+        entity_freq: int,
+        entity_vector: List[float],
+        alias_hash: int,
+        prior_prob: float,
+    ):
+        """
+        retrieve_string_from_hash (Callable[[int], str]): Callable retrieveing entity name from provided entity/vocab
```
svlandeg, line 60:
```suggestion
        retrieve_string_from_hash (Callable[[int], str]): Callable retrieving entity name from provided entity/vocab
```

svlandeg (commented):
Few more small comments

spacy/kb/candidate.py
svlandeg: Hm, or maybe not, I suppose the current EL implementation needs this, right?

rmitsch: [Yes](https://github.com/explosion/spaCy/blob/master/spacy/pipeline/entity_linker.py#L279).

spacy/pipeline/entity_linker.py
rmitsch: I think that doesn't apply here, as `._add_activations()` seems to be a v4 addition. `master` uses [this](https://github.com/explosion/spaCy/blob/master/spacy/pipeline/entity_linker.py#L499), which is correct.

I can split off another v4-targeted PR for this, if you want

rmitsch: It's just that `mypy` raises an error due to `str`/`int` type mismatch here (with good reason, I don't know how this passed when it was merged originally), so in case of a PR I'd have to slap on an `type: ignore` in this line.

svlandeg: Oh, right. Ok, let's just go ahead and merge it as part of this PR then.

spacy/kb/candidate.py
svlandeg: I didn't see the commit addressing this?

svlandeg:
Is this PR still draft?

spacy/kb/candidate.py
svlandeg: To clarify: the mention of "Candidate" and it having to be a sub-class.

rmitsch: Ah, my push timed out when my connection dropped :smiling_face_with_tear:

rmitsch: It's this one: https://github.com/explosion/spaCy/pull/12341/commits/3da0712582e272bd757cfe41a6a477c9d875bd9e

spacy/kb/candidate.py
```diff
@@ -0,0 +1,109 @@
...
+        return self._mention
+
+    @property
+    def entity_vector(self) -> List[float]:
+        """RETURNS (List[float]): Entity vector."""
+        return self._entity_vector
+
+
+class Candidate(BaseCandidate):
+    """`Candidate` for InMemoryLookupKBCandidate."""
```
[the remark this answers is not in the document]
svlandeg: Should we call this `InMemoryCandidate`?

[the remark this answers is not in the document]
rmitsch: It'd be clearer...and since we're aiming for v4, I think it'd make sense. Also `BaseCandidate` -> `Candidate` then?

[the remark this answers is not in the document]
svlandeg: Ah, yes!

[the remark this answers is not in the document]
svlandeg: I think by now we'll just have to accept that the entity linker & KB stuff will be pretty different in `v4`. We'll make sure to have a clear "convert" tutorial.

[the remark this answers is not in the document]
rmitsch: I think so too. And since stuff will be breaking anyway, we should aim to get things as right as possible.

spacy/kb/candidate.py
```diff
@@ -0,0 +1,118 @@
...
+class InMemoryCandidate(Candidate):
+    """Candidate for InMemoryLookupKB."""
+
+    def __init__(
+        self,
+        retrieve_string_from_hash: Callable[[int], str],
+        entity_hash: int,
+        entity_freq: int,
+        entity_vector: List[float],
+        alias_hash: int,
```
svlandeg, line 73: `alias` and `mention` are used interchangeabily here, right? I suggest we stick with one?

rmitsch: That's true. I'd go with mention then, since `alias` doesn't apply anymore. Agreed?

svlandeg: I don't really get why we need the additional `alias methods on this class, maybe I'm missing something.

website/docs/api/kb.mdx
```diff
@@ -190,25 +190,27 @@ Restore the state of the knowledge base from a given directory. Note that the
...
 
-## Candidate {id="candidate",tag="class"}
+## InMemoryCandidate {id="candidate",tag="class"}
 
-A `Candidate` object refers to a textual mention (alias) that may or may not be
-resolved to a specific entity from a `KnowledgeBase`. This will be used as input
-for the entity linking algorithm which will disambiguate the various candidates
-to the correct one. Each candidate `(alias, entity)` pair is assigned to a
-certain prior probability.
+A `InMemoryCandidate` object refers to a textual mention (alias) that may or may
```
svlandeg, line 195:
```suggestion
An `InMemoryCandidate` object refers to a textual mention (alias) that may or may
```

website/docs/api/kb.mdx
```diff
@@ -190,25 +190,27 @@ Restore the state of the knowledge base from a given directory. Note that the
...
+candidates to the correct one. Each candidate `(alias, entity)` pair is assigned
+to a certain prior probability.
 
-### Candidate.\_\_init\_\_ {id="candidate-init",tag="method"}
+### InMemoryCandidate.\_\_init\_\_ {id="candidate-init",tag="method"}
 
-Construct a `Candidate` object. Usually this constructor is not called directly,
-but instead these objects are returned by the `get_candidates` method of the
-[`entity_linker`](/api/entitylinker) pipe.
+Construct a `InMemoryCandidate` object. Usually this constructor is not called
```
svlandeg, line 203:
```suggestion
Construct an `InMemoryCandidate` object. Usually this constructor is not called
```

spacy/kb/candidate.py
rmitsch: So far I kept them for backwards compatibility. It's true that they aren't used in `EntityLinker` anywhere (otherwise we'd have to have this in `Candidate` as well anyway). Without `alias` and `alias_` won't have the possibility to retrieve any information about the alias in an instance of `InMemoryCandidate`, which would be annoying, I think?

website/docs/api/kb.mdx
rmitsch: PyCharm refactoring doesn't take articles into account :stuck_out_tongue:

rmitsch: I'd drop "(alias)" though and phrase everything as "mention". I think "alias" is confusing, as mentions don't have to be aliases of entity names already (and even less so once we primarily work in the embedding space). What do you think?

spacy/kb/candidate.py
svlandeg: but the properties `_alias` and `mention` are the same, right? So maybe we should just add one method `mention_hash`?

spacy/kb/candidate.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @property
+    def entity(self) -> int:
+        """RETURNS (int): Unique entity ID."""
+        return self._entity_id
+
+    @property
+    def entity_(self) -> str:
+        """RETURNS (int): Entity name."""
+        return self._entity_name
```
svlandeg, line 46: I know how this historically came to be because of storing the hashes etc, but I want to take a moment and consider whether we still want this. Also whether we should insist that the `entity_id` is an `int`, which might get annoying for use-cases where the KB identifiers are actually strings, like for Wikidata.

spacy/kb/candidate.py
rmitsch: Ah, that's what you mean. Yes, that makes sense!

website/docs/api/kb.mdx
svlandeg: agreed!

spacy/kb/candidate.py
rmitsch: I actually had this as `Union[str, int]` recently. I reverted to `int` because `v4` [changes](https://github.com/explosion/spaCy/blob/v4/spacy/pipeline/entity_linker.py#L588) expect `.entity` to be an `int` and I'm not sure how trivial it would be to modify the `_add_activation` apparatus.

That said, we could always allow passing `entity_id` as `Union[int, str]` - if it's a `str`, hash it in `Candidate` and return the internal hash `int` in `.entity`, if it's an `int`, return that. Does that make sense to you?

rmitsch: I can just draft this, if it's easier to discuss then.

spacy/kb/candidate.py
rmitsch: Do we want to keep the `alias` terminology in `InMemoryLookupKB`?

rmitsch: I'd say yes, since it still makes sense there.

rmitsch: Updates to variable and property naming in https://github.com/explosion/spaCy/pull/12341/commits/46fe069f8731c3d591963fe4a1e3e1c1a4b1eef9 and https://github.com/explosion/spaCy/pull/12341/commits/94e57d0ed5fe1981cf5ac2b54964d0e3f14533a4.

spacy/kb/candidate.py
svlandeg: I knew I'd seen `Union[str, int]` somewhere! Yea, I think that's better. And then we can have a property `entity_hash` that returns an `int` by hashing if it's a `str`. So yea, I think we're proposing the same thing.

rmitsch: I restructured `Candidate` attributes in https://github.com/explosion/spaCy/pull/12341/commits/38dce966e55dbeb61d1d085ff97ce50f5095dea8. This includes doing away with storing the mention hash, since in the current version `Candidate` stores the mention string anyway; renaming attributes for more clarity; and using `Union[str, int]` for `entity_id`, with support for converting `entity_id` into a string or integer respectively.

Consider this a draft and let me know what you think.

spacy/errors.py
```diff
@@ -960,6 +960,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E4003 = ("Training examples for distillation must have the exact same tokens in the "
              "reference and predicted docs.")
     E4004 = ("Backprop is not supported when is_train is not set.")
+    E4005 = ("Expected `entity_id` to be of type {should_type}, but is of type {is_type}.")
```
svlandeg, line 963: I wouldn't use "should" or "is" here as part of the variable names, but instead something like "exp" and "value"/"found" to be more in line with other error messages (not that we're that consistent, but this feels off to me)

spacy/kb/candidate.py
```diff
@@ -0,0 +1,116 @@
...
+    @property
+    def entity_id_int(self) -> int:
+        """RETURNS (int): Numerical representation of entity ID (if entity ID is numerical, this is just the entity ID,
+        otherwise the hash of the entity ID string)."""
+        return self._entity_id_hash
+
+    @property
+    def entity_id_str(self) -> str:
+        """RETURNS (str): String representation of entity ID."""
+        return str(self._entity_id)
```
svlandeg, line 53: Can't we use the same convention as what we do on `Token`, to have the underscored version return the `str` and the normal one the hash? And not have a method that returns the `Union`, I don't think we actually need it?

spacy/kb/candidate.py
```diff
@@ -0,0 +1,116 @@
...
+        hash_to_str: Callable[[int], str],
+        entity_id: int,
+        mention: str,
+        entity_vector: List[float],
+        prior_prob: float,
+        entity_freq: int,
+    ):
+        """
+        hash_to_str (Callable[[int], str]): Callable retrieving entity name from provided entity/vocab hash.
+        entity_id (str): Entity ID as hash that can be looked up with InMemoryKB.vocab.strings.__getitem__().
```
svlandeg, line 85:
```suggestion
        entity_id (int): Entity ID as hash that can be looked up with InMemoryKB.vocab.strings.__getitem__().
```

svlandeg (commented):
Few more nitpicking comments...

spacy/kb/candidate.py
rmitsch: Yeah, that's cleaner. Done in https://github.com/explosion/spaCy/pull/12341/commits/cea58ade8969592266483848feae3fd06f537e6b.

rmitsch: It is a bit confusing though possibly that we ask to specify `entity_id: Union[str, int]` and then have a property `entity_id() -> int`. Perhaps we should rename the `__init__()` argument?

rmitsch: Or we rename the properties `entity_id` and `entity_id_` to `entity` and `entity_` instead. Also not that nice.

svlandeg: Hmmmm. I think the current solution is fine. It's not 100% ideal but I don't think any of the alternatives are better.

website/docs/api/kb.mdx
svlandeg: This wasn't addressed I think?

rmitsch: Updated in https://github.com/explosion/spaCy/pull/12341/commits/1c937db3af6a66fa36fbb0538dc975450254d301.

spacy/kb/candidate.pxd
```diff
@@ -1,12 +1,16 @@
...
-    cdef float prior_prob
+    cdef readonly str _entity_id_
+    cdef readonly hash_t _entity_id
+    cpdef vector[float] _entity_vector
+    cdef float _prior_prob
+
+
+cdef class InMemoryCandidate(Candidate):
+    cdef readonly InMemoryLookupKB _kb
+    cdef hash_t _entity_hash
```
svlandeg, line 14: Isn't `entity_hash` the same as `_entity_id` now?

spacy/kb/candidate.pxd
```diff
@@ -1,12 +1,16 @@
...
 
-# Object used by the Entity Linker that summarizes one entity-alias candidate combination.
 cdef class Candidate:
-    cdef readonly KnowledgeBase kb
-    cdef hash_t entity_hash
-    cdef float entity_freq
-    cdef vector[float] entity_vector
-    cdef hash_t alias_hash
-    cdef float prior_prob
+    cdef readonly str _entity_id_
```
svlandeg, line 6: Ah, hm. Storing the string specifically sort of defeats the point of having the hash and storing that though...

spacy/kb/candidate.pyx
```diff
@@ -1,74 +1,116 @@
...
 from .kb cimport KnowledgeBase
-from ..tokens import Span
+from .kb_in_memory cimport InMemoryLookupKB
+from ..errors import Errors
 
 cdef class Candidate:
-    """A `Candidate` object refers to a textual mention (`alias`) that may or may not be resolved
-    to a specific `entity` from a Knowledge Base. This will be used as input for the entity linking
+    """A `Candidate` object refers to a textual mention that may or may not be resolved
+    to a specific `entity_id` from a Knowledge Base. This will be used as input for the entity_id linking
```
svlandeg, line 11: replace all gone wrong :p

spacy/kb/candidate.pyx
```diff
@@ -1,74 +1,116 @@
...
-        self.entity_freq = entity_freq
-        self.entity_vector = entity_vector
-        self.alias_hash = alias_hash
-        self.prior_prob = prior_prob
+    def __init__(
+        self,
+        entity_id: str,
+        entity_vector: vector[float],
+        prior_prob: float,
+    ):
```
svlandeg, line 23: The alias/mention is not here?

spacy/kb/candidate.pxd
svlandeg: Maybe this should be a function instead?

spacy/kb/candidate.pyx
```diff
@@ -1,74 +1,116 @@
...
+                Errors.E1046.format(cls_name=self.__class__.__name__)
+            )
 
-    @property
-    def entity(self) -> int:
-        """RETURNS (uint64): hash of the entity's KB ID/name"""
-        return self.entity_hash
+        self._entity_id_ = entity_id
+        # Note that hashing an int value yields the same int value.
+        self._entity_id = hash(entity_id)
```
svlandeg, line 40: Continuing with the same reasoning. The original thought was to use `vocab` to do the `str`-`int` conversions so we'd only have to store the `int`

svlandeg: Why aren't we doing that anymore? `KnowledgeBase` still has a `vocab`?

spacy/kb/candidate.pyx
```diff
@@ -1,74 +1,116 @@
...
-    """
-    Return candidate entities for the given mentions and fetching appropriate entries from the index.
-    kb (KnowledgeBase): Knowledge base to query.
-    mention (Iterable[Span]): Entity mentions for which to identify candidates.
-    RETURNS (Iterable[Iterable[Candidate]]): Identified candidates.
-    """
-    return kb.get_candidates_batch(mentions)
+    @property
+    def mention(self) -> str:
+        """RETURNS (str): ID/name of this entity in the KB"""
```
svlandeg, line 106: update doc string

svlandeg (commented):
More comments now that Cython is back ;-)

spacy/kb/candidate.pxd
rmitsch: Ya, it's not ideal. If we don't do that though we can't support string IDs.

spacy/kb/candidate.pyx
rmitsch: - One motivation behind the design was that we'd support `str` identifier for entities (see above).
- We have `vocab`, but an entity ID isn't necessarily in there IMO. If you have "Q123456789", this might never occur in `vocab`, but be a valid ID. Or am I missing something here?

spacy/kb/candidate.pyx
rmitsch: Yes, because it should be up to the KB and `Candidate` child class implementations how to store and retrieve mentions. Originally `Candidate` stored `mention` as string, but this meant that `InMemoryLookupKB` had an unnecessarily large memory footprint. Instead `Candidate` only implements the (theoretically abstract :cry:) property to return `mention`.

E. g. for `WikiKB` I wouldn't want to look up mentions every time, since that would mean querying the DB on disk for each individual lookup. Instead I'd store `mention` in my candidate, which can be an implementation detail of `WikiKBCandidate`.

spacy/kb/candidate.pxd
rmitsch: Alternatively: we don't have an attribute for `entity_id` / `entity_id` in `Candidate` and leave that as an implementation detail for the inheriting class. I think that'd be fine as the `Candidate` interface would still describe all the functionality that inheriting classes have to inherit.

spacy/kb/candidate.pyx
rmitsch: Updated in https://github.com/explosion/spaCy/pull/12341/commits/348dd1c87ec853f837907ef8b804b103be1ebfcc.

spacy/kb/candidate.pxd
rmitsch: Ah, I forgot to remove this. It's not used and is indeed the same as `_entity_id`. Updated in https://github.com/explosion/spaCy/pull/12341/commits/27053912da0fbaae216e2d862371298228346063.

spacy/kb/candidate.pxd
rmitsch: I actually do think that'd be better. This would avoid the weirdness of always having to specify your `entity_id` as string even if it's an int.

svlandeg: Maybe we just want to remove most of these from `Candidate` and put them in `InMemoryCandidate` itself, so that inheritance remains as flexible as possible?

spacy/kb/candidate.pyx
svlandeg: Hm. It just feels weird to me have a constructor of an abstract class that can never be used as such. Because semantically it makes no sense to have an entity ID and a prior probability without actually also having a mention?

svlandeg: And so it feels like this constructor here doesn't really explain to users what a `Candidate` is / should be.

spacy/kb/candidate.pyx
svlandeg: The `vocab` can be extended though, right? That's how it worked before? I guess I'm just confused again why we needed to change this.

spacy/kb/candidate.pyx
rmitsch: > Hm. It just feels weird to me have a constructor of an abstract class that can never be used as such.

I'd argue that it can be used as a base constructor called from the inheriting class. And an abstract class can never be used as such. Having said that, I agree that this is a not-quite-here-not-quite-there situation. Ideally I'd have the mention in here as well, but then we burden `InMemoryCandidate` with storing this too.

rmitsch: Alternative: we change `Candidate` to be purely functional, i. e. an interface, not storing _any_ attributes. This would at least be more consistent.

rmitsch: And the only place candidates are instantiated are in the knowledge base classes, so there is no need for a unified `__init__()` signature.

rmitsch:
@svlandeg Considering the two open threads, I'd suggest to convert `Candidate` to an interface / delegate responsibility for all attributes to inheriting classes. This way we are as flexible as possible in terms of how storage and retrieval of those attributes is handled.
This should address the issues in both threads IMO. What do you think?

spacy/kb/candidate.pyx
svlandeg: Yes, I was arriving to the same conclusion :-) https://github.com/explosion/spaCy/pull/12341/files#r1132433003

spacy/kb/candidate.pyx
```diff
@@ -1,74 +1,114 @@
...
+        entity_vector: vector[float],
+        prior_prob: float,
+    ):
+        """Initializes properties of abstract base class `Candidate`.
+        entity_id (Union[str, int]): Unique entity ID.
+        entity_vector (List[float]): Entity embedding.
+        prior_prob (float): Prior probability of entity for this mention - i.e. the probability that, independent of
+            the context, this mention resolves to this entity_id in the corpus used to build the knowledge base. In
+            cases in which this isn't always possible (e.g.: the corpus to analyse contains mentions that the KB corpus
+            doesn't) it might be better to eschew this information and always supply the same value.
```
rmitsch, line 27: Regarding the mention/alias terminology: this is a bit tricky with `prior_prob`. It only really makes sense for _aliases_, since we don't and realistically can't have information on the prior probabilities of arbitrary mentions resolving to specific entities, just for the specific aliases stored in the KB.

The context in which this is used though is a "mention". As in, for "Boston play tonight." candidates for "Boston" would refer to a _mention_, not an alias. This points to a deeper problem: 
- Prior probabilities only make sense if we are operating if mention == alias. This won't always be the case. How do we cope with situations in which some mentions have exactly matching aliases, and therefore a non-zero prior, and others don't? Will the former always be preferred then?
E. g.: "Boston" may refer to the basketball team "Boston Celtics", but ["Boston" isn't listed as an alias](https://www.wikidata.org/wiki/Q131371). If we accept that the city of Boston has a non-zero prior (as "Boston" would certainly be a listed alias), this automatically puts "Boston Celtics" at an disadvantage. This doesn't seem great.
- The concept of prior probability for candidate retrieval without exact lexical matching seems questionable for me due to that. Should this be an implementation detail of `InMemoryLookupKB` (and `InMemoryCandidate`)?
- If it should be an implementation detail - how should `EntityLinker` deal with that? Assume `prior_prob` is zero if it's not available in the candidate instance?

website/docs/api/inmemorylookupkb.mdx
```diff
@@ -194,26 +194,26 @@ to you.
...
-| Name        | Description                                                                                  |
-| ----------- | -------------------------------------------------------------------------------------------- |
-| `mentions`  | The textual mention or alias. ~~Iterable[Span]~~                                             |
-| **RETURNS** | An iterable of iterable with relevant `Candidate` objects. ~~Iterable[Iterable[Candidate]]~~ |
+| Name        | Description                                                                                                  |
+| ----------- | ------------------------------------------------------------------------------------------------------------ |
+| `mentions`  | The textual mention or alias. ~~Iterable[Span]~~                                                             |
+| **RETURNS** | An iterable of iterable with relevant `InMemoryCandidate` objects. ~~Iterable[Iterable[InMemoryCandidate]]~~ |
 
 ## InMemoryLookupKB.get_alias_candidates {id="get_alias_candidates",tag="method"}
```
rmitsch, line 202: Actually we should drop this, considering `_get_alias_candidates()` is private now. Alternatively we can re-expose the method.

spacy/kb/candidate.pyx
```diff
@@ -1,74 +1,100 @@
...
+        """RETURNS (str): String representation of entity ID."""
+        raise NotImplementedError
 
     @property
-    def alias(self) -> int:
-        """RETURNS (uint64): hash of the alias"""
-        return self.alias_hash
+    def mention(self) -> str:
+        """RETURNS (str): Mention."""
+        raise NotImplementedError
```
svlandeg, line 37: I thought we were going to remove this one altogether from `Candidate`, no?

spacy/kb/candidate.pyx
```diff
@@ -1,74 +1,100 @@
...
-    """
-    Return candidate entities for a given mention and fetching appropriate entries from the index.
-    kb (KnowledgeBase): Knowledge base to query.
-    mention (Span): Entity mention for which to identify candidates.
-    RETURNS (Iterable[Candidate]): Identified candidates.
-    """
-    return kb.get_candidates(mention)
+    @property
+    def mention(self) -> str:
+        return self._kb.vocab.strings[self._mention]
```
svlandeg, line 91: I'd name these `self._alias` and property `alias`, as they both refer to the string and not the `Span`.

spacy/kb/candidate.pyx
svlandeg: (was that not the latest conclusion? I might have lost track of the various discussions)

spacy/kb/kb.pyx
```diff
@@ -32,8 +32,8 @@ cdef class KnowledgeBase:
 
     def get_candidates_batch(self, mentions: Iterable[Span]) -> Iterable[Iterable[Candidate]]:
         """
-        Return candidate entities for specified texts. Each candidate defines the entity, the original alias,
-        and the prior probability of that alias resolving to that entity.
+        Return candidate entities for specified texts. Each candidate defines the entity, the original mention,
+        and the prior probability of this mention resolving to that entity.
```
svlandeg, line 36: Same question here (and in the next method) about whether or not the candidate will actually include the alias/mention.

spacy/ml/models/entity_linker.py
```diff
@@ -117,3 +117,25 @@ def create_candidates_batch() -> Callable[
...
 ]:
     return get_candidates_batch
+
+
+def get_candidates(kb: KnowledgeBase, mention: Span) -> Iterable[Candidate]:
+    """
+    Return candidate entities for a given mention and fetching appropriate entries from the index.
+    kb (KnowledgeBase): Knowledge base to query.
+    mention (Span): Entity mention for which to identify candidates.
+    RETURNS (Iterable[InMemoryCandidate]): Identified candidates.
```
svlandeg, line 127:
```suggestion
    RETURNS (Iterable[Candidate]): Identified candidates.
```
(right?)

spacy/ml/models/entity_linker.py
```diff
@@ -117,3 +117,25 @@ def create_candidates_batch() -> Callable[
...
+
+
+def get_candidates_batch(
+    kb: KnowledgeBase, mentions: Iterable[Span]
+) -> Iterable[Iterable[Candidate]]:
+    """
+    Return candidate entities for the given mentions and fetching appropriate entries from the index.
+    kb (KnowledgeBase): Knowledge base to query.
+    mention (Iterable[Span]): Entity mentions for which to identify candidates.
+    RETURNS (Iterable[Iterable[InMemoryCandidate]]): Identified candidates.
```
svlandeg, line 139:
```suggestion
    RETURNS (Iterable[Iterable[Candidate]]): Identified candidates.
```

website/docs/api/inmemorylookupkb.mdx
svlandeg: I agree that we can drop this, and have users query by mention instead of by alias as best practice.

svlandeg (commented):
This is feeling pretty close to be merged! Few more comments and we can probably wrap this up.

spacy/ml/models/entity_linker.py
rmitsch: Yes!

spacy/kb/kb.pyx
rmitsch: Updated in https://github.com/explosion/spaCy/pull/12341/commits/3cfc1c6accb4f75b141b075aa53d8d9ae12166f0 and https://github.com/explosion/spaCy/pull/12341/commits/80fb0666b98ff0bd96c36daf5f7d0e5e00ecdb25.

spacy/kb/candidate.pyx
rmitsch: Done in https://github.com/explosion/spaCy/pull/12341/commits/3cfc1c6accb4f75b141b075aa53d8d9ae12166f0.

spacy/kb/candidate.pyx
rmitsch: Yes - I forgot to drop it. Done in https://github.com/explosion/spaCy/pull/12341/commits/3cfc1c6accb4f75b141b075aa53d8d9ae12166f0.

website/docs/api/inmemorylookupkb.mdx
rmitsch: Done in https://github.com/explosion/spaCy/pull/12341/commits/3cfc1c6accb4f75b141b075aa53d8d9ae12166f0.

rmitsch:
A thought: is the name `InMemoryCandidate` perhaps misleading? Kinda makes it sound like other candidates are not "in memory" and this is special to this class. It should be clear that's not the case after like 10 secs of research, but still. Maybe `InMemoryKBCandidate` instead?

spacy/errors.py
```diff
@@ -209,7 +209,10 @@ class Warnings(metaclass=ErrorsWithCodes):
             "`enabled` ({enabled}). Be aware that this might affect other components in your pipeline.")
     W124 = ("{host}:{port} is already in use, using the nearest available port {serve_port} as an alternative.")
 
+    # v4 warning strings
     W400 = ("`use_upper=False` is ignored, the upper layer is always enabled")
+    W401 = ("`incl_prior is True`, but the selected knowledge base type {kb_type} doesn't support prior probability "
+            "lookups.")
```
svlandeg, line 215: We should add here what this will mean in practice: prior probabilities will in fact not taken into account, even if the KB has them available - if it does it should change the flag to `True`. Maybe we should even say that specifically, that if it's a custom KB that does support prior probabilities they need to change that method. If not, `incl_prior` simply won't work.

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -502,20 +503,20 @@ def create_kb(vocab):
...
-        return kb.get_alias_candidates(span.text.lower())
+        return kb._get_alias_candidates(span.text.lower())
 
     def get_lowercased_candidates_batch(kb, spans):
         return [get_lowercased_candidates(kb, span) for span in spans]
 
     @registry.misc("spacy.LowercaseCandidateGenerator.v1")
     def create_candidates() -> Callable[
-        [InMemoryLookupKB, "Span"], Iterable[Candidate]
+        [InMemoryLookupKB, "Span"], Iterable[InMemoryCandidate]
```
svlandeg, line 513: Why do we need to change the type here to `InMemoryCandidate`? Does it really matter what the exact type of the candidate is, can't the code be more generic to deal with any kind?

svlandeg: (same question for the next one)

svlandeg (commented):
I'm so sorry, I found more nitpicks 🙈

spacy/tests/pipeline/test_entity_linker.py
rmitsch: We don't need to. By default I try to be as specific as possible when it comes to typing. If you prefer, we can stick with `Iterable[Candidate]` here, I don't mind.

svlandeg: Basically my thinking was that all candidate generators would have the same return type, and thus be more uniform. But I can also see the reasoning why to make it more specific if we know that that's the exact subtype that'll come up. I'm not sure which is best, tbh.

svlandeg: Basically my thinking was that all candidate generators would have the same return type, and thus be more uniform. But I can also see the reasoning why to make it more specific if we know that that's the exact subtype that'll come up. I'm not sure which is best, tbh.

rmitsch: Right, so my approach is usally: when writing interfaces/classes/functions, i. e. things that will be used in multiple places, go with the most abstract type supporting the operations in there. When writing "consumer" code, i. e. you are writing something that won't be reused anywhere else, use the most specific types possible as this makes reading the code easier and supports better auto-complete.

This is my two cents. Having said that, I don't care much in this case anyway, as it's just a test.

spacy/errors.py
rmitsch: Updated in https://github.com/explosion/spaCy/pull/12341/commits/2377b67f81d5707f46ba6ca085591dcc9256f334. Feel free to close if you think that's good.

spacy/kb/kb_in_memory.pyx
```diff
@@ -18,7 +18,7 @@ from .. import util
 from ..util import SimpleFrozenList, ensure_path
 from ..vocab cimport Vocab
 from .kb cimport KnowledgeBase
-from .candidate import Candidate as Candidate
+from .candidate import InMemoryCandidate
 
 
 cdef class InMemoryLookupKB(KnowledgeBase):
```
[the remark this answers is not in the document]
rmitsch: I know that's your favorite thing, yes. Hm...`MemoryKB` and `MemoryKBCandidate`? Slightly less verbose and still gets the most important point across IMO.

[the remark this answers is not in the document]
svlandeg: Yea, why not?

spacy/tests/pipeline/test_entity_linker.py
svlandeg: Yea, to me it was more of a question also how we type the actual default candidate generator.

In my opinion, we don't get any added advantages of typing the more specific type, because we wouldn't want any other code to rely on that specific subtype, because at that point we can't switch in a different candidate generator.

spacy/errors.py
```diff
@@ -209,7 +209,11 @@ class Warnings(metaclass=ErrorsWithCodes):
             "`enabled` ({enabled}). Be aware that this might affect other components in your pipeline.")
     W124 = ("{host}:{port} is already in use, using the nearest available port {serve_port} as an alternative.")
 
+    # v4 warning strings
     W400 = ("`use_upper=False` is ignored, the upper layer is always enabled")
+    W401 = ("`incl_prior is True`, but the selected knowledge base type {kb_type} doesn't support prior probability "
+            "lookups (`.supports_prior_probs is False`). If your KB does support prior probability lookups, make sure "
+            "to return True in `.supports_prior_probs`.")
```
svlandeg, line 216:
```suggestion
    W401 = ("`incl_prior is True`, but the selected knowledge base type {kb_type} doesn't support prior probability "
            "lookups so this setting will be ignored. If your KB does support prior probability lookups, make sure "
            "to return `True` in `.supports_prior_probs`.")
```

spacy/tests/pipeline/test_entity_linker.py
rmitsch: Isn't developers' knowledge about the returned type an advantage in itself? Typing is also a form of documentation and it would be more understandable if we type the specific type that is returned by the default candidate retrieval function IMO.

I defer to your judgement here though, this is not important enough for a debate :grin:

svlandeg: But the developer shouldn't even really need to know about the returned type other than it being a `Candidate`, if we want the candidate generators to be uniform in their output, you know?

So yea I do think it's best practice throughout the code base to have architectures that are meant to be swappable with eachother, to have the exact same output type, even if in reality a subtype is returned.

spacy/kb/kb_in_memory.pyx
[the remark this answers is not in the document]
svlandeg: (let's continue the naming discussion in private and for a different PR)

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -502,20 +503,20 @@ def create_kb(vocab):
...
     ]:
         return get_lowercased_candidates
 
     @registry.misc("spacy.LowercaseCandidateBatchGenerator.v1")
     def create_candidates_batch() -> Callable[
-        [InMemoryLookupKB, Iterable["Span"]], Iterable[Iterable[Candidate]]
+        [InMemoryLookupKB, Iterable["Span"]], Iterable[Iterable[InMemoryCandidate]]
```
svlandeg, line 519:
```suggestion
        [InMemoryLookupKB, Iterable["Span"]], Iterable[Iterable[Candidate]]
```

spacy/tests/pipeline/test_entity_linker.py
svlandeg, line 513:
```suggestion
        [InMemoryLookupKB, "Span"], Iterable[Candidate]
```

website/docs/api/inmemorylookupkb.mdx
```diff
@@ -194,26 +194,10 @@ to you.
...
-> candidates = kb.get_alias_candidates("Douglas")
-> ```
-
-| Name        | Description                                                   |
-| ----------- | ------------------------------------------------------------- |
-| `alias`     | The textual mention or alias. ~~str~~                         |
-| **RETURNS** | The list of relevant `Candidate` objects. ~~List[Candidate]~~ |
+| Name        | Description                                                                                                  |
+| ----------- | ------------------------------------------------------------------------------------------------------------ |
+| `mentions`  | The textual mention or alias. ~~Iterable[Span]~~                                                             |
```
svlandeg, line 199:
```suggestion
| `mentions`  | The textual mentions. ~~Iterable[Span]~~                                                                     |
```

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -7,10 +7,10 @@
 from spacy import registry, util
 from spacy.attrs import ENT_KB_ID
 from spacy.compat import pickle
-from spacy.kb import Candidate, InMemoryLookupKB, get_candidates, KnowledgeBase
+from spacy.kb import InMemoryCandidate, InMemoryLookupKB, KnowledgeBase
```
svlandeg, line 10:
```suggestion
from spacy.kb import Candidate, InMemoryLookupKB, KnowledgeBase
```

spacy/tests/pipeline/test_entity_linker.py
rmitsch: That's a good point.

PR #12344, opened 2023-02-28
Entity linking: use `SpanGroup` instead of `Iterable[Span]` for mentions

rmitsch:
## Description

Use `SpanGroup` instead of `Iterable[Span]` for mentions to be looked up in entity linking.

Split off of https://github.com/explosion/spaCy/pull/11669/.

## Checklist


svlandeg (approved):
After you've run `prettier` on the `mdx` files, this should be good to merge!

PR #12347, opened 2023-03-01
KnowledgeBase length fix

svlandeg:
## Description
The `entity_linker` was still querying the `KnowledgeBase` `__len__` property, which got moved to `InMemoryLookupDB` since 3.5. Instead, proposing to have an `is_empty` function. Internals for now, but from `v4` onwards we could have this as an abstract method on `KnowledgeBase` so that downstream implementations always provide it.

### Types of change
bug fix

## Checklist

### DIFF
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -444,8 +444,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E133 = ("The sum of prior probabilities for alias '{alias}' should not "
             "exceed 1, but found {sum}.")
     E134 = ("Entity '{entity}' is not defined in the Knowledge Base.")
-    E139 = ("Knowledge base for component '{name}' is empty. Use the methods "
-            "`kb.add_entity` and `kb.add_alias` to add entries.")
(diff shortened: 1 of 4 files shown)


spacy/errors.py
```diff
@@ -444,8 +444,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E133 = ("The sum of prior probabilities for alias '{alias}' should not "
             "exceed 1, but found {sum}.")
     E134 = ("Entity '{entity}' is not defined in the Knowledge Base.")
-    E139 = ("Knowledge base for component '{name}' is empty. Use the methods "
-            "`kb.add_entity` and `kb.add_alias` to add entries.")
+    E139 = ("Knowledge base for component '{name}' is empty.")
```
svlandeg, line 447: I shortened this error message because the `add_*` methods are implementation specific and might not be relevant for the user's custom KB.

rmitsch (approved):
Good idea, and it's as far as we can go w/o breaking things in v3.

PR #12349, opened 2023-03-01
Allow setting last=False when adding a component

svlandeg:
## Description
When a component is added, the default order is "last". However, when you specify specifically that `last=False`, arguably the pipeline should honour that. Here, I propose to set it as the first component in that case. 

### Types of change
feature

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -743,6 +743,7 @@ class Language:
         """Add a component to the processing pipeline. Valid components are
         callables that take a `Doc` object, modify it and return it. Only one
         of before/after/first/last can be set. Default behaviour is "last".
+        If "last" is set to False, then the default is "first".
 
         factory_name (str): Name of the component factory.
         name (str): Name of pipeline component. Overwrites existing
@@ -816,10 +817,10 @@ class Language:
             raise ValueError(
                 Errors.E006.format(args=all_args, opts=self.component_names)
             )
-        if last or not any(value is not None for value in [first, before, after]):
-            return len(self._components)
-        elif first:
+        if first or last is False:
             return 0
+        elif last or not any(value is not None for value in [first, before, after]):
+            return len(self._components)
         elif isinstance(before, str):
             if before not in self.component_names:
                 raise ValueError(
--- a/spacy/tests/pipeline/test_pipe_methods.py
+++ b/spacy/tests/pipeline/test_pipe_methods.py
@@ -188,6 +188,14 @@ def test_add_pipe_last(nlp, name1, name2):
     assert nlp.pipeline[-1][0] == name1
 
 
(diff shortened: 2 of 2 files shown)


adrianeboyd:
I feel like defaulting to `first=True` for `last=False` is a bit unexpected. What about just rejecting this instead?

Where has this come up?

svlandeg:
> I feel like defaulting to `first=True` for `last=False` is a bit unexpected. What about just rejecting this instead?
> 
> Where has this come up?

In tests I was running myself. It feels like a perfectly reasonable thing to do, considering the abundance of options we give with setting booleans on `add_pipe` and if "first" is unexpected, then the user can specify it better. But setting it "last" when you specifically say "don't set it last" feels like a bug to me.

adrianeboyd:
Why not second-to-last then? I don't really like second-to-last either, I just don't see how `first` would be at all expected?

svlandeg:
> Why not second-to-last then? I don't really like second-to-last either, I just don't see how `first` would be at all expected?

Honestly, I don't think it's any more unexpected than setting "first=False" and getting "last=True" - this is the default behaviour and might be unexpected as well if you haven't read the docs that specify the default behaviour.

adrianeboyd:
I think appending is much less unexpected as a default? I think someone who has provided only `last=False` has not provided enough information and an error would be better.

svlandeg:
I'll likely redo the PR to fix this more properly in `v4`, and leave `master` alone 🤷

honnibal:
I'd expect `last=False` to be anywhere but last, so if we add it I would make it second last. But it feels oddly vague -- like the behaviour could vary by implementation or something. I think that's why we just declared it an illegal value.

The other thing is, if you set `last=False`, should it raise if the pipeline is empty? Because it will indeed be last in that case.

svlandeg:
> I'd expect `last=False` to be anywhere but last, so if we add it I would make it second last. But it feels oddly vague -- like the behaviour could vary by implementation or something. I think that's why we just declared it an illegal value.

That's the thing - it's not an illegal value right now on `master`. Setting `last=False` simply gets ignored, and the component is added last.
```
    nlp = spacy.blank("en")
    nlp.add_pipe("parser")
    nlp.add_pipe("ner")
    nlp.add_pipe("tagger", last=False)
    print(nlp.pipe_names)
```
> ['parser', 'ner', 'tagger']

This is what this PR aimed to fix. Yes, it's underdefined whether we should put the `tagger` first or second-to-last and that's a bit arbitrary, but simply ignoring the given option and doing it anyway feels worse.

svlandeg:
Closing this PR as this functionality was fixed for `v4`: https://github.com/explosion/spaCy/pull/12793

PR #12369, opened 2023-03-06
Add more information to custom code docs

thomashacker:
## Description

This `PR` adds a small warning section to the `usage/training` docs that mentions that unpackaged models, trained with custom code, do not internally contain custom code, but need to be imported into the script before loading in the models. 

### Types of change

Documentation

## Checklist


website/docs/usage/training.mdx
```diff
@@ -758,6 +758,14 @@ any custom architectures, functions or
 your pipeline and registered when it's loaded. See the documentation on
 [saving and loading pipelines](/usage/saving-loading#models-custom) for details.
 
+<Infobox variant="warning">
+
+Please note that unpackaged models produced by `spacy train` do **not include
+any custom code**, you need to import the code to your script before loading in
```
adrianeboyd, line 764:
```suggestion
any custom code**, you need to import the code in your script before loading
```

adrianeboyd:
This is a good addition! I'm thinking a little about whether it should also include an explanation related to `spacy.load("/path")` vs. `spacy.load("package_name") in terms of whether code can be included/loaded. I wonder if we can link to a section with more details, but I'm not sure there is one in the docs?

thomashacker:
We could link to this section https://spacy.io/api/top-level#spacy.load
It mentions that installed packages run their own `.load()` method. We could mention that this is where the custom code gets imported.

rmitsch:
> We could link to this section https://spacy.io/api/top-level#spacy.load
It mentions that installed packages run their own .load() method. We could mention that this is where the custom code gets imported.

I think this doesn't cover the difference between `spacy.load("/path")` and `spacy.load("package_name")` that Adriane referred to. [This](https://spacy.io/usage/saving-loading#models) could also be linked and has some more info, but also doesn't explicitly discuss a difference. Maybe we should extend one of these two doc pages and then link them from the new doc section in this PR?  @adrianeboyd 

Otherwise LGTM.

adrianeboyd:
I think it would make sense to expand the top paragraph at https://spacy.io/api/top-level#spacy.load to clarify that loading from a package name includes loading any code packaged with it, but loading from data directories only loads data and no additional code?

And then link that?

adrianeboyd:
Ah, this should also be switched to `master`. I'll switch it here, but it will probably need to be redone not to include unintended commits (force-pushing to fix it is fine).

PR #12375, opened 2023-03-07
Add `.ents_spangroup` to `Doc`

rmitsch:
## Description

Add property `.ents_spangroup` to `Doc`.

Split off of https://github.com/explosion/spaCy/pull/11669.

### Types of change

New feature.

## Checklist

### DIFF
--- a/website/docs/api/doc.mdx
+++ b/website/docs/api/doc.mdx
@@ -752,22 +752,23 @@ The L2 norm of the document's vector representation.
 
 ## Attributes {id="attributes"}
 
-| Name                                       | Description                                                                                                                                    |
-| ------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------- |
-| `text`                                     | A string representation of the document text. ~~str~~                                                                                          |
-| `text_with_ws`                             | An alias of `Doc.text`, provided for duck-type compatibility with `Span` and `Token`. ~~str~~                                                  |
-| `mem`                                      | The document's local memory heap, for all C data it owns. ~~cymem.Pool~~                                                                       |
-| `vocab`                                    | The store of lexical types. ~~Vocab~~                                                                                                          |
-| `tensor`                                   | Container for dense vector representations. ~~numpy.ndarray~~                                                                                  |
-| `user_data`                                | A generic storage area, for user custom data. ~~Dict[str, Any]~~                                                                               |
-| `lang`                                     | Language of the document's vocabulary. ~~int~~                                                                                                 |
-| `lang_`                                    | Language of the document's vocabulary. ~~str~~                                                                                                 |
-| `user_hooks`                               | A dictionary that allows customization of the `Doc`'s properties. ~~Dict[str, Callable]~~                                                      |
-| `user_token_hooks`                         | A dictionary that allows customization of properties of `Token` children. ~~Dict[str, Callable]~~                                              |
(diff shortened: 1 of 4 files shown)


svlandeg (approved):
Looks good to me, just want to run this by Adriane before we merge.

website/docs/api/doc.mdx
```diff
@@ -752,22 +752,23 @@ The L2 norm of the document's vector representation.
...
+| `user_token_hooks`                            | A dictionary that allows customization of properties of `Token` children. ~~Dict[str, Callable]~~                                              |
+| `user_span_hooks`                             | A dictionary that allows customization of properties of `Span` children. ~~Dict[str, Callable]~~                                               |
+| `has_unknown_spaces`                          | Whether the document was constructed without known spacing between tokens (typically when created from gold tokenization). ~~bool~~            |
+| `_`                                           | User space for adding custom [attribute extensions](/usage/processing-pipelines#custom-components-attributes). ~~Underscore~~                  |
+| `activations` <Tag variant="new">4.0</Tag>    | A dictionary of activations per trainable pipe (available when the `save_activations` option of a pipe is enabled). ~~Dict[str, Option[Any]]~~ |
+| `ents_spangroup` <Tag variant="new">4.0</Tag> | All entitity `Span` instances (as stored in `.ents`) as [`SpanGroup`](/api/spangroup). ~~SpanGroup~~                                           |
```
adrianeboyd, line 771: This isn't a `Doc` attribute, it should be added as a property similar to `Doc.ents`.

adrianeboyd:
My main feeling is that this feels like it might be better as separate utility method rather than a `Doc` property, and the name doesn't seem especially clear to me?

adrianeboyd:
On the other hand it's not that different as `Doc.ents`, but naming feels kind of weird / arbitrary, like `Doc.ents` isn't `Doc.ents_tuple`. I will think about it a bit...

adrianeboyd:
Is the goal to be able to add this to `doc.spans` or to use it separately?

rmitsch:
To use it separately. This change is motivated by `EntityLinker` expecting `SpanGroup` instances instead of `Iterable[Span]` from v4 onwards, so @svlandeg suggested adding something to obtain entities as `SpanGroup` directly (IIRC).

adrianeboyd:
Doing this just for `Doc.ents` feels overly specific. It feels like it ought to be more like `spacy.util.as_spangroup(Union[Iterable[Span], whatever-else-you-can-convert, etc.])`.

adrianeboyd:
I don't want you to be blocked on this. How about we add `Doc._ents_spangroup` for now and plan to think about the more general use of `SpanGroup` in the future?

rmitsch:
We can do that, but it's not blocking me / the EL stuff. It's purely for convenience' sake, it's fine to just do
```python
from spacy.tokens import SpanGroup
sg = SpanGroup(doc, spans=doc.ents)
```
wherever I need it. Is `Docs._ents_spangroup` really ok for you or would you rather figure out a proper solution before merging?

adrianeboyd:
Okay, then let's just mark this as a draft for now.

honnibal:
I'd lean slightly against this I think. I think it's worth being pretty protective of the API surface on the `Doc`. The alternative code isn't bad I don't think.

adrianeboyd:
Let's close this for now. We can always revisit it in the future.

PR #12380, opened 2023-03-07
TEST: Doc.ents as SpanGroup

adrianeboyd:
## Description

Overview of required changes to support `SpanGroup` rather than `Tuple[Span]`:

* implement slice for `SpanGroup`
* return `SpanGroup` for `SpanGroup + x` or `x + SpanGroup` rather than refusing to concatenate (currently without good error handling)

Side effects:

* if appending to `Doc.ents`, only `Iterable[Span]` is supported rather than other formats like raw entity tuples from matcher results, but you can still assign mixed data in any of the currently supported formats to `Doc.ents`
  * for the `Matcher` case, `as_spans` provides a good alternative

### Types of change

?

## Checklist


adrianeboyd:
I'm surprised that no one has complained about slicing span groups.

spacy/tokens/span_group.pyx
```diff
@@ -103,6 +106,10 @@ cdef class SpanGroup:
 
         DOCS: https://spacy.io/api/spangroup#getitem
         """
+        if isinstance(i, slice):
+            start, stop = util.normalize_slice(len(self), i.start, i.stop, i.step)
+            spans = [self[i] for i in range(start, stop)]
+            return SpanGroup(self.doc, spans=spans)
```
adrianeboyd, line 112: This is probably not the fastest way to do this.

adrianeboyd: And I suppose we could support `step` here, but I'm not sure why you'd want to do this.

adrianeboyd:
And if we did this, we'd want to do it for all the `Tuple[Span]` cases from #12288.

spacy/tests/matcher/test_matcher_logic.py
```diff
@@ -116,7 +119,8 @@ def test_issue242(en_tokenizer):
     with pytest.raises(ValueError):
         # One token can only be part of one entity, so test that the matches
         # can't be added as entities
-        doc.ents += tuple(matches)
+        matches_spans = matcher(doc, as_spans=True)
+        doc.ents += tuple(matches_spans)
```
adrianeboyd, line 123:
```suggestion
        doc.ents += matches_spans
```

spacy/tokens/span_group.pyi
```diff
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Iterable, Optional
+from typing import Any, Dict, Iterable, Optional, Union, overload
```
adrianeboyd, line 1:
```suggestion
from typing import Any, Dict, Iterable, Optional, overload
```

adrianeboyd:
So the current `SpanGroup` design with a weakref for the doc is basically a deal breaker for this proposal:


```python
import spacy
nlp = spacy.blank("en")
ruler = nlp.add_pipe("entity_ruler")
ruler.add_patterns([{"label": "A", "pattern": "the"}])
print(nlp("the cat").ents)
```

```
  File "<stdin>", line 1, in <module>
  File "spacy/tokens/span_group.pyx", line 62, in spacy.tokens.span_group.SpanGroup.__repr__
    return str(list(self))
  File "spacy/tokens/span_group.pyx", line 179, in __iter__
    yield self[i]
  File "spacy/tokens/span_group.pyx", line 114, in spacy.tokens.span_group.SpanGroup.__getitem__
    return Span.cinit(self.doc, self.c[i])
  File "spacy/tokens/span_group.pyx", line 73, in spacy.tokens.span_group.SpanGroup.doc.__get__
    raise RuntimeError(Errors.E865)
RuntimeError: [E865] A SpanGroup is not functional after the corresponding Doc has been garbage collected. To keep using the spans, make sure that the corresponding Doc object is still available in the scope of your function.
```

I like it otherwise, but this (by itself) is definitely too breaking.

honnibal:
I see the thinking here.

Is it too awkward to make `SpanGroup` take an argument that tells us whether to hold a real ref or a weakref?

svlandeg:

PR #12385, opened 2023-03-08
Add links in website and readme for survey

victorialslocum:
## Description

Added links to the spaCy user survey in a bottom bar alert for the website and in the README. 

### Types of change

Enhancement, additional documentation

## Checklist

### DIFF
--- a/website/src/templates/index.js
+++ b/website/src/templates/index.js
@@ -25,6 +25,11 @@ const AlertSpace = ({ nightly, legacy }) => {
     const isOnline = useOnlineStatus()
     return (
         <>
+            {isOnline && (
+                <Alert title="💥 We'd love to learn more about your experience with spaCy!">
+                    <Link to="https://form.typeform.com/to/aMel9q9f">Take our survey here.</Link>
+                </Alert>
+            )}
             {nightly && (
                 <Alert
                     title="You're viewing the pre-release docs."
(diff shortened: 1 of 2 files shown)


website/src/templates/index.js
```diff
@@ -25,6 +25,11 @@ const AlertSpace = ({ nightly, legacy }) => {
     const isOnline = useOnlineStatus()
     return (
         <>
+            {isOnline && (
+                <Alert title="💥 We'd love to learn more about your experience with spaCy!">
+                    <Link to="https://form.typeform.com/to/aMel9q9f">Take our survey here.</Link>
```
svlandeg, line 30:
```suggestion
                    <Link to="https://form.typeform.com/to/aMel9q9f">Take our 2023 survey here.</Link>
```
Maybe this makes it clear this is a current, limited opportunity?

victorialslocum: To me, this kind of makes it feel like it will be available for the rest of the year. I'm not sure of a better option though

svlandeg: Oh yea, that's a good point. Hm. Maybe just leave as is then? We'd likely run this for about a month, until April 10th.

victorialslocum: I would say leave as is, or change to something like `<LINK>Take our survey here</LINK> until April 10th`. I think I prefer as is though.

PR #12398, opened 2023-03-10
Fix EL failure with sentence-crossing entities

rmitsch:
### Types of change

`EntityLinker` crashes when `incl_context is True and `len(list(ent.sents)) > 1` and the first of the sentences in `ent.sents` (which is the one used in the context calculation) doesn't have an entity. 

This currently adds a test to reproduce this behavior. A fix will be added later - probably in this PR, hence this is marked as draft.

## Checklist

### DIFF
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -476,26 +476,7 @@ class EntityLinker(TrainablePipe):
                 for j, ent in enumerate(ent_batch):
                     assert hasattr(ent, "sents")
                     sents = list(ent.sents)
-                    # Note: the last sentence associated with an sentence-crossing entity isn't complete. E. g. if you
-                    # have "Mahler's Symphony No. 8 was beautiful", the entity being "No. 8", ent.sents would be:
-                    #   1. "Mahler's Symphony No."
-                    #   2. "8"
-                    # whereas doc.sents would be:
-                    #   1. "Mahler's Symphony No."
-                    #   2. "8 was beautiful"
-                    # This makes it tricky to receive the last sentence by indexing doc.sents - hence we use an offset
-                    # to determine sent_indices[1].
-                    sent_indices = (
-                        (
-                            sentences.index(sents[0]),
-                            sentences.index(sents[0]) + len(sents) - 1,
-                        )
-                        if len(sents) > 1
-                        else (
-                            sentences.index(ent.sent),
-                            sentences.index(ent.sent),
-                        )
-                    )
+                    sent_indices = (sentences.index(sents[0]), sentences.index(sents[-1]))
                     assert all([si >= 0 for si in sent_indices])
 
                     if self.incl_context:
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -484,10 +484,13 @@ class EntityLinker(TrainablePipe):
                     #   2. "8 was beautiful"
                     # This makes it tricky to receive the last sentence by indexing doc.sents - hence we use an offset
                     # to determine sent_indices[1].
-                    sent_indices = (
-                        sentences.index(sents[0]),
-                        sentences.index(sents[0]) + len(sents) - 1,
-                    )
+                    if len(sents) > 1:
+                        sent_indices = (
+                            sentences.index(sents[0]),
+                            sentences.index(sents[0]) + len(sents) - 1,
+                        )
+                    else:
+                        sent_indices = (sentences.index(ent.sent), sentences.index(ent.sent))
                     assert all([si >= 0 for si in sent_indices])
 
                     if self.incl_context:
--- a/spacy/pipeline/entity_linker.py
+++ b/spacy/pipeline/entity_linker.py
@@ -474,18 +474,29 @@ class EntityLinker(TrainablePipe):
 
                 # Looping through each entity in batch (TODO: rewrite)
                 for j, ent in enumerate(ent_batch):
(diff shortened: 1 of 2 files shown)


spacy/pipeline/entity_linker.py
```diff
@@ -474,18 +474,29 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
+                    # have "Mahler's Symphony No. 8 was beautiful", the entity being "No. 8", ent.sents would be:
+                    #   1. "Mahler's Symphony No."
+                    #   2. "8"
+                    # whereas doc.sents would be:
+                    #   1. "Mahler's Symphony No."
+                    #   2. "8 was beautiful"
+                    # This makes it tricky to receive the last sentence by indexing doc.sents - hence we use an offset
+                    # to determine sent_indices[1].
+                    sent_indices = (sentences.index(sents[0]), sentences.index(sents[0]) + len(sents) - 1)
+                    assert all([si >= 0 for si in sent_indices])
```
rmitsch, line 491: We could add an assertion checking that `sent_indices` is a strictly increasing sequence.

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -1219,3 +1219,47 @@ def test_span_maker_forward_with_empty():
     # just to get a model
     span_maker = build_span_maker()
     span_maker([doc1, doc2], False)
+
+
+def test_sentence_crossing_ents():
```
svlandeg, line 1224: Could you incorporate this test into `test_issue7065_b`? It's basically the same except for which entities to define, so `entities` and `links` can be parametrized.

spacy/pipeline/entity_linker.py
```diff
@@ -474,18 +474,40 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
+                    sents = list(ent.sents)
+                    # Note: the last sentence associated with an sentence-crossing entity isn't complete. E. g. if you
+                    # have "Mahler's Symphony No. 8 was beautiful", the entity being "No. 8", ent.sents would be:
+                    #   1. "Mahler's Symphony No."
+                    #   2. "8"
+                    # whereas doc.sents would be:
+                    #   1. "Mahler's Symphony No."
+                    #   2. "8 was beautiful"
+                    # This makes it tricky to receive the last sentence by indexing doc.sents - hence we use an offset
+                    # to determine sent_indices[1].
```
svlandeg, line 487: This looks like a bug in `span.ents` - the second/last sentence shouldn't be partial. It looks like there's a bug in the indexing near the end of the doc. If you would a sentence after "beautiful", you'll see that the last sentence will in fact be complete. This bug only shows up when the entity crosses boundaries into the very last sentence, in which case the last sentence is truncated to the end of the span.

Let me know if you want to look into solving this bug (in a separate PR) or whether you'd like someone else to look into it.

rmitsch: Ah, ok. I'd want to look into this myself, if we're ok with it not happening right away (due to other EL stuff I'd like to take of first).

svlandeg: The thing is that I don't want this PR to be unnecessarily complicated because it's working around a bug in another part of the code base, you know?

rmitsch: Right, so you think it'd be better to address the sentence indexing bug before we implement the solution for the EL bug?

svlandeg: Alternatively we can also write this PR assuming that the `sents` work correctly, but then we really shouldn't forget to follow up with the other fix.

rmitsch: Yeah, it's better to fix the sentence indexing problem first then. I should have a look at https://github.com/explosion/spaCy/blob/master/spacy/tokens/span.pyx#L427 then, no?

rmitsch: [PR](https://github.com/explosion/spaCy/pull/12405) with test and fix.

svlandeg: Merged!

rmitsch:
Simplified the `ent.sents` handling in https://github.com/explosion/spaCy/pull/12398/commits/9b6f47f1acdd90ccaedc4baae1aacf9c176f9bd9 and https://github.com/explosion/spaCy/pull/12398/commits/fc0f954d9b8b85b5821b9420042ab55b7c1ef458 as `Span.sents` has been [fixed](https://github.com/explosion/spaCy/pull/12405) and doesn't produce partial sentences anymore.

spacy/pipeline/entity_linker.py
```diff
@@ -474,18 +474,24 @@ def predict(self, docs: Iterable[Doc]) -> List[str]:
...
                 for j, ent in enumerate(ent_batch):
-                    sent_index = sentences.index(ent.sent)
-                    assert sent_index >= 0
+                    assert hasattr(ent, "sents")
+                    sents = list(ent.sents)
+                    sent_indices = (
+                        sentences.index(sents[0]),
+                        sentences.index(sents[-1]),
+                    )
+                    assert sent_indices[-1] >= sent_indices[0] >= 0
```
svlandeg, line 483:
```suggestion
                    assert 0 >= sent_indices[0] >= sent_indices[1]
```
This should be the same but feels more natural to me, as the code below uses these exact formulations. We know that `sent_indices` is always of length 2 so we might as well just index directly with `0` and `1`.

svlandeg (approved):
That all looks good!

PR #12427, opened 2023-03-15
Add info that Vocab and StringStore are not static in docs

thomashacker:
## Description

Adds a warning to the `Vocab` and `StringStore` docs that their size is not static and increases with every newly seen token. This adjustment resulted from this [discussion](https://github.com/explosion/spaCy/discussions/10015).

### Types of change

Documentation

## Checklist

### DIFF
--- a/website/docs/api/stringstore.mdx
+++ b/website/docs/api/stringstore.mdx
@@ -10,8 +10,8 @@ different `StringStores`.
 
 <Infobox variant="warning">
 
-Please note that `StringStore` instances are not static. It increases in size as 
-texts with new tokens are processed.
+Note that a `StringStore` instance is not static. It increases in size as texts
+with new tokens are processed.
 
 </Infobox>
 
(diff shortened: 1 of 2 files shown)


website/docs/api/stringstore.mdx
```diff
@@ -8,6 +8,13 @@ Look up strings by 64-bit hashes. As of v2.0, spaCy uses hash values instead of
 integer IDs. This ensures that strings always map to the same ID, even from
 different `StringStores`.
 
+<Infobox variant="warning">
+
+Please note that the `StringStore` size is not static and increases as texts are
+processed and new tokens are seen.
```
rmitsch, line 14: I feel like this is clearer, but feel free to ignore & resolve if you disagree.
```suggestion
Please note that `StringStore` instances are not static. It increases in size as 
texts with new tokens are processed.
```

rmitsch (approved):

PR #12429, opened 2023-03-16
Remove redundant strings.add for Doc.char_span

adrianeboyd:
## Description

Remove redundant `strings.add` for `Doc.char_span`.

### Types of change

?

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyx
+++ b/spacy/tokens/doc.pyx
@@ -544,10 +544,6 @@ cdef class Doc:
 
         DOCS: https://spacy.io/api/doc#char_span
         """
-        if not isinstance(label, int):
-            label = self.vocab.strings.add(label)
-        if not isinstance(kb_id, int):
-            kb_id = self.vocab.strings.add(kb_id)
         alignment_modes = ("strict", "contract", "expand")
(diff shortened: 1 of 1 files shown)


adrianeboyd:
A larger question here is whether we want to support `Span.kb_id` and `Span.span_id` as ints that aren't in the strings. As in this test (which I added, probably naively):

https://github.com/explosion/spaCy/blob/96b61d06712e8de7dd59a6ab8ae153893e5d7d2c/spacy/tests/parser/test_ner.py#L602

But then `span.span_id_` is going to raise errors in the future.

If not, we could extend these errors in `Span.__cinit__` to `kb_id` and `span_id`, but this is slightly breaking:

https://github.com/explosion/spaCy/blob/96b61d06712e8de7dd59a6ab8ae153893e5d7d2c/spacy/tokens/span.pyx#L109-L110

adrianeboyd:
> A larger question here is whether we want to support `Span.kb_id` and `Span.span_id` as ints that aren't in the strings.

I think the answer to this is: no, this shouldn't be supported because `Span.kb_id_` etc. breaks.

svlandeg (approved):
Good catch!

PR #12436, opened 2023-03-17
CI: Only run test suite once with thinc-apple-ops for macos python 3.11

adrianeboyd:
## Description

Only run test suite once with `thinc-apple-ops` for macos python 3.11. The motivation is to reduce the total CI time when all jobs are run in parallel.

### Types of change

CI.

## Checklist


adrianeboyd:
This is diverging from the azure CI but I didn't want to forget to make this change after the azure CI is disabled.

And... this did not run as intended.

adrianeboyd:
Okay, this gets the total time down to 22 min.

PR #12439, opened 2023-03-17
Add Span.kb_id and Span.id strings to Doc/DocBin serialization

adrianeboyd:
## Description

Add `Span.kb_id` and `Span.id` strings to `Doc`/`DocBin` serialization

* Raise errors for missing strings in `Span` constructor.
* Switch tests that use int-only `kb_id/id` to strings.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tests/doc/test_json_doc_conversion.py
+++ b/spacy/tests/doc/test_json_doc_conversion.py
@@ -313,7 +313,7 @@ def test_json_to_doc_spans(doc):
     """Test that Doc.from_json() includes correct.spans."""
     doc.spans["test"] = [
         Span(doc, 0, 2, label="test"),
-        Span(doc, 0, 1, label="test", kb_id=7),
+        Span(doc, 0, 1, label="test", kb_id="7"),
     ]
     json_doc = doc.to_json()
     new_doc = Doc(doc.vocab).from_json(json_doc, validate=True)
--- a/spacy/tests/parser/test_ner.py
+++ b/spacy/tests/parser/test_ner.py
@@ -599,22 +599,22 @@ def test_overfitting_IO(use_upper):
     # test that kb_id is preserved
     test_text = "I like London and London."
     doc = nlp.make_doc(test_text)
-    doc.ents = [Span(doc, 2, 3, label="LOC", kb_id=1234)]
+    doc.ents = [Span(doc, 2, 3, label="LOC", kb_id="1234")]
     ents = doc.ents
     assert len(ents) == 1
     assert ents[0].text == "London"
     assert ents[0].label_ == "LOC"
-    assert ents[0].kb_id == 1234
+    assert ents[0].kb_id_ == "1234"
     doc = nlp.get_pipe("ner")(doc)
     ents = doc.ents
     assert len(ents) == 2
     assert ents[0].text == "London"
     assert ents[0].label_ == "LOC"
-    assert ents[0].kb_id == 1234
-    # ent added by ner has kb_id == 0
+    assert ents[0].kb_id_ == "1234"
+    # ent added by ner has kb_id == ""
     assert ents[1].text == "London"
     assert ents[1].label_ == "LOC"
-    assert ents[1].kb_id == 0
+    assert ents[1].kb_id_ == ""
 
 
 def test_beam_ner_scores():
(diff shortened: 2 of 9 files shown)


spacy/tokens/doc.pyx
```diff
@@ -1350,6 +1350,8 @@ cdef class Doc:
         for group in self.spans.values():
             for span in group:
                 strings.add(span.label_)
+                strings.add(span.kb_id_)
+                strings.add(span.id_)
```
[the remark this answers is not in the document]
adrianeboyd: Tests, yes. I don't see a strong reason to have separate test marked with the issue, though?

rmitsch (approved):

spacy/tests/doc/test_json_doc_conversion.py
```diff
@@ -313,7 +313,7 @@ def test_json_to_doc_spans(doc):
     """Test that Doc.from_json() includes correct.spans."""
     doc.spans["test"] = [
         Span(doc, 0, 2, label="test"),
-        Span(doc, 0, 1, label="test", kb_id=7),
+        Span(doc, 0, 1, label="test", kb_id="7"),
```
svlandeg, line 316: Why don't we test for both?

spacy/tokens/span.pyx
```diff
@@ -107,7 +107,11 @@ cdef class Span:
         if isinstance(span_id, str):
             span_id = doc.vocab.strings.add(span_id)
         if label not in doc.vocab.strings:
-            raise ValueError(Errors.E084.format(label=label))
+            raise ValueError(Errors.E084.format(name="label", value=label))
+        if kb_id not in doc.vocab.strings:
+            raise ValueError(Errors.E084.format(name="kb_id", value=kb_id))
+        if span_id not in doc.vocab.strings:
+            raise ValueError(Errors.E084.format(name="span_id", value=span_id))
```
svlandeg, line 114: Could we have these as `else` clauses instead? It seems wasteful to check if they were already a `str` and we had just added it specifically?

spacy/tests/parser/test_ner.py
```diff
@@ -599,22 +599,22 @@ def test_overfitting_IO(use_upper):
...
     assert ents[0].text == "London"
     assert ents[0].label_ == "LOC"
-    assert ents[0].kb_id == 1234
-    # ent added by ner has kb_id == 0
+    assert ents[0].kb_id_ == "1234"
+    # ent added by ner has kb_id == ""
     assert ents[1].text == "London"
     assert ents[1].label_ == "LOC"
-    assert ents[1].kb_id == 0
+    assert ents[1].kb_id_ == ""
```
svlandeg, line 617: I might be missing something - but I'm not sure I understand why these tests were changed?

adrianeboyd: There's no string corresponding to `1234` so you run into errors. You could have kept `0` because it's predefined as `""`, but it seemed nicer to make it clear that what we care about is setting from a string?

adrianeboyd:
This comment is relevant: https://github.com/explosion/spaCy/pull/12429#issuecomment-1471465218

svlandeg (commented):
We might want to think about this more carefully, and I wouldn't add this PR to the upcoming bugfix release. The [API docs](https://spacy.io/api/span#init) currently state that the `kb_id` can be of `Union[str, int]` and it's not necessarily clear to users that the `int` means a hash from our `StringStore` - it seems reasonable to assume that you can "just" use ints for KB IDs. Do we need this option of passing in the hash in the first place? Are users doing that? Or are they assuming that passing `1234` will actually be parsed into `"1234"` internally? Why don't we do that instead, actually?

We need to document this better - especially with respect to the KB ID here. It'll be a common use-case where such IDs are in fact `int`'s and things will become confusing to the user. The fact that we want to change our test suite so much here, means I worry that user code might be impacted in the same way.

svlandeg:
Yea I think that's the best way forward for now: assume that if users are passing a non-existing hash, they kind of wanted to add an `int` as an actual KB ID and we take care of this so it doesn't break in the short term.

Then in the longer term I agree we should clean this up and probably disallow passing `int`'s in the first place.

svlandeg:
Follow-up PR: https://github.com/explosion/spaCy/pull/12493

PR #12459, opened 2023-03-23
Add model-last saving mechanism to pretraining

thomashacker:
## Description

This PR takes inspiration from the training loop to save the last epoch of pretraining as `model_last.bin` instead of `model<last_epoch>.bin`. The PR aims to make it easier for users to work with pretrained weights in an automated workflow (e.g. spaCy projects) and reduce manual adjustment based on the set `max_epochs` in the training config.

### Types of change

Feature

## Checklist


thomashacker:
I'll add some unit tests. If we decide to implement this feature, do we want to mention this in the documentation?

spacy/training/pretrain.py
```diff
@@ -60,10 +60,14 @@ def pretrain(
     row_settings = {"widths": (3, 10, 10, 6, 4), "aligns": ("r", "r", "r", "r", "r")}
     msg.row(("#", "# Words", "Total Loss", "Loss", "w/s"), **row_settings)
 
-    def _save_model(epoch, is_temp=False):
+    def _save_model(epoch, is_temp=False, is_last=False):
         is_temp_str = ".temp" if is_temp else ""
         with model.use_params(optimizer.averages):
-            with (output_dir / f"model{epoch}{is_temp_str}.bin").open("wb") as file_:
+            if is_last:
+                save_path = output_dir / f"model_last.bin"
```
adrianeboyd, line 67: I think it would be nicer if the name is more similar to `spacy train`, but I don't have really strong feelings about it if there's a good reason to keep `model_last`?

```suggestion
                save_path = output_dir / f"model-last.bin"
```

thomashacker: Oh I didn't realize it was a "-"! I agree; let's change it

adrianeboyd:
And I would definitely add this to the docs.

adrianeboyd:
The CI seems to have gotten into a weird state, let me close and reopen this.

website/docs/usage/embeddings-transformers.mdx
```diff
@@ -745,14 +745,17 @@ To benefit from pretraining, your training step needs to know to initialize its
 this by setting `initialize.init_tok2vec` to the filename of the `.bin` file
 that you want to use from pretraining.
 
-A pretraining step that runs for 5 epochs with an output path of `pretrain/`, as
-an example, produces `pretrain/model0.bin` through `pretrain/model4.bin`. To
-make use of the final output, you could fill in this value in your config file:
```
adrianeboyd, line 750: The default still produces all the intermediate files, right? I think I'd leave the first sentence in as something like:

> A pretraining step that runs for 5 epochs with an output path of `pretrain/`, as an example, produces `pretrain/model0.bin` through `pretrain/model4.bin` plus a copy of the final output as `pretrain/model-last.bin`.

website/docs/usage/embeddings-transformers.mdx
```diff
@@ -745,14 +745,17 @@ To benefit from pretraining, your training step needs to know to initialize its
...
+Similar to training, pretraining produces a `model-last.bin` file which is the
+last iteration of the trained weights which you can use to initialize your
+`tok2vec` layer. Additionally, you can configure `n_save_epoch` to tell
+pretraining in which epoch interval it should save the current training
+progress. To make use of the final output, you could fill in this value in your
+config file:
```
adrianeboyd, line 753:
```suggestion
A pretraining step that runs for 5 epochs with an output path of `pretrain/`, as
an example, produces `pretrain/model0.bin` through `pretrain/model4.bin` plus a
copy of the last iteration as `pretrain/model-last.bin`. Additionally, you can
configure `n_save_epoch` to tell pretraining in which epoch interval it should
save the current training progress. To use the final output to initialize your
`tok2vec` layer, you could fill in this value in your config file:
```

svlandeg (commented):
Looks good to me.

The one thing I wonder about is whether this could influence existing workflows somehow, like users iterating over all `.bin` files in the output folder, assuming the names all look like `modelX.bin` with `X` an `int`, and that that could now fail when they encounter `model-last.bin`.

Could we make this opt-in for the upcoming bugfix release? Or at least provide an easy switch to turn this off if users don't want the additional duplicate file?

PR #12465, opened 2023-03-24
NER: Ensure zero-cost sequence with sentence split in entity

danieldk:
## Description

If we use a sentence splitter as one of the annotating components during training, an entity can become split in the predicted `Doc`. Before this change, training would fail, because no zero-cost transition sequence could be found.

This fixes two scenarios:

1. When the gold action is `B` and a split occurs after the current token, the `BEGIN` action is invalid. However, this was the only possible zero-cost action. This change makes `OUT` a zero-cost action in this case.
2. When the gold action is `I` and a split occurs after the current token, the `IN` action is invalid, removing the only zero-cost action. This change makes `LAST` a zero-cost action, so that the entity can be properly closed.

### Types of change

Bugfix

## Checklist

### DIFF
--- a/spacy/pipeline/_parser_internals/ner.pyx
+++ b/spacy/pipeline/_parser_internals/ner.pyx
@@ -120,6 +120,10 @@ cdef bint _entity_is_sunk(const StateC* state, Transition* golds) nogil:
         return False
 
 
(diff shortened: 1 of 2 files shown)


danieldk:
One thing that I am not very sure of: maybe in case 1, `U` should also be a zero-cost transition?

PR #12492, opened 2023-03-31
Support registered vectors

adrianeboyd:
## Description

Support registered vectors.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
For example using bpemb vectors (@koaning):

```python
from typing import cast, Callable, Optional
from pathlib import Path
import warnings
from bpemb import BPEmb
from spacy.util import registry
from spacy.vectors import BaseVectors
from spacy.vocab import Vocab
from thinc.api import Ops, get_current_ops
from thinc.backends import get_array_ops
from thinc.types import Floats2d


class BPEmbVectors(BaseVectors):
    def __init__(
        self,
        *,
        strings: Optional[str] = None,
        lang: Optional[str] = None,
        vs: Optional[int] = None,
        dim: Optional[int] = None,
        cache_dir: Optional[Path] = None,
        encode_extra_options: Optional[str] = None,
        model_file: Optional[Path] = None,
        emb_file: Optional[Path] = None,
    ):
        kwargs = {"lang": lang, "vs": vs, "dim": dim}
        if cache_dir is not None:
            kwargs["cache_dir"] = cache_dir
        if encode_extra_options is not None:
            kwargs["encode_extra_options"] = encode_extra_options
        if model_file is not None:
            kwargs["model_file"] = model_file
        if emb_file is not None:
            kwargs["emb_file"] = emb_file
        self.bpemb = BPEmb(**kwargs)
        self.strings = strings
        self.name = repr(self.bpemb)
        self.n_keys = -1
        self.mode = "BPEmb"
        self.to_ops(get_current_ops())

    def __contains__(self, key):
        return True

    def is_full(self):
        return True

    def add(self, key, *, vector=None, row=None):
        warnings.warn(
            (
                "Skipping BPEmbVectors.add: the bpemb vector table cannot be "
                "modified. Vectors are calculated from bytepieces."
            )
        )
        return -1

    def __getitem__(self, key):
        return self.get_batch([key])[0]

    def get_batch(self, keys):
        keys = [self.strings.as_string(key) for key in keys]
        bp_ids = self.bpemb.encode_ids(keys)
        ops = get_array_ops(self.bpemb.emb.vectors)
        indices = ops.asarray(ops.xp.hstack(bp_ids), dtype="int32")
        lengths = ops.asarray([len(x) for x in bp_ids], dtype="int32")
        vecs = ops.reduce_mean(cast(Floats2d, self.bpemb.emb.vectors[indices]), lengths)
        return vecs

    @property
    def shape(self):
        return self.bpemb.vectors.shape

    def __len__(self):
        return self.shape[0]

    @property
    def vectors_length(self):
        return self.shape[1]

    def to_ops(self, ops: Ops):
        self.bpemb.emb.vectors = ops.asarray(self.bpemb.emb.vectors)


@registry.misc("BPEmbVectors.v1")
def create_bpemb_vectors(
    lang: Optional[str],
    vs: Optional[int],
    dim: Optional[int],
    cache_dir: Optional[Path],
    encode_extra_options: Optional[str],
    model_file: Optional[Path],
    emb_file: Optional[Path],
) -> Callable[[Vocab], BPEmbVectors]:
    def bpemb_vectors_factory(vocab: Vocab) -> BPEmbVectors:
        return BPEmbVectors(
            strings=vocab.strings,
            lang=lang,
            vs=vs,
            dim=dim,
            cache_dir=cache_dir,
            encode_extra_options=encode_extra_options,
            model_file=model_file,
            emb_file=emb_file,
        )

    return bpemb_vectors_factory
```

adrianeboyd:
(At least the immediate CI failures are github glitches.)

koaning:
Super cool! Thanks for the ping. 

Just sayin', that bpemb example might serve as a great plugin, if only for the language support, but also because it seems to be a nice tangible example of an implementation.

Something like `spacy-bpemb`?

adrianeboyd:
Yeah, it would definitely make sense to publish this kind of thing in an extra package and we could also include support for `finalfusion` and other similar packages. (Do you have any other recommendations? Implementing this for bpemb was not at all difficult.)

We'd have to decide whether we want to support internal serialization or not. And I wouldn't know what to call the package because something like `spacy-vectors` sounds pretty ambiguous.

koaning:
Maybe something like `spacy-embeddings`? Or `spacy-representations`? 

The main idea that I have is that we may want to consider embeddings that go beyond words. Maybe sentences? I've made [an "embeddings for sklearn" project](https://github.com/koaning/embetter) and the main embeddings that I end up using are the ones from [sentence-bert](https://sbert.net/). I also recall using the universal sentence encoder a lot, but supporting `tfhub` was such a pain that I decided not supporting it anymore.

Another idea might be to have projects that are specific. The downside of having `spacy-embeddings` is that folks might want us to implement/maintain this months unicorn vectors. It could be nicer to perhaps have `spacy-bpemb`, `spacy-sentence-encoders`, etc. That way, if people want to implement their own, they are totally welcome to clone one of our example projects. Seems easier to maintain in the long run.

adrianeboyd:
I'm not sure I would want to maintain a bunch of small packages and we can be selective about which ones we would add to our own package, making it a bit like `spacy-loggers`.

We wouldn't plan to host any embeddings, just as embeddings at least. (As part of a distributed model if we implement internal serialization: maybe.)

adrianeboyd:
Would come back to this after #12625 is merged.

spacy/vocab.pyx
```diff
@@ -289,12 +290,17 @@ cdef class Vocab:
...
+            return self.vectors.shape[1]
+        else:
+            return -1
 
     def reset_vectors(self, *, width=None, shape=None):
         """Drop the current vector table. Because all vectors must be the same
         width, you have to call this to change the size of the vectors.
         """
+        if not isinstance(self.vectors, Vectors):
+            raise ValueError(Errors.E849.format("reset", vectors_type=type(self.vectors)))
```
svlandeg, line 303:
```suggestion
            raise ValueError(Errors.E849.format(action="reset", vectors_type=type(self.vectors)))
```

spacy/vocab.pyx
```diff
@@ -357,6 +365,8 @@ cdef class Vocab:
 
         DOCS: https://spacy.io/api/vocab#prune_vectors
         """
+        if not isinstance(self.vectors, Vectors):
+            raise ValueError(Errors.E849.format(action="prune", vectors_type=type(self.vectors)))
         if self.vectors.mode != VectorsMode.default:
```
svlandeg, line 370: These three methods that now raise `E849`, could we have them somehow as methods on `(Base)Vectors` directly? So that other implementations have the option to support these operations, and we don't have this awkward type-checking done by the vocab?

spacy/default_config.cfg
```diff
@@ -26,6 +26,9 @@ batch_size = 1000
 [nlp.tokenizer]
 @tokenizers = "spacy.Tokenizer.v1"
 
+[nlp.vectors]
+@misc = "spacy.Vectors.v1"
```
svlandeg, line 30: Should we have a separate registry for them?

adrianeboyd: Yes, I guess we may as well just add `vectors`.

spacy/vocab.pyx
adrianeboyd: I agree that it's clunky, but the implementations in `Vocab` are so tied to the specific `Vectors` implementation that I think this would be difficult to do. (If the `Vocab` versions work with your custom vectors, you may as well be using `Vectors` instead.)

I think if you need pruning, deduplication, etc. for your custom vectors, you could implement this for your vocab/pipeline outside of `Vocab`? I don't think there's any requirement that this needs to be implemented as a `Vocab` method, you'd just need to pass in the full `Vocab` in order to have access to the lexical info, etc. if you need it.

svlandeg: It's the fact that these implementations in `Vocab` are so tied to the specific `Vectors` implementation that made me wonder whether we shouldn't consider moving those implementations to `Vectors` instead. Then like you say you can just pass the `vocab` to that method and let the (custom) vectors do everything else?

adrianeboyd: In general I think they could be separate `util`-style methods instead that take a `Vocab` arg, but I don't want to change the current `Vocab` API (at least not in v3). I don't think it's a huge deal to leave it like it is, since for the most part we're the only ones using these methods internally, and basically only for `init vectors`.

spacy/vocab.pyx
```diff
@@ -289,12 +290,17 @@ cdef class Vocab:
...
+            return self.vectors.shape[1]
+        else:
+            return -1
 
     def reset_vectors(self, *, width=None, shape=None):
         """Drop the current vector table. Because all vectors must be the same
         width, you have to call this to change the size of the vectors.
         """
+        if not isinstance(self.vectors, Vectors):
+            raise ValueError(Errors.E849.format(method="reset_vectors", vectors_type=type(Vectors)))
```
svlandeg, line 303:
```suggestion
            raise ValueError(Errors.E849.format(method="reset_vectors", vectors_type=type(self.vectors)))
```

spacy/vocab.pyx
```diff
@@ -357,6 +365,8 @@ cdef class Vocab:
 
         DOCS: https://spacy.io/api/vocab#prune_vectors
         """
+        if not isinstance(self.vectors, Vectors):
+            raise ValueError(Errors.E849.format(method="prune_vectors", vectors_type=type(Vectors)))
```
svlandeg, line 369:
```suggestion
            raise ValueError(Errors.E849.format(method="prune_vectors", vectors_type=type(self.vectors)))
```

website/docs/api/basevectors.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+Return the number of vectors in the table.
+
+| Name        | Description                                 |
+| ----------- | ------------------------------------------- |
+| **RETURNS** | The number of vectors in the table. ~~int~~ |
+
+## BaseVectors.\_\_contains\_\_ {id="contains",tag="method"}
+
+Check whether there is a vector entry for key.
```
svlandeg, line 47:
```suggestion
Check whether there is a vector entry for the given key.
```

website/docs/usage/embeddings-transformers.mdx
```diff
@@ -632,6 +632,162 @@ def MyCustomVectors(
...
+from thinc.api import Ops, get_current_ops
+from thinc.backends import get_array_ops
+from thinc.types import Floats2d
+
+
+class BPEmbVectors(BaseVectors):
+    def __init__(
+        self,
+        *,
+        strings: Optional[str] = None,
```
svlandeg, line 663:
```suggestion
        strings: Optional[StringStore] = None,
```

website/docs/usage/embeddings-transformers.mdx
```diff
@@ -632,6 +632,162 @@ def MyCustomVectors(
...
+As an example, the following `BPEmbVectors` class implements support for
+[BPEmb subword embeddings](https://bpemb.h-its.org/):
+
+```python
+# requires: pip install bpemb
+from typing import cast, Callable, Optional
+from pathlib import Path
+import warnings
+from bpemb import BPEmb
+from spacy.util import registry
```
svlandeg, line 651:
```suggestion
from spacy.strings import StringStore
from spacy.util import registry
```

svlandeg (approved):
Only found a few more typos, otherwise should be good to merge!

website/docs/api/basevectors.mdx
```diff
@@ -0,0 +1,143 @@
...
+---
+
+`BaseVectors` is an abstract class to support the development of custom vectors
+implementations.
+
+For use in training with [`StaticVectors`](/api/architectures#staticvectors),
+`get_batch` must be implemented. For improved performance, use efficient
+batching in `get_batch` and implement `to_ops` to copy the vector data to the
+current device. See an example custom implementation for
+[BPEMb subword embeddings](/usage/embeddings-transformers#custom-vectors).
```
adrianeboyd, line 16:
```suggestion
[BPEmb subword embeddings](/usage/embeddings-transformers#custom-vectors).
```

PR #12499, opened 2023-04-04
Always redo listener state on pipeline modification

adrianeboyd:
## Description

Always redo listener state on pipeline modification

* Modify `Language._link_components` to reset listener map and re-add all components from scratch.
* Run `Language._link_components` when pipes are added or removed.
* Fix replace listeners for sourced components:
  * Make sure that the source pipeline has the listener state corresponding to the source pipeline and not the new pipeline when listeners are replaced.
  * Remove removal of unused listeners (this is now always updated by `_link_components` at the point where pipes are added).
  * Remove incorrect `replace listeners` after the pipeline is created.
    * For components where `replace_listeners` was specified, the listeners have already been replaced when the component was sourced+added.
    * This incorrectly ran `replace_listeners` twice for components that had `replace_listeners` AND the listened-to component was also sourced into the pipeline (but at this point the listened-to component is irrelevant for those components).

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -793,6 +793,7 @@ class Language:
         pipe_index = self._get_pipe_index(before, after, first, last)
         self._pipe_meta[name] = self.get_factory_meta(factory_name)
         self._components.insert(pipe_index, (name, pipe_component))
+        self._link_components()
         return pipe_component
 
     def _get_pipe_index(
@@ -951,6 +952,7 @@ class Language:
         # Make sure the name is also removed from the set of disabled components
         if name in self.disabled:
             self._disabled.remove(name)
+        self._link_components()
         return removed
 
     def disable_pipe(self, name: str) -> None:
@@ -1675,6 +1677,7 @@ class Language:
         # here :(
         for i, (name1, proc1) in enumerate(self.pipeline):
             if isinstance(proc1, ty.ListenedToComponent):
+                proc1.listener_map = {}
                 for name2, proc2 in self.pipeline[i + 1 :]:
                     proc1.find_listeners(proc2)
 
@@ -1827,6 +1830,12 @@ class Language:
                     source_name = pipe_cfg.get("component", pipe_name)
                     listeners_replaced = False
                     if "replace_listeners" in pipe_cfg:
+                        # HACK: Reset any listened-to components to the listener
+                        # state of the source pipeline for the purpose of
+                        # replacing listeners. The add_pipe below will set
+                        # the state back to the listener state for the new
+                        # pipeline.
+                        source_nlps[model]._link_components()
                         for name, proc in source_nlps[model].pipeline:
                             if source_name in getattr(proc, "listening_components", []):
                                 source_nlps[model].replace_listeners(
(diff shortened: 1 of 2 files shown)


spacy/tests/pipeline/test_tok2vec.py
```diff
@@ -189,9 +189,8 @@ def test_tok2vec_listener(with_vectors):
             tagger.add_label(tag)
 
     # Check that the Tok2Vec component finds it listeners
-    assert tok2vec.listeners == []
-    optimizer = nlp.initialize(lambda: train_examples)
     assert tok2vec.listeners == [tagger_tok2vec]
+    optimizer = nlp.initialize(lambda: train_examples)
```
adrianeboyd, line 193: Now the listeners are also set when the pipe is added rather than on initialize, so it's not empty before initialization anymore. I think you could remove `_link_components` from `initialize` now, but I was trying to change as little as possible for now.

adrianeboyd:
What we actually need is for `source` to make a copy of the sourced component.

spacy/language.py
```diff
@@ -951,6 +952,7 @@ def remove_pipe(self, name: str) -> Tuple[str, PipeCallable]:
         # Make sure the name is also removed from the set of disabled components
         if name in self.disabled:
             self._disabled.remove(name)
+        self._link_components()
```
svlandeg, line 955: We should add a test for the `remove_pipe` functionality and double check that it's removed from the internal `listener_map` (which it probably wasn't before this PR)

spacy/language.py
```diff
@@ -1675,6 +1677,7 @@ def _link_components(self) -> None:
         # here :(
         for i, (name1, proc1) in enumerate(self.pipeline):
             if isinstance(proc1, ty.ListenedToComponent):
+                proc1.listener_map = {}
```
svlandeg, line 1680: Perhaps we could put this functionality behind a `reset` flag that is `False` by default? It's really only relevant when deleting pipes, right? Otherwise we can just have the incremental adding behaviour as before?

I don't care strongly, it probably won't have much impact either way as we're looping through everything anyway, and only preventing `append` when the listener had already been recorded, so the traversal will be equally long either way. It's just that we're kind of reconstructing the same map every time again which feels wasteful.

spacy/language.py
```diff
@@ -1827,6 +1830,12 @@ def from_config(
                     source_name = pipe_cfg.get("component", pipe_name)
                     listeners_replaced = False
                     if "replace_listeners" in pipe_cfg:
+                        # HACK: Reset any listened-to components to the listener
+                        # state of the source pipeline for the purpose of
+                        # replacing listeners. The add_pipe below will set
+                        # the state back to the listener state for the new
+                        # pipeline.
+                        source_nlps[model]._link_components()
```
svlandeg, line 1838: It's nice to be able to remove the whole extra block L1895-L1915, but this hack is giving me a headache. I've been trying to think of ways on how we could do this more cleanly instead... Because right now it feels a bit obscure and difficult to maintain in the future...

svlandeg: Does this PR need the rewrite from the original L1895-L1915 block to this `source_nlps[model]._link_components()` call? It seems like the tests are passing without this edit?

svlandeg: I do think we need to clean this up, but it feels like a separate thing and we probably need to write more unit tests for it, too?

adrianeboyd: You need the source pipeline to have the right listener state or `replace_listeners` (in particular from `spacy-transformers`) does not work correctly. Use this test case: https://github.com/explosion/spacy-transformers/pull/381

adrianeboyd: Ah, wait, I may have this backwards again. Let me send you a test case separately, but it's mainly related to the `replace_listeners` code in `spacy-transformers`, not something that's easy to test in the spacy test suite.

spacy/language.py
adrianeboyd: I think it's such a small number of things and it's only called when pipes are added or removed, so there's no real benefit to making this marginally faster?

spacy/language.py
adrianeboyd: The code in L1895-L1915 now always does nothing because the listener state in the new pipeline has already been configured correctly.

svlandeg: If by "now" you mean "in this PR", I get that. 

My question is: what is a test case that fails with the original block in place vs. the one-liner in this PR? Because if I switch them back, I don't get any failing tests (also not in the `spacy-transformers` PR you linked). It would be helpful to understand the refactor & also to have the test suite cover that.

adrianeboyd: Yeah, it looks like there are no tests for `replace_listener` and `replace_listener_cfg` outside of `spacy-transformers`, and `spacy-transformers` tests don't cover this particular case either because you need a case where multiple components and the `transformer` component are sourced at the same time.

PR #12507, opened 2023-04-06
SpanFinder into spaCy from experimental

kadarakos:
Integrating the `span_finder` from [`spacy-experimental`](https://github.com/explosion/spacy-experimental/tree/master/spacy_experimental/span_finder) into `spaCy`. 

## Description
The `SpanFinder` is a component that suggest spans with a very simple tokenwise classifier: it decides for each token if it could be a start and/or end of a span. The code is largely just copied from `spacy-experimental` with the minimum amount of changes required to run the tests.

I added a couple of comments with the `XXX` annotation to ask a couple of initial questions about the parts I would consider changing after reading the code. 

### Types of change
Moving `spacy-experimental` component into `spaCy`. 

## Checklist


spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,389 @@
...
+width = ${model.tok2vec.embed.width}
+window_size = 1
+maxout_pieces = 3
+depth = 4
+"""
+
+DEFAULT_SPAN_FINDER_MODEL = Config().from_str(span_finder_default_config)["model"]
+DEFAULT_PREDICTED_KEY = "span_candidates"
+# XXX What was this TODO for?
+DEFAULT_TRAINING_KEY = "sc"  # TODO: define in spancat
```
adrianeboyd, line 46: Use the value from `spancat` instead, or if it's not defined separately there, define it there and import it here.

kadarakos:
Not sure why the project clone test fails in the Test Python39Linux?

spacy/pipeline/span_finder.py
```diff
@@ -1,8 +1,7 @@
 from functools import partial
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, cast
 
-from thinc.api import (Config, Model, Ops, Optimizer, get_current_ops,
-                       set_dropout_rate)
+from thinc.api import Config, Model, Ops, Optimizer, get_current_ops, set_dropout_rate
```
kadarakos, line 4: I ran `black` first then `isort` and then the PR got rejected so I had to run `black` again. Do we have to run `isort` first than `black`?

spacy/pipeline/spancat.py
```diff
@@ -71,6 +69,7 @@
 depth = 4
 """
 
+DEFAULT_SPAN_KEY = "sc"
```
adrianeboyd, line 72:
```suggestion
DEFAULT_SPANS_KEY = "sc"
```

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,386 @@
...
+        """
+        reference_truths = self._get_aligned_truth_scores(examples)
+        d_scores = scores - self.model.ops.asarray2f(reference_truths)
+        loss = float((d_scores**2).sum())
+        return loss, d_scores
+
+    def _get_aligned_truth_scores(self, examples) -> List[Tuple[int, int]]:
+        """Align scores of the predictions to the references for calculating the loss"""
+        # TODO: handle misaligned (None) alignments
+        # TODO: handle cases with differing whitespace in texts
```
adrianeboyd, line 286: At least the first point is blocking for inclusion in `spacy`. All pipeline components need to be able to handle misaligned tokenization.

For the second point we could potentially error out and say that this just not supported for this component, but it's something that `Example` supports in general and if it's possible to support it it would be strongly preferred. (You can't use plain character offsets at this point, though, so it would be a pretty major refactoring.)

spacy/tests/pipeline/test_span_finder.py
```diff
@@ -0,0 +1,173 @@
...
+
+    docs = [nlp("This is an example."), nlp("This is the second example.")]
+    docs[0].spans[TRAINING_KEY] = [docs[0][3:4]]
+    docs[1].spans[TRAINING_KEY] = [docs[1][3:5]]
+
+    span_finder = nlp.add_pipe("span_finder", config={"training_key": TRAINING_KEY})
+    nlp.initialize()
+    docs = list(span_finder.pipe(docs))
+
+    # TODO: update hard-coded name
```
adrianeboyd, line 104: A review note about this TODO.

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,386 @@
...
+    def __init__(
+        self,
+        nlp: Language,
+        model: Model[Iterable[Doc], Floats2d],
+        name: str = "span_finder",
+        *,
+        threshold: float = 0.5,
+        max_length: int = 0,
+        min_length: int = 0,
+        # XXX I think this is weird and should be just None like in
```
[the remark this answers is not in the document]
adrianeboyd: Many of the existing components have defaults here so that the behavior of `nlp.add_pipe("name")` didn't change as scorers were added as separate registered functions. It's possible we could do this different for a new component?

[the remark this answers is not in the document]
kadarakos: Sorry. The default `max_positive` is `None` in the `SpanFinder` and I think that makes more sense for `max_length` as well because `max_length=0` is really weird semantically I think that the "unlimited" is 0.

spacy/pipeline/span_finder.py
adrianeboyd: I misread the TODOs a bit initially.

The first case is related to returning `None` instead of `0` in cases where the span in the reference document does not align with token boundaries in the predicted doc. In this case, we want to ignore this span for `get_loss` rather than treating it as `0` as is currently the case (I think, anyway, please double-check the details!).

For the second case, we can check that the texts are the same in advance, similar to this in coref:

https://github.com/explosion/spacy-experimental/blob/9dd89d5ed2e4abd160f898c4a4ebae1b297ffb61/spacy_experimental/coref/coref_component.py#L232-L239

Many components in spacy do support training for referenced/predicted docs with differing whitespace, but it's a really niche case and since working from character offsets makes this easier, it's okay to have this constraint here as long as it doesn't accidentally use character offsets for cases where this would be incorrect.

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,386 @@
...
+
+@Language.factory(
+    "span_finder",
+    assigns=["doc.spans"],
+    default_config={
+        "threshold": 0.5,
+        "model": DEFAULT_SPAN_FINDER_MODEL,
+        "predicted_key": DEFAULT_PREDICTED_KEY,
+        "training_key": DEFAULT_SPANS_KEY,
+        # XXX Doesn't 0 seem bad compared to None instead?
```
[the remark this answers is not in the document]
adrianeboyd: This comment can be removed now, right?

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,395 @@
...
+        is given a list of documents and predicts a probability for each token.
+    threshold (float): Minimum probability to consider a prediction positive.
+    predicted_key (str): Name of the span group the predicted spans are saved
+        to
+    training_key (str): Name of the span group the training spans are read
+        from
+    max_length (int): Max length of the produced spans (no max limitation when
+        set to 0)
+    min_length (int): Min length of the produced spans (no min limitation when
+        set to 0)
```
adrianeboyd, line 94: Update the types here, too.

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,395 @@
...
+        model (thinc.api.Model): The Thinc Model powering the pipeline component.
+        name (str): The component instance name, used to add entries to the
+            losses during training.
+        threshold (float): Minimum probability to consider a prediction
+            positive.
+        scorer (Optional[Callable]): The scoring method.
+        predicted_key (str): Name of the span group the candidate spans are saved to
+        training_key (str): Name of the span group the training spans are read from
+        max_length (int): Max length of the produced spans (unlimited when set to 0)
+        min_length (int): Min length of the produced spans (unlimited when set to 0)
```
adrianeboyd, line 196: Update the types and the descriptions here, too. I'd add `Defaults to None` for users who are just reading the docstrings.

spacy/errors.py
```diff
@@ -970,6 +970,8 @@ class Errors(metaclass=ErrorsWithCodes):
     E1050 = ("Port {port} is already in use. Please specify an available port with `displacy.serve(doc, port=port)` "
              "or use `auto_select_port=True` to pick an available port automatically.")
     E1051 = ("'allow_overlap' can only be False when max_positive is 1, but found 'max_positive': {max_positive}.")
+    E1052 = ("Both 'min_length' and 'max_length' should be larger than 1, but found"
```
adrianeboyd, line 973:
```suggestion
    E1052 = ("Both 'min_length' and 'max_length' should be larger than 0, but found"
```

spacy/pipeline/span_finder.py
kadarakos: I guess I can do it in reverse that makes more sense right? For each `span` in `eg.predicted` i can try to look for corresponding annotation in the `eg.reference` and if I find something I set it to `1` and if I don't I set it to `None`. For all other tokens its `0`.

kadarakos: I will push this version, but while we are at it I thought it could be helpful for the model to teach it about the "in" label as well even though it is not used for predictions. This would be a 3rd label used to just tag tokens that are inside a span as "in" for just additional training signal.

kadarakos: Then I think this suggestion doesn't really work right?

adrianeboyd: Let me see if I understand everything that's going on here:

- currently the loss function only considers the start and end tokens
- it is a problem that `Doc.char_span` fails if only one is incorrect, since you could still use one of the boundary tokens
  - here maybe you could use `expand` and then filter out cases that aren't the same as the input?
- in `get_loss` should we switch to a loss from thinc that supports missing values?
- you think it might be useful to incorporate the middle tokens so there's more signal for misaligned data but you're not sure this will work?

kadarakos: Currently I pushed is what we discussed on Slack. The label assignment decisions are based on the first and last character offsets of the `predicted` and `reference` spans. Using `alignment_mode="expand"` if the first characters and last characters line up then the labeling is what it used to be.

If they don't both line up then the one that lines up gets supervision and the other one is skipped from the loss. This is achieved by masking the token's class specific logit (multiplying the corresponding entry in `d_scores` by 0).

spacy/tests/pipeline/test_span_finder.py
```diff
@@ -0,0 +1,267 @@
...
+
+def test_span_finder_suggester():
+    nlp = Language()
+    docs = [nlp("This is an example."), nlp("This is the second example.")]
+    docs[0].spans[SPANS_KEY] = [docs[0][3:4]]
+    docs[1].spans[SPANS_KEY] = [docs[1][3:5]]
+    span_finder = nlp.add_pipe("span_finder", config={"spans_key": SPANS_KEY})
+    nlp.initialize()
+    span_finder.set_annotations(docs, span_finder.predict(docs))
+
```
adrianeboyd, line 205:
```suggestion
```

spacy/tests/pipeline/test_span_finder.py
```diff
@@ -0,0 +1,272 @@
...
+            "London",
+            "Berlin",
+            "London and Berlin",
+        }
+
+    # Test scoring
+    scores = nlp.evaluate(train_examples)
+    assert f"span_finder_{span_finder.spans_key}_f" in scores
+    # XXX Its not perfect 1.0 F1 because we want it to overgenerate for now.
+    assert scores[f"span_finder_{span_finder.spans_key}_f"] == 0.4
```
adrianeboyd, line 268: Maybe to make it a little more clear what we're going for with the overfitting?

```suggestion
    assert scores[f"span_finder_{span_finder.spans_key}_f"] == 0.4
    assert scores[f"span_finder_{span_finder.spans_key}_r"] == 1.0
```

spacy/tests/pipeline/test_span_finder.py
```diff
@@ -0,0 +1,261 @@
...
+
+    docs = [nlp("This is an example."), nlp("This is the second example.")]
+    docs[0].spans[SPANS_KEY] = [docs[0][3:4]]
+    docs[1].spans[SPANS_KEY] = [docs[1][3:5]]
+
+    span_finder = nlp.add_pipe("span_finder", config={"spans_key": SPANS_KEY})
+    nlp.initialize()
+    docs = list(span_finder.pipe(docs))
+
+    # TODO: update hard-coded name
```
adrianeboyd, line 138:
```suggestion
```

spacy/tests/pipeline/test_span_finder.py
```diff
@@ -0,0 +1,261 @@
...
+        assert set([span.text for span in spans2]) == {
+            "London",
+            "Berlin",
+            "London and Berlin",
+        }
+
+    # Test scoring
+    scores = nlp.evaluate(train_examples)
+    assert f"span_finder_{span_finder.spans_key}_f" in scores
+    # XXX Its not perfect 1.0 F1 because we want it to overgenerate for now.
```
adrianeboyd, line 255:
```suggestion
    # It's not perfect 1.0 F1 because it's designed to overgenerate for now.
```

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,312 @@
...
+def span_finder_score(examples: Iterable[Example], **kwargs) -> Dict[str, Any]:
+    kwargs = dict(kwargs)
+    attr_prefix = "span_finder_"
+    key = kwargs["spans_key"]
+    kwargs.setdefault("attr", f"{attr_prefix}{key}")
+    kwargs.setdefault(
+        "getter", lambda doc, key: doc.spans.get(key[len(attr_prefix) :], [])
+    )
+    kwargs.setdefault("has_annotation", lambda doc: key in doc.spans)
+    return Scorer.score_spans(examples, **kwargs)
```
adrianeboyd, line 110:
```suggestion
    scores = Scorer.score_spans(examples, **kwargs)
    scores.pop(f"{kwargs['attr']}_per_type", None)
    return scores
```

spacy/ml/models/span_finder.py
```diff
@@ -0,0 +1,43 @@
...
+
+@registry.architectures("spacy.SpanFinder.v1")
+def build_finder_model(
+    tok2vec: Model[InT, List[Floats2d]], scorer: Model[OutT, OutT]
+) -> Model[InT, OutT]:
+
+    logistic_layer: Model[List[Floats2d], List[Floats2d]] = with_array(scorer)
+    model: Model[InT, OutT] = chain(tok2vec, logistic_layer, flattener())
+    model.set_ref("tok2vec", tok2vec)
+    # XXX Why do we have reference for both the scorer and it being wrapped in with_array?
```
adrianeboyd, line 22: What is the status of this comment?

kadarakos: Nothing really. Its unclear to me when we decide to add references or not, but it doesn't really matter actually so I'll just remove it!

spacy/pipeline/__init__.py
```diff
@@ -32,6 +33,7 @@
     "Sentencizer",
     "SpanCategorizer",
     "SpanRuler",
+    "SpanFinder",
```
svlandeg, line 36:
```suggestion
    "SpanFinder",
    "SpanRuler",
```

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,336 @@
...
+
+from thinc.api import Config, Model, Optimizer, set_dropout_rate
+from thinc.types import Floats2d
+
+from spacy.language import Language
+from spacy.pipeline.trainable_pipe import TrainablePipe
+from spacy.scorer import Scorer
+from spacy.tokens import Doc, Span
+from spacy.training import Example
+from spacy.errors import Errors
```
svlandeg, line 11: These should all be relative imports

spacy/pipeline/span_finder.py
```diff
@@ -0,0 +1,336 @@
...
+def _char_indices(span: Span) -> Tuple[int, int]:
+    start = span[0].idx
+    end = span[-1].idx + len(span[-1])
+    return start, end
+
+
+class SpanFinder(TrainablePipe):
+    """Pipeline that learns span boundaries.
+
+    DOCS: https://spacy.io/api/spancategorizer
```
svlandeg, line 131:
```suggestion
    DOCS: https://spacy.io/api/spanfinder
```

svlandeg (commented):
Looks good to me! I only had a few minor comments.

adrianeboyd:
I'll go ahead and merge this so we can include it in a dev release. I'll be testing it a lot with several demo projects, so if anything crops up we can fix it before the v3.6.0 release.

svlandeg:
Thanks @kadarakos and @adrianeboyd for pushing this one over the line! 🙏

PR #12528, opened 2023-04-14
perf(REL_OP): Replace token.children with token.rights or token.lefts

tanloong:
## Description

In some cases, not children at both the right and left side are relevant to the corresponding REL_OP. This PR limits the iteration range only to children on the required side.

### Types of change

Enhancement on the parsing speed of some REL_OPs

## Checklist


adrianeboyd:
Thanks, looks good!

PR #12538, opened 2023-04-17
Update LatinDefaults for lang 'la'

diyclassics:
## Description
Building on the work of issue #11349, this PR makes the following changes:
- Adds noun chunking to syntax iterators (with test)
- Expands list of numeral/ordinal words
- Expands list of abbreviations in tokenizer exceptions
- Adds example sentences

### Types of change
Enhancement to lang defaults

## Checklist

NB: Noun chunking requires use of trained Latin pipeline with POS tagging and dependency parsing, like e.g. this pipeline which I am preparing to submit to the general spaCy trained pipeline offerings: https://huggingface.co/diyclassics/la_core_cltk_md)


spacy/lang/la/syntax_iterators.py
```diff
@@ -0,0 +1,85 @@
+# NB: Modified from da on suggestion from https://github.com/explosion/spaCy/issues/7457#issuecomment-800349751 [PJB]
+
+from typing import Union, Iterator, Tuple
+from ...tokens import Doc, Span
+from ...symbols import NOUN, PROPN, PRON, VERB, AUX
+from ...errors import Errors
+
+
+def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Tuple[int, int, int]]:
```
svlandeg, line 9:
```suggestion
from typing import Union, Iterator, Tuple
from ...tokens import Doc, Span
from ...symbols import NOUN, PROPN, PRON, VERB, AUX
from ...errors import Errors

# NB: Modified from da on suggestion from https://github.com/explosion/spaCy/issues/7457#issuecomment-800349751 [PJB]
def noun_chunks(doclike: Union[Doc, Span]) -> Iterator[Tuple[int, int, int]]:
```

spacy/lang/la/syntax_iterators.py
```diff
@@ -0,0 +1,85 @@
...
+    def is_verb_token(tok):
+        return tok.pos in [VERB, AUX]
+
+    def get_left_bound(doc, root):
```
svlandeg, line 13: It looks like the var 'doc' isn't actually used here (nor in the `da` implementation)

spacy/lang/la/tokenizer_exceptions.py
```diff
@@ -12,65 +12,15 @@
...
-    "Suff.",
-    "Trib.",
-]:
+_abbrev_exc = """A. A.D. Aa. Aaa. Acc. Agr. Ap. Apr. April. A.U.C. Aug. C. Caes. Caess. Cc. Cn. Coll. Cons. Conss. Cos. Coss. D. D.N. Dat. Dd. Dec. Decemb. Decembr. F. Feb. Febr. Februar. Ian. Id. Imp. Impp. Imppp. Iul. Iun. K. Kal. L. M'. M. Mai. Mam. Mar. Mart. Med. N. Nn. Nob. Non. Nov. Novemb. Oct. Octob. Opet. Ord. P. Paul. Pf. Pl. Plur. Post. Pp. Prid. Pro. Procos. Q. Quint. S. S.C. Scr. Sept. Septemb. Ser. Sert. Sex. Sext. St. Sta. Suff. T. Ti. Trib. V. Vol. Vop. Vv.""".split()
+
+_abbrev_exc += [item.lower() for item in _abbrev_exc]
+_abbrev_exc += [item.upper() for item in _abbrev_exc]
+_abbrev_exc += [item.replace("v", "u").replace("V", "U") for item in _abbrev_exc]
+
+_abbrev_exc += ["d.N."]
```
svlandeg, line 21: I think we should filter out the duplicates before we put these into `_exc`. Right now the final list will contain duplicates, like `A.` appearing in the original list and in the `upper` additions, as well as strings that remain the same after doing the v-u replacement.

spacy/lang/la/lex_attrs.py
```diff
@@ -6,17 +6,14 @@
...
+_num_words = """unus una unum duo duae tres tria quattuor quinque sex septem octo novem decem undecim duodecim tredecim quattuordecim quindecim sedecim septendecim duodeviginti undeviginti viginti triginta quadraginta quinquaginta sexaginta septuaginta octoginta nonaginta centum ducenti ducentae ducenta trecenti trecentae trecenta quadringenti quadringentae quadringenta quingenti quingentae quingenta sescenti sescentae sescenta septingenti septingentae septingenta octingenti octingentae octingenta nongenti nongentae nongenta mille
 """.split()
-)
 
-_ordinal_words = set(
-    """
-primus prima primum secundus secunda secundum tertius tertia tertium
-""".split()
-)
+_num_words += [item.replace("v", "u") for item in _num_words]
```
svlandeg, line 12: Here as well, we should consider removing the duplicates (and in `_ordinal_words`)

svlandeg (commented):
Thank you so much for these additions, it all looks really good and extensive! 

I only had a few small comments. Let me know if you'd like to address them or if you'd like us to do so.

spacy/lang/la/lex_attrs.py
diyclassics: Updating to `_num_words` and `ordinal_words` to sets

spacy/lang/la/tokenizer_exceptions.py
```diff
@@ -12,65 +12,15 @@
...
+
+for orth in _abbrev_exc:
```
diyclassics, line 23: Updating this so that this loop is over `set(_abbrev_exc)`

spacy/lang/la/syntax_iterators.py
```diff
@@ -0,0 +1,85 @@
...
+                        doc[root.i : right.i],
+                    )
+                ):
+                    break
+                else:
+                    right_bound = right
+        return right_bound
+
+    def get_bounds(doc, root):
+        return get_left_bound(doc, root), get_right_bound(doc, root)
```
svlandeg, line 37:
```suggestion
        return get_left_bound(root), get_right_bound(doc, root)
```

PR #12545, opened 2023-04-19
Add default option to MorphAnalysis.get

adrianeboyd:
## Description

Similar to `dict`, allow a `default` option for `MorphAnalysis.get` for the user to provide a default return value if the field is not found. The default return value remains `[]`, which is not the same as `dict.get`, but is already established as this method's default return value with the return type `List[str]`. However the new `default` option does not enforce that the user-provided default is actually `List[str]`.

### Types of change

Enhancement.

## Checklist


spacy/tests/doc/test_morphanalysis.py
```diff
@@ -32,7 +32,8 @@ def test_token_morph_key(i_has):
 
 def test_morph_props(i_has):
     assert i_has[0].morph.get("PronType") == ["prs"]
-    assert i_has[1].morph.get("PronType") == []
```
svlandeg, line 35: Why did you remove the original assert, I thought that behaviour would still be valid?

svlandeg (commented):
Looks good! Just had one small question.

spacy/tests/doc/test_morphanalysis.py
adrianeboyd: No reason, I'll add it back.

bryant1410:
Thanks, folks!

PR #12553, opened 2023-04-20
Load exceptions last in Tokenizer.from_bytes

adrianeboyd:
## Description

In `Tokenizer.from_bytes`, the exceptions should be loaded last so that they are only processed once as part of loading the model.

The exceptions are tokenized as phrase matcher patterns in the background and the internal tokenization needs to be synced with all the remaining tokenizer settings. If the exceptions are not loaded last, there are speed regressions for `Tokenizer.from_bytes/disk` vs. `Tokenizer.add_special_case` as the caches are reloaded more than necessary during deserialization.

### Types of change

Bug fix.

## Checklist


svlandeg (approved):
Nice catch!

PR #12582, opened 2023-04-28
TEST: preshed v4

adrianeboyd:
## Description

Test preshed v4.0.0.

### Types of change

n/a

## Checklist


adrianeboyd:
This is probably going to fail. Locally it compiles but doesn't past the first tokenizer init in the test suite.

adrianeboyd:
Ah, I have done this too soon. But whatever is going on, the `preshed` tests don't appear to have covered it.

PR #12602, opened 2023-05-07
Malay language support

khursani8:
## Description
Add Malay language

### Types of change
new feature

## Checklist


adrianeboyd:
Thanks for the PR! An initial question about the tokenizer exceptions list: would it make sense for Indonesian and Malay to use the exact same list of exceptions? This looks like it's just a copy of the Indonesian list, so Malay could import this list from Indonesian rather than copying it if there aren't any differences.

khursani8:
Hi, currently the exceptions list not perfect, some of the Indonesian exceptions can be use for Malay and some might need to be remove in future. I think better separate it so that in future easy to maintain

adrianeboyd:
Thanks for the info! It's fine to maintain a separate list.

Are you planning on updating the exceptions list for Malay for this PR? It's a little unexpected for users to get default exceptions that weren't customized for this language.

adrianeboyd:
Thanks, this looks like a good initial version for the all the defaults!

PR #12607, opened 2023-05-08
In initialize only calculate current vectors hash if needed

adrianeboyd:
## Description

In initialize only calculate current vectors hash if needed. Otherwise this takes unnecessary time and RAM, and fails on extremely large vectors due to msgpack limits.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
A case where this fails is in #12584.

rmitsch (approved):

PR #12620, opened 2023-05-10
universe: Update examples Adept Augementation

davidberenstein1957:
@koaning Suggested changing to a golden data example.

## Description
"Extra open-source tools are always appreciated, but ... just a genuine concern/question. Won't this generate a whole bunch of biased training data? As in, if you have examples that the spaCy model mispredicts ... won't it then generate examples with mispredictions? "

### Types of change
Documentation

## Checklist


bdura (approved):
Looks great!

PR #12623, opened 2023-05-11
Support overriding registered functions in configs

adrianeboyd:
## Description

Support overriding registered functions in configs. Previously the registry name was parsed as a section name rather than as a registry name.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
This still needs to address the CLI commands that call `walk_dict`.

adrianeboyd:
This is getting messy for nested things, let me think about this a bit more...

adrianeboyd:
So the flag is kind of overkill (and can be renamed if anyone has any better ideas), but if anyone happened to be using `walk_dict`, then its default behavior doesn't change.

I think since overrides are applied before the config is resolved, it shouldn't be a problem to treat anything at or below the first registered functions as a value, even if it contains nested registered functions.

PR #12624, opened 2023-05-11
Removed invalid argument

KennethEnevoldsen:
Removed arguments which is no longer valid

## Checklist


bdura:
Hello @KennethEnevoldsen, nice catch!

PR #12625, opened 2023-05-12
Support custom token/lexeme attribute for vectors

adrianeboyd:
## Description

Support custom token/lexeme attribute for vectors.

The main goal is to support `LOWER` or `NORM` for vector lookups instead of only `ORTH`. Because this attribute should be determined by the vectors themselves (primarily how the data was normalized for training) rather than the model config, the attribute for model vector lookup is moved from `StaticVectors` to a property of the `Vectors`.

Because a `Token` can have a custom `NORM` that overrides the `Lexeme` `NORM`, `Token.get_struct_attr` is used to retrieve the attr value in contexts where a `Token` is available.

### Types of change

Enhancement.

## Checklist


honnibal:
All makes sense and looks good 👍

PR #12636, opened 2023-05-16
Remove shorthand for output-file in spacy apply

ljvmiranda921:
The output-file argument is positional, so can't use a shorthand like -o.

## Description

Noticed that there's an extra `-o` option in `spacy apply`'s `output-file` argument. It's unnecessary.

### Types of change

Minor documentation change

## Checklist


adrianeboyd:
Thanks for the PR! Was there anything else you needed to double-check here or is this ready to be merged?

PR #12652, opened 2023-05-22
feat: add comparison operators in `span.pyi`

bdura:
## Description

This PR adds all six comparison operators to the `Span` object. These are defined internally using `__richcmp__`, which is Cython-specific and not recognised by Mypy, for instance.

Hence, this PR also removes type hinting for `__richcmp__`.

### Types of change

This is a typing enhancement.

## Checklist


spacy/tokens/span.pyi
```diff
@@ -51,7 +53,12 @@ class Span:
         kb_id: Union[str, int] = ...,
         span_id: Union[str, int] = ...,
     ) -> None: ...
-    def __richcmp__(self, other: Span, op: int) -> bool: ...
```
adrianeboyd, line 54: Does it make sense to remove this entirely?

bdura:
```python
from spacy.tokens import Span
Span.__richcmp__
```

Fails with an `AttributeError: 'spacy.tokens.span.Span' object has no attribute '__richcmp__'`.

I think Cython mangles the method name in some way while creating the six operators.

PR #12679, opened 2023-05-26
feat: add example stubs

bdura:
## Description

This PR adds a stubs file for `spacy.training.example`. It also fixes a few typing-related issues.

### Types of change

This is an enhancement.

## Checklist


bdura:
I only added evident type hints. I'll add those later on, either in this PR or the next - let me know what works best since I'll likely have little time for that in the next few days.

spacy/tokens/doc.pyi
```diff
@@ -1,16 +1,30 @@
...
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Protocol,
+    Sequence,
+    Tuple,
+    Union,
+    overload,
+)
```
adrianeboyd, line 15:
```suggestion
from typing import Any, Callable, Dict, Iterable, Iterator, List, Optional
from typing import Protocol, Sequence, Tuple, Union, overload
from pathlib import Path
```

spacy/tokens/doc.pyi
```diff
@@ -1,16 +1,30 @@
...
+
+import numpy as np
```
adrianeboyd, line 17:
```suggestion
import numpy
```

spacy/tokens/doc.pyi
adrianeboyd: Let me just do this directly...

spacy/training/corpus.py
```diff
@@ -19,14 +20,19 @@
 FILE_TYPE = ".spacy"
 
 
+class ReaderProtocol(Protocol):
+    def __call__(self, nlp: "Language") -> Iterable[Example]:
```
adrianeboyd, line 24: I thought `Iterator[Example]` was correct throughout here?

svlandeg: Answered [here](https://github.com/explosion/spaCy/pull/12801#pullrequestreview-1518042573).

PR #12689, opened 2023-05-31
Always redo listener state (v2)

svlandeg:
Created an alternative to https://github.com/explosion/spaCy/pull/12499, as a separate branch by Adriane's request.

The reason to create this alternative is that #12499 addresses several different problems, I think this is the set of changes that's minimal to fix the problems we've been seeing with Prodigy. The other issue should ideally be addressed & tested separately in follow-up work.

## Description
Always redo listener state on pipeline modification

* Modify `Language._link_components` to reset listener map and re-add all components from scratch.
* Run `Language._link_components` when pipes are added or removed.
* Remove call to `Language._link_components` from `Language.initialize`

### Types of change
Bug fix.

## Checklist


spacy/language.py
```diff
@@ -951,6 +952,7 @@ def remove_pipe(self, name: str) -> Tuple[str, PipeCallable]:
         # Make sure the name is also removed from the set of disabled components
         if name in self.disabled:
             self._disabled.remove(name)
+        self._link_components()
```
svlandeg, line 955: We should add a test for the `remove_pipe` functionality and double check that it's removed from the internal `listener_map` (which it probably wasn't before this PR)

svlandeg:
Closing in favour of https://github.com/explosion/spaCy/pull/12701

PR #12694, opened 2023-06-02
ci: add cython linter

bdura:
## Description

This PR adds [cython-lint](https://github.com/MarcoGorelli/cython-lint) to the CI.

### Types of change

It's a CI chore.

## Checklist


bdura:
The failing test is expected at this point. I'm stopping the PR here because I'd love some feedback before I go on fixing other Cython files (which will result in a monstrous PR).

Let me know what you think.

shadeMe:
I think it won't be a bad idea to disable the line length and trailing whitespace lints (or for that matter any formatting-related lints) seeing how we don't have a formatter for Cython code?

bdura:
> I think it won't be a bad idea to disable the line length and trailing whitespace lints (or for that matter any formatting-related lints) seeing how we don't have a formatter for Cython code?

I'd be ok with cython-lint forcing us to format Cython files ourselves, I think it would improve legibility which is ultimately a good thing... But that's probably just me. I don't mind fixing all current violations within this PR for instance

.github/workflows/tests.yml
```diff
@@ -41,6 +41,11 @@ jobs:
         run: |
           python -m pip install flake8==5.0.4
           python -m flake8 spacy --count --select=E901,E999,F821,F822,F823,W605 --show-source --statistics
+      - name: cython-lint
+        run: |
+          python -m pip install cython-lint -c requirements.txt
+          cython-lint spacy
```
MarcoGorelli, line 47: hey, thanks for using cython-lint

if the changes are too many, you might want to start by enabling it with `--no-pycodestyle`? that at least is what arrow did

rmitsch: Thanks for the hint!

rmitsch:
`cython-lint` checks are passing. There's a bunch of other tests that are failing now. I don't have a good idea yet why, will look into it (if anybody wants to give the changes a once-over to see if they spot something - highly appreciated).

spacy/tokenizer.pyx
```diff
@@ -458,9 +455,11 @@ cdef class Tokenizer:
         if string:
             if self._try_specials_and_cache(hash_string(string), tokens, has_special, with_special_cases):
                 pass
-            elif (self.token_match and self.token_match(string)) or \
-                    (self.url_match and \
-                    self.url_match(string)):
+            elif (
+                self.token_match and self.token_match(string) or
+                self.url_match and self.url_match(string)
```
adrianeboyd, line 460: Does the linter actually want to remove the parentheses here? (I think it's technically correct but I strongly prefer keeping the parentheses for clarity.)

rmitsch: No, I've taken that liberty. Readded in https://github.com/explosion/spaCy/pull/12694/commits/3583f94dab4ef46fd8322c93746fa88d061707df.

rmitsch:
Seems I've removed a serialization routine call in the first pass. Tests should succeed now (both linting and functional tests).

requirements.txt
```diff
@@ -38,4 +38,5 @@ types-setuptools>=57.0.0
 types-requests
 types-setuptools>=57.0.0
 black==22.3.0
+cython-lint>=0.15.0; python_version >= "3.7"
```
rmitsch, line 41: `cython-lint` is available only for Python 3.7+, and we run the `validate` step with 3.7.

spacy/tokenizer.pyx
svlandeg: Yea, I also strongly prefer keeping the parentheses for clarity in cases like this!

svlandeg (approved):
Nice cleanup, thanks!

I'll follow up with a sync PR to `develop` right after this.

PR #12701, opened 2023-06-06
Address issues with source with component names and replacing listeners

adrianeboyd:
## Description

When sourcing a component, the object from the original pipeline is added to the new pipeline as the same object. This creates a situation where there are several attributes that cannot be in sync between the original pipeline and the new pipeline at the same time for this one object:

* `component.name`
* `component.listener_map` / `component.listening_components` for `tok2vec` and `transformer`

When running `replace_listeners` on a component, the config is not updated correctly if the state of the component is incorrect for the current pipeline (in particular changes that should be applied from `model.attrs["replace_listener_cfg"]` as used in `spacy-transformers`) due to the fact that:

* `find_listeners` relies on `component.name` to set the name in the `listener_map`
* `replace_listeners` relies on `listener_map` to determine how to modify the configs

In addition, there are several places where pipeline components are modified and the listener map and/or internal component names aren't currently updated.

In cases where there is a component shared by two pipelines that cannot be in sync, this PR chooses to prioritize the most recently modified or initialized pipeline. There is no actual solution with the current `source` behavior that will make both pipelines usable, so the current pipeline is updated whenever components are added/renamed/removed or the pipeline is initialized for training.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -934,9 +934,6 @@ class Language:
         if old_name in self._config["initialize"]["components"]:
             init_cfg = self._config["initialize"]["components"].pop(old_name)
             self._config["initialize"]["components"][new_name] = init_cfg
-        pipe = self.get_pipe(new_name)
-        if hasattr(pipe, "name"):
-            pipe.name = new_name
         self._link_components()
 
     def remove_pipe(self, name: str) -> Tuple[str, PipeCallable]:
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1921,27 +1921,6 @@ class Language:
                 raise ValueError(
                     Errors.E942.format(name="pipeline_creation", value=type(nlp))
                 )
-        # Detect components with listeners that are not frozen consistently
-        for name, proc in nlp.pipeline:
-            if isinstance(proc, ty.ListenedToComponent):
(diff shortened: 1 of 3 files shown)


adrianeboyd:
I'll let the tests run in the CI first to show the issues. The new tests have a few temporary modifications (marked with TODOs for later removal) so that the test failures show the interesting failures rather than more minor technical issues.

adrianeboyd:
With the pipeline state kept in sync for the new pipeline by `add_pipe`, the code block at the end of `Language.from_config` to sync listener status should no longer be needed.

It doesn't cause any problems, but it should never apply. In particular, it was it looks like it was unintentional that `replace_listeners` was being checked in two different places and I suspect that at least that part probably should have been removed in #7620.

adrianeboyd:
Tests for `spacy-transformers` using `replace_listener_cfg`: https://github.com/explosion/spacy-transformers/pull/381

* fail with current `master`: https://github.com/explosion/spacy-transformers/actions/runs/5221455649/jobs/9425733452?pr=381#step:13:210
* succeed with this PR: https://buildkite.com/explosion-ai/spacy-transformers-gpu-test-suite/builds/26

PR #12710, opened 2023-06-10
[WIP] Support adding pipeline component instances

honnibal:
Add a method `Language.add_pipe_instance` that allows you to add component instances to the pipeline. This is primarily intended to make it easier to assemble a pipeline dynamically in a script, which is useful for exploration or for situations where you don't care so much about serialisation.

Components that are added this way cannot be reconstructed automatically during deserialisation. Instead, a `pipe_instances` argument is added to `spacy.load()`.

Example usage 1: creating an NLP object with a function, and using `nlp.to_disk()` and `nlp.from_disk()`.

If users wish to assemble the pipeline themselves in a function, they don't need to worry about the config system. They just have to make sure they use the same function to construct the pipeline each time.

```python
def make_nlp():
    nlp = spacy.blank("en")
    tagger = Tagger(nlp.vocab, model=DEFAULT_TAGGER_MODEL, overwrite=False, scorer=tagger_score, neg_prefix="!", label_smoothing=0.0)
    nlp.add_pipe_instance(tagger)
    return nlp

nlp = make_nlp()
nlp.to_disk("/tmp/my_nlp")
# Later
nlp = make_nlp()
nlp.from_disk("/tmp/my_nlp")
```

Example usage 2: Passing component instances to `spacy.load()`

If pipe instances are added directly, we won't be able to reconstruct them in `spacy.load()`. Instead of just making the pipeline impossible to load this way, you can pass the missing instances using the `pipe_instances` argument. This approach is kind of impractical for components that need the `Vocab` instance, but I definitely think it's better than just not supporting it.

```python
vocab = Vocab()
tagger = Tagger(vocab, model=DEFAULT_TAGGER_MODEL, overwrite=False, scorer=tagger_score, neg_prefix="!", label_smoothing=0.0)
nlp = spacy.load("/tmp/my_nlp", vocab=vocab, pipe_instances={"tagger": my_tagger})
```

I added some logic to fetch the `vocab` instance from a pipe instance, so that `nlp = spacy.load("/tmp/my_nlp", pipe_instances={"tagger": my_tagger})` would work. The value of that logic is a bit dubious.

## Questions for discussion

* We could overload `add_pipe`, so that you can just pass the pipe instance in there. I've kept this separate to avoid interaction with existing features, at least until we're happy with it. In the long term I think the overload is a bit more consistent with the general design of the API.
* We could support pipe metadata in `add_pipe_instance`, but then how would we serialise it? I felt it was better to avoid the complication, and if you want those things to be correct you can assemble the pipeline via the config system. I could easily be missing implications of this metadata not being there though.
* Naming of `pipe_instance` argument?
* Fishing the `vocab` instance out of the pipe instances seems dubious. Probably we should just not do this, and ask that the `vocab` argument be set?
* Wording of the warnings about `vocab` instances?
* Should we check for `vocab` instances more robustly? We could go through the `__dict__`, but that still doesn't cover nested objects. I felt that the `vocab` member was a decent value thing to check, albeit not watertight.

## Checklist

### DIFF
--- a/spacy/__init__.py
+++ b/spacy/__init__.py
@@ -35,6 +35,7 @@ def load(
     enable: Union[str, Iterable[str]] = util._DEFAULT_EMPTY_PIPES,
     exclude: Union[str, Iterable[str]] = util._DEFAULT_EMPTY_PIPES,
     config: Union[Dict[str, Any], Config] = util.SimpleFrozenDict(),
+    pipe_instances: Dict[str, Any] = util.SimpleFrozenDict(),
 ) -> Language:
     """Load a spaCy model from an installed package or a local path.
 
@@ -58,6 +59,7 @@ def load(
         enable=enable,
         exclude=exclude,
         config=config,
+        pipe_instances=pipe_instances,
     )
 
 
--- a/spacy/errors.py
+++ b/spacy/errors.py
@@ -215,6 +215,11 @@ class Warnings(metaclass=ErrorsWithCodes):
     W123 = ("Argument `enable` with value {enable} does not contain all values specified in the config option "
             "`enabled` ({enabled}). Be aware that this might affect other components in your pipeline.")
     W124 = ("{host}:{port} is already in use, using the nearest available port {serve_port} as an alternative.")
+    W125 = (
+        "Pipe instance '{name}' is being added with a vocab "
+        "instance that will not match other components. This is "
+        "usually an error."
+    )
 
 
 class Errors(metaclass=ErrorsWithCodes):
@@ -970,6 +975,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E1050 = ("Port {port} is already in use. Please specify an available port with `displacy.serve(doc, port=port)` "
              "or use `auto_select_port=True` to pick an available port automatically.")
     E1051 = ("'allow_overlap' can only be False when max_positive is 1, but found 'max_positive': {max_positive}.")
+    E1052 = ("Cannot create Language instance from config: missing pipeline components. The following components were added by instance (rather than config) via the 'Language.add_pipe_instance()' method, but are not present in the 'pipe_instances' variable: {names}")
 
 
 # Deprecated model shortcuts, only used in errors and warnings
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -801,8 +801,11 @@ class Language:
         self._components.insert(pipe_index, (name, pipe_component))
         return pipe_component
 
-    def add_pipe_instance(self, component: PipeCallable,
-        /, name: Optional[str] = None,
+    def add_pipe_instance(
+        self,
+        component: PipeCallable,
+        /,
+        name: Optional[str] = None,
         *,
         before: Optional[Union[str, int]] = None,
         after: Optional[Union[str, int]] = None,
@@ -1743,7 +1746,7 @@ class Language:
         meta: Dict[str, Any] = SimpleFrozenDict(),
         auto_fill: bool = True,
         validate: bool = True,
-        pipe_instances: Dict[str, Any] = SimpleFrozenDict()
+        pipe_instances: Dict[str, Any] = SimpleFrozenDict(),
     ) -> "Language":
         """Create the nlp object from a loaded config. Will set up the tokenizer
         and language data, add pipeline components etc. If no config is provided,
@@ -1844,7 +1847,7 @@ class Language:
         # and aren't built by factory.
         missing_components = _find_missing_components(pipeline, pipe_instances, exclude)
         if missing_components:
-            raise ValueError(Errors.E1052.format(", ",join(missing_components)))
+            raise ValueError(Errors.E1052.format(", ", join(missing_components)))
         # If components are loaded from a source (existing models), we cache
         # them here so they're only loaded once
         source_nlps = {}
@@ -1858,9 +1861,7 @@ class Language:
                 if pipe_name in exclude:
                     continue
                 else:
-                    nlp.add_pipe_instance(
-                        pipe_instances[pipe_name]
-                    )
+                    nlp.add_pipe_instance(pipe_instances[pipe_name])
             # Is it important that we instantiate pipes that
             # aren't excluded? It seems like we would want
             # the exclude check above. I've left it how it
@@ -2384,7 +2385,9 @@ class _Sender:
             self.send()
 
 
-def _get_instantiated_vocab(vocab: Union[bool, Vocab], pipe_instances: Dict[str, Any]) -> Union[bool, Vocab]:
+def _get_instantiated_vocab(
+    vocab: Union[bool, Vocab], pipe_instances: Dict[str, Any]
+) -> Union[bool, Vocab]:
     vocab_instances = {}
     for name, instance in pipe_instances.items():
         if hasattr(instance, "vocab") and isinstance(instance.vocab, Vocab):
@@ -2410,8 +2413,8 @@ def _get_instantiated_vocab(vocab: Union[bool, Vocab], pipe_instances: Dict[str,
 
 
 def _find_missing_components(
-    pipeline: List[str],
-    pipe_instances: Dict[str, Any],
-    exclude: List[str]
+    pipeline: List[str], pipe_instances: Dict[str, Any], exclude: List[str]
 ) -> List[str]:
-    return [name for name in pipeline if name not in pipe_instances and name not in exclude]
+    return [
+        name for name in pipeline if name not in pipe_instances and name not in exclude
+    ]
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -1,6 +1,6 @@
 from typing import Iterator, Optional, Any, Dict, Callable, Iterable
-from typing import Union, Tuple, List, Set, Pattern, Sequence
-from typing import NoReturn, TYPE_CHECKING, TypeVar, cast, overload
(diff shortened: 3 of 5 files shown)


svlandeg:
> They just have to make sure they use the same function to construct the pipeline each time.

To continue a bit on the usage 1 example, this could become problematic inbetween spaCy versions, no? If we'd change the default tagger model, this `make_nlp` function bypasses the versioning and could break or result in different results?

honnibal:
> The "Example usage 1" to me is crucial to understand how this functionality would/should look like in terms of UX.

I used a tagger for this, which is admittedly a bit of a fiddly use-case for it. Assembling a matcher or something by hand might be more common, or if you're just putting together some arbitrary components.

> Clearly this isn't user-friendly either, because we've now lost a way to just go with the default settings when defining the tok2vec. Also, if we want to better support this approach, we'll need to better document the functions underlying the registry names (right now docs focus on "spacy.HashEmbedCNN.v2" and don't even mention build_hash_embed_cnn_tok2vec)

These are good points to keep in mind.  I would suggest getting the core functionality in first, and then at least theres _some_ way to do this, even if the support for the overall workflow in the docs and the library is shaky.

We don't necessarily need to document every one of these registered functions. What we do need are good ways for users to navigate from registry names to the location in the source code that the name resolves to. The VSCode extension I think does this, right? We need something like that on the command-line as well. 

> To continue a bit on the usage 1 example, this could become problematic in between spaCy versions, no? If we'd change the default tagger model, this make_nlp function bypasses the versioning and could break or result in different results?

Well, let's say people go and reference undocumented stuff with this workflow. Yes, they can have their things break from this between versions. That's a risk that they should take in a well-informed way.

Trying to prevent every mistake or shield users from potential problems can be at the expense of flexibility. If we ask what we expect the Python API to be able to do, from a simple flexibility standpoint, to me it definitely should be able to add components by instance.

svlandeg:
> I used a tagger for this, which is admittedly a bit of a fiddly use-case for it. Assembling a matcher or something by hand might be more common, or if you're just putting together some arbitrary components.

So which use-cases are we really aiming to support here? The matcher doesn't seem applicable either - at least the built-in `Matcher` can't actually be used as a pipeline component, as it returns a list of matches on `__call__`, not the annotated docs like other components do.

If we're thinking about stateless functions like the `evil_component` test case added here, it feels like a small effort to add a one-liner to allow for `nlp.add_pipe("evil_instance")`:

```
@Language.component("evil_instance")
def evil_component(doc):
    ...
```

I can imagine other use-cases where you want to initiate the component class directly from within code though, like for `spacy-llm`'s (v0.3.0) LLM component:
```
from spacy_llm.pipeline import LLMWrapper
from spacy_llm.tasks import NERTask
from spacy_llm.backends.rest.openai import OpenAIBackend
from spacy_llm.cache import BatchCache

cache = BatchCache(path=None, batch_size=342, max_batches_in_mem=42)
backend = OpenAIBackend(config={"model": "text-davinci-003"}, strict=False, max_tries=4, interval=1.0, max_request_time=40)
instance = LLMWrapper(vocab=nlp.vocab, task=NERTask, backend=backend, cache=cache, save_io=False)

nlp.add_pipe_instance(instance, name="my_llm")
```
Like with the `tagger` use-case I tried to work through in my first example, you end up specifying all values of the Python objects you need to use, because we're currently handling all the defaults through the registered functions & the config. 

Does the above code feel like a satisfying solution to offer to users? Or do we need to rethink again how we define these default values? I just want to understand the level of support we want to offer here.

> We don't necessarily need to document every one of these registered functions. What we do need are good ways for users to navigate from registry names to the location in the source code that the name resolves to. The VSCode extension I think does this, right? We need something like that on the command-line as well.

Agreed - this would be good functionality to have!

spacy/util.py
```diff
@@ -675,6 +675,7 @@ def load_model_from_init_py(
     enable: Union[str, Iterable[str]] = _DEFAULT_EMPTY_PIPES,
     exclude: Union[str, Iterable[str]] = _DEFAULT_EMPTY_PIPES,
     config: Union[Dict[str, Any], Config] = SimpleFrozenDict(),
+    pipe_instances: Dict[str, Any] = SimpleFrozenDict(),
```
svlandeg, line 678: Not having this broke `nlp = spacy.load("en_core_web_sm")`,  which wasn't caught by the test suite :|

honnibal:
> Does the above code feel like a satisfying solution to offer to users? Or do we need to rethink again how we define these default values? I just want to understand the level of support we want to offer here.

I think it's better than nothing, yeah.

I'm interested in whether there's a way we can somehow provide the defaults in Python without duplicating them in the configs. Perhaps we could inspect the function to build the default config? Anyway we can leave that for a separate question. I'd still want `add_pipe_instance` even if we don't.

If people build a suite of their own functions and things, they can manage their own defaults, and they should be able to express a reasonable Python API for assembling a pipeline without having to go through our config system.

spacy/__init__.py
```diff
@@ -32,6 +32,7 @@ def load(
     enable: Union[str, Iterable[str]] = util._DEFAULT_EMPTY_PIPES,
     exclude: Union[str, Iterable[str]] = util._DEFAULT_EMPTY_PIPES,
     config: Union[Dict[str, Any], Config] = util.SimpleFrozenDict(),
+    pipe_instances: Dict[str, Any] = util.SimpleFrozenDict(),
```
svlandeg, line 35:
```suggestion
    pipe_instances: Dict[str, PipeCallable] = util.SimpleFrozenDict(),
```
cf the type we're using in `add_pipe_instance()`

spacy/language.py
```diff
@@ -818,12 +824,61 @@ def add_pipe(
...
+        """Add a component instance to the processing pipeline. Valid components
+        are callables that take a `Doc` object, modify it and return it. Only one
+        of before/after/first/last can be set. Default behaviour is "last".
+
+        A limitation of this method is that spaCy will not know how to reconstruct
+        your pipeline after you save it out (unlike the 'Language.add_pipe()' method,
+        where you provide a config and let spaCy construct the instance). See 'spacy.load'
+        for details of how to load back a pipeline with components added by instance.
+
+        pipe_instance (Callable[[Doc], Doc]): The component to add.
```
svlandeg, line 852:
```suggestion
        component (Callable[[Doc], Doc]): The component to add.
```

spacy/language.py
```diff
@@ -818,12 +824,61 @@ def add_pipe(
...
+            used. An error is raised if a name already exists in the pipeline.
+        before (Union[str, int]): Name or index of the component to insert new
+            component directly before.
+        after (Union[str, int]): Name or index of the component to insert new
+            component directly after.
+        first (bool): If True, insert component first in the pipeline.
+        last (bool): If True, insert component last in the pipeline.
+        RETURNS (Callable[[Doc], Doc]): The pipeline component.
+
+        DOCS: https://spacy.io/api/language#add_pipe_instance
```
svlandeg, line 865: TODO: add documentation to API page

spacy/language.py
```diff
@@ -818,12 +824,61 @@ def add_pipe(
...
+        """
+        name = name if name is not None else getattr(component, "name")
+        if name is None:
+            raise ValueError("TODO error")
```
svlandeg, line 869: Alternatively, we could require `name` to be a `str` in this method, and not `Optional`.

spacy/language.py
```diff
@@ -2337,3 +2413,46 @@ def step(self) -> None:
...
+        resolved_vocab = None
+        for name, inst_voc in vocab_instances.items():
+            if resolved_vocab is None:
+                resolved_vocab = inst_voc
+            elif inst_voc is not resolved_vocab:
+                warnings.warn(Warnings.W126.format(name=name))
+        # This is supposed to only be for the type checker --
+        # it should be unreachable
+        assert resolved_vocab is not None
+        return resolved_vocab
```
svlandeg, line 2442: This doesn't look "unreachable"? The `for` loop above doesn't return anything.

spacy/language.py
```diff
@@ -1835,6 +1901,16 @@ def from_config(
...
+                else:
+                    nlp.add_pipe_instance(pipe_instances[pipe_name])
+            # Is it important that we instantiate pipes that
+            # aren't excluded? It seems like we would want
+            # the exclude check above. I've left it how it
+            # is though, in case there's some sort of crazy
+            # load-bearing side-effects someone is relying on?
             pipe_cfg = util.copy_config(pipeline[pipe_name])
             raw_config = Config(filled["components"][pipe_name])
             if pipe_name not in exclude:
```
svlandeg, line 1916: I agree that it looks like we should move the `exclude` check to be the first thing in the `for` loop - we specifically say that excluded components won't be loaded.

Let's move that to a separate PR though?

spacy/__init__.py
svlandeg: Same comment for the other methods, too.

spacy/errors.py
```diff
@@ -219,6 +219,9 @@ class Warnings(metaclass=ErrorsWithCodes):
     W125 = ("The StaticVectors key_attr is no longer used. To set a custom "
             "key attribute for vectors, configure it through Vectors(attr=) or "
             "'spacy init vectors --attr'")
+    W126 = ("Pipe instance '{name}' is being added with a vocab "
+            "instance that will not match other components. This is "
```
svlandeg, line 223:
```suggestion
            "instance that does not match other components. This is "
```

PR #12718, opened 2023-06-12
Fix #12716 does not update the `config` generation section

### LINKED ISSUE #12716: Firefox does not update the `config` generation section.
The spaCy site seems broken on Firefox when it comes to config generation.

When going for “accuracy”, it’s not generating a model that uses `en_core_web_lg`. But if you go to Chrome, it does.

Here's what Firefox seems to look like:

image

essenmitsosse:
## Description

This is a really odd bug, where Firefox doesn't re-render the `code` element, even though `children` changed.

I couldn't find any explanation for why this is happening and why it only happens in Firefox. I assume it is a bug caused by one of our many dependencies (or their interplay). Updating dependencies didn't fix the bug.

To make matters worse: This bug *doesn't* occure when running the site in dev mode. You have to build and serve the site to recreate it.

### Types of change
Two things fixed that:
- remove the `language-ini` `className`
- replace the `code` block with a `div`

Both are not ideal. Therefor this solution adds an inner `div` that now has the classes while still maintaining the semantic `code` element.

## Checklist

### DIFF
--- a/website/src/components/quickstart.js
+++ b/website/src/components/quickstart.js
@@ -215,15 +215,17 @@ const Quickstart = ({
                     }
                 )}
                 <pre className={classes['code']}>
-                    <code
-                        className={classNames(classes['results'], {
-                            [classes['small']]: !!small,
-                            [`language-${codeLang}`]: !!codeLang,
-                        })}
-                        data-quickstart-results=""
-                        ref={contentRef}
-                    >
-                        {Children.toArray(children).flat().filter(isRelevant)}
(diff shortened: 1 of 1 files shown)


essenmitsosse:
Working [Deploy Preview](https://deploy-preview-12718--spacy.netlify.app/usage/training)

svlandeg (approved):
Thanks Marcus! 🙏 

(internally checking first when the best time is to actually merge this and push to the live website)

PR #12733, opened 2023-06-16
Update callbacks.py

PierreNAPOLETANO:
Use ternary

## Checklist


svlandeg:
Hi! Thanks for the contribution. This is really just a style change, which is a matter of preference, and we often use the longer, more readable format throughout the library. So I suggest to keep the code as is. Thanks!

PR #12744, opened 2023-06-21
spacyspellcheck

Vishnunkumar:
# Spacyspellcheck - Simple spell checker using spacy vocab and in-built Levenshtein module 

## Description
A simple spell checker using spacy vocab strings and the in-built Levenshtein module.

### Types of change
New feature - spell check library. 

## Checklist


adrianeboyd:
Thanks for your submission! I can see how you're trying to do simple spell checking using only built-in functionality in spacy, and I think it shows a good understanding of some of the built-in methods and data structures.

I used to do research on spell checking so this PR may end up with a slightly unfair degree of scrutiny for a universe contribution, however I don't think that it makes sense for us to include this package as a suggestion in the universe:

* None of the trained pipelines from https://spacy.io/models such as `en_core_web_sm` provide word lists that are appropriate for spell checking. In particular, the strings in `nlp.vocab.strings` mainly include tokens seen during training (the training data may be noisy and is unlikely to be comprehensive) and the strings also include non-word labels like `conj` or `VBZ` or `WORK_OF_ART` from the pipeline components.
* This module uses whitespace tokenization rather than spacy's tokenization, so adding punctuation to the end of the example results in unexpected suggestions for "approach".
* The suggestions are lowercased, so using "We" in the example returns the suggestion "we".
* The suggestions are provided in a format that's difficult to line up with the input.

Vishnunkumar:
Hey,

Thanks for the update, will work on the same to enhance this.

PR #12783, opened 2023-07-04
Universe docs: Use 'exclude' instead of 'disable'

tomaarsen:
Hello!

## Description
Replace `disable` with `exclude`, as the former will unnecessarily load the unused "ner" model still.

Thank you @svlandeg for the suggestion on LinkedIn.

### Types of change
Documentation.

## Checklist
- [-] I ran the tests, and all new and existing tests passed.

---

- Tom Aarsen

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -4372,7 +4372,7 @@
             "code_example": [
                 "import spacy",
                 "",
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks! 😄

PR #12785, opened 2023-07-04
`Language.replace_listeners`: Pass the replaced listener and the `tok2vec` pipe to the callback

shadeMe:
## Description

Prior to this PR, `Language.replace_listeners` only passed the copied `tok2vec`/`transformer` model to the `replace_listener` callback. This worked for `spacy-transformers` as the library only uses a single `TransformerListener` class, i.e., the replacement model was always the same. However, `spacy-curated-transformers` introduces the concept of different listener classes for downstream components. So, the replacement process needs to provide extra information about the listener model being replaced to correctly select the target model.

This PR adds support for such situations by passing the following extra arguments to the `replace_listener` callback: the replaced listener instance and the `tok2vec` pipe whose model is copied into the downstream component. These argumentss are only passed when the callback accepts them, so the implementation is backwards-compatible with the callback used in `spacy-transformers`.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/extra/DEVELOPER_DOCS/Listeners.md
+++ b/extra/DEVELOPER_DOCS/Listeners.md
@@ -1,14 +1,18 @@
 # Listeners
 
-1. [Overview](#1-overview)
-2. [Initialization](#2-initialization)
-   - [A. Linking listeners to the embedding component](#2a-linking-listeners-to-the-embedding-component)
-   - [B. Shape inference](#2b-shape-inference)
-3. [Internal communication](#3-internal-communication)
-   - [A. During prediction](#3a-during-prediction)
-   - [B. During training](#3b-during-training)
-   - [C. Frozen components](#3c-frozen-components)
-4. [Replacing listener with standalone](#4-replacing-listener-with-standalone)
+- [Listeners](#listeners)
+  - [1. Overview](#1-overview)
+  - [2. Initialization](#2-initialization)
+    - [2A. Linking listeners to the embedding component](#2a-linking-listeners-to-the-embedding-component)
+    - [2B. Shape inference](#2b-shape-inference)
+  - [3. Internal communication](#3-internal-communication)
+    - [3A. During prediction](#3a-during-prediction)
+    - [3B. During training](#3b-during-training)
+      - [Training with multiple listeners](#training-with-multiple-listeners)
+    - [3C. Frozen components](#3c-frozen-components)
+      - [The Tok2Vec or Transformer is frozen](#the-tok2vec-or-transformer-is-frozen)
+      - [The upstream component is frozen](#the-upstream-component-is-frozen)
+  - [4. Replacing listener with standalone](#4-replacing-listener-with-standalone)
 
 ## 1. Overview
 
@@ -62,7 +66,7 @@ of this `find_listener()` method will specifically identify sublayers of a model
 
 If it's a Transformer-based pipeline, a
 [`transformer` component](https://github.com/explosion/spacy-transformers/blob/master/spacy_transformers/pipeline_component.py)
-has a similar implementation but its `find_listener()` function will specifically look for `TransformerListener` 
+has a similar implementation but its `find_listener()` function will specifically look for `TransformerListener`
 sublayers of downstream components.
 
 ### 2B. Shape inference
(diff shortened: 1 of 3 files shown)


spacy/language.py
```diff
@@ -2033,8 +2034,18 @@ def replace_listeners(
...
-                if "replace_listener" in tok2vec_model.attrs:
-                    new_model = tok2vec_model.attrs["replace_listener"](new_model)
+                replace_listener_func = tok2vec_model.attrs.get("replace_listener")
+                if replace_listener_func is not None:
+                    # Pass the extra args to the callback without breaking compatibility with
+                    # old library versions that only expect a single parameter.
+                    num_params = len(
+                        inspect.signature(replace_listener_func).parameters
+                    )
+                    assert num_params in (1, 3)
```
danieldk, line 2044: I think it's nicer to have a proper exception hare. This is an attribute that downstream implementations can define. Or otherwise at least an assert with a message.

extra/DEVELOPER_DOCS/Listeners.md
```diff
@@ -216,5 +220,12 @@ new_model = tok2vec_model.attrs["replace_listener"](new_model)
 ```
 
 The new config and model are then properly stored on the `nlp` object.
-Note that this functionality (running the replacement for a transformer listener) was broken prior to 
+Note that this functionality (running the replacement for a transformer listener) was broken prior to
 `spacy-transformers` 1.0.5.
+
+As of spaCy 3.7, the `replace_listener` callback accepts three arguments instead of just one: the new (copied) model,
```
danieldk, line 226: The text here is a bit ambiguous, since the callback can also accept just one argument. Maybe it's best to write out both cases in the docs here? The listener mechanism is pretty tricky, so I think it's good to be precise in these docs.

spacy/language.py
shadeMe: Good point! Will fix.

extra/DEVELOPER_DOCS/Listeners.md
shadeMe: Clarified that sentence.

spacy/language.py
```diff
@@ -2033,8 +2034,21 @@ def replace_listeners(
...
+                if replace_listener_func is not None:
+                    # Pass the extra args to the callback without breaking compatibility with
+                    # old library versions that only expect a single parameter.
+                    num_params = len(
+                        inspect.signature(replace_listener_func).parameters
+                    )
+                    assert num_params in (
+                        1,
+                        3,
+                    ), f"'replace_listener' callback expects {num_params} parameters, but we only support one or three"
```
svlandeg, line 2047: To come back to Daniël's suggestion, I'd make this an actual exception and raise an error instead of the assert-with-message. If it's user-facing, a proper exception feels more appropriate. We can just put this at the end of the `if-else` structure as the fall-back case in which we raise an exception.

extra/DEVELOPER_DOCS/Listeners.md
```diff
@@ -216,5 +220,18 @@ new_model = tok2vec_model.attrs["replace_listener"](new_model)
 ```
 
 The new config and model are then properly stored on the `nlp` object.
-Note that this functionality (running the replacement for a transformer listener) was broken prior to 
+Note that this functionality (running the replacement for a transformer listener) was broken prior to
 `spacy-transformers` 1.0.5.
+
+In spaCy 3.7, `Language.replace_listeners` was updated to pass the following additional arguments to the `replace_listener` callback:
+the listener to be replaced, the `tok2vec`/`transformer` pipe from which the new model was copied. To maintain backwards-compatiblity,
```
svlandeg, line 227:
```suggestion
the listener to be replaced, and the `tok2vec`/`transformer` pipe from which the new model was copied. To maintain backwards-compatiblity,
```

svlandeg (commented):
Looks good!

There's nowhere in the public docs that we need to update, right?

extra/DEVELOPER_DOCS/Listeners.md
```diff
@@ -1,14 +1,18 @@
...
-1. [Overview](#1-overview)
-2. [Initialization](#2-initialization)
-   - [A. Linking listeners to the embedding component](#2a-linking-listeners-to-the-embedding-component)
-   - [B. Shape inference](#2b-shape-inference)
-3. [Internal communication](#3-internal-communication)
-   - [A. During prediction](#3a-during-prediction)
-   - [B. During training](#3b-during-training)
-   - [C. Frozen components](#3c-frozen-components)
-4. [Replacing listener with standalone](#4-replacing-listener-with-standalone)
+- [Listeners](#listeners)
```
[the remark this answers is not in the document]
shadeMe: Fixed.

spacy/language.py
shadeMe: Changed to an exception.

extra/DEVELOPER_DOCS/Listeners.md
shadeMe: Fixed.

shadeMe:
> There's nowhere in the public docs that we need to update, right?

No, I don't believe so.

PR #12793, opened 2023-07-06
Disallow False for first/last arguments of add_pipe

svlandeg:
Follow up to https://github.com/explosion/spaCy/pull/12349, which is on `master` - this PR targets `v4` and can get rid of the option `False` alltogether

## Description
The `first` and `last` arguments of `add_pipe` were meant to only take `None` or `True`, and disallow `False`. This PR makes that explicit in both the typing, UX and docs.

### Types of change
fix

## Checklist

### DIFF
--- a/spacy/language.py
+++ b/spacy/language.py
@@ -757,8 +757,8 @@ class Language:
         *,
         before: Optional[Union[str, int]] = None,
         after: Optional[Union[str, int]] = None,
-        first: Optional[bool] = None,
-        last: Optional[bool] = None,
+        first: Optional[Literal[True]] = None,
+        last: Optional[Literal[True]] = None,
         source: Optional["Language"] = None,
         config: Dict[str, Any] = SimpleFrozenDict(),
         raw_config: Optional[Config] = None,
@@ -777,8 +777,8 @@ class Language:
             component directly before.
         after (Union[str, int]): Name or index of the component to insert new
             component directly after.
-        first (bool): If True, insert component first in the pipeline.
-        last (bool): If True, insert component last in the pipeline.
+        first (True or None): If True, insert component first in the pipeline.
+        last (True or None): If True, insert component last in the pipeline.
         source (Language): Optional loaded nlp object to copy the pipeline
             component from.
         config (Dict[str, Any]): Config parameters to use for this component.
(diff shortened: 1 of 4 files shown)


spacy/language.py
```diff
@@ -777,8 +777,8 @@ def add_pipe(
             component directly before.
         after (Union[str, int]): Name or index of the component to insert new
             component directly after.
-        first (bool): If True, insert component first in the pipeline.
-        last (bool): If True, insert component last in the pipeline.
+        first (True or None): If True, insert component first in the pipeline.
```
rmitsch, line 780: Wouldn't this be more consistent?
```suggestion
        first (Literal[True]): If True, insert component first in the pipeline.
```

spacy/language.py
```diff
@@ -777,8 +777,8 @@ def add_pipe(
...
+        last (True or None): If True, insert component last in the pipeline.
```
rmitsch, line 781:
```suggestion
        last (Optional[Literal[True]]): If True, insert component last in the pipeline.
```

spacy/language.py
```diff
@@ -823,18 +823,22 @@ def _get_pipe_index(
...
+        last: Optional[Literal[True]] = None,
     ) -> int:
         """Determine where to insert a pipeline component based on the before/
         after/first/last values.
 
         before (str): Name or index of the component to insert directly before.
         after (str): Name or index of component to insert directly after.
-        first (bool): If True, insert component first in the pipeline.
-        last (bool): If True, insert component last in the pipeline.
+        first (True or None): If True, insert component first in the pipeline.
```
rmitsch, line 834:
```suggestion
        first (Optional[Literal[True]]): If True, insert component first in the pipeline.
```

spacy/language.py
```diff
@@ -823,18 +823,22 @@ def _get_pipe_index(
...
+        last (True or None): If True, insert component last in the pipeline.
```
rmitsch, line 835:
```suggestion
        last (Optional[Literal[True]]): If True, insert component last in the pipeline.
```

spacy/language.py
svlandeg: In terms of consistency, you're right that throughout the code base we use `Optional` much more often than `or None` in the doc strings, though there are a handful cases of the latter. Happy to change it, though in that case it should be `Optional[Literal[True]]`.

rmitsch: Fixed the suggestion. Up to you :slightly_smiling_face:

rmitsch (approved):

PR #12795, opened 2023-07-06
Trainable lemmatizer docs link

svlandeg:
## Description
Add an anchor to the docs subsection on "trainable lemmatizer" so we can link to it directly.

Also move the "needs model" bit to the trainable subsection.

### Types of change
docs fix

## Checklist

### DIFF
--- a/website/docs/usage/linguistic-features.mdx
+++ b/website/docs/usage/linguistic-features.mdx
@@ -170,7 +170,7 @@ nlp = spacy.blank("sv")
 nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
 ```
 
-### Rule-based lemmatizer {id="lemmatizer-rule",model="morphologizer,tagger"}
+### Rule-based lemmatizer {id="lemmatizer-rule",model="morphologizer"}
 
 When training pipelines that include a component that assigns part-of-speech
 tags (a morphologizer or a tagger with a [POS mapping](#mappings-exceptions)), a
--- a/website/docs/usage/linguistic-features.mdx
+++ b/website/docs/usage/linguistic-features.mdx
@@ -170,7 +170,7 @@ nlp = spacy.blank("sv")
 nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
 ```
 
-### Rule-based lemmatizer {id="lemmatizer-rule"}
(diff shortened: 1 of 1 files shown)


website/docs/usage/linguistic-features.mdx
```diff
@@ -113,7 +113,7 @@ print(doc[2].morph)  # 'Case=Nom|Person=2|PronType=Prs'
 print(doc[2].pos_)  # 'PRON'
 ```
 
-## Lemmatization {id="lemmatization",model="lemmatizer",version="3"}
+## Lemmatization {id="lemmatization",version="3"}
```
adrianeboyd, line 116: If you're going to remove `model="lemmatizer"` here, then I think it makes sense to also add it to the rule-based lemmatizer section?

adrianeboyd: The rule-based lemmatizers all require at least POS from a model?

svlandeg: Ok, so for the rule-based section you want to add `model="tagger"`?

adrianeboyd: Hmm, there's not a simple description here that applies across the board. I guess `morphologizer` is accurate enough for most cases is the closest.

website/docs/usage/linguistic-features.mdx
```diff
@@ -194,7 +194,7 @@ information, without consulting the context of the token. The rule-based
 lemmatizer also accepts list-based exception files. For English, these are
 acquired from [WordNet](https://wordnet.princeton.edu/).
 
-### Trainable lemmatizer
+### Trainable lemmatizer {id="lemmatizer-train",model="lemmatizer"}
```
adrianeboyd, line 197: I'm trying to think about what exactly the tags should do. Should this be `trainable_lemmatizer` then? (I might just punt and go back to `lemmatizer` on the section? I'm not sure how much this is confusing users.)

svlandeg: Having it on the "lemmatizer" section is wrong, IMO. These tags define which trainable component should be in the pipeline, and the 3 subsections are quite different in that regard.

svlandeg: So yes, let's make this `trainable_lemmatizer`.

PR #12797, opened 2023-07-06
black formatting

svlandeg:
## Description
There was a strange github issue with syncing on PR https://github.com/explosion/spaCy/pull/12679, it was green and I could merge it, then only after the merge it showed the test failure. Pretty strange. Anyway, adding the `black` formatting.

### Types of change
format fix

## Checklist


adrianeboyd:
I noticed with GHA there's often a delay before it updates to the most recent commit, so sometimes it's still showing the green check from the previous run in the wrong spot.

But can we revert https://github.com/explosion/spaCy/pull/12679 instead and move it to `develop`?

svlandeg:
> I noticed with GHA there's often a delay before it updates to the most recent commit, so sometimes it's still showing the green check from the previous run in the wrong spot.

Ah interesting, I hadn't seen that before. But this time I literally saw it updating the page as I was merging, even though it was showing the last commit be green before.

> 
> But can we revert #12679 instead and move it to `develop`?

We can definitely revert if you prefer, but why target `develop`? That PR shouldn't be breaking?

svlandeg:
Reverted, closing in favour of https://github.com/explosion/spaCy/pull/12798 (`master` vs `develop` still needs to be discussed though)

adrianeboyd:
I thought we had already configured `develop` for v3.7?

I thought we had agreed not modify types on patch releases, except for possibly minor cases where we needed to widen a type?

svlandeg:
> I thought we had agreed not modify types on patch releases, except for possibly minor cases where we needed to widen a type?

Right - that's what I'm trying to understand, which change you see as modifying or breaking in that original PR.

adrianeboyd:
I think adding types where there weren't types before is narrowing and has the potential to possibly trigger new type checking errors for users?

Let me also comment on the new PR...

PR #12817, opened 2023-07-11
🐛 Escape annotated HTML tags in span renderer

### LINKED ISSUE #12816: 🐛 Annotated HTML tags are not escaped in displaCy span renderer output

Here's a test that currently fails:
```python
def test_span_escaping(en_vocab) -> None:
    """Test that displaCy's span visualizer escapes annotated HTML tags correctly."""
    # Create a doc containing an annotated word and an unannotated HTML tag
    doc = Doc(en_vocab, words=["test", "<TEST>"])
    doc.spans["sc"] = [Span(doc, 0, 1, label="test")]

    # Verify that the HTML tag is escaped when unannotated
    html = displacy.render(doc, style="span")
    assert "&lt;TEST&gt;" in html

    # Annotate the HTML tag
    doc.spans["sc"].append(Span(doc, 1, 2, label="test"))

    # Verify that the HTML tag is still escaped
    html = displacy.render(doc, style="span")
    assert "&lt;TEST&gt;" in html
```

The test currently fails on the last line, since the annotated HTML tag is not escaped by the displaCy renderer. Adding a call to `escape_html` here fixes the issue:
https://github.com/explosion/spaCy/blob/ddffd096024004f27a0dee3701dc248c4647b3a7/spacy/displacy/render.py#L221

I ran into this issue when trying to visualize some annotated code documents with `<span>` in some of the documents. This resulted in the documents and span underlines rendering on top of each other at the beginning of the visualization. Adding `escape_html` as described above fixed the rendering issues.

~I have a PR ready to fix this that I'll post as soon as I update the test <-> issue links in the code!~ Posted! 😄 

## Your Environment

- **spaCy version:** 3.6.0
- **Platform:** macOS-13.4.1-arm64-arm-64bit
- **Python version:** 3.11.2

connorbrinton:
## Description

These changes add a missing call to `escape_html` in the displaCy span renderer. Previously span-annotated tokens would be inserted into the page markup without being escaped, resulting in potentially incorrect rendering. When I encountered this issue, it resulted in some docs and span underlines being superimposed on top of properly rendered docs and span underlines near the beginning of the visualization (due to an unescaped `<span>` tag).

Fixes #12816.

### Types of change

Bug fix

## Checklist

- [x] I ran the tests, and all new and existing tests passed (except for a few errors related to a missing `Vectors.attr` field, but they don't seem related to my changes)

### DIFF
--- a/spacy/displacy/render.py
+++ b/spacy/displacy/render.py
@@ -1,4 +1,3 @@
-import itertools
 import uuid
 from typing import Any, Dict, List, Optional, Tuple, Union
 
@@ -218,7 +217,7 @@ class SpanRenderer:
                     + (self.offset_step * (len(entities) - 1))
                 )
                 markup += self.span_template.format(
-                    text=token["text"],
+                    text=escape_html(token["text"]),
                     span_slices=slices,
                     span_starts=starts,
                     total_height=total_height,
--- a/spacy/tests/test_displacy.py
+++ b/spacy/tests/test_displacy.py
@@ -377,3 +377,22 @@ def test_displacy_manual_sorted_entities():
 
     html = displacy.render(doc, style="ent", manual=True)
     assert html.find("FIRST") < html.find("SECOND")
+
+
+@pytest.mark.issue(12816)
+def test_issue12816(en_vocab) -> None:
+    """Test that displaCy's span visualizer escapes annotated HTML tags correctly."""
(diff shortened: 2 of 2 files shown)


svlandeg (approved):
Thanks so much for the report and the PR! We appreciate it 🙏

Issue #12731, opened 2023-06-16
i am not able to download en modules like sm, md.


ipratham101:
please help i am trying this from 2 days tried around 50 methods none worked
now am kinda lost the motivation to learn about spacy, this is going to be my last hope

## Which page or section is this issue related to?

error->

C:\Users\Divyansh Ahuja>python -m spacy download en

Traceback (most recent call last):
  File "<frozen runpy>", line 189, in _run_module_as_main
  File "<frozen runpy>", line 148, in _get_module_details
  File "<frozen runpy>", line 112, in _get_module_details
  File "C:\Users\Divyansh Ahuja\AppData\Local\Programs\Python\Python311\Lib\site-packages\spacy\__init__.py", line 14, in <module>
    from . import pipeline  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Divyansh Ahuja\AppData\Local\Programs\Python\Python311\Lib\site-packages\spacy\pipeline\__init__.py", line 1, in <module>
    from .attributeruler import AttributeRuler
  File "C:\Users\Divyansh Ahuja\AppData\Local\Programs\Python\Python311\Lib\site-packages\spacy\pipeline\attributeruler.py", line 6, in <module>
    from .pipe import Pipe
  File "spacy\pipeline\pipe.pyx", line 1, in init spacy.pipeline.pipe
  File "spacy\vocab.pyx", line 1, in init spacy.vocab
  File "C:\Users\Divyansh Ahuja\AppData\Local\Programs\Python\Python311\Lib\site-packages\spacy\tokens\__init__.py", line 1, in <module>
    from .doc import Doc
  File "spacy\tokens\doc.pyx", line 36, in init spacy.tokens.doc
  File "C:\Users\Divyansh Ahuja\AppData\Local\Programs\Python\Python311\Lib\site-packages\spacy\schemas.py", line 158, in <module>
    class TokenPatternString(BaseModel):
  File "C:\Users\Divyansh Ahuja\AppData\Local\Programs\Python\Python311\Lib\site-packages\pydantic\main.py", line 369, in __new__
    cls.__signature__ = ClassAttribute('__signature__', generate_model_signature(cls.__init__, fields, config))
                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #12819, opened 2023-07-12
fix: Resolve Python keyword conflict in Pydantic models

teleprint-me:
- Renamed 'IN' attribute in TokenPatternString and TokenPatternNumber models to 'IS_IN' to avoid conflict with Python's 'in' keyword.
- Updated the corresponding Field alias to 'is_in' from 'in'.
- This fix prevents a TypeError upon the initialization of these models due to the improper use of a reserved Python keyword as an attribute.

## Description
This PR addresses an issue in which the use of the 'in' keyword as an attribute in the Pydantic models `TokenPatternString` and `TokenPatternNumber` resulted in a TypeError. To resolve this, 'IN' has been renamed to 'IS_IN', and the associated Field alias has been updated from 'in' to 'is_in'. As a result, these models can now be initialized without running into a conflict with Python's reserved keywords.

The changes were not tested due to issues with relative imports in the development environment. 

Note that these changes may affect the loading of existing trained models or saved data that rely on the 'in' field. As such, proper migration strategies should be considered.

### Types of change
This PR covers a bug fix.

## Checklist

This is a suggested fix for issue #12731.

### DIFF
--- a/spacy/schemas.py
+++ b/spacy/schemas.py
@@ -179,7 +179,7 @@ def validate_token_pattern(obj: list) -> List[str]:
 
 class TokenPatternString(BaseModel):
     REGEX: Optional[Union[StrictStr, "TokenPatternString"]] = Field(None, alias="regex")
-    IS_IN: Optional[List[StrictStr]] = Field(None, alias="is_in")
+    IN: Optional[List[StrictStr]] = Field(None, alias="is_in")
     NOT_IN: Optional[List[StrictStr]] = Field(None, alias="not_in")
     IS_SUBSET: Optional[List[StrictStr]] = Field(None, alias="is_subset")
     IS_SUPERSET: Optional[List[StrictStr]] = Field(None, alias="is_superset")
@@ -226,7 +226,7 @@ class TokenPatternString(BaseModel):
 
 class TokenPatternNumber(BaseModel):
     REGEX: Optional[StrictStr] = Field(None, alias="regex")
-    IS_IN: Optional[List[StrictInt]] = Field(None, alias="is_in")
+    IN: Optional[List[StrictInt]] = Field(None, alias="is_in")
     NOT_IN: Optional[List[StrictInt]] = Field(None, alias="not_in")
     IS_SUBSET: Optional[List[StrictInt]] = Field(None, alias="is_subset")
     IS_SUPERSET: Optional[List[StrictInt]] = Field(None, alias="is_superset")
--- a/spacy/schemas.py
+++ b/spacy/schemas.py
@@ -179,7 +179,7 @@ def validate_token_pattern(obj: list) -> List[str]:
 
 class TokenPatternString(BaseModel):
     REGEX: Optional[Union[StrictStr, "TokenPatternString"]] = Field(None, alias="regex")
-    IN: Optional[List[StrictStr]] = Field(None, alias="in")
+    IS_IN: Optional[List[StrictStr]] = Field(None, alias="is_in")
     NOT_IN: Optional[List[StrictStr]] = Field(None, alias="not_in")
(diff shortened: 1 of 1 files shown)


teleprint-me:
@adrianeboyd 

I appreciate the response and information. 

Did you review [the version information](https://github.com/explosion/spaCy/issues/12731#issuecomment-1634841932)?

I will review, test, and verify when I get a chance to.

I use `poetry`, so handling versions is much simpler as a result.

The issue isn't resolved though. I've already validated it and reviewed the source and ran the tests. The `alias` parameter cannot be a Python keyword. This is a fact, not an opinion.

adrianeboyd:
Yes, I saw the version information, which is why my best guess is that a package install might be corrupted in your current environment, since the error `ValueError: 'in' is not a valid parameter name` should be fixed with pydantic v1.10+.

I can see how using a keyword as an alias causes limitations for the pydantic model in general, but I haven't been able to find an example where it causes problems for this usage for token pattern validation in spacy. I double-checked and extended the test cases to confirm: #12835

If you're running into a different error (you mention a `TypeError` above?) it might be something different than in the original issue. If that's the case, please open a new issue with the exact steps you're using to install spacy in a new venv, the output of `spacy info` and `pip freeze`, and the full error traceback.

adrianeboyd:
I'm sure that this problem is fixed in pydantic v1.10.x, so if you're still having trouble, this really points to an installation or environment problem. `pip` can definitely be frustrating here.

I can recommend:

In an existing venv, try to uninstall packages multiple times before installing a different version, especially if you've been using any editable installs:

```shell
python -m pip uninstall -y pydantic
python -m pip uninstall -y pydantic
python -m pip install 'pydantic~=1.10.0'
```

However, it's easiest to get things working in a brand new venv, without upgrading or downgrading any packages. Install all packages with one single install command like `pip install -r requirements.txt` command or whatever you use.

Skip your pip cache while installing in case a cached download is corrupted:

```shell
python -m pip install 'pydantic~=1.10.0' --no-cache-dir
```

If skipping the cache solves the problem, then purge your pip cache to delete any corrupted packages:

```shell
python -m pip cache purge
```

teleprint-me:
Yeah, I've been polite, insightful, and have even given valid feedback.

It just feels like you're ignoring what's happening at this point which makes my efforts feel moot and makes me feel like no longer relying on a package like this.

Walking me through a .venv step-by-step when I've clearly showcased competency in navigating a complex code base is just mind-numbing and pointless to me.

I've reported the details, the output, the logs, the tests, and even included a PR.

Why didn't I think of that? Like, just delete my '.venv' and restart from scratch. It did occur to me, I did do that, and I've clearly outlined how and why.

This isn't tricky, it's logic that a computer follows. There's a version constraint and you have conflicting outstanding issues that need to be resolved.

Come up with a plan on how to tackle it. It requires time and planning via review and testing.

It would be less disingenuous if it didn't feel like you were attempting to kill me with kindness here.

teleprint-me:
[pydantic/HISTORY.md#v1108-2023-05-23](https://github.com/pydantic/pydantic/blob/4fbfe53e78b99c651567bf7a16c8b1e2b9720e91/HISTORY.md#v1108-2023-05-23)

The version required here is `v1.10.8`. So, if a project, say like [chroma](https://github.com/chroma-core/chroma/blob/main/requirements.txt), requires anything lower, and I specify that in my `pyproject.toml` and run `poetry install`, then it will generate an error stating a package incompatibility.

Simply upgrading isn't necessarily a valid solution as a result.

adrianeboyd:
While your frustration related to managing python dependencies is understandable, the rest of this interaction is much less so. We have provided the version information you should need to sort this out for your environment and your project-specific requirements. Sorry, I'm afraid I don't see a productive way forward for this discussion.

PR #12820, opened 2023-07-13
Add support for `llm` in `spacy init config`

rmitsch:
## Description

Add support for `llm` component. in `spacy init config`

### Types of change

Enhancement.

## Checklist

### DIFF
--- a/spacy/cli/init_config.py
+++ b/spacy/cli/init_config.py
@@ -1,7 +1,7 @@
 import re
 from enum import Enum
 from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Any, Dict
 
 import srsly
 from jinja2 import Template
@@ -9,9 +9,10 @@ from thinc.api import Config
 from wasabi import Printer, diff_strings
 
 from .. import util
+from ..errors import Errors
 from ..language import DEFAULT_CONFIG_PRETRAIN_PATH
 from ..schemas import RecommendationSchema
-from ..util import SimpleFrozenList
+from ..util import SimpleFrozenList, registry
 from ._util import (
     COMMAND,
     Arg,
@@ -40,6 +41,8 @@ class InitValues:
 
     lang = "en"
     pipeline = SimpleFrozenList(["tagger", "parser", "ner"])
+    llm_task: Optional[str] = None
+    llm_model: Optional[str] = None
     optimize = Optimizations.efficiency
     gpu = False
     pretraining = False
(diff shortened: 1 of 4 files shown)


rmitsch:
As of now `spacy-llm` is required if `llm` is added in the pipeline. This is because we need its registry entries to resolve `--llm.task ner` to e. g. `spacy.NER.v2` (note that it the current state we don't select the highest version, this is a todo). This means we'd have to install `spacy-llm` for the CI tests though. Are we ok with that?

Alternatively we could also required the registry handle to be set explicitly: e. g. `--llm.task spacy.NER.v2`. In this case there's no need to require `spacy-llm`, but it's somewhat less consistent with the way `--pipeline` is configured.

Opinions?

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -3,7 +3,7 @@ the docs and the init config command. It encodes various best practices and
 can help generate the best possible configuration, given a user's requirements. #}
 {%- set use_transformer = hardware != "cpu" and transformer_data -%}
 {%- set transformer = transformer_data[optimize] if use_transformer else {} -%}
-{%- set listener_components = ["tagger", "morphologizer", "parser", "ner", "textcat", "textcat_multilabel", "entity_linker", "span_finder", "spancat", "spancat_singlelabel", "trainable_lemmatizer"] -%}
+{%- set listener_components = ["tagger", "morphologizer", "parser", "ner", "textcat", "textcat_multilabel", "entity_linker", "span_finder", "spancat", "spancat_singlelabel", "trainable_lemmatizer", "llm"] -%}
```
adrianeboyd, line 6: `llm` shouldn't be in this list of components with listeners, which will cause `tok2vec` or `transformers` to be added (depending on the rest of the template, of course).

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -585,6 +597,17 @@ no_output_layer = false
...
+[components.llm]
+factory = "llm"
+
+[components.llm.model]
+@llm_models = "{{ llm_spec['model'] }}"
+
+[components.llm.task]
+@llm_tasks = "{{ llm_spec['task'] }}"
+{% endif -%}
+
```
adrianeboyd, line 610: I think (depending on exactly how this gets designed in the end), it should be possible to only have this block once because it does not depend on the listener setup.

spacy/tests/test_cli.py
```diff
@@ -651,6 +650,26 @@ def test_init_config(lang, pipeline, optimize, pretraining):
     load_model_from_config(config, auto_fill=True)
 
 
+@pytest.mark.parametrize("pipeline", [["llm"]])
+@pytest.mark.parametrize("llm_model", ["noop"])
+@pytest.mark.parametrize("llm_task", ["ner", "sentiment"])
+def test_init_config_llm(pipeline, llm_model, llm_task):
```
adrianeboyd, line 656: I think this test should probably be in `spacy-llm` unless it's installed by default (similar to `spacy-transformers`, where I'm not sure this is tested at all, but where it could be tested).

svlandeg: IMO we should consider installing `spacy-llm` by default going forward. At some point we want to move it to the core codebase, but having it installed by default + kept as separate repo has the advantages that we can still release updates more quickly with `spacy-llm` while not troubling users with an additional install command. The [requirements](https://github.com/explosion/spacy-llm/blob/main/requirements.txt) are minimal.

svlandeg (commented):
This is currently not working for most OS HF models, as they don't have a default `name` like e.g. `""spacy.GPT-3-5.v1"` does.

For instance, `spacy init config myconfig.cfg -p "llm" --llm.model "Falcon" --llm.task 'ner'`

will fail with 

```
✘ Config validation error
llm.model -> name       field required
{'@llm_models': 'spacy.Falcon.v1'}
```

spacy/cli/init_config.py
```diff
@@ -165,8 +174,57 @@ def init_config(
...
+
+        for spec_type, spec in llm_spec.items():
+            valid_values = set()
+            user_value = spec["arg"].lower().replace(".", "-")
+            for reg_handle in spec["reg_handles"]:
+                reg_name = reg_handle.split(".")[1]
+                valid_values.add(reg_name)
+                if reg_name.lower() == user_value:
+                    spec["matched_reg_handle"] = reg_handle
+                    break
```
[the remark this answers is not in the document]
svlandeg: It doesn't - I just checked - it produces `spacy.NER.v1` instead of `spacy.NER.v2` when given `"ner"`

spacy/cli/templates/quickstart_training.jinja
rmitsch: Makes sense. I vaguely remember that the LLM block was not output properly if "llm" was not in `listener_components`. I'll check.

spacy/cli/init_config.py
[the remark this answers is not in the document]
rmitsch: No, it just grabs the first occurence as of now. I left that as TBD since it wasn't clear how we want to go forward.

PR #12828, opened 2023-07-15
added entry for SaysWho

afriedman412:
## Checklist


adrianeboyd (commented):
Thanks for this contribution! I was able to install and run the examples following the instructions in the README without any problems.

I'd recommend specifying that this package (currently) only supports English.

I'd also recommend using the most recent coreference model:

https://github.com/explosion/spacy-experimental/releases/download/v0.6.1/en_coreference_web_trf-3.4.0a2-py3-none-any.whl

Error messages that explain how to install missing models might be helpful, especially for the experimental coreference model, since that might not be obvious to many users.

And in the universe entry it would be nice to have a short code example and a bit more detail about how this package relates to spacy.

website/meta/universe.json
```diff
@@ -16,6 +16,22 @@
             },
             "category": ["extension"],
             "tags": []
+        },
+                {
+            "id": "sayswho",
+            "title": "SaysWho",
+            "slogan": "Quote identification, attribution and resolution.",
+            "description": "A Python package for identifying and attributing quotes in text. It uses a combination of logic and grammer to find quotes and their speakers, then uses a coreferencing model to better clarify who is speaking.",
```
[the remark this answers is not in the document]
afriedman412: wow this is embarassing lol

afriedman412:
is there a way to install the coreferencing model without automatically installing SpaCy 3.4?

adrianeboyd:
You can install the model with `--no-deps`, but then it doesn't pull in `spacy-experimental`. It basically still is all on the experimental side and maybe not quite ready to be used in another package like this, although it's still really nice to see what people are doing with the coref model!

I think that this particular model should currently be compatible with any newer versions of spacy, but I haven't tested it and you'll see warnings.

Is there a particular reason you need to use a newer version of spacy?

afriedman412:
ok noted -- yeah it works fine with the newer versions as far as I can tell so you might as well use the latest version if possible, you know?

website/meta/universe.json
```diff
@@ -16,6 +16,31 @@
             },
             "category": ["extension"],
             "tags": []
+        },
+                {
+            "id": "sayswho",
+            "title": "SaysWho",
+            "slogan": "Quote identification, attribution and resolution",
+            "description": "A Python package for identifying and attributing quotes in text. It uses a combination of SpaCy functionality, logic and grammar to find quotes and their speakers, then uses the SpaCy coreferencing model to better clarify who is speaking. Currently English only.",
```
adrianeboyd, line 24:
```suggestion
       {
            "id": "sayswho",
            "title": "SaysWho",
            "slogan": "Quote identification, attribution and resolution",
            "description": "A Python package for identifying and attributing quotes in text. It uses a combination of spaCy functionality, logic and grammar to find quotes and their speakers, then uses the spaCy coreferencing model to better clarify who is speaking. Currently English only.",
```

adrianeboyd:
Thanks for the updates!

website/meta/universe.json
```diff
@@ -17,6 +17,31 @@
             "category": ["extension"],
             "tags": []
         },
+       {
```
adrianeboyd, line 20:
```suggestion
        {
```

PR #12829, opened 2023-07-15
Add Left and Right Pointing Angle Brackets as punctuation to ancient Greek

jmyerston:
## Description

This PR adds 〈 (U+2329) and 〉(U+232A) to the ancient Greek. These two symbols are very common in ancient Greek texts to mark editorial reconstructions. As right now, spaCy does not separate these symbol from words what causes improper tokenization.

### Types of change

Adds the two symbols as prefixes and suffixes in the punctuation file.  

## Checklist

- [ X] I confirm that I have the right to submit this contribution under the project's MIT license.
- [X ] I ran the tests, and all new and existing tests passed.
- [ X] My changes don't require a change to the documentation, or if they do, I've added all required information.


adrianeboyd:
Ah, I'll switch this back until `develop` is back up-to-date so the history isn't incorrect. But we want this to target `develop` so this change is first in v3.7.0 and doesn't affect existing v3.6.x pipelines.

adrianeboyd:
Thanks for the PR!

PR #12839, opened 2023-07-18
Update numpy build constraints for numpy 1.25

adrianeboyd:
## Description

Starting in numpy 1.25 (see
https://github.com/numpy/numpy/releases/tag/v1.25.0, https://numpy.org/doc/1.25/dev/depending_on_numpy.html), the numpy C API is backwards-compatible by default.

For python 3.9+, we should be able to drop the specific numpy build requirements and use `numpy>=1.25`, which is currently backwards-compatible to `numpy>=1.19`.

In the future when python 3.8 is dropped, the specific python <3.9 requirements could be dropped and the lower numpy pin could correspond to the oldest supported version for the current lower python pin.

### Types of change

?

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,8 @@ project_urls =
 zip_safe = false
 include_package_data = true
 python_requires = >=3.6
-# NOTE: This section is superseded by pyproject.toml will be removed in spaCy v4
+# NOTE: This section is superseded by pyproject.toml and will be removed in
+# spaCy v4
 setup_requires =
     cython>=0.25,<3.0
     # The newest supported pip for python 3.6 has bugs related to markers in


adrianeboyd:
The CI does not currently really test this, since numpy isn't downgraded after building.

From what I understand the lower numpy build constraint will need to be `numpy>=2.0` once numpy 2 is released.

spacy/tests/package/test_requirements.py
```diff
@@ -22,6 +22,7 @@ def test_build_dependencies():
     ]
     # ignore language-specific packages that shouldn't be installed by all
     libs_ignore_setup = [
+        "numpy",
```
rmitsch, line 25: Why is `numpy` in `libs_ignore_requirements` and `libs_ignore_setup` now?

setup.cfg
```diff
@@ -32,8 +32,12 @@ project_urls =
 zip_safe = false
 include_package_data = true
 python_requires = >=3.6
+# NOTE: This section is superseded by pyproject.toml will be removed in spaCy v4
```
rmitsch, line 35: ?
```suggestion
# NOTE: This section is superseded by pyproject.toml and will be removed in spaCy v4
```

spacy/tests/package/test_requirements.py
adrianeboyd: The test can't deal with the duplicate entries.

adrianeboyd: Probably also not the diverging versions in build vs. install requirements, which we didn't have before.

rmitsch (approved):

PR #12853, opened 2023-07-24
Switch from distutils to setuptools/sysconfig

adrianeboyd:
## Description

Additionally remove outdated `is_new_osx` check and settings.

### Types of change

?

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,9 @@
 #!/usr/bin/env python
 from setuptools import Extension, setup, find_packages
 import sys
-import platform
 import numpy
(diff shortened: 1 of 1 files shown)


svlandeg (approved):
Thanks for taking care of all of these!

adrianeboyd:
For the `is_new_osx` thing, there is a slight possibility that something might complain that it breaks something, but we don't have any good way to test it at this point. We could consider adding it back in (also for a new other packages, but with the `platform.mac_ver()` check from `spacy` instead of `distutils.util.get_platform()` that's in a few other places).

PR #12860, opened 2023-07-26
`SpanCat`: Remove invalid `threshold` config argument from example

shadeMe:
## Description

As discussed in #12858.

The `threshold` argument is only valid for the multi-class SpanCat component.

### Types of change

Doc fix

## Checklist


svlandeg (approved):
Thanks!

PR #12882, opened 2023-08-02
Update br tags

adrianeboyd:
## Description

I think the intention in displacy was `<br/>`, but maybe now `<br>` would be better (says the w3c markup validator anyway)?

### Types of change

Bug fix?

## Checklist

### DIFF
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -403,7 +403,7 @@ def _format_sources(data: Any) -> str:
         if author:
             result += " ({})".format(author)
         sources.append(result)
-    return "<br />".join(sources)
+    return "<br>".join(sources)
 
 
 def _format_accuracy(data: Dict[str, Any], exclude: List[str] = ["speed"]) -> str:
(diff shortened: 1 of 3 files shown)


rmitsch:
Does this fix a bug? I just wouldn't want to assume the closing tag is an opening tag. `<br>` should be sufficient now AFAIK, yes (not an expert though).

adrianeboyd:
I think all browsers probably render `<br>`, `</br>`, and `<br/>` the same so it's not exactly a bug? It's been like this for ages. I'd noticed a few times in the past when inspecting the rendering but never worried about it beyond that. But I don't think that `</br>` has ever been an actual valid tag?

rmitsch:
Ah right, it's just line break.

rmitsch (approved):

svlandeg (approved):
If the w3c markup validator suggests `<br>` I would yet change it into that, but I suspect that it won't make a practical difference either way.

PR #12945, opened 2023-08-31
spacy-llm docs cleanup

svlandeg:
## Description
Significant clean-up of the `spacy-llm` docs. It'll probably be easier to review the summary below + [the preview](https://deploy-preview-12945--spacy.netlify.app/api/large-language-models) instead of (only) looking at the final diff.

### Fixes
* HF models don't have a default model name
* REST models don't have a `timeout` parameter, but `interval` and `max_request_time`
* The default of `max_tries` is 5, not 3
* Fix template references (adding ".v1" and fixing URLs)
* Set `temperature` to `0.0` in example code
* Add `Llama2` to the usage overview table

### Cosmetic changes
* Removed duplicate information from tasks with increasing versions, focusing instead on what is different between implementations
* Changed some level headings and add task-specific sublevels so we can link to "the NER task" without having to specify a version, and add version-inspecific information there.
* Grouped models with the same parameters together, as the current page had become too difficult to process. This also avoids having to repeat the same GPU-install or env instructions over and over again.

### Types of change
docs update

## Checklist


svlandeg:
Configured Netlify, let me try to close & reopen to trigger a preview 🤞

website/docs/api/large-language-models.mdx
```diff
@@ -164,71 +175,27 @@ descriptions.
...
 > Label descriptions can also be used with explicit examples to give as much
 > info to the LLM model as possible.
 
-You can also write definitions for each label and provide them via the
-`label_definitions` argument. This lets you tell the LLM exactly what you're
-looking for rather than relying on the LLM to interpret its task given just the
-label name. Label descriptions are freeform so you can write whatever you want
-here, but through some experiments a brief description along with some examples
-and counter examples seems to work quite well.
+New to v2, is the fact that you can write definitions for each label and provide
```
rmitsch, line 193:
```suggestion
New to v2 is the fact that you can write definitions for each label and provide
```

website/docs/api/large-language-models.mdx
```diff
@@ -712,6 +644,53 @@ it's a function of type `Callable[[Iterable[Any]], Iterable[Any]]`, but specific
 implementations can have other signatures, like
 `Callable[[Iterable[str]], Iterable[str]]`.
 
+### Models via REST API {id="models-rest"}
+
+These models all take the same parameters:
```
rmitsch, line 651: This might be somewhat misleading. Maybe we should single out `config` and emphasize that its content has to depend on the LLM provider?

svlandeg: Good point. What about https://github.com/explosion/spaCy/pull/12945/commits/ccc3fc1318824c9006c0273f09a90431f32d3ac3?

rmitsch: Sounds good!

rmitsch (approved):

PR #12952, opened 2023-09-03
Add error handling if import of info function in __init__.py fails

Vuizur:
## Description
The info function import currently imports the entire CLI. So it imports packages like typer and requests and importing spacy fails if they are not available on the current system. When we run spacy on webassembly, we might not want to bundle all of these CLI libraries, and using requests requires currently an ugly workaround (https://github.com/koenvo/pyodide-http) and the other CLI libraries can't really be used as of now and would be "dead weight". In the case of failure, by replacing info with a placeholder function that raises an error when it is attempted to be used, we can simply use spacy as is.

### Types of change

Bug fix

## Checklist

### DIFF
--- a/spacy/__init__.py
+++ b/spacy/__init__.py
@@ -21,7 +21,7 @@ try:
     from .cli.info import info  # noqa: F401
 except ImportError:
 
-    def info(*args, **kwargs):
+    def info(*args, **kwargs):  # type: ignore
         raise RuntimeError("Importing the CLI failed. Info function not available.")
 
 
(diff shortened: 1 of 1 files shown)


spacy/__init__.py
```diff
@@ -13,7 +13,13 @@
...
 from . import util
 from .about import __version__  # noqa: F401
-from .cli.info import info  # noqa: F401
+try:  
+    # Emscripten does not make use of the CLI and can not use the CLI libraries
+    # (typer etc.) that are imported here. Additionally importing requests,
+    # which is also done here, requires a workaround on this platform.
+    from .cli.info import info  # noqa: F401
+except ImportError:
+    info = lambda *args: "Importing the CLI failed. Info function not available."
```
hoodmane, line 22: Maybe:
```suggestion
    def info(*args, **kwargs):
        raise RuntimeError("Importing the CLI failed. Info function not available.")
```

Vuizur: True, this is really better.

PR #12955, opened 2023-09-05
Fix in BertTokenizer docs

svlandeg:
## Description
If I understand the intended use correctly, this function registry should refer to the example `BertTokenizer` implementation from the section above it.

### Types of change
docs fix

## Checklist


website/docs/usage/linguistic-features.mdx
```diff
@@ -1301,7 +1301,7 @@ correct type.
 @spacy.registry.tokenizers("bert_word_piece_tokenizer")
 def create_whitespace_tokenizer(vocab_file: str, lowercase: bool):
```
adrianeboyd, line 1302:
```suggestion
def create_bert_tokenizer(vocab_file: str, lowercase: bool):
```

PR #12963, opened 2023-09-06
fix training.batch_size example

magdaaniol:
## Description

Just fixing the `training.batch_size` examples as `batch_size` is not a component of the `training` section of the config so using it causes Pydantic validation error.

### Types of change

change to the documentation

## Checklist


svlandeg (approved):
Nice catch, and thanks so much for consistently fixing this!

PR #12979, opened 2023-09-13
Redesigned cython profiling and other minor updates for python 3.12

adrianeboyd:
## Description

Swap `profile` default to `True` and disable entirely in setup for python 3.12+.

- Add `profile=False` to currently unprofiled cython
- Remove `profile=True` from currently profiled cython
- Set cython profiling default to `True` for <3.12, `False` for >=3.12

### Types of change

?

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,7 +58,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python_version: ["3.11"]
+        python_version: ["3.11", "3.12.0-rc.2"]
         include:
           - os: windows-latest
             python_version: "3.7"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -93,6 +93,7 @@ jobs:
       - name: Run mypy
         run: |
           python -m mypy spacy
+        if: matrix.python_version != '3.7'
 
       - name: Delete source directory and .egg-info
         run: |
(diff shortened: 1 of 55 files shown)


adrianeboyd:
This is intended to preserve the current behavior. We could potentially consider disabling profiling entirely by default.

rmitsch:
> This is intended to preserve the current behavior. We could potentially consider disabling profiling entirely by default.

Why would we disable it - wouldn't it better to stick with the current behavior?

adrianeboyd:
Most people never use profiling and there are slight performance hits, so it could be reasonable to have the default published wheels have no profiling, and have across-the-board profiling be an option that you can enable when compiling from source. (Managing that kind of option for `pip` through `setup.py` is probably kind of a hassle, though.)

rmitsch (approved):

rmitsch:
> (Managing that kind of option for pip through setup.py is probably kind of a hassle, though.)

In that case I'd stick with the current behavior.

adrianeboyd:
A short summary of the problems with tqdm before I force-push to clean up the history a bit:

* `tqdm` v4.66.1 produces a deprecation warning starting in python 3.12, which is then raised to an error in the test suite
* when the warning is raised to an error in the test suite, the test suite deadlocks consistently when both these tests are run:
  * `spacy/tests/test_cli.py::test_applycli_docbin`
  * `spacy/tests/test_cli.py::test_cli_find_threshold`

As in some older uses of `tqdm` in `spacy`, it looks like the best workaround is use `disable=None` so that tqdm output is disabled in the test suite and other non-interactive uses. This is not a solution for the underlying problem with deadlocks in `tqdm` (looking into the details makes me want to take tqdm out of spacy entirely, I feel like it's not really needed and would at least be better if it's entirely optional), but avoids the deadlocks and other problems related to raising errors in many everyday situations.

The test suite will not pass until the next version of `weasel` is released with some 3.12 updates and the remote tests are disabled in spacy.

PR #12981, opened 2023-09-13
add --spans-key option for CLI spancat evaluation

evornov:
## Description
The internal `evaluate` method called by the `evaluate`/`benchmark accuracy` CLI commands accepts a spans_key value, but the CLI command itself does not. This PR adds a `--spans-key` option so that the CLI can be used to evaluate spancat models with span key values other than the default `sc`.

I also updated the CLI documentation, but I didn't add a "new" tag because I'm not sure what the version would be there.

### Types of change
Minor CLI enhancement

## Checklist


adrianeboyd:
Thanks for the useful PR, especially with docs! I'm not sure why this wasn't added much sooner.

Do you mind if I push to your branch directly to rephrase the docs a bit? No problem if you'd rather not, I could do it in a follow-up PR instead.

evornov:
Yes, you can go ahead and push to my branch. Thanks!

PR #12962, opened 2023-09-06
Load the cli module lazily for spacy.info

danieldk:
## Description

This avoids that the `spacy` module cannot be imported when the users chooses not to install `typer`/`requests`.

Proposed alternative to #12952. With lazy loading, people that do have typer/requests installed will not pay the loading penalty when they just import `import spacy`.

### Types of change

Enhancement

## Checklist

- [x] I confirm that I have the right to submit this contribution under the project's MIT license.
- [x] I ran the tests, and all new and existing tests passed.
- [x] My changes don't require a change to the documentation, or if they do, I've added all required information.

PR #13040, opened 2023-10-04
Revert "Load the cli module lazily for spacy.info (#12962)"

adrianeboyd:
## Description

This reverts commit beda27a91eadd70563dbaffd844d8c9d5e245928.

### Types of change

Bug fix.

## Checklist


adrianeboyd:
I am not 100% sure that this is the source of the difference.

adrianeboyd:
It does look like this is the source of the problem (ugh).

adrianeboyd:
I ... have no idea how to write a test for this, so let's move forward just with the revert.

PR #13063, opened 2023-10-12
Update __all__ fields

svlandeg:
Update `__all__` fields across the main modules 
 
## Description
Ensure that the imported classes/functions can be easily imported from the main module instead of a submodule (and doesn't get flagged by type checking). For easier maintenance, alphabetize the lists.

### Types of change
fix

## Checklist

### DIFF
--- a/spacy/kb/__init__.py
+++ b/spacy/kb/__init__.py
@@ -1,3 +1,11 @@
 from .candidate import Candidate, get_candidates, get_candidates_batch
 from .kb import KnowledgeBase
 from .kb_in_memory import InMemoryLookupKB
+
+__all__ = [
+    "Candidate",
+    "KnowledgeBase",
+    "InMemoryLookupKB",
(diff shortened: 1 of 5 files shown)


rmitsch (approved):

kabirkhan:
Looks great, wondering if we should add an `__all__` to the main `__init__.py` file as well though
Currently for type checking (with pyright at least) you need to do:

```python
```

instead of just:

```python
from spacy import Language
```

adrianeboyd:
I'm actually not sure it's the intention of the API to support/prefer `from spacy import Language`? It looks like it's only imported in `__init__.py` for typing.

svlandeg:
As there is (only) discussion about the `__all__` field in the main `__init__.py` but the actual edits of this PR do not touch that file, I'll go ahead and get this PR merged and leave the rest up for further discussion, to ensure this one does not get blocked.

Issue #13056, opened 2023-10-10
Displacy render: Spans are overlapped when rendered.


mchlsam:
Hello,

I'm trying to display spans with displacy render where I built the span manually. 

Several spans can be overlapped.

Sometimes when there are more than 3 spans overlapping, the rendering fails to render properly. 

In the following image spans on the second line have 1 token in common so the render should have done 3 lines, instead it overlapped them.
![image](https://github.com/explosion/spaCy/assets/74212431/4856e75c-d6ef-46ff-aeb3-7578eb166186)

## How to reproduce the behaviour

```code
doc_rendering = {
    "text": "Welcome to the Bank of China.",
    "spans": [
        {"start_token": 2, "end_token": 5, "label": "SkillNC"},
        {"start_token": 0, "end_token": 2, "label": "Skill"},
        {"start_token": 1, "end_token": 3, "label": "Skill"},
    ],
    "tokens": ["Welcome", "to", "the", "Bank", "of", "China", "."],
}
```

```code
from spacy import displacy

html = displacy.render(
        doc_rendering,
        style="span",
        manual=True,
        options={"colors": {"Skill": "#56B4E9", "SkillNC": "#FF5733"}},
    )
```

## Your Environment

* Operating System: linux
* Python Version Used: 3.10
* spaCy Version Used: 3.6.1
* Environment Information: conda  23.5.2

Thanks for your help :)

PR #13068, opened 2023-10-18
Fix displacy span stacking

rmitsch:
## Description

Fix displacy span stacking by using the stack's last span's level instead of a counter not taking previous spans' levels into account.
Fixes https://github.com/explosion/spaCy/issues/13056.

Before: 
image

After:
image

### Types of change

Bug fix

## Checklist


svlandeg (commented):
Is there a unit test we could write to double check this behaviour?

rmitsch:
> Is there a unit test we could write to double check this behaviour?

Added one. Also refactored the relevant functionality to allow checking the data prior to rendering it as HTML.

PR #13081, opened 2023-10-23
Warn about reloading dependencies after downloading models

shadeMe:
## Description

This change accounts for model installations that took place during the current Python session.

### Types of change

Fix/Feature

## Checklist


spacy/errors.py
```diff
@@ -229,8 +229,9 @@ class Errors(metaclass=ErrorsWithCodes):
             "If you're using a Transformer, make sure to install 'spacy-transformers'. "
```
adrianeboyd, line 229: I think this may now be more confusing than helpful.

```suggestion
```

spacy/errors.py
```diff
@@ -229,8 +229,9 @@ class Errors(metaclass=ErrorsWithCodes):
...
             "If you're using a custom component, make sure you've added the "
             "decorator `@Language.component` (for function components) or "
-            "`@Language.factory` (for class components).\n\nAvailable "
-            "factories: {opts}")
+            "`@Language.factory` (for class components). If you installed the model "
+            "during the current Python session, you'll need to restart Python to "
```
adrianeboyd, line 233: I'm not sure people working in notebooks are going to understand "restart your runtime" from "restart Python"? But I may not be a very good judge of this...

spacy/errors.py
shadeMe: Ah, true. Overlooked that. Are we fine with suggesting both `spacy-transformers` and `spacy-curated-transformers` then? The other option would be to display the former for <=3.6.x and the latter for the rest.

spacy/errors.py
shadeMe: I've added Jupyter notebooks to the error message. Open to suggestions if it's still unclear.

spacy/errors.py
adrianeboyd: I think we'd just change it for the current v3.7 release, so this wouldn't affect v3.6. Overall it's getting too hairy (the error message is already extremely long) and I think it's better in general to assume that model packages handle their own dependencies. I think suggesting that people install both is not a good idea.

The main edge cases are mainly related to local saved files/directories, where you have a config or a local model directory that was generated in a venv with `spacy-transformers` (or whichever package) that you're now trying to load in a different venv. The package records the dependencies but the local files/dirs don't.

I guess I wouldn't be opposed to adding more targeted info to the error message if we can see from the missing registered function name that we should probably suggest package X (but we still can't do it 100% correctly because of the lack of actual namespacing).

spacy/errors.py
adrianeboyd: "restart your Python interpreter or runtime" (longer is just what we need here!)

adrianeboyd: Sorry, I think we posted at about the same time before and I didn't see your update before I wrote the comment above.

I'm also not sure that people will understand "reload their notebook". Let's see, jupyter says "restart kernel" and colab says "restart runtime".

> "If you're in an interactive session or notebook, you'll need to restart your runtime to ..." ???

Another idea: maybe we can put this in the download CLI message instead. And possibly detect notebooks / interactive sessions? We already have utils to detect jupyter for some other warnings/uses.

shadeMe: > Another idea: maybe we can put this in the download CLI message instead. And possibly detect notebooks / interactive sessions? We already have utils to detect jupyter for some other warnings/uses.

Yeah, that might be a better place for it. I'll give it a shot.

adrianeboyd: (These are basically the very hardest kind of error messages to write.)

shadeMe: I've given it the ol' college try. Feel free to change it.

spacy/errors.py
adrianeboyd: I still think we should take this line out now.

spacy/errors.py
adrianeboyd: I think this message needs to be pitched more at "novice notebook user" and we can also make an attempt to identify interactive sessions. Let me see...

spacy/errors.py
shadeMe: Removed that line.

spacy/errors.py
shadeMe: I've added an extra utility function to detect interactive sessions.

adrianeboyd:
I'd wondered a bit earlier this year why displacy in colab didn't seem to work as easily as in jupyter, so I suspect some of the auto-detection methods weren't working as intended or had gotten out-of-date.

These messages only work for `spacy.cli.download` not for `spacy download`, so it still be an issue in notebooks, but I think most users are aware of the need to restart at this point. There's probably a few more places this could be emphasized in the docs.

spacy/util.py
```diff
@@ -1077,20 +1077,38 @@ def force_remove(rmfunc, path, ex):
...
+            return True  # Colab notebook
     except NameError:
         return False  # Probably standard Python interpreter
+    # additional check for Colab
+    try:
+        import google.colab
+
+        return True  # Colab notebook
+    except ImportError:
+        return False
```
adrianeboyd, line 1099:
```suggestion
        pass  # Probably standard Python interpreter
    # additional check for Colab
    try:
        import google.colab

        return True  # Colab notebook
    except ImportError:
        pass
```

svlandeg (approved):
Looks good to me! Feel free to merge if you're happy with this, Adriane.

PR #13086, opened 2023-10-25
Update Tokenizer.explain for special cases with whitespace

adrianeboyd:
## Description

Update `Tokenizer.explain` to skip special case matches if the exact text has not been matched due to intervening whitespace.

### Types of change

Bug fix, tests.

## Checklist

### DIFF
--- a/spacy/tests/tokenizer/test_explain.py
+++ b/spacy/tests/tokenizer/test_explain.py
@@ -85,6 +85,18 @@ def test_tokenizer_explain_special_matcher(en_vocab):
     assert tokens == explain_tokens
 
 
+def test_tokenizer_explain_special_matcher_whitespace(en_vocab):
+    rules = {":]": [{"ORTH": ":]"}]}
+    tokenizer = Tokenizer(
+        en_vocab,
+        rules=rules,
+    )
+    text = ": ]"
+    tokens = [t.text for t in tokenizer(text)]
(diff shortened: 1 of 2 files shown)


spacy/tests/tokenizer/test_explain.py
```diff
@@ -123,6 +135,9 @@ def test_tokenizer_explain_fuzzy(lang: str, sentence: str) -> None:
     """
 
     tokenizer: Tokenizer = spacy.blank(lang).tokenizer
-    tokens = [t.text for t in tokenizer(sentence) if not t.is_space]
+    # Tokenizer.explain is not intended to handle whitespace or control
+    # characters in the same way as Tokenizer
```
rmitsch, line 139: Out of curiosity, why is that?

rmitsch (commented):
Looks reasonable to me, but I'm lacking experience with the tokenizer.

rmitsch (approved):
(Meant to approve before.)

spacy/tests/tokenizer/test_explain.py
adrianeboyd: Because it wasn't a priority when implementing it. The main thing was to make it easier to debug all the tokenizer regexes, which mostly operate on a single whitespace-bounded string. (Not 100% given the phrase matcher for special cases, which is what is affected here.)

PR #13108, opened 2023-11-05
Switch to build for package CLI

adrianeboyd:
## Description

Switch to `build` for package CLI.

### Types of change

Enhancement?

## Checklist

### DIFF
--- a/spacy/cli/package.py
+++ b/spacy/cli/package.py
@@ -11,6 +11,7 @@ from thinc.api import Config
 from wasabi import MarkdownRenderer, Printer, get_raw_input
 
 from .. import about, util
+from ..compat import importlib_metadata
 from ..schemas import ModelMetaSchema, validate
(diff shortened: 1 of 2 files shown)


adrianeboyd:
`python setup.py bdist_wheel` already shows warnings that it's deprecated and I suspect that `python setup.py sdist` will follow (although I haven't been following the details here carefully). In general calling `python setup.py` in any way is discouraged.

It's not a hassle exactly, but detecting whether `build` is installed is annoying. We could consider just adding `build` as a dependency and removing the check. We don't currently have `wheel` as

adrianeboyd:
We could also fall back to the previous `sdist`/`bdist_wheel` methods if `build` isn't installed, and then this could potentially target `master` instead. But `build` is the (current) future-friendly option.

adrianeboyd:
Closing in favor of #13109.

PR #13116, opened 2023-11-08
Feature/nn and fo language extensions

lise-brinck:
## Description
Added language extensions for Faroese and Norwegian Nynorsk.

### Types of change
This is an enhancement as it enhances the support of the spaCy language extensions.

## Checklist


adrianeboyd:
Thanks for the nice PR, especially for including all the citations and tests!

I'm a little concerned about the duplication between the `nb` and `nn` tokenizer settings. Are there many practical differences, e.g. with removing `'s` as a suffix for one or the other? (`nb` doesn't remove the kind of English-specific suffixes, but `nn` does.)

And do you mind if I push directly to this branch to add these languages to the website docs?

spacy/lang/nn/tokenizer_exceptions.py
```diff
@@ -0,0 +1,227 @@
...
+
+_exc = {}
+
+
+for exc_data in [
+    {ORTH: "jan.", NORM: "januar"},
+    {ORTH: "feb.", NORM: "februar"},
+    {ORTH: "mar.", NORM: "mars"},
+    {ORTH: "apr.", NORM: "april"},
+    {ORTH: "jun.", NORM: "juni"},
```
adrianeboyd, line 13: Is "juli" missing on purpose?

lise-brinck: Yes. The abbreviation "jul(.)" for "juli" also means "Christmas" i Norwegian Nynorsk.

lise-brinck:
> Thanks for the nice PR, especially for including all the citations and tests!
> 
> I'm a little concerned about the duplication between the `nb` and `nn` tokenizer settings. Are there many practical differences, e.g. with removing `'s` as a suffix for one or the other? (`nb` doesn't remove the kind of English-specific suffixes, but `nn` does.)
> 
> And do you mind if I push directly to this branch to add these languages to the website docs?

The Norwegian Nynorsk tokenizer is a mix of the Norwegian Bokmål and Danish tokenizers with the addition of some language-specific abbreviations. I'm honestly not too certain about the exact differences between the two variations of Norwegian, and I'm sure there's room for improvements by someone who knows Norwegian Nynorsk.

And sure, go ahead and push to this branch :)

spacy/lang/nn/tokenizer_exceptions.py
```diff
@@ -0,0 +1,227 @@
...
+    "hoh.",
+    "hr.",
+    "ifb.",
+    "ifm.",
+    "iht.",
+    "inkl.",
+    "istf.",
+    "jf.",
+    "jr.",
+    "jun.",
```
adrianeboyd, line 105: So you still have the abbreviation `jul.`, just not with the `NORM` `juli`?

Since `jun.` is above, I think this list can just be modified to have `jul.` instead:

```suggestion
    "jul.",
```

adrianeboyd: I'm not sure whether you ended to leave "jul." out as an abbreviation entirely. Since `nb` does include it, I decided to make the same decision to keep it here, just without the `NORM`, even though "jul." at the end of a sentence would be tokenized as intended for the non-July/abbreviation usage.

PR #13149, opened 2023-11-23
Update `TextCatBOW` to use the fixed `SparseLinear` layer

danieldk:
## Description

A while ago, we fixed the `SparseLinear` layer to use all available
parameters: https://github.com/explosion/thinc/pull/754

This change updates `TextCatBOW` to `v3` which uses the new
`SparseLinear_v2` layer. This results in a sizeable improvement on a
text categorization task that was tested.

While at it, this `spacy.TextCatBOW.v3` also adds the `length_exponent`
option to make it possible to change the hidden size. Ideally, we'd just
have an option called `length`. But the way that `TextCatBOW` uses
hashes results in a non-uniform distribution of parameters when the
length is not a power of two.

### Types of change

Bugfix

## Checklist

### DIFF
--- a/spacy/ml/models/textcat.py
+++ b/spacy/ml/models/textcat.py
@@ -111,15 +111,21 @@ def build_bow_text_classifier_v3(
     exclusive_classes: bool,
     ngram_size: int,
     no_output_layer: bool,
-    length_exponent: int = 18,
+    length: int = 262144,
     nO: Optional[int] = None,
 ) -> Model[List[Doc], Floats2d]:
+    if length < 1:
+        raise ValueError(Errors.E1056.format(length=length))
+
+    # Find k such that 2**(k-1) < length <= 2**k.
+    length = 2 ** (length - 1).bit_length()
+
     return _build_bow_text_classifier(
         exclusive_classes=exclusive_classes,
         ngram_size=ngram_size,
         no_output_layer=no_output_layer,
         nO=nO,
-        sparse_linear=SparseLinear_v2(nO=nO, length=2**length_exponent),
+        sparse_linear=SparseLinear_v2(nO=nO, length=length),
     )
 
 
--- a/spacy/ml/models/textcat.py
+++ b/spacy/ml/models/textcat.py
@@ -29,6 +29,7 @@ from thinc.layers.resizable import resize_linear_weighted, resize_model
 from thinc.types import ArrayXd, Floats2d
 
 from ...attrs import ORTH
+from ...errors import Errors
 from ...tokens import Doc
 from ...util import registry
 from ..extract_ngrams import extract_ngrams
--- a/spacy/ml/models/textcat.py
+++ b/spacy/ml/models/textcat.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import List, Optional, cast
+from typing import List, Optional, Tuple, cast
 
 from thinc.api import (
     Dropout,
@@ -12,6 +12,7 @@ from thinc.api import (
     Relu,
     Softmax,
     SparseLinear,
+    SparseLinear_v2,
     chain,
     clone,
     concatenate,
@@ -25,7 +26,7 @@ from thinc.api import (
 )
 from thinc.layers.chain import init as init_chain
 from thinc.layers.resizable import resize_linear_weighted, resize_model
-from thinc.types import Floats2d
+from thinc.types import ArrayXd, Floats2d
 
 from ...attrs import ORTH
 from ...tokens import Doc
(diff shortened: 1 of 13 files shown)


spacy/tests/pipeline/test_textcat.py
```diff
@@ -451,11 +451,16 @@ def test_no_resize(name, textcat_config):
...
+        # BOW V2
         ("textcat", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": True, "no_output_layer": False, "ngram_size": 3}),
         ("textcat", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": True, "no_output_layer": True, "ngram_size": 3}),
         ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": False, "no_output_layer": False, "ngram_size": 3}),
         ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": False, "no_output_layer": True, "ngram_size": 3}),
+        # BOW V3
+        ("textcat", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": True, "no_output_layer": False, "ngram_size": 3}),
+        ("textcat", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": True, "no_output_layer": True, "ngram_size": 3}),
+        ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": False, "no_output_layer": False, "ngram_size": 3}),
+        ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": False, "no_output_layer": True, "ngram_size": 3}),
```
svlandeg, line 463: Let's not copy all these tests, it'll blow up the test suite if we always do. We've also got `test_textcat_configs` that specifically tests older architectures. Here, I suggest to just update to the new one.

spacy/tests/pipeline/test_textcat.py
```diff
@@ -480,11 +485,16 @@ def test_resize(name, textcat_config):
...
+        # BOW v2
         ("textcat", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": True, "no_output_layer": False, "ngram_size": 3}),
         ("textcat", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": True, "no_output_layer": True, "ngram_size": 3}),
         ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": False, "no_output_layer": False, "ngram_size": 3}),
         ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v2", "exclusive_classes": False, "no_output_layer": True, "ngram_size": 3}),
+        # BOW v3
+        ("textcat", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": True, "no_output_layer": False, "ngram_size": 3}),
+        ("textcat", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": True, "no_output_layer": True, "ngram_size": 3}),
+        ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": False, "no_output_layer": False, "ngram_size": 3}),
+        ("textcat_multilabel", {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": False, "no_output_layer": True, "ngram_size": 3}),
```
svlandeg, line 497: Here as well I'd just update to use v3.

svlandeg (commented):
Looks great!

> Ideally, we'd just have an option called `length`. But the way that TextCatBOW uses
hashes results in a non-uniform distribution of parameters when the
length is not a power of two.

One alternative would be to take the highest power of two fitting within `length` if we don't want to expose these internals too much to users.

adrianeboyd:
If you do keep `length_exponent` as a direct setting, I would vote strongly for a different name that's easier for users to understand. I don't immediately know what, though...

danieldk:
> One alternative would be to take the highest power of two fitting within `length` if we don't want to expose these internals too much to users.

Rounding up seems like a good strategy. I'll update the PR to do that.

spacy/ml/models/textcat.py
```diff
@@ -95,10 +97,48 @@ def build_bow_text_classifier(
...
+    exclusive_classes: bool,
+    ngram_size: int,
+    no_output_layer: bool,
+    length: int = 262144,
+    nO: Optional[int] = None,
+) -> Model[List[Doc], Floats2d]:
+    if length < 1:
+        raise ValueError(Errors.E1056.format(length=length))
+
+    # Find k such that 2**(k-1) < length <= 2**k.
```
adrianeboyd, line 121: I'd include a note about why you are doing this (and I think it's confusing that you're reusing the variable name).

spacy/ml/models/textcat.py
```diff
@@ -95,10 +97,48 @@ def build_bow_text_classifier(
...
+        sparse_linear=SparseLinear(nO=nO),
+    )
+
+
+@registry.architectures("spacy.TextCatBOW.v3")
+def build_bow_text_classifier_v3(
+    exclusive_classes: bool,
+    ngram_size: int,
+    no_output_layer: bool,
+    length: int = 262144,
```
adrianeboyd, line 115: To line up more with `SparseLinear`:

```suggestion
    length: int = 2 ** 18,
```

adrianeboyd (commented):
I feel like the power-of-2 bit could be emphasized a bit more for people coming back to this code in the future.

PR #13181, opened 2023-12-05
Add TextCatReduce.v1

danieldk:
## Description

This is a textcat classifier that pools the vectors generated by a tok2vec implementation and then applies a classifier to the pooled representation. Three reductions are supported for pooling: first, max, and mean. When multiple reductions are enabled, the reductions are concatenated before providing them to the classification layer.

This model is a generalization of the TextCatCNN model, which only supports mean reductions and is a bit of a misnomer, because it can also be used with transformers. This change also reimplements `TextCatCNN.v2` using the new `TextCatReduce.v1` layer.

One thing I wasn't certain of is what to do with the `TextCatCNN.v2` docs. We could move them to legacy, however, this would mean that the existing _TextCatCNN_ does not work anymore and there may be existing pages and tutorials linking to that.

### Types of change

Feature

## Checklist


spacy/ml/models/textcat.py
```diff
@@ -230,3 +207,75 @@ def build_text_classifier_lowdata(
...
+) -> Model[List[Doc], Floats2d]:
+    """Build a model that classifies pooled `Doc` representations.
+
+    Pooling is performed using reductions. Reductions are concatenated when
+    multible reductions are used.
+
+    tok2vec (Model): the tok2vec layer to pool over.
+    exclusive_classes (bool): Whether or not classes are mutually exclusive.
+    use_reduce_first (bool): Pool by using the hidden representation of the
+        first token of a `Doc`
```
svlandeg, line 229:
```suggestion
        first token of a `Doc`.
```

spacy/ml/models/textcat.py
```diff
@@ -230,3 +207,75 @@ def build_text_classifier_lowdata(
...
+    exclusive_classes: bool,
+    use_reduce_first: bool,
+    use_reduce_max: bool,
+    use_reduce_mean: bool,
+    nO: Optional[int] = None,
+) -> Model[List[Doc], Floats2d]:
+    """Build a model that classifies pooled `Doc` representations.
+
+    Pooling is performed using reductions. Reductions are concatenated when
+    multible reductions are used.
```
svlandeg, line 224:
```suggestion
    multiple reductions are used.
```

svlandeg, line 224: Out of curiosity: have you experimented with having multiple reductions?

spacy/tests/pipeline/test_textcat.py
```diff
@@ -701,9 +701,9 @@ def test_overfitting_IO_multi():
...
         ("textcat", TRAIN_DATA_SINGLE_LABEL, {"@architectures": "spacy.TextCatEnsemble.v2", "tok2vec": DEFAULT_TOK2VEC_MODEL, "linear_model": {"@architectures": "spacy.TextCatBOW.v3", "exclusive_classes": True, "ngram_size": 5, "no_output_layer": False}}),
-        # CNN V2
-        ("textcat", TRAIN_DATA_SINGLE_LABEL, {"@architectures": "spacy.TextCatCNN.v2", "tok2vec": DEFAULT_TOK2VEC_MODEL, "exclusive_classes": True}),
-        ("textcat_multilabel", TRAIN_DATA_MULTI_LABEL, {"@architectures": "spacy.TextCatCNN.v2", "tok2vec": DEFAULT_TOK2VEC_MODEL, "exclusive_classes": False}),
+        # REDUCE V1
+        ("textcat", TRAIN_DATA_SINGLE_LABEL, {"@architectures": "spacy.TextCatReduce.v1", "tok2vec": DEFAULT_TOK2VEC_MODEL, "exclusive_classes": True, "use_reduce_first": True, "use_reduce_max": True, "use_reduce_mean": True}),
+        ("textcat_multilabel", TRAIN_DATA_MULTI_LABEL, {"@architectures": "spacy.TextCatReduce.v1", "tok2vec": DEFAULT_TOK2VEC_MODEL, "exclusive_classes": False, "use_reduce_first": True, "use_reduce_max": True, "use_reduce_mean": True}),
```
svlandeg, line 706: Here, I would keep the CNN V2 tests because we need to have at least one test checking that `spacy.TextCatCNN.v2` still works. We can put "CNN V2 (obsolete)" and then add the "REDUCE V1" tests as you have them here.

website/docs/api/architectures.mdx
```diff
@@ -1043,6 +1043,9 @@ A neural network model where token vectors are calculated using a CNN. The
 vectors are mean pooled and used as features in a feed-forward network. This
 architecture is usually less accurate than the ensemble, but runs faster.
 
+This model is identical to [TexCatReduce.v1](#TextCatReduce) with
+`use_reduce_mean=true`.
```
svlandeg, line 1047: Should we add that the other two should be set to `False`?

svlandeg (commented):
Nice!

> One thing I wasn't certain of is what to do with the `TextCatCNN.v2` docs. We could move them to legacy, however, this would mean that the existing TextCatCNN does not work anymore and there may be existing pages and tutorials linking to that.

The docs can be moved to [legacy](https://spacy.io/api/legacy), and the code can go into [spacy-legacy](https://github.com/explosion/spacy-legacy/), but all existing code should ideally just continue to work as-is. Why wouldn't/couldn't it?

https://spacy.io/api/architectures#TextCatCNN could just refer to the legacy docs and/or the new `TextCatReduce`.

danieldk:
> Nice!
> 
> > One thing I wasn't certain of is what to do with the `TextCatCNN.v2` docs. We could move them to legacy, however, this would mean that the existing TextCatCNN does not work anymore and there may be existing pages and tutorials linking to that.
> 
> The docs can be moved to [legacy](https://spacy.io/api/legacy), and the code can go into [spacy-legacy](https://github.com/explosion/spacy-legacy/), but all existing code should ideally just continue to work as-is. Why wouldn't/couldn't it?

I just noticed that there is a word missing in my initial comment:

> One thing I wasn't certain of is what to do with the `TextCatCNN.v2` docs. We could move them to legacy, however, this would mean that the existing TextCatCNN (documentation) **anchor** does not work anymore and there may be existing pages and tutorials linking to that.

Currently, pages can link to https://spacy.io/api/architectures#TextCatCNN

spacy/tests/pipeline/test_textcat.py
danieldk: Done.

website/docs/api/architectures.mdx
danieldk: Done.

danieldk:
> > One thing I wasn't certain of is what to do with the `TextCatCNN.v2` docs. We could move them to legacy, however, this would mean that the existing _TextCatCNN_ does not work anymore and there may be existing pages and tutorials linking to that.
> 
> I think you can add a redirect to `netlify.toml` to handle that kind of change.

Cool, works with anchors too 🙏 .

netlify.toml
```diff
@@ -55,6 +55,7 @@ redirects = [
     {from = "/models/comparison", to = "/models", force = true},
     {from = "/api/#section-cython", to = "/api/cython", force = true},
     {from = "/api/#cython", to = "/api/cython", force = true},
+    {from = "/api/architectures#TextCatCNN", to = "/api/legacy#TextCatCNN_v2", force = true},
```
adrianeboyd, line 58: Do we want to remove this now since it's kind of misleading?

danieldk: Done.

spacy/cli/templates/quickstart_training.jinja
```diff
@@ -279,8 +279,12 @@ no_output_layer = false
 
 {% else -%}
 [components.textcat.model]
-@architectures = "spacy.TextCatCNN.v2"
+@architectures = "spacy.TextCatReduce.v1"
```
svlandeg, line 282: This is used by the [generator](https://spacy.io/usage/training#quickstart) in the docs too, right? For the sake of compatibility with older spaCy versions, I think that we should consider keeping the old name here for a bit, as it's functionally the same anyway.

danieldk: Fixed.

spacy/errors.py
```diff
@@ -984,6 +984,9 @@ class Errors(metaclass=ErrorsWithCodes):
     E1055 = ("The 'replace_listener' callback expects {num_params} parameters, "
              "but only callbacks with one or three parameters are supported")
     E1056 = ("The `TextCatBOW` architecture expects a length of at least 1, was {length}.")
+    E1057 = ("The `TextCatReduce` architecture must be used with at least one reduction. "
+             "Please enable one of `use_reduce_first`, `reduce_last`, `use_reduce_max` or "
```
adrianeboyd, line 988:
```suggestion
             "Please enable one of `use_reduce_first`, `use_reduce_last`, `use_reduce_max` or "
```

spacy/ml/models/textcat.py
```diff
@@ -230,3 +209,80 @@ def build_text_classifier_lowdata(
...
+    """Build a model that classifies pooled `Doc` representations.
+
+    Pooling is performed using reductions. Reductions are concatenated when
+    multiple reductions are used.
+
+    tok2vec (Model): the tok2vec layer to pool over.
+    exclusive_classes (bool): Whether or not classes are mutually exclusive.
+    use_reduce_first (bool): Pool by using the hidden representation of the
+        first token of a `Doc`.
+    use_reduce_first (bool): Pool by using the hidden representation of the
```
adrianeboyd, line 233:
```suggestion
    use_reduce_last (bool): Pool by using the hidden representation of the
```

spacy/tests/test_models.py
```diff
@@ -26,6 +26,8 @@
     build_Tok2Vec_model,
 )
 from spacy.ml.staticvectors import StaticVectors
+from spacy.pipeline import tok2vec
```
adrianeboyd, line 29:
```suggestion
```

PR #13187, opened 2023-12-07
Type documentation fixes for Doc

svlandeg:
## Description

- Fix `doc.pyi` to include the fact that `char_span()` can return `None`
- Unify type documentation of `exclude`: it's a `tuple` by default, originally documented to be a `List`. Can be documented as an `Iterable[str]`, like we do for `util.to_dict` and the `to_disk` method in `doc.pyi`.

### Types of change
docs fixes

## Checklist


spacy/tokens/doc.pyx
```diff
@@ -1361,11 +1361,9 @@ cdef class Doc:
     def to_dict(self, *, exclude=tuple()):
         """Export the document contents to a dictionary for serialization.
 
-        exclude (list): String names of serialization fields to exclude.
+        exclude (Iterable[str]): String names of serialization fields to exclude.
         RETURNS (bytes): A losslessly serialized copy of the `Doc`, including
             all annotations.
-
-        DOCS: https://spacy.io/api/doc#to_bytes
```
svlandeg, line 1368: This link was wrong, this method is currently not documented. Alternatively, we could go ahead and document it and fix this link. Not sure whether this needs to be in the public API though.

adrianeboyd: I'm confused, this is part of the public API and is here: https://spacy.io/api/doc#to_bytes

spacy/tokens/doc.pyx
```diff
@@ -1361,11 +1361,9 @@ cdef class Doc:
     def to_dict(self, *, exclude=tuple()):
         """Export the document contents to a dictionary for serialization.
 
-        exclude (list): String names of serialization fields to exclude.
+        exclude (Iterable[str]): String names of serialization fields to exclude.
         RETURNS (bytes): A losslessly serialized copy of the `Doc`, including
```
adrianeboyd, line 1365:
```suggestion
        RETURNS (Dict): A losslessly serialized copy of the `Doc`, including
```

spacy/tokens/doc.pyi
```diff
@@ -179,16 +179,12 @@ class Doc:
...
-    ) -> Doc: ...
-    def to_dict(self, *, exclude: Union[List[str], Tuple[str]] = ...) -> bytes: ...
-    def from_dict(
-        self, msg: bytes, *, exclude: Union[List[str], Tuple[str]] = ...
+        self, path: Union[str, Path], *, exclude: Iterable[str] = ...
     ) -> Doc: ...
+    def to_bytes(self, *, exclude: Iterable[str] = ...) -> bytes: ...
+    def from_bytes(self, bytes_data: bytes, *, exclude: Iterable[str] = ...) -> Doc: ...
+    def to_dict(self, *, exclude: Iterable[str] = ...) -> bytes: ...
+    def from_dict(self, msg: bytes, *, exclude: Iterable[str] = ...) -> Doc: ...
```
adrianeboyd, line 187:
```suggestion
    def to_dict(self, *, exclude: Iterable[str] = ...) -> Dict: ...
    def from_dict(self, msg: Dict, *, exclude: Iterable[str] = ...) -> Doc: ...
```

spacy/tokens/doc.pyx
adrianeboyd: `from_dict` needs the same change, but I can't propose an edit here.

PR #13191, opened 2023-12-08
Revert the parser refactor

danieldk:
## Description

Reverse the parser refactor until we shake out the remaining issues.

### Types of change

Maintenance.

## Checklist

### DIFF
--- a/spacy/tests/test_cli.py
+++ b/spacy/tests/test_cli.py
@@ -1042,6 +1042,13 @@ def test_applycli_user_data():
         assert result[0]._.ext == val
 
 
+# TODO: remove this xfail after merging master into v4. The issue
+#       is that for local files, pathy started returning os.stat_result,
+#       which doesn't have a last_modified property. So, recency-sorting
+#       fails and the test fails. However, once we merge master into
+#       v4, we'll use weasel, which in turn uses cloudpathlib, which
+#       should resolve this issue.
+@pytest.mark.xfail(reason="Recency sorting is broken on some platforms")
 def test_local_remote_storage():
     with make_tempdir() as d:
         filename = "a.txt"


spacy/errors.py
```diff
@@ -215,12 +215,6 @@ class Warnings(metaclass=ErrorsWithCodes):
             "key attribute for vectors, configure it through Vectors(attr=) or "
             "'spacy init vectors --attr'")
 
-    # v4 warning strings
-    W400 = ("`use_upper=False` is ignored, the upper layer is always enabled")
-    W401 = ("`incl_prior is True`, but the selected knowledge base type {kb_type} doesn't support prior probability "
-            "lookups so this setting will be ignored. If your KB does support prior probability lookups, make sure "
-            "to return `True` in `.supports_prior_probs`.")
-
```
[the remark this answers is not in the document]
danieldk: Fixed.

website/docs/usage/embeddings-transformers.mdx
```diff
@@ -140,7 +140,7 @@ factory = "tok2vec"
 factory = "ner"
 
 [components.ner.model]
-@architectures = "spacy.TransitionBasedParser.v3"
+@architectures = "spacy.TransitionBasedParser.v1"
```
adrianeboyd, line 143: ? and below

```suggestion
@architectures = "spacy.TransitionBasedParser.v2"
```

danieldk: Thanks! Fixed.

spacy/compat.py
```diff
@@ -23,6 +23,11 @@
 except ImportError:
     cupy = None
 
+if sys.version_info[:2] >= (3, 8):  # Python 3.8+
+    from typing import Literal, Protocol, runtime_checkable
+else:
+    from typing_extensions import Literal, Protocol, runtime_checkable  # noqa: F401
+
```
adrianeboyd, line 30: This shouldn't be necessary since we've dropped python 3.7 already.

spacy/errors.py
```diff
@@ -1000,6 +999,7 @@ class Errors(metaclass=ErrorsWithCodes):
     E4011 = ("Server error ({status_code}), couldn't fetch {url}")
 
 
+
```
adrianeboyd, line 1002:
```suggestion
```

spacy/ml/models/parser.py
```diff
@@ -1,66 +1,24 @@
-import warnings
-from typing import Any, List, Literal, Optional, Tuple
+from typing import List, Optional, cast
```
adrianeboyd, line 1:
```suggestion
from typing import List, Literal, Optional, cast
```

spacy/ml/models/parser.py
```diff
@@ -1,66 +1,24 @@
...
 
-from thinc.api import Model
+from thinc.api import Linear, Model, chain, list2array, use_ops, zero_init
 from thinc.types import Floats2d
 
-from ...errors import Errors, Warnings
-from ...tokens.doc import Doc
+from ...compat import Literal
```
adrianeboyd, line 6:
```suggestion
```

adrianeboyd (commented):
We don't need to bring back python <3.8 compat.

spacy/compat.py
danieldk: Fixed.

website/docs/api/cli.mdx
```diff
@@ -361,7 +361,7 @@ Module     spacy.language
 File       /path/to/spacy/language.py (line 64)
 ℹ [components.ner.model]
 Registry   @architectures
-Name       spacy.TransitionBasedParser.v3
+Name       spacy.TransitionBasedParser.v1
```
adrianeboyd, line 364: Also ?

```suggestion
Name       spacy.TransitionBasedParser.v2
```

website/docs/api/cli.mdx
```diff
@@ -371,7 +371,7 @@ Module     spacy.ml.models.tok2vec
 File       /path/to/spacy/ml/models/tok2vec.py (line 16)
 ℹ [components.parser.model]
 Registry   @architectures
-Name       spacy.TransitionBasedParser.v3
+Name       spacy.TransitionBasedParser.v1
```
adrianeboyd, line 374:
```suggestion
Name       spacy.TransitionBasedParser.v2
```

website/docs/api/cli.mdx
danieldk: I wasn't to sure about that, since it's an example with program output.

adrianeboyd:
This is mildly challenging to review!

I can confirm that the changes to the tests look fine (everything reverted except the distillation tests, right?).

Nothing else jumps out at me at this point, so I think we'll just have to keep this mind when merging in `develop` later. I think in particular we might have to sort out the cython profiling settings a bit, but that should be clear from the python 3.12 tests.

website/docs/api/cli.mdx
adrianeboyd: Ah, you're right. I think this doesn't belong in this PR. (On the other hand, it doesn't matter that much either at this point...)

svlandeg (approved):
As Adriane states, pretty challenging to review 🙈 

I think we could go ahead and merge though, so we can get through the next steps...

PR #13201, opened 2023-12-18
Add spacy.TextCatParametricAttention.v1

danieldk:
## Description

This layer provides is a simplification of the ensemble classifier that only uses paramteric attention. We have found empirically that with a sufficient amount of training data, using the ensemble classifier with BoW does not provide significant improvement in classifier accuracy. However, plugging in a BoW classifier does reduce GPU training and inference performance substantially, since it uses a GPU-only kernel.

### Types of change

Feature

## Checklist


svlandeg (approved):
Looks great!

PR #13203, opened 2023-12-19
Remove debug data normalization for span analysis

adrianeboyd:
## Description

As a result of this normalization, `debug data` could show a user tokens that do not exist in their data.

### Types of change

?

## Checklist


adrianeboyd:
I don't really like it, but I don't know whether the lowercasing is something that's typically used in related work. We can at least remove the confusing quote normalization. Not that it really matters that much either way...

spacy/cli/debug_data.py
```diff
@@ -1073,9 +1073,7 @@ def _get_distribution(docs, normalize: bool = True) -> Counter:
     word_counts: Counter = Counter()
     for doc in docs:
         for token in doc:
-            # Normalize the text
-            t = token.text.lower().replace("``", '"').replace("''", '"')
-            word_counts[t] += 1
+            word_counts[token.text] += 1
```
adrianeboyd, line 1076:
```suggestion
            t = token.text.lower()
            word_counts[t] += 1
```

svlandeg:
Yea I think it makes sense to remove the quote normalization.

svlandeg (approved):

PR #13240, opened 2024-01-16
Fix typo in method name

mauricesvp:
## Checklist


svlandeg (commented):
Thanks! This method doesn't seem to be internally used or even documented, so I guess we could just go ahead and merge the fix - let me just run this by Daniël who's been working on the parser code as well.

danieldk:
Thanks for fixing this! Are you using this property? If not, we might as well remove it altogether, as it's an internal class and the property is not used in spaCy.

mauricesvp:
I'm not using it myself, I only noticed the typo while browsing through some internals.

PR #13247, opened 2024-01-18
add custom code support to CLI speed benchmark

evornov:
## Description
Adds `--code` option to the speed benchmark CLI command so that pipelines with custom components can be benchmarked.
I used `[doc.from_docs([doc]) for doc in docs]` to deep-copy the warmup docs to prevent errors with pipelines that modify sentence boundaries. Happy to change this to something else if there's a more idiomatic way to copy those docs!

### Types of change
CLI enhancement + associated documentation update

## Checklist

### DIFF
--- a/spacy/cli/benchmark_speed.py
+++ b/spacy/cli/benchmark_speed.py
@@ -13,7 +13,7 @@ from .. import util
 from ..language import Language
 from ..tokens import Doc
 from ..training import Corpus
-from ._util import Arg, Opt, benchmark_cli, setup_gpu
+from ._util import Arg, Opt, benchmark_cli, setup_gpu, import_code
 
 
 @benchmark_cli.command(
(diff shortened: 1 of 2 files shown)


spacy/cli/benchmark_speed.py
```diff
@@ -171,5 +173,7 @@ def print_outliers(sample: numpy.ndarray):
 def warmup(
     nlp: Language, docs: List[Doc], warmup_epochs: int, batch_size: Optional[int]
 ) -> numpy.ndarray:
-    docs = warmup_epochs * docs
-    return annotate(nlp, docs, batch_size)
+    warmup_docs = []
+    for _ in range(warmup_epochs):
+        warmup_docs += [doc.from_docs([doc]) for doc in docs]
+    return annotate(nlp, warmup_docs, batch_size)
```
danieldk, line 179: `Doc` has a `copy` method, so something like the following should work:

```suggestion
    return annotate(nlp, [doc.copy() for doc in docs], batch_size)
```

danieldk (commented):
Thanks for the PR, we forgot this option when adding the `benchmark speed` subcommand.

danieldk (approved):

danieldk:
Awesome, thanks again for the PR!

PR #13249, opened 2024-01-19
`TextCatParametricAttention.v1`: set key transform dimensions

danieldk:
This is necessary for tok2vec implementations that initialize lazily (e.g. curated transformers).

## Description

Set key transform dimensions in `TextCatParametricAttention.v1`, otherwise the key transformation dimensionality cannot be inferred for lazily-initialized tok2vecs listeners like transformers.

Not sure how we can test this without making a new tok2vec model in spaCy that is lazily initialized (does know its width until initialization). Maybe we should add a separate model to the tests?

## Checklist

### DIFF
--- a/spacy/tests/tok2vec.py
+++ b/spacy/tests/tok2vec.py
@@ -1,9 +1,10 @@
 from typing import List
 
-from spacy.util import registry
-from thinc.types import Floats2d
-from spacy.tokens import Doc
 from thinc.api import Model
+from thinc.types import Floats2d
+
+from spacy.tokens import Doc
+from spacy.util import registry
 
 
 @registry.architectures("LazyInitTok2Vec.v1")
--- a/spacy/tests/tok2vec.py
+++ b/spacy/tests/tok2vec.py
@@ -1,0 +1,35 @@
+from typing import List
+
+from spacy.util import registry
+from thinc.types import Floats2d
+from spacy.tokens import Doc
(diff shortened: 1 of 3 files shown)


svlandeg (commented):
Looks good.

Would it be much work to make a sort of mockup `tok2vec` in the test suite for this purpose? I do think it's the sort of thing where we should have better tests because these things may come to bite us again in the future...

danieldk:
> Would it be much work to make a sort of mockup `tok2vec` in the test suite for this purpose? I do think it's the sort of thing where we should have better tests because these things may come to bite us again in the future...

Agreed, I'll give it a try. Might reveal issues in the other implementations as well.

danieldk:
Added, this uncovered two more issues.

danieldk:
Tests break currently because of the issue that #13284 fixes.

danieldk:
Ready for another review round.

spacy/tests/tok2vec.py
```diff
@@ -0,0 +1,36 @@
+from typing import List
+
+from thinc.api import Model
+from thinc.types import Floats2d
+
+from spacy.tokens import Doc
+from spacy.util import registry
+
+
+@registry.architectures("LazyInitTok2Vec.v1")
```
svlandeg, line 10: Maybe we should add a "test." prefix just to be fully clear this isn't meant to be used in any way?

PR #13255, opened 2024-01-20
test_find_available_port: use port 5001

danieldk:
## Description

macOS now uses port 5000 for the AirPlay receiver functionality, so this test will always fail on a macOS desktop (unless AirPlay receiver functionality is disabled like in CI).

### Types of change

Bugfix

## Checklist


svlandeg (commented):
Hm, it looks like we're still seeing the same error on the CI?

Issue #13230, opened 2024-01-10
Misleading spacy apply behavior when using .spacy file


didmar:
When using the `spacy apply` command line with a .spacy format file as the input, existing entities from that input file are kept and the model won't predict anything for the corresponding tokens.
I was using it to run a model on a dataset that has gold standard annotations, and was initially surprised that there was not a single false negative in the resulting output file.

The documentation says nothing about this behavior, and my expectation was that those entities would be overridden by my predictions.

Maybe there could be an option to override this behavior and decide to ignore any entities from the input? Or does it goes against the design philosophy?

## Which page or section is this issue related to?

https://spacy.io/api/cli#apply

didmar:
OK, it makes sense.

I think a simple mention in the description for `data_path` could help, such as "Note that any existing properties will **not** be overridden by the model's prediction."

PR #13272, opened 2024-01-25
Clarify data_path loading for apply CLI command

svlandeg:
Fixes https://github.com/explosion/spaCy/issues/13230

## Description
Attempt to clarify that if a specified .spacy file contains annotations, that these will be preserved by the `apply` CLI command, and run through the pipeline as is. There are valid cases to do for this, like e.g. evaluating an EL pipeline on files that have gold entities, but the behaviour may be surprising for users assuming that the pipeline will be applied to raw text only.

### Types of change
doc update

## Checklist

### DIFF
--- a/website/docs/api/cli.mdx
+++ b/website/docs/api/cli.mdx
@@ -1296,7 +1296,7 @@ input formats are:
 
 When a directory is provided it is traversed recursively to collect all files.
 
-When loading a .spacy file, any potential annotations stored on the `Doc` that are not overwritten by a pipe will be preserved.
+When loading a .spacy file, any potential annotations stored on the `Doc` that are not overwritten by the pipeline will be preserved.
 If you want to evaluate the pipeline on raw text only, make sure that the .spacy file does not contain any annotations.
 
 ```bash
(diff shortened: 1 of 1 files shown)


website/docs/api/cli.mdx
```diff
@@ -1296,6 +1296,9 @@ input formats are:
 
 When a directory is provided it is traversed recursively to collect all files.
 
+When loading a .spacy file, any potential annotations stored on the `Doc` will be preserved.
```
danieldk, line 1299: Mostly for my understanding: whether the annotation will be preserved depends on whether the pipe is set to `overwrite`?

If so, maybe it should say something like:

> When loading a .spacy file, any potential annotations stored on the `Doc` that are not overwritten by a pipe will be preserved.

svlandeg: What I intended to say is that annotations are preserved upon loading the `Doc`. It entirely depends on the implementation of the actual pipeline what happens to them afterwards... I'll think about how to rephrase this more clearly.

danieldk (approved):

Issue #13252, opened 2024-01-19
Vocab Issue


lordsoffallen:
## How to reproduce the behaviour

![image](https://github.com/explosion/spaCy/assets/20232088/0174a4e4-b2aa-44fb-b7af-b817abc88dba)
I am trying to find a word in the vocab and testing the example provided in the documentation. However I see that word apple is not in the vocab. Am I doing something wrong here? How can I check if a word exist in the vocab?

## Info about spaCy

- **spaCy version:** 3.7.2
- **Platform:** Linux-5.15.133+-x86_64-with-glibc2.31
- **Python version:** 3.10.12
- **Pipelines:** en_core_web_sm (3.7.1), en_core_web_lg (3.7.1)

danieldk:
Reference to docs with the example: https://spacy.io/api/vocab#contains

svlandeg:
You're right - this is confusing and the example is not complete. It sort of just assumes that at some point, the pipeline has seen the word "apple" because it's so common.

I fixed the example in the following PR: https://github.com/explosion/spaCy/pull/13273

If you try this, you'll see it works, because the `Vocab` works as a sort of cache.

lordsoffallen:
hey @danieldk @svlandeg,

Thanks for the update. My question would be how can i access a let's say english vocab of a model. In nltk we can just download this and it has about 250k words there i believe. Is there a way to get a whole vocabulary in spacy? To me that's still not clear

PR #13273, opened 2024-01-25
Clarify vocab docs

svlandeg:
Fixes https://github.com/explosion/spaCy/issues/13252

## Description
The example in the docs assumes that `"apple"` is in the vocab but this isn't necessarily the case. Expanding the example to ensure that it is, and to (hopefully) clarify that the vocab works as a cache.

### Types of change
docs update

## Checklist

### DIFF
--- a/website/docs/api/vocab.mdx
+++ b/website/docs/api/vocab.mdx
@@ -13,7 +13,7 @@ between `Doc` objects.
 <Infobox variant ="warning">
 
 Note that a `Vocab` instance is not static. It increases in size as texts with
-new tokens are processed.
+new tokens are processed. Some models may have an empty vocab at initialization.
 
 </Infobox>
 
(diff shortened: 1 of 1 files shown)


website/docs/api/vocab.mdx
```diff
@@ -93,6 +93,7 @@ given string, you need to look it up in
 > #### Example
 >
 > ```python
+> nlp("I'm eating an apple")
```
danieldk, line 96: Maybe we should also add something to the blurb at the beginning of the page? It currently says:

> Note that a `Vocab` instance is not static. It increases in size as texts with
new tokens are processed.

I think it might not be clear that models do not have a vocabulary per sé (perhaps outside the word embeddings if a non-transformer model used).

svlandeg: I added 
> Some models may have an empty vocab at initialization.

What do you think?

danieldk (approved):

PR #13284, opened 2024-01-29
Extension serialization attr tests: add teardown

danieldk:
## Description

The doc/token extension serialization tests add extensions that are not
serializable with pickle. This didn't cause issues before due to the
implicit run order of tests. However, test ordering has changed with
pytest 8.0.0, leading to failed tests in test_language.

Update the fixtures in the extension serialization tests to do proper
teardown and remove the extensions.

### Types of change

Bugfix

## Checklist


svlandeg (approved):
Thanks for the quick fix!

PR #13317, opened 2024-02-09
extend abbreviations list in Faroese tokenizer exceptions

lise-brinck:
Extended list of abbreviations in Faroese language extension's tokenizer exceptions.

## Checklist


svlandeg (commented):
Thanks! We could merge this into `develop`, as we typically don't want to make breaking changes to the language data in bugfix releases - instead it would go into the next minor release (so 3.8 instead of 3.7.5, for instance)

Could you update your branch to base off from `develop` cleanly? (or start a new PR to target `develop` if that's easier). I can't do it automatically as the branches aren't currently synced, and I don't want other contributions from `develop` showing up on this PR.

lise-brinck:
New pull request into develop: #13366

PR #13337, opened 2024-02-19
add coordination ruler

india-kerle:
## Checklist

### DIFF
--- a/spacy/pipeline/coordinationruler.py
+++ b/spacy/pipeline/coordinationruler.py
@@ -1,0 +1,321 @@
+from typing import List, Callable, Optional, Union
+from pydantic import BaseModel, validator
+import re
+import en_core_web_sm
+
+from ..tokens import Doc
+from ..language import Language
+from ..vocab import Vocab
+from .pipe import Pipe
+
+########### DEFAULT COORDINATION SPLITTING RULES ##############
+
+
+def _split_duplicate_object(doc: Doc) -> Union[List[str], None]:
+    """Split a text with 2 verbs and 1 object (and optionally a subject) into
+       2 texts each with 1 verb, the shared object (and its modifiers), and the subject if present.
+
+    i.e. 'I use and provide clinical supervision' -->
+    ['I use clinical supervision', 'I provide clinical supervision']
+
+    Args:
+        doc (Doc): The spaCy Doc object.
+
+    Returns:
+        List[str]: The split texts.
+    """
+    sentences = []
+
+    for token in doc:
+        if token.pos_ == "VERB" and (token.dep_ == "ROOT" or token.dep_ == "conj"):
+
+            has_AND = False
+            has_second_verb = False
+            has_dobj = False
+            subject = None
+
+            # Find the subject if it exists
+            for possible_subject in token.head.children:
+                if possible_subject.dep_ in ["nsubj", "nsubjpass"]:
+                    subject = possible_subject
+                    break
+
+            for child in token.children:
+
+                if child.pos_ == "CCONJ" and child.lemma_ == "and":
+                    has_AND = True
+
+                if child.pos_ == "VERB" and child.dep_ == "conj":
+                    has_second_verb = True
+                    second_verb = child
+                    first_verb = token.head if token.dep_ == "conj" else token
+
+                    for descendant in second_verb.subtree:
+                        if descendant.dep_ == "dobj":
+                            has_dobj = True
+                            # Collect the full noun phrase for the direct object
+                            dobj_span = doc[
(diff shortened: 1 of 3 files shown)


honnibal:
Thanks! Really excited to have something like this in the library.

> I've expanded the initial splitting rules very slightly to be more generalisable to full sentences and not the original skill spans. Should I add additional generalisable splitting rules? There is also a very specific skill splitting function i.e. the token skill must be at the end of phrase.

I think one construction that will be especially useful to people is coordination of modifiers in noun phrases. This could be coordination of adjectives, or nouns themselves. Section 2.2 of this thesis has a nice background on one type of construction that will be important to think about, compound nouns: https://www.researchgate.net/profile/Mark-Lauer-2/publication/2784243_Designing_Statistical_Language_Learners_Experiments_on_Noun_Compounds/links/53f9ccf60cf2e3cbf5604ec4/Designing-Statistical-Language-Learners-Experiments-on-Noun-Compounds.pdf

In general we'd like to detect and process stuff like "green and red apples" into "green apples" and "red apples". But we can have deeper nesting than that: stuff like "hot and cold chicken soup", which ends up as "hot chicken soup" and "cold chicken soup". Ultimately we're going to trust the tree structure in the parser (which isn't always fantastic on these things, due to limitations in the training data annotation) but we still want to have some concept of the range of tree shapes so we can make the test cases for them.

I would suggest first focussing on the cases where we have coordination inside a noun phrase. These will be the ones most useful for entity recognition. If we can enumerate the main construction cases we want to cover, we can then put together the target trees for them, and then test for those. For the tests, we definitely want to specify the dependency parse as part of the test case rather than letting it be predicted by the model. This way the test describes the tree, and also if we have different versions of the model the test doesn't break because it predicted something unexpected.

> I made this a factory as opposed to a function component because I thought it would be nice for users to be able to add their own custom rules - thoughts?

Yes the extensibility is definitely good. Arguably we also want to support matcher or dependency matcher patterns directly, but this could be done via a function that takes the patterns as an argument.

spacy/pipeline/coordinationruler.py
```diff
@@ -0,0 +1,321 @@
+from typing import List, Callable, Optional, Union
+from pydantic import BaseModel, validator
+import re
+import en_core_web_sm
```
svlandeg, line 4: We'll want to find another solution for this, because we don't want to enforce all users to have exactly this model in their environment

spacy/pipeline/coordinationruler.py
```diff
@@ -0,0 +1,321 @@
...
+
+from ..tokens import Doc
+from ..language import Language
+from ..vocab import Vocab
+from .pipe import Pipe
```
svlandeg, line 9: Could you run `isort` on all files? (the test suite will fail otherwise)

spacy/pipeline/coordinationruler.py
```diff
@@ -0,0 +1,187 @@
...
+
+########### DEFAULT COORDINATION SPLITTING RULES ##############
+ 
+def split_noun_coordination(doc: Doc) -> Union[List[str], None]:
+    """Identifies and splits phrases with multiple nouns, a modifier
+        and a conjunction.
```
india-kerle, line 14: this doesn't work yet on phrases with multiple modifiers i.e. "juicy apples and rotten oranges" but this is what I'm thinking in terms of how to approach it i.e. a test to ensure the phrase should be split and then a splitting rule

spacy/tests/pipeline/test_coordinationruler.py
```diff
@@ -0,0 +1,166 @@
...
+from spacy.tokens import Doc
+import spacy
+
+from spacy.pipeline.coordinationruler import split_noun_coordination
+
+@pytest.fixture
+def nlp():
+    return spacy.blank("en")
+
+### NOUN CONSTRUCTION CASES ###
```
india-kerle, line 13: and then this is how i'm testing it - manually defining pos_tags, dep_relations and relationships between toks for different construction cases

spacy/pipeline/coordinationruler.py
india-kerle: FYI @honnibal

spacy/tests/pipeline/test_coordinationruler.py
```diff
@@ -0,0 +1,399 @@
...
+from spacy.pipeline.coordinationruler import split_noun_coordination
+from spacy.tokens import Doc
+
+
+@pytest.fixture
+def nlp():
+    return spacy.blank("en")
+
+
+### CONSTRUCTION CASES ###
```
india-kerle, line 15: doesn't account for i.e. the "water and power meters and electrical sockets"

PR #13378, opened 2024-03-15
comprehensive french tokenizer without exceptions list

thjbdvlt:
The current french tokenizer doesn't handle hyphens and apostrophes in an accurate, comprehensive and optimal way. It uses a gigantic (15600) list of words with hyphen that must not be split on the hyphen. This list is not only huge (full of village names such as _Minaucourt-le-Mesnil-lès-Hurlus_, or _Beaujeu-Saint-Vallier-Pierrejux-et-Quitteur_), but also very incomplete. This list has no chance to become ever exhaustive, because common nouns and proper names in french with a hyphen that must not be split are virtually infinite: the hyphen is called in french _trait d'union_ (union trait), it unifies, it joins separate words into one semantic word (and token). There is cases where a hyphen must split a substring into two words, and these cases are easily handled with a simple regex, because unlike the infinite exceptions, they are not very diverse: a) verb-subject inversion where subject is pronominalized; b) verb-object form where object is pronominalized; for a total of 21 words. This current pull requests ask to replace the tokenizer exceptions by a new 're_infixes' function, that easily handles each of the 15600 exceptions, and many more. It reverses the rule-exception relation: rule = keep as one token the words containing a hyphen; exception = split words containing a hyphen if the hyphen is followed by one of the registered word (pronominalized subject/object).


thjbdvlt:
i apologize for all these failed test!! it's the first time i contribute to a project (i'm not a programmer: i study french literature) and i finally understood that i could do it myself: now it doesnt fail anymore. sorry again and thank's for having look at my pull request :)

PR #13400, opened 2024-03-27
Fix use_gold_ents behaviour for EntityLinker

svlandeg:
## Description
The `use_gold_ents` flag was [introduced](https://github.com/explosion/spaCy/pull/9669) to allow the `entity_linker` to train on gold entities, even if there's no (annotating) NER component in the pipeline.

I think this behaviour was buggy because of a few reasons:
1. In `initialize()`, NER "predictions" from `eg.reference` were added to `eg.predicted` for the first 10 examples, and never cleaned up/restored afterwards.
2. In `update()`, this transfer happened on all examples, but here the ents were "restored" before calling the loss function. In theory, this should have prevented the EL to learn anything at all, except that in the corresponding unit test, this bug got masked by bug 1, which resulted in a few spurious annotations on the first 10 documents
3.  Because of how a spaCy pipeline works internally, the scoring could never work out of the box, because `Language.evaluate()` calls `pipe()` on the predicted docs, which won't have entities if there is no (annotating) NER in the pipeline.

To test some of this behaviour, I used different configs with the EL Emerson example, cf https://github.com/explosion/projects/pull/207. The "EL only" config would produce all-zero lines with `master`:
```
E    #       LOSS ENTIT...  NEL_MICRO_F  NEL_MICRO_R  NEL_MICRO_P  SCORE
---  ------  -------------  -----------  -----------  -----------  ------
  0       0           0.00         0.00         0.00         0.00    0.00
 33     200           0.00         0.00         0.00         0.00    0.00
 73     400           0.00         0.00         0.00         0.00    0.00
123     600           0.00         0.00         0.00         0.00    0.00
```

Then it would produce actual loss scores after fixing 1 and 2:
```
E    #       LOSS ENTIT...  NEL_MICRO_F  NEL_MICRO_R  NEL_MICRO_P  SCORE
---  ------  -------------  -----------  -----------  -----------  ------
  0       0           3.30         0.00         0.00         0.00    0.00
 33     200          55.77         0.00         0.00         0.00    0.00
 73     400           3.93         0.00         0.00         0.00    0.00
123     600           1.99         0.00         0.00         0.00    0.00
```

And finally, after fixing 3, it would give actual scores:
```
E    #       LOSS ENTIT...  NEL_MICRO_F  NEL_MICRO_R  NEL_MICRO_P  SCORE
---  ------  -------------  -----------  -----------  -----------  ------
  0       0           3.30        33.33        33.33        33.33    0.33
 33     200          57.55        83.33        83.33        83.33    0.83
 74     400           4.25        83.33        83.33        83.33    0.83
124     600           1.95        83.33        83.33        83.33    0.83

```

### Types of change
bug fixes & enhancement

## Checklist


spacy/pipeline/entity_linker.py
```diff
@@ -235,14 +234,42 @@ def __init__(
...
+            else:
+                examples = self._augment_examples(examples)
+                docs = self.pipe(
+                    (eg.predicted for eg in examples),
+                )
+                for eg, doc in zip(examples, docs):
+                    eg.predicted = doc
+                return scorer(examples, **kwargs)
+
+        self.scorer = _score_augmented
```
svlandeg, line 258: This whole bit is surely pretty hacky, but considering bug 3 as explained in the PR, I don't see a better option other than changing the entire mechanism how evaluation/scoring of a pipeline works...

spacy/pipeline/entity_linker.py
```diff
@@ -235,14 +234,42 @@ def __init__(
...
+
+    def _augment_examples(self, examples: Iterable[Example]) -> Iterable[Example]:
+        """If use_gold_ents is true, set the gold entities to (a copy of) eg.predicted."""
+        if not self.use_gold_ents:
+            return examples
+
+        new_examples = []
+        for eg in examples:
+            ents, _ = eg.get_aligned_ents_and_ner()
+            new_eg = eg.copy()
```
svlandeg, line 268: Making a copy here feels safest? Not 100% about all the possible interactions with all other components in the pipeline, before or after, annotated or not, and frozen or not...

spacy/pipeline/entity_linker.py
```diff
@@ -235,14 +234,42 @@ def __init__(
...
         self.kb = generate_empty_kb(self.vocab, entity_vector_length)
-        self.scorer = scorer
         self.use_gold_ents = use_gold_ents
         self.candidates_batch_size = candidates_batch_size
         self.threshold = threshold
 
         if candidates_batch_size < 1:
             raise ValueError(Errors.E1044)
 
+        def _score_augmented(examples, **kwargs):
```
rmitsch, line 244:
```suggestion
        def _score_augmented(examples: Iterable[Example], **kwargs):
```

spacy/pipeline/entity_linker.py
rmitsch: Agreed, this is not really satisfying. The workaround makes sense in this context though.

spacy/pipeline/entity_linker.py
rmitsch: Hm, do we manipulate examples in other components? I'm also unsure about this. Either way :+1: for copying it.

spacy/pipeline/entity_linker.py
```diff
@@ -235,14 +234,42 @@ def __init__(
...
+                docs = self.pipe(
+                    (eg.predicted for eg in examples),
+                )
+                for eg, doc in zip(examples, docs):
+                    eg.predicted = doc
+                return scorer(examples, **kwargs)
+
+        self.scorer = _score_augmented
+
+    def _augment_examples(self, examples: Iterable[Example]) -> Iterable[Example]:
```
rmitsch, line 260: I'm not very happy with the method name - this doesn't "augment" in the usual ML "data augmentation" sense, so this may be confusing. Maybe something like `_adopt_[predicted_]examples` or `_set_gold_ents`?

spacy/pipeline/entity_linker.py
```diff
@@ -284,11 +311,9 @@ def initialize(
         nO = self.kb.entity_vector_length
         doc_sample = []
         vector_sample = []
-        for eg in islice(get_examples(), 10):
+        examples = self._augment_examples(islice(get_examples(), 10))
```
rmitsch, line 314: Why exactly 10 here? I don't remember the specifics about this anymore.

spacy/pipeline/entity_linker.py
```diff
@@ -354,41 +379,29 @@ def update(
...
         )
         bp_context(d_scores)
         if sgd is not None:
             self.finish_update(sgd)
         losses[self.name] += loss
+
         return losses
 
     def get_loss(self, examples: Iterable[Example], sentence_encodings: Floats2d):
+        """Here, we assume that get_loss is called with augmented examples if need be"""
```
rmitsch, line 404: Params in docstring would be great for public methods :grin:

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -807,6 +809,107 @@ def create_kb(vocab):
...
     assert_equal(batch_deps_1, no_batch_deps)
 
+    eval = nlp.evaluate(train_examples)
+    assert "nel_macro_p" in eval
+    assert "nel_macro_r" in eval
+    assert "nel_macro_f" in eval
+    assert "nel_micro_p" in eval
+    assert "nel_micro_r" in eval
+    assert "nel_micro_f" in eval
+    assert "nel_f_per_type" in eval
```
rmitsch, line 819: Formatting in suggestion TBD. Ignore if you separated on purpose for easier debugging.
```suggestion
    assert all([metric in eval for metric in ("nel_macro_p", "nel_macro_r", "nel_macro_f", "nel_micro_p", "nel_micro_r", "nel_micro_f", "nel_f_per_type")]
```

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -807,6 +809,107 @@ def create_kb(vocab):
...
+            alias="Russ Cochran",
+            entities=["Q2146908", "Q7381115"],
+            probabilities=[0.5, 0.5],
+        )
+        return mykb
+
+    # Create the NER and EL components and add them to the pipeline
+    ner = nlp.add_pipe("ner", first=True)
+    entity_linker = nlp.add_pipe(
+        "entity_linker", last=True, config={"use_gold_ents": False}
```
rmitsch, line 855: Why `"use_gold_ents": False` here?

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -807,6 +809,107 @@ def create_kb(vocab):
...
+    test_text = "Russ Cochran captured his first major title with his son as caddie."
+    doc = nlp(test_text)
+    ents = doc.ents
+    assert len(ents) == 1
+    assert ents[0].text == "Russ Cochran"
+    assert ents[0].label_ == "PERSON"
+    assert ents[0].kb_id_ != "NIL"
+
+    # TODO: below assert is still flaky - EL doesn't properly overfit quite yet
+    # assert ents[0].kb_id_ == "Q2146908"
```
rmitsch, line 886: TBD in this PR?

spacy/tests/pipeline/test_entity_linker.py
```diff
@@ -807,6 +809,107 @@ def create_kb(vocab):
...
+        assert ents2[0].kb_id_ != "NIL"
+
+    eval = nlp.evaluate(train_examples)
+    assert "nel_macro_f" in eval
+    assert "nel_micro_f" in eval
+    assert "ents_f" in eval
+    assert "nel_f_per_type" in eval
+    assert "ents_per_type" in eval
+    assert "PERSON" in eval["nel_f_per_type"]
+    assert "PERSON" in eval["ents_per_type"]
```
rmitsch, line 907: Formatting in suggestion TBD. Ignore if you separated on purpose for easier debugging.
```suggestion
    assert all([metric in eval for metric in ("nel_macro_p", "nel_macro_r", "nel_macro_f", "nel_micro_p", "nel_micro_r", "nel_micro_f", "nel_f_per_type")]
```

rmitsch (commented):
Great spot :eyes: Can you elaborate on the cleaning up/restoration from reason 1.? Not sure what you mean by that.

spacy/tests/pipeline/test_entity_linker.py
svlandeg: In unit tests I prefer to test one thing per line so it's clear what fails when it does ;-)

spacy/tests/pipeline/test_entity_linker.py
svlandeg: This is a different issue so a separate PR would be prefered

spacy/tests/pipeline/test_entity_linker.py
svlandeg: Because there is an `ner` component in this pipeline

spacy/pipeline/entity_linker.py
svlandeg: We do this everywhere just to have some data for `model.initialize()` so that Thinc can infer the correct `nI`/`nO` dimensions.

spacy/pipeline/entity_linker.py
svlandeg: True! I renamed it to `_ensure_ents` as the function doesn't set any ents if `use_gold_ents` is `False`.

svlandeg:
> Can you elaborate on the cleaning up/restoration from reason 1.?

When the pipeline gets initialized, all the individual components their `initialize()` method is being called on a set of `Example` objects. The entity linker was changing these examples by setting gold entities, and not cleaning up afterwards.

Issue #13405, opened 2024-04-01
Code example discrepancy for `Span.lemma_` in API docs


schorfma:
Hello spaCy team,

I found a small discrepancy in the documentation.

---

The attribute `lemma_` for a `Span` is described as follows in the API docs:
> The span's lemma. Equivalent to `"".join(token.text_with_ws for token in span)`.

### Suggested Change

The equivalent code example should not contain `token.text_with_ws` in the comprehension, but `token.lemma_ + token.whitespace_`:

```diff
- | `lemma_`       | The span's lemma. Equivalent to `"".join(token.text_with_ws for token in span)`. ~~str~~                                      |
+ | `lemma_`       | The span's lemma. Equivalent to `"".join(token.lemma_ + token.whitespace_ for token in span).strip()`. ~~str~~                |
````

## Which page or section is this issue related to?

* Docs: [API Docs / `Span`](https://spacy.io/api/span#attributes)
* Code: [`website/docs/api/span.mdx` / Line 564](https://github.com/explosion/spaCy/blob/5f8a398/website/docs/api/span.mdx?plain=1#L564)

schorfma:
I provided corrections for this issue and two further small errors in the PR #13436.

Greetings
Martin

PR #13436, opened 2024-04-12
Correct API docs for Span.lemma_, Vocab.to_bytes and Vectors.__init__

schorfma:
## Description

This pull request combines three small changes to the API docs to fix errors, I have stumbled upon while carefully reading on the API details.

### Types of change

This correction for the API docs addresses:

1. Wrong code example for the property `Span.lemma_` (First reported in #13405, Addressed in commit b334a7a)
2. Wrong return type (`Vocab` instead of `bytes`) for the method `Vocab.to_bytes` (Addressed in commit ce88925)
3. Slightly incorrect wording (`Sequence` instead of `Iterable`) in constructor method `Vectors.__init__` (Addressed in commit d9c368d)

## Checklist

- [ ] ~~I ran the tests, and all new and existing tests passed.~~


honnibal:
Thanks!

PR #13466, opened 2024-04-28
Fix typos in docs

strickvl:
Fixed typos in Spacy docs.

Would be happy to add in a GH CI actions that automates this and catches these pre-emptively as part of this PR, but thought maybe that's a bit invasive into your codebase so just starting with the typos fixes first.


svlandeg (commented):
Hi, thanks a lot for this PR!

It looks like the CI is failing with an unrelated error, which we'll look into separately.
Otherwise I think the only thing left to do is run `prettier` on the docs that have changed, cf [this setup guide](https://github.com/explosion/spaCy/tree/2e2334632beb0e91abc1d7820a0471a10af61489/website#setup-and-installation).

svlandeg (approved):
Thanks a lot for fixing all these typos! Super appreciated. PR looks good to me.

I've also fixed the CI with this PR in the meantime: https://github.com/explosion/spaCy/pull/13469

PR #13510, opened 2024-05-28
add Tibetan support

ykyogoku:
## Description

Add Tibetan language class.

### Types of change

Enhancement

## Checklist


ykyogoku:
It’s been a while since I requested the addition of the Tibetan framework. Is it still in progress? Could you update me on how things are going?

By the way, I successfully trained a POS-Tagger using spaCy and would like to add it to the framework.

honnibal:
Sorry for the delay on this. I haven't been able to review it properly but I'm happy to merge it as an inclusion. Thanks for your work!

Regarding the models, we need to incorporate the training into our pipeline, because we need to be able to retrain models as we release new versions. So we can't accept binary artifacts as user contributions. I'm not sure when I'll have time to do the work on this, but there's a couple of others to add so I might be able to do them together later this year.

PR #13613, opened 2024-09-02
Fix typing errors with set_extenion methods

mribbons:
## Description
eg:
```
Argument of type "(span: Span, tag: str) -> str" cannot be assigned to parameter "method" of type "SpanMethod | None" in function "set_extension"
  Type "(span: Span, tag: str) -> str" is incompatible with type "SpanMethod | None"
    "function" is incompatible with protocol "SpanMethod"
      "__call__" is not present
    "function" is incompatible with "None"PylancereportArgumentType
```

Rationale:
`Callable` only accepts a single parameter, optional parameters don't work ... isn't recognized as variadic arguments (... not valid here) SomeMethod(Protocol) __call__ approach doesn't support mapping non variadic arguments to variadic arguments. Therefore, to not break the currently documented interface, implement a union of Callable overloads.

### Types of change
Typing fix

## Checklist

### DIFF
--- a/spacy/tokens/doc.pyi
+++ b/spacy/tokens/doc.pyi
@@ -28,8 +28,12 @@ from .underscore import Underscore
 
 DOCBIN_ALL_ATTRS: Tuple[str, ...]
 
-class DocMethod(Protocol):
-    def __call__(self: Doc, *args: Any, **kwargs: Any) -> Any: ...  # type: ignore[misc]
+DocMethod = Union[
+ Callable[[Doc], Any],
+ Callable[[Doc, Any], Any],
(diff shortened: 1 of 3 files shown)


honnibal:
The current way does indeed look broken, but if we can't make it correct I'd rather just declare them as `Callable`? The union up to an arbitrary number isn't very satisfying.

mribbons:
Hey @honnibal,
> The union up to an arbitrary number isn't very satisfying.

I agree, I just didn't want to break the current API.

Turns out the docs in master use `Callable` anyway. 

PR updated.

PR #13627, opened 2024-09-18
universe-pipeline-solipCysme-french

thjbdvlt:
Add a pipeline ([solipcysme](https://github.com/thjbdvlt/solipCysme)) for french language, designed for the analysis of fictions and first person point of view texts, primarly focused on personal pronouns. It aims to fill a gap in french models: the currently available pipelines for french are trained on 'news' data, a special kind of data that lack of personal pronouns (as an example: 'tu', which means 'you', is not in the data), and of some verb modes (e.g. imperative) very usual in fictions.

PS: The pipeline integrates two small projects that are already in the spaCy universe (_presque_, a normalizer for french, and _quelquhui_, a tokenizer for french). Maybe i should remove these projects, so i only have one there?

### DIFF
--- a/website/meta/universe.json
+++ b/website/meta/universe.json
@@ -277,6 +277,47 @@
             ]
         },
         {
+            "id": "solipcysme",
+            "title": "solipCysme",
+            "slogan": "spaCy pipeline for french fictions and first person point of view texts.",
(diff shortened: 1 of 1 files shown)


ines:
Thank you for all the universe contributions! 💙 And if the two other projects can also be used independently, it's fine to keep them separate – but if you want people to just use the new project instead, feel free to submit another PR that removes the others!

PR #13646, opened 2024-10-01
[WIP] Investigate windows CI failures

honnibal:
Need to investigate this


danieldk:
```
Windows fatal exception: access violation
```

might come from BLIS. The upstream OOB writes were never fully solved in upstream 0.9.x BLIS, which is why we had to revert to 0.7.x.

PR #13655, opened 2024-10-10
Sentiment analysis using logistic regression (only spacy)

samhithamuvva:
## Checklist


samhithamuvva:
The provided functionality will perform Sentiment Analysis Using spaCy (Without using sklearn). It can classify the text into positive and negative (binary classification) using a logistic regression-based text classification model using only spaCy, without the need for external libraries like sklearn. The usage is specified in the Readme.md documentation as well.

Issue #13701, opened 2024-11-30
reference to obsolete thinc backend linalg


envolution:
```
ml/parser_model.pyx
line8: from thinc.backends.linalg cimport Vec, VecVec
```
https://github.com/explosion/thinc/pull/742

was obsoleted a couple of years ago, which is causing some downstream issues:
```
  File "/usr/lib/python3.12/site-packages/auralis/models/xttsv2/config/tokenizer.py", line 12, in <module>
    from spacy.lang.ar import Arabic
  File "/usr/lib/python3.12/site-packages/spacy/__init__.py", line 13, in <module>
    from . import pipeline  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/spacy/pipeline/__init__.py", line 2, in <module>
    from .dep_parser import DependencyParser
  File "spacy/pipeline/dep_parser.pyx", line 1, in init spacy.pipeline.dep_parser
  File "spacy/pipeline/transition_parser.pyx", line 1, in init spacy.pipeline.transition_parser
  File "/usr/lib/python3.12/site-packages/spacy/ml/__init__.py", line 2, in <module>
    from .models import *  # noqa: F401, F403
    ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/spacy/ml/models/__init__.py", line 3, in <module>
    from .parser import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/spacy/ml/models/parser.py", line 11, in <module>
    from ..tb_framework import TransitionModel
  File "/usr/lib/python3.12/site-packages/spacy/ml/tb_framework.py", line 4, in <module>
    from .parser_model import ParserStepModel
  File "spacy/ml/parser_model.pyx", line 1, in init spacy.ml.parser_model
ModuleNotFoundError: No module named 'thinc.backends.linalg'
```
## Your Environment
Arch linux
Name: spacy
Version: 3.8.2
Summary: Industrial-strength Natural Language Processing (NLP) in Python
Home-page: https://spacy.io
Author: Explosion
Author-email: contact@explosion.ai
License: MIT
Location: /usr/lib/python3.12/site-packages
[...]

PR #13702, opened 2024-11-30
remove linalg backend as per #13701

envolution:
https://github.com/explosion/spaCy/issues/13701

I have not tested this due to some external issues, but the edits were compared against https://github.com/explosion/spaCy/pull/11292/files#diff-f19f042dd7350fc29ab39ab36483a189e482a86592f20bb5b07f0d085306bd9d


envolution:
I seem to be solving issues in unsupported (as yet) thinc - none of this was necessary with think 8.3.2

PR #13721, opened 2024-12-20
Lazily load cli with module __getattr__

hoodmane:
This is a reland of #12962. That PR broke accesses to `spacy.cli`:
```py
import spacy
spacy.cli # NameError
```
This avoids such a  problem by using module `__getattr__`.

### DIFF
--- a/spacy/__init__.py
+++ b/spacy/__init__.py
@@ -5,23 +5,40 @@ from typing import Any, Dict, Iterable, Union
 # set library-specific custom warning handling before doing anything else
 from .errors import setup_default_warnings
 
-setup_default_warnings()  # noqa: E402
+setup_default_warnings()
 
 # These are imported as part of the API
-from thinc.api import Config, prefer_gpu, require_cpu, require_gpu  # noqa: F401
+from thinc.api import Config, prefer_gpu, require_cpu, require_gpu
 
-from . import pipeline  # noqa: F401
-from . import util
-from .about import __version__  # noqa: F401
-from .cli.info import info  # noqa: F401
+from . import pipeline, util
+from .about import __version__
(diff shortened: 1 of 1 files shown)


hoodmane:
Well this does not work quite yet.

hoodmane:
Okay now it's better.

hoodmane:
Yes. And `__all__` is good practice but it's also needed to avoid dropping cli from the things imported by `from mod import *`

honnibal:
@nro-bot Sorry for the lack of attention on this. Adriane no longer works at Explosion (we're no longer venture-backed), and my time for support has been split over a lot of different things.

I'm now getting more on top of it, having just finished the support for Python 3.13.

Could you summarise a little more what the lazy loading is intending to do? Is it that you want to avoid the import of the pipeline at loading, so that you can avoid triggering stuff that doesn't work?

An issue with the lazy-loading is that spaCy heavily depends on entry-points. These end up importing everything to get decorators to run as side-effects. I've refactored this in the patch for 3.13, so we no longer rely on these import side-effects to run decorators. That said, the registration functions (which need to run to do anything much) still need to import pretty much everything. So it's not clear to me what the net impact for your use-case would be.

hoodmane:
> Could you sumarise a little more what the lazy loading is intending to do? Is it that you want to avoid the import of the pipeline at loading, so that you can avoid triggering stuff that doesn't work?

Yeah that's what we're after.

> it's not clear to me what the net impact for your use-case would be.

The test here works fine with this patch:
https://github.com/pyodide/pyodide/blob/144c3de286ad41575b7d7fcc10438465a1e093fe/packages/spacy/test_spacy.py

PR #13749, opened 2025-02-11
docs: fix README.md for compatible Python versions

bolinocroustibat:
## Description
SpaCy 3.8 is not compatible with Python 3.13 yet but the doc says otherwise, which can lead for users to loose time on that if they are ill-informed.
See https://github.com/explosion/spaCy/issues/13658

### Types of change
Update README.md to fix compatible Python versions.


bolinocroustibat:
Any news on this?

honnibal:
Thanks, I've been working on Python 3.13 support but it's actually quite difficult. Definitely best to update the docs.

PR #13764, opened 2025-03-06
Support Cython 3

davispuh:
## Description

Cython 0.x doesn't support Python 3.13 so for that need to update to Cython 3

### Types of change

Update to build with Cython 3

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "setuptools",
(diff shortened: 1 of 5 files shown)


honnibal:
I've been working on this and it's actually very difficult. Cython 3 forces `from __future__ import annotations`, which messes with Pydantic. Separately I'm also trying to get Pydantic support for Pydantic v2 and there's a lot of problems.

PR #13787, opened 2025-04-04
Fix: Update display import for compatibility with latest IPython

rnihesh:
**Fix: Update display import for compatibility with latest IPython**

---

## Description

This PR updates the import statement in `displacy.render()` to fix an `ImportError` caused by breaking changes in IPython 9+.

Replaced:

```python
from IPython.core.display import HTML, display
```

With:

```python
from IPython.display import HTML, display
```

The `display` function is no longer available in `IPython.core.display` in IPython 9 and above. This change ensures `displacy.render()` works correctly in Jupyter and notebook environments using newer versions of IPython.

**Tested in:**
- Python 3.12.9 (venv)
- IPython 9.0.2
- JupyterLab
- Confirmed `displacy.render()` now works without error.

---

### Types of change

- [x] Bug fix
- [ ] Enhancement
- [ ] New feature
- [ ] Documentation update

---

## Checklist

---

Let me know if you want to attach an issue number or want help creating tests for the fix!

### DIFF
--- a/spacy/displacy/__init__.py
+++ b/spacy/displacy/__init__.py
@@ -66,7 +66,7 @@ def render(
     if jupyter or (jupyter is None and is_in_jupyter()):
         # return HTML rendered by IPython display()
         # See #4840 for details on span wrapper to disable mathjax
-        from IPython.core.display import HTML, display
+        from IPython.display import HTML, display
 
         return display(HTML('<span class="tex2jax_ignore">{}</span>'.format(html)))
     return html


majorgilles:
also ran into this issue, was about to submit my own PR

ines:
Thanks a lot and sorry for only getting to this now!

Just wondering if we should include a `try`/`except` here for backwards compatibility, even though it's not very pretty. Because I wouldn't want users on older IPython versions to run into problems going forward. This is probably the one thing I would add and then we can ship this (happy to do it before we release or you can also make an update) 🙂

rnihesh:
Yup working on that

celestinoxp:
@ines ping...

wildfluss:
this is very annoying especially when working with prodigy :/ had to patch version which prodigy pulls etc

it176131:
Temporary solution until this is merged:
```shell
$ pip install -U "ipython<9"
```

PR #13788, opened 2025-04-04
Support Cython 3

honnibal:
Need to support Cython 3 for Python 3.13


h-vetinari:
Note that cython 3 supports tracing for python 3.13 again (c.f. https://github.com/explosion/spaCy/commit/1adf79414e14f8f45a64c7cdb6cb098b4cf1f46f), see [docs](https://docs.cython.org/en/latest/src/userguide/source_files_and_compilation.html).

honnibal:
@h-vetinari Thanks I'll fix that. It's currently off in the Python 3.13 support patch that's been released.

PR #13816, opened 2025-05-09
fix: match hyphenated words to lemmas in index_table (e.g. "co-author…

d0ngw:
Previously, words with hyphens such as "co-authored" could not be matched correctly against the index_table, even though their lemma forms like "co-author" existed.
This fix normalizes such words by lemmatizing and aligning them with their base forms, ensuring accurate lookup and processing.

Example:

Input: "co-authored"
Lemma: "co-author"
Result: Now correctly matched in index_table

This improves the robustness of word matching, especially for common hyphenated expressions.


honnibal:
Thanks!

d0ngw:
> Thanks!

You are welcome! 😄

PR #13822, opened 2025-05-16
fix: DeprecationWarning: Importing 'parser.split_arg_string' is deprecated, it will only be available in 'shell_completion' in Click 9.0.

otakutyrant:
This appears as Click release 8.2.0: https://click.palletsprojects.com/en/stable/changes

But I also noticed that `requirements.txt` does not include `click`. Any idea?

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -20,7 +20,7 @@ from typing import (
 import srsly
 import typer
 from click import NoSuchOption
-from click.parser import split_arg_string
+from click.shell_completion import split_arg_string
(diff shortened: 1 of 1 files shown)


honnibal:
We depend on click via typer. I expect they'll fix the problem upstream shortly, if they don't we can definitely have a look.

zhemaituk:
Same warning is generated for a dependency as well: https://github.com/explosion/weasel/blob/main/weasel/util/config.py#L8, might need the same fix.

weezymatt:
Hi you may close your PR the same fix #13843 was merged recently.

PR #13840, opened 2025-06-24
Fix LLM EntityLinker task factory typo

cdjameson:
## Description
Change the factory string for the EntityLinker to "llm_entitylinker" from "llm_entity_linker" so that it works, as per https://github.com/explosion/spacy-llm/blob/main/spacy_llm/tasks/__init__.py#L33.
This appears to be the only place it is documented.

### Types of change
Documentation change to reflect the API as currently implemented.

## Checklist
~~- [ ] I ran the tests, and all new and existing tests passed.~~


themavik (commented):
@/run/media/zoom/STORAGE/Github/._rb/s3.txt

PR #13922, opened 2026-02-11
Switch dependency back from `typer-slim` to `typer`

svlandeg:
Hi, long time no see! 😀 
Here with my Typer-maintainer-hat on.

## Description

Recently, spaCy [switched](https://github.com/explosion/spaCy/pull/13759) from using `typer` to `typer-slim`. We used to offer `typer-slim` as a more light-weight package, which didn't depend on `rich` and `shellingham`. Unfortuntately, the way it was set up meant that we ran into issues with package managers (e.g. https://github.com/fastapi/typer/issues/1503). We ran a user survey and eventually decided to drop support for `typer-slim` altogether (since [0.22.0](https://github.com/fastapi/typer/releases/tag/0.22.0)). Going forward, it's better to just use `typer` as dependency.

What this means for this repo:
- `shellingham` and `rich` will be added as indirect requirement, adding about 12MB extra.
- With `rich` present in the environment, any Typer app will automatically set `rich_markup_mode` to `"rich"` by default (it was `None` when Rich wasn't installed). This can be overwritten when creating a `Typer()` app by setting `rich_markup_mode` to `None`. Alternatively, since Typer 0.22.0 there is a new env var `TYPER_USE_RICH` that you can set to `False` to disable Rich entirely everywhere and always. I've gone ahead and went for the first option by setting `rich_markup_mode=None` in the main `app`. Personally, I do like the Rich formatting better, so you can pick which you like most as spaCy has had different behaviour in the past anyway (before PR #13759 it would be formatted with Rich).

I'll submit a PR to `weasel` as well.

Let me know if there's any concerns/questions!

### Types of change
External dependency change

## Checklist

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -68,7 +68,7 @@ INIT_HELP = """Commands for initializing configs and pipeline packages."""
 Arg = typer.Argument
 Opt = typer.Option
 
-app = typer.Typer(name=NAME, help=HELP)
+app = typer.Typer(name=NAME, help=HELP, rich_markup_mode=None)
 benchmark_cli = typer.Typer(name="benchmark", help=BENCHMARK_HELP, no_args_is_help=True)
 debug_cli = typer.Typer(name="debug", help=DEBUG_HELP, no_args_is_help=True)
 init_cli = typer.Typer(name="init", help=INIT_HELP, no_args_is_help=True)


honnibal:
Hey sorry for missing this before! I do prefer the non-rich output, and I haven't been doing much LLM-powered spaCy stuff but I think the less verbose output is probably better there too.

PR #13947, opened 2026-03-28
fix: check pip module availability instead of PATH binary in download

### LINKED ISSUE #13946: spacy download fails in venvs with pip3 but no pip binary (regression in 3.8.12)
## The problem

`spacy download en_core_web_lg` fails with "No package installer found" in venvs where `pip3` exists but bare `pip` doesn't. Regression in v3.8.12.

```
✘ No package installer found
spaCy requires either pip or uv to download models. Please install one of
them and try again.
```

## Reproduction

```bash
# Debian/Ubuntu, or any ensurepip setup that creates pip3 but not pip
python3 -m venv testenv
source testenv/bin/activate
pip3 install spacy
python -m spacy download en_core_web_lg  # fails
python -m pip install some-package       # works fine
```

## Root cause

`_get_pip_install_cmd()` (added in c6c78d6c33) checks `shutil.which("pip")` — a PATH binary search — then returns `[sys.executable, "-m", "pip", "install"]`, which invokes pip as a module. The binary check and the module invocation test different things. When only `pip3` is on PATH, the binary check fails even though the module invocation would succeed.

The pre-3.8.12 code was correct: it unconditionally used `sys.executable -m pip`.

## Affected environments

- Debian/Ubuntu venvs (`ensurepip` creates `pip3`, not `pip`)
- Docker containers with Debian/Ubuntu base images
- Any setup where pip is installed as a module but the bare `pip` binary isn't on PATH

Not affected: macOS, conda, pyenv, Windows (all create both `pip` and `pip3`).

## Fix

Replace the PATH binary check with a module availability check — the actual precondition for `sys.executable -m pip` to work:

```python
# Before
if shutil.which("pip"):

# After
if importlib.util.find_spec("pip") is not None:
```

The `uv` branch stays as-is since `uv` is a standalone binary.

PR incoming.

- **spaCy version:** 3.8.12+
- **Platform:** Linux (Debian/Ubuntu)
- **Python version:** 3.10+

taylorsatula:
Fixes #13946.

`_get_pip_install_cmd()` checks `shutil.which("pip")` then returns `[sys.executable, "-m", "pip", "install"]`. The binary check and the module invocation test different things. In venvs where `ensurepip` creates `pip3` but not bare `pip` (standard on Debian/Ubuntu), the binary check fails even though `python -m pip` works.

Replaces `shutil.which("pip")` with `importlib.util.find_spec("pip")` — the actual precondition for `sys.executable -m pip`. The `uv` branch is unchanged since `uv` is a standalone binary where `shutil.which` is the right check.

No new dependencies (stdlib). No change in behavior for environments where both `pip` and `pip3` exist.

### DIFF
--- a/spacy/cli/download.py
+++ b/spacy/cli/download.py
@@ -1,3 +1,4 @@
+import importlib.util
 import shutil
 import sys
 from typing import Optional, Sequence
@@ -189,7 +190,7 @@ def download_model(
 
 
 def _get_pip_install_cmd() -> list:
-    if shutil.which("pip"):
+    if importlib.util.find_spec("pip") is not None:
         return [sys.executable, "-m", "pip", "install"]
     elif shutil.which("uv"):
         return ["uv", "pip", "install"]


honnibal:
Thanks! I should've paid more attention on this.

taylorsatula:
Absolutely! That is the power of open source! I encountered it and thought I might as well fix it if I'm in there. Have a great week!

PR #13951, opened 2026-03-31
Fix #13950: tokenizer issue: special spans cause every subsequent span to not be cached

### LINKED ISSUE #13950: Performance/caching issue: tokenizer fails to reset has_special flag after encountering special span, effectively disabling caching

```python
nlp = English()
doc = nlp("I can't believe you have done this")
```

"can't" is a tokenizer exception because of the funky contraction (yay english)

https://github.com/explosion/spaCy/blob/0069cf99b60023363ba5f47404819f59c8a39b3e/spacy/lang/en/tokenizer_exceptions.py#L233

Spans that contain these exceptions are marked as `has_special`
declared here:
https://github.com/explosion/spaCy/blob/0069cf99b60023363ba5f47404819f59c8a39b3e/spacy/tokenizer.pyx#L179
set here:
https://github.com/explosion/spaCy/blob/0069cf99b60023363ba5f47404819f59c8a39b3e/spacy/tokenizer.pyx#L375

And `has_special` spans are not cached:

https://github.com/explosion/spaCy/blob/0069cf99b60023363ba5f47404819f59c8a39b3e/spacy/tokenizer.pyx#L523

The problem is that `has_special`, once set to a nonzero value, is never reset. And this means that once the tokenizer encounters a special case, *every subsequent span is also marked as special, and none of them get cached, even if they should be*.

This has some fairly significant tokenizer performance implications. It should be much faster than it is. I'll put some benchmarking in my PR.

## Info about spaCy

- **spaCy version:** 3.8.14
- **Platform:** macOS-26.3.1-arm64-arm-64bit
- **Python version:** 3.12.4
- **Pipelines:** en_core_web_md (3.8.0), en_core_web_sm (3.8.0)

jberg5:
## Description

Fixes https://github.com/explosion/spaCy/issues/13950

Basically once a span with special cases appeared (like "can't" or any nonstandard contraction), `has_special` would be set and remain set for any subsequent spans, even if they weren't special. These spans wouldn't get cached, and performance suffered as a result. Fortunately the fix is pretty straightforward, and we immediately see some decent tokenizer speedups.

To illustrate the worst case scenario: if you stick "can't" at the beginning of Huckleberry Finn, it takes 691ms to tokenize on my macbook pro. Reusing the same tokenizer on the same text is another 637ms. After the fix, the first pass takes 230ms, and reusing (everything cached now) takes 87ms. So a **3x speedup on the cold run, and a 7x speedup on the warm run**.

```python
text = open("/tmp/huck_finn.txt").read()
tok = spacy.blank("en").tokenizer
doc = tok(text)
```

Important note: in the average case of tokenizing a diverse, real-world corpora (like a bunch of tweets or something), the impact is more modest because eventually most words will get cached because they appear in a document where they don't follow a special span.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/spacy/tokenizer.pxd
+++ b/spacy/tokenizer.pxd
@@ -12,7 +12,7 @@ from .vocab cimport LexemesOrTokens, Vocab, _Cached
 
 cdef class Tokenizer:
     cdef Pool mem
-    cdef PreshMap _cache
+    cdef readonly PreshMap _cache  # readonly so tests can check state
     cdef PreshMap _specials
     cdef readonly Vocab vocab
 
--- a/spacy/tests/tokenizer/test_tokenizer.py
+++ b/spacy/tests/tokenizer/test_tokenizer.py
@@ -5,6 +5,7 @@ import pytest
 
 from spacy.lang.de import German
 from spacy.lang.en import English
+from spacy.strings import hash_string
 from spacy.symbols import ORTH
 from spacy.tokenizer import Tokenizer
 from spacy.tokens import Doc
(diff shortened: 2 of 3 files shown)


spacy/tokenizer.pxd
```diff
@@ -12,7 +12,7 @@ from .vocab cimport LexemesOrTokens, Vocab, _Cached
 
 cdef class Tokenizer:
     cdef Pool mem
-    cdef PreshMap _cache
+    cdef readonly PreshMap _cache  # readonly so tests can check state
```
jberg5, line 15: I don't love exposing private state so that tests can read it, but couldn't think of an easy alternative for testing the behavior I wanted to test. Open to suggestions!

PR #14000, opened 2026-07-19
Fix bare except: use specific exception types

koteshyelamati:
Replace bare `except:` clauses with specific exception types:

- `spacy/ml/callbacks.py`: `except Exception:` (guards `inspect.signature()` which can fail for Cython methods)
- `spacy/cli/_util.py`: `except Exception:` (guards `run_command("git --version")`)

Bare `except:` catches `BaseException` including `SystemExit` and `KeyboardInterrupt`.

### DIFF
--- a/spacy/cli/_util.py
+++ b/spacy/cli/_util.py
@@ -206,7 +206,7 @@ def get_git_version(
     """
     try:
         ret = run_command("git --version", capture=True)
-    except:
+    except Exception:
         raise RuntimeError(error)
     stdout = ret.stdout.strip()
     if not stdout or not stdout.startswith("git version"):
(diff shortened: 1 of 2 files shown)


rkfshakti (approved):
Clean fix. Replacing bare `except:` with `except Exception:` is the right approach — it prevents catching `SystemExit`/`KeyboardInterrupt` while still catching the expected runtime errors from `inspect.signature()` and `run_command`.

Both call sites are appropriate: `inspect.signature()` can raise `ValueError` for builtins, and `run_command` could raise various `OSError`/`subprocess` exceptions. Using `except Exception` is the minimal safe widening.

PR #14018, opened 2026-08-18
Fix scoring of explicit whitespace tokens

### LINKED ISSUE #13739: Inconsistent handling of whitespace tokens in Scorer.score_token_attr
I've been training tagger and parser on heavily augmented data and was surprised by poor performance (comparing to what I calculated manually on the test dataset). I narrowed it down to `Scorer.score_token_attr`. 

Sorry, I don't have a code example, but it's pretty straight-forward. In this function, in the gold dataset, all tokens (except for those with a missing attribute) are included in evaluation. However, in the predicted dataset, whitespace tokens are excluded. If the gold dataset contains whitespace tokens (which is true in my case), we are not comparing apples to apples here and inflate the error rate.

I just created my own scorer for now, but this behavior is kind of unexpected.

Let me know if you want me to change the behavior, and I will do a PR. My own fix to the function was to add `exclude_spaces` parameter, defaulting to the current behavior (i.e. `True`), and either include or exclude whitespace tokens in **both** datasets.

```
# line 240 of scorer.py
for gold_i, token in enumerate(gold_doc):
        value = getter(token, attr)
        if value not in missing_values:
            gold_tags.add((gold_i, getter(token, attr)))
        else:
            missing_indices.add(gold_i)
    pred_tags = set()
    for token in pred_doc:
        if token.orth_.isspace(): # HERE: excluding whitespace tokens
            continue
        if align.x2y.lengths[token.i] == 1:
            gold_i = align.x2y[token.i][0]
```

tandede:
## Description

Explicit whitespace tokens were skipped while collecting predicted token attributes and dependencies, but the corresponding reference loops still included them. As a result, even otherwise identical predicted and reference documents containing a whitespace token could receive `0.8` instead of `1.0` for token attributes and dependency scores.

This change applies the existing whitespace-token policy symmetrically to the reference side of:

- `score_token_attr`
- `score_token_attr_per_feat`
- `score_deps`

Alignment indices remain unchanged because both sides continue to use the original token positions. The regression deliberately assigns different tag, POS, morphology, and dependency values to the predicted whitespace token and verifies that all affected scores remain `1.0`, while the non-whitespace annotations still match.

Closes #13739.

### Types of change

Bug fix.

## Validation

- Before the fix, the regression example produced `0.8` for `tag_acc`, `pos_acc`, `morph_acc`, `morph_micro_f`, `dep_uas`, and `dep_las`; after the fix all six are `1.0`.
- `python -m pytest -q spacy/tests/test_scorer.py` — 11 passed
- `python -m pytest -q spacy` — 3,615 passed, 1,229 skipped, 24 xfailed, 7 xpassed, 5 subtests passed
- `python -m ruff format spacy --check` — 886 files already formatted
- `python -m ruff check spacy --select I` — passed
- `git diff --check` — passed

The repository-wide all-rule Ruff hook reports pre-existing complexity and legacy-style findings in `spacy/scorer.py`; the official CI format and import-order checks pass.

## Checklist

- [ ] I confirm that I have the right to submit this contribution under the project MIT license.
- [x] My changes do not require a documentation change.


tandede:
Hi maintainers, I’m following up on this PR in case it was missed. It fixes scoring for explicitly represented whitespace tokens and is ready for review. Please let me know if any changes or additional validation would be helpful. Thank you!

PR #14019, opened 2026-08-18
Fix missing cp314 wheels on PyPI (#13949)

### LINKED ISSUE #13949: Missing cp314 wheels for v3.8.14 on PyPI
https://github.com/explosion/spaCy/releases/tag/release-v3.8.14 lists wheels for Python 3.14 (cp314).

https://pypi.org/project/spacy/3.8.14/#files does not list these wheels.

(The 3.8.13 release still had them, see https://pypi.org/project/spacy/3.8.13/#files)

bertonha:
## Problem

Fixes #13949. The v3.8.14 and v3.8.15 GitHub releases each have **40 assets**, but PyPI received only **30**. The 10 missing files are exactly the last-uploaded batch:

- all 8 `cp314` wheels
- the `cp313 win_arm64` wheel
- the `sdist` (`.tar.gz`)

## Root cause

`publish_pypi.yml` downloaded release assets with `robinraju/release-downloader`, which reads only the **first page** of a release's assets. GitHub paginates the release-assets listing at **30 per page** (the endpoint returns a `Link: …page=2` header and only 30 items), so the tail was never fetched and never uploaded to PyPI — even though all 40 files are present on the GitHub release.

This regressed when the action was pinned from the floating `@v1` tag to a specific SHA (v1.12) for supply-chain hardening in 4216738. v3.8.13 was published *before* that pin and got all 40 files; v3.8.14 was the first release *after* it and lost the tail.

## Fix

- Replace the downloader action with `gh release download`, which paginates correctly (verified locally: it fetches all 40 assets, including every cp314 wheel and the sdist).
- Add a verification step that compares the number of downloaded files against the release's asset count and fails the job on mismatch, so a partial upload can never silently reach PyPI again.

## Note

This workflow only runs on new release publications, so the already-affected v3.8.14/v3.8.15 wheels won't appear on PyPI retroactively — those need a one-off manual `gh release download` + `twine upload` of the missing files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -23,9 +23,30 @@ jobs:
     # or, alternatively, upload to PyPI on every tag starting with 'v' (remove on: release above to use this)
     # if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     steps:
-      - uses: robinraju/release-downloader@daf26c55d821e836577a15f77d86ddc078948b05  # v1
-        with:
-          tag: ${{ github.event.release.tag_name }}
-          fileName: '*'
-          out-file-path: 'dist'
+      # Use `gh release download` rather than a third-party downloader action:
+      # GitHub paginates release assets at 30 per page, and a downloader that
+      # only reads the first page silently drops the tail. That's how the
+      # cp314 wheels + sdist went missing from PyPI for v3.8.14/v3.8.15 while
+      # being present on the GitHub release (#13949). `gh` paginates correctly.
+      - name: Download release assets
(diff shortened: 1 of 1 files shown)


grahamknapp-sofia:
Hello, I don't think this will work because the wheels on pypi are published using the private repo explosion/wheelmonger, not by the action in this repo - see the links on this page https://pypi.org/project/spacy/#spacy-3.8.15-cp313-cp313-musllinux_1_2_x86_64.whl - I think the error lies in that private repo not in the robinraju downloader.

